PR #10, opened 2023-02-11
Proof of concept: login a user via default browser

schacon:
This is a somewhat working example of how to use the API to log in via a default OS browser. It requests a temp login token on page load (we should probably do this just before redirecting to the login screen, but that's a little difficult because `window.open()` doesn't work in <script> in Tauri), replaces the href in a log in link, then starts polling to see if you've logged in. When it sees you have, it stops polling and shows you the response, which eventually we should store to disk and then you're *logged in*.

Currently it is hardcoded to my local box via ngrok, but we can change that to a variable or the staging server or something.

Don't actually merge this, I just wanted a place to put the more or less working code so we could discuss the approach.

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -315,6 +315,9 @@ fn main() {
                         projects_storage,
                     });
 
+                    let window = app.get_window("main").unwrap();
+                    window.open_devtools();
+
                     Ok(())
                 })
                 .plugin(sentry_plugin)
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -21,6 +21,9 @@
         "all": false,
         "startDragging": true
       },
+      "shell": {
+        "open": true
+      },
       "path": {
         "all": true
       },
--- a/src/routes/user/+page.svelte
+++ b/src/routes/user/+page.svelte
@@ -1,0 +1,64 @@
+<script>
+  let response = null;
+  let interval = null;
+  let token = null;
+
+  let login;
+  let user;
+
+  async function postData(url = "", data = {}) {
+    const response = await fetch(url, {
+      method: "POST", // *GET, POST, PUT, DELETE, etc.
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify(data), // body data type must match "Content-Type" header
+    });
+    return response.json(); // parses JSON response into native JavaScript objects
+  }
+
+  const fetchData = async (token) => {
+    try {
+      const res = await fetch(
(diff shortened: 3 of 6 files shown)


src-tauri/tauri.conf.json
```diff
@@ -21,6 +21,9 @@
         "all": false,
         "startDragging": true
       },
+      "shell": {
```
ngalaiko, line 24: why did you add this permission?

src-tauri/src/main.rs
```diff
@@ -315,6 +315,9 @@ fn main() {
                         projects_storage,
                     });
 
+                    let window = app.get_window("main").unwrap();
+                    window.open_devtools();
```
ngalaiko, line 319: nice! i was thinking looking into some sort of flag to only do this in dev, but we can keep it for now

src/routes/user/+page.svelte
```diff
@@ -0,0 +1,64 @@
+<script>
+  let response = null;
+  let interval = null;
+  let token = null;
+
+  let login;
+  let user;
```
ngalaiko, line 7: i understand this is a poc, but this should be stored on disk, right? also, the user info should be stored globally for the frontend app, so it's easy to access

src/routes/user/+page.svelte
```diff
@@ -0,0 +1,64 @@
...
+        console.log(user);
+        clearInterval(interval);
+      }
+    } catch (error) {
+      console.error(error);
+    }
+  };
+
+  const logIn = () => {
+    postData("https://chacons.eu.ngrok.io/api/login/token")
```
ngalaiko, line 37: is this on gitbutler.com?

ngalaiko (commented):
good as a poc, let me know if you need some help with a proper setup

src-tauri/tauri.conf.json
schacon: I was having problems getting it to open an external browser when you click on the link, I think I needed this to get it to work.

src-tauri/src/main.rs
schacon: Yeah, I needed it for debugging, a flag would be great.

src/routes/user/+page.svelte
schacon: right, this should be stored globally on disk as like `user.json` or something.

src/routes/user/+page.svelte
schacon: It is now, at `https://app.gitbutler.com/api`

PR #13, opened 2023-02-14
POC: Push to Git server after a flush

schacon:
This PR implements the concept of having the client try to push the gitbutler data to a server after a GB flush activity. 

In this PR, it pushes every time it flushes and the server and auth token are hard coded, since I'm not sure what the best way to get them out of the projects.json and user.json file is. I assume it's not too difficult as that's in Rust anyways, right?

We should probably make the function return immediately and no-op if the project is not registered with the server, or if the user is not logged in.

Before this is working, we would need to have a way to have the client create a project on the server and store the git url that is produced. You can do this with the user's auth token by doing a POST to `/api/projects` with a `name` parameter. The JSON returned should have a `git_url`. If we store this in `projects.json`, we should be able to pull it out easily here.

The user auth token should already exist in `user.json` if they are logged in.

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -5,6 +5,7 @@ mod fs;
 mod projects;
 mod sessions;
 mod storage;
+mod users;
 mod watchers;
 
 use deltas::Delta;
@@ -23,6 +24,7 @@ use watchers::WatcherCollection;
 struct AppState {
     watchers: WatcherCollection,
     projects_storage: projects::storage::Storage,
+    user_storage: users::storage::Storage,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -95,13 +97,14 @@ fn add_project<R: Runtime>(
     let project = projects::project::Project::from_path(path.to_string());
     if project.is_ok() {
         let project = project.unwrap();
+        let user = state.user_storage.get_user().unwrap();
         state.projects_storage.add_project(&project).map_err(|e| {
             log::error!("{}", e);
             Error {
                 message: "Failed to add project".to_string(),
             }
         })?;
-        watchers::watch(window, &state.watchers, &project).map_err(|e| {
+        watchers::watch(window, &state.watchers, &project, &user).map_err(|e| {
             log::error!("{}", e);
             Error {
                 message: "Failed to watch project".to_string(),
@@ -299,15 +302,25 @@ fn main() {
                 })
                 .setup(move |app| {
                     let resolver = app.path_resolver();
-                    let storage = Storage::new(&resolver);
-                    let projects_storage = projects::storage::Storage::new(storage);
+
+                    let pstorage = Storage::new(&resolver);
+                    let projects_storage = projects::storage::Storage::new(pstorage);
+
+                    let ustorage = Storage::new(&resolver);
+                    let user_storage = users::storage::Storage::new(ustorage);
+                    let user = user_storage.get_user().unwrap();
 
                     let watchers = watchers::WatcherCollection::default();
 
                     if let Ok(projects) = projects_storage.list_projects() {
                         for project in projects {
-                            watchers::watch(app.get_window("main").unwrap(), &watchers, &project)
-                                .map_err(|e| e.to_string())?;
+                            watchers::watch(
+                                app.get_window("main").unwrap(),
+                                &watchers,
+                                &project,
+                                &user,
+                            )
+                            .map_err(|e| e.to_string())?;
                         }
                     } else {
                         log::error!("Failed to list projects");
@@ -319,6 +332,7 @@ fn main() {
                     app.manage(AppState {
                         watchers,
                         projects_storage,
+                        user_storage,
                     });
 
                     Ok(())
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -119,6 +119,44 @@ fn add_project<R: Runtime>(
 }
 
 #[tauri::command]
+fn add_git_url(state: State<'_, AppState>, url: &str, path: &str) -> Result<(), Error> {
+    log::info!("ADD GIT URL: {} {}", url, path);
+
+    // add '.git/gb-url' to the path
+    let mut path = std::path::PathBuf::from(path);
+    path.push(".git");
+    path.push("gb-url");
+
+    // write contents to a file under the path
+    std::fs::write(path, url).map_err(|e| {
+        log::error!("{}", e);
+        Error {
+            message: "Failed to write git url".to_string(),
+        }
+    })?;
+
+    Ok(())
+}
+
+#[tauri::command]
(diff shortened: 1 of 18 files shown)


src-tauri/src/sessions.rs
```diff
@@ -499,10 +500,48 @@ pub fn flush_current_session(repo: &git2::Repository) -> Result<Session> {
...
+    // Create an anonymous remote
+    let mut remote = repo.remote_anonymous(remote_url).unwrap();
+
+    // Set the remote's callbacks
+    let mut callbacks = RemoteCallbacks::new();
+    callbacks.push_update_reference(move |refname, message| {
+        println!("Pushing reference '{}': {:?}", refname, message);
+        Ok(())
+    });
+    callbacks.push_transfer_progress(move |one, two, three| {
```
schacon, line 560: Oh, also, this is an example of a progress callback, in case we wanted to expose progress to the client somehow.

src-tauri/src/main.rs
```diff
@@ -299,15 +302,25 @@ fn main() {
                 })
                 .setup(move |app| {
                     let resolver = app.path_resolver();
-                    let storage = Storage::new(&resolver);
-                    let projects_storage = projects::storage::Storage::new(storage);
+
+                    let pstorage = Storage::new(&resolver);
```
ngalaiko, line 344: both project storage and use storage should share the storage instance, or is there a reason they do not?

src-tauri/src/users/storage.rs
```diff
@@ -0,0 +1,41 @@
...
+                        }
+                        Err(e) => {
+                            println!("Error: {:#?}", e);
+                            return Err(anyhow::anyhow!("Error: {:?}", e));
+                        }
+                    };
+                }
+                None => return Ok(None),
+            },
+            Err(_) => return Ok(None),
```
ngalaiko, line 38: instead of handling `Err` everywhere manually, use `?` macro like in other parts of the code. it basically does what you do here, but without code

src-tauri/src/watchers/git.rs
```diff
@@ -32,12 +32,14 @@ impl From<std::io::Error> for WatchError {
     }
 }
 
-const FIVE_MINUTES: u64 = Duration::new(5 * 60, 0).as_secs();
+//const FIVE_MINUTES: u64 = Duration::new(5 * 60, 0).as_secs();
+const FIVE_MINUTES: u64 = Duration::new(5, 0).as_secs();
```
ngalaiko: remove this?

src-tauri/src/users/user.rs
```diff
@@ -0,0 +1,18 @@
...
+    pub access_token: String,
+}
+
+impl AsRef<User> for User {
+    fn as_ref(&self) -> &User {
+        self
+    }
+}
+
+impl User {}
```
ngalaiko: if you copied it for no reason, remove it please

ngalaiko (commented):
you got it right! some rust related fixes though

src-tauri/src/main.rs
schacon: there was some type error or reuse or something, I can't remember. But it worked if I created a second. 🤷

src-tauri/src/watchers/git.rs
schacon: oh yeah, sorry - for testing quickly. maybe we should have a "flush" button in dev where we can make this happen

src/routes/projects/[projectId]/settings/+page.svelte
```diff
@@ -1,29 +1,62 @@
...
 
     const authApi = Authentication();
+    const projectsFile = 'projects.json';
 
     export let data: PageData;
     $: project = data.project;
     $: projects = data.projects;
     $: user = data.user;
 
+    Projects().then((projectStore) => {
```
Comment (ngalaiko) [inline src/routes/projects/[projectId]/settings/+page.svelte]: you should get the store from `data` here instead of creating a new one. 

see here: https://kit.svelte.dev/docs/load

src/routes/projects/[projectId]/settings/+page.svelte
```diff
@@ -1,29 +1,62 @@
...
+
+    function updateProject(project: Project) {
+        console.log("Update Project", project);
+        const items = $projects;
+        const newProjects = items.map(item => item.id === project.id ? project : item);
+
+        console.log(newProjects);
+        // omg
+
+        writeTextFile(projectsFile, JSON.stringify(newProjects), {
```
Comment (ngalaiko) [inline src/routes/projects/[projectId]/settings/+page.svelte]: didn't you want to use rust to save file? then, you should call rust api to update the file, and not do that directly. it would be a similar flow to the `add_project` call.

src-tauri/src/users/storage.rs
schacon: Is that equivalent? It won't return None on an error in that case, it will raise the error, no? The problem I can't seem to solve well here is that `serde_json` is super picky and will error out if there's any field that doesn't perfectly match or is missing or whatever, so I have to get it to map to `<Value>` which is more general and then parse each field. Ugh. But the point is that if there's a problem, I don't want the entire app to panic and crash because the user file was malformated or whatever.

src-tauri/src/main.rs
schacon: CleanShot 2023-02-15 at 14 02 07@2x

src/routes/projects/[projectId]/settings/+page.svelte
Comment (schacon) [reply src/routes/projects/[projectId]/settings/+page.svelte]: Yeah, I was just trying to figure out how any of this works. I was hoping I could import the store and bind to it normally like the [examples](https://svelte.dev/tutorial/auto-subscriptions), but we're doing it way differently and I can't figure out how to actually call things like `set` and `update` the way we're doing it. So I've deleted all this stuff for a totally different initial approach because it's the only thing I can actually get working after a few hours of frustrated flailing.

src/routes/projects/[projectId]/settings/+page.svelte
Comment (schacon) [reply src/routes/projects/[projectId]/settings/+page.svelte]: Yeah, but my experience with JSON handling in Rust is that it will take days to get this to work properly, because we're using a JSON file as a database. I want to update fields in one element of an array and that seems like it quickly becomes massively complicated, especially because of how insanely picky `serde_json` is. As far as I can tell, as soon as we add a field somewhere, everything will break unless we do the `<Value>` thing, which is horrible.

I don't hate the idea of having a json file per project for settings and whatnot, but I'm not sure.

Comment (schacon) [reply src/routes/projects/[projectId]/settings/+page.svelte]: Also, in the latest iteration, I'm actually doing both - writing the url to `.git/gb-url` which is removed if the setting is toggled off, so pushes only happen if that's there. But I'm also writing this file so I don't have to have another exposed rust call to check if a file is there. But also, I don't really know how we should do this. I cannot for the life of me get Svelte to update _part_ of the JSON file that backs the stores. If we can get that to work, then toggling doesn't have to completely remove the file, but we need to keep a bunch of settings and stuff somewhere.

PR #14, opened 2023-02-15
s3 upload

ngalaiko:
testing actions for s3 upload here


scripts/release.sh
```diff
@@ -189,12 +188,4 @@ info "  - $MACOS_UPDATER"
 info "  - $MACOS_UPDATER_SIG"
 info
 
-if [ "$DO_UPLOAD" = "true" ]; then
-	info "uploading release to s3"
-
-	aws s3 cp "$MACOS_DMG" "s3://releases.gitbutler.com/release/$VERSION/$OS/$ARCH/GitButler.dmg"
```
ngalaiko: please make sure to preserve `<channel>/<version>/<os>/<arch>` file structure in the release bucket, it'll be easier to work with in the future (from our experience with sturdy desktop app)

PR #16, opened 2023-02-17
do semver

ngalaiko:


releases/GitButler.app.tar.gz.sig
```diff
@@ -0,0 +1 @@
+dW50cnVzdGVkIGNvbW1lbnQ6IHNpZ25hdHVyZSBmcm9tIHRhdXJpIHNlY3JldCBrZXkKUlVUck9FSStpbTFYWUxrb1I5aXZza2hkb3pKZDF6N1FQNkt0SUdmVXkxYVEvWGRGQldrdmxiSU5YSFBqNnVyTjd1bmk3TnYyQ1lHS1hKKzYxM2FOQkV3eWlybEhCVjBQMHc4PQp0cnVzdGVkIGNvbW1lbnQ6IHRpbWVzdGFtcDoxNjc2NjMzODQ2CWZpbGU6R2l0QnV0bGVyLmFwcC50YXIuZ3oKKzhOTWRsTk5TbTBFeVo3OTNwTDNnNHZrdHFlSWdpNmw3NkdmTXRuQ0ZYM29rOWRKUkdqMXFxdEtla1M3QkpES1BnQ0hwN1hkaVNPa2g1eDZRU2FCQkE9PQo=
```
schacon: probably didn't mean to commit the `releases` directory

.github/workflows/publish.yaml
```diff
@@ -46,10 +47,9 @@ jobs:
                     --apple-id                   "${{ secrets.APPLE_ID }}" \
                     --apple-password             "${{ secrets.APPLE_PASSWORD }}"
 
-            - name: Set Current Date As Environment Variable
+            - name: Set env variable with version
               run: |
-                  ts=$(date +"%y.%m.%d")
-                  echo "version=$ts-${{ github.run_number }}" >> $GITHUB_ENV
```
schacon, line 52: the other thing we could do is just change the `-` to a `+` here, which makes this semver compatible. it's a little strange, because it's not really following the rules of semver, so it's strange that it's so strict about it.

schacon (approved):
looks good 👍

PR #17, opened 2023-02-18
Read dev server from .env

schacon:
I think this addresses the dev problem I have with wanting to run the client against localhost, but it requires a `.env` file for everyone, it doesn't appear to be able to be left blank.


README.md
```diff
@@ -13,6 +13,14 @@ then, make sure to install app dependencies:
 $ pnpm install
 ```
 
+### setup the env
```
ngalaiko: ideally, you shouldn't do that, see https://vitejs.dev/guide/env-and-mode.html#env-variables-and-modes

ngalaiko (approved):

README.md
schacon: I'm not sure I follow. Is Vite a subset of Svelte? Or are you saying the env file management is the same? If there is a way you prefer to do this, I don't much care how it's done, I just need to be able to point my build to `http://localhost:3000` and have it fall back to test and prod builds point to production. How would you like to do this?

ngalaiko: svelte is built with vite, and it has integration to get env variables: https://kit.svelte.dev/docs/modules#$env-static-public - this is perhaps a better link

if your solution works, let's just merge it, this is not very important with only one variable, that's why i approved

ngalaiko: then, once you have `.env.development` and `.env.production`, make sure that `pnpm build` is ran with a correct `--mode` setting (should be correct by default), and your variables will be accessible as you want.

ngalaiko (approved):

PR #18, opened 2023-02-22
Posthog integration GB-36

schacon:
I setup Posthog and this looks like it's initializing on load and identifying a user if we have user data. It also records sessions and page loads:

CleanShot 2023-02-22 at 13 49 12@2x

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -20,7 +20,7 @@
 
     user.subscribe((user) => {
         if (user) {
-            posthog.identify("user_" + $user.id.toString());
+            posthog.identify("user_" + user.id.toString());
         }
     });
 </script>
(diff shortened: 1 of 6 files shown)


src/routes/+layout.svelte
```diff
@@ -16,6 +17,11 @@
 
     export let data: LayoutData;
     const { user } = data;
+
+    if ($user) {
+        console.log("ID USER", $user.id);
+        posthog.identify("user_" + $user.id.toString());
+    }
```
ngalaiko:
```suggestion
  user.subscribe(user => {
    if (user) {
      // identify
    } else {
      // remove identification
  }
})
```

ngalaiko: since user is dynamic, we should support both set/unset

schacon: I don't think we can remove id, it's just to connect an anonymous user with an ID'd user, it basically just does nothing after the first time with a posthog internal private id. But I'll subscribe instead.

PR #21, opened 2023-02-24
Better default project home screen

schacon:
Added an empty state screen and rearranged the project listing.

CleanShot 2023-02-23 at 23 06 31@2x


ngalaiko (approved):
i like it!

PR #422, opened 2023-06-12
Rename a couple of components and types

mtsgrd:
When a type and a component name collide we've decided to alter the component name, typically by concatenating the type name and a noun for the intended visualisation. For example, Branch -> BranchLane.


krlvi (approved):
lgtm

PR #477, opened 2023-06-21
Move files corner case

ngalaiko:


src-tauri/src/virtual_branches/mod.rs
```diff
@@ -19,7 +19,6 @@ pub struct VirtualBranch {
     pub id: String,
     pub name: String,
     pub active: bool,
-    pub kind: String,
```
schacon, line 22: @ngalaiko this was here because it maps to the data structure that is used in the Typescript side. I'm not sure if it's needed there or not. 🤷

ngalaiko: not anymore at least, i've checked

PR #495, opened 2023-06-21
Make page load code a bit more succinct

mtsgrd:


src/routes/projects_new/[projectId]/types.ts
```diff
@@ -8,6 +8,7 @@ class DndItem {
 export class Hunk extends DndItem {
 	name!: string;
 	diff!: string;
+	@Transform((obj) => new Date(obj.value))
```
Comment (mtsgrd) [inline src/routes/projects_new/[projectId]/types.ts:11]: So we don't need to .map this to a date!

Comment (mtsgrd) [reply src/routes/projects_new/[projectId]/types.ts:11]: Is a value always guaranteed here btw?

PR #526, opened 2023-06-26
fun with stores

krlvi:
- something may or may not work here
- fix stupid mistake
- add refresh even when getting from cache
- use branches from store
- refresh store on updates
- explicitly define VirtualBranch interface

### DIFF
--- a/src/routes/projects_new/[projectId]/vbranches.ts
+++ b/src/routes/projects_new/[projectId]/vbranches.ts
@@ -1,8 +1,63 @@
 import { invoke } from '$lib/ipc';
 import { Branch } from './types';
 import { stores } from '$lib';
-import { Value } from 'svelte-loadable-store';
+import { writable, type Loadable, Value } from 'svelte-loadable-store';
 import { plainToInstance } from 'class-transformer';
+import type { Readable } from '@square/svelte-store';
+
+const cache: Record<string, Readable<Loadable<Branch[]>>> = {};
+
+export function getStore(projectId: string) {
+	if (projectId in cache) return cache[projectId];
+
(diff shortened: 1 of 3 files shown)


src/routes/projects_new/[projectId]/vbranches.ts
```diff
@@ -1,10 +1,67 @@
...
 import { Branch } from './types';
 import { stores } from '$lib';
-import { Value } from 'svelte-loadable-store';
+import { writable, type Loadable, Value } from 'svelte-loadable-store';
 import { plainToInstance } from 'class-transformer';
+import type { Readable, Writable } from '@square/svelte-store';
 
-export function sortBranchHunks(branches: Branch[]): Branch[] {
+/** Virtual Branch interface with custom operations on top of subscribing */
+export interface VirtualBranch extends Readable<Loadable<Branch[]>> {
```
Comment (mtsgrd) [inline src/routes/projects_new/[projectId]/vbranches.ts:9]: This feels unnecessary to me since we're not distributing a typescript library.

src/routes/projects_new/[projectId]/vbranches.ts
```diff
@@ -1,10 +1,67 @@
...
+	/**
+	 * Force re-fetch of the branches. Exists temporarily until we have all mutations on this interface.
+	 */
+	refresh(this: void): void;
+	// TODO: The other operations on branchesa, like create, delete, etc.
+}
+
+const cache: Record<string, Writable<Loadable<Branch[]>>> = {};
+
+export default (projectId: string): VirtualBranch => {
```
Comment (mtsgrd) [inline src/routes/projects_new/[projectId]/vbranches.ts:19]: Why not make this a regular function called `getStore(projectId)`? The call `vbranches(projectId);` doesn't give you any clue on what to expect the return value to be.

PR #659, opened 2023-07-08
vbranches feature package

krlvi:
- first stab
- use async
- make it a class
- use the new thing
- why is caching not working


src/lib/vbranches/BranchController.ts
```diff
@@ -0,0 +1,163 @@
+import type { Refreshable } from './BranchStoresCache';
+import type { Readable } from '@square/svelte-store';
+import type { Loadable } from 'svelte-loadable-store';
+import type { Branch, BranchData, Target } from './types';
+import { toasts } from '$lib';
+import * as ipc from './ipc';
```
mtsgrd, line 6: On the principle that clearer code is better code, you should explicitly import the things you want to use.

src/lib/vbranches/BranchController.ts
```diff
@@ -0,0 +1,163 @@
...
+	constructor(
+		readonly projectId: string,
+		readonly virtualBranchStore: Refreshable & Readable<Loadable<Branch[]>>,
+		readonly remoteBranchStore: Refreshable & Readable<Loadable<BranchData[]>>,
+		readonly targetBranchStore: Refreshable & Readable<Loadable<Target>>
+	) {}
+
+	async setTarget(branch: string) {
+		try {
+			const projectId = this.projectId;
```
mtsgrd: Unless it really helps with readability I wouldn't declare a variable that only gets used once. You have thirteen of them in this file. Maybe remove and just use `this.projectId`?

src/routes/repo/[projectId]/+page.ts
```diff
@@ -12,11 +12,24 @@ function sortBranchData(branchData: BranchData[]): BranchData[] {
 	return branchData.sort((a, b) => b.lastCommitTs - a.lastCommitTs);
 }
 
-export async function load(e: PageLoadEvent) {
-	const projectId = e.params.projectId;
+export async function load({ parent, params }: PageLoadEvent) {
+	const projectId = params.projectId;
 	const remoteBranchNames = await getRemoteBranches({ projectId });
```
Comment (mtsgrd) [inline src/routes/repo/[projectId]/+page.ts:17]: This probably goes away now since you have `targetBranchStore`?

src/lib/vbranches/index.ts
```diff
@@ -0,0 +1,25 @@
+/**
```
mtsgrd, line 1: In general, perhaps we don't need index.ts files? They feel most useful to me in the context of a published library rather than a product repo. For example, we never have to worry about breaking imports since we'd refactor all of them if the layout changes.

mtsgrd (approved):

src/lib/vbranches/BranchController.ts
krlvi: for some reason, typescript doesn't understand it that way :(

src/lib/vbranches/BranchController.ts
krlvi: you are right! i was being lazy

src/routes/repo/[projectId]/+page.ts
Comment (krlvi) [reply src/routes/repo/[projectId]/+page.ts:17]: nice catch!

src/lib/vbranches/BranchController.ts
mtsgrd: It's because you're using a short form for creating objects, typescript understands if you use e.g. `{ projectId: this.projectId, branch }`.

krlvi: 🤦‍♂️ thx!

src/routes/repo/[projectId]/+page.ts
Comment (krlvi) [reply src/routes/repo/[projectId]/+page.ts:17]: actually no, this is another thing - it gets just a list of branches for the dropdown, selecting a target. I will think about how to integrate this as well

PR #690, opened 2023-07-12
nightly builds

ngalaiko:


.github/workflows/publish.yaml
```diff
@@ -93,12 +116,12 @@ jobs:
...
                   aws_bucket: "releases.gitbutler.com"
                   source_dir: "release/"
-                  destination_dir: "releases/${{ env.version }}-${{ github.run_number }}"
+                  destination_dir: "releases/${{ github.event.inputs.channel }}/${{ env.version }}-${{ github.run_number }}"
 
             # tell our server to update with the version number
             - name: Tell our server to update
               run: |
-                  curl 'https://app.gitbutler.com/api/releases' \
+                  curl 'https://app.gitbutler.com/api/releases/${{ github.event.inputs.channel }}' \
```
schacon: Can we change this to be POST `/api/releases` that takes a version and a channel as params rather than the channel in the URL?

schacon: yeah, a little farther down there is a `--data` line, let's put the channel in there.

ngalaiko: ok!

PR #705, opened 2023-07-13
Ian/button fixes july 13

idonahue:


src/styles/main.postcss
```diff
@@ -94,34 +94,48 @@ button.active {
...
-		py-2
-		text-base
-		font-medium
-		leading-5
-		underline
-		transition
-		duration-150
-		ease-in-out
-		hover:ease-in;
+	@apply relative inline-flex w-fit cursor-pointer items-center gap-0.5 whitespace-nowrap rounded py-2 text-base font-medium leading-5 underline transition duration-150 ease-in-out hover:ease-in;
```
mtsgrd, line 97: Can you check what formatter you're using? Prettier is what we want, but I know we don't have any shared settings.

PR #931, opened 2023-08-03
Unsubscribe streams onDestroy

mtsgrd:
This is better than not unsubscribing, but there might be a cleaner way
achieving it.

### DIFF
--- a/src/lib/stores/deltas.ts
+++ b/src/lib/stores/deltas.ts
@@ -1,33 +1,35 @@
-import { writable, type Loadable, Loaded } from 'svelte-loadable-store';
+import { writable, Loaded } from 'svelte-loadable-store';
 import * as deltas from '$lib/api/ipc/deltas';
-import { get, type Readable } from '@square/svelte-store';
-
-const stores: Record<string, Readable<Loadable<Record<string, deltas.Delta[]>>>> = {};
+import { get } from '@square/svelte-store';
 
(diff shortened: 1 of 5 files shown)


src/lib/stores/deltas.ts
```diff
@@ -1,33 +1,35 @@
...
-			if (oldValue.isLoading) {
-				deltas.list(params).then(set);
-			} else if (Loaded.isError(oldValue)) {
-				deltas.list(params).then(set);
-			} else {
-				set({
+	const { store, unsubscribe } = getDeltasStore2(params);
+	return store;
+}
+export function getDeltasStore2(params: { projectId: string; sessionId: string }) {
```
mtsgrd, line 9: Just leaving a note here to say I introduce this function to avoid refactoring the whole codebase right away. There are a few call-sites that get this store.

src/lib/stores/sessions.ts
```diff
@@ -1,11 +1,14 @@
...
+import { asyncWritable, get } from '@square/svelte-store';
 
 export function getSessionStore(params: { projectId: string }) {
-	if (params.projectId in stores) return stores[params.projectId];
+	const { store, unsubscribe } = getSessionStore2(params);
+	return store;
+}
+
+export function getSessionStore2(params: { projectId: string }) {
+	console.log('getting sessions!');
```
mtsgrd, line 11: removed upstream

PR #991, opened 2023-08-10
Integration

ngalaiko:


src/routes/repo/[projectId]/+page.svelte
```diff
@@ -240,8 +240,9 @@
 					<Button
 						color="purple"
 						height="small"
-						on:click={() =>
-							invoke('git_switch_branch', { projectId, branch: 'gitbutler/integration' })}
+						on:click={() => {
+							if ($baseBranchStore) branchController.setTarget($baseBranchStore.branchName);
```
Comment (mtsgrd) [inline src/routes/repo/[projectId]/+page.svelte:244]: When we get here we know that `baseBranchState.isError == true`, so we don't have a value in the baseBranchStore. I discovered this when I tried to click the button. ✌️

PR #1006, opened 2023-08-11
detect commits to gitbutler/integration

ngalaiko:


src-tauri/src/virtual_branches/integration.rs
```diff
@@ -189,13 +198,132 @@ pub enum VerifyError {
...
+
+    let new_branch_head = project_repository
+        .git_repository
+        .find_commit(new_branch.head)
+        .context("failed to find new branch head")?;
+
+    // rebasing the extra commits onto the new branch
+    let writer = super::branch::Writer::new(gb_repository);
+    extra_commits.reverse();
+    for commit in extra_commits {
```
schacon, line 285: I'm a little confused as to what this is doing. It looks like if there are two commits, it will loop through them, rebase them both on top of the base and update a virtual branch once, then overwrite it with the second one. Maybe we can use the result from the first to be the parent of the second?

ngalaiko: hm yea, you are right

PR #1050, opened 2023-08-18
Testing resizable lanes

mtsgrd:
- removes the dblclick expansion feature
- we should find a different way of opening a lane in full-width


src/lib/components/Differ/CodeHighlighter.ts
```diff
@@ -164,7 +164,7 @@ export function languageFromFilename(filename: string): LanguageSupport | null {
 		//     return new LanguageSupport(await CodeMirror.scss());
 
 		case 'svelte':
-			return svelte();
+			return javascript({ typescript: true, jsx: true });
```
mtsgrd: This shouldn't be here, but it _does_ improve .svelte highlighting.

PR #1193, opened 2023-09-11
Signing commits

schacon:
Almost got commit signing working with the ed25519 key we write out, but we can't quite do it this way, because we don't specify an email address in the key when we write the key out, which I believe GitHub needs in order to verify. So the code signing part may well work correctly, but the key is possibly not formatted correctly for this.

I looked into this a bit more and I don't think this is easily possible. I can't even figure out how the email address is written into these keys. It will probably just be easier to be able to select a personal key the way that we do in the project settings for the push key. We might want to allow them to select different keys for each. But signing with the `id_ed25519` key that I generated by hand vs the random one we generate and uploading them both as signing keys to github results in one being verified and the other not, so I'm pretty sure that's the difference.

We may have to make the `keys` part a little more generic, to be able to use our keys_controller for more than populating the git credential thing.

### DIFF
--- a/packages/tauri/src/keys/key.rs
+++ b/packages/tauri/src/keys/key.rs
@@ -20,6 +20,10 @@ impl PrivateKey {
         Self::default()
     }
 
+    pub fn private_ssh_key(&self) -> ssh_key::PrivateKey {
+        self.0.clone()
+    }
+
     pub fn public_key(&self) -> PublicKey {
         PublicKey::from(self)
     }
--- a/packages/tauri/src/virtual_branches/vbranch.rs
+++ b/packages/tauri/src/virtual_branches/vbranch.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::{HashMap, HashSet},
+    env,
     os::unix::fs::PermissionsExt,
     path, time, vec,
 };
@@ -10,11 +11,13 @@ use serde::Serialize;
 
 use uuid::Uuid;
 
+use ssh_key::{HashAlg, LineEnding, PrivateKey, SshSig};
+
 use crate::{
     dedup::{dedup, dedup_fmt},
     gb_repository,
-    git::{self, diff},
-    keys::Key,
+    git::{self, credentials, diff, Commit, Oid, Repository, Tree},
+    keys,
     project_repository::{self, conflicts, LogUntil},
     reader, sessions,
 };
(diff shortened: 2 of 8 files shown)


packages/tauri/src/keys/key.rs
```diff
@@ -20,6 +20,10 @@ impl PrivateKey {
         Self::default()
     }
 
+    pub fn private_ssh_key(&self) -> ssh_key::PrivateKey {
+        self.0.clone()
+    }
+
```
ngalaiko, line 26: it doesn't make sense to return private ssh key from within a private ssh key. what do you want to use it for? implement a public method to do it here instead of opening internal details to a consumer

packages/tauri/src/virtual_branches/vbranch.rs
```diff
@@ -1814,6 +1817,7 @@ fn _print_tree(repo: &git2::Repository, tree: &git2::Tree) -> Result<()> {
 }
 
 pub fn commit(
```
ngalaiko, line 1819: instead, add another method called 

```
pub fn commit_signed(
    key: &keys::PrivateKey,
    gb_repository: &gb_repository::Repository,
    project_repository: &project_repository::Repository,
    branch_id: &str,
    message: &str,
    ownership: Option<&branch::Ownership>,
) -> Result<()> {
```

and do all the logic to figure out if a commit should be signed or not somewhere above it. perhaps in the controller.

we don't need to be forced to pass keys_storage everywhere if we are not sure it's needed

packages/tauri/src/virtual_branches/vbranch.rs
```diff
@@ -1928,6 +1938,56 @@ pub fn commit(
...
+    let value = config
+        .get_string("gitbutler.signCommits")
+        .unwrap_or("false".to_string());
+    if value == "true" {
+        dbg!("signing the commit");
+
+        if let Some(keys_storage) = keys_storage {
+            if let Ok(key) = keys_storage.get_or_create() {
+                dbg!(&key);
+
```
ngalaiko, line 1962: move this to the controller? and replace keys_storage with `Option<keys::PrivateKey>` in this private helper

packages/tauri/src/keys/key.rs
schacon: I need the actual ssh_key::PrivateKey object because I need to pass it to another actual `ssh_key` function and our wrapper doesn't satisfy the traits it needs.

packages/tauri/src/virtual_branches/vbranch.rs
schacon: what controller?

ngalaiko: virtual branches/controller.rs, where all the frontend requests are coming through

packages/tauri/src/keys/key.rs
ngalaiko: maybe just add a `sign` method that takes bytes to sign and returns signed bytes to our type?

PR #1279, opened 2023-10-03
refactor: use content that is either utf8, large or binary

ngalaiko:


packages/ui/src/lib/api/ipc/files.ts
```diff
@@ -1,14 +1,37 @@
 import { invoke, listen } from '$lib/ipc';
 
+export type Contents = { type: 'utf8'; value: string } | { type: 'binary' } | { type: 'large' };
+
+export namespace Contents {
```
mtsgrd: I'd recommend we follow the Google style guide and avoid using namespaces: https://google.github.io/styleguide/tsguide.html#namespaces-vs-modules

packages/ui/src/lib/api/ipc/files.ts
```diff
@@ -1,14 +1,37 @@
 import { invoke, listen } from '$lib/ipc';
 
+export type Contents = { type: 'utf8'; value: string } | { type: 'binary' } | { type: 'large' };
```
mtsgrd: I'm also a bit curious what we gain with nested structure for content rather than applying the properties flat on the file object?

mtsgrd: Actually, I didn't read the surrounding code properly. This has nothing to do with the files you are working with, and only to do with reading full file contents. *edit: could we call it `FileContent` instead of `Contents`?

packages/ui/src/lib/api/ipc/files.ts
ngalaiko: removed it completely

packages/ui/src/lib/api/ipc/files.ts
ngalaiko: renamed

PR #1316, opened 2023-10-09
GB-610: flush session if GB_FLUSH file triggers watcher

extrawurst:
trigger `flush_session` if `GB_FLUSH` file triggers watcher git-event

### DIFF
--- a/packages/tauri/Cargo.toml
+++ b/packages/tauri/Cargo.toml
@@ -7,6 +7,9 @@ rust-version = "1.57"
 [build-dependencies]
 tauri-build = { version = "1.5", features = [] }
 
+[dev-dependencies]
+pretty_assertions = "1.4"
+
 [dependencies]
 anyhow = "1.0.72"
 async-trait = "0.1.73"
--- a/packages/tauri/src/watcher/handlers/git_file_change.rs
+++ b/packages/tauri/src/watcher/handlers/git_file_change.rs
@@ -1,20 +1,32 @@
+use std::path;
+
 use anyhow::{Context, Result};
 use tauri::AppHandle;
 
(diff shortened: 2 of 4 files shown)


schacon (approved):
Looks good!

packages/tauri/src/watcher/handlers/git_file_change.rs
```diff
@@ -64,3 +103,32 @@ impl Handler {
...
+        let flush_file_path = project.path.join(".git/GB_FLUSH");
+        std::fs::write(flush_file_path.as_path(), "")?;
+
+        listener.handle("GB_FLUSH", &project.id)?;
+
+        assert!(!flush_file_path.exists());
+
+        Ok(())
+    }
+}
```
ngalaiko: I think you should also assert if the session was flushed, and add cases like:
* no session, no file -> nothing happens
* session is there, no file -> session is still there
* no session, file -> no flushing
* session, file -> flushed

you can get access to sessions from using gb_repository that is a part of the Case type

it's mostly to get familiar with the codebase, otherwise it looks good

ngalaiko (approved):

packages/tauri/src/watcher/handlers/git_file_change.rs
extrawurst: I think i covered these cases now. 
is there a better way to simulate having a session other than creating a file and triggering a `project_file_change` event?

packages/tauri/Cargo.toml
extrawurst, line 1: let me clean this up. i have this toml vs code extension which also auto formats it

ngalaiko (approved):

PR #1328, opened 2023-10-10
GB-609: create/update `.git/gitbutler.json`

extrawurst:
this happens when a new session is created

### DIFF
--- a/packages/tauri/src/gb_repository/repository.rs
+++ b/packages/tauri/src/gb_repository/repository.rs
@@ -344,6 +344,8 @@ impl Repository {
             "created new session"
         );
 
+        self.flush_gitbutler_file(&session.id, &project_repository)?;
+
         Ok(session)
     }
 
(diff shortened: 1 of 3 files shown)


packages/tauri/src/gb_repository/repository.rs
```diff
@@ -593,6 +595,28 @@ impl Repository {
...
+        let gb_file_content = serde_json::json!({
+            "sessionId": session_id,
+            "repositoryId": project_id,
+            "gbPath": gb_path,
+        });
+
+        let gb_file_path = project_repository.path().join(".git/gitbutler.json");
+        std::fs::write(&gb_file_path, gb_file_content.to_string())?;
+
+        tracing::info!("gitbutler file updated: {:?}", gb_file_path);
```
ngalaiko: let's make it debug, @mtsgrd doesn't like noisy logs 🙂

packages/tauri/src/project_repository/repository.rs
```diff
@@ -99,6 +99,10 @@ impl Repository {
         self.git_repository.path().parent().unwrap()
     }
 
+    pub fn gb_path(&self) -> &std::path::Path {
+        self.git_repository.path()
+    }
```
ngalaiko: i think what @schacon meant by gb_path is the path of the `gb_repository.git_repository`

anyhow, `gb_` is usually related to our data repository, and here you have a path to the user's project repository that is not gb.

ngalaiko (changes requested):
you have to use a different path, and also maybe add a simple test? we have those in the `gb_repository/repository_tests.rs`

packages/tauri/src/project_repository/repository.rs
schacon: Yes, I wanted the path to the gitbutler git repository, not the project git repository. 👍

ngalaiko (approved):

PR #1329, opened 2023-10-10
Reset branch to specific commit

ngalaiko:


schacon (approved):
OK, yeah, this all looks good. It does allow resetting past the last push point, correct? So we'll enforce that in the UI later if we want to not allow force-pushes? Otherwise right now if you reset beyond the last pushed commit, pushing again won't work because we don't allow forced pushes, correct?

PR #1330, opened 2023-10-10
make sure clippy does not allow us to use unsafe

extrawurst:


packages/tauri/src/virtual_branches/target/reader.rs
extrawurst, line 1: @ngalaiko any reason keep using unsafe for unittests using global statics? this way we can project wide enfore non-unsafe via clippy

ngalaiko: no particular reason, that's just what I found on stackoverflow 🙂

ngalaiko (approved):

PR #1331, opened 2023-10-10
more clippy lints and the according fixes

extrawurst:


ngalaiko (approved):
how can I get the same suggestions? some flags I should use?

packages/tauri/src/lib.rs
```diff
@@ -1,3 +1,11 @@
+#![deny(
```
extrawurst, line 1: this is how i enabled more lints. without my fixes it would have made clippy complain about all the places.

i just started, this is the first batch. 
eventually we might want to be able to enable:
```
#![deny(clippy::all, clippy::perf)]
```

but for now this generates too much noise. i will work through that step by step

extrawurst: cc @ngalaiko

PR #1333, opened 2023-10-11
fix: handle errors properly in init_device_oauth and check_auth_statu…

ngalaiko:
…s functions

- Previously, the functions `init_device_oauth` and `check_auth_status` did not properly handle errors when sending requests and parsing response bodies.
- This commit updates the error handling in both functions to provide more informative error messages and properly propagate errors.
- Now, if there is an error sending the request or parsing the response body, the functions will return an `Error::Unknown` error with the appropriate context.
- This improves the reliability and error reporting of the login functionality in the client application.

### DIFF
--- a/packages/tauri/src/github/commands.rs
+++ b/packages/tauri/src/github/commands.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 
+use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
-use anyhow::{Result, Context};
 use tracing::instrument;
 
 use crate::error::Error;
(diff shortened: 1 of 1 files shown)


packages/tauri/src/github/commands.rs
```diff
@@ -32,37 +33,14 @@ pub async fn init_device_oauth() -> Result<Verification, Error> {
...
-                Err(_) => {
-                    return Err(Error::Unknown);
-                }
-            }
-        }
-        Err(_) => Err(Error::Unknown),
-    }
+    let rsp_body = res.text().await.context("Failed to get response body")?;
+
+    serde_json::from_str(&rsp_body)
```
krlvi, line 41: shouldnt this add `< CodesContainer >` for deserialization?

ngalaiko: CodesContainer is the same as Verification. and because here we return from the function, you don't need to give it type hints, they'll be derived from the function return type

krlvi: Oh i see! this is really cool that it can do that

PR #1356, opened 2023-10-12
conflicts tests & fixes

ngalaiko:


packages/tauri/src/git/repository.rs
```diff
@@ -17,7 +17,6 @@ impl<'a> From<&'a Repository> for &'a git2::Repository {
 }
 
 impl Repository {
-    #[cfg(test)]
```
extrawurst: hm its a bit bad that this now leaks test code into the production build .. the only way to fix this is by using a feature flag AFAIK

packages/tauri/src/virtual_branches/controller.rs
```diff
@@ -150,14 +175,16 @@ impl Controller {
         &self,
         project_id: &str,
         create: &super::branch::BranchCreateRequest,
-    ) -> Result<(), Error> {
+    ) -> Result<String, Error> {
```
extrawurst, line 178: what do you think about newtyping this to a type `VirtualBranchId` to benefit from more type safety?

packages/tauri/src/virtual_branches/mod.rs
```diff
@@ -12,6 +12,7 @@ mod base;
 pub use base::*;
 
 pub mod controller;
+pub use controller::{Controller, Error as ControllerError};
```
extrawurst, line 15: not for this PR but i think we could make things more digestible by having less generic type names in some places. the various error types are a good example for that so that aliasing them is not needed anymore and the name is more expressive

packages/tauri/src/git/repository.rs
ngalaiko: i've used another method instead, and hid this one back.

but then also, it's a convenience wrapper over libgit, so not a big issue I think, since those are in libgit anyway

packages/tauri/tests/virtual_branches/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+                controller,
+                ..
+            } = Test::default();
+
+            assert!(controller
+                .set_base_branch(
+                    &project_id,
+                    &git::RemoteBranchName::from_str("refs/remotes/origin/missing").unwrap(),
+                )
+                .is_err());
```
extrawurst: we can check more precise here by matching what error we expect, right?

packages/tauri/src/git/repository.rs
```diff
@@ -381,7 +382,6 @@ impl Repository {
         self.0.find_tree(oid).map(Into::into).map_err(Into::into)
     }
 
-    #[cfg(test)]
```
extrawurst, line 384: same as with the other comment

packages/tauri/tests/virtual_branches/mod.rs
ngalaiko: right!

packages/tauri/src/virtual_branches/controller.rs
ngalaiko: yea I think it's time to add those. i'll make another pr for all the typed ids

packages/tauri/src/virtual_branches/mod.rs
ngalaiko: yes, this is a leftover from the time when I thought one package should have one Error type. now I am more on `one error type per method` 🤔

extrawurst: one per method?? i am not sure I agree 😄

extrawurst: i would argue one per logical unit/module

PR #1364, opened 2023-10-13
Add id module

ngalaiko:


packages/tauri/src/bookmarks/writer.rs
```diff
@@ -24,7 +24,7 @@ impl<'writer> BookmarksWriter<'writer> {
         )?;
 
         tracing::debug!(
-            project_id = self.repository.get_project_id(),
+            project_id = %self.repository.get_project_id(),
```
extrawurst, line 27: did that fail to build without or is just weird formatting? if its the latter we have to be careful to not miss a spot where the new ids are traced

packages/tauri/src/commands.rs
```diff
@@ -99,7 +127,11 @@ pub async fn git_remote_branches(
     project_id: &str,
 ) -> Result<Vec<git::RemoteBranchName>, Error> {
     let app = handle.state::<app::App>();
-    let branches = app.git_remote_branches(project_id)?;
+    let project_id = project_id.parse().map_err(|_| Error::UserError {
```
extrawurst, line 130: why do we have to manually map here? cant we have a Into/From for the parse error?

packages/tauri/src/commands.rs
```diff
@@ -148,6 +188,10 @@ pub async fn upsert_bookmark(
         .elapsed()
         .context("failed to get time")?
         .as_millis();
+    let project_id = project_id.parse().map_err(|_| Error::UserError {
```
extrawurst, line 191: see above that is a lot of duplication

packages/tauri/src/deltas/database.rs
```diff
@@ -141,19 +141,19 @@ mod tests {
         let db = test_utils::test_database();
         let database = Database::from(db);
 
-        let project_id = "project_id";
-        let session_id = "session_id";
+        let project_id = ProjectId::generate();
```
extrawurst, line 144: i love it :)

packages/tauri/src/projects/storage.rs
```diff
@@ -125,9 +129,9 @@ impl Storage {
             .clone())
     }
 
-    pub fn purge(&self, id: &str) -> Result<(), Error> {
+    pub fn purge(&self, id: &ProjectId) -> Result<(), Error> {
         let mut projects = self.list()?;
-        if let Some(index) = projects.iter().position(|p| p.id == id) {
+        if let Some(index) = projects.iter().position(|p| p.id == *id) {
```
extrawurst, line 134: maybe slightly more efficient to compare via ref instead of deref'd

extrawurst (commented):
i reviewed most of it, its mingled with a lot clippy-in-test fixes. i think I saw most of the new patterns and left a few minor comments

packages/tauri/src/bookmarks/writer.rs
ngalaiko: that failed to build. it's a macro to use fmt::Display basically

packages/tauri/src/commands.rs
ngalaiko: the error type there is uuid::Error, and I wanted to have a message that is related to projects here. same for other kind of ids

PR #1383, opened 2023-10-16
feat: add flush_vbranches method to Controller struct

extrawurst:
move `flush_vbranches` functionality in controller and make sure it locks the right mutexes

### DIFF
--- a/packages/tauri/src/virtual_branches/controller.rs
+++ b/packages/tauri/src/virtual_branches/controller.rs
@@ -1,7 +1,6 @@
 use std::{collections::HashMap, sync::Arc};
 
 use anyhow::Context;
-use futures::executor::block_on;
 use tauri::AppHandle;
 use tokio::sync::Semaphore;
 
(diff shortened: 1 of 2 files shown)


packages/tauri/src/virtual_branches/controller.rs
```diff
@@ -530,4 +532,32 @@ impl Controller {
         let _permit = semaphore.acquire().await;
         action()
     }
+
+    pub fn flush_vbranches(&self, project_id: ProjectId) -> Result<(), Error> {
+        block_on(async {
```
ngalaiko: let's make this async for consistency, and move block_on to the caller

packages/tauri/src/virtual_branches/controller.rs
```diff
@@ -530,4 +532,32 @@ impl Controller {
...
+            let _ = self
```
ngalaiko: it won't be relevant if you fix the other comment, but I think this would make it ignore the error that might be returned from the block_on's callback

extrawurst: yeah i fixed that already :D

ngalaiko (approved):

PR #1398, opened 2023-10-17
fix vbranch flushing to not flush non-applied vbranches

extrawurst:
the previous implementation had two problems:
* it flushed (wrote to disk) vbranches that were not applied
* it also wrote them to disk with the flag `applied` set to true


packages/tauri/src/virtual_branches/virtual.rs
```diff
@@ -463,11 +463,11 @@ pub fn unapply_branch(
     Ok(())
 }
 
-pub fn flush_vbranch_as_tree(
+fn flush_vbranch_as_tree(
     gb_repository: &gb_repository::Repository,
     project_repository: &project_repository::Repository,
     branch_id: &BranchId,
-    applied: bool,
+    applied: bool, // set `Branch.applied` before writing
```
ngalaiko: is it though? in the unapply function it's `false` even though before "writing" is's true

extrawurst: originally i extracted the functionality of `flush_vbranch_as_tree` out of `unapply_branch` it also just set the `applied` to false before the write. the opposite happens in `apply_branch`

ngalaiko: let's just have a `Branch` here?

ngalaiko: I suspect this will trigger a chain of simplifications that will lead to removing `flush_vbranch_as_tree` completely

extrawurst: ok done

extrawurst: > I suspect this will trigger a chain of simplifications that will lead to removing flush_vbranch_as_tree completely

ok wait that one i do not see

packages/tauri/src/virtual_branches/virtual.rs
```diff
@@ -530,27 +536,52 @@ fn unapply_all_branches(
...
+    }
+
+    Ok(())
+}
+
+pub fn flush_applied_vbranches(
+    gb_repository: &gb_repository::Repository,
+    project_repository: &project_repository::Repository,
+) -> Result<()> {
+    let vbranches = list_applied_vbranches(gb_repository)?;
```
ngalaiko: since flush_vbranch_as_tree already does
```  
  let applied_branches = Iterator::new(&current_session_reader)
        .context("failed to create branch iterator")?
        .collect::<Result<Vec<branch::Branch>, reader::Error>>()
        .context("failed to read virtual branches")?
        .into_iter()
        .filter(|b| b.applied)
        .collect::<Vec<_>>();
    let applied_statuses = get_applied_status(
        gb_repository,
        project_repository,
        &default_target,
        applied_branches,
    )
    .context("failed to get status by branch")?;
```  

    
 i think we should just call 
 
```
 let tree = write_tree(project_repository, &default_target, files)?;
 let mut branch = branch;
 branch.tree = tree;
 branch_writer.write(&branch)?;
````

 here  for every applied branch's status

extrawurst: now i get what you mean. reducing the code duplication

ngalaiko (approved):

PR #1399, opened 2023-10-17
use github token for fetching / pushing

ngalaiko:


packages/tauri/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![forbid(unsafe_code, dead_code)]
```
extrawurst: are you sure you want it to not fail at all? if we put it in `deny` we can explicitly allow it in places where we want to

ngalaiko: brought it back!

extrawurst (approved):
looks good to me, nice abstraction.

PR #1401, opened 2023-10-17
refactor: update type conversions and remove unnecessary clippy lints

ngalaiko:


packages/tauri/src/git/index.rs
```diff
@@ -138,8 +138,14 @@ impl From<git2::IndexEntry> for IndexEntry {
 impl From<IndexEntry> for git2::IndexEntry {
     fn from(entry: IndexEntry) -> Self {
         Self {
-            ctime: git2::IndexTime::new(entry.ctime.seconds() as i32, entry.ctime.nanoseconds()),
-            mtime: git2::IndexTime::new(entry.mtime.seconds() as i32, entry.mtime.nanoseconds()),
+            ctime: git2::IndexTime::new(
+                i32::try_from(entry.ctime.seconds()).unwrap(),
```
extrawurst, line 142: wtf IndexTime is not 64bit? 2038 -> 💥

packages/tauri/src/lib.rs
```diff
@@ -77,7 +77,6 @@
 )]
 //TODO: should probably be cleaned up as any of these could lead to panics or unexpected behaviour (the cast-ones)
 #![allow(
-    clippy::cast_possible_truncation,
```
extrawurst, line 80: ❤️

packages/tauri/src/git/index.rs
extrawurst: @ngalaiko found this in libgit2:
https://github.com/libgit2/libgit2/blob/a6205fa5c1f856b9a7b3908077aa935a1b82fed9/include/git2/index.h#L48

> Note that the time and size fields are truncated to 32 bits. This is enough to detect changes

we probably should not panic but truncate too?

PR #1402, opened 2023-10-17
reduce dynamic allocations

extrawurst:
use `map` to make two relevant fields a tuple and then `unzip` it into two collections


ngalaiko (approved):
neat!

PR #1406, opened 2023-10-18
check unused dependencies in ci

extrawurst:
* also fixes the ci warnings about `set-output` usage due to using a maintained action to install the rust toolchain

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -92,6 +92,7 @@ jobs:
       - uses: ./.github/actions/init-env-rust
         with:
           rust: "nightly"
-      - run: cargo install cargo-udeps --locked
+      - name: install cargo udeps
+        run: cargo install cargo-udeps --locked
       - run: pnpm build
       - run: cargo +nightly udeps --all-targets
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -92,7 +92,7 @@ jobs:
       - uses: ./.github/actions/init-env-rust
         with:
           rust: "nightly"
-      - name: install cargo udeps
+      - name: Install cargo udeps
         run: cargo install cargo-udeps --locked
       - run: pnpm build
       - run: cargo +nightly udeps --all-targets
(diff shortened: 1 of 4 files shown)


.github/workflows/push.yaml
```diff
@@ -81,3 +81,25 @@ jobs:
...
+      - name: Install cargo-udeps
+        env:
+          UDEPS_VERSION: v0.1.42
+        run: |
+          wget -qO- https://github.com/est31/cargo-udeps/releases/download/${UDEPS_VERSION}/cargo-udeps-${UDEPS_VERSION}-x86_64-unknown-linux-gnu.tar.gz | tar -xzf-
+          cd cargo-udeps-${UDEPS_VERSION}-x86_64-unknown-linux-gnu
+          mv cargo-udeps ~/.cargo/bin/
+          cargo-udeps help
+          cargo udeps -V
+      - run: mkdir -p packages/ui/build
```
extrawurst: @ngalaiko any idea why this folder-creation is needed? it fails otherwise like here: https://github.com/gitbutlerapp/gitbutler-client/actions/runs/6558692461/job/17812807010#step:6:1155

ngalaiko: @extrawurst when you compile tauri app, it uses `build` as a source for the ui files (index.html and so on)

that's why we run clippy after building frontend. i think you should run udeps similarly (in the test step)

.github/workflows/push.yaml
```diff
@@ -81,3 +81,25 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/init-env-node
+      - uses: ./.github/actions/init-env-rust
+        with:
+          rust: "nightly"
+      - name: Install cargo-udeps
+        env:
+          UDEPS_VERSION: v0.1.42
+        run: |
+          wget -qO- https://github.com/est31/cargo-udeps/releases/download/${UDEPS_VERSION}/cargo-udeps-${UDEPS_VERSION}-x86_64-unknown-linux-gnu.tar.gz | tar -xzf-
```
ngalaiko: can we cargo install? then we'll also cache it

extrawurst: good idea. lets do that

ngalaiko (approved):

PR #1442, opened 2023-10-23
GB-623: make sure integration is up to date before flushing all vbranches

extrawurst:


packages/tauri/src/virtual_branches/controller.rs
```diff
@@ -518,6 +518,9 @@ impl Controller {
     pub async fn flush_vbranches(&self, project_id: ProjectId) -> Result<(), Error> {
         self.with_lock(&project_id, || {
             self.with_verify_branch(&project_id, |gb_repository, project_repository, _| {
+                super::integration::update_gitbutler_integration(gb_repository, project_repository)
```
extrawurst: @ngalaiko or would it be better fit to always happen inside `flush_applied_vbranches`?

ngalaiko: i don't think it matters too much. i think we want all the code form virtual.rs to be here, inside the controllers. i'd say depends on how and if you test it. if it's only flush_applied_vbranches, then we'd have to put it there

PR #1479, opened 2023-10-31
GB-613 push project code to gb repo

extrawurst:


packages/tauri/src/project_repository/repository.rs
```diff
@@ -275,6 +305,71 @@ impl Repository {
         }
     }
 
+    pub fn push_to_gitbutler_server(
+        &self,
+        oid: &git::Oid,
+        user: Option<&users::User>,
+        ref_prefix: &str,
```
ngalaiko, line 312: do we have to keep `oid`  and `tmp / non tmp` calculation outside?

ngalaiko (approved):

packages/tauri/src/watcher/handlers/push_project_to_gitbutler.rs
```diff
@@ -0,0 +1,128 @@
...
+                    .context("failed to update last push")?;
+
+                tracing::debug!(
+                    %project_id,
+                    "project batch pushed: {}/{}",ids.len()-idx,ids.len(),
+                );
+            }
+
+            project_repository
+                .push_to_gitbutler_server(&head_id, user.as_ref(), "")
```
schacon, line 112: So a problem is that we're only pushing the `HEAD` ref here, which is fine for getting the combination of the virtual branch data, but we don't get the trees and commits that are written for each branch, which are kept in `refs/gitbutler/*`. We need to essentially run a `git push [gbserver] +refs/gitbutler/*:refs/gitbutler/*` after this. It's fine to push the HEAD, since that should have most of the data we need, but after the HEAD is updated, we need to also push the gb refs.

extrawurst: ok i will merge this and address that in a separate PR 👍

packages/tauri/src/project_repository/repository.rs
extrawurst: i am doing this to be able to reuse the same function to push to the tmp-ref and to the final ref as well as to push the in-between-steps (ids) and the final one

PR #1483, opened 2023-10-31
Refactor HandlerInner usage of Mutex

extrawurst:
cleanup to be more idiomatic rust by wrapping the shared resource inside a Mutex


ngalaiko (approved):
👍 it is indeed more clear

PR #1510, opened 2023-11-03
Add draggable action gb-607

ngalaiko:


packages/ui/src/lib/utils/dragable.ts
```diff
@@ -0,0 +1,180 @@
...
+	hover: string;
+	accepts: (data: any) => boolean;
+	onDrop: (data: any) => Promise<void> | void;
+}
+
+const defaultDropzoneOptions: Dropzone = {
+	disabled: false,
+	active: 'dropzone-active',
+	hover: 'dropzone-hover',
+	accepts: (data) => data === 'default',
```
mtsgrd: For consistency in code base let's avoid using strict equality checks.

packages/ui/src/lib/utils/dragable.ts
```diff
@@ -0,0 +1,180 @@
...
+			unregister(options);
+
+			node.removeEventListener('dragenter', handleDragEnter);
+			node.removeEventListener('dragleave', handleDragLeave);
+			node.removeEventListener('dragover', handleDragOver);
+		}
+	};
+}
+
+//
```
mtsgrd: ☝️

packages/ui/src/lib/utils/dragable.ts
```diff
@@ -0,0 +1,180 @@
...
+function register(node: HTMLElement, dropzone: Dropzone) {
+	registry.push([node, dropzone]);
+}
+
+function unregister(dropzone: Dropzone) {
+	const index = registry.findIndex(([, dz]) => dz === dropzone);
+	if (index >= 0) registry.splice(index, 1);
+}
+
+//
```
mtsgrd: ☝️

packages/ui/src/lib/utils/dragable.ts
```diff
@@ -0,0 +1,180 @@
...
+
+export interface Dragable {
```
mtsgrd: Draggable with two d's pls

packages/ui/src/lib/utils/dragable.ts
```diff
@@ -0,0 +1,180 @@
...
+	data: any;
+	disabled: boolean;
+}
+
+const defaultDragableOptions: Dragable = {
```
mtsgrd: Same

packages/ui/src/lib/utils/dragable.ts
```diff
@@ -0,0 +1,180 @@
...
+	data: 'default',
+	disabled: false
+};
+
+export function dragable(node: HTMLElement, opts: Partial<Dragable> | undefined) {
```
mtsgrd: Same

packages/ui/src/lib/utils/dragable.ts
mtsgrd, line 1: -> `draggable.ts`

packages/ui/src/lib/vbranches/branchController.ts
```diff
@@ -243,4 +243,16 @@ export class BranchController {
 			toasts.error(`Failed to mark file resolved`);
 		}
 	}
+
+	async amendBranch(params: { branchId: string; ownership: string }) {
```
mtsgrd: For consistency in the code base, can we please phase out the use of pseudo named parameters?

mtsgrd (approved):
Hey, just taking off from London to Berlin. Lgtm, but can you see my comments  before merging  🙏

packages/ui/src/lib/vbranches/branchController.ts
ngalaiko: in this case both parameters are strings, so it's easy to make a mistake calling the method. 

or is it not what you mean?

ngalaiko: i've changed it to `amendBranch({branchId, ownership}: {branchId: string, ownership: string})’` for both not having pseudo name and having distinguishable parameters

mtsgrd: > in this case both parameters are strings, so it's easy to make a mistake calling the method

Imho it's rare to encounter these kinds of mistakes, especially for functions that only take two parameters. But then again, I would like to follow the Google style guide for typescript, and it explicitly allows for what you've refactored it into.

So if you don't have a strong opinion, let's go with regular params, otherwise I'm onboard with what you now have. ✌️ 

See: https://google.github.io/styleguide/tsguide.html#object-destructuring

ngalaiko: what you've linked is about object destructing with default values, this case is about defining input parameters, which is a different topic

if you really think that having flat parameters with same types like
```
function doSomething(parameterOne: string, parameterTwo: string, parameterThree: string) { ... }

doSomething("one", "two", "tree")
```

is more clear than forcing a named inputs with
```
function doSomething(params: {parameterOne: string, parameterTwo: string, parameterThree: string}) { ... }

doSomething({parameterOne: "one", parameterTwo: "two", parameterThree: "tree")
```

i'll do it, since you contribute more to typescript and having a consistent style is more important

krlvi: Named parameters are much clearer to read. This style guide from google doesn't articulate the drawbacks of this approach. Do you know what we'd be sacrificing (could it be making is similar to other languages in their ecosystem)?  
Other styleguides out there permit this, I would rather us focus on clarity

packages/ui/src/lib/utils/dragable.ts
krlvi: @mtsgrd let's have any conventions we establish and agree on added as lint rules

PR #1539, opened 2023-11-08
GB-627: Session flush command

extrawurst:


packages/tauri/src/sessions/controller.rs
```diff
@@ -84,4 +133,53 @@ impl Controller {
...
+            user.as_ref(),
+        )
+        .context("failed to open gb repository")?;
+
+        let current_session = gb_repository.get_current_session()?;
+
+        Ok(current_session)
+    }
+
+    pub fn flush(&self, project_id: &ProjectId, session: &Session) -> Result<Session, FlushError> {
```
ngalaiko: since you can only flush the current session, there is no need to pass it to the arguments here

ngalaiko: you can call gb_repository::Repository::flush

PR #1701, opened 2023-11-24
Css fixes

PavelLaptev:
- removed tailwind classes from body (except `text-base`). and move all body styles into the `main` css file.
- added props that optimize font smoothing and legibility
- added conditional cursors to `Resizer` component
- Because Svelte isolate classes like `css-modules` in React I cleaned some class names and renamed e.g. `card__button` to just `button`. because there are no other buttons in the component and no need to add extra specificity. 
- Updated `sync-btn` component at the top.
- updated `project selector` component styles
- added `transition` variables into `tokens` file
- updated `text` classes file


packages/ui/src/styles/text-classes.css
```diff
@@ -144,3 +136,13 @@
 	letter-spacing: 0;
 	line-height: 100%;
 }
+
+/* modifiers */
+
+.text-semibold {
+	font-weight: 500;
```
mtsgrd: Since this is the new default, should semibold be something thicker? Also, as I'm writing this I'm unsure if the new default is actually default, or just where I use the new classes.

packages/ui/src/routes/[projectId]/navigation/ProjectSelector.svelte
```diff
@@ -36,22 +36,25 @@
...
-<style>
+<style lang="postcss">
 	.wrapper {
 		margin-top: var(--space-10);
 		margin-bottom: var(--space-16);
 	}
+
 	.button {
 		display: flex;
+		height: var(--space-40);
```
Comment (mtsgrd) [inline packages/ui/src/routes/[projectId]/navigation/ProjectSelector.svelte]: I don't like having both padding and height set since the padding doesn't do anything. Wdyt?

packages/ui/src/routes/[projectId]/navigation/ProjectSelector.svelte
```diff
@@ -27,7 +27,7 @@
 				e.preventDefault();
 			}}
 		>
-			{project.title}
+			<span class="button__label text-base-14 text-bold">{project.title}</span>
```
Comment (mtsgrd) [inline packages/ui/src/routes/[projectId]/navigation/ProjectSelector.svelte:30]: Could we maybe set the font stuff on the button instead? Otherwise the line-height in the button is inherited from its parent.

mtsgrd (approved):

packages/ui/src/routes/[projectId]/navigation/Header.svelte
```diff
@@ -48,17 +59,51 @@
...
 		justify-content: right;
-		color: var(--clr-theme-scale-ntrl-50);
 	}
-	.header__sync {
+
+	.sync-btn {
 		display: flex;
 		align-items: center;
-		gap: var(--space-4);
+		height: 20px;
```
Comment (mtsgrd) [inline packages/ui/src/routes/[projectId]/navigation/Header.svelte]: height thing, but also not a variable?

packages/ui/src/styles/text-classes.css
PavelLaptev: thanks, yes. It should be `600` I'll fix it

packages/ui/src/routes/[projectId]/navigation/ProjectSelector.svelte
Comment (PavelLaptev) [reply packages/ui/src/routes/[projectId]/navigation/ProjectSelector.svelte]: will fix it :-)

packages/ui/src/routes/[projectId]/navigation/Header.svelte
Comment (PavelLaptev) [reply packages/ui/src/routes/[projectId]/navigation/Header.svelte]: Let's actually leave `height` here, but I'll add a variable instead px
I can't do `20px` height with padding because of the text frame. The easiest way to achieve 20px will be just to set it as a `height`
image

PR #1714, opened 2023-11-24
fix push code

extrawurst:
* unit tests that pushing works correct
* fix last pushed id to be `default_target.sha` (chunked rev walk misbehaved in single batch cases)

- [x] no pushes at all (when all is up to date)
- [x] multiple batches pushed (when there are more then batch_size commits)
- [ ] pushing of remote branches
- [x] pushing of local branches


ngalaiko (dismissed):
please add a test case with:
1. no pushes at all (when all is up to date)
2. multiple batches pushed (when there are more then batch_size commits)
3. pushing of remote branches
4. pushing of local branches

krlvi (approved):
lets try this out

PR #1741, opened 2023-11-27
Refactor apply branch function

krlvi:


packages/ui/src/lib/vbranches/types.ts
```diff
@@ -53,6 +53,7 @@ export class Branch {
 	@Type(() => RemoteBranch)
 	upstream?: RemoteBranch;
 	conflicted!: boolean;
+	// TODO: to be removed from the API
```
ngalaiko, line 56: @krlvi could you please fix it? remove the field from the rust structure, and then remove all the code that compiler will mark as unused

PR #1867, opened 2023-11-30
pushing all remotes

extrawurst:
* handle network error gracefully
* clearify error handling
* cleanup code
* unittest for pushing fetched remotes that are on refs that were never checkedout


ngalaiko (approved):
:shipit:

PR #1890, opened 2023-12-03
DropDown and Segmented control

PavelLaptev:
- segmented control — centered icons
- some CSS fixes
- updated `DropDown` component. Now the component uses `Button` component. Did it in order to reduce code duplication.

### DIFF
--- a/packages/ui/src/lib/components/DropDown.svelte
+++ b/packages/ui/src/lib/components/DropDown.svelte
@@ -1,8 +1,11 @@
 <script lang="ts">
+	import { joinClassNames } from '$lib/utils/joinClassNames';
 	import Icon from '$lib/icons/Icon.svelte';
+	import Button from '$lib/components/Button.svelte';
 	import { clickOutside } from '$lib/clickOutside';
 
-	export let type: 'filled' | 'outlined' = 'filled';
+	export let color: 'primary' | 'error' = 'primary';
+	export let kind: 'filled' | 'outlined' = 'filled';
 	export let disabled = false;
 	export let loading = false;
 	export let wide = false;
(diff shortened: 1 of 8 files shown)


packages/ui/src/lib/components/DropDown.svelte
```diff
@@ -51,88 +59,47 @@
...
+		&:hover {
+			z-index: 2;
 		}
 	}
 
-	.label {
-		display: inline-flex;
-		padding: 0 var(--space-2);
-	}
+	.dropdown-wrapper :global(.dropdown__icon-btn) {
```
mtsgrd, line 77: Could we try and do this differently so we don't have to rely on `:global()`?

packages/ui/src/lib/components/DropDown.svelte
```diff
@@ -51,88 +59,47 @@
...
+		z-index: 1;
```
mtsgrd, line 78: Why is this necessary?

packages/ui/src/styles/text-classes.css
```diff
@@ -119,6 +119,7 @@
 	font-weight: 700;
 	line-height: 100%;
 }
+sdf
```
mtsgrd, line 122: This looks accidental.

packages/ui/src/lib/utils/joinClassNames.ts
```diff
@@ -0,0 +1,3 @@
+export function joinClassNames(classNames: string[]): string {
+	return classNames.filter(Boolean).join(' ');
```
mtsgrd, line 2: What is this?

packages/ui/src/lib/icons/Icon.svelte
```diff
@@ -12,11 +12,8 @@
 </script>
 
 <svg
-	class={className}
+	class={`icon-wrapper ${className}`}
```
mtsgrd, line 15: Let's discuss this at some point. I think I'm slowly turning against this pattern, there seems to always be a cleaner solution imho.

packages/ui/src/lib/components/DropDown.svelte
PavelLaptev: Because Svelte isolates class names, you can't pass a class from one component to another, like you do for example in React. But don't worry, this `:global()` thing will only affect this component. Other components with the same names won't be affected. It's isolated to this component.

Here is an example: https://svelte.dev/repl/c16aa7a6ef714067af6f5eee0a0f453c?version=3.29.0

packages/ui/src/lib/utils/joinClassNames.ts
PavelLaptev: I use it here:
image

I tried to add classes like this `class:outlined={kind}` but seems like it doesn't work for components
image

packages/ui/src/lib/icons/Icon.svelte
PavelLaptev: Yes, let's discuss this tomorrow 🙂

packages/ui/src/lib/utils/joinClassNames.ts
PavelLaptev: We can change the approach. It's just how I did it here.

packages/ui/src/lib/components/DropDown.svelte
mtsgrd: The point I'm making isn't about working/not working, or affecting other components, but rather about readability and maintainability. It gets difficult to reason about the style of components when classes are passed in dynamically.

packages/ui/src/styles/text-classes.css
PavelLaptev: yes. that's already removed

packages/ui/src/lib/components/DropDown.svelte
PavelLaptev: Yes, I need to bring the hovered button to the top

image

It's because I'm overlapping one outline button with another

image

packages/ui/src/lib/utils/joinClassNames.ts
mtsgrd: Oh sorry, that was too ambiguous. I was looking at filtering using `Boolean`. At first I didn't understand how it's related, but I see it's for filtering out any empty/undefined items. But this doesn't add value here, and imho more appropriate when you're working on a version library used by other people. Let's talk about this one too over ☕.

PR #1960, opened 2023-12-07
GB-700: support user triggered pre commit hook

extrawurst:


packages/tauri/src/git/repository.rs
```diff
@@ -421,6 +422,11 @@ impl Repository {
             .map(|iter| iter.map(|reference| reference.map(Into::into).map_err(Into::into)))
             .map_err(Into::into)
     }
+
+    pub fn run_hook_pre_commit(&self) -> Result<HookResult> {
+        let res = git2_hooks::hooks_pre_commit(&self.0, Some(&["../.husky"]))?;
```
ngalaiko, line 427: what is .husky?

packages/tauri/src/virtual_branches/virtual.rs
```diff
@@ -2006,6 +2007,15 @@ pub fn commit(
     signing_key: Option<&keys::PrivateKey>,
     user: Option<&users::User>,
 ) -> Result<git::Oid, errors::CommitError> {
+    let hook_result = project_repository
+        .git_repository
+        .run_hook_pre_commit()
```
ngalaiko, line 2012: this runs hook on the whole repository, i suppose we want to do it per vbranch instead, right?

ngalaiko (approved):

PR #1961, opened 2023-12-07
Improve label editing using keyboard

PavelLaptev:
Current behaviour: when we hit "Enter" on the input nothing happens.

the idea:
focus the label name → hit "Enter" → mount input and `focus` it → hit "Enter" → swap input with the name label and `focus` it. So we can hit "Enter" and edit input again.

Right now I did it also with `setTimeout` but wondaring if there is another way.


packages/ui/src/routes/[projectId]/components/BranchLabel.svelte
```diff
@@ -1,15 +1,29 @@
...
 
-	function activate() {
-		active = true;
+	function activateInput() {
+		activeInput = true;
 		setTimeout(() => input.select(), 0);
 	}
+
+	function activateInputOnEnter(e: KeyboardEvent) {
+		if (e.key === 'Enter') {
```
Comment (mtsgrd) [inline packages/ui/src/routes/[projectId]/components/BranchLabel.svelte]: Please use regular equality `==` over strict since it's shorter and has the exact same result.

packages/ui/src/routes/[projectId]/components/BranchLabel.svelte
```diff
@@ -1,15 +1,29 @@
...
 	let input: HTMLInputElement;
 
-	function activate() {
-		active = true;
+	function activateInput() {
+		activeInput = true;
 		setTimeout(() => input.select(), 0);
 	}
+
+	function activateInputOnEnter(e: KeyboardEvent) {
```
Comment (mtsgrd) [inline packages/ui/src/routes/[projectId]/components/BranchLabel.svelte]: Can we inline these functions since they are short and only called from one place?
```
on:keydown={(e)=> e.key == 'Enter' && activateInput()}
```

packages/ui/src/routes/[projectId]/components/BranchLabel.svelte
```diff
@@ -1,15 +1,29 @@
...
+		if (e.key === 'Enter') {
+			activateInput();
+		}
+	}
+
+	function selectNameLabelOnEnter(e: KeyboardEvent) {
```
Comment (mtsgrd) [inline packages/ui/src/routes/[projectId]/components/BranchLabel.svelte:12]: Yeah `setTimeout` is fine here, but like above it might be nicer to just inline this short thing.

```
on:keydown={(e)=> {
	if (e.key == 'Enter') {
		activeInput = false;
		setTimeout(() => label.focus(), 0);
	}
}
```

packages/ui/src/routes/[projectId]/components/BranchLabel.svelte
```diff
@@ -1,15 +1,29 @@
...
+		if (e.key === 'Enter') {
+			activeInput = false;
+			setTimeout(() => label.focus(), 0);
+		}
+	}
 </script>
 
-{#if active}
+{#if activeInput}
```
Comment (mtsgrd) [inline packages/ui/src/routes/[projectId]/components/BranchLabel.svelte]: Can you change this to `inputActive`? `activeInput` sounds like a reference to a currently active input, while `inputActive` sounds more like a boolean state.

packages/ui/src/routes/[projectId]/components/BranchLabel.svelte
```diff
@@ -19,18 +33,20 @@
 		class="branch-name-input text-base-13"
 		on:dblclick|stopPropagation
 		on:click={(e) => e.currentTarget.select()}
-		on:blur={() => (active = false)}
+		on:blur={() => (activeInput = false)}
+		on:keydown={selectNameLabelOnEnter}
```
Comment (mtsgrd) [inline packages/ui/src/routes/[projectId]/components/BranchLabel.svelte:31]: The correct term here is focus, select is something you can only do on input elements.

mtsgrd (approved):
Hey, this is great!

Feel free to ignore my feedback, I'm just thinking out loud so you and I can eventually converge on patterns and styles.

Also, feel free to pick apart my feedback if I'm saying silly things.

packages/ui/src/routes/[projectId]/components/BranchLabel.svelte
Comment (PavelLaptev) [reply packages/ui/src/routes/[projectId]/components/BranchLabel.svelte:31]: Didn't change anything here except the name of the variable `active`

packages/ui/src/routes/[projectId]/components/BranchLabel.svelte
Comment (PavelLaptev) [reply packages/ui/src/routes/[projectId]/components/BranchLabel.svelte:12]: I prefer wrapping long functions into separate ones because it makes it easier to distinguish them from the surrounding code. Let's discuss it next week 🙂

Also, here I followed the pattern in the code; we already had this as a separate function.
image

Comment (mtsgrd) [reply packages/ui/src/routes/[projectId]/components/BranchLabel.svelte:12]: This is true, but one difference was that it's called from multiple locations.

packages/ui/src/routes/[projectId]/components/BranchLabel.svelte
Comment (mtsgrd) [reply packages/ui/src/routes/[projectId]/components/BranchLabel.svelte:31]: I should have commented above, the new function you introduced is called "select...", and that's not the action you're performing. It's a focus(), not a select().

Comment (PavelLaptev) [reply packages/ui/src/routes/[projectId]/components/BranchLabel.svelte:31]: Ah, sorry, got it now. I'll fix it

packages/ui/src/routes/[projectId]/components/BranchLabel.svelte
Comment (PavelLaptev) [reply packages/ui/src/routes/[projectId]/components/BranchLabel.svelte:12]: I see, thanks

PR #2011, opened 2023-12-12
Add force push preference

krlvi:


ngalaiko (approved):
i'd prefer simple bool instead, up to you though

PR #2018, opened 2023-12-13
Add test for diff correct after update GB-716

ngalaiko:


packages/ui/src/lib/vbranches/types.ts
```diff
@@ -156,7 +156,11 @@ export class BaseBranch {
 	upstreamCommits!: RemoteCommit[];
 	@Type(() => RemoteCommit)
 	recentCommits!: RemoteCommit[];
-	fetchedAt!: Date;
+	lastFetchedMs?: number;
```
mtsgrd, line 159: We use the suffix `At` for timestamps fyi.

PR #2114, opened 2023-12-20
refactor reader / writer

ngalaiko:
todo:
* [x] implement writer.batch
* [x] use write.batch in all places where it must write atomically (session files, delta, session, branch, branch.target)


packages/tauri/src/deltas/reader.rs
```diff
@@ -1,58 +1,89 @@
...
+    reader: &'reader reader::Reader<'reader>,
 }
 
-impl<'reader> DeltasReader<'reader> {
-    pub fn new(reader: &'reader dyn reader::Reader) -> Self {
+impl<'reader> From<&'reader reader::Reader<'reader>> for DeltasReader<'reader> {
+    fn from(reader: &'reader reader::Reader<'reader>) -> Self {
         DeltasReader { reader }
     }
+}
```
Qix-: Perhaps we can just express this with a `with_reader()` method instead of an `impl From`? I'm bikeshedding a bit but `From` is typically to describe transformation, not encapsulation, since anything that `impl From<T> for U` has the blanket implementation of `impl Into<U> for T`.

This would mean you'd have to do `DeltasReader::with_reader(reader)` each time, which IMO is a lot clearer. I know that writing `.into()` is faster to type but it's not any clearer. Just my opinion, though.

packages/tauri/src/deltas/reader.rs
```diff
@@ -1,58 +1,89 @@
...
+
+#[derive(thiserror::Error, Debug)]
+pub enum ReadError {
+    #[error("not found")]
+    NotFound,
+    #[error(transparent)]
+    Other(#[from] anyhow::Error),
+}
```
Qix-: Since we're already using `anyhow` here, why not just have methods return an `anyhow::Result<>` and leave this to just a single variant `ReadError::NotFound`? That would remove a level of indirection for other error types since `anyhow` errors use `Box` under the hood.

packages/tauri/src/deltas/reader.rs
```diff
@@ -1,58 +1,89 @@
...
+
+impl<'reader> DeltasReader<'reader> {
+    pub fn new(reader: &'reader sessions::Reader<'reader>) -> Self {
```
Qix-: Is there a reason why `DeltasReader` doesn't take ownership of the reader? Does the reader need to be re-used after the deltas reader has consumed it?

ngalaiko: it might be. you could read deltas + something else from the same session reader

packages/tauri/src/deltas/reader.rs
ngalaiko: it's because i want to get rid of anyhow at some point. i thought this will make is simpler in the future, no?

packages/tauri/src/deltas/reader.rs
ngalaiko: that sounds good, we can make it separately

packages/tauri/src/deltas/reader.rs
Qix-: Ehh no, in this case it'd actually be harder because we'd have to change any match statements against a `ReadError` type. Letting the `anyhow::Result<>` take care of encapsulating all types instead of doing the encapsulation ourselves will make the migration easier down the road.

ngalaiko: maybe i don't get something. How would I handle NotFound variant when calling `read()` for example, if it returns anyhow?

packages/tauri/src/deltas/reader.rs
ngalaiko: https://github.com/gitbutlerapp/gitbutler-client/pull/2187

PR #2200, opened 2024-01-08
Add test for locking disjointed hunks

krlvi:


gitbutler-app/tests/virtual_branches.rs
```diff
@@ -110,6 +110,136 @@ mod create_commit {
...
+                .list_virtual_branches(&project_id)
+                .await
+                .unwrap()
+                .into_iter()
+                .find(|b| b.id == branch_id)
+                .unwrap();
+            assert_eq!(branch.files.len(), 1);
+            assert_eq!(branch.files[0].path.display().to_string(), "file.txt");
+            assert_eq!(branch.files[0].hunks.len(), 1);
+            // TODO: We lock this hunk, but can we afford not lock it?
```
krlvi, line 217: @ngalaiko would we break something if we dont lock hunks when the overlap is in the context?

ngalaiko: you can see how hunks locking is done [here](https://github.com/gitbutlerapp/gitbutler-client/blob/aefeba8/gitbutler-app/src/virtual_branches/virtual.rs#L955-L965)

1. we take diff between target and branch tree
1. we take diff between branch head and branch tree
1. hunks that are in `2`, but not in `1` are locked. we compare by diff.

we might be able to solve it by setting diff context to 0, not sure how it will affect other things though

PR #2226, opened 2024-01-11
GB-781 Add is default to branch

ngalaiko:
1. add `is_default` field to virtual branch, indicating if it's a selected default target for new hunks
1. add is_default to create / update parameters
1. consider is_default when looking for a default branch target

### DIFF
--- a/gitbutler-app/src/virtual_branches/branch.rs
+++ b/gitbutler-app/src/virtual_branches/branch.rs
@@ -40,8 +40,10 @@ pub struct Branch {
     pub ownership: Ownership,
     // order is the number by which UI should sort branches
     pub order: usize,
-    // is default is true, this branch will become an owner of any new hunks.
-    pub is_default: bool,
+    // is true, this branch will become an owner of any new hunks.
+    // when more then one branch has this field set to true (shouldn't happen)
+    // the branch with the lowest order will be selected
+    pub selected_for_changes: bool,
 }
 
 impl Branch {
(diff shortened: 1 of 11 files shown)


gitbutler-app/src/virtual_branches/branch.rs
```diff
@@ -40,6 +40,8 @@ pub struct Branch {
     pub ownership: Ownership,
     // order is the number by which UI should sort branches
     pub order: usize,
+    // is default is true, this branch will become an owner of any new hunks.
+    pub is_default: bool,
```
Qix-: So the problem with adding this as a property to the vbranch itself is that if there are multiple that are `true` then it's kind of undefinable behavior what happens.

To make this a tiny bit more robust, could we instead make this a `u64` and populate it with the current timestamp in milliseconds? Then when we iterate over the branches to see which one is the default, we just take the highest number (the latest). It'll at least make it a tad more robust for the user.

Just a thought :)

ngalaiko: it is accounted for, because we have an _order_ field. currently we pick first branch, when ordered by number. this will protect us in the same way

Qix-: Right but in a race condition that's not always the correct choice. The correct choice isn't "default and left-most" it's "the most recent default the user set", right?

ngalaiko: updated, check it out

Qix-: That should work :)

PR #2540, opened 2024-02-05
adjust sentry rate limiter

ngalaiko:


krlvi (approved):
seems like reasonable limits

PR #2552, opened 2024-02-06
GB-818 Add test for unapplying ownership

ngalaiko:


gitbutler-app/src/git/diff.rs
```diff
@@ -259,6 +259,81 @@ fn hunks_by_filepath(
...
+        Some("@@") => {}
+        _ => return None,
+    };
+
+    Some(format!(
+        "@@ {} {} @@ {}",
+        new_range.replace('+', "-"),
+        old_range.replace('-', "+"),
+        parts.join(" ")
+    ))
```
Qix-: You can condense a lot of those with...

```rust
let new_range = parts.next()?;
```

ngalaiko: only two, others i need to check

PR #2562, opened 2024-02-07
update dependabot schedule and grouping

Qix-:
Right now the PRs are a bit time consuming and mostly involve patch versions. Going to set this to monthly until we can figure out a better way to handle them.


Qix-:
Will also need to figure out if there's a better way to handle security updates on a more urgent basis. Not sure if that's possible; if someone knows, please let us know!

PR #2581, opened 2024-02-09
Replace fork of Context Menu Plugin with Upstream

namanlp:
Hi there!

So, Tauri Plugin Context Menu do have Linux Support now 🎉 

Hence, the fork of it can be replaced with the upstream. I have not tested it yet though, and hence, making it as a draft PR for now. 

Also, now looking at the source, I am wondering why `tauri-plugins` are explicitly fetched from git? This may cause issue because they might use unreleased version.

Thank You


namanlp:
Ok, So I am do not have much Svelte knowledge, hence can't help you guys to make context menu work on Linux.

Qix-:
You'll have to make sure Cargo.lock is updated prior to pushing - we use `--locked` in CI. :)

gitbutler-app/Cargo.toml
```diff
@@ -66,7 +66,7 @@ slug = "0.1.5"
 ssh-key = { version = "0.6.4", features = [ "alloc", "ed25519" ] }
 ssh2 = { version = "0.9.4", features = ["vendored-openssl"] }
 tauri = { version = "1.5.4", features = ["dialog-open", "fs-read-file", "path-all", "process-relaunch", "protocol-asset", "shell-open", "system-tray", "window-maximize", "window-start-dragging", "window-unmaximize"] }
```
qpgt2, line 68: 👍

qpgt2, line 68: 👍

PR #2635, opened 2024-02-13
chore: create tar.gz Linux release

p5:
Creates a .tar.gz release so others can package and release GB for other distros.


scripts/release.sh
p5, line 1: Please don't do anything with this PR yet - I'm trying to get CI running on my fork's PR to test out these builds
https://github.com/p5/gitbutler/pull/1

PR #2688, opened 2024-02-14
Fix branch name change when pressing Enter

### LINKED ISSUE #2743: When setting branch name, enter should "submit" the input, currently it does nothing

askielboe:
This fixes a bug where the branch name could not be changed by pressing Enter after typing a new branch name in the `BranchHeader`'s `BranchLabel` field.

I believe the issue is that the `<input/>` field is unmounted by setting `inputActive = false` in `on:keydown` before the `on:change` function can be executed.

Running `inputActive = false` asynchronously, as is already done for `label.focus()`, seems to fix the issue for me.

I realize there may be better ways to fix this, but this is all I had time for. I also added a comment to explain the change but feel free to edit or remove it if it doesn't fit your style.

---

Versions of GitButler and OS where this was reproduced and fixed:

macOS 14.2.1 (23C71).
GitButler Version 0.10.7 (20240213.214246).
GitButler `master` as of now.

---

Discussed (briefly) on Discord: https://discord.com/channels/1060193121130000425/1206670506271707156/1207405803661951036.

---

Closes #2743.

### DIFF
--- a/gitbutler-ui/src/lib/components/BranchLabel.svelte
+++ b/gitbutler-ui/src/lib/components/BranchLabel.svelte
@@ -44,7 +44,8 @@
 		on:blur={() => (inputActive = false)}
 		on:keydown={(e) => {
 			if (e.key == 'Enter') {
-				inputActive = false;
+				// Unmount input field asynchronously to ensure on:change gets executed.
+				setTimeout(() => (inputActive = false), 0);
 				setTimeout(() => label.focus(), 0);
 			}
 


gitbutler-ui/src/lib/components/BranchLabel.svelte
```diff
@@ -44,7 +44,8 @@
 		on:blur={() => (inputActive = false)}
 		on:keydown={(e) => {
 			if (e.key == 'Enter') {
-				inputActive = false;
+				// Unmount input field asynchronously to ensure on:change gets executed.
+				setTimeout(() => (inputActive = false), 0);
```
mtsgrd, line 48: Before we merge, can we maybe combine it with the setTimeout on the next line?

askielboe: I tried that, but `label` is only bound when the `<div/>` is mounted, which only happens after `inputActive` is set to `false`. So unfortunately we have to wait for the next event loop for `label` to not be null.

askielboe: A better solution would probably be to try to avoid unmounting the `input` field at all, but I can see why it was done this way.

krlvi: i suggest we could probably refactor this down the line if needed

PR #2689, opened 2024-02-14
Replace fork of Context Menu Plugin with Upstream

namanlp:
Hi there!

(This PR is copy of https://github.com/gitbutlerapp/gitbutler/pull/2581, but is cleaner and cargo.lock is also updated )

So, Tauri Plugin Context Menu do have Linux Support now 🎉

Hence, the fork of it can be replaced with the upstream. I have not tested it yet though, and hence, making it as a draft PR for now.

Also, now looking at the source, I am wondering why tauri-plugins are explicitly fetched from git? This may cause issue because they might use unreleased version.

Thank You

### DIFF
--- a/gitbutler-app/Cargo.toml
+++ b/gitbutler-app/Cargo.toml
@@ -66,7 +66,7 @@ slug = "0.1.5"
 ssh-key = { version = "0.6.4", features = [ "alloc", "ed25519" ] }
 ssh2 = { version = "0.9.4", features = ["vendored-openssl"] }
 tauri = { version = "1.5.4", features = ["dialog-open", "fs-read-file", "path-all", "process-relaunch", "protocol-asset", "shell-open", "system-tray", "window-maximize", "window-start-dragging", "window-unmaximize"] }
-tauri-plugin-context-menu = { git = "https://github.com/gitbutlerapp/tauri-plugin-context-menu", branch = "main" }
+tauri-plugin-context-menu = { git = "https://github.com/c2r0b/tauri-plugin-context-menu", branch = "main" }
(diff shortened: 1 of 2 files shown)


gitbutler-app/Cargo.toml
```diff
@@ -66,7 +66,7 @@ slug = "0.1.5"
 ssh-key = { version = "0.6.4", features = [ "alloc", "ed25519" ] }
 ssh2 = { version = "0.9.4", features = ["vendored-openssl"] }
 tauri = { version = "1.5.4", features = ["dialog-open", "fs-read-file", "path-all", "process-relaunch", "protocol-asset", "shell-open", "system-tray", "window-maximize", "window-start-dragging", "window-unmaximize"] }
-tauri-plugin-context-menu = { git = "https://github.com/gitbutlerapp/tauri-plugin-context-menu", branch = "main" }
+tauri-plugin-context-menu = { git = "https://github.com/c2r0b/tauri-plugin-context-menu", branch = "main" }
```
03juan, line 69:
```suggestion
tauri-plugin-context-menu = "0.7.0"
```

It's recommended to specify official version, as per https://crates.io/crates/tauri-plugin-context-menu

Tested locally and it works fine.

03juan (changes requested):

gitbutler-app/Cargo.toml
namanlp: Well, I kept it as Github URL, because other Dependencies are Git as well. Kind off makes sense as GitButler is in Beta Phase, it is always nice to have latest version straight from the repository.

Once GitButler stabilise itself, it would be good to use stable version. 

But obviously, management has to decide on this.

03juan: Good point

krlvi (approved):
thank you !

PR #2697, opened 2024-02-15
Minor Grammatical Update to Cargo.toml

lawyered0:
Corrects the usage of the indefinite article from 'an' to 'a' before the acronym 'URL' in documentation and comments, aligning with grammatical rules regarding consonant-sounding acronyms.

Since "URL" is pronounced "you-are-el," it starts with a consonant sound, making "a" the correct article to use.


krlvi (approved):
thank you for this fix

PR #2705, opened 2024-02-15
Capture push and pr success/failure in posthog

mtsgrd:
- just a start since we currently are not getting much data on this


mtsgrd:
@krlvi this a pretty quick pr, but I reckon it's important we capture these events a bit better. The autocapture we use for commit isn't really good enough since it only signals the user clicked the button (albeit commit would rarely fail).

krlvi (approved):

PR #2714, opened 2024-02-15
fix: discard diff hunk incorrectly replaces all "-" with "+"

Meesayen:
I noticed that when discarding a diff hunk that contains dashes (i.e. a line in a package.json like this `"@typescript-eslint/eslint-plugin": "^5.62.0"`), the discard action would replace all occurrences of `-` with a `+`.

This was caused by Rust's `str::replace` being a replace all by default.

Switched to `replacen()` to only replace the first occurrence of `-` (and applied the same logic for the opposite case, `+` -> `-`)


krlvi (approved):
Thank you for fixing this!

PR #2726, opened 2024-02-15
feat: ability to discard changes to an entire file

### LINKED ISSUE #2710: Add the ability to discard changes to an entire file (similar to discarding hunks)

Meesayen:
Added the ability to discard the changes of a whole file by right-clicking it and clicking "discard" on the context menu.

Fixes https://github.com/gitbutlerapp/gitbutler/issues/2710

~This one depends on this other PR https://github.com/gitbutlerapp/gitbutler/pull/2714 to work properly.~ (done)


gitbutler-app/src/virtual_branches/commands.rs
```diff
@@ -326,6 +326,25 @@ pub async fn unapply_ownership(
...
+#[tauri::command(async)]
+#[instrument(skip(handle))]
+pub async fn unapply_file_changes(
+    handle: AppHandle,
+    project_id: &str,
+    file_path: &str,
+) -> Result<(), Error> {
+    let project_id = project_id.parse().map_err(|_| Error::UserError {
+        code: Code::Validation,
+        message: "Malformed project id".to_string(),
```
jdicke: `message: format!("Malformed project id: {}", project_id),`

ngalaiko:
> is pretty much a 1:1 copy of the unapply_ownership with the sole difference of collecting every hunk of a matching file_path

that's true. instead of having a duplicate method in rust, it's more appropriate to construct full file ownership from the current diff on the frontend and then invoke `unapply_ownership` instead.

gitbutler-app/src/virtual_branches/virtual.rs
```diff
@@ -494,11 +494,15 @@ pub fn unapply_ownership(
...
-                                hunks_to_unapply.push((path.clone(), hunk));
-                            }
+                    let ownership_hunks: Vec<&Hunk> = ownership
+                        .files
+                        .iter()
+                        .filter(|o| o.file_path == path)
+                        .flat_map(|f| &f.hunks)
+                        .collect();
+                    for hunk in hunks {
+                        if ownership_hunks.contains(&&Hunk::from(&hunk)) {
```
Meesayen, line 504: Ok, removed the duplicated code and handled it like suggested: compose a list of hunks from the frontend and pass it to `unapply_ownership`.

Now I'm just not sure this code here is the best Rust code. The need for this double `&` on `&&Hunk` is giving me bad vibes.

krlvi:
I tested this with binary files (e.g. png) and they are not being handled, and we error out on those.

Another thought that I also got while playing with this is how easy it is to accidentally nuke a file, perhaps we should have a confirmation prompt (like on branch deletion), at least until we can support an "undo" for operations

Meesayen:
@krlvi I've added the confirmation modal and also handled multi-selection of files (via shift click). 
For binary files, I'm now disabling the discard button, until the Rust api can suppor that kind of reverse diff.

I also noticed Added and Deleted file changes are not handled correctly in general: if we hit discard on an Added file (either via this new context menu or the diff panel) the content of the file is erased but the file is kept there; Deleted files, instead, are not restored at all.

Fixing this is probably a separate issue, but surely this new feature highlights the problem a bit more, unfortunately :/

krlvi:
@Meesayen tested it out, it works great! Such an awesome contribution!

krlvi (approved):

krlvi (approved):

krlvi:
hmm, there still seem to be conflicts

Meesayen:
Sorry, I just saw the comment. Thanks for fixing the conflicts!

krlvi:
@Meesayen 
image
still conflicts, this is weird

Meesayen:
Resolved.

krlvi (approved):

krlvi:
> Resolved.

github _must_ be glitching out... it is showing conflicts again, without specifying a file, and nothing else was merged

Issue #2709, opened 2024-02-15
Improve Branch Search and Case Sensitivity in Branch View


gitbutler-client:
For branches that have PRs, in the branch view, I can't search for their branch names. I can see the branch name if I click on them, bit its quite cumbersome. It would be neat if it still showed their names below the PR titles, and if I could still search for them.

The search box also seems to be case-sensitive, which would be fine, if it wasn't for the fact that macOS auto correct is coming in and capitalising the first letter of what I type in 😆

https://discord.com/channels/1060193121130000425/1073202153163857920/1207643377890369537

krlvi:
fixed in https://github.com/gitbutlerapp/gitbutler/pull/2732

PR #2732, opened 2024-02-15
Ensure searches on PRs also includes the branch name

Caleb-T-Owens:
* Related https://github.com/gitbutlerapp/gitbutler/issues/2709

## The problem

For PRs in the branch list, I'm unable to search by the branch name; currently, I can only search by the PR title.

## The solution

I thought about how to do this and I didn't want to encode too many of the specifics of searching onto the CombinedBranch class, so I thought a happy medium would be to have a getter to access the various identifiers which would be relevant for searching, and then make use of them in the Branches component.

### DIFF
--- a/gitbutler-ui/src/lib/branches/types.ts
+++ b/gitbutler-ui/src/lib/branches/types.ts
@@ -85,6 +85,19 @@ export class CombinedBranch {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


krlvi (approved):
<3 thank you for this

Issue #2728, opened 2024-02-15
bug: quotes are changing from " to « by themselves


nikolaistarostin:
I am trying to type quotes (") in commit message, but the keep changing to different quotes («»)

I have ABC (English) and Russian keyboards enabled on my laptop (MacBook Pro, macOS 14.1.1), and I am typing using ABC keyboard.

Here is a video with an example.
https://github.com/gitbutlerapp/gitbutler/assets/48692605/e53aef8e-ada3-4a26-b072-2dc209c4622b

03juan:
Is it possible you have smart quotes enabled for double quotes in your keyboard input sources settings?

Refs:
https://support.apple.com/en-gb/guide/mac-help/mchl84525d76/14.0/mac/14.0
https://support.apple.com/en-gb/guide/mac-help/mh35735/14.0/mac/14.0

krlvi:
I think we need to do `autocomplete='off' spellcheck='false' autocorrect='off'` on inputs

krlvi:
This should be fixed by https://github.com/gitbutlerapp/gitbutler/pull/2737
I'm going to make a release tomorrow morning

PR #2737, opened 2024-02-15
Disable auto completions in branch filter field

Caleb-T-Owens:
* Related https://github.com/gitbutlerapp/gitbutler/issues/2709

## The problem

macOs spell checker is correcting words and capitalisation in the branch name filter.

Often we use incomplete words, non-word acronyms, and search terms that don't start with capitals, and it can be quite bothersome.

## The solution

The solution was to make use of the spellCheck property which does pretty much what it says on the tin

### DIFF
--- a/gitbutler-ui/src/lib/components/Branches.svelte
+++ b/gitbutler-ui/src/lib/components/Branches.svelte
@@ -163,7 +163,6 @@
 					<TextBox
 						icon="filter"
 						placeholder="Search"
-						spellcheck={false}
 						on:input={(e) => textFilter$.next(e.detail)}
 					/>
 					<div bind:this={contents} class="content">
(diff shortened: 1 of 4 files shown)


krlvi:
Nice! Thank you for fixing this. By the way, if it's of interest, theres also this issue, which is very much the same thing, just in another place of the app: https://github.com/gitbutlerapp/gitbutler/issues/2728

Caleb-T-Owens:
@krlvi Interesting. I would have thought that the commit message was one of the places you do want spell check on 😆.

Regardless, is it desired that spellCheck be false by default?

krlvi:
I imagine a lot of people would wanna put things like names of subsystems, ticket numbers or other technical things in the commit messages, making spellchecking there more annoying than helpful.

Caleb-T-Owens:
Makes sense - I'll go make sure that spellcheck=false is set everywhere then (taking a look, the git commit message is its own text area element)

krlvi (approved):

PR #2747, opened 2024-02-16
File diff removed lines prefix should be -

askielboe:
Before:

image

After:

image


krlvi (approved):
thank you so much for fixing this!

Issue #2621, opened 2024-02-13
Support moving commits between virtual branches


krlvi:
It would be useful to be able to drag at leat the topmost commit between vbranches.

PR #2750, opened 2024-02-16
move commit functionality

ngalaiko:
https://github.com/gitbutlerapp/gitbutler/issues/2621

### DIFF
--- a/gitbutler-app/src/bin.rs
+++ b/gitbutler-app/src/bin.rs
@@ -217,6 +217,7 @@ fn main() {
                     virtual_branches::commands::get_remote_branch_data,
                     virtual_branches::commands::squash_branch_commit,
                     virtual_branches::commands::fetch_from_target,
+                    virtual_branches::commands::move_commit,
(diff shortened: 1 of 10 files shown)


ngalaiko:
@mtsgrd this will need some frontend inputs, perhaps refactoring. we have to have a way to verify that moving commit is only possible for: 
* a head commit (that is done)
* that does not have any hunks locked to self <- this should be implemented

another unknown for me from a design perspective is where to display dropzone for the commit. now it's rendered on top of the files list, but maybe is there are local commits present, we should render it inside of the local commits box instead?

in general, it looks like since the lane was split into boxes with files, local commits, remote commits, etc. some drop zones render in a wrong place (like for example cherry picking drop zone)

mtsgrd:
Hey, thanks. Does it need to refactored prior to landing this or after?

ngalaiko:
after would be fine.

krlvi (approved):

PR #2751, opened 2024-02-16
Fix hunk grab cursors

askielboe:
This PR fixes two issues with the cursor style for dragging hunks:

1. Fixes issue where the grab cursor would be shown when hovering diff lines, even though dragging was disabled (e.g. if the hunk was locked).
2. Fixes issue where a grab cursor would never show when hovering hunk diff line numbers, even though dragging on the line numbers was allowed.

Discord discussion: https://discord.com/channels/1060193121130000425/1208015050082619423.


krlvi (approved):
Nice fix, thank you very much

PR #2754, opened 2024-02-16
Fix push button behaviour after undoing a commit

### LINKED ISSUE #2733: When a (pushed) commit is removed from a virtual branch there should be an option to force push

Caleb-T-Owens:
* Closes https://github.com/gitbutlerapp/gitbutler/issues/2733

## The problem

Currently in git butler, after undoing a commit from the remote, there is no option to force push to get rid of the commit that is still on the remote.

## The solution

The solution was to change the logic of the PushButton to include a new state that will show if there should be the option to force push

## Video of the change in action:

https://youtu.be/puGAWTrq6xM

### DIFF
--- a/gitbutler-ui/src/lib/components/BranchCommits.svelte
+++ b/gitbutler-ui/src/lib/components/BranchCommits.svelte
@@ -18,50 +18,48 @@
 	export let branchCount: number;
 </script>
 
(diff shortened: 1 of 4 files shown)


krlvi (approved):
tested this and it's working for me too! thanks for implementing it :)

PR #2764, opened 2024-02-16
Update the default value for signCommits in frontend to false to match backend

### LINKED ISSUE #2763: Commit signing appear to no longer work

Caleb-T-Owens:
* Closes https://github.com/gitbutlerapp/gitbutler/issues/2763

## The problem

Enabling signCommits for the first time doesn't seem to enable signing commits.

## The solution

In the backend, a nil value was interpreted as false, whereas in the frontend it is interpreted as true, and it seems like this dissonance was the cause of the problem so I've chosen to default to false in both locations.


krlvi (approved):
🚀

PR #2766, opened 2024-02-16
allow periods in branch names

krlvi:


gitbutler-app/src/virtual_branches/virtual.rs
```diff
@@ -155,8 +155,7 @@ impl From<git::Signature<'_>> for Author {
 }
 
 pub fn normalize_branch_name(name: &str) -> String {
-    // Technically this pattern should include ".", but it makes libgit2 throws an error
-    let pattern = Regex::new("[^A-Za-z0-9_/]+").unwrap();
+    let pattern = Regex::new("[^A-Za-z0-9_/.]+").unwrap();
```
mtsgrd, line 158: There is a corresponding regex in typescript that also needed to be updated: https://github.com/gitbutlerapp/gitbutler/pull/2827

PR #2775, opened 2024-02-17
Add loading state to PushButton when in "pr" state

Caleb-T-Owens:
## The problem

When clicking the push button in the PR state, it feels unresponsive because it doesn't have the loading state set up

## The solution

I added it :D


krlvi (approved):
thank you!

PR #2777, opened 2024-02-17
Only show "Force push to remote" when pushing zero commits to a branch

Caleb-T-Owens:
## The problem

GitHub won't let you make a PR with zero commits in it, so when we have the PushButton component rendered on a remote branch with zero commits, we shouldn't give the option to make a PR.

Example: I've pushed a commit to a remote, and undone it. Having the option to create a PR is non-sensical, and confusing if its actually defaulted to that option on you.
Screenshot 2024-02-17 at 12 33 04

## The solution

I ended up passing in the branch and type into the PushButton component because there was a lot of PushButton specific logic sitting in the CommitListFooter, so moving it inside the PushButton felt like a sensible move.

I've added a second condition for only showing the single "Force push to remote" button.
Previously, it was:
- If the branch is a PR && its not pushed
Now its shown if either these conditions are met:
- If the branch is a PR && its not pushed
- If the branch has no commits && its not pushed

## Demo

In this screenshot, I've pushed a commit to remote and undone the commit. I'm no longer provided the option to make a PR.

Screenshot 2024-02-17 at 12 31 49

### DIFF
--- a/gitbutler-ui/src/lib/components/CommitListFooter.svelte
+++ b/gitbutler-ui/src/lib/components/CommitListFooter.svelte
@@ -75,9 +75,9 @@
 			<PushButton
 				wide
 				isLoading={isPushing || $githubServiceState$?.busy}
-				isPushed={type == 'remote' && !branch.requiresForce}
-				requiresForcePush={branch.requiresForce}
 				isPr={!!$pr$}
+				{type}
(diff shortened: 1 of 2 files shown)


krlvi (approved):
thanks for handling this

PR #2785, opened 2024-02-17
Remove tray menu functionality

krlvi:
Realising that the tray icon is not very valuable functionality-wise, creates clutter and there are a few bugs related to it

@ngalaiko i hope i didnt delete something i shouldnt have by accident

### DIFF
--- a/gitbutler-app/src/bin.rs
+++ b/gitbutler-app/src/bin.rs
@@ -20,43 +20,8 @@ fn main() {
         .unwrap()
         .block_on(async {
             tauri::async_runtime::set(tokio::runtime::Handle::current());
-            let app_title = tauri_context.package_info().name.clone();
-
-            let quit = tauri::CustomMenuItem::new("quit".to_string(), "Quit");
-            let hide =
-                tauri::CustomMenuItem::new("toggle".to_string(), format!("Hide {}", app_title));
-            let tray_menu = tauri::SystemTrayMenu::new().add_item(hide).add_item(quit);
-            let tray = tauri::SystemTray::new().with_menu(tray_menu);
 
             tauri::Builder::default()
-                .system_tray(tray)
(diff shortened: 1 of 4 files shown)


gitbutler-app/src/bin.rs
```diff
@@ -226,32 +188,10 @@ fn main() {
...
-                                .expect("Failed to set toggle menu hide");
-                        } else {
-                            set_toggle_menu_show(app_handle)
-                                .expect("Failed to set toggle menu show");
-                        }
-                    }
-                    tauri::RunEvent::ExitRequested { api, .. } => {
-                        hide_window(app_handle).expect("Failed to hide window");
-                        api.prevent_exit();
-                    }
```
ngalaiko: @krlvi i think this would virtually disable background mode which is a core mac experience. meaning when you click `close`, the app will not just be hidden, but instead stop completely.

navtoj: > @krlvi i think this would virtually disable background mode which is a core mac experience. meaning when you click `close`, the app will not just be hidden, but instead stop completely.

the `minimize` button is for hiding, `close` is for closing

ngalaiko: > > @krlvi i think this would virtually disable background mode which is a core mac experience. meaning when you click `close`, the app will not just be hidden, but instead stop completely.
> 
> the `minimize` button is for hiding, `close` is for closing

not on mac though. close on mac closes the window, not the process. the process remains in the dock

krlvi: you are right... without this the red icon kills the process (not the case for VSCode). gonna re-add it

navtoj: @krlvi tauri can't keep the app process running without the tray menu?

PR #2791, opened 2024-02-18
fix: show better error message when loading a repository without commits

### LINKED ISSUE #2778: Better user error message when an added repository has no commits in it

Pranav2612000:
Fixes #2778


Pranav2612000:
Would love to get a review! @krlvi
There may be a better way to do it ( and with proper conventions ). Would love to get your thoughts

krlvi (approved):
thank you for this - it's deff an improvement over the state before

Issue #2718, opened 2024-02-15
Deleting an empty virtual branch shouldn't show a warning


TheKnarf:
- If you double-click the "New branch+" button you might get two new virtual branches

- Getting a warning to when trying to delete an empty virtual branch is just annoying and makes the experience less smooth

mtsgrd:
Fixed in #2801

PR #2801, opened 2024-02-18
 fix - Avoid showing delete branch modal for an empty virtual branch.

AbhishekThorat:
Fix for #2718

### DIFF
--- a/gitbutler-ui/src/lib/components/BranchLanePopupMenu.svelte
+++ b/gitbutler-ui/src/lib/components/BranchLanePopupMenu.svelte
@@ -47,8 +47,12 @@
 
 			<ContextMenuItem
 				label="Delete"
-				on:click={() => {
-					deleteBranchModal.show(branch);
+				on:click={async () => {
+					if (commits.length == 0 && branch.files?.length == 0) {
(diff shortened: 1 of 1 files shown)


gitbutler-ui/src/lib/components/BranchLanePopupMenu.svelte
```diff
@@ -47,8 +47,16 @@
 
 			<ContextMenuItem
 				label="Delete"
-				on:click={() => {
-					deleteBranchModal.show(branch);
+				on:click={async () => {
+					if (
+						branch.name.toLowerCase().includes('virtual branch') &&
+						commits.length == 0 &&
+						branch.files?.length == 0
```
AbhishekThorat, line 54: Not 100% sure whether this is elegant way of detecting whether virtual branch has been updated or not. Adding/maintaining property something like `touched` / [`dirty`](https://formik.org/docs/api/formik#dirty-boolean) might be an option.

mtsgrd: No this looks decent, thank you!

PR #2805, opened 2024-02-19
fix: #2799 Add a custom renderer for links, same as in ToastController

### LINKED ISSUE #2799: Links inside PR descriptions should be opened in the browser
If a PR description contains a link, as GB renders it, if a user clicks on it, it should be opened in their browsers, as opposed to GB itself.

dinolindlau:
Fixes: https://github.com/gitbutlerapp/gitbutler/issues/2799

The code I used is also inside the [ToastController](https://github.com/gitbutlerapp/gitbutler/blob/master/gitbutler-ui/src/lib/notifications/ToastController.svelte#L10-L13). I guess it would be appropriate to extract this an extra file in lib/marked ?


mtsgrd:
Nice one, thanks!

PR #2810, opened 2024-02-19
Add tooltip to show full commit message

Caleb-T-Owens:
# The problem

Git butler only shows a portion of the first line of a commit

# The solution

Implement new designs which facilitate displaying the commit's full message.

Screenshot 2024-02-19 at 22 04 52

### DIFF
--- a/gitbutler-ui/src/lib/components/CommitCard.svelte
+++ b/gitbutler-ui/src/lib/components/CommitCard.svelte
@@ -6,7 +6,6 @@
 	import { projectCurrentCommitMessage } from '$lib/config/config';
 	import { draggable } from '$lib/dragging/draggable';
 	import { draggableCommit, nonDraggable } from '$lib/dragging/draggables';
-	import { tooltip } from '$lib/utils/tooltip';
 	import { openExternalUrl } from '$lib/utils/url';
 	import { Ownership } from '$lib/vbranches/ownership';
 	import { listRemoteCommitFiles } from '$lib/vbranches/remoteCommits';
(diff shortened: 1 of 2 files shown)


mtsgrd:
Thanks, this is a good stop-gap measure! I had a look and we probably need something like this to make it behave well when the description shorter than available width.

image

@PavelLaptev in general, what could we do here? The tooltip isn't ideal for this since it's less than 80 characters in wide.

Caleb-T-Owens:
Hi @PavelLaptev! Not showing the tooltip if it provides no extra information makes sense - I'll make that change this afternoon.

I'm not so sure about cutting it off at five lines though. My thought would be that it's uncommon for a commit message to have more than 5 lines, and if there are more than 5 lines, they're probably worth reading!

PavelLaptev:
Hi @Caleb-T-Owens 🙂 I understand. Long commits are not commits issue, but UI. I'm just saying that current UI can't handle it.
Let's hold it a little I'll share a new design concept on Discord today

Caleb-T-Owens:
@PavelLaptev Sounds good 👍

PavelLaptev:
@Caleb-T-Owens here is a link to design in Figma — https://www.figma.com/file/FbeLt0yjY9RiNn8MXUXsYs/Client-Design?type=design&node-id=1408%3A60581&mode=design&t=Am6YqAdsel1jLjoB-1

image

Feel free to ask if you have any questions.
Also, you can do the logic and leave styles to me.

Caleb-T-Owens:
Hi! This is ready for review

PavelLaptev (approved):
Great PR @Caleb-T-Owens, thanks!
I just made a few updates to UI _(in Figma as well)_

I removed tailwind classes. We're planning to get rid of it in the future.

Caleb-T-Owens:
Those smaller tag buttons do look better! Thanks for that update 👍. I do wonder though if the clickable version of Tag really ought to be a state of Button...

PR #2812, opened 2024-02-19
fix: hunk and file context menu now correctly dismiss

Meesayen:
I noticed that sometimes the file and hunk context menu wouldn't dismiss itself after you take an action (i.e. click "Discard" in the menu).
I've added a `dismiss` callback to the `PopupMenu` component that consumers can use to dismiss it when the action requires no further interaction with the menu.

#### before: 
https://github.com/gitbutlerapp/gitbutler/assets/298675/07c4b023-b94b-4558-bcca-5e05338a7811

#### after:
https://github.com/gitbutlerapp/gitbutler/assets/298675/fce6bc5a-6436-4b05-ade7-995fe42ac52b

### DIFF
--- a/gitbutler-ui/src/lib/components/FileContextMenu.svelte
+++ b/gitbutler-ui/src/lib/components/FileContextMenu.svelte
@@ -21,14 +21,20 @@
 	}
 </script>
 
-<PopupMenu bind:this={popupMenu} let:item>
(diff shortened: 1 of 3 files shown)


mtsgrd:
Nice, thanks so much! :)

PR #2814, opened 2024-02-19
chore: consider "contextmenu" event as a click outside event too

Meesayen:
Follow up to this PR https://github.com/gitbutlerapp/gitbutler/pull/2812

The `clickOutside` directive should also include the `contextmenu` event to correctly trigger when the user clicks outside by right-clicking.

#### before:

https://github.com/gitbutlerapp/gitbutler/assets/298675/c9220663-3ac2-4227-aff9-6fc60459d204

#### after:

https://github.com/gitbutlerapp/gitbutler/assets/298675/5f4820da-c7b4-4a94-bcdc-eb91cd1bdc6c

### DIFF
--- a/gitbutler-ui/src/lib/clickOutside.ts
+++ b/gitbutler-ui/src/lib/clickOutside.ts
@@ -15,15 +15,19 @@ export function clickOutside(
 		}
 	}
 	document.addEventListener('click', onClick, true);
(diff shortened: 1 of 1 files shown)


mtsgrd:
Nice one, thanks for contributing! 🎉

PR #2816, opened 2024-02-19
chore: applied context menu to discard file to the tree view

Meesayen:
Follow up to this PR https://github.com/gitbutlerapp/gitbutler/pull/2726 

I've added the same context menu to the files in tree view.

Also fixed an issue where the `allowMultiple` property was not propagated to recursive instances of FileTree.


krlvi (approved):
awesome! thanks

PR #2822, opened 2024-02-19
Prevent divider line from scrolling up with branch card

Caleb-T-Owens:
# The problem

Notice how the divider line in the middle is missing in the bottom part. When you scroll the branch card, the divider line will scroll with it.
image

# The solution

The issue was caused by the divider line being inside the scrollable container rather than outside it. Moving it outside solves the issue.


Caleb-T-Owens:
Looks like this has been resolved upstream from where I was already 👍

PR #2825, opened 2024-02-19
Add option to copy PR URL to clipboard

Caleb-T-Owens:
My workflow currently requires me to copy the URL of my PRs to put them into Trello. Having to follow the link to GH to then copy it is quite cumbersome so I've added in a context menu that has an option to copy the link to the clipboard

### DIFF
--- a/gitbutler-ui/src/lib/components/ActiveBranchStatus.svelte
+++ b/gitbutler-ui/src/lib/components/ActiveBranchStatus.svelte
@@ -1,7 +1,9 @@
 <script lang="ts">
 	import Tag from '$lib/components/Tag.svelte';
+	import ViewPrContextMenu from '$lib/components/ViewPrContextMenu.svelte';
 	import { normalizeBranchName } from '$lib/utils/branch';
 	import { open } from '@tauri-apps/api/shell';
+	import { onDestroy } from 'svelte';
 	import type { Persisted } from '$lib/persisted/persisted';
 	import type { BaseBranch, Branch } from '$lib/vbranches/types';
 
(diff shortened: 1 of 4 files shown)


gitbutler-ui/src/lib/components/ActiveBranchStatus.svelte
```diff
@@ -11,6 +13,22 @@
...
+
+	function updateContextMenu() {
+		if (popupMenu) popupMenu.$destroy();
+		return new ViewPrContextMenu({
+			target: document.body,
+			props: { prUrl: prUrl || '' }
+		});
+	}
+
+	$: popupMenu = updateContextMenu();
```
mtsgrd: I haven't tested this out, but it looks like we ought to pass `prUrl` as a parameter to this function to make it react to new values?

Caleb-T-Owens: Thanks for pointing that out - I think you're probably right

mtsgrd:
Perfect, thank you!

As a side note, I'm starting to wonder if it'd be better to have one instance of each type of context menu living closer to the root in the dom.. that way we wouldn't need to render them dynamically.

Caleb-T-Owens:
@mtsgrd I was thinking perhaps something one step further would be better. I was wondering if it would make sense to have just one context menu at the top that you could pass an object which contained all the different options and handlers and it would dynamically render it out

So you'd just make one call like when you want to have a contextMenu
```ts
<MyThing
on:contextmenu={(e) => {
  openContextMenu({
    "Copy to Clipboard": () => { copyToClipboard(prUrl) }
  })
}}>
</MyThing>
```

PR #2835, opened 2024-02-20
feat: generate PR title and body from commit

### LINKED ISSUE #2795: Consider using commit message for PR title and description in case of a single commit
When creating a PR directly in GitHub (or via the GitHub CLI) for a branch with a _single_ commit, the default values of the PR title and description is the commit message—first paragraph for the title and the rest for the description.

This behavior is quite convenient.

Please consider implementing the same behavior when creating a PR via GitButler.

After a quick glance at the code, I came up with the following:

```diff
diff --git a/gitbutler-ui/src/lib/branches/service.ts b/gitbutler-ui/src/lib/branches/service.ts
--- gitbutler-ui/src/lib/branches/service.ts
+++ gitbutler-ui/src/lib/branches/service.ts
@@ -67,13 +67,25 @@
 		}
 
 		const createPrSpan = sentryTxn.startChild({ op: 'pr_api_create' });
 
+		let title = newBranch.name;
+		let body = newBranch.notes;
+
+		// In case of a single commit, use the commit summary and description for the title and
+		// description of the PR
+		if (newBranch.commits.length === 1) {
+			const paragraphs = newBranch.commits[0].description.split("\n\n");
+
+			title = paragraphs[0];
+			body = paragraphs.slice(1).join("\n\n");
+		}
+
 		try {
 			const resp = await this.githubService.createPullRequest(
 				baseBranch,
-				newBranch.name,
-				newBranch.notes,
+				title,
+				body,
 				newBranch.id,
 				newBranch.upstreamName,
 				draft
 			);
```

Please note that I haven't tested this. Let me know if you think this is a good idea and whether it's the correct approach.

Pranav2612000:
If the branch contains only 1 commit, we use the first line to generate the PR title and the remaining commit body to generate the PR body.
Co-author: @eyalch

Fixes #2795

### DIFF
--- a/gitbutler-ui/src/lib/branches/service.ts
+++ b/gitbutler-ui/src/lib/branches/service.ts
@@ -68,11 +68,22 @@ export class BranchService {
 
 		const createPrSpan = sentryTxn.startChild({ op: 'pr_api_create' });
 
+		let title = newBranch.name;
+		let body = newBranch.notes;
+
+		// In case of a single commit, use the commit summary and description for the title and
+		// description of the PR
+		if (newBranch.commits.length === 1) {
+			const commit = newBranch.commits[0];
+			if (commit.descriptionTitle) title = commit.descriptionTitle;
+			if (commit.descriptionBody) body = commit.descriptionBody;
+		}
+
 		try {
 			const resp = await this.githubService.createPullRequest(
 				baseBranch,
-				newBranch.name,
-				newBranch.notes,
+				title,
+				body,
 				newBranch.id,
 				newBranch.upstreamName,
 				draft


gitbutler-ui/src/lib/branches/service.ts
```diff
@@ -68,11 +68,21 @@ export class BranchService {
 
 		const createPrSpan = sentryTxn.startChild({ op: 'pr_api_create' });
 
+		let title = newBranch.name;
+		let body = newBranch.notes;
+
+		// In case of a single commit, use the commit summary and description for the title and
+		// description of the PR
+		if (newBranch?.commits?.length === 1) {
```
eyalch:
```suggestion
		if (newBranch.commits.length === 1) {
```

gitbutler-ui/src/lib/branches/service.ts
```diff
@@ -68,11 +68,21 @@ export class BranchService {
...
+			const paragraphs = newBranch.commits[0].description.split("\n\n");
+			title = paragraphs[0];
+			body = paragraphs.slice(1).join("\n\n");
+		}
```
eyalch, line 80:
```suggestion
		}

```

eyalch (changes requested):

gitbutler-ui/src/lib/branches/service.ts
Pranav2612000: @eyalch Why do you suggest to remove the chaining operator? Especially the `newBranch?.commits` one because `newBranch` has the type `Branch | undefined`

eyalch: There's already a check that throws an error in case `newBranch` is `undefined`:
https://github.com/gitbutlerapp/gitbutler/blob/065e3264e625dc75d9eed977606d2270e4941cc9/gitbutler-ui/src/lib/branches/service.ts#L59-L63

Also, if you glance 2 lines above, you'll see that we already rely on `newBranch` having a value, which means that, if `newBranch` were `undefined` in some case, this optional chaining is already too late and we'd have got an error.

https://github.com/gitbutlerapp/gitbutler/blob/065e3264e625dc75d9eed977606d2270e4941cc9/gitbutler-ui/src/lib/branches/service.ts#L71-L72

Pranav2612000: 0kay. Makes sense. Thank you.

Pranav2612000: Resolving the comment now.

Pranav2612000:
@Caleb-T-Owens I've updated the code to use `descriptionTitle` and `descriptionBody`. I've also incorporated @eyalch 's suggestions.

Pranav2612000:
Also, Github automatically only includes the first 80 characters of the commit title in the PR title and moves the remaining title to the body. We are not supporting this at the moment.

As a further improvement, if the branch has more than 1 commit, we can  include the titles of all the commits in the PR body ( I think Bitbucket does this )

eyalch:
> As a further improvement, if the branch has more than 1 commit, we can include the titles of all the commits in the PR body ( I think Bitbucket does this )

GitHub also does this. I thought of adding this functionality too, so I think it's a good idea. I'm not sure if we should add this as part of the same PR, though.

Pranav2612000:
> I'm not sure if we should add this as part of the same PR, though.
Makes sense. Let's wait for other reviewers

gitbutler-ui/src/lib/branches/service.ts
```diff
@@ -68,11 +68,28 @@ export class BranchService {
...
+		if (newBranch.commits.length === 1) {
+			const commit = newBranch.commits[0];
+
+			if (commit.descriptionTitle) {
+				title = commit.descriptionTitle;
+			}
+
+			if (commit.descriptionBody) {
+				body = commit.descriptionBody;
+			}
```
mtsgrd: Great stuff! Before merging, consider condensing it just a bit, e.g. like this:

```
const commit = newBranch.commits[0];
if (commit.descriptionTitle) title = commit.descriptionTitle;
if (commit.descriptionBody) body = commit.descriptionBody;
```

mtsgrd:
This is really great, I've been wanting this functionality for a long time!

>GitHub also does this. I thought of adding this functionality too, so I think it's a good idea. I'm not sure if we should add this as part of the same PR, though.

Let's merge this in and tackle that in another PR?

mtsgrd (approved):

gitbutler-ui/src/lib/branches/service.ts
eyalch: For convenience, here's a suggestion you can accept with a single click 😄

```suggestion
			const commit = newBranch.commits[0];
			if (commit.descriptionTitle) title = commit.descriptionTitle;
			if (commit.descriptionBody) body = commit.descriptionBody;
```

eyalch (approved):

Pranav2612000:
Thank you @mtsgrd and @eyalch .
I've updated the code to make it succint. Please feel to take another look.

mtsgrd:
Thanks, I'll merge this now. :)

PR #2852, opened 2024-02-21
feat: add search branch by author functionality

Meesayen:
I thought it might be useful to someone else other than me to have a way to filter the branches list by author. I'm very forgetful so I have a hard time remembering what name I used for a specific branch and navigating through a long list other branches can be a nightmare 😅

So, I've come up with this solution where a user can type `author:` followed by any text, in the search field, and the search would trigger a "search by author" flow where the part after `author:` and before a space will be matched against committers' username or email.

I wonder if fetching the remote branches data is too expensive and that's why this component is currently using a very streamlined RemoteBranch structure with minimal properties. If, so, what could be the best way to fetch authors for each branch?

Secondly, I thought neat to enhance the search text field, but it's kinda limiting, as in, users can only search for one author at max. A nice select menu could be more useful in case one wants to filter by many authors. Thoughts?

### DIFF
--- a/gitbutler-ui/src/lib/components/Branches.svelte
+++ b/gitbutler-ui/src/lib/components/Branches.svelte
@@ -9,13 +9,17 @@
 	import { persisted } from '$lib/persisted/persisted';
 	import { storeToObservable } from '$lib/rxjs/store';
 	import { SETTINGS_CONTEXT, type SettingsStore } from '$lib/settings/userSettings';
-	import { BehaviorSubject, combineLatest } from 'rxjs';
+	import { getRemoteBranchData } from '$lib/stores/remoteBranches';
+	import { BehaviorSubject, Subject, combineLatest, map } from 'rxjs';
 	import { getContext, onDestroy, onMount } from 'svelte';
 	import { derived } from 'svelte/store';
 	import type { BranchService } from '$lib/branches/service';
 	import type { CombinedBranch } from '$lib/branches/types';
 	import type { GitHubService } from '$lib/github/service';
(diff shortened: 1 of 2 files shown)


mtsgrd:
@krlvi is getting authors for commits also an expensive operation? I know you recently refactored the way we deal with remote branches.

krlvi:
@Meesayen the remote branch data attributes used to be part of the branches when listing them. Unfortunately, there were some performance issues getting those items for all branches while listing. 

The reason for that is that in the backend it does revwalking from the head of the branch to the merge base, to get the commits and to get the counts. 

We want to get RemoteBranchData only on individual basis. 

But how do we get the author without getting all the commits? 🤔

Meesayen:
Does the underlying Rust git library used by GB have an implementation of `for-each-ref`? 
I don't know how git works under the hood, but something like `git for-each-ref --format='%(authorname) %(authoremail) %09 -%(refname)'` seems pretty fast. (assuming it is not doing the same revwalking you do to analyze commits, under the hood)
Maybe we could use something like that to parse and extract author info data for each branch in a cheap way. But I'm really no expert here. I'm ok to park this PR if the cost outweighs the benefit.

krlvi:
hey @Meesayen! as of yesterday, author is back as an attribute on branches so this should be possible to add now :)

krlvi:
@Meesayen since this pr was quite a bit behind, i ported it here https://github.com/gitbutlerapp/gitbutler/pull/3581
With the next release it will be possible to search by author

Meesayen:
@krlvi  amazing, thanks!

PR #2862, opened 2024-02-21
Fix GitHub error mapping

mtsgrd:
- correctly parses the error response


mtsgrd:
@krlvi fyi - we should get this one into a release asap.

krlvi:
i will cut a release in the morning

PR #2864, opened 2024-02-22
Improve types for github error handling

mtsgrd:
- adds explicit dependency on @octokit/request-error

https://github.com/octokit/octokit.js?tab=readme-ov-file#request-error-handling


mtsgrd:
Hmm, it's unclear if this would work, will park this and revisit at a later date.

PR #2874, opened 2024-02-22
Profile page redesign

PavelLaptev:
- new icons
- icon size could be adjusted with `size` prop
- new focus color for focus states
- `Spacer` component updated
- new `Profile settings` page layout


gitbutler-ui/src/lib/components/AnalyticsSettings.svelte
```diff
@@ -25,14 +22,11 @@
...
-		{#if showTitle}
-			<h2 class="text-base-16 text-bold">Telemetry</h2>
-		{/if}
-		<p class="text-base-body-12 analytics-settings__text">
+		<p class="text-base-body-13 analytics-settings__text">
 			GitButler uses telemetry strictly to help us improve the client. We do not collect any
 			personal information.
 		</p>
-		<p class="text-base-body-12 analytics-settings__text">
+		<p class="text-base-body-113analytics-settings__text">
```
mtsgrd: 👀

gitbutler-ui/src/lib/components/Icon.svelte
```diff
@@ -9,6 +9,9 @@
 	export let color: IconColor = undefined;
 	export let opacity: number | undefined = 1;
 	export let spinnerRadius: number | undefined = 5;
+	export let size = 16;
+
+	const pxToRem = (px: number) => `${px / 16}rem`;
```
mtsgrd, line 14: Could you test this on different zoom levels? I think it'll give you an incorrect result.

This is how I've done it in other places, but it's a bit tedious. Maybe we could put this conversion function into `userSettings.ts`?
```js
const pxToRem = (px: number) => `${px / (16 * $userSettings.zoom)}rem`; 
```

gitbutler-ui/src/lib/components/Login.svelte
```diff
@@ -36,18 +38,35 @@
...
+			.finally(() => {
+				signUpOrLoginLoading = false;
+				onLogin();
+			});
 	};
 	const token = writable<LoginToken | null>(null);
 	const authUrl = derived(token, ($token) => $token?.url as string);
+
+	// create on:login event and on:logout event
+	const dispatch = createEventDispatcher();
```
mtsgrd: For better typing when listening to these events you can do this:

```js
const dispatch = createEventDispatcher<{login: void, logout: void}>();
```

gitbutler-ui/src/lib/components/AnalyticsSettings.svelte
PavelLaptev: oh

gitbutler-ui/src/lib/components/Spacer.svelte
```diff
@@ -1,7 +1,21 @@
 <script lang="ts">
+	export let margin: number | undefined = 16;
+
+	const getPadding = () => {
```
mtsgrd: Named padding but is actually margins. :)

gitbutler-ui/src/lib/components/AnalyticsSettings.svelte
PavelLaptev: thanks for noticing

gitbutler-ui/src/lib/components/Spacer.svelte
PavelLaptev: it should be `getMargins`. Thanks will fix this

PR #2884, opened 2024-02-23
update navigation component to support collapsing and expanding

KroniK907:
I updated the navigation pane on the left side of the app to allow it to be collapsed and expanded. I also refactored the DomainButton component to make it simpler to implement in both places.


gitbutler-ui/src/lib/config/config.ts
```diff
@@ -1,60 +1,65 @@
 import { persisted, type Persisted } from '$lib/persisted/persisted';
 
 export function projectHttpsWarningBannerDismissed(projectId: string): Persisted<boolean> {
-	const key = 'projectHttpsWarningBannerDismissed_';
-	return persisted(false, key + projectId);
+    const key = 'projectHttpsWarningBannerDismissed_';
+    return persisted(false, key + projectId);
```
mtsgrd: Looks like you're changing tabs -> spaces in this file. 👀

gitbutler-ui/src/lib/components/AccountLink.svelte
```diff
@@ -2,25 +2,29 @@
 	import Icon from '$lib/components/Icon.svelte';
 	import type { User } from '$lib/backend/cloud';
 	import { goto } from '$app/navigation';
+	import type { Persisted } from '$lib/persisted/persisted';
 
 	export let user: User | undefined;
 	export let pop = false;
+	export let isNavCollapsed: Persisted<boolean>;
```
mtsgrd: Could we make this a regular boolean? We only need this writable persisted store if we want to also toggle the value (which we don't do in this file).

gitbutler-ui/src/lib/components/Navigation.svelte
```diff
@@ -30,58 +32,93 @@
...
-					<rect opacity="0.7" x="10" y="3" width="4" height="10" rx="2" fill="#D9F3F2" />
-				</svg>
+	const foldNav = () => {
+		console.log('BeforeSet: ' + $isNavCollapsed);
+		$isNavCollapsed = true;
+		console.log('AfterSet: ' + $isNavCollapsed);
+	};
+	const unfoldNav = () => {
+		$isNavCollapsed = false;
+	};
```
mtsgrd: Could we make this one function that's called e.g. toggleNavigation? Also, for sake of consistency, can we do a regular function declaration instead of using `const`.

```
function toggleNavigation() {
  $isNavCollapsed = !$isNavCollapsed;
}
```

gitbutler-ui/src/lib/components/AccountLink.svelte
mtsgrd: The same should apply to the other components where this is passed around.

KroniK907:
I can make the suggested changes. Though most of the parts you have commented on I copied directly from the lane collapse code for the vbranches. So you may want to refractor that code too.

gitbutler-ui/src/lib/components/AccountLink.svelte
KroniK907: I'm not sure how you would like me to deal with the type issue. Should I create a second variable in the Navigation.svelte file that is a normal boolean type that follows the Persisted boolean variable, and then propagate the value of the normal boolean?

KroniK907:
Sorry for the extra commits for formatting. I thought I had got the linter to pass but must have accidently undid the change or something.

KroniK907:
Ok, I really think its passing now. Apologies for the sloppy PR.

gitbutler-ui/src/lib/components/AccountLink.svelte
```diff
@@ -5,22 +5,25 @@
 
 	export let user: User | undefined;
 	export let pop = false;
+	export let isNavCollapsed: boolean;
```
mtsgrd: Could we make this `export let isNavCollapsed = false` so that we can omit the parameter completely in `DecorativeSplitView.svelte`?

gitbutler-ui/src/lib/components/Navigation.svelte
```diff
@@ -30,58 +31,90 @@
...
+			<div class="collapsed-nav__info">
+				<h3 class="collapsed-nav__label text-base-13 text-bold">
+					{project?.title}
+				</h3>
+				<DomainButton
+					href={`/${project.id}/board`}
+					domain="workspace"
+					{branchController}
+					{baseBranchService}
+					{isNavCollapsed}
```
mtsgrd: Here I would I would pass the value using the $ syntax, like this: `isNavCollapsed={$isNavCollapsed}`. Then we can remove the extra `isNavCollapsed` variable, and the other parts similarly simplify.

gitbutler-ui/src/lib/components/Navigation.svelte
```diff
@@ -30,58 +31,90 @@
...
+	>
+		<div class="drag-region" data-tauri-drag-region></div>
+		<div class="hide-nav-button">
+			<Button
+				icon="fold-lane"
+				kind="outlined"
+				color="neutral"
+				help="Collapse Nav"
+				align="flex-end"
+				on:click={toggleNavCollapse}
```
mtsgrd: Since the function called here is so short I would recommend we go with something like this:
```svelte
  on:click={() => $isNavCollapsed = !$isNavCollapsed}
```
*edit: but I see you're doing this in two places, so if you want keep it as a declared function.

gitbutler-ui/src/lib/config/config.ts
```diff
@@ -51,6 +51,11 @@ export function projectLaneCollapsed(projectId: string, laneId: string): Persist
 	return persisted(false, key + projectId + '_' + laneId);
 }
 
+export function navCollapsed(): Persisted<boolean> {
+	const key = 'projectNavCollapsed_';
+	return persisted(false, key);
+}
+
```
mtsgrd: I know this isn't documented anywhere, but I think keeping all the configs in one file is a bit of an anti pattern that I want to move away from. Instead, I would put `const isNavCollapsed = persisted(false, 'projectNavCollapsed_' + projectId)` into `Navigation.svelte`.

Note that I added projectId to the key, but want to confirm it should be a per project setting. At minimum when we start supporting multiple windows it will make sense if it isn't global.

mtsgrd:
Hey sorry for the delay, but great work! I've added some feedback. Would you mind having a look and then I'll try this out locally?

PavelLaptev:
@KroniK907 Thank you for the PR! I like the idea. Before merging, I would like to tweak these design issues.
1. The fold button is small, but it takes too much space in the sidebar. The focus on the button should be secondary or even tertiary.
2. The button looks the same as the lane folding button.
3. Traffic lights overlay the folded sidebar
4. I would like to distinguish folded sidebar from the folded lane design.

image
image

KroniK907:
Makes sense! I was also not sure about where exactly to place the button when the nav was open. I knew you would probably want to fit it into the design better.

Also on Linux I don't have the traffic light, so I didn't realize that would overlap on Mac.

Glad you are willing to consider the functionality though! Thanks!

gitbutler-ui/src/lib/components/AccountLink.svelte
KroniK907: ooh that is a good idea

KroniK907:
@PavelLaptev would you like me to address @mtsgrd's comments or would you like to do that when you make your adjustments?

PavelLaptev:
@KroniK907 let me update the design, and then you can refine the code 🙂. What do you think, does that make sense?

PavelLaptev:
@KroniK907 can you please send a screenshot how the sidebar looks like on Linux? if you don't have the traffic lights, you probably have this empty space here, right?

image

KroniK907:
Yes, that is how it looks. I'm not at my desk but here is a short video that I posted a while back on discord for an unrelated issue that shows what the window looks like for me: https://cdn.discordapp.com/attachments/1073202153163857920/1209972011468591204/msrdc_2024-02-21_12-14-16.mp4?ex=65e8dd3f&is=65d6683f&hm=077e1850c57548a3e8732f1586fa24bf51d874b0a249d925ae520b5a6792d83d&

PavelLaptev:
@KroniK907 thank you, exactly what I need :-)

PavelLaptev:
@KroniK907 I'm finished with my changes 🙂
We can proceed with other fixes

https://github.com/gitbutlerapp/gitbutler/assets/18498712/45067fb2-74a3-4437-906f-6b419617bd13

Changes I made:
1. The folding button now appears when hovering over the sidebar.
2. Updated the `Resizer` UI
3. Added a project icon to the selector so you can identify the project when the navigation is folded.
4. Updated sidebar components, mostly styles

PavelLaptev:
@KroniK907 one thing I noticed is when I switch between projects, it doesn't remember the state and show the sidebar open again. And looks like after that, I need to click twice on the folding button in order to switch `true` to `false` and to `true` again.

https://github.com/gitbutlerapp/gitbutler/assets/18498712/40e68e04-e766-4173-b58f-7d6a16bd2d23

KroniK907:
Love the way that looks!!

PavelLaptev:
@KroniK907 thanks :-) Meanwhile I'll investigate how to add a hotkey for this and add the action to the menu bar.

KroniK907:
@mtsgrd I believe I have addressed all your comments.

PavelLaptev:
@KroniK907 thank you!

gitbutler-app/Cargo.toml
```diff
@@ -65,7 +65,7 @@ similar = { version = "2.4.0", features = ["unicode"] }
 slug = "0.1.5"
 ssh-key = { version = "0.6.4", features = [ "alloc", "ed25519" ] }
 ssh2 = { version = "0.9.4", features = ["vendored-openssl"] }
-tauri = { version = "1.5.4", features = ["dialog-open", "fs-read-file", "path-all", "process-relaunch", "protocol-asset", "shell-open", "window-maximize", "window-start-dragging", "window-unmaximize"] }
+tauri = { version = "1.5.4", features = [ "os-all", "dialog-open", "fs-read-file", "path-all", "process-relaunch", "protocol-asset", "shell-open", "window-maximize", "window-start-dragging", "window-unmaximize"] }
```
mtsgrd, line 68: @krlvi @Qix- what do you make of this? I have also seen modifications like this locally (not done by me) as I'm coding, but for a different feature.

mtsgrd:
Sorry for the delay, but this is great stuff. 🎉  I also love the attention to detail in e.g. how when you resize beyond min-width it switches state to collapsed.

> I can make the suggested changes. Though most of the parts you have commented on I copied directly from the lane collapse code for the vbranches. So you may want to refractor that code too.

Thanks for flagging, I'll have a look. 🙏

gitbutler-ui/src/lib/components/BranchLane.svelte
```diff
@@ -165,7 +165,7 @@
 		align-items: self-start;
 
 		padding: var(--space-12) var(--space-12) var(--space-12) 0;
-		border-right: 1px solid var(--clr-theme-container-outline-light);
+		/* border-right: 1px solid var(--clr-theme-container-outline-light); */
```
mtsgrd: Should this line be deleted?

mtsgrd: I went ahead and deleted it.

gitbutler-ui/src/lib/components/ProjectAvatar.svelte
```diff
@@ -14,32 +14,24 @@
 		'#5FD2B0'
 	];
 
-	const stringToGradient = (string: string | undefined) => {
-		if (!string) {
-			return `linear-gradient(45deg, ${gradients[0][0]} 15%, ${gradients[0][1]} 90%)`;
+	function nameToColor(name: string | undefined) {
+		const trimmed = name?.replace(/\s/g, '');
+		if (!trimmed) {
+			return `linear-gradient(45deg, ${colors[0][0]} 15%, ${colors[0][1]} 90%)`;
```
mtsgrd, line 20: @PavelLaptev I forgot to ask about this, I don't see under what circumstances the parameter to this function can be `undefined`?

Issue #2720, opened 2024-02-15
When a repository is removed from GitButler, the file .git/gitbutler.json should be cleaned up


krlvi:

krlvi:
This is now implemented

PR #2892, opened 2024-02-23
#2720 - Cleanup .git/gitbutler.json file with deletion of project

AbhishekThorat:
- Fix for #2720

### DIFF
--- a/gitbutler-app/src/projects/controller.rs
+++ b/gitbutler-app/src/projects/controller.rs
@@ -201,6 +201,10 @@ impl Controller {
             tracing::error!(project_id = %id, ?error, "failed to remove project data",);
         }
 
+        if let Err(error) = std::fs::remove_file(project.path.join(".git/gitbutler.json")) {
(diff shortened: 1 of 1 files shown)


gitbutler-app/src/projects/controller.rs
```diff
@@ -201,6 +201,10 @@ impl Controller {
             tracing::error!(project_id = %id, ?error, "failed to remove project data",);
         }
 
+        if let Err(error) = std::fs::remove_file(project.path.join(".git/gitbutler.json")) {
+            tracing::error!(project_id = %project.id, ?error, "failed to remove .git/gitbutler.json data",);
+        }
+
```
AbhishekThorat, line 207: I got bit confused with [this function](https://github.com/gitbutlerapp/gitbutler/blob/master/gitbutler-app/src/gb_repository/repository.rs#L542) so not sure whether proposed fix is good enough or not

mtsgrd: @krlvi this seems reasonable to me, what do you think?

krlvi (approved):
thanks

PR #2899, opened 2024-02-24
Add pointer cursor to buttons

Caleb-T-Owens:
# The problem

People have been confused not knowing certain parts of the app are clickable. (especially some of the tags)

# The solution

I've gone and used the pointer cursor to all clickable button-esq elements.
I've tried to go for elements which are "safe bets" in terms of "should it have the cursor pointer?".

In the long term, it would be good to have some answers to things like "Should the commit card have the pointer cursor or the drag cursor?" (at the minute I've left it as is) so we can take a more holistic approach


PavelLaptev:
@Caleb-T-Owens, thanks for the PR!
Several elements in the app still behave like buttons, but have a `default` cursor. If we agree that all clickable elements should have a pointer, I'll update the rest of the elements before merging this PR :-)

PR #2925, opened 2024-02-27
Add keyboard shortcut to commit

### LINKED ISSUE #2896: "cmd + return” to Commit
Might be nice to have more keyboard shortcuts, but specifically one to commit while still focused in the text box would be nice.

alabhyajindal:
Fixes #2896 

Tested on Linux and `Ctrl + Enter` to commit works perfectly. It should work fine on Mac with `Cmd + Return` because of `e.metaKey` - please test this.

I have copy pasted a block of code from the click handler of the 'Commit' button. Another function `checkAndCommit` could be declared and reused in both places. I haven't done this, let me know if you want me to.

### DIFF
--- a/gitbutler-ui/src/lib/components/CommitDialog.svelte
+++ b/gitbutler-ui/src/lib/components/CommitDialog.svelte
@@ -134,6 +134,17 @@
 					on:input={useAutoHeight}
 					on:focus={useAutoHeight}
 					on:change={() => currentCommitMessage.set(commitMessage)}
+					on:keydown={(e) => {
+						if ((e.ctrlKey || e.metaKey) && e.key === 'Enter') {
+							if ($expanded) {
+								if (commitMessage) {
+									commit();
+								}
+							} else {
+								$expanded = true;
+							}
+						}
+					}}
 					spellcheck={false}
 					class="text-input text-base-body-13 commit-box__textarea"
 					rows="1"
--- a/gitbutler-ui/src/lib/components/CommitDialog.svelte
+++ b/gitbutler-ui/src/lib/components/CommitDialog.svelte
@@ -58,6 +58,7 @@
 	};
 
 	function commit() {
+		if (!commitMessage) return;
 		isCommitting = true;
 		branchController
 			.commitBranch(branch.id, commitMessage, $selectedOwnership.toString(), $runCommitHooks)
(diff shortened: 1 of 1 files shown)


gitbutler-ui/src/lib/components/CommitDialog.svelte
```diff
@@ -134,10 +134,21 @@
...
 					on:focus={useAutoHeight}
 					on:change={() => currentCommitMessage.set(commitMessage)}
+					on:keydown={(e) => {
+						if ((e.ctrlKey || e.metaKey) && e.key === 'Enter') {
+							if ($expanded) {
+								if (commitMessage) {
+									commit();
+								}
+							} else {
+								$expanded = true;
```
mtsgrd: This will never happen because of the `{#if expanded}` on line 127.

gitbutler-ui/src/lib/components/CommitDialog.svelte
```diff
@@ -134,10 +134,21 @@
 					on:input={useAutoHeight}
 					on:focus={useAutoHeight}
 					on:change={() => currentCommitMessage.set(commitMessage)}
+					on:keydown={(e) => {
+						if ((e.ctrlKey || e.metaKey) && e.key === 'Enter') {
+							if ($expanded) {
+								if (commitMessage) {
```
mtsgrd: Let's move this check into the commit function? That way we won't need an intermediary function to make things dry.

mtsgrd:
Thanks, this is great. Lgtm if you just check the comments I left.

gitbutler-ui/src/lib/components/CommitDialog.svelte
alabhyajindal: Removed 👍🏼

gitbutler-ui/src/lib/components/CommitDialog.svelte
alabhyajindal: The `commit` function is called on button click, line 209. 

Do you want me to modify the `commit` function to handle both the click and keyboard event. I could do this by passing a parameter but it won't look pretty.

mtsgrd: What you have now looks good to me? I'm not sure what parameter you would want to pass.

alabhyajindal: Great! I thought you wanted me to move the code which checks for the key pressed inside the `commit` function. My bad.

mtsgrd:
Tested this out locally, works great! Feel free to merge.

mtsgrd (approved):

schacon:
Thanks @alabhyajindal !!

Issue #2694, opened 2024-02-15
Support keyboard navigation when browsing changes


steffandroid:
I commonly use the up and down arrows on my keyboard to iterate through the list of changed files in other Git GUI apps (e.g. GitHub Desktop), so would really like this to be added in GitButler.

Currently, the arrow keys scroll the whole workspace, but I don't find this to be particularly useful.

This is with macOS, not sure if it behaves differently on other platforms.

steffandroid:
The arrows keys now iterate through the files, but in a seemingly arbitrary order rather than the displayed order.

PavelLaptev:
@steffandroid thank you for the report!
Do you still encounter the issue? If yes, could you please write the version you're currently on and which OS are you using.
It would be ideal if you can also capture a video of this bug. Thanks 🙂

steffandroid:
@PavelLaptev pleased to see this is no longer an issue – closing.

Issue #2663, opened 2024-02-14
Support arrow key navigation between files in a virtual branch lane 


krlvi:
discord discussion https://discord.com/channels/1060193121130000425/1176875943873232947/1207235171636609036

PR #2928, opened 2024-02-27
Added keyboard shortcuts for file navigation

alabhyajindal:
Possible fix for #2694 and #2663. 

https://github.com/gitbutlerapp/gitbutler/assets/52493077/d1c6c85d-3665-463d-a4d4-b84a9eca26fe

The focus remains on the initially selected item, I don't know how to move it. Please help out if you do!


alabhyajindal:
Thanks @mtsgrd! The PR is not done yet. I'll let you know once it's ready for review.

alabhyajindal:
This is ready for review. The keyboard navigation now works fine for the file list view with correct focus states. 

Unfortunately I'm unable to get it working for the  tree view. I think we need to pass down `files` to `FileTree.svelte` and add the keydown handler there. I'm unable to do this.

I tried passing `files` to `FileTree.svelte` from `BranchFiles.svelte` but I get an error: `<FileTree> was created without expected prop 'files'`.

gitbutler-ui/src/lib/components/BranchFilesList.svelte
```diff
@@ -39,6 +42,27 @@
...
+			if (e.key === 'ArrowUp') {
+				if (selectedFileIndex - 1 < 0) return;
+				$selectedFiles = [sortedFiles[selectedFileIndex - 1]];
+			} else if (e.key === 'ArrowDown') {
+				if (selectedFileIndex + 1 > sortedFiles.length - 1) return;
+				$selectedFiles = [sortedFiles[selectedFileIndex + 1]];
+			}
+
+			// Focus on the newly selected file
+			const fileElement = document.getElementById('file-' + $selectedFiles[0].id);
```
mtsgrd: While this technically works, I'm a bit reluctant to perform action at a distance. If we were to change the id of that element, it would be difficult to remember/predict that this line also needs to change.

I'm checking this out locally now, give me an hour or two and I'll get back to you.

mtsgrd:
> I tried passing files to FileTree.svelte from BranchFiles.svelte but I get an error: <FileTree> was created without expected prop 'files'.

It sounds like you need to pass files recursively, so add it here as well?
https://github.com/gitbutlerapp/gitbutler/blob/5f928c6d4455058e9940310bed911bccef8c2583/gitbutler-ui/src/lib/components/FileTree.svelte#L136-L149

alabhyajindal:
> It sounds like you need to pass files recursively, so add it here as well?

Thanks! I've added that. Sorry but I'll be unable to work on adding the navigation to the Tree View as well. Maybe that could be addressed in a different PR.

gitbutler-ui/src/lib/components/BranchFilesList.svelte
alabhyajindal: I have slightly improved the function but it still references the element id which is fragile as you said. We need to get the child DOM to the parent and then call `focus()` on it. 

I'm unsuccessful here - it'll be better if you give this a try sometime later.

- [Getting a Svelte component reference from child](https://www.sveltesociety.dev/recipes/component-recipes/getting-references-to-components-in-each-blocks) returns a Svelte component. DOM methods can't be called there.
- There's a way to get the [DOM to parent](https://stackoverflow.com/a/62282925/11484041) as mentioned in the answer here. This needs to be modified to work for all elements in the each block.

mtsgrd:
@alabhyajindal thanks again for working on this. I've added a commit that enables this for the file tree as well, and uses a different method for updating the focus. What do you think?

mtsgrd (approved):

krlvi:
fixing a merge conflict

PR #2934, opened 2024-02-28
Refactor diff/signature.rs

MichaelOwenDyer:
Signature.rs had a few FIXMEs asking for the identifiers to be cleaned up, so I thought I would give it a shot. Tried to rename variables as intuitively as I could, and it's certainly better than it was before. Perhaps someone with a better understanding of the algorithm at play would have different opinions; feel free to comment.
Also added some documentation to the `bigrams` function and consolidated a duplicate macro in the test module.

### DIFF
--- a/gitbutler-diff/src/signature.rs
+++ b/gitbutler-diff/src/signature.rs
@@ -122,7 +122,10 @@ impl<S: AsRef<str>> From<S> for Signature {
     #[inline]
     fn from(source: S) -> Self {
         let source = source.as_ref();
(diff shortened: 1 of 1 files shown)


Qix-:
:D thank you so much!

MichaelOwenDyer:
You got it!

MichaelOwenDyer:
Added a couple more last-minute tweaks. GitButler made me force push, hope that's okay

Qix- (approved):

Qix-:
Force push on PR branches is fine :) That is so much, looks much better now! Appreciated!

PR #2936, opened 2024-02-28
Making the image in empty board not draggable.

pouyakary:
I made this image not selectable/draggable as it was a UI-component and acted like content:

Screenshot 1402-12-09 at 1 48 55 PM


Qix-:
Thanks! :)

PavelLaptev:
@pouyakary thank you for your PR 🙏 
I moved these CSS to the image component instead of just the board, so it will work across all images

PR #2939, opened 2024-02-28
Fix app updater bug

mtsgrd:
- move `distinctUntilChanged` before combining with status
- add appropriate async to install function
- small reformatting

### DIFF
--- a/gitbutler-ui/src/lib/components/AppUpdater.svelte
+++ b/gitbutler-ui/src/lib/components/AppUpdater.svelte
@@ -2,18 +2,13 @@
 	import Button from './Button.svelte';
 	import IconButton from './IconButton.svelte';
 	import { showToast } from '$lib/notifications/toasts';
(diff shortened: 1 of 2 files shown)


gitbutler-ui/src/lib/components/AppUpdater.svelte
```diff
@@ -2,18 +2,13 @@
...
+	import { tap } from 'rxjs';
 	import { fade } from 'svelte/transition';
 	import type { UpdaterService } from '$lib/backend/updater';
 
 	export let updaterService: UpdaterService;
 
-	// Extrend update stream to allow dismissing updater by version
 	$: update$ = updaterService.update$.pipe(
-		// Only run operators after this one once per version
-		distinctUntilChanged((prev, curr) => prev?.version == curr?.version),
```
mtsgrd, line 16: This line broke the "install button", moved it into `updater.ts`.

PR #2957, opened 2024-03-01
Add horizontal scrolling functionality

KroniK907:
Add horizontal mousewheel scrolling functionality to the scroll-viewport div when the mouse is above a certain threshold to allow for easier horizontal scrolling within the contents. Threshold is currently manually set to 145px which is the bottom of the branch header. Unsure how to set this programmatically in a nice way.

### DIFF
--- a/gitbutler-ui/src/routes/[projectId]/board/+page.svelte
+++ b/gitbutler-ui/src/routes/[projectId]/board/+page.svelte
@@ -62,6 +62,7 @@
 			}}
 			on:wheel={(e) => {
 				if (!horizontalScrolling) return;
+				if (Math.abs(e.deltaX)) return;
 				e.preventDefault();
 				viewport.scrollBy({
 					left: e.deltaY,
--- a/gitbutler-ui/src/routes/[projectId]/board/+page.svelte
+++ b/gitbutler-ui/src/routes/[projectId]/board/+page.svelte
@@ -25,6 +25,9 @@
 	let viewport: HTMLDivElement;
 	let contents: HTMLDivElement;
 
+	let horizontalScrolling: boolean = false;
+	const horizontalScrollAbove = 145;
+
 	const httpsWarningBannerDismissed = projectHttpsWarningBannerDismissed(projectId);
 
 	function shouldShowHttpsWarning() {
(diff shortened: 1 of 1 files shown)


KroniK907:
@mtsgrd & @PavelLaptev Thoughts?

KroniK907:
One thing I didn't really think about is horizontal scrolling using trackpads. Maybe the `on:wheel` should check for `deltaX` and use that if its not zero.

PavelLaptev:
@KroniK907 could be a good UX case. 
Do you use a regular mouse, right?
Asking because I mostly use a trackpad or magic mouse, which supports both scrolling directions. But I have a regular mouse too 🐁. Let me test how it feels, and I'll write you back.

KroniK907:
@PavelLaptev Yeah, I rarely use a laptop with a trackpad, and all my mice only offer vertical scrolling. I just added a commit that should help make the trackpad and magic mouse experience much more similar to how it worked before, and only trigger the horizontal scroll with a vertical mouse wheel movement if no horizontal scroll delta is detected.

PavelLaptev:
@KroniK907, I like the idea.

When using a mouse, it's easy and convenient. But when using a trackpad and scrolling over the header, it can be confusing. I prefer to keep the scrolling natural.

I found that holding the "**shift**" key allows horizontal scrolling with the mouse. Can you check if it works for you?
If not, I'm thinking of mimicking this behavior instead of creating different scroll zones.

Also, what do you think about adding a setting to show the horizontal scrollbar by default?

KroniK907:
@PavelLaptev unfortunately, using shift does not work on my setup. When using GitButler in WSL with a standard mouse, there is no way for me to horizontal scroll at all besides using the arrow keys on the keyboard.

Could we maybe have a setting that enables this mode that is off by default? Or maybe it's an invisible setting in local storage that disables this mode if a deltaX value is ever seen?

PavelLaptev:
I see. It seems this combination doesn't function on all browsers.

I've implemented a workaround where using the shift key + scroll should enable horizontal scrolling. Could you please test it on [this branch](https://github.com/gitbutlerapp/gitbutler/tree/Horizontal-scroll) to see if it works for you?

The settings idea is good, I'm considering something like the option to show/hide the board scrollbar. I'll create an issue for it.

FaithfulDev:
FYI: I tested it with GitButler running natively on Windows. Works as intended.

PavelLaptev:
@FaithfulDev cool, thanks for testing! Let's see how it works on Linux 👀

KroniK907:
@PavelLaptev This works great except for one tiny detail! The `on:keyup` event should use a ternary to set the value to false because `e.shiftKey` will return `true` on the keyup not `false`.

PavelLaptev:
@KroniK907 thanks! Would you mind implementing this fix into this PR for me?
I'll remove my branch after 🙂

KroniK907:
After building the dev of the current master, the shift+scroll is working on my setup now. Not sure why it did not work before.

PavelLaptev:
Interesting, but it doesn't work for you on prod right?

PavelLaptev:
Thanks, that's good news 🙂 Could you then close this PR. And I'll still work on show/hide scrollbar feature [here](https://github.com/gitbutlerapp/gitbutler/issues/2985)

KroniK907:
Ended up not being necessary

PavelLaptev:
Related PR https://github.com/gitbutlerapp/gitbutler/pull/3083

PR #2959, opened 2024-03-01
File context menu improvements

### LINKED ISSUE #2950: Add the ability to "copy path to file" and "open in VSCode" on file right click

Pranav2612000:
Fixes #2950

### DIFF
--- a/gitbutler-ui/src/lib/components/FileContextMenu.svelte
+++ b/gitbutler-ui/src/lib/components/FileContextMenu.svelte
@@ -29,7 +29,7 @@
 					<ContextMenuItem label="Discard changes (Binary files not yet supported)" disabled />
 				{:else}
 					<ContextMenuItem
-						label="Discard changes"
+						label={item.files.length === 1 ? "Discard file" : "Discard files"}
 						on:click={() => {
 							confirmationModal.show(item);
 							dismiss();
--- a/gitbutler-ui/src/lib/components/FileContextMenu.svelte
+++ b/gitbutler-ui/src/lib/components/FileContextMenu.svelte
@@ -5,9 +5,14 @@
 	import ContextMenu from '$lib/components/contextmenu/ContextMenu.svelte';
 	import ContextMenuItem from '$lib/components/contextmenu/ContextMenuItem.svelte';
 	import ContextMenuSection from '$lib/components/contextmenu/ContextMenuSection.svelte';
+	import * as toasts from '$lib/utils/toasts';
+	import { join } from '@tauri-apps/api/path';
+	import { open } from '@tauri-apps/api/shell';
+	import type { Project } from '$lib/backend/projects';
 	import type { BranchController } from '$lib/vbranches/branchController';
 	import type { AnyFile } from '$lib/vbranches/types';
 
+	export let project: Project | undefined;
 	export let branchController: BranchController;
 	let confirmationModal: Modal;
 	let popupMenu: PopupMenu;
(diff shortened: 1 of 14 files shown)


Pranav2612000:
For `Open in VSCode` & `Copy Absolute Path` we'll need the base project directory location. I couldn't find a way to get this directly.

One solution is to pass the project ( or `projectService` ) as props to the `FileContextMenu` component, but that would involve a fair bit of prop-drilling.
The other option is to add an `absolutePath` property to the `LocalFile` type we have.

@krlvi What do you think should we do? Or is there a different better approach?

mtsgrd:
You're right about prop drilling, but that is how we have been doing things to date. What I'd prefer is some form of automated dependency injection the way Angular does it, but Svelte's `setContext` and `getContext` methods provide no lifecycle management. Currently what this means is when you e.g. switch between projects components that are recycled rather than destroyed and recreated would not be given the latest dependencies created in the `layout.ts` files.

So we do need a solution here, but until we have a plan it feels prop drilling is better than e.g. going with global singletons. What do you think?

Pranav2612000:
Makes sense @mtsgrd
I was also thinking we need a global state management tool like React Context.
I'm not very well versed but with Svelte, but I get the general idea of what you are trying to say. 
I think the stale context would be a common issue. Do you have an idea of how others do it? I'll be happy to spend more time and see if there are any workarounds we can come up with.

In the meantime, I'll update this PR to pass the props to the `FileContextMenu` component.
Thank you so much for your help.

Pranav2612000:
Updated the PR to use prop drilling. Feel free to review

gitbutler-ui/src/lib/components/FileContextMenu.svelte
```diff
@@ -29,13 +33,57 @@
...
+							}
+						}}
+					/>
+					<ContextMenuItem
+						label="Copy Relative Path"
+						on:click={() => {
+							try {
+								if (!project) {
+									return;
+								}
```
mtsgrd: For consistency with the code base I'd recommend expressing this as: `if(!project) return;`

gitbutler-ui/src/lib/components/FileContextMenu.svelte
```diff
@@ -29,13 +33,57 @@
...
+								navigator.clipboard.writeText(item.files[0].path);
+								dismiss();
+							} catch (err) {
+								console.log('Failed to copy relative path', err);
```
mtsgrd, line 67: I think we want to show a toast here, could you add something like this?

```
import * as toasts from '$lib/utils/toasts';
toasts.error('Failed to copy relative path', err);
```

Pranav2612000:
@mtsgrd I've updated the code to add error toasts. Can you please take another look

gitbutler-ui/src/lib/components/FileContextMenu.svelte
```diff
@@ -29,13 +34,58 @@
...
+					on:click={async () => {
+						try {
+							if (!project) return;
+							for (let file of item.files) {
+								const absPath = await join(project.path, file.path);
+								open(`vscode://file${absPath}`);
+							}
+							dismiss();
+						} catch {
+							console.log('Failed to open in VSCode');
```
mtsgrd, line 84: Can we change this to  catch the error, switch log to `console.error()` and pass the err as a second argument?

gitbutler-ui/src/lib/components/FileContextMenu.svelte
```diff
@@ -29,13 +34,58 @@
...
+					/>
+					<ContextMenuItem
+						label="Copy Relative Path"
+						on:click={() => {
+							try {
+								if (!project) return;
+								navigator.clipboard.writeText(item.files[0].path);
+								dismiss();
+							} catch (err) {
+								console.log('Failed to copy relative path', err);
```
mtsgrd, line 67: Same here, let's use `console.error()`?

mtsgrd (approved):

mtsgrd:
Added a couple of comments, but have approved so feel free to merge when ready!

krlvi:
Thanks for the contribution! I am gonna merge this and change the logs to `console.error()` in a separate PR

PR #2968, opened 2024-03-01
move from lib+bin to bin, refactor tests into crate

Qix-:
Title. This is a relatively large, sweeping change that simplifies the app codebase a bit.

Since there's no `[[bin]]` and tests have been refactored into in-crate tests, all modules can now be marked `pub(crate)` at the top level, which allows for much more static analysis to be performed.


Qix-:
Outage with Github Actions / PRs right now. But should be ready.

PR #2982, opened 2024-03-01
Create new pull request card

mtsgrd:
Moves pull requested tags and merge button into a separate card.

image

### DIFF
--- a/gitbutler-ui/src/lib/components/ActiveBranchStatus.svelte
+++ b/gitbutler-ui/src/lib/components/ActiveBranchStatus.svelte
@@ -1,34 +1,14 @@
 <script lang="ts">
 	import Tag from '$lib/components/Tag.svelte';
-	import ViewPrContextMenu from '$lib/components/ViewPrContextMenu.svelte';
 	import { normalizeBranchName } from '$lib/utils/branch';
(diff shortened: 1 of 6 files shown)


mtsgrd:
@PavelLaptev just fyi, we don't have contrast colors for `Tag.svelte` yet.

PavelLaptev:
@mtsgrd In Figma, it's labeled as `contrast`, while in the code, it's referred to as `filled`. I'll update the name in Figma to match our code.

mtsgrd:
I'm an idiot :facepalm:

But before we leave this thread, I want to mention we need to think about the way we collapse drop down left/right [border in this file](https://github.com/gitbutlerapp/gitbutler/blob/cd879ab014112e57efd249636bce40dd43fa1ec2/gitbutler-ui/src/lib/components/DropDownButton.svelte#L101). It prevents left border of a merge button from aligning exactly with any previous containers' left border in a column layout. Can you think of a good way to solve it?

With 1px translation
image

Without
image

PavelLaptev:
No worries, thank you :-) I'll check this out and fix with the next PR

mtsgrd:
Awesome, thanks! Fyi, switching to filled tag here: #2987

Issue #2983, opened 2024-03-01
'Add key to the GitHub' action from Project settings screen is not opening github in browser.


AbhishekThorat:
- Seems like we are missing import to `open` api from `tauri`

Quick screenshot for reference -
![open-browser-tab-issue](https://github.com/gitbutlerapp/gitbutler/assets/18628649/5515c54c-dc07-4a81-89e9-c304db7ff119)

PR #2984, opened 2024-03-01
'Add key to the GitHub' action from Project settings screen is not opening github in browser.

AbhishekThorat:
- Add missing import for `open` tauri api
Fix for #2983

### DIFF
--- a/gitbutler-ui/src/lib/components/KeysForm.svelte
+++ b/gitbutler-ui/src/lib/components/KeysForm.svelte
@@ -9,7 +9,7 @@
 	import Link from '$lib/components/Link.svelte';
 	import { copyToClipboard } from '$lib/utils/clipboard';
 	import { debounce } from '$lib/utils/debounce';
-	import { open } from '@tauri-apps/api/shell';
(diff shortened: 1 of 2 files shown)


gitbutler-ui/src/lib/components/KeysForm.svelte
```diff
@@ -9,6 +9,7 @@
 	import Link from '$lib/components/Link.svelte';
 	import { copyToClipboard } from '$lib/utils/clipboard';
 	import { debounce } from '$lib/utils/debounce';
+	import { open } from '@tauri-apps/api/shell';
```
mtsgrd: Can we use `openExternalUrl` here instead? This has unfortunately happened before since `open` is a global type.

AbhishekThorat: this make sense @mtsgrd. I should have checked the definition for `openExternalUrl` earlier, my bad.
Addressed your review comment with [this commit](https://github.com/gitbutlerapp/gitbutler/pull/2984/commits/01cff8eaac5e48bd1aacb56765387cf23b4b0db8).

PR #2986, opened 2024-03-02
Fix merge button 1px left margin effect

mtsgrd:


PavelLaptev:
I think I found a solution to handle the dropdown style. Will ship it with my next PR. Let's close this one 🙂

Issue #3017, opened 2024-03-04
branch name generator stuck in loop for changes that heavily use the string 'virtual branch'


KroniK907:
If a local change has the string 'virtual branch' in multiple places such that it would reasonably be a string included in an AI generated branch name, then the branch name generation will end up in a loop of generating new branch names.

PR #3019, opened 2024-03-04
Improve new branch name detection

KroniK907:
Improved the new branch name detection for auto generating branch names. Now uses a regex to match new branch names more explicitly and will avoid branch name generation loops when 'virtual branch' is a string used often in the diff. Fixes issue #3017


mtsgrd:
Thank you! :)

PR #3028, opened 2024-03-04
style: update background color for FileListItem. To make it visible while dragging

PavelLaptev:


mtsgrd:
Sorry, I should have mentioned I did this already.

PR #3038, opened 2024-03-05
move most `TryFrom`  initialization into main instead of initialization on demand

Qix-:


gitbutler-app/src/main.rs
```diff
@@ -98,48 +98,74 @@ fn main() {
...
-                        virtual_branches::controller::Controller::try_from(&app_handle)
-                            .expect("failed to initialize virtual branches controller");
-                    app_handle.manage(vbranch_contoller);
+                    let deltas_database_controller = deltas::database::Database::new(database_controller.clone());
+                    app_handle.manage(deltas_database_controller.clone());
 
-                    let keys_controller = keys::Controller::try_from(&app_handle).expect("failed to initialize keys controller");
-                    app_handle.manage(keys_controller);
+                    let deltas_controller = deltas::Controller::new(deltas_database_controller.clone());
+                    app_handle.manage(deltas_controller);
```
mtsgrd, line 147: Variable is only used here, could drop the declaration like you've done in some places?

Qix-: Yeah I'll clean it up tomorrow. :) good catch

PR #3041, opened 2024-03-05
Move over to evaluatePrompt endpoint for AI integration

Caleb-T-Owens:
This PR stops using the old summarization endpoints in favour of the new evaluatePrompt endpoint. This enables us to generate the commit messages on the client side.

## AI Providers

We've already talked about having the ability to plug in different AI sources, and potentially have local models, so I decided to leave an open door there and create an AIProvider interface that can be implemented.

## Change from ES2020 to ES2021 target

ES2021 provides the `String#replaceAll` method which is handy so I've updated the target to 2021.

My UI seemed not to update after some commit operations but seemed to be fine after I properly restarted GB. However, it would be helpful if someone could try testing doing some gitty operations and see if the UI responds as expected.

## Summarizer

I'm not entirely happy with all the formatting logic in there, but at the minute I don't see a better way to achieve it without overcomplicating it for the sake of it, so any feedback would be appreciated 👍

### DIFF
--- a/gitbutler-ui/src/lib/backend/ai_providers.ts
+++ b/gitbutler-ui/src/lib/backend/ai_providers.ts
@@ -1,0 +1,30 @@
+import type { User, getCloudApiClient } from '$lib/backend/cloud';
+
+enum ButlerMessageRole {
+	User = 'user',
+	System = 'system'
+}
+
+export interface ButlerPromptMessage {
+	content: string;
+	role: ButlerMessageRole;
+}
+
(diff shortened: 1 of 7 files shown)


krlvi:
Behaves correctly and this sets the ground for doing the prompt tweaking client side, by users, which is awesome! As well as customizing the end point.

@mtsgrd do you see any issues with bumping the ES target version? (i hope this doesnt mess with the linux builds in some way)

PR #3044, opened 2024-03-06
make go back to integration more predictable

ngalaiko:


krlvi:
@ngalaiko what happens if there is uncommitted work on the non-integration branch? it would be reasonable to reject the operation if that is the case, and have the UI inform the user that they need to stash or commit their stuff

ngalaiko:
> uld be reasonable to reject the operation if that is the case, and have the UI inform the user that they need to stash or commit t

if it conflicts with the integration tree, then there will be an error. otherwise, it'll be imported into a new vbranch. This is how git behaves, see conversation with @mtsgrd  in linear.

krlvi:
perhaps the adding to a new vbranch is unexpected in this case. Scott put some thoughts in here https://github.com/gitbutlerapp/gitbutler/issues/3048

ngalaiko:
@krlvi take a look again please

krlvi:
did some testing - it seems like after going back to gb integration, the branches that used to be applied are not applied

gitbutler-app/src/virtual_branches/integration.rs
```diff
@@ -319,14 +318,8 @@ fn verify_head_is_set(
         Some(refname) if refname.to_string() == GITBUTLER_INTEGRATION_REFERENCE.to_string() => {
             Ok(())
         }
-        None => {
-            super::mark_all_unapplied(gb_repository).map_err(errors::VerifyError::Other)?;
-            Err(errors::VerifyError::DetachedHead)
-        }
-        Some(head_name) => {
-            super::mark_all_unapplied(gb_repository).map_err(errors::VerifyError::Other)?;
```
ngalaiko, line 327: @schacon maybe you remember why we had this? anyways now it seems not needed

ngalaiko: @krlvi anyways, this is why it was unapplying. i think we've added it to avoid conflicts when switching to and from git, but since now we handle that by other means, should be safe to remove.

schacon: I don't remember this. It looks like you introduced it here:  490318e8ea9415a66b2f3d0e17d48f41be7c96a8

schacon: https://github.com/gitbutlerapp/gitbutler/pull/991

PR #3050, opened 2024-03-06
feat: Update button text from "Authenticate" to "Authorize" in GitHub authentication workflow

PavelLaptev:
Changed the button text from "Authenticate" to "Authorize" for better clarity.


gitbutler-ui/src/lib/components/GithubIntegration.svelte
```diff
@@ -102,15 +100,15 @@
 		</svelte:fragment>
 		<Button {disabled} kind="filled" color="primary" on:click={gitHubStartOauth}>
 			{#if $user$?.github_access_token}
-				Reauthenticate
+				Reauthorize
```
mtsgrd: As discussed, we might want to allow the user to first remove any existing integration before offering to set it up again.

PavelLaptev: Fixed that with the commit below

PR #3055, opened 2024-03-06
refactor remaining low-hanging-fruit `TryFrom` implementations

Qix-:


gitbutler-app/src/keys/controller.rs
```diff
@@ -14,7 +12,8 @@ impl Controller {
         Self { storage }
     }
 
-    pub fn from_path<P: AsRef<Path>>(path: P) -> Self {
+    #[cfg(test)]
+    pub fn from_path<P: AsRef<std::path::Path>>(path: P) -> Self {
```
mtsgrd, line 16: I'm curious how you determine when to `use std::path::Path` or when to spell it out inline?

Qix-: In this case it's because this method is only used in tests, hence `#[cfg(test)]`. If I did a `use std::path::Path` at the top, this would be the only usage of `Path`, and would cause a warning in normal (non-test) builds for being an unused `use`. Something I wish Rust handled better but alas, it doesn't.

In theory I could do a

```rust
#[cfg(test)]
use std::path::Path
```

but that's just adding more lines for no real gain. Thus, I just spell out the full canonical type to eliminate those two lines altogether.

Qix-: Also, it's not a great pattern to have `#[cfg(test)]` on otherwise non-test `impl`s of types. Kind of crufts up the docs/code/etc. You'd normally just do a helper function somewhere in the test code itself, to contain the alternative construction case.

However, it's not easy to do right now since the types are so heavily inter-dependent and thus test helpers would have to be able to access others modules' test helpers, etc. which also isn't ideal. This was the lowest-friction path to getting rid of `TryFrom` in these cases, for now.

mtsgrd: Awesome, makes sense!

PR #3068, opened 2024-03-07
Use SectionCards instead of ClickableCard

mtsgrd:
- commit grew a bit in scope
- lots of refactoring to settings pages
- clickable card dropped
- should not change any behavior


PavelLaptev:
@mtsgrd cards are not clickable anymore. Can you please fix this?
Also, cards not always clickable, sometimes they just show the content, so they shoudn't have `pointer` cursor in this case.

https://github.com/gitbutlerapp/gitbutler/assets/18498712/f8ea928f-a193-43ac-9111-64dcff67f4e4

PavelLaptev:
the layout for these sections now broken

image

also I don't see the generated SSH key
image

PavelLaptev:
the layout for this section also broken

image

PavelLaptev:
@mtsgrd could you ping me next time before merge? There are now many broken styles we need to fix

mtsgrd:
Sure thing, but let me take a look before you fix anything.

PavelLaptev:
Sounds good. I leave it to you just ping me when you finish and I'll check it then 🙂

PavelLaptev:
This is how I did it before for these cards. I pass the additional content as a `slot`.
We also disscuss the naming confution today. I had `body` slot which was besically a `description`

image

mtsgrd:
I think I fixed it all, thanks for being so vigilant 🙏

PavelLaptev:
@mtsgrd many thanks! I'll check it in the Morning!
And sorry I take my words back. I render clickable and not clickable as different card separatly, sorry. 
I did it because the top part was clickable and the content not

https://github.com/gitbutlerapp/gitbutler/assets/18498712/4e988575-9c18-47bb-a108-6d00f6217c25

image

PR #3073, opened 2024-03-07
fix: branches filter checkboxes and toggle behavior

LeoCaprile:
Hey guys!
Amazing app, just wanted to fix this weird bug, when clicking on the checkbox of the branches filters, the problem occurs because the on:mousedown don't work for some reason, so change it to a on:click, solve the problem.

Here's a little video showing the bug then the fix:
!

https://github.com/gitbutlerapp/gitbutler/assets/71847541/cdd5294c-34ce-4cd0-b228-b54818c9e4b6


mtsgrd:
Thanks so much. I did a bunch of `on:click` -> `on:mousedown` changes the other day, but missed this regression. Could you run `pnpm format` and push again?

LeoCaprile:
Done!

PR #3075, opened 2024-03-08
Fix settings page refactor regressions

mtsgrd:
- fixes click on some missing elements
- load ssh key in onMount
- happened with removal of clickable card
- fixes missed prop renames


PavelLaptev:
I mean we could keep `label` if there is an action inside the card. But for not clickable case I think it should be different. Let me check what we can do

Issue #2612, opened 2024-02-13
Duplicated files visible on drag and drop


BloodStainedCrow:
I seem to have a ghost version of a couple files attached to my drag and drop:

![image](https://github.com/gitbutlerapp/gitbutler/assets/23032812/721c8a4e-e012-49d0-b82e-dd6bd0e90039)

![image](https://github.com/gitbutlerapp/gitbutler/assets/23032812/a9db9bd6-9ac7-4a7d-b4bd-b095dd33fa99)

![image](https://github.com/gitbutlerapp/gitbutler/assets/23032812/3f8fb527-ba7a-4080-9831-0f58888534aa)

BloodStainedCrow:
This happens when I drag around any file, you cannot see my cursor in the screenshots, but I am holding the files

BloodStainedCrow:
I also cannot move these files, now that they are quantum entangled.

mtsgrd:
I just looked into this and discovered a bug. Any chance you have these other files selected in a different branch? I can probably submit a fix later today.

mtsgrd:
@BloodStainedCrow are you still seeing this bug?

LeoCaprile:
I think this may happen when you drag by the item icon, don't know wich OS @BloodStainedCrow is using but in macOS looks like this:


https://github.com/gitbutlerapp/gitbutler/assets/71847541/f4dc0475-093f-463c-9a27-40c3fd5cbec4

Maybe this is related to another bug, but that some weird behavior.

LeoCaprile:
Okay, got it.
After doing the above, I get the same behaviour as @BloodStainedCrow

<img width="1728" alt="Captura de pantalla 2024-03-07 a la(s) 21 55 33" src="https://github.com/gitbutlerapp/gitbutler/assets/71847541/eb5f70f4-d76d-40bb-ab7c-4073d04b9d80">

gonna make a PR fixing this.

LeoCaprile:
PR on #3076 solve this bug

PavelLaptev:
Looks like this has been fixed. Let's reopen the issue if you still encounter this issue.

PR #3076, opened 2024-03-08
Fixed duplicated files on drag

LeoCaprile:
The bug occurs when trying to drag the file by clicking on the icon of the file, then the function `handleDragStart` of the `use:draggable`action, starts the draggin, but the function `handleDragEnd` it's never called, because the dropzone wouldn't admit the image(file icon). This causes that the clean up function to never be called and leaves orphan cloned nodes arround that will be selected by the query selector ".selected-draggable" on the options of the `use:draggable`

So the easy solution was making the image not draggable and that's it.

This video is replicating the bug.

https://github.com/gitbutlerapp/gitbutler/assets/71847541/b7a04603-6e6d-4b4b-a4de-8f978bda15b4

This video was after adding `draggable="false"` attribute to the file-icon.

https://github.com/gitbutlerapp/gitbutler/assets/71847541/7d15a973-9620-4e3e-a872-531c176fc0cc

Solves: #2612

### DIFF
--- a/gitbutler-ui/src/lib/components/FileListItem.svelte
+++ b/gitbutler-ui/src/lib/components/FileListItem.svelte
@@ -98,7 +98,7 @@
 	>
 		<div class="info-wrap">
 			<div class="info">
-				<img src={getVSIFileIcon(file.path)} alt="js" style="width: var(--space-12)" />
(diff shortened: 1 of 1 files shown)


mtsgrd:
Thanks! 🙏

PR #3084, opened 2024-03-09
AI Provider Settings

Caleb-T-Owens:
This PR introduces the ability to better configure the AI within GitButler, allowing for bring your own key configurations and choosing between different strength models.

App preferences whilst logged in
image

App preferences whilst logged out
image

App preferences choosing a bring your own key option
image

Updated project settings will always show the AI settings
image


gitbutler-ui/src/lib/components/BranchCard.svelte
```diff
@@ -99,16 +101,12 @@
...
-			const aiProvider = new ButlerAIProvider(cloud, user);
-
-			new Summarizer(aiProvider).branch(diff).then((message) => {
-				if (message !== branch.name) {
-					branch.name = message;
-					branchController.updateBranchName(branch.id, branch.name);
-				}
-			});
-		}
+		summarizer.branch(diff).then((message) => {
```
mtsgrd: Are you intentionally using `then` or would this be better written with `await`?

gitbutler-ui/src/routes/settings/+page.svelte
```diff
@@ -303,6 +304,10 @@
 				<GithubIntegration {userService} />
 			{/if}
 		</ContentWrapper>
+	{:else if currentSection === 'ai'}
```
mtsgrd: For sake of consistency with the rest of the codebase can we please use `==`?

mtsgrd:
Looking really solid 💪

gitbutler-ui/src/lib/components/BranchCard.svelte
Caleb-T-Owens: When I modified that part of the function, it wasn't yet declared as `async` but now it is so I'll move it over; thanks for the spot 👍

Caleb-T-Owens:
> Looking really solid 💪

Thanks for taking a look, I realise it turned into quite a large PR so I'd be happy to talk over any of the changes if you have any questions/thoughts about any choices

gitbutler-ui/src/lib/backend/aiProviders.ts
```diff
@@ -17,14 +20,73 @@ export interface AIProvider {
...
 }
+
+export class OpenAIProvider implements AIProvider {
+	private openAI: OpenAI;
+
+	constructor(
+		apiKey: string,
+		private model: OpenAIModel
+	) {
+		this.openAI = new OpenAI({ apiKey, dangerouslyAllowBrowser: true });
```
mtsgrd: I think it would be good to avoid calling constructors inside constructors, it will make it easier to test this code in the future. What if we made the OpenAI client a parameter?

gitbutler-ui/src/lib/backend/aiProviders.ts
```diff
@@ -17,14 +20,73 @@ export interface AIProvider {
...
+	}
+
+	async evaluate(prompt: string) {
+		const messages: PromptMessage[] = [{ role: MessageRole.User, content: prompt }];
+
+		const response = await this.openAI.chat.completions.create({
+			// @ts-expect-error There is a type mismatch where it seems to want a "name" paramater that isn't required https://github.com/openai/openai-openapi/issues/118#issuecomment-1847667988
```
mtsgrd: Make this a multiline comment?

mtsgrd:
For things operating on git like `gitGetConfig`, could we integrate them into an existing service or create a new one? I think we want to stick with using dependancy injection to keep the code testable, even though we currently don't have front end tests.

Everything else looks great to me!

gitbutler-ui/src/lib/components/AISettings.svelte
```diff
@@ -0,0 +1,239 @@
...
+	$: if (modelKind) summarizerSettings.setModelKind(modelKind);
+	$: if (form && modelKind) form.modelKind.value = modelKind;
+
+	let keyOption: { name: string; value: KeyOption } | undefined;
+	summarizerSettings
+		.getKeyOption()
+		.then(
+			(persistedKeyOption) =>
+				(keyOption = keyOptions.find((option) => option.value == persistedKeyOption))
+		);
```
mtsgrd, line 61: How about moving these into an `onMount(async ()=> {})` so you can use `await` instead of `then`?

gitbutler-ui/src/lib/components/AISettings.svelte
```diff
@@ -0,0 +1,239 @@
...
+	$: if (keyOption) summarizerSettings.setKeyOption(keyOption.value);
```
mtsgrd: Do these run on first initialization atm?

gitbutler-ui/src/lib/backend/summarizer.ts
```diff
@@ -80,3 +91,45 @@ export class Summarizer {
...
+
+interface SummarizerContext {
+	user?: User;
+}
+
+// This optionally returns a summarizer. There are a few conditions for how this may occur
+// Firstly, if the user has opted to use the GB API and isn't logged in, it will return undefined
+// Secondly, if the user has opted to bring their own key but hasn't provided one, it will return undefined
+export async function buildSummarizer(context: SummarizerContext): Promise<Summarizer | undefined> {
+	const gitConfig = new GitConfig();
```
mtsgrd: There is still something I want to figure out here. If we were testing either of the components where this is used, how would we swap this out for a mock `GitConfig` object? To me it feels like we need an `AiService` created in one of the `layout.ts` files and passed down to `BranchCard` and `CommitDialog`.

mtsgrd:
Great progress! 💪

gitbutler-ui/src/lib/components/AISettings.svelte
Caleb-T-Owens: Yes. Is there a clever way to prevent this, perhaps with the onMount? I looked to see if there was a svelte-y way to track changes and the solutions I found suggested having a second variable you assigned the previous value to which seemed very cumbersome.

gitbutler-ui/src/lib/components/AISettings.svelte
Caleb-T-Owens: I like this suggestion, will do 👍

gitbutler-ui/src/lib/backend/summarizer.ts
Caleb-T-Owens: I'll have a go at adding something like that.

In terms of how to provide the service, I'm not a huge fan of drilling the props down. There doesn't seem to be much use of the setContext and getContext API, is there a particular reason for this (and would it be reasonable to provide the GitConfig and AIService this way)?

Also, is AiSomething preferred over AISomething?

mtsgrd: Yeah I'm not a fan either, and in fact, we're going to move to `setContext` and `getContext`. The only obstacle is that we need to replace the `projectId` constructor params with a store.. or something. I will have a look at that this coming week!

> Also, is AiSomething preferred over AISomething

Yeah, but it's just my personal preference. I don't like having e.g. three capital letters in a row. 😏

gitbutler-ui/src/lib/components/AISettings.svelte
Caleb-T-Owens: Thinking about the AI service and how I want to interact with the RXJS, I might end up structuring them in such a way where this isn't a problem

Caleb-T-Owens: So, even after using a writable, I needed to add in an "initialized" check which I'm not super thrilled about https://github.com/gitbutlerapp/gitbutler/pull/3084/commits/206d6cce07edfc4bf4d082b440423b8ad0da7844.

I thought registering the subscriber after the set call would have done the trick.

mtsgrd: I'll pull down your branch locally so I can have a better look.

Caleb-T-Owens:
A musing I had was, the aiService is more or less the summarizer with the creation of the provider just brought inside the class.

I'm going to have a think about this, because we're currently not succeeding at making the provider easy to inject

Caleb-T-Owens:
> A musing I had was, the aiService is more or less the summarizer with the creation of the provider/client just brought inside the class.
>
> I'm going to have a think about this, because we're currently not succeeding at making the provider easy to inject

I was concerned about how to test it partially because I've not tried vitest before apart from seeing someone use it on stream, so I thought I'd give it a go, and I've been able to reasonably cover most of the logic

mtsgrd:
Just leaving a note that I'll review the new changes as soon as I get a chance!

Caleb-T-Owens:
I've just seen the new context changes, I'll go rebase in a bit and now the userService can be gotten from the userContext, I'll have a go passing it into the `commit` and `branch` methods... which I think I'll rename to `summarizeCommit` and `summarizeBranch` because it would then read as `aiService.summarizeCommit`

Caleb-T-Owens:
I've done the rebase. I'll do the tweaks I said I'll do in a little bit, but then I've got some other things I need to work on this weekend

mtsgrd:
I just sat down to look at this again. Sorry for my absence, I was deep in the weeds trying to wrap my head around both streams and dependency injection. ✌️

Caleb-T-Owens:
No worries @mtsgrd!

I've gone and done that refactor I mentioned.

Part of me is wanting to extract the prompt construction and post processing into their own functions to better test them. What are your thoughts on that?

mtsgrd:
@Caleb-T-Owens hey, sorry, I pushed here even though I meant to push to my own branch. I didn't realize GitButler can push to other peoples' forks. :)

Anyway, feel free to force push if you want to take some suggestions but tweak it. I just wanted to get this pushed before I go have some beers tonight. 🥳

Caleb-T-Owens:
@mtsgrd 

> hey, sorry, I pushed here even though I meant to push to my own branch. I didn't realize GitButler can push to other peoples' forks. :)
>
> Anyway, feel free to force push if you want to take some suggestions but tweak it. I just wanted to get this pushed before I go have some beers tonight. 🥳

No worries! GitHub has an option to permit maintainers to push to your fork which is pretty nifty. I can now pull them into my branch with the upstream commits card now too 💪!

Enjoy your evening!

---

I see you've changed https://github.com/gitbutlerapp/gitbutler/pull/3084/commits/1247e5253ad8d3d9cc130599686075f8a39e7e58#diff-c828d978a2adbaa41707e94fea99a318b081ed67c3c3fa2fb610416dbc4431bfR23-R24 from using null to undefined. I typed that using `null` because `invoke` seems to be returning `null` but it was more inconvenient due to the bind value of text boxes wanting to receive `string | undefined` so I'm happy to make the move over. I am going to add in a `?? undefined` to ensure our types are consistent with our actual values.

---

Unrelated to your changes, I was having a think about https://github.com/gitbutlerapp/gitbutler/pull/3084/files#diff-17085abea331547efcd2493dbd64eb700ea1f3108919330b2a6748eb56ab9060R30-R35 where we're saying `$: setContext(...)`.

To me this seems a little odd because if we're in the scenario where `data.something` changes and we set the state, none of the `getContext` calls will react to that as their components won't have re-rendered. Aside from that, I was also under the impression that the load() method is awaited before the `+layout.svelte` is rendered meaning that the items in the data object shouldn't be getting updated after the `+layout.svelte` is rendered. If that assumption is true, doesn't that mean that we don't need to react regardless?

Perhaps I'm missing some sveltey wizardry that is going on in the background that I don't know about, so I'm keen to hear about this

mtsgrd:
Just quickly before I head out.
- Yeah let's convert the `null` to `undefined`, and pretend `null` barely even exists in this code base
- The services located under `/[projectId]/` get update in the corresponding `layout.ts` file, but changing projects re-renders everything with a `getContext` call, so we're cool for now.
- I would perhaps prefer if we only had global services rather than some that depend on `projectId`, but let's refactor that later

gitbutler-ui/src/lib/backend/aiService.test.ts
```diff
@@ -0,0 +1,223 @@
...
+import type { GitConfigService } from './gitConfigService';
+
+const defaultGitConfig = Object.freeze({
+	'gitbutler.aiModelProvider': ModelKind.OpenAI,
+	'gitbutler.aiOpenAIKeyOption': KeyOption.ButlerAPI,
+	'gitbutler.aiAnthropicKeyOption': KeyOption.ButlerAPI,
+	'gitbutler.aiOpenAIKey': undefined,
+	'gitbutler.aiOpenAIModelName': OpenAIModelName.GPT35Turbo,
+	'gitbutler.aiAnthropicKey': undefined,
+	'gitbutler.aiAnthropicModelName': AnthropicModelName.Haiku
```
mtsgrd: Just realized we probably want to tie this to the new enum?

Caleb-T-Owens: Yes - I was going to do that 👍

gitbutler-ui/src/lib/backend/aiService.test.ts
```diff
@@ -0,0 +1,243 @@
...
+			expect(await aiService.buildClient()).toBe(undefined);
+			expect(toastErrorSpy).toHaveBeenLastCalledWith(
+				"When using GitButler's API to summarize code, you must be logged in"
+			);
+		});
+
+		test('When token is bring your own, When a openAI token is present. It returns OpenAIClient', async () => {
+			const gitConfig = new DummyGitConfigService({
+				...defaultGitConfig,
+				'gitbutler.aiOpenAIKeyOption': KeyOption.BringYourOwn,
```
mtsgrd: Could you still update these strings to point to `ConfigKeys`? On the same note, I didn't necessarily mean to leave it named what it is. Perhaps it would be better to rename it e.g. `GitConfigKey`?

mtsgrd:
I'm pretty stoked how well written this feature has become. Feel free to land the changes at any time 👍

mtsgrd (approved):

gitbutler-ui/src/lib/components/Select.svelte
```diff
@@ -13,8 +13,13 @@
 	export let labelId = 'label';
 	export let itemId = 'value';
 	export let value: any = undefined;
+	export let selectedItemId: any = undefined;
 	export let placeholder = '';
 
+	$: if (selectedItemId) {
+		value = items.find((item) => item[itemId] == selectedItemId);
+	}
```
mtsgrd: Does this perhaps fit on one row?

package.json
```diff
@@ -1,19 +1,20 @@
...
+		"build:development": "pnpm --filter @gitbutler/ui run build:development",
+		"check": "pnpm --filter @gitbutler/ui run check",
+		"lint": "pnpm --filter @gitbutler/ui run lint",
+		"format": "pnpm --filter @gitbutler/ui run format",
+		"tauri": "tauri",
+		"prepare": "pnpm --filter @gitbutler/ui run prepare"
+	},
+	"devDependencies": {
+		"@tauri-apps/cli": "^1.5.10"
+	}
```
mtsgrd, line 19: Side note, I wonder if we don't have a formatting specified for json in ci since it doesn't complain? We should perhaps keep the original unless we're now doing tabs.

Caleb-T-Owens: The tauri.conf.*.json files are all tabs and the gitbutler-ui/package.json is also tabs, but, I'd be happy to put that package.json back if we're not ready to make a particular decision on that.

mtsgrd: You're right, we should prefer consistency with the other json files. :)

gitbutler-ui/src/lib/backend/aiService.test.ts
Caleb-T-Owens: Yep, I missed those strings, I'll go and do those.

As for the name, part of me wants to indicate that they're related to the AI functionality and go for `GitAIConfigKey` to differentiate between some other config key enum in the future.

Part of me still likes the idea of having getters and setters for the fields, perhaps now on the GitConfigService, so we can say `await gitConfigService.getOpenAIKeyOption()` (or maybe as a method declared as `get`) so we don't need to worry about being consistent with default values in other areas of the code

mtsgrd: >await gitConfigService.getOpenAIKeyOption()

How many getters and setters do you think we could eventually end up with?

Caleb-T-Owens: > How many getters and setters do you think we could eventually end up with?

At the minute we've got two wrappers for local storage (persisted.ts & userSettings.ts), the gitbutler.json, various data stored in the new TOML files, and the git config.

I think there are several considerations around reducing the quantity of locations we store data, whether we want to sync the users settings to the could, and deciding on if we want functionality to provide project level overrides like vscode.

In the long run, my thought is that it would be helpful to have one unified interface (even if behind the scenes it writes to different providers) for accessing reading and writing settings so we don't end up with duplicate functionality and inconsistent interfaces.

---

Now is certainly not the time to reinvent that particular wheel. However, by adding these getters and setters, we're going to completely avoid any key name problems, inconsistent default problems, and typing problems in application code. To me it seems like a very sensible thing to do given that the cost of removing this abstraction is going to be pretty much equivalent to finding and replacing the `get` and `set` methods.

Caleb-T-Owens: I've use the enum in the other tests and renamed it.

Perhaps refactoring into setters and getters can be left to another time.

Something I do want to do before we merge in, is create a better metric for when the AI generation button should be disabled, because currently you must be logged in if you're providing your own token, but apart from that, if we're happy from a UI perspective, I'd be happy for this to go ahead and be merged in

Caleb-T-Owens: I've made the change to allow using AI without a user if a key is provided and updated the settings to better reflect that

gitbutler-ui/src/lib/components/BranchCard.svelte
```diff
@@ -112,7 +112,7 @@
 		generateBranchName();
 	}
 
-	onMount(() => {
+	onMount(async () => {
```
mtsgrd: Since there is no async calls in this onMount we don't need to pass it an async arrow function.

gitbutler-ui/src/lib/backend/aiService.ts
```diff
@@ -89,6 +89,35 @@ export class AIService {
 		private cloud: ReturnType<typeof getCloudApiClient>
 	) {}
 
+	async configurationValid(userToken?: string) {
```
mtsgrd: I think this function name is hard to interpret, could it be better described as `validateConfiguration`?

gitbutler-ui/src/lib/backend/aiService.ts
```diff
@@ -89,6 +89,35 @@ export class AIService {
...
+			(modelKind == ModelKind.OpenAI && openAIKeyOption == KeyOption.ButlerAPI) ||
+			(modelKind == ModelKind.Anthropic && anthropicKeyOption == KeyOption.ButlerAPI)
+		) {
+			return Boolean(userToken);
+		}
+
+		return Boolean(
+			(modelKind == ModelKind.OpenAI && openAIKey) ||
+				(modelKind == ModelKind.Anthropic && anthropicKey)
+		);
```
mtsgrd, line 118: Can we rewrite these two checks without the object type `Boolean`? We only have primitive strings and primitive boolean types in this code base.

In the latter I would just remove the boolean, and in the former I would return `!!userToken`.

mtsgrd:
How about we land this today and then iterate?

gitbutler-ui/src/lib/backend/aiService.ts
Caleb-T-Owens: Sure! I think I was going for the ruby pattern where we define proposition methods because `?` can be part of a name, so we'd be able to call `aiService.configurationValid?` but `validateConfiguration` or maybe doing `hasValidConfiguration` would be better.

Caleb-T-Owens: I also want to extract the large propositions into some intermediary constants to clarify the intent

gitbutler-ui/src/lib/components/BranchCard.svelte
Caleb-T-Owens: Thanks for spotting that. That is just something I missed when committing

Leopere:
this is exciting but does it add ollama support?  I'd like either ollama or grok or something that I get more choice or privacy with preferrably.

Caleb-T-Owens:
Hi @Leopere, this certainly sets the groundwork for adding in supporting custom models over a user provided API endpoint, but doesn't yet provide that option.

In my dev environment I hacked in a basic AIClient that talked to ollama (which took about 5 minutes to do), but it made it clear that there is a good portion of work that needs to be done, before these weaker models can be anywhere near useful.

I summarised what I deemed to be prerequisite on discord: https://discord.com/channels/1060193121130000425/1204524097241878629/1219776902978600970. (invite to the server if needed: https://discord.gg/B7rv6qBejb)

Leopere:
Yeah it’s kind of a shame that we’re not quite close to parity in oss to accomplish onprem llms yet

PavelLaptev:
@Caleb-T-Owens @mtsgrd if you do new updated and this involves UI — include me to reviewers

PavelLaptev:
Otherwise, I'll change what you did afterward. Which is bad practice.

Caleb-T-Owens:
@PavelLaptev will do 👍

PavelLaptev:
@Caleb-T-Owens thanks 🙏

PR #3089, opened 2024-03-09
Integrate alls-green in rust-check job to improve status checks

### LINKED ISSUE #2848: [CI bug][drive-by recommendation] An incorrectly configured `check-rust` job Unified gate setup recommendation
I see the project's CI setup has a job called `check-rust` that is set up *incorrectly*.
It seems to have been added in hopes to act as a unified gate for use in the branch protection, allowing only having a single GHA job listed in the repo settings. However, with the way it's currently set up, this job is always treated by GitHub's branch protection as successful — whether it's skipped because the job dependencies failed or actually got executed. I've seen this problem so many times that I came up with a GitHub Action that allows setting up a job that actually works as an ultimate gate for letting PRs go through merge or blocking them reliably.

Here it is, with a few paragraphs of explanation: https://github.com/marketplace/actions/alls-green/#why. Wiring it up properly may allow you drop the unnecessary statuses from the branch protection rules leaving only one GHA status as required.

OlenaYefymenko:
This patch integrates alls-green, that checks whether all dependency jobs have passed successfully. This change also simplifies the CI setup by requiring only the rust-check job for status checks.

Resolves #2848

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -170,14 +170,28 @@ jobs:
           features: ${{ toJson(matrix.features) }}
           action: ${{ matrix.action }}
 
+# See: https://github.com/re-actors/alls-green
   check-rust:
+    if: always()
     needs:
       - changes
       - check-gitbutler-app
       - check-gitbutler-changeset
       - check-gitbutler-core
     runs-on: ubuntu-latest
-    if: ${{ needs.changes.outputs.rust == 'true' }}
     steps:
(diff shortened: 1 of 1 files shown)


mtsgrd:
@Qix- would you be able to review this? It loos great to me, but I don't understand it well enough to merge.

Qix-:
I didn't see this for some reason, was just about to do this myself. Looks good to me, let's try it :)

PR #3110, opened 2024-03-11
Do not filter to remote branches only

schacon:
Way back in November, we reverted functionality where we would list local branches in addition to remote branches in our sidebar. It makes sense, we want to be able to convert local branches to virtual branches. But due to naming confusion, the branches have been restricted to remote branches ever since.

This PR is exploring what we need minimally to revert this back to listing all branches that you can turn into virtual branches.

- [x] filter out `gitbutler/integration`
- [ ] filter out remote branches that are already fully contained in any virtual branch head


krlvi:
Consider renaming the command or alternatively I can do it in a new PR

schacon:
I'm mainly just pushing up my notes to make sure you see them, but we need to decide if we want to refactor _all_ the code around "remote" branches or just get something minimal working and address it later. Now all the Typescript code has been renamed to `RemoteBranch` (it used to be the more generic `BranchData`), the Rust `struct`s are all "RemoteX", there's Remote stuff all over the place, when they all meant "normal git branch" for a while.

But now I'm seeing some other bugs in applying remote branches that are also issues converting local branches into virtual branches, so I'm going down the rabbit hole of figuring out why that is exactly.

PR #3115, opened 2024-03-11
fixup more instances of string path handling

Qix-:


Qix-:
(this PR will most likely fail, in which case it'll be abandoned because I pushed the boundaries a bit with these particular changes)

PR #2789, opened 2024-02-18
Add character count feedback for commit title

dinolindlau:
- display the remaining character count, advice is 50
- only available for the first line of the commit, commit title

https://github.com/gitbutlerapp/gitbutler/assets/61562467/53df04ab-13d3-409d-b663-548f3efc4216

PR #3134, opened 2024-03-13
Add summary and description fields. Show tooltip for long summary messages

PavelLaptev:
Based on https://github.com/gitbutlerapp/gitbutler/pull/2789

Changes:
- Added two fields for summary and description
- If summary is too long, show tooltip icon

https://github.com/gitbutlerapp/gitbutler/assets/18498712/402a6dcb-9051-49b8-9882-d6f11056cdd3


gitbutler-ui/src/lib/components/CommitDialog.svelte
```diff
@@ -39,31 +40,55 @@
...
-		textareaElement.style.height = 'auto';
-		textareaElement.style.height = `${textareaElement.scrollHeight + 2}px`;
+	function getCommitMessageTitleAndDescription(commitMessage: string) {
+		// Split the commit message into title and description
+		// get the first line as title and the rest as description
+		const [summary, ...description] = commitMessage.split('\n');
+		return {
+			summary: summary || '',
+			// remove top new line
+			description: description.join('\n').replace(/^\n/, '') || ''
```
mtsgrd: Perhaps we could `.trim()` rather than `.replace(/^\n/,'')` since it would remove both leading and trailing whitespace?

*edit: actually, it might be nicer to split with regex to limit the number of splits.

```ts
const [summary, description] = commitMessage.split(/\n(.*)/s)
```

*edit2: for trimming both strings I think you'd need the following:

```ts
const [summary, description] = commitMessage.trim().split(/\n+(.*)/s)
```

PavelLaptev: Make sense

gitbutler-ui/src/lib/components/CommitDialog.svelte
```diff
@@ -39,31 +40,55 @@
...
+	let isTextareaFocused = false;
+	let isSecondTextareaFocused = false;
+
 	function focusTextareaOnMount(el: HTMLTextAreaElement) {
 		if (el) el.focus();
 	}
 
 	function commit() {
-		if (!commitMessage) return;
+		// console.log('commitMessage', commitMessage);
```
mtsgrd: 👀

gitbutler-ui/src/lib/components/CommitDialog.svelte
```diff
@@ -39,31 +40,55 @@
...
+		// get the first line as title and the rest as description
+		const [summary, ...description] = commitMessage.split('\n');
+		return {
+			summary: summary || '',
+			// remove top new line
+			description: description.join('\n').replace(/^\n/, '') || ''
+		};
+	}
+
+	function concatCommitMessage(message: { summary: string; description: string }) {
```
mtsgrd: This looks a bit weird to me, can this function take two regular parameters called summary and description?

gitbutler-ui/src/lib/components/CommitDialog.svelte
PavelLaptev: Sowwy

gitbutler-ui/src/lib/components/CommitDialog.svelte
PavelLaptev: Yes, will fix it

gitbutler-ui/src/lib/components/CommitDialog.svelte
```diff
@@ -39,31 +40,55 @@
...
+		return `${message.summary}\n${message.description}`;
 	}
 
+	let commitMessageSet = getCommitMessageTitleAndDescription(get(currentCommitMessage));
```
mtsgrd: You should be able to replace `get(currentCommitMessage)` with `$currentCommitMessage`, but I would also see if we can omit having a `commitMessageSet` type object and instead destructure this declaration into commitTitle and commitDescription.

PavelLaptev: Got it

mtsgrd:
Sorry for leaving so many comments here, but I'd love for you to do one more thing, in the spirit of constantly improving the quality of this code base. Could you rewrite the `commit()` function to an `async function` that uses await rather than `.then(...)`? I know you didn't write that function, but it would be nice thing to do. :)

PavelLaptev:
All good, my aim is to make it right 🙂

gitbutler-ui/src/lib/components/CommitDialog.svelte
PavelLaptev: Working on these fixes now. I added an object as a parameter here because `commitMessageSet` is an object, so it's easier just to pass the whole object like this `concatCommitMessage(commitMessageSet)`

gitbutler-ui/src/lib/components/CommitDialog.svelte
PavelLaptev: Done, description returning as an array, so I just take the first item `description[0]`

gitbutler-ui/src/lib/components/CommitDialog.svelte
```diff
@@ -49,19 +48,19 @@
 	function getCommitMessageTitleAndDescription(commitMessage: string) {
 		// Split the commit message into title and description
 		// get the first line as title and the rest as description
-		const [summary, ...description] = commitMessage.split('\n');
+		const [summary, ...description] = commitMessage.trim().split(/\n+(.*)/s);
```
mtsgrd: You can remove the `...` now, making description a `string` rather than `string[]`.

PavelLaptev: Ah, I see. That's why it's an array. I completely forgot about this one. Thanks!

PavelLaptev:
@mtsgrd thank you! I'll QA it again and then merge

PavelLaptev:
Seems like a condition race here - textarea takes the height of the previous condition. I'll check if I can fix it 🙂 

https://github.com/gitbutlerapp/gitbutler/assets/18498712/a5e5cf02-fa9a-4bc7-b863-f474f70e28b4

PavelLaptev:
> Seems like a race condition here - textarea takes the height of the previous condition. I'll check if I can fix it 🙂
> 
>  Screen.Recording.2024-03-14.at.21.59.50.mov

@mtsgrd I can do something like this:
```
$: if ($commitMessage) {
		setTimeout(() => {
			setAutoHeight(titleTextArea);
			setAutoHeight(descriptionTextArea);
            
			if (descriptionTextArea) {
				descriptionTextArea.focus();
			}
		}, 0);
	}
```
But I'm not sure if using `setTimeout` is the best way to handle this. Let me know if you have any suggestions.

PR #3150, opened 2024-03-14
send more context with feedback (OS, browser, index size)

### LINKED ISSUE #3148: send more context with feedback
We have a `context` field in our feedback API that we're not using. It would be really helpful for some of these support requests to fill that with some useful contextual data:

* GitButler version
* Operating system information (something like `uname -v`?)
* Index size (something like `project_repo.index.len()`)
* Browser (something like `window.navigator.userAgent`)

schacon:
Now sends `context` field with feedback. Something like this:

```
GitButler Version: 0.5.346
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)
URL: http://localhost:1420/aa587eef-2aa2-4ed4-9fb7-f90c7e5f2527/board
Length of index: 133
```

Closes https://github.com/gitbutlerapp/gitbutler/issues/3148

### DIFF
--- a/gitbutler-app/src/app.rs
+++ b/gitbutler-app/src/app.rs
@@ -145,6 +145,15 @@ impl App {
             .map_err(|e| Error::Other(anyhow::anyhow!(e.to_string())))
     }
 
+    pub fn git_index_size(&self, project_id: &ProjectId) -> Result<usize, Error> {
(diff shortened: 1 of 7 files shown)


schacon:
Nevermind, figured it out.

PR #3159, opened 2024-03-15
refactor: save virtual branches state in a flat file

krlvi:


Caleb-T-Owens:
Hi! I was wondering why a separate TOML file was chosen when we already have a sqlite database that is storing various deltas?

krlvi:
The sqlite database is only used as a cache layer for the CRDTs the app stores (we had a local replay functionality which is currently not exposed in the app) - but it's not related to vbranches.

Right now the state of vbranches is persisted in `<app_data_dir>/projects/<project_id>/gitbutler/branches` 
That's the `branch` and `target` structs. Attributes are placed in files nested in the file tree, for example ownership info is in:
`<app_data_dir>/projects/<project_id>/gitbutler/branches/<branch_id>/meta/ownership`. Its quite tedious to debug and manage actually, so I'm exploring some options :)

Caleb-T-Owens:
Cool cool!

mtsgrd:
🤘

PR #3185, opened 2024-03-18
Rename/refactor a few things related to `list_virtual_branches` and remove the ownership bubble up functionality

mtsgrd:


krlvi:
The refactor part looks good. The part where you remove the bubble up ownership functionality - you should create a new test asserting that order is sequential

Qix-:
Regarding the Hunk to GitHunk, is this clearing up some ambiguity?

mtsgrd:
>Regarding the Hunk to GitHunk, is this clearing up some ambiguity?

Yeah I was looking at the code in `fn get_applied_status` and it was converting between `git::diff::Hunk` and `virtualbranches::branch::Hunk` and it was really confusing how we were converting one hunk to another with `Hunk::from`. Even after figuring out that these were different hunks I found it really hard to read the code.

krlvi (approved):

PR #3197, opened 2024-03-18
add PID lockfile crate

Qix-:


Qix-:
Closing this in favor of tempfile-rename strategy.

PR #3198, opened 2024-03-18
Simplify padding in commit dialog

mtsgrd:
- under some conditions we lost all padding in commit dialog
- with these changes the bottom padding is always correct


PavelLaptev:
@mtsgrd hmm. I'll check it. With this changes paddings now are too big 
image

PR #3220, opened 2024-03-19
Add `head: string` property to `Branch` class for front end

mtsgrd:
Needed for interacting with input prompts from the back end.


mtsgrd:
Not needed

PR #3232, opened 2024-03-20
Consolidate on one commit splitting implementation

Caleb-T-Owens:
We had four different instances in the app where we implement commit message splitting making it clear that its a common operation so I've added in a method that does this so we can use one consistent and tested implementation.

I chose to move away from the strange usage of `.split` over to regular RegExp capturing groups because IMO by the time you're reading [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/@@split) documentation page and paying attention to [this bulletpoint](https://arc.net/l/quote/qrgagdgr), you're no longer reading "obvious" code 😆

### DIFF
--- a/gitbutler-ui/src/lib/utils/commitMessage.ts
+++ b/gitbutler-ui/src/lib/utils/commitMessage.ts
@@ -1,0 +1,10 @@
+const commitMessageSplitRegExp = /(.*?)(?:\n+|$)(.*)/s;
+
+export function splitMessage(message: string) {
+	const matches = message.match(commitMessageSplitRegExp);
+
+	return {
+		title: matches?.[1] || '',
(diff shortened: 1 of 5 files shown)


gitbutler-ui/src/lib/utils/commitMessage.ts
```diff
@@ -0,0 +1,10 @@
+const commitMessageSplitRegExp = /(.*?)(?:\n+|$)(.*)/s;
```
mtsgrd, line 1: 🔥

mtsgrd (approved):

Caleb-T-Owens:
I'm happy for these changes to be merged when ready

PR #3243, opened 2024-03-20
feat: allow signing rebased commits

### LINKED ISSUE #3147: rebased commits are not re-signed
When a branch is updated and commits are rebased and SSH signing is on, we do not properly re-sign the commits. We should be checking for the signing option and re-signing commits that are rebased.

Pranav2612000:
Not sure if this is the best way to do it. Especially, setting `rebase_options.inmemory(false);` to false.
If not, I'll try to make it work by adding `git_commit_create_cb` support to git2

Fixes #3147 

cc: @schacon @krlvi

### DIFF
--- a/gitbutler-app/src/project_repository/repository.rs
+++ b/gitbutler-app/src/project_repository/repository.rs
@@ -8,7 +8,7 @@ use anyhow::{Context, Result};
 
 use crate::{
     askpass::AskpassBroker,
-    git::{self, credentials::HelpError, Url},
+    git::{self, credentials::HelpError, Oid, Url},
     keys,
     projects::{self, AuthKey},
     ssh, users,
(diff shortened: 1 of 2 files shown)


schacon:
I don't believe that this will work. If rebase is not in `in_memory` mode, it will try messing with the HEAD and working directory, which is going to put the project in a weird state most likely. 

I think what's probably much simpler would be to do exactly what we're doing, but at the end, if the rebase is successful, simply rewrite all the rebased commits, reuse the trees and manually write the commits with the signature.

So, basically, here: https://github.com/gitbutlerapp/gitbutler/blob/a1e2b92abcbfecf038f8b00c2ca25512731fcb0b/gitbutler-app/src/virtual_branches/virtual.rs#L1557

If there is a signing key, instead of just saving the rebased head here, we should calculate the list of commits from the base (`merge_base`) to that head (`last_rebase_head`), then walk through them starting from the first child of the base and for each one, rerun `project_repository.commit()` with the user, message and tree from that commit, and the parent that was written last. (does that make sense?)

Byron:
Thanks for contributing!

Since the PR was created, a lot has changed which puts it into a conflicting stage unfortunately. Also I hope you will find time to try out the suggestions.

If that's not the case, please feel free to close it and resubmit when that changes.

Thank you.

schacon:
Hey @Pranav2612000, I just refactored how commits are done in general in a way that should sign everything properly now. The change is in #3721. I'll close this, since it's out of date and based on code that has been pretty massively changed and now should address this.

Pranav2612000:
Sure. Thank you @schacon 
Sorry for not replying earlier.

I think the cleanest way to do this would be to use the [git_commit_create_cb](https://libgit2.org/libgit2/#HEAD/group/callback/git_commit_create_cb) method of libgit2. Since git2rs does not yet have support for it, I've opened up a PR [ https://github.com/rust-lang/git2-rs/pull/1047 ] to add this support and was waiting for this to be merged.

I'll open up this conversation again once the method is available in libgit2 and we can then decided if we want to use it or not. Does that work?

PR #3254, opened 2024-03-20
Pass in hunk array rather than diff

Caleb-T-Owens:
I wanted to move the generation of the diff into a function inside the aiService.ts so we could perform standard formatting and shuffling operations.

This PR specifically:
- Updates the `diffLengthLimit` from `20000` to `5000`. Turns out that before the diff was getting passed to the models, it was getting trimmed down to 5000 characters and it seems to work fine, so I've updated this constant now that we are *actually* making use of it.
- Includes the file name before each hunk. This provides a little more context to the AI and I've found that it will occasionally pick up on the file names and include them in the generated commit messages
- Introduced hunk shuffling. This means that we're not going to fill up the context with 5000 characters of one particular file, and in stead take a smattering of changes from all the files present.
- Changed the summarizeXXX methods to take in `Hunk[]` so we can perform the diff generation in a consistent manner

### DIFF
--- a/gitbutler-ui/src/lib/backend/aiService.test.ts
+++ b/gitbutler-ui/src/lib/backend/aiService.test.ts
@@ -8,9 +8,12 @@ import {
 	GitAIConfigKey,
 	KeyOption,
 	ModelKind,
-	OpenAIModelName
+	OpenAIModelName,
+	buildDiff
 } from '$lib/backend/aiService';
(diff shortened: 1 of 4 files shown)


Caleb-T-Owens:
Screenshot 2024-03-20 at 22 13 54

Random example of that file name getting used

mtsgrd (approved):

PR #3256, opened 2024-03-21
Refactor prompt service to be easier to use

mtsgrd:


gitbutler-ui/src/lib/backend/prompt.ts
```diff
@@ -10,40 +24,121 @@ export type SystemPrompt = {
...
-			this.prompt$.next(e.payload);
-		} else {
-			// Always cancel actions that are marked "auto", e.g. periodic sync
-			await this.cancel(e.payload.id);
-		}
-	});
-
-	constructor() {}
+	// Timeout id for any ongoing prompt
+	private timeout: any;
```
mtsgrd: I should remove this, not used anymore.

gitbutler-ui/src/lib/backend/prompt.ts
```diff
@@ -10,40 +24,121 @@ export type SystemPrompt = {
...
+type FilterParams = {
+	// This can be e.g. push, fetch, auto etc
+	action?: string | undefined;
+	// Filter to a specific `BranchLane`
+	branchId?: string | undefined;
+	// Time until auto closing prompt
+	timeoutMs?: number | undefined;
+};
+
 export class PromptService {
```
mtsgrd, line 47: Needs clarification that this class exposes stores through the filter method, and as long as at least one of them is subscribed to this service listens to tauri for git_prompt events.

Caleb-T-Owens: I was having a think about the name of this service, and perhaps I've been thinking too much about prompting AI over the last week, but I was wondering if we should be slightly more specific in the name to avoid confusion between the two?

That said, I can't think of a particularly nice with the best I've come up with being `UserInputDialogService`, so perhaps this is fine and we just need to be more specific on the AI side if we want to use the name `Prompt` in public interfaces there

PR #3282, opened 2024-03-22
refactor: Update function parameter to use sortedFiles instead of files

schacon:
This fixes the issue where arrow keys seem to select random order of files. I think the `files` list is by last modified rather than alphabetical as it's shown visually.


mtsgrd:
Thanks @schacon!

PR #3294, opened 2024-03-23
refactor(AISettings): Remove unneeded config set calls

Caleb-T-Owens:
There were still some calls to set the configuration values which where superfluous, so I've gone and fixed the problem by adding in a function which reads an `initialized` variable which is set after the onMount has finished running.

I have also made all the values optionally undefined as we were previously getting runtime errors because before it is set from in the onMount hook, it is an undefined value.

---

This is an behind the scenes change


krlvi (approved):
lgtm

PR #3297, opened 2024-03-24
Fix bugs occasionally preventing hunks from being moved

mtsgrd:
- hunks equal if hash is same, regardless of line numbers
- send hash from front end when updating ownership
- use updated line numbers after hunk intersection match


krlvi:
Should work. Just keep in mind that if there's the same change there would be a hash collision. Probably more rare than the line number issue though

mtsgrd:
Yeah, should be rare, especially with 3 lines of context!

Issue #3121, opened 2024-03-12
UX of virtual branch commit states (discussion)


krlvi:
## Background:

The nature of branches having a local and remote equivalent means that commits can be in multiple states.
1. If the local and remote are identical, there’s nothing to do
2. If the local has commit(s) that are not on the remote, those commits can be pushed
3. If the remote has commit(s) (from another user or the GitHub GUI) that are not local, they can be pulled
4. If the local branch has non-punished commits and the remote has non-pulled commits the branches are diverged

I find this graphic by Julia Evans to be quite nice at illustrating this:

![diverged](https://github.com/gitbutlerapp/gitbutler/assets/4030927/c8ebd9b0-4b98-4ad4-9f1e-de9e3507d359)

## Problem

The lanes GitButler renders for virtual branches don’t represent the `need to pull` and `diverged` states clearlly. To illustrate this, I made a graphic showcasing the way the 4 states are shown by GitButler

## Current state in GitButler

Below are 4 screenshots next to each other, covering the 4 states mentioned previously and how they are represented within GitButler virtual branches, alongside some commentary. I generated these states by making a commits using the GitHub website. 
 
<img width="1570" alt="current_lanes" src="https://github.com/gitbutlerapp/gitbutler/assets/4030927/b25b4d7b-3e84-408f-a909-0b66f0c61273">

Comments from the graphic 👆 (the numbers in blue)
- (1) “Upstream commits” duplicate what is local
- (2) “Merge” is misleading as this would mean simply pulling the remote commit
- (3) “Remote branch” is a misleading term as here we show only local commits
- (3) “Force push to remote” doesn’t make sense as this is in the context of remote commits
- (3) “Force push” may have unintended consequences (losing the remote commit)
- (4) Duplicated from below
- (5) “Merge” is a bit misleading
- (6) “Force push to remote”, unclear difference from the other force push button, possibly unintended consequences 
- (7) Possibly unintended consequences 

## Lets discuss 💬

The purpose of this GH issue is to invite a discussion and maybe go over some drafts / ideas that come up. @schacon, @mtsgrd, @Qix- @PavelLaptev let me know if I have missed something

mtsgrd:
@krlvi I should perhaps know the answer by now, but what is the difference between "upstream commits" and commits on "remote branch"?

krlvi:
> @krlvi I should perhaps know the answer by now, but what is the difference between "upstream commits" and commits on "remote branch"?

[...]

PR #3298, opened 2024-03-24
feat: make code tab size configurable

mayfieldiv:
I'm fully open to feedback on the location/text of the setting, but here's what it looks like initially:
image

Example code section with a tab size of 2 rather than the default 4:
image


gitbutler-ui/src/lib/components/HunkLine.svelte
```diff
@@ -1,9 +1,12 @@
 <script lang="ts">
 	import { create } from '$lib/components/Differ/CodeHighlighter';
+	import { SETTINGS_CONTEXT, type SettingsStore } from '$lib/settings/userSettings';
 	import { SectionType } from '$lib/utils/fileSections';
-	import { createEventDispatcher } from 'svelte';
+	import { createEventDispatcher, getContext } from 'svelte';
 	import type { Line } from '$lib/utils/fileSections';
 
+	const userSettings = getContext(SETTINGS_CONTEXT) as SettingsStore;
```
mtsgrd: This is great, but instead of getting this context here each time per line, how about moving it to `HunkViewer` and passing in the tab size as a prop to this component?

mayfieldiv: ah yes, that makes much more sense :p

PavelLaptev:
@mayfieldiv thank you for the PR :-)
Let me make a few design changes, and it's ready to go, unless there are any new code fixes required.

gitbutler-ui/src/lib/components/HunkLine.svelte
mtsgrd: Sorry, I have to ask.. Would it make even more sense to set this as a property on the hunk viewer level? The `HunkLine` component currently takes quite primitive properties, and can be instantiated without context dependencies.

mayfieldiv: Potentially, but at that point should `project` also be passed down? Would they be gotten at the `FileCard` level and prop-drilled down?

PavelLaptev:
Done with my changes
@mtsgrd if all good we can merge this

image

gitbutler-ui/src/routes/settings/+page.svelte
```diff
@@ -176,10 +176,32 @@
...
+			<SectionCard orientation="row" centerAlign>
+				<svelte:fragment slot="title">Hunks tab size</svelte:fragment>
+
+				<svelte:fragment slot="actions">
+					<TextBox
+						type="number"
+						width={100}
+						textAlign="center"
+						value={$userSettings.tabSize.toString()}
+						minVal={2}
```
mayfieldiv: I'd imagine some people prefer a tab size of 1, so maybe that should be the minimum?

gitbutler-ui/src/lib/components/TextBox.svelte
```diff
@@ -60,20 +72,54 @@
...
+			}}
 			on:change={(e) => dispatch('change', e.currentTarget.value)}
 			on:keydown={(e) => dispatch('keydown', e)}
 		/>
 
+		{#if type == 'number'}
+			<div class="textbox__count-actions">
+				<button
+					class="textbox__count-btn"
+					on:click={() => {
```
mayfieldiv, line 100: From quick testing, it seems like the setting is only being saved when typing numbers, but not when just clicking the +/-

mayfieldiv: So maybe these buttons should also dispatch `change` events?

gitbutler-ui/src/routes/settings/+page.svelte
mayfieldiv: I went ahead and made this change, so please let me know if you disagree

gitbutler-ui/src/lib/components/TextBox.svelte
mayfieldiv: I went ahead and made this change, so please let me know if you disagree

gitbutler-ui/src/routes/settings/+page.svelte
PavelLaptev: All good. Thanks, didn't know about this

gitbutler-ui/src/lib/components/TextBox.svelte
PavelLaptev: 👍

PavelLaptev (approved):

gitbutler-ui/src/lib/components/HunkLine.svelte
mtsgrd: You're right.. and then we're just creating more props. Let's go with what you have right now!

mtsgrd:
> making a mess of things with force-pushes and duplicate commits and I'm not quite sure what I'm doing wrong

We're aware, and we're going to try and make things better as soon as possible. Meanwhile, check this issue for more on what we're planning: https://github.com/gitbutlerapp/gitbutler/issues/3121

mtsgrd (approved):
lgtm!

mtsgrd:
Also, thank you @mayfieldiv for contributing! 💪

PR #3301, opened 2024-03-24
fix: improve `Segment` button selected state style

### LINKED ISSUE #3300: Style of list/tree button group is confusing
I find this styling of the buttons confusing:

image

Which button is currently pressed? For me, it looks like the "tree" button (right) is pressed, which means that the clickable button is the "list" button (left), but this isn't the case.

I suggest switching the styles of the pressed and unpressed states.

If this sounds like a good idea, I can create a PR.

eyalch:
The changes here swap the styles of the _selected_ and _unselected_ states of the `Segment` button component.

Resolves #3300

### DIFF
--- a/gitbutler-ui/src/lib/components/SegmentControl/Segment.svelte
+++ b/gitbutler-ui/src/lib/components/SegmentControl/Segment.svelte
@@ -37,7 +37,7 @@
 	role="tab"
 	aria-selected={isSelected}
 	aria-disabled={disabled}
-	tabindex={isSelected ? 0 : -1}
+	tabindex={!isSelected ? 0 : -1}
 	{...$$restProps}
 	on:mousedown|preventDefault={() => {
 		if (index !== $selectedSegmentIndex && !disabled) {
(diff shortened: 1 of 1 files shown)


eyalch:
I haven't yet tested this locally, so if someone could help that'll be great!

eyalch:
> I haven't yet tested this locally, so if someone could help that'll be great!

Tested locally. LGTM and behaves as I expect.

PavelLaptev:
@eyalch thanks for the PR! I'll check it out 🙂

gitbutler-ui/src/lib/components/SegmentControl/Segment.svelte
```diff
@@ -37,7 +37,7 @@
 	role="tab"
 	aria-selected={isSelected}
 	aria-disabled={disabled}
-	tabindex={isSelected ? 0 : -1}
+	tabindex={!isSelected ? 0 : -1}
```
eyalch: @PavelLaptev Consider avoiding this double-negative.

```suggestion
	tabindex={isSelected ? -1 : 0}
```

PavelLaptev: Thanks!

PavelLaptev:
Looks good to me. I'll merge it in 10 mins 🙂 

https://github.com/gitbutlerapp/gitbutler/assets/18498712/7cf735dd-0e7d-4980-83e0-9da544afabfb

PR #3303, opened 2024-03-24
feat: allow submitting rename branch form on pressing Enter

### LINKED ISSUE #3302: Improve accessibility of branch rename form
The accessibility of the branch rename form/modal ("Set branch name" menu item) could be improved.

Two simple improvements I can think of are:

- Allow the user to press <kbd>Enter</kbd> to submit
- Auto-focus the branch name input

eyalch:
In addition, auto-focus the branch name input.

Resolves #3302

### DIFF
--- a/gitbutler-ui/src/lib/components/BranchLanePopupMenu.svelte
+++ b/gitbutler-ui/src/lib/components/BranchLanePopupMenu.svelte
@@ -141,7 +141,7 @@
 
 	<svelte:fragment slot="controls" let:close>
 		<Button color="neutral" kind="outlined" on:click={close}>Cancel</Button>
-		<Button color="primary" type="submit" form="branch-name-form">Rename</Button>
+		<Button color="primary" form="branch-name-form">Rename</Button>
 	</svelte:fragment>
 </Modal>
 
--- a/gitbutler-ui/src/lib/components/BranchLanePopupMenu.svelte
+++ b/gitbutler-ui/src/lib/components/BranchLanePopupMenu.svelte
@@ -120,23 +120,22 @@
 	</ContextMenu>
 {/if}
 
-<Modal width="small" bind:this={renameRemoteModal}>
-	<svelte:fragment let:close>
-		<form
-			id="branch-name-form"
-			on:submit={() => {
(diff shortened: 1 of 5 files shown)


PavelLaptev:
@eyalch thanks for the PR! I also made the copy update you mentioned in the issue

gitbutler-ui/src/lib/components/BranchLanePopupMenu.svelte
```diff
@@ -121,20 +121,22 @@
...
 <Modal width="small" bind:this={renameRemoteModal}>
-	<TextBox label="Remote branch name" id="newRemoteName" bind:value={newRemoteName}></TextBox>
-
-	<svelte:fragment slot="controls" let:close>
-		<Button color="neutral" kind="outlined" on:click={close}>Cancel</Button>
-		<Button
-			color="primary"
-			on:click={() => {
+	<svelte:fragment let:close>
+		<form
```
mtsgrd: Instead of introducing a form with `on:submit`, how about having an `on:keydown` handler on the textbox? If we make a submit function we could do e.g, and we don't need to introduce restprops.

```ts
  on:keydown={(e) => {
    if (e.detail.key == 'Enter') submit();
  }}
```

eyalch: @mtsgrd What would you do if we had more than one input?

I know that this isn't the case, but I'm not sure if a form with single input should be treated as a special case.

eyalch: Also, with the `on:submit` approach, you don't have to hard-code the `Enter` key.

mtsgrd: Hey sorry for the late response. You're right.. it's a bit clumsy to check the enter key everywhere. But.. in this case, what if we made it part of the `Modal` component so we don't have to ad-hoc add form elements? Then we could e.g. remove the key check from `PromptModal` as well.

mtsgrd:
I moved the form to the modal so that it is accessible to all components using it. Thanks for contributing @eyalch!

PR #3285, opened 2024-03-22
fix: Change help text for lane icon to "Expand lane"

schacon:
CleanShot 2024-03-22 at 11 47 24@2x

PR #3304, opened 2024-03-24
feat: Update "Collapse lane" help text to "Expand lane"

krlvi:


krlvi:
strange that this one here https://github.com/gitbutlerapp/gitbutler/pull/3285 is red while this one passes... they are identical

PR #3307, opened 2024-03-24
feat: detect conflicting ownership claims

krlvi:


gitbutler-app/src/virtual_branches/virtual.rs
```diff
@@ -1799,18 +1800,39 @@ fn set_ownership(
...
-            if !taken.is_empty() {
-                branch_writer.write(branch).context(format!(
-                    "failed to write source branch for {}",
-                    file_ownership
-                ))?;
+            branch_claims.push((branch.clone(), taken));
+        }
+    }
+
+    // Check consistency and reject changes if they would result in a file being claimed by multiple branches
```
mtsgrd: I think s/file/hunk on this line?

But also, just want to make sure we're on the same page. This function won't result in a hunk being claimed by multiple files, instead if two branches claimed the same file this would remove duplicates?

krlvi: yeah, used the wrong word in the comment.

regarding the check - its putting `file-startline-endline` in a hash set, and if that gets repeated errors out, thus asserting that a hunk cant be in two branches at the same time with the update branch operation.

mtsgrd: Right, makes sense.

PR #3334, opened 2024-03-26
tests work with init.defaultBranch

Byron:
The test-suite hardcodes `refs/heads/master`. However, `git` somewhat selectively picks up configuration that affects the default branch name, like `GIT_BRANCH_DEFAULT` and `init.defaultBranch`.

This causes `HEAD` to contain `refs/heads/main`, which typically isn't created, causing 199 tests to fail.

Now `git2` repository initialization will override the branch name. The only notable exception to this is a submodule test, which needs the `HEAD` branch to be set specifically after initialization.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -262,7 +262,7 @@ PUBLIC_API_BASE_URL=https://app.gitbutler.com/
 
 ## Joining the GitButler Team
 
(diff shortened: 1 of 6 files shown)


krlvi:
Thank you for fixing this, I wasn't aware of the `init.defaultBranch` config! 🚀

PR #3337, opened 2024-03-26
accessibility fix: commit card folding behaviour on keyup

PavelLaptev:
I was able to fold/unfold it pressing any key. Fixed to `Enter` and `Space` keys only

https://github.com/gitbutlerapp/gitbutler/assets/18498712/d126f527-add6-4b27-867f-d5afb9fd2210

### DIFF
--- a/gitbutler-ui/src/lib/components/CommitCard.svelte
+++ b/gitbutler-ui/src/lib/components/CommitCard.svelte
@@ -46,17 +46,14 @@
 		files = await listRemoteCommitFiles(project.id, commit.id);
 	}
 
-	function onClick() {
+	function toggleFiles() {
 		showFiles = !showFiles;
 		if (showFiles) loadFiles();
 	}
 
 	function onKeyup(e: KeyboardEvent) {
-		if (e.key === 'Enter' || e.key === ' ') {
-			showFiles = !showFiles;
-			if (showFiles) {
-				loadFiles();
-			}
+		if (e.key == 'Enter' || e.key == ' ') {
+			toggleFiles();
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


gitbutler-ui/src/lib/components/CommitCard.svelte
```diff
@@ -51,6 +51,15 @@
...
 
+	function onKeyup(e: KeyboardEvent) {
+		if (e.key === 'Enter' || e.key === ' ') {
+			showFiles = !showFiles;
+			if (showFiles) {
+				loadFiles();
+			}
+		}
+	}
+
```
Caleb-T-Owens: I was wondering if it would be neater if we renamed `onClick` to `toggleFiles` (or something to that effect) so we can then change `onKeyup` to

```
	function onKeyup(e: KeyboardEvent) {
		if (e.key === 'Enter' || e.key === ' ') {
			toggleFiles();
		}
	}
```

Caleb-T-Owens: Could we also use `==` instead of `===` to match the standard in the codebase?

PavelLaptev: Good idea @Caleb-T-Owens thanks! And yeah, there should be `==` instead of `===`

PavelLaptev: telegram-cloud-photo-size-2-5463392310151730834-y

Caleb-T-Owens (approved):
Thanks for the refactor. Just tried it out and behaves correctly 👍

PavelLaptev:
Cool, I'll merge this then 🙂

PR #3340, opened 2024-03-26
Ensure project settings button goes to correct project

### LINKED ISSUE #3184: "Project" -> "Project Settings" switches active project
## context
When opening GitButler and selecting a different project (than it opened with) and then you click on "Project" -> "Project Settings" it selects the project it was opening when it was started. 

_The project settings link in the bottom bar is working._

## reproduce
[gitbutlerBug.webm](https://github.com/gitbutlerapp/gitbutler/assets/7146844/36209428-582f-481e-9a7d-ea69e95949f9)

Caleb-T-Owens:
* Closes https://github.com/gitbutlerapp/gitbutler/issues/3184

The primary issue was around the first call being inside `onMount()` which doesn't get re-called when the projectId gets updated. There was a secondary issue where the listener wasn't correctly getting unsubscribed.

## Unsubscribe changes

The first change I made was to allow `unsubscribe()` to take in `undefined`, improving ergonomics when dealing with a potentially undefined subscription. I also modified it so we can await all the unsubscription of all the subscriptions passed into it.

There should be no change of behaviour to existing uses of this method

## ProjectSettingsMenuAction

This component uses its lifecycle hooks to manage the events of the menu bar button


gitbutler-ui/src/lib/menuBarController.ts
```diff
@@ -0,0 +1,43 @@
+import { invoke, listen as listenIpc } from '$lib/backend/ipc';
+import { unsubscribe } from '$lib/utils/unsubscribe';
+import { goto } from '$app/navigation';
+
+export class MenuBarController {
```
mtsgrd: The concepts of controller and service are in my head a bit interchangeable. Do you have any thoughts on what we should prefer, and how to decide if both?

mtsgrd:
Great work!

PR #3360, opened 2024-03-27
separate integration tests for 'app' crate

Byron:
This PR is a massive undertaking as it takes all tests out of the `src/` folder and puts it into the `tests/`. This means that the crate first becomes a library, and the app initialization as a binary now lives in `main.rs` without anything else.

This also means we are only looking at 21k lines of production code, and 12k lines of integration tests, which makes working with the codebase a whole lot easier at least to me, enabling further improvements and maybe a proper separation of tauri-backend code with all the rest (probably).

### Tasks

* [x] try to libify crate, by having `lib.rs` at least, step by step.
* [x] separate tests
* [x] figure out if the App is still working (which it should, tests work too and nothing else should have changed)
* [x] fix TODOs

### Development Notes

* I learned the hard way that the app is actually a binary and thus integration tests don't work. This explains why tests are the way they are (and why this pattern was duplicated into the utility crates which are libraries). In order to get there, first there would have to be a `lib` part of the crate which is used by the `binary` part of it. Then the `lib` portion can be integration tested like everything else, but I will see how tangled it really is. My expectation is that `tauri` *won't* be a problem here, but maybe I am wrong due to `build.rs`.

### Notes for the Reviewer

**Reviewing this probably won't work 'the usual' way** - the change is simply too big and it can't reasonable be split up. My recommendation is to check that all tests are still there, compared to `master`. Validating that nothing related to functionality was changed in `src/` should also be possible to do. However, assuring the tests weren't changed beyond their structure is probably not feasible, and a merge would thus have to be based on trust. And I am saying this because if the merge takes weeks, maintaining this PR might become a nightmare as new tests should ideally be added to `tests/` from now on.

* This crate is different from the utility 'git' and 'changeset' crates as it contained `#[cfg(test)]` markings in production types, to provide special APIs just when under tests. These are not testable by integration tests, but probably they shouldn't have to exist to begin with.
* Some import paths in the tests might not be consistent yet as some of these were touched up by the IDE. I didn't worry about it.
* `use` statements are probably a little messy and probably inconsistent. `cargo +nightly fmt` with the right settings fixes them all up in one go.
* I made a couple of access (read-only) methods public to be able to use them from the integration tests - this is usually a good indication that these are useful/needed.
* Some constructors that are only useful for tauri-free initialization are also available now.
* Because it was easiest (and I think it's not a problem) I made `watcher::handlers` accessible for purpose of moving out their tests. This could also be private/unit-tests, of course, but I'd hope that this could be made testable through the `Watcher` itself (but I don't know) yet.


krlvi:
Tested out and appears to function as before. Thanks for the refactor

PR #3367, opened 2024-03-28
Follow-up to 3360

Byron:
Created on the fly while perusing the test-suite, and it turned out that #3360 wasn't complete at all.

Now I am counting 19.5k lines in production code, and 13k in tests.

This PR roughly does the following:

* avoid unnecessary `test_` prefix
* move more (and previously missed) test modules to integration level
* make `dedup` crate-private to have a reason to keep the tests where they are, same with similar functions

Please note that some unit-level tests remain in the `app` crate nonetheless, as it didn't seem worth it to move them around as the overhead in terms of perceived complexity is quite high with integration tests due to the extra `folder/mod.rs`.


krlvi (approved):
looks good

PR #3368, opened 2024-03-28
Use svelte component context in even more places

mtsgrd:


gitbutler-ui/src/lib/utils/context.ts
```diff
@@ -9,10 +20,43 @@ export function getContextByClass<T extends new (...args: any) => InstanceType<T
...
+// ): Readable<InstanceType<T>> {
+// 	const instance = svelteGetContext<Readable<InstanceType<T>> | undefined>(key);
+// 	if (!instance) throw new Error(`no instance of \`Readable<${key.name}>\` in context`);
+// 	return instance;
+// }
+
+export function getContextStore<
+	T extends new (...args: any) => InstanceType<T>,
+	S extends Readable<InstanceType<T>> = Readable<InstanceType<T>>
+>(key: T): S {
```
Caleb-T-Owens: If we're going for the name `getContextStore` over `getContextStoreByClass`, it feels appropriate that we should also rename `getContextByClass` to just `getContext`.

gitbutler-ui/src/lib/vbranches/tooltip.ts
```diff
@@ -0,0 +1,23 @@
+import type { Commit } from './types';
+
+export function getLockText(commitId: string[], commits: Commit[]): string;
+export function getLockText(commitId: string, commits: Commit[]): string;
+export function getLockText(commitId: any, commits: Commit[]): string {
+	if (!commitId || commits === undefined) return 'Depends on a committed change';
```
Caleb-T-Owens: Is the typescript overloading syntax something we want to go for? With it not actually having any polymorphism to back it up, found it to be a more cumbersome alternative to using a union type:
```ts
export function getLockText(commitId: string[] | string | any, commits: Commit[]): string {
```


Style aside, I am wondering why we need the `any` type and whether `undefined` would suffice.

gitbutler-ui/src/lib/components/BranchLane.svelte
```diff
@@ -17,11 +27,37 @@
...
-	export let branchCount = 1;
 
-	$: selectedOwnership = writable(Ownership.fromBranch(branch));
+	const selectedOwnership = Ownership.fromBranch(branch);
+	$: setContextStore(Ownership, selectedOwnership);
 	$: selected = setSelected($selectedFiles, branch);
 
+	$: setContextStore(Branch, branch);
+	$: setContextStore(
+		LOCAL_COMMITS,
```
Caleb-T-Owens: I'm not loving the fact that we're still dealing with type-unsafe getters and setters.

Part of me is thinking about a public interface like:

```ts
// context.ts
function buildContextStore<T>() {
  const identifier = Symbol();
  const get = () => getContextStore<T>(identifier),
  const set = (value: T) => setContextStore(identifier, value)
  return [get, set]
}

// branchContexts.ts
const [getLocalCommitContextStore, setLocalCommitContextStore] = buildContextStore<LocalCommit[]>

export default { getLocalCommitContextStore, setLocalCommitContextStore }

// BranchLane.svelte
$: setLocalCommitContextStore(branch.commits.filter((c) => c.status == 'local'))

// Where its used
const localCommits = getLocalCommitContextStore()
```
would be nice.


I'm also not convinced about the symbols living in the types.ts file. I'm willing to be swayed on this, but it feels like it would make sense to have a file that is correlated with the component that creates the contexts which contains the relevant symbols

gitbutler-ui/src/lib/vbranches/tooltip.ts
mtsgrd: Thanks!

gitbutler-ui/src/lib/utils/context.ts
mtsgrd: Done!

gitbutler-ui/src/lib/components/BranchLane.svelte
mtsgrd: That looks pretty great to me. Would you want to try and add something like that to this pr?

Caleb-T-Owens: Sure, I'm happy to have a go tomorrow 👍

mtsgrd:
Thanks for helping with reviews @Caleb-T-Owens!

Caleb-T-Owens:
No problem! Glad I could help

PR #3369, opened 2024-03-28
auto-remove tempdirs after testrun

Byron:
Currently each test creates multiple temp-directories to place various disjoint portions of GitButler in. As only their path is used, these accumulate as they are never cleaned up, or in other words, `tempfile::TempDir` is only used to create a temporary directory name which then is racily re-created.

After having worked with the testsuite quite intensively, I noticed that 680k files in thousands of tempdirectories had accumulated already, each with their own set of files. After deleting them and running the tests just a few times, 38k files were there once again, surprisingly quickly.

Ideally, testfiles and directories are removed after usage so developers don't have to manually clean their machines (just judging from me, I'd probably be obsessive about it).

### Tasks

* [x] make all tests work again 
* [x] validate that directories don't accumulate
* [x] add a way to keep temporary directories on demand
* [x] possibly use this 'way' on CI to avoid unnecessary cleanup for the extra-bit of performance

Overall, I'd expect performance locally to not be worse, and performance on CI to be better (at least in theory) as temporary directories are only created once (instead of created, deleted, recreated).

### Notes for the Reviewer

* The first commit is another forgotten TODO and its fix as left-over from #3360 .
* I am not 100% sure that my CI environment adjustment has the desired effect on the basis that I might have gotten the yaml schema wrong. If that would be the case, CI would be slower than before.
* The `GITBUTLER_TESTS_NO_CLEANUP` name was made-up and naming is hard. It should be easy to change as it's mentioned in only two places.

### Before: files accumulate

```
❯ find /var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/ -type f 2>/dev/null | wc -l
       1
```

After a full app test run with...

```
❯ time cargo test -p gitbutler-app
[.. no build happens, only tests run] 
test result: ok. 309 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.73s

cargo test -p gitbutler-app  3.78s user 30.58s system 505% cpu 6.792 tota
```

…and counting new files…

```
❯ find /var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/ -type f 2>/dev/null | wc -l
   15201
```

…we see that 15k files accumulate per full run. This would mean to get 680k files  previously, I had to run the suite just 45 times, something that happens quickly since each run only takes 6s.

### After: auto-cleanup

```
❯ find /var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/ -type f 2>/dev/null | wc -l
       1
```

Now run the app-tests:

```
❯ time cargo test -p gitbutler-app
[.. no build happens, only tests run] 
test result: ok. 310 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 8.50s

cargo test -p gitbutler-app  4.04s user 48.78s system 567% cpu 9.309 total
```

```
❯ find /var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/ -type f 2>/dev/null | wc -l
      10
```

Some files with `.lock` extension still stay, and I couldn't see where these would be created easily, and didn't worry about these quite as much.
**It's notable that tests now run much more slowly, at 67% of the original speed.** Parallelism is reduced as well, probably due to increased filesystem load.

### After: auto-cleanup with GITBUTLER_TESTS_NO_CLEANUP

```
❯ find /var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/ -type f 2>/dev/null | wc -l
       1
```

Now when running the tests with the environment variable set:

```
❯ time GITBUTLER_TESTS_NO_CLEANUP=1 cargo test -p gitbutler-app
[..]
test result: ok. 310 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.69s

GITBUTLER_TESTS_NO_CLEANUP=1 cargo test -p gitbutler-app  3.78s user 29.90s system 557% cpu 6.043 total
```

And we are back to where we were in terms of performance.

```
❯ find /var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/ -type f 2>/dev/null | wc -l
   14412
```

It's notable that it now leaks about 700 files less, which is probably due to some self-contained tests that now also don't leak their files, but don't respect the environment variable either. It's good that these don't affect the runtime.

### DIFF
--- a/.github/actions/check-crate/action.yaml
+++ b/.github/actions/check-crate/action.yaml
@@ -25,6 +25,8 @@ runs:
 
     - run: cargo test --locked -p ${{ inputs.crate }} --all-targets $(cat /tmp/features)
       if: inputs.action == 'test'
+      env:
+        GITBUTLER_TESTS_NO_CLEANUP: "1"
       shell: bash
 
     - run: cargo clippy -p ${{ inputs.crate }} --all-targets $(cat /tmp/features) -- -D warnings
--- a/gitbutler-app/src/gb_repository/repository.rs
+++ b/gitbutler-app/src/gb_repository/repository.rs
@@ -520,6 +520,10 @@ impl Repository {
         self.root().join("session")
     }
 
+    pub fn git_repository_path(&self) -> &std::path::Path {
+        self.git_repository.path()
+    }
+
     pub fn session_wd_path(&self) -> std::path::PathBuf {
         self.session_path().join("wd")
     }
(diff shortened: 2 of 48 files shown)


krlvi (approved):
CI seems to take about the same duration as before (at least for this run) and nice to not pollute with temp dirs from running the tests locally 👍

PR #3370, opened 2024-03-29
Remove invalid attribute from svg file

tobyjwebb:
This SVG file had a `height="auto"` attribute, which was making the following error display in the dev console:

image

After removing it, the error went away, and no appreciable change to the image was observed.


krlvi (approved):
thanks for spotting this!

PR #3371, opened 2024-03-29
Open project in VSCode in new window

tobyjwebb:
I was running GitButler from a VSCode terminal, and clicked on a "Open in VSCode" button for another project, and the project that opened was opened in the same window, causing GitButler to stop.

IMO it would be a better behaviour to always open projects in new windows (note that even with this change, if the project is already opened in a VSCode window, it will not open a new one).

(Details about opening in new window [here](https://github.com/microsoft/vscode/issues/141548#issuecomment-1102200617))

### DIFF
--- a/gitbutler-ui/src/lib/components/Board.svelte
+++ b/gitbutler-ui/src/lib/components/Board.svelte
@@ -142,8 +142,8 @@
 										class="empty-board__suggestions__link"
 										role="button"
 										tabindex="0"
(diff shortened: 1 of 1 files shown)


krlvi (approved):
I didnt realise this was possible! this has been annoying me for a while, thanks for fixing

tobyjwebb:
> I didnt realise this was possible! this has been annoying me for a while, thanks for fixing

You're welcome! Glad it helps

PR #3373, opened 2024-03-29
extract general library from `app` crate

Byron:
This allows the tauri-specific parts to remain in the `app` crate, which will help to eventually release a `cli` crate as well.
Much of the library portion is now in the top-level, with only minor adaptations to make it integratable into `app` without needing tauri.

### Tasks

* [x] library crate whose tests run (291 of 310 in app)
* [x] minimize dependencies in lib
* [x] integrate library into `app`
* [x] minimize dependencies in app
* [x] assure CI can handle this (there are path-based filters)

### Notes for the Reviewer

* When adjusting CI, it felt like the top-level `gitbutler` crate *maybe* wants to be called `gitbutler-core` (or similar). Then it *might* be clearer what it truly is. An alternative argument also can be made that the library implementing GitButler can reasonably identify itself as `gitbutler`, like is the case now. 
* I applied minimal changes to get it to work, it's merely a first of many more steps to turn the library that looks like a typical API crate where borrowcheck will do (more) work.
* As follow-up, one *could* move the tauri-specific building blocks into a `tauri` crate. From there it should be straightforward to create a CLI crate (or a binary in `app`) to interact with `tauri`. However, if perfect separation of dependencies isn't too important, one could also build a CLI library from the `app` crate. I think it's cleaner to split though, so one would have `tauri`, `app` and `cli`.
* Once again, there seemingly is many changes, but the only the only actual change was `src/askpass.rs` and `src/projects/controller.rs`, to make them tauri-independent.
* The crate doesn't actually compile without the `error-context` feature, but that's the same for the `app` crate. While it's a bit uncertain how the error-handling story should look like I decided not to fix it (or add CI tests to catch that). My thinking is that when errors are sorted, this will probably just always be available.
* I left the custom error implementation in the `app` crate to be able to fix it independently of the library, should that be needed.
* In `virtual_branches.rs` the error handling definitely got uglier due to additional `From` implementations, but I think it's temporary as error handling definitely needs to be addressed. There is now some duplicate manual conversions as well.


Caleb-T-Owens:
Personally, I'm *very* excited to see this and the possibilities of a CLI tool getting opened up

krlvi (approved):
nice, this is a very juicy refactor! i'm very excited to see the tauri parts separated out! tested it working locally as well

PR #3377, opened 2024-03-30
[draft] Refactoring services

Caleb-T-Owens:
This is a work in progress PR where I'm having a go at cleaning up some of the services, and hopefully will be able all state from services.

Here are some avenues of exploration:
- [ ] Removing RX.js from the public APIs exposed by services. (RX.js should only be used internally)
- [ ] Fix services that are dependant on the project
  - Services should have no state and have no need to reactively depend on the projectId.
- [ ] GitHubService
  - Demangle project specific code (get rid of remoteUrl$)
  - Make generic & replaceable (to allow the implementation GitLab/BitBucket clients)
- [ ] BaseBranchService, VBranchService, BranchController, BranchService, RemoteBranchService.
  - These are all dependant on the projectId, and some feel they should really be injected into some Project class (exploring options)

## Reviewing

I'm still getting an idea of the direction I want to take these services, so anything here may change drastically. I just wanted to open a Draft PR so it was visible.

### DIFF
--- a/gitbutler-ui/src/lib/backend/projects.ts
+++ b/gitbutler-ui/src/lib/backend/projects.ts
@@ -1,11 +1,9 @@
 import { invoke } from '$lib/backend/ipc';
 import { persisted } from '$lib/persisted/persisted';
-import { observableToStore } from '$lib/rxjs/store';
 import * as toasts from '$lib/utils/toasts';
 import { open } from '@tauri-apps/api/dialog';
 import { plainToInstance } from 'class-transformer';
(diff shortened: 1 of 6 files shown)


mtsgrd:
Sorry about the conflicts in `projects.ts`, I was unaware you were working on this!

Caleb-T-Owens:
No worries @mtsgrd!

I don't want to block any other refactoring attempts that may be touching these files, so I'm more than happy to drop out various changes if other ones come along.

Caleb-T-Owens:
I'm going to close this for now as its quite far behind and I don't have the time at the minute to do the renovations that I want to do. I am keen to come back to this though.

PR #3378, opened 2024-03-30
establish `crates/` directory

Byron:
It is similar to `nu-shell`, which has a lot of crates by now, which is something we probably see here as well.
The benefit of having many small crates is to be precise when handling their dependencies, and faster compile times for test-driven development.

### Tasks

* [x] move `app` to `crates/`
* [x] setup `rustfmt.toml` and configure imports to unify them (but only if this doesn't mess with the currently unmerged PRs)
* [x] validate CI and dev-setup still works

### Notes for the Reviewer

* I also checked the `publish` CI configuration and it looked like it won't be affected by this change. But since I didn't actually run it, it's nothing more than a guess.
* I shoe-horned a way to, for now manually, auto-organize use statements which I had left messier than I found them. I also executed it once, but only for the crates that aren't touched by the outstanding PRs. If in doubt, this could be its own PR, or be postponed/skipped entirely.


rustfmt-nightly.toml
```diff
@@ -0,0 +1,4 @@
+group_imports = "StdExternalCrate"
+imports_granularity = "Crate"
+# This is useful to adapt to larger screens of today, for better use of horizontal screen space.
+# max_width = 120
```
Byron, line 4: Changing the line-length is a big change and would need `cargo fmt` to pick up this file, which it would if it was named `rustfmt.toml`. However, if there is any doubt about this, let's not do it.

rustfmt-nightly.toml
```diff
@@ -0,0 +1,4 @@
+group_imports = "StdExternalCrate"
+imports_granularity = "Crate"
```
Byron, line 2: These two statements configure the style with which use-statements are organised, and there are others well: https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=imports . I copied this configuration from `gitoxide`, but that doesn't mean it's perfect. Maybe by now there are more and better options.

Maybe this should rather be a separate PR as well, but I thought during easter it's the least annoying.

krlvi (approved):
looks good. merging this, and will test the publish job from master

PR #3380, opened 2024-03-30
Make more type conversion implicit in `commands.rs`

mtsgrd:


krlvi:
Using anyhow we will do even less error type conversion. I wonder, how does the argument error look like from Tauri? As it what does the frontend get when you provide wrong parameter type @mtsgrd ?

krlvi (changes requested):
moving some directories, pls hold off 30 mins or so :)

krlvi (approved):
okay done, sorry for the merge conflict, it's just a file move tho

PR #3385, opened 2024-03-31
consistent error-handling

Byron:
This PR is an attempt to remove the legacy error handling in favor of a unified and simple way of handling errors.

> [!NOTE]
> This could be a big PR and the goal here is to get as much of it done as possible, while possibly considering intermediate merges. When started, it must be completed or error handling might look even messier than before.

### Tasks

* [x] provide extension trait in `core` to core to allow attaching a `Code` to an error as `anyhow` context.
    - This is still used by the frontend, so seems significant, even though it's only available in the legacy error
    - The core is expected to provide the `Code` as form of context so the frontend knows what to do
* [x] add `ErrorWithContext` implementations to `thiserror` instances that need it as they are currently using `legacy::Error` conversions.
     - This should probably be replaced by `anyhow`, and instead this is only used where it truly is required/useful, but even then there are alternatives (for instance, when doing `git2` error handling)
* [x] in `app`, provide an Error type that serializes in the desired format. Can be `thiserror` actually.
    - Such an error type already exists, but it's currently coming from `core` mostly, but should be coming from the `app` crate as it's specific to tauri.
* [x] Resolve thiserrors that wrap a single anyhow error
* [x] remove the legacy error type and make the new one the default.
* [x] Check how errors look like in `tracing`-based logs - is it possible to see where they are coming from?
    - How about backtraces, and should these be added to some `thiserror` variants?
* [x] Write a `how-to` in the error docs, with doc-test examples.
    - There is probably now more than two things to know about this, but one has to be aware or context for users won't show up.
* ~~add `backtrace` to `#[from]` conversions on the lower level for more debug info~~
    - This should probably be done on demand

### Test Results

I did some manual testing to assure lower-level context bubbles up, that the error chain is kept, and that backtraces work at least when `anyhow` is used (`thiserror` can also do it, but that isn't automatic).

Screenshot 2024-04-01 at 16 32 47

The above is after using`context(error::Context::new(Code, "message"))` to bubble up custom messages, even though the Code is only relevant in some portions of the UI.

Screenshot 2024-04-01 at 16 33 04

This is how it looks in the logs, without backtrace. The context message seems duplicated, and I don't really know why that is. I wouldn't fret it now, either.

Screenshot 2024-04-01 at 16 34 06

And if the application is launched with `RUST_BACKTRACE=1`, then any `anyhow` error automatically collects a full backtrace. `thiserror` won't do that, but [can be told to](https://docs.rs/thiserror/latest/thiserror/).

### Things to Remember

* The `Code` attached to the root cause is what we want to surface, and there is a lot of code which assures just that when `thiserror` is used and conversions are made to get the legacy Error with Code.
* In some places, the code is attached based on the analysis of a complex Code (like `git2::Error`), which is currently neatly integrated into the `core`. Ideally, as much of that code as possible goes directly into the 
* `#[error(transparent)]` effectively flattens the error by one level, removing a link from the error chain. Unless there is a backtrace at the start of the chain, it might be hard to see where the error is coming from.

### Notes for the Reviewer

* When reading some of the error messages destined for the frontend, I thought they are probably not adding much, but are better than a bare error code. However, this is something that can be improved once it's clear which errors happen a lot.
* It does seem as if errors will be logged multiple times, once in a manual trace, and another time with instruments.
    - I removed the error tracing that I though was now superfluous - the reason these happened is to avoid loosing error detail, which now is naturally kept.
* I refrained from changing the granularity of errors, in that unless an error has a `Code` or `Context` attached, they will show up as "Something went wrong" only. It's easily possible to make more errors visible, but I think it's more of a product decision of what these should be.
* The new system introduced here is made under the assumption that the lower levels (i.e. `core`) should be able to control which errors should be visible in the UI. If that's not needed, so that the `app`-layer is controlling it, then this could be simplified more. However, I think it's good to allow lower layers to control it 'at the source' so the app-layer doesn't have to duplicate additions of context. Maybe, with the `ErrorWithContext` extension trait, it would be possible to control context by type just in the `app` layer - the impls could just move 'up' from `core` to `app`. There is definitely options.
* If separate crates want to use the new context-facilities, we'd have to place the `error` module into its own crate. I refrained from doing it now as there was no need.

### Out of Scope

* a custom `log` or `tracing` formatter that displays backtraces more succinctly
    - This will work, but is neatly disjoint from this PR as well as it's about displaying data that is readily available.

<details><summary>Research</summary>

### Requirements

* serializable for UI
* backtrace collection
* ability to format it nicely for logs, but also the console, without producing noise (particularly for backtraces)
* Easy to use without overhead, just like `anyhow`
    - `thiserror` can optionally be used, like is the case for lower-level crates `git` and `changeset`

### Approach

* use `anyhow` everywhere, even in lower-level crates that currently use `thiserror`.
* higher-level `cli` and `app` crates should use `anyhow`

#### How to backtraces in `thiserror` interact with `anyhow`? 

In `thiserror`, it's stored via `std::error::Error::provide()`, providing a `std::backtrace::Backtrace` reference, if the `thiserror` variant stored it in a `backtrace` field.

Anyhow will use the backtrace stored in the underlying Error, or store its own, so interaction between both should be sane.

Note that backtraces are only available [if certain environment variables are set](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.backtrace).

#### How does `tauri` interact with errors?

It's part of its `generate_handler!()` macro to generalize a function call. It seems to come with special support for anyhow, using the alternate anyhow display implementation (which includes causes, but is unstructured):

```Rust
  #[inline(always)]
  pub fn from_anyhow(error: anyhow::Error) -> Self {
    Self(JsonValue::String(format!("{error:#}")))
  }
```

This means that either this displays custom context, like `Code`, or one has to wrap it into a custom type which implements `serde::Serialize`

```Rust
impl<T: Serialize> From<T> for InvokeError {
  #[inline]
  fn from(value: T) -> Self {
    serde_json::to_value(value)
      .map(Self)
      .unwrap_or_else(Self::from_serde_json)
  }
}
```

#### How does `log` display errors currently?

The `log` crate seems to be used to display UI errors using the backend and the `log` framework, which [uses a plugin](https://github.com/gitbutlerapp/gitbutler/blob/72291ce4cb039f8675ec143a0ad672936a183ce1/gitbutler-app/src/main.rs#L46) to receive these 'log' events and writes them to a file. Thus, on Rust tauri listens to `log` events and writes them, but their format is certainly defined by the frontend and it's not really related to backend error handling.

#### How does `tracing` display errors currently?

Tracing picks them up using `instrument` decorators, which probably puts them into a field. Sometimes errors are logged by hand, typically using its `?error` debug implementation. Note that the Debug implementation prints everything, including stacktraces, if the error is `anyhow`, similar to what happens when such error is returned from `main()`.

#### What does the UI expect of errors?

[Here is the code](https://github.com/gitbutlerapp/gitbutler/blob/9d4a709012fa88a3f568eaa190a9f50a1f34556c/gitbutler-ui/src/lib/backend/ipc.ts#L72-L87) that handles calls to tauri and deals with errors, which are supposed to look like this right now:

https://github.com/gitbutlerapp/gitbutler/blob/9d4a709012fa88a3f568eaa190a9f50a1f34556c/gitbutler-ui/src/lib/backend/ipc.ts#L20-L31

However, that should only be relevant for errors provided returned by tauri-commands, which is the only way the frontend interacts with them.

This raises another question: 

#### How important is the error code in the application design?

It's used in two places, checking for value `errors.git.authentication` and `errors.projects.git.auth`, but there is more that the GUI could differentiate. The `UserError` is used on the Rust side to nominate an error that should be presentable in the frontend. There, it seems it's only used on occasion.

The GUI also keeps the `cause`, but doesn't seem to use it at all. It's also not really the cause, but the original error, which makes me think that it's not meant to get an error chain.

#### How can metadata (like `Code`) be attached to `anyhow` errors?

Anyhow provides the ability to attach `context()`, which can be anything as long as it's `Display`. The [downcasting manual](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#effect-on-downcasting) makes clear that this can be any type, so any context is supported as well, which then can be retrieved at a later time.

This means `anyhow` can be self-contained, and could have an extension trait which allows to attach and retrieve application-specific context easily.

This also means that in order for Tauri to serialize errors correctly, all one has to do is to make `anyhow::Error` convertable into a utility type that can serialize into errors with the fields the frontend needs.

#### Can errors that are used by `tracing` be formatted more consisely?

I think so, as `anyhow` and `thiserror` both provide complete access to the underlying data, which can be formatted at will once a corresponding formatting implementation was created.

#### Sentry-Errors

These are implemented by the `gb::Error` and it appears to be entirely unused. 

</details>


krlvi:
Just want to add that the current uses of `Code` in the UI should not necessarily be a constraint. We may still want to propagate an error code to the UI if we feel this is a useful pattern, or remove it.

Here's my perceived shortcomings with the current design:
1. Bad user-facing error messages. Because creation of new error types or error enum variants can be tedious, often we'd resort to using a catch-all variant (`Other`), and in those cases the error that gets surfaced to the the user often looks like "something went wrong" (or effectively that)
2. Poor debug-ability. When an undesirable event occurs (i.e. something breaks in the app), we would look at an error in the logs to understand what and why. If the error enum variant is only used in one place, then we are lucky, but if it is used in multiple places, we would have to guess where in the code the error originated.
3. Boilerplate around error management. Converting between error types (eg. git2 / app errors) can be quite verbose which makes the application code more difficult to follow than necessary.

Byron:
Thanks so much for chiming in, this information truly helps! Here is my takeaways:

1. I find it beneficial to be able to drive the GUI with error codes, which is definitely happening in a few places already. Even though it can be removed, for now I think it needs to be supported to not break anything. However, what `Code` really is, is the notion of custom context, which I think the system should support for future changes and/or extension.
2. The new system won't 'flatten' the error anymore just to retain the code, and thus keep the whole error chain available in the logs. The UI will still only see the frontmost error though, but that of course can be improved as well (later).
3. I will work on making unnecessary boilerplate go away, even though I would like to keep it if it's helpful. Like when matching on `git2` error codes to turn them into own error codes. However, maybe this can be streamlined, and I will look into this as well.

Point 2 is the most exciting to me, as it's a definitive improvement in debuggability.

krlvi:
And additional comment so that it's on the PR as well - we can remove / not consider sentry from the rust code. We have a better approach for bug reporting (gh issues / discord).

krlvi (approved):
i tried this pr locally and works nicely. merging this now because the only way to see how it feels is to try it out for real! thanks

PR #3405, opened 2024-04-02
Restructure crates

Qix-:


Qix-:
Please do not merge - @mtsgrd has asked that we wait until he's landed a few changes.

.github/workflows/push.yaml
```diff
@@ -194,12 +193,36 @@ jobs:
           features: ${{ toJson(matrix.features) }}
           action: ${{ matrix.action }}
 
+  check-gitbutler-core:
```
krlvi, line 174: I think you may also wanna delete `check-gitbutler` on line `104` since its for the same target

Qix-: Great catch, thanks! Didn't realize it was for core.

crates/gitbutler-tauri/src/tests/mod.rs
```diff
@@ -0,0 +1,3 @@
+#[path = "../../../gitbutler-core/tests/shared/mod.rs"]
```
Byron: I think the culprit is this hack that I introduced to allow code-sharing, without having to put the test code into its own crate which would have been the right thing to do.

To resolve the warnings, the shared test-setup code would have to become its own crate, `gitbutler-testtools` maybe (but I never come up with good names).

PS: I did try to put a hack on the hack by using `pub mod shared;` instead, but Rust is smart enough to know that nothing is truly exported from a binary.

Byron: Since this is a large PR, one could consider to follow it up with the proper fix as well, and just silence the warnings for now. As I introduced the issue, I'd offer to wrangle it down as well :D.

Qix-: The tests under `gitbutler-core/` still have `pub mod shared` unless I'm looking in the wrong spot. Would this not be fixed just by having these tests as part of the crate like they were before?

Qix-: > Since this is a large PR, one could consider to follow it up with the proper fix as well, and just silence the warnings for now. 

Sure, I can do that :)

Byron: > The tests under `gitbutler-core/` still have `pub mod shared` unless I'm looking in the wrong spot. Would this not be fixed just by having these tests as part of the crate like they were before?

Do you mean to turn them back into unit-tests and put them under `src/`. I think figuring out what parts need sharing, and making them shareable via a separate crate is the main difficulty that thus far I skipped over.

> Sure, I can do that :)

Ok, please feel free to add a `TODO(ST)` there near the suppression as well.

Qix-:
@Byron w.r.t https://github.com/gitbutlerapp/gitbutler/pull/3344 and https://github.com/gitbutlerapp/gitbutler/pull/3347 do we want to merge this one first and then continue work after the restructure to make sure no new warnings are added? Might make things easier.

Byron:
I have lost track a little of what's best, and trust you if you think it should be easier. Right know I am a bit lost, there is conflicts here and it seemed the changes here would conflict with #3344 and #3347 as well. 

But like you say, it's probably easier to merge this one first and redo the test-separation for the `git` and `changeset` crates, as the respective PRs will probably too hard to salvage, while they are easy to redo.

krlvi:
May I suggest that we merge this contribution first https://github.com/gitbutlerapp/gitbutler/pull/3386 to make life easier for community contributors? It seems to me like a valuable fix to something we overlooked, wdyt @Qix- ?

Qix-:
@krlvi I've posted my thoughts there and need to process the newest comment. I don't think it's related to this PR at all and should be trivially rebased (this is mostly just a rename PR). I want to merge this one to unblock @Byron on the test splitting front.

Qix-:
Rebasing this now and will push + merge ASAP.

PR #3411, opened 2024-04-02
Add difflenth to settings

Caleb-T-Owens:
## Implementing Diff Length settings

When used in the API, I've capped it at 5000 characters so we don't end up inundating the servers. @PavelLaptev I wonder if we need to add a line to the UI component letting the user know this.

## Adding getters for GitConfig values

We've got several defaults specified for a handful of values, and these end up getting duplicated each time they are read which is not particularly ergonomic. I've added in some getter methods and changed over to using them in the AIService and AISettings.

## Refactoring logic

I've added in a helper method to help reduce some duplicate logic and added in some constants which I think make the code more self explanitory

### DIFF
--- a/gitbutler-ui/src/lib/backend/aiService.ts
+++ b/gitbutler-ui/src/lib/backend/aiService.ts
@@ -105,56 +105,80 @@ export class AIService {
 		private cloud: CloudClient
 	) {}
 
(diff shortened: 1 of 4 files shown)


PavelLaptev:
@Caleb-T-Owens thank you! Looks good, I just made a few changes

PavelLaptev (approved):

PR #3432, opened 2024-04-04
wiggle lock files if dragging

PavelLaptev:
https://github.com/gitbutlerapp/gitbutler/assets/18498712/ab1e1eb3-9ae4-493e-9444-ff92e2422ca1

### DIFF
--- a/app/src/lib/dragging/draggable.ts
+++ b/app/src/lib/dragging/draggable.ts
@@ -82,6 +82,7 @@ export function draggable(node: HTMLElement, opts: DraggableConfig) {
 
 	function handleDragStart(e: DragEvent) {
 		let elt: HTMLElement | null = dragHandle;
+
 		while (elt) {
 			if (elt.dataset.noDrag !== undefined) {
 				e.stopPropagation();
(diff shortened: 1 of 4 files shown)


app/src/lib/dragging/draggable.ts
```diff
@@ -91,14 +92,26 @@ export function draggable(node: HTMLElement, opts: DraggableConfig) {
 			elt = elt.parentElement;
 		}
 
-		// If the draggable specifies a selector then we check if we're dragging selected
-		// elements, falling back to the single node executing the drag.
+		// If the draggable specifies a selector then we check
+		// if we're dragging selected elements
 		if (opts.selector) {
-			const parentNode = node.parentNode;
+			const parentNode = node.parentNode?.parentNode; // the parent has a container, so we need to go up one more level
```
mtsgrd: How about this?
```
// Checking for selected siblings in the parent of the parent container likely works
// for most use-cases but it was done here primarily for dragging multiple files.
```

app/src/lib/dragging/draggable.ts
```diff
@@ -91,14 +92,26 @@ export function draggable(node: HTMLElement, opts: DraggableConfig) {
 			elt = elt.parentElement;
 		}
 
-		// If the draggable specifies a selector then we check if we're dragging selected
-		// elements, falling back to the single node executing the drag.
+		// If the draggable specifies a selector then we check
+		// if we're dragging selected elements
```
mtsgrd: Maybe this comment can be on one line instead of two?

mtsgrd (approved):

PR #3437, opened 2024-04-04
Move AI Client related stuff into its own folder

Caleb-T-Owens:
Sorts AI related code into its own folder/package. I would welcome the discussion as to whether this is better or worse than having it all flattened.


mtsgrd:
I can't see that we gain much from having a separate `aiTypes` folder, could just go into `/ai/types.ts`. Wdyt?

Caleb-T-Owens:
> I can't see that we gain much from having a separate `aiTypes` folder, could just go into `/ai/types.ts`. Wdyt?

@mtsgrd It would mean that the `backend` and `ai` folders would be cyclically dependant on each other's types.

Typescript is good at resolving circular types and it would be simpler, so I wouldn't be opposed

mtsgrd:
Right sorry, I'm on the same page again. Hmm, there's something a bit awkard in this setup, and I'm wondering if `cloud.ts` would be better imagined as a `GitButlerApi`? Its puprose could be providing a fetch function that is optionally authenticated, and standardises error handling. If we did that then AiService could depend on HttpClient, and define its own api calls, which in turns means aiTypes.ts could live in the same directory as the related files.

Caleb-T-Owens:
> Right sorry, I'm on the same page again. Hmm, there's something a bit awkard in this setup, and I'm wondering if `cloud.ts` would be better imagined as a `GitButlerApi`? Its puprose could be providing a fetch function that is optionally authenticated, and standardises error handling. If we did that then AiService could depend on HttpClient, and define its own api calls, which in turns means aiTypes.ts could live in the same directory as the related files.

@mtsgrd I'm not thrilled about that from a perspective of code duplication and including potentially dangerous typing in more parts of the application.

It's likely that we will want multiple modules to access the same endpoints which would mean either creating some other shared module (which is somewhat where we are already), or we just accept having multiple `fetch<PotentiallyWrongType>('/endpoint')` calls around the application which will make updating the expected return type harder.

---

We could make use of the fact that Typescript is very good at duck typing. IE: if you have interface A, and interface B. If B is a superset of A, even if it's *not* declared as extending, you can assign an object of type B to type A as if it was.

We could decide to type our APIs using its own set of interfaces that don't depend on the rest of the application. It would mean that there could be slight code duplication in interface definitions if we've designed the API to match our client data structures or vice versa, but on the other hand, it would mean that we're maintaining more type safety around the application which seems like it should be a priority to me.

app/src/lib/backend/cloud.ts
```diff
@@ -60,30 +58,66 @@ async function parseResponseJSON(response: Response) {
...
+	private formatBody(body?: FormData | object) {
+		if (body instanceof FormData) {
+			return body;
+		} else if (body) {
+			return JSON.stringify(body);
+		}
+	}
+
+	// TODO: consider renaming
+	async makeRequest<T>(params: {
```
mtsgrd: I would make this just `request`, and provide convenience functions as necessary e.g. `post`. It makes for nice to read code like `httpClient.post('/some/path', data)`. Wdyt?

app/src/lib/backend/cloud.ts
```diff
@@ -60,30 +58,66 @@ async function parseResponseJSON(response: Response) {
...
-			method: 'POST',
-			headers: {
-				'Content-Type': 'application/json'
-			},
-			body: JSON.stringify({})
+		const token = await this.makeRequest<LoginToken>({
+			path: 'login/token.json',
+			method: RequestMethod.POST,
+			body: {}
 		});
```
mtsgrd: Could we make it more like this?
```ts
const token = await this.post<LoginToken>({path: 'login/token.json'});
```

app/src/lib/backend/cloud.ts
Caleb-T-Owens: I've gone with object arguments because I'm not a fan of having multiple optional arguments in a row

PR #3438, opened 2024-04-04
(Testing linting/CI) testing cyclical dependency linting

Caleb-T-Owens:


Caleb-T-Owens:
Screenshot 2024-04-04 at 19 25 51

In my editor and in CI it doesn't appear to be complaining about cyclic types

Caleb-T-Owens:
Screenshot 2024-04-04 at 19 31 59

In my editor and in CI, cyclic types are not reported, but objects are reported

Caleb-T-Owens:
Screenshot 2024-04-04 at 19 36 07
My editor and CI agree that there are no warnings. Taking a read through the documentation, I don't think that we should expect an error https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/no-cycle.md

Caleb-T-Owens:
Screenshot 2024-04-04 at 19 43 50

reverting the flattening of aiClients folder (while keeping aiClient.ts in the backend folder) doesn't create any warnings in CI or in my editor

Caleb-T-Owens:
Screenshot 2024-04-04 at 19 48 15

Reverted back to its original state, there doesn't appear to be any cyclic import warnings

Caleb-T-Owens:
It seems like `import/no-cycle` does not catch:
- Cyclic type imports
- Cyclic folders/packages
  - I tested having base/b/foo.ts, base/c/bar.ts, and base/c/baz.ts where foo imported bar, and baz imported foo, creating a codewise safe setup, but meant that the two folders were cyclically dependant on each other

mtsgrd:
Thanks for investigating :)

PR #3444, opened 2024-04-05
Code fixes, update `on:close` Function for Files, Button styles update

PavelLaptev:
- **fix: update `on:close` funcion fro files**. The previous method didn't work for committed files.
- **Renamed `FileSelection` to `FileIdSelection`**. In order to make it more specific,since we already have `selectedFiles`


app/src/lib/utils/selectFilesInList.ts
```diff
@@ -1,24 +1,24 @@
...
+	fileIdSelection: FileIdSelection,
 	sortedFiles: AnyFile[],
 	allowMultiple: boolean,
 	commit: AnyCommit | undefined
 ) {
-	let selectedFileIds = get(fileSelection);
+	let selectedFileIds = get(fileIdSelection);
 	e.stopPropagation();
-	const isAlreadySelected = selectedFileIds && fileSelection.has(file.id, commit?.id);
+	const isAlreadySelected = selectedFileIds && fileIdSelection.has(file.id, commit?.id);
```
mtsgrd, line 15: Let's simplify since `selectedFiles` is never undefined, we can remove the boolean check. I likely introduced it, but it's unnecessary.

mtsgrd (approved):

PR #3448, opened 2024-04-06
Remove unused (?) allow clippy statements

tobyjwebb:
:shrug: 

It looks like Clippy isn't being run in pedantic mode in CI.


Caleb-T-Owens:
Would it not make sense to run the CI in pedantic mode?

mtsgrd:
@Qix- should we be running in pedantic mode?

Qix-:
Yes we should. Can we undo this please? These aren't addressed AFAIK.

PR #3463, opened 2024-04-09
Revert "Remove unused (?) allow clippy statements"

Qix-:
Reverts gitbutlerapp/gitbutler#3448

The current codebase has a ton of clippy lints that we should be testing for but cannot because the changes would be widespread. The goal is to enable pedantic warnings but right now there's a lot that needs to be fixed to make them work. The solution to this isn't to revert the clippy lints we have already.

### DIFF
--- a/crates/gitbutler-changeset/src/signature.rs
+++ b/crates/gitbutler-changeset/src/signature.rs
@@ -109,6 +109,7 @@ impl Signature {
             }
         }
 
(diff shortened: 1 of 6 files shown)


tobyjwebb:
I'll have a look into this, but it might take a while, I'm busy for the next few weeks.

mtsgrd:
What's pending here?

Qix-:
Just fell through the cracks :)

Issue #3430, opened 2024-04-04
Corner cases for the PR card


PavelLaptev:
Currently, we could have states where all checks are passed, but we are unable to merge the PR and the merge button is disabled. It could happen because there are some merge conflicts with the `main` branch.
But the PR card doesn't say anything about it. So, there should be a clearer communication about this.

![image](https://github.com/gitbutlerapp/gitbutler/assets/18498712/bd56e35a-7bab-4663-981d-063189fae6d8)

PavelLaptev:
Done here https://github.com/gitbutlerapp/gitbutler/pull/3465

PR #3465, opened 2024-04-09
added status messages + code refactor

PavelLaptev:


PavelLaptev:
related issue https://github.com/gitbutlerapp/gitbutler/issues/3430

app/src/lib/components/PullRequestCard.svelte
```diff
@@ -34,21 +35,27 @@
...
 	async function updateDetailsAndChecks() {
-		if (!isFetchingChecks) fetchChecks();
-		if (!isFetchingDetails) updateDetailedPullRequest($pr$?.targetBranch, false);
+		if (!isFetchingDetails) {
+			await updateDetailedPullRequest($pr$?.targetBranch, true);
+		}
+
+		if (!isFetchingChecks) {
+			await fetchChecks();
+		}
```
mtsgrd: Can we make this one-liners again?

app/src/lib/components/PullRequestCard.svelte
```diff
@@ -115,67 +120,123 @@
...
+		mergeableState: string | undefined,
+		checksStatus: ChecksStatus | null | undefined,
+		isFetchingChecks: boolean
+	):
+		| {
+				icon: keyof typeof iconsJson;
+				style: MessageStyle;
+				text: string;
+		  }
+		| undefined {
```
mtsgrd: How about defining a type for this? Looks like there's repetitions, but one of them has `label` instead of `text`.

mtsgrd (approved):
Ah nice, a big improvement on my code. 💪

app/src/lib/components/PullRequestCard.svelte
```diff
@@ -115,67 +120,123 @@
...
-		if (status.completed) {
-			return status.success ? 'success' : 'error';
-		}
-		return 'warning';
+	function getChecksCount(status: ChecksStatus): string {
+		if (!status) return 'Running checks';
+
+		const skipped = status.skipped || 0;
+		const total = (status.totalCount || 0) - skipped;
+		const quieed = total - (status.queued || 0);
```
mtsgrd: typo

app/src/lib/components/PullRequestCard.svelte
```diff
@@ -34,21 +35,27 @@
...
 	$: if ($branch && $pr$) updateDetailsAndChecks();
 
-	$: checksIcon = getChecksIcon(checksStatus, isFetchingChecks);
-	$: checksColor = getChecksColor(checksStatus);
-	$: checksText = getChecksText(checksStatus);
-	$: statusIcon = getStatusIcon(detailedPr);
-	$: statusColor = getStatusColor(detailedPr);
-	$: statusLabel = getPrStatusLabel(detailedPr);
+	$: checksTagInfo = getChecksTagInfo(checksStatus, isFetchingChecks);
+	$: infoMessageInfo = getInfoMessageInfo(
```
mtsgrd: Maybe rename this just `infoMessage`? That way it gets used like this:

```svelte
{#if infoMessage}
  <... icon={infoMessage.icon} />
{/if}
```
Or even better if we could:
```svelte
{#if infoProps}
  <InfoMessage {...infoProps} />
{/if}
```

PavelLaptev: Thanks! Updated it to `infoMessage` but didn't use spread operator to add props, because the `InfoMessage` using `slot` inside and its color type slight different.

PavelLaptev:
I think I've addressed all comments 🙂 will merge it in the morning if there wouldn't any fixes

app/src/lib/components/Button.svelte
```diff
@@ -34,8 +23,8 @@
 	export let align: 'flex-start' | 'center' | 'flex-end' | 'stretch' | 'baseline' | 'auto' = 'auto';
 	export let isDropdownChild = false;
 	// Style props
-	export let style: ButtonStyle = 'neutral';
-	export let kind: ButtonKind = 'soft';
+	export let style: ColorStyle = 'neutral';
+	export let kind: KindStyle = 'soft';
```
mtsgrd: `KindStyle` sounds like tautology btw, i.e. both words have similar meaning. I would also argue that color and style can't be combined, it should be one or the other. Wdyt?

PavelLaptev: I agreed but don't want to make the name too vague

PavelLaptev:
@mtsgrd I updated component style types but not props, yet they still `style`. I'll rename this prop to `color` next refactor PR

PR #3446, opened 2024-04-06
share test code via separate crate

Byron:
In the process of the restructuring of crates, some test support code shared among multiple crates was shared by 'manufacturing modules', which came with is own challenges.

This PR remedies this by moving shared code into its own crate.

### Tasks

* [x] a new crate with shared test code
* [x] integrate the new crate

### Notes for the Reviewer

* I accidentally pushed the branch to this repository instead of my fork via the `gh` tool (which makes that very easy to do). Maybe it's best to remove write access to this repo as it seems to be a risk, particularly because I don't have any use for write access at this time.

PR #3467, opened 2024-04-09
share test code via separate crate

Byron:
In the process of the restructuring of crates, some test support code shared among multiple crates was shared by 'manufacturing modules', which came with is own challenges.

This PR remedies this by moving shared code into its own crate.

### Tasks

* [x] a new crate with shared test code
* [x] integrate the new crate

### Notes for the Reviewer

* I accidentally pushed the branch to this repository instead of my fork via the `gh` tool (which makes that very easy to do). Maybe it's best to remove write access to this repo as it seems to be a risk, particularly because I don't have any use for write access at this time.
* This PR supersedes #3446 which accidentally created a branch in the origin repository instead of in the fork.


crates/gitbutler-core/Cargo.toml
```diff
@@ -8,6 +8,7 @@ publish = false
 [dev-dependencies]
 once_cell = "1.19"
 pretty_assertions = "1.4"
+gitbutler-testsupport = { path = "../gitbutler-testsupport" }
```
Qix-: I've been putting crate deps in the workspace Cargo.toml and then using `.workspace = true` - is it cool if we follow that convention?

Byron: Thanks for letting me know, I will remember that. It's fixed via 4542c73fdb4ab51c14ca03ae036d869e00e482d5.

Qix-:
Thanks :D

Issue #3447, opened 2024-04-06
Watcher Refactor


Byron:
This issue keeps track of initial research to understand `watcher.rs` better, with the intend to formulate a plan to revise it, and make it more maintainable in the process.

### PRs

* [x] #3471 
* [x] #3541 

### Refactor opportunities fixed In PRs

* Excessive use of `try_from()` which should rather be `new()` for better readability of what's going on. `try_from` is still called by hand each time, so `new()` is really the intend here.
* ~~Allow handlers to be stopped using a message/event through a channel. They are already using this, and all it takes is to expose it. That way `Watcher` doesn't have to be `Clone`, which makes `WatcherInner` obsolete, removing a level of indirection.~~
    - This turned out not to be possible yet, `Clone` is still required. All abstractions were removed though.
* `Handler` should gain a method that has all the code currently distributed in various `handlers` modules, then can also reuse more state, like opened git repositories. This would also do away with `Arc<Mutex>>`, which I doubt are necessary everywhere. It doesn't need to be clone anymore either.
     - Mostly addressed, even though there still is a `handle()` method.
     - ~~The internal message queue… it probably resolves itself then.~~ - a channel is still used to communicate with this long-running process.
* Only keep one active `Watcher`, depending on the project.
     
### Issues related to watcher

* Deletions of the project itself should be recognized as event and cause the UI to act accordingly (Related to #3001). I'd find it unsurprising if the UI would at least go back to the project selector, while showing a message of some kind that indicates what happened.

### Discovered Issues

Issues (re)discovered during the watcher refactoring, but not related to the watcher alone.

* #3511 
* #3062 (lack of worktree support) 
* #3512
* `git::credential::Helpers::from_path()` is used in tests, but depends on the actual `HOME` environment variables, isn't isolated
    - Maybe check other `from_path()` implementations as well.

### Notes

The notes here are partially derived from [my notes branch](https://github.com/gitbutlerapp/gitbutler/compare/master...Byron:gitbutler:notes).

* Application keeps watchers for all registered projects, even though the user can only work in one project at a time right now.
    - It's probably to allow quick-switching of projects. However, while the application is closed, it still has a way to catch up with changes, so it should be possible to do the same when switching a project, and watch only when needed. Maybe it's an optimization though that nobody needs. But in a way, the question is what the app should do if a project triggers filesystem events that aren't relevant to what's currently displayed - seems like unnecessary overhead.
[...]

Issue #3062, opened 2024-03-07
"Add new project" in a worktree fails due to not being able to read the objects directory.


lf-:
Context: I am working on a fork of something that is deliberately worktree'd with its diverged upstream to make cherry-picks faster, and I am using worktree-specific configs on this particular worktree. I know that git-butler is an anti-worktree workflow; that's not what I'm using it for :)

Relevant log:

```
2024-03-07T06:55:27.971806Z  INFO gitbutler-app/src/main.rs:101: starting app version=0.10.22 name=GitButler
2024-03-07T06:55:28.264710Z  INFO get_user: gitbutler-app/src/users/commands.rs:23: close time.busy=13.0µs time.idle=19.7µs
2024-03-07T06:55:28.266643Z  INFO list_projects: gitbutler-app/src/projects/commands.rs:132: close time.busy=4.36µs time.idle=9.38µs
2024-03-07T06:55:38.918216Z ERROR add_project: gitbutler-app/src/projects/commands.rs:79: failed to add project error=failed to open repository

Caused by:
    path not found: /home/jade/dev/sekrit/.git/objects

Stack backtrace:
   0: <E as anyhow::context::ext::StdError>::ext_context
             at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.80/src/context.rs:27:29
   1: anyhow::context::<impl anyhow::Context<T,E> for core::result::Result<T,E>>::context
             at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.80/src/context.rs:54:31
   2: gitbutler_app::projects::controller::Controller::add
             at /home/runner/work/gitbutler/gitbutler/gitbutler-app/src/projects/controller.rs:93:9
   3: gitbutler_app::projects::commands::add_project::{{closure}}::{{closure}}
             at /home/runner/work/gitbutler/gitbutler/gitbutler-app/src/projects/commands.rs:92:5
   4: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
             at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tracing-0.1.40/src/instrument.rs:321:9
   5: gitbutler_app::projects::commands::add_project::{{closure}}
             at /home/runner/work/gitbutler/gitbutler/gitbutler-app/src/projects/commands.rs:87:1
   6: <F as futures_core::future::TryFuture>::try_poll
[...]

PR #3471, opened 2024-04-09
watcher refactor

Byron:
This PR attemps to minimize the watcher implementation to only the needed complexity.

Based on the analysis in #3447.

### Tasks

* [x] *(warmup)*: `try_from()` implementation becomes a constructor method.
* [x] See if consuming `run()` is accidental (it was)
* [x] reorganize dispatcher to work without extra objects and without cancellation token (channels are all that's needed for producers to shut down when consumers shut down, and the dispatcher is a producer)
* [x] See if each `Handler` needs to be managed - are they used and if so, how? Those who aren't shouldn't be managed.
    - It seems they are only managed because they check if they are managed - a catch22?
    - Careful: this could break the app, but if it does, hopefully thoroughly so it's obvious.
* [x] Remove `Clone` from `Watcher` by allowing communication to it using a channel
    - That way, the channel can be `Clone`, but the watcher doesn't have to be.
      This is a precondition for removing all the interior Send+Sync and locking protections,
      while the Watcher will have to watch in the background and thus be in a loop.
* [x] try to implement each `Handler` as function or method.
* [x] remove the event loop by making relevant calls directly (as much as possible)
    - [x] remove last `pure` suffixed function and fixup tests to allow dissolve the last remaining private events
    - [x] fix tests - needs tauri support, which works with mocks 
* [x] fix as many remaining TODO(ST) as possible (only 3 remaining, 2 of which I might fix here)
* [x] make `events::Event` just `Event`
* [x] check usage (also after changing of `.gitignore` at app runtime) of ignore-filter to see if it can be removed/improved to be faster
* [x] try to parallelise delta computation
    - Performance test was `git clone https://github.com/BurntSushi/ripgrep`, then `gco @~200 -- crates/`. After each run, the worktree was reset to `@`.
    - It's definitely working with granularity of a single file, and it's way faster than the single-threaded version
* [x] validate that it is not slower than the application on `master`
    - It's not necessarily faster than the ultra-parallel auto-parallelizing version of the watcher in master. Something I don't understand is that a single file here costs 12s runtime on a single thread, yet master is done way earlier than that (or at least, logging stops earlier). It's like it is not seeing all the file or something else is going on that causes it to skip computations maybe?
* [x] Create issues for application bugs after re-validating them
* [x] Only watch one project at a time

### Notes for the Reviewer

* The delta computation consumes the most CPU, and if it's truly unused, I think it should be removed from the handler at least.
* I removed the rate-limited, as it never kicked in during my testing due to the handling of multiple paths at once. This was an issue to me as well as simply not calculating something would have meant that sometimes, it's not arriving at the correct result.
* I am pretty sure that I never got to test the `fetch` and `push` GB repo functionality, maybe much more. Better test everything that is triggered by watchers.
* LOG_LEVEL=info should probably be preferred when testing the performance of `git checkout @~100 -- src/` or similar as it will still emit a lot of debug messages that are buried deeply in the `core` part. This makes it seem like a lot is happening, even though that's not the case anymore in comparison to the before-times.
* I adjusted the log level of a few traces in `core` to allow launching the application in DEBUG mode without being spammed with low-level information. Instead, DEBUG is now used for select information, and everything else was demoted to the `TRACE` level. It's probably the way things go, what's useful in `DEBUG` one day is better as `TRACE` in another as the focus changes.

### Application Bugs 📋

These bugs were encountered during testing.

* **Adding a file is registered, but removing it often is not**. #3511 
* **Worktrees probably aren't supported** - judging by the assumption in `file_monitor.rs`, `.git` has to be a directory which isn't the case for worktrees.
    - still needs to be formally reproduced, but #3062 already talks about this. and what's mentioned here is definitely related.
    - **possible fix** - detect this case and install a separate notification stream on the extra git directory.
* **Trying to open the linux kernel freezes** for a long time, probably because it wants to download gravatar images of 30k or so contributors. See #3512 .

### Out of Scope but TODO

* `git::credential::Helpers::from_path()` is used in tests, but depends on the actual `HOME` environment variables, isn't isolated
    - Maybe check other `from_path()` implementations as well.

### My Notes

Definitely not a mandatory read to review this PR.

* There seem to be redundant calls to `is_path_ignored()` in many places, given that the dispatcher filter already avoids ignored files (and can do so much more efficiently, but we must assure that it always gets most recent information .gitignore files as the watcher is long-running).
* In theory, since it's going to be communicating with channels *to the outside*, there should be no fear of parallelism at all, so all `Handlers` should have no need. One can hope they are isolated which they are probably not as they are all managed in tauri. This should probably be rolled back as much as possible.
* `Send+Sync` + shared state and `Clone` are poison, and a way to feel unsafe even without the keyword. In such an environment, one has no garantuees anymore, so one will want all code *not* to be dealing with that.
    - This is probably the entrypoint to unravelling `Send+Sync` in `core` as well, even though it might also be a good hint at how these `core` instances can be used. My feeling is that many will need a tauri-specific wrapper, so that `core` can be pure.
* There is a large amount of boilerplate just to deal with all the blanket 'protections' needed for `Send+Sync`, many handlers are just a couple of lines. One handler is just sending an event, effectively.
* The current `event queue` system is actually very cleverly multi-threading everything. Each event coming in is processed in its own (or pooled) thread, and as each even can spawn new events, these will equally be queued and run in parallel. This is why everything is `Send + Sync` in the watcher.
* Seeing the `.git/GB_Flush` file, I wonder if this file-based notification system is truly necessary, or if that could be kept in `tauri`. Maybe the plan is to support multiple windows to be open on the same repository (rather than prevent it), which probably would make this file-based approach necessary, and maybe even the most simple solution if filesystem events can be assumed to be reliable.
* There definitely is some business logic in the watcher that probably wants to move to `core` at some point,.
* There is generally a lot of rather generic string-only `context()` calls and it feels these are 'just there to be there' even though the callee probably could deliver decent error messages from the beginning (and maybe already does so).
* Tests are definitely easier to write with the Events system as side-effects are nicely decoupled via returned events. Now that everything is coupled again, it becomes clear how much state is required for certain functions, making the initial test setup more involved.


crates/gitbutler-tauri/src/watcher/handler.rs
```diff
@@ -243,17 +285,15 @@ impl Handler {
...
-                        },
-                    ),
-                )])
+                self.emit_app_event(&app_events::Event::virtual_branches(
+                    project_id,
+                    &VirtualBranches {
+                        branches,
+                        skipped_files,
+                    },
+                ))
```
mtsgrd, line 294: Good stuff :v:

app/src/routes/[projectId]/+layout.ts
```diff
@@ -27,6 +28,7 @@ export async function load({ params, parent }) {
 	let project: Project | undefined = undefined;
 	try {
 		project = await projectService.getProject(projectId);
+		invoke('set_project_active', { id: projectId }).then((_r) => {});
```
Comment (Byron) [inline app/src/routes/[projectId]/+layout.ts:31]: It explicitly doesn't await here to not block, but I don't know if there is any syntax that makes the intent clearer.

crates/gitbutler-core/src/deltas/writer.rs
```diff
@@ -26,7 +26,7 @@ impl<'writer> DeltasWriter<'writer> {
         self.writer
             .write_string(PathBuf::from("session/deltas").join(path), &raw_deltas)?;
 
-        tracing::debug!(
+        tracing::trace!(
```
Byron, line 29: I hope that by now it's fine to downgrade the severity of these low-level log message to make `debug` more useful for where it matters now. I assume that most `debug` traces will eventually be downgraded or downright removed as confidence (and test-coverage) grows.

crates/gitbutler-tauri/src/watcher/handler/calculate_deltas.rs
```diff
@@ -0,0 +1,178 @@
...
+                }
+                drop(tx);
+
+                for thread in threads {
+                    num_no_delta += thread.join().unwrap()?;
+                }
+                let (_, session) = make_processor()?;
+                session
+            };
+            self.index_session(project_id, &current_session)?;
```
Byron, line 161: Note that this call was previously between line 98 and 99, and thus called per `path`. Thus it seems like a waste to call it so often as it doesn't depend on it. This app seems to work fine like this, but there may be side-effects I am unaware of, particularly because the delta-computation doesn't seem to be displayed in the UI.

app/src/routes/[projectId]/+layout.ts
Comment (Qix-) [reply app/src/routes/[projectId]/+layout.ts:31]: Does this work? If it's returning a future then not awaiting it won't make it tick, will it?

crates/gitbutler-core/src/deltas/writer.rs
Qix-: That's just fine :)

crates/gitbutler-tauri/Cargo.toml
```diff
@@ -29,6 +29,7 @@ backoff = "0.4.0"
 backtrace = { version = "0.3.71", optional = true }
 chrono = { version = "0.4.37", features = ["serde"] }
 console-subscriber = "0.2.0"
+crossbeam-channel = "0.5.12"
```
Qix-, line 32: Is there a reason for crossbeam channels vs using tokio's channels?

crates/gitbutler-tauri/src/events.rs
```diff
@@ -101,22 +76,22 @@ impl Event {
...
         }
     }
 
     pub fn deltas(
-        project_id: &ProjectId,
-        session_id: &SessionId,
-        deltas: &Vec<deltas::Delta>,
+        project_id: ProjectId,
+        session_id: SessionId,
+        deltas: &[deltas::Delta],
```
Qix-, line 94: Just a thought: we could also use a generic here for `AsRef<[...]>`.

crates/gitbutler-tauri/src/projects.rs
```diff
@@ -57,6 +58,23 @@ pub mod commands {
...
+    /// This trigger is the GUI telling us that the project with `id` is now displayed.
+    ///
+    /// We use it to start watching for filesystem events.
+    #[tauri::command(async)]
+    #[instrument(skip(handle), err(Debug))]
+    pub async fn set_project_active(handle: tauri::AppHandle, id: &str) -> Result<(), Error> {
+        let id: ProjectId = id.parse().context(error::Context::new_static(
+            Code::Validation,
+            "Malformed project id",
+        ))?;
```
Qix-: @mtsgrd didn't we end up fixing the project ID acquisition stuff? Or did that never end up landing?

crates/gitbutler-tauri/tests/watcher/handler/push_project_to_gitbutler.rs
```diff
@@ -38,13 +38,14 @@ async fn push_error() -> Result<()> {
         })
         .await?;
 
-    let listener = Handler::new(
-        suite.local_app_data().into(),
-        suite.projects.clone(),
-        suite.users.clone(),
+    let res = Handler::push_project_to_gitbutler_pure(
```
Qix-: What does "pure" mean here?

crates/gitbutler-tauri/src/projects.rs
mtsgrd: We didn't change anything, we just dropped the idea of custom validation failure responses since they are not actually handled by the front end. Instead we can rely on the built-in conversion to respond with appropriate failures (including a description of why the param conversion failed).

A different thing here, where could use `Code::Validation` is if e.g. a project id does not exist.. but we're not there yet.

app/src/routes/[projectId]/+layout.ts
Comment (mtsgrd) [reply app/src/routes/[projectId]/+layout.ts:31]: No it will, promises always execute immediately whether or not you await them.

Comment (Byron) [reply app/src/routes/[projectId]/+layout.ts:31]: It is effective, and won't block the frontend. It is like `tokio::spawn()`, which is done for each and every promise in the frontend.

crates/gitbutler-tauri/Cargo.toml
Byron: Actually, `tokio` channels might be usable here and I didn't try. It would certainly be preferable if it worked.

crates/gitbutler-tauri/src/events.rs
Byron: Absolutely - I usually avoid generics until tests tell me they want it to make it more pleasant to work with the API under test.

crates/gitbutler-tauri/src/projects.rs
Byron: I took a look and apparently it's permissible to just use `ProjectId` there. However, I copied the function and thus inherited not doing so. In `commands.rs` the `ProjectId` type is used directly. This should probably be fixed here.

crates/gitbutler-tauri/tests/watcher/handler/push_project_to_gitbutler.rs
Byron: It's my way of saying that `x_pure()` is the version of `x()` that doesn't have  `&self` parameter. It's used in tests, which then doesn't have to build a whole `Handler`.

However, since this was written, building a `Handler` just for testing is much more common-place, and the tests that use this `pure` function could transition to using the handler, so the function can be moved to the respective method.

Byron:
Thanks for starting the review, @Qix- !

I recommend to `gh pr checkout 3471` this PR and push fixes and changes directly into it. For bigger questions, I will answer everything that might be coming up here. Thanks again.

crates/gitbutler-tauri/src/watcher/file_monitor.rs
```diff
@@ -0,0 +1,226 @@
...
+use anyhow::{anyhow, Context, Result};
+use gitbutler_core::{git, projects::ProjectId};
+use notify::Watcher;
+use notify_debouncer_full::new_debouncer;
+use tokio::task;
+use tracing::Level;
+
+/// The timeout for debouncing file change events.
+/// This is used to prevent multiple events from being sent for a single file change.
+static DEBOUNCE_TIMEOUT: Duration = Duration::from_millis(100);
```
Qix-: Can probably be a `const` (`from_millis` is marked as `const fn`).

app/src/routes/[projectId]/+layout.ts
Comment (Qix-) [reply app/src/routes/[projectId]/+layout.ts:31]: Oh derp this is Typescript, not Rust. My bad :D

crates/gitbutler-tauri/src/watcher/handler/calculate_deltas.rs
Qix-: Can always adjust if something isn't right.

crates/gitbutler-tauri/Cargo.toml
Qix-: Let's try tokio then, lmk if it doesn't work for some reason. Just trying to reduce our number of dependencies.

crates/gitbutler-tauri/src/projects.rs
Qix-: Yeah let's do that. We want to move to that app-wide anyway, I could have sworn we had done that a while back. Maybe some regressions seeped in from somewhere, not sure.

Byron: Alright, https://github.com/gitbutlerapp/gitbutler/pull/3471/commits/62b1c49372972139feec9750a198da8802db79fb should bring it back to the desired state.

crates/gitbutler-tauri/Cargo.toml
Byron: I did take a look and realised that [`tokio::sync::broadcast`](https://docs.rs/tokio/latest/tokio/sync/broadcast/) is indeed able to provide the multi-multi case. Probably last time I didn't really look there though since it requires `async`, whereas this code is all non-async.

`crossbeam-channel` was previously used in the dependency graph by 5 other dependencies and isn't adding to the compile time.

Do you still think `crossbeam-channel` should be replaced with `tokio::sync::broadcast`? If so, I'd probably go with `block_on` where needed as this implementation relies on `std::thread::scope()`. Turning it into `async`-proper for parallelisation would be more effort.

Qix-: That's probably fine, we can move more toward async over time.

Also TIL about `std::thread::scope()` - neat!

crates/gitbutler-tauri/tests/watcher/handler/push_project_to_gitbutler.rs
Byron: Actually, in https://github.com/gitbutlerapp/gitbutler/pull/3471/commits/fb9db89e1bc5cf90575846d2daf138c4234b45b4 I removed this way of running tests in favor of just creating the `Handler` like in all other tests. Thats more consistent, and I think I was just a little too 'träge' earlier 😅.

crates/gitbutler-tauri/src/watcher/file_monitor.rs
Byron: That's a great catch! It's changed in a3fd06897 .

Byron (commented):
Thanks for the review! I think I have addressed all comments thus far, please let me know if anything else comes up.

Qix-:
Thanks!

PR #3476, opened 2024-04-09
Update README.md to include a link to the nightly release page

2Friendly4You:
I've added a link to "https://app.gitbutler.com/downloads" to download the nightly versions as I couldn't find them on Google and other users may be trying to find the links.


schacon:
We debated this a bit, since the nightlies are less stable, but decided that people can take the risk if they want to. I might change the text to "unstable nightlies" or something later. Thanks for the contribution!

PR #3485, opened 2024-04-10
Only allow use of system executable git on Windows

mtsgrd:
- libgit2 does not work on windows


mtsgrd:
@Qix- let me know if this works?

Qix-:
image
there's a double hr here, is that intentional?

mtsgrd:
> there's a double hr here, is that intentional?

Not at all, thanks! Updating in a sec.

PR #3498, opened 2024-04-11
Adding missing "/" to open in vscode

2Friendly4You:
Adding missing "/" to open correctly in vscode

### DIFF
--- a/app/src/lib/components/Board.svelte
+++ b/app/src/lib/components/Board.svelte
@@ -144,8 +144,8 @@
 										class="empty-board__suggestions__link"
 										role="button"
 										tabindex="0"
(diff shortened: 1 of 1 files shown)


2Friendly4You:
Fixes bug in #3495

mtsgrd:
It looks to me like this isn't necessary. It's already part of the path in unix environments, and it's added conditionally here for windows: https://github.com/gitbutlerapp/gitbutler/blob/eb80df0450de3b46f77b77feafb69319e0314abe/app/src/lib/backend/projects.ts#L35

2Friendly4You:
That's absolutely right... I overlooked that part. Sorry for the waste of time

PR #3501, opened 2024-04-12
Always await returned promises

Caleb-T-Owens:
This PR is based on some prior discussion about how we want to handle returning promises.

The main topic of discussion being whether we should prefer:
```ts
createCloudProject(
	token: string,
	params: {
		name: string;
		description?: string;
		uid?: string;
	}
): Promise<CloudProject> {
	return this.httpClient.post('projects.json', {
		body: params,
		token
	});
}
```
or
```ts
async createCloudProject(
	token: string,
	params: {
		name: string;
		description?: string;
		uid?: string;
	}
): Promise<CloudProject> {
	return await this.httpClient.post('projects.json', {
		body: params,
		token
	});
}
```

I have taken a look into the typescript-eslint documentation which provides some relevant rules (which this PR applies), and some compelling justifications for why they should be used.

## Promise async return

Ref: https://typescript-eslint.io/rules/promise-function-async/
Options: default

### Ensuring more predictable exception raising
This rule ensures that each function is only capable of:
- returning a rejected promise, or
- throwing an Error object.

This means that you can't have a function like:

```ts
function foo() {
  if (Math.random() < 0.1) throw new Error("oops");

  const promise = Promise((resolve, reject) => {
    if (Math.random() < 0.1) {
      reject("oh no");
    } else {
       resolve(69);
    }
  }

  return promise;
}
```

Which, if you were to try to consume in a non-async context (which admittedly shouldn't be often as we are preferring async/await syntax over `.then`), would mean that you need to use both a `try catch` and `.catch()` method:
```ts 
// eww - one or the other please!
try {
  foo().then(console.log).catch(console.error);
} catch (e) {
  console.error(e);
}
```

### How does this rule mitigate the aforementioned problem?

This rule requires any function that returns a promise to be marked as async.

This removes the dual case because a function which marked async represents any thrown errors as rejected promises.

IE:
```ts
async function rejectedPromise() {
  throw new Error("oops");

  return somePromise
}

rejectedPromise().catch(console.error) // "oops" is logged
```

### Ensuring functions either return a value or a promise

This rule has the additional side effect of ensuring that you can't return a union of T and Promise<OtherT> which often is a mistake.

### Downsides

This rule means a function like `() => goto('/settings')` must be declared as `async` because `goto()` returns a promise, which can feel a little excessively verbose.

## Return Await

Ref: https://typescript-eslint.io/rules/return-await/
Options: always

There is a lot of conflicting information about this rule out there, which stems from the fact that over 6 years ago, async await was poorly optimised, creating 3 promises behind the scenes for each await. This is no longer the case.

Relevant documents:
- https://v8.dev/blog/fast-async
- http://bit.ly/v8-zero-cost-async-stack-traces

The performance of
```ts
async function foo() {
  return somePromise()
}
```

is the same as
```ts
async function foo() {
  return await somePromise()
}
```

By awaiting the promise, both WebKit and v8 are able to create more relevant stack traces.

Enforcing awaiting promises can also prevent mistakes in try catch blocks

```ts
async function rejectedPromise() {
  throw new Error("whoops!")
}

async function foo() {
  try {
    return rejectedPromise()
  } catch (e) {
    console.log("This is never run")
  }
}
```

In the above example, the error is never caught because the rejected promise will only throw an error when its awaited. This will instead just return a rejected promise from foo().

By enforcing awaiting promises in try catch blocks, we can avoid this mistake too.

## Await Thenable

Ref: https://typescript-eslint.io/rules/await-thenable/

This rule ensures that only objects that implement `.then()` are awaited.


mtsgrd:
This is great, thank you! And, it was a nice read. :)

PR #3510, opened 2024-04-13
Use integration commit for calculating current diffs

mtsgrd:
I investigated why you sometimes see pairs of inverse hunks when working on more than one branch. The main takeaway is that we've been needlessly including diffs for already committed code in the calculations about what uncommitted hunks belong to what branch.

When use the right base then we don't need any complicated `calculate_uncommitted_files`, we simply assign uncommitted hunks based on either branch lock or ownership claim.

### DIFF
--- a/crates/gitbutler-core/src/virtual_branches/virtual.rs
+++ b/crates/gitbutler-core/src/virtual_branches/virtual.rs
@@ -2717,9 +2717,11 @@ pub fn commit(
             })
         })?;
 
+    let uncommitted_base =
+        super::integration::update_gitbutler_integration(gb_repository, project_repository)?;
     // get the files to commit
     let (mut statuses, _) =
-        get_status_by_branch(gb_repository, project_repository, Some(&default_target.sha))
+        get_status_by_branch(gb_repository, project_repository, Some(&uncommitted_base))
             .context("failed to get status by branch")?;
 
     let (ref mut branch, files) = statuses
--- a/crates/gitbutler-core/src/virtual_branches/virtual.rs
+++ b/crates/gitbutler-core/src/virtual_branches/virtual.rs
@@ -15,6 +15,7 @@ use regex::Regex;
 use serde::Serialize;
 
 use super::{
+    base,
     branch::{
         self, Branch, BranchCreateRequest, BranchId, BranchOwnershipClaims, Hunk, OwnershipClaim,
     },
@@ -28,7 +29,7 @@ use crate::{
     gb_repository,
     git::{
         self,
-        diff::{self, diff_files_to_hunks, GitHunk},
+        diff::{self, diff_files_to_hunks},
         show, Commit, Refname, RemoteRefname,
     },
     keys,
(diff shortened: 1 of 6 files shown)


crates/gitbutler-core/src/virtual_branches/virtual.rs
```diff
@@ -833,19 +838,18 @@ pub fn list_virtual_branches(
...
         .iter()
         .filter_map(|(branch, _)| branch.selected_for_changes)
         .max()
         .unwrap_or(-1);
 
     for (branch, files) in &statuses {
-        // check if head tree does not match target tree
-        // if so, we diff the head tree and the new write_tree output to see what is new and filter the hunks to just those
-        let files =
-            calculate_non_commited_diffs(project_repository, branch, &default_target, files)?;
```
krlvi, line 847: Trying to test this out and understand, does skipping this work? Or does `get_status_by_branch` give us something else now?

krlvi: Just saw the commit message, what do you mean by produces it's own input? (there is shadowing of the local var)

mtsgrd: By producing its own input, I mean it did something complicated that returned its original input. That and it resolved conflict markers, I missed that, but also just fixed it.

*edit: it makes sense if you think of it from the perspective that now we start out with only uncommitted hunks, so no need to calculate them.

crates/gitbutler-core/tests/suite/virtual_branches/unapply.rs
```diff
@@ -142,7 +142,7 @@ async fn conflicting() {
         assert!(!branch.active);
         assert!(!branch.base_current);
         assert!(!branch.conflicted);
-        assert_eq!(branch.files[0].hunks[0].diff, "@@ -1 +1 @@\n-first\n\\ No newline at end of file\n+conflict\n\\ No newline at end of file\n");
+        assert_eq!(branch.files[0].hunks[0].diff, "@@ -1 +1 @@\n-second\n\\ No newline at end of file\n+conflict\n\\ No newline at end of file\n");
```
mtsgrd: This one doesn't feel right, need to debug still. *edit: actually, I think it does.. when unapplied, [we are diffing](https://github.com/gitbutlerapp/gitbutler/blob/master/crates/gitbutler-core/src/virtual_branches/virtual.rs#L2104) against `default_target` which has now updated to include commit 2.

crates/gitbutler-core/tests/suite/virtual_branches/apply_virtual_branch.rs
```diff
@@ -124,7 +124,7 @@ async fn rebase_commit() {
         let (branches, _, _) = controller.list_virtual_branches(project_id).await.unwrap();
         assert_eq!(branches.len(), 1);
         assert_eq!(branches[0].id, branch1_id);
-        assert_eq!(branches[0].files.len(), 0);
+        assert_eq!(branches[0].files.len(), 1);
```
mtsgrd: Looks wrong, I'm debugging these.

PR #3517, opened 2024-04-15
more error information when diffy-apply fails

Byron:
This PR adds the most basic error information I can imagine for the case when `diffy::apply()` fails. 
It contains the base-image`and complete hunk information, but with each string replaced by its hash.

The question really is if this kind of `Debug` based output is good enough, or if ti should more closely
resemble how `diffy` hunks are actually printed.

The version you find here merely provides the information, but it's unlikely to be very human-readable.
For now, I also haven't tested it at all.


mtsgrd:
Looks great.

Byron:
Alright - the option here is to make it look prettier and probably more easily graspable, but it's probably OK to see how *this* fares and then make improvements as needed.

PR #3520, opened 2024-04-16
Fix hunk locking when default target is different from locked_to

mtsgrd:


crates/gitbutler-core/src/virtual_branches/virtual.rs
```diff
@@ -1073,7 +1073,8 @@ fn branches_with_hunk_locks(
 }
 
 fn joined(start_a: u32, end_a: u32, start_b: u32, end_b: u32) -> bool {
-    (start_a <= start_b && end_a >= start_b) || (start_a <= end_b && end_a >= end_b)
+    ((start_a >= start_b && start_a <= end_b) || (end_a >= start_b && end_a <= end_b))
+        || ((start_b >= start_a && start_b <= end_a) || (end_b >= start_a && end_b <= end_a))
```
mtsgrd, line 1077: @krlvi fyi this one has been giving us problems, range intersection requires that you check both ways.

PR #3533, opened 2024-04-16
Fix old hunk vs new hunk range comparisons

mtsgrd:
When checking for lock status we should be asking: does the new range of a previously committed hunk intersect with the old range of a new hunk?


krlvi:
Nice catch! Lets encode the desired behavior with a test that fails before this change :)

krlvi:
with the redesign of locking this PR is not likely obsolete

PR #3541, opened 2024-04-17
split `tauri/watcher` into crate

Byron:
This PR further separates the watcher logic from `tauri`.

### Tasks

* [x] copy code from `tauri/watcher` into own crate, bluntly
* [x] crate `analytics` crate *(required by watcher and currently has minor tauri-entanglement)*
* [x] assure tests in `watcher`crate run
* [x] re-integrate `watcher` crate into `tauri`
* [x] assure `watcher`-exclusive dependencies are removed form `tauri` crate
* [x] disable default features in `tokio` to properly setup its features on a per-crate basis

### Notes for the Reviewer

* I added https://github.com/gitbutlerapp/gitbutler/pull/3541/commits/c30ac3ba62ab49e5d41773a3af1e76fb2d205ba4 which is entirely optional here, but think it requires extra-scrutiny to assure there are no unintended side-effects. For the `tauri` crate in particular, I enabled all features that are relevant to the application performance.


Cargo.toml
```diff
@@ -14,10 +16,12 @@ uuid = { version = "1.8.0", features = ["serde"] }
 serde = { version = "1.0", features = ["derive"] }
 thiserror = "1.0.58"
 rusqlite = { version = "0.29.0", features = [ "bundled", "blob" ] }
-tokio = { version = "1.37.0" }
+tokio = { version = "1.37.0", default-features = false }
```
Byron, line 19: The `default-features = false` part has no effect as `tokio` doesn't declare default features, but I thought it's a good indicator for the desire to take full control of any feature that is used.

Cargo.toml
```diff
@@ -1,9 +1,11 @@
 [workspace]
 members = [
+    "crates/gitbutler-analytics",
```
Byron, line 3: The `analytics` crate was a side-effect of making `watcher` standalone, which led to further separating dependencies. Added dependencies that are only used in one module is definitely a good motivation for a separate crate, even if otherwise it might not seem to be quite worth it.

Byron (commented):
Assuming that CI will be green, this PR would be ready for review.

krlvi (approved):

krlvi:
Btw, @Byron could this change https://github.com/gitbutlerapp/gitbutler/commit/c30ac3ba62ab49e5d41773a3af1e76fb2d205ba4 be causing the windows build errors here -> 
https://github.com/gitbutlerapp/gitbutler/actions/runs/8735820206/job/23969386863
?

Byron:
Definitely! So `sync` is needed on windows only. The publishing job is a tricky one as CI doesn't currently trigger all the error paths naturally.

PR #3546, opened 2024-04-18
outcome of researching `hunks_by_filepath`

Byron:
This is the result of looking into the following question:

**Can `gitoxide` solve the problem [described here](https://discord.com/channels/1060193121130000425/1073202153163857920/1230388984048517181)?**

My study showed that that `gitoxide` wouldn't solve the issue, as the problem is the assumption (throughout the codebase) that paths and contents decode as UTF-8.
If that's not the case, some portions of the code might not handle this correctly.

**Possible Solutions**

- `diffy` can work with binary data (assuming lines can be determined), so UTF-8 conversions aren't necessary there.
- Use binary data everywhere, but convert it lossily for display in the user interface.

> [!NOTE]
> The changes proposed here are merely flyby-refactors, and are unrelated to fixing any problem.

### DIFF
--- a/crates/gitbutler-core/src/git/diff.rs
+++ b/crates/gitbutler-core/src/git/diff.rs
@@ -1,7 +1,9 @@
-use std::{collections::HashMap, path, str};
+use std::path::PathBuf;
+use std::{collections::HashMap, str};
 
 use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
+use tracing::instrument;
 
 use super::Repository;
 use crate::git;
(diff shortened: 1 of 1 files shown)


crates/gitbutler-core/src/git/diff.rs
```diff
@@ -174,44 +179,35 @@ fn hunks_by_filepath(
...
                         Some((format!("{}{}", line.origin(), content), false))
                     } else {
-                        let full_path = repository.workdir().unwrap().join(file_path);
-                        // save the file_path to the odb
-                        if !delta.new_file().id().is_zero() && full_path.exists() {
-                            // the binary file wasnt deleted
-                            repository.blob_path(full_path.as_path()).unwrap();
-                        }
-                        Some((delta.new_file().id().to_string(), true))
+                        assume_binary()
```
krlvi, line 197: Oh this is a nice way of removing repetition! I learned something :)

krlvi (approved):

PR #3547, opened 2024-04-18
lossy utf8 conversion on diff read

krlvi:


Qix-:
@krlvi are we sure that this content isn't written back out to the disk anywhere?

crates/gitbutler-core/src/git/diff.rs
```diff
@@ -191,20 +191,14 @@ fn hunks_by_filepath(
...
                 'B' => assume_binary(),
                 'F' => None,
                 _ => {
-                    if let Ok(content) = str::from_utf8(line.content()) {
-                        Some((content.to_string(), false))
-                    } else {
-                        assume_binary()
-                    }
+                    let content = String::from_utf8_lossy(line.content()).to_string();
+                    Some((content.to_string(), false))
```
Qix-:
```suggestion
                    let content = String::from_utf8_lossy(line.content());
                    Some((content.into(), false))
```

Should be sufficient.

krlvi: thanks!

krlvi:
> @krlvi are we sure that this content isn't written back out to the disk anywhere?

yeah, actually. i was a bit trigger happy.

krlvi:
I discussed a bit with @Byron earlier - we probably wanna have to have two separate fields on the struct, with explicitly stating that the string bit is display only

Qix-:
What is stored in the non-display field?

PR #3550, opened 2024-04-18
fix build on Windows

Byron:
As per https://github.com/gitbutlerapp/gitbutler/pull/3541#issuecomment-2063644317, the windows build is broken.

This PR should fix it.

### Tasks

* [x] fix it
* [x] validate fix with quick CI check on Windows

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -203,6 +203,7 @@ jobs:
       - check-gitbutler-core
       - check-gitbutler-changeset
       - check-gitbutler-git
+      - check-rust-windows
     runs-on: ubuntu-latest
     steps:
       - name: Decide whether the needed jobs succeeded or failed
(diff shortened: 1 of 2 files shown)


.github/workflows/push.yaml
```diff
@@ -211,3 +211,10 @@ jobs:
           allowed-skips: ${{ toJSON(needs) }}
           jobs: ${{ toJSON(needs) }}
 
+  check-rust-windows:
+    runs-on: windows-latest
```
Qix-, line 217: can you add an `if` here:

```
    if: ${{ needs.changes.outputs.rust == 'true' }}
```

Qix-: Just makes sure that our typescript-only changes don't take time to build all of rust on windows

Byron: Makes sense, great catch! It's in there now.

Qix-:
Ach sorry, one more thing: The windows check needs to be added to the `needs` list above in the final rust check job, otherwise failures won't fail the entire CI >.>

Qix-:
Thank you 🙏🏻

Issue #3451, opened 2024-04-08
Hunks not in order


schacon:
Since about release 10.28 I believe, we've started getting in some circumstances, a "Hunk not in order or overlap" issue that prevents people from committing and looks something like this:
```
2024-03-27T16:06:54.482959Z ERROR commit_virtual_branch: gitbutler-app/src/virtual_branches/errors.rs:496: commit error error=error parsing patch: Hunks not in order or overlap project_id=[PROJECT_ID] branch=[BRANCH_ID] commit message ownership=Some(BranchOwnershipClaims { claims: [OwnershipClaim { file_path: "src/example.ts", hunks: [Hunk { hash: None, timestamp_ms: None, start: 104, end: 105 }] }] }) run_hooks=false
```

Also seen in Discord: 

* https://discord.com/channels/1060193121130000425/1223970695042568224/1223972691967475875
* https://discord.com/channels/1060193121130000425/1220833850650001428/1223399606675701811
* https://discord.com/channels/1060193121130000425/1073202153163857920/1222582292447363193

Certainly, this should be recoverable and not prevent further usage.

Byron:
Here I my attempts to reproduce the issue, with a local dev build at e7377132bf902b6ee3637cf367aea36a86897915.

I tried to reproduce with [this comment](https://discord.com/channels/1060193121130000425/1223970695042568224/1224840846613151846) from the  [first discord thread](https://discord.com/channels/1060193121130000425/1223970695042568224/1223972691967475875) listed above, but it detects the conflict and refuses to apply the branch. However, the conflict-resolution people talk about also doesn't show up. The UI shows an error in the console though.

```
code: "errors.branches"
message: "Branch f8be11a5-6f21-4302-8f46-c1ca9dab53b5 is in a conflicing state"
```

Interestingly, I also wasn't able to find the error message strings anywhere, and I trued substrings of these in this repo, in `git2-rs` and in `libgit2` (even though a version of `libgit2` is included in the repository of `git2-rs`).

I also noticed that errors aren't actually traced into the log anymore, and I will bring up a PR with a fix later today.

mtsgrd:
I had a look at that comment, and I'm wondering if we have one or two problems? I wasn't expecting "Hunks not in order or overlap" to only show when you try and make a commit, not when applying a conflicting branch.

schacon:
[...]

PR #3552, opened 2024-04-18
binary diff

Byron:
Currently, UTF-8 encoding is assumed in many places in code both for filesystem paths, as well as for file content.

This can lead to abnormal program behaviour in the worst case, or to diffs being considered binary in the best case.

This PR adjusts the types of struct-fields where needed to remove that assumption.

### Tasks

* [x] convert `diff` to `BString` and follow through, everywhere
* [x] graceful fallback if commit-messages can' t be (fully) decoded as UTF8
* [x] validate application
* [x] avoid unnecessary copies of hunks
* [x] ~~use blake3 binary digest as hash~~ abandoned, see "abandoned patch" section
* [x] use binary md5 digest and convert on demand
* [x] fix or resolve all added `TODOs` *(some deferred to review comments)*
* [x] figure out what to do with `Hunk::hash`
* [x] figure out hash-issue for good! (*kind-of*)
* [x] make the `SHA1 in difflines` case typesafe if feasible. (it might not be) *(limited to the hunk generator)*

#### How do line-endings work?

- `GitHunk` - from `git2`, it pre-split lines, they contain terminators.

### Notes for the Reviewer

* For validation, I only moved hunks between virtual branches, and listed commits in trunk, also showing the diff of files.
* The change around `Hunk::hash` is bigger, as it fixes a hash-misuse at least to some extend. Still, it's somewhat unclear how this is supposed to work, despite tests now passing again.
    - **Special attention is needed to assure there is no side-effects with `virtual_branches.toml`** - it *seemed* alright, despite the hashes being changed. Maybe that file is rewritten a lot as the actual source of truth is Git itself.
* I didn't really do performance testing, but didn't notice anything either. In theory, it should be faster as it doesn't have to deal with chars anymore.
* https://github.com/gitbutlerapp/gitbutler/pull/3552/commits/73462182bed888293d73603888e67d93c936f1c2 accidentally pulled me into `virtual.rs`, a monstrosity that deserves attention separately, one day, and with care 😅. The PR was already big enough as well.

### Out of Scope

* I did not specifically look for file-paths which are converted to strings unnecessarily.

### Assumptions

* The GUI doesn't send strings back for the backend to process, at least not those that the backend previously sent to the frontend.

### Abandoned Patch

Trying to use a more efficient representation of hashes, and switching from md5 to `sip` failed, but only for a less obvious reason: the virtual-branches serialization relies on these hashes being MD5, and it's stored as MD5 hex. In order to remain consistent and backward compatible, one would have to put more effort in than I think is warranted as a 'side-task.

For posterity, below the fold is the patch that didn't make it.

<details>

```patch
commit f3849ab2ffa79c70a5df90cd521fba6055f23b18
Date:   Sat Apr 20 14:03:03 2024 +0200

    Use binary hash digest but transform it at the boundary

index e1e7b1d1f..a3d9b6323 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -136,11 +136,10 @@ dependencies = [
 
 [[package]]
 name = "async-executor"
-version = "1.9.1"
+version = "1.11.0"
-checksum = "10b3e585719c2358d2660232671ca8ca4ddb4be4ce8a1842d6c2dc8685303316"
+checksum = "b10202063978b3351199d68f8b22c4e47e4b1b822f8d43fd862d5ea8c006b29a"
 dependencies = [
- "async-lock 3.3.0",
  "async-task",
  "concurrent-queue",
  "fastrand 2.0.2",
@@ -325,7 +324,7 @@ dependencies = [
  "glib-sys",
  "gobject-sys",
  "libc",
- "system-deps 6.2.2",
+ "system-deps 6.2.0",
 ]
 
 [[package]]
@@ -674,7 +673,7 @@ checksum = "3c55d429bef56ac9172d25fecb85dc8068307d17acd74b377866b7a1ef25d3c8"
 dependencies = [
  "glib-sys",
  "libc",
- "system-deps 6.2.2",
+ "system-deps 6.2.0",
 ]
 
 [[package]]
@@ -1848,7 +1847,7 @@ dependencies = [
  "glib-sys",
  "gobject-sys",
  "libc",
- "system-deps 6.2.2",
+ "system-deps 6.2.0",
 ]
 
 [[package]]
@@ -1865,7 +1864,7 @@ dependencies = [
  "libc",
  "pango-sys",
  "pkg-config",
- "system-deps 6.2.2",
+ "system-deps 6.2.0",
 ]
 
 [[package]]
@@ -1879,7 +1878,7 @@ dependencies = [
  "gobject-sys",
  "libc",
  "pkg-config",
- "system-deps 6.2.2",
+ "system-deps 6.2.0",
 ]
 
 [[package]]
@@ -1891,7 +1890,7 @@ dependencies = [
  "gdk-sys",
  "glib-sys",
  "libc",
- "system-deps 6.2.2",
+ "system-deps 6.2.0",
  "x11",
 ]
 
@@ -1983,7 +1982,7 @@ dependencies = [
  "glib-sys",
  "gobject-sys",
  "libc",
- "system-deps 6.2.2",
+ "system-deps 6.2.0",
  "winapi 0.3.9",
 ]
 
@@ -2072,6 +2071,7 @@ dependencies = [
  "serde_json",
  "sha2",
  "similar",
+ "siphasher 1.0.1",
  "slug",
  "ssh-key",
  "ssh2",
@@ -2219,7 +2219,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ef4b192f8e65e9cf76cbf4ea71fa8e3be4a0e18ffe3d68b8da6836974cc5bad4"
 dependencies = [
  "libc",
- "system-deps 6.2.2",
+ "system-deps 6.2.0",
 ]
 
 [[package]]
@@ -2249,7 +2249,7 @@ checksum = "0d57ce44246becd17153bd035ab4d32cfee096a657fc01f2231c9278378d1e0a"
 dependencies = [
  "glib-sys",
  "libc",
- "system-deps 6.2.2",
+ "system-deps 6.2.0",
 ]
 
 [[package]]
@@ -2321,7 +2321,7 @@ dependencies = [
  "gobject-sys",
  "libc",
  "pango-sys",
- "system-deps 6.2.2",
+ "system-deps 6.2.0",
 ]
 
 [[package]]
@@ -2442,12 +2442,6 @@ version = "0.4.1"
 checksum = "95505c38b4572b2d910cecb0281560f54b440a19336cbbcb27bf6ce6adc6f5a8"
 
-[[package]]
-name = "heck"
-version = "0.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
-
 [[package]]
 name = "hermit-abi"
 version = "0.3.9"
@@ -3046,12 +3040,13 @@ checksum = "4ec2a862134d2a7d32d7983ddcdd1c4923530833c9f2ea1a44fc5fa473989058"
 
 [[package]]
 name = "libredox"
-version = "0.1.3"
+version = "0.0.1"
-checksum = "c0ff37bd590ca25063e35af745c343cb7a0271906fb7b37e4813e8f79f00268d"
+checksum = "85c833ca1e66078851dba29046874e38f08b2c883700aa29a03ddd3b23814ee8"
 dependencies = [
  "bitflags 2.5.0",
  "libc",
+ "redox_syscall 0.4.1",
 ]
 
 [[package]]
@@ -3724,7 +3719,7 @@ dependencies = [
  "glib-sys",
  "gobject-sys",
  "libc",
- "system-deps 6.2.2",
+ "system-deps 6.2.0",
 ]
 
 [[package]]
@@ -4458,9 +4453,9 @@ dependencies = [
 
 [[package]]
 name = "redox_users"
-version = "0.4.5"
+version = "0.4.4"
-checksum = "bd283d9651eeda4b2a83a43c1c91b266c40fd76ecd39a50a8c630ae69dc72891"
+checksum = "a18479200779601e498ada4e8c1e1f50e3ee19deb0259c25825a98b5603b2cb4"
 dependencies = [
  "getrandom 0.2.12",
  "libredox",
@@ -5608,15 +5603,15 @@ dependencies = [
 
 [[package]]
 name = "system-deps"
-version = "6.2.2"
+version = "6.2.0"
-checksum = "a3e535eb8dded36d55ec13eddacd30dec501792ff23a0b1682c38601b8cf2349"
+checksum = "2a2d580ff6a20c55dfb86be5f9c238f67835d0e81cbdea8bf5680e0897320331"
 dependencies = [
  "cfg-expr 0.15.7",
- "heck 0.5.0",
+ "heck 0.4.1",
  "pkg-config",
  "toml 0.8.12",
- "version-compare 0.2.0",
+ "version-compare 0.1.1",
 ]
 
 [[package]]
@@ -6523,9 +6518,9 @@ checksum = "1c18c859eead79d8b95d09e4678566e8d70105c4e7b251f707a03df32442661b"
 
 [[package]]
 name = "version-compare"
-version = "0.2.0"
+version = "0.1.1"
-checksum = "852e951cb7832cb45cb1169900d19760cfa39b82bc0ea9c0e5a14ae88411c98b"
+checksum = "579a42fc0b8e0c63b76519a339be31bed574929511fa53c1a3acae26eb258f29"
 
 [[package]]
 name = "version_check"
@@ -6723,7 +6718,7 @@ dependencies = [
  "pango-sys",
  "pkg-config",
  "soup2-sys",
- "system-deps 6.2.2",
+ "system-deps 6.2.0",
 ]
 
 [[package]]
diff --git a/app/src/lib/vbranches/types.ts b/app/src/lib/vbranches/types.ts
index 367b6c943..bc80999ce 100644
--- a/app/src/lib/vbranches/types.ts
+++ b/app/src/lib/vbranches/types.ts
@@ -19,7 +19,7 @@ export class Hunk {
 	})
 	modifiedAt!: Date;
 	filePath!: string;
-	hash?: string;
+	hash?: number;
 	locked!: boolean;
 	lockedTo!: string | undefined;
 	changeType!: ChangeType;
@@ -209,7 +209,7 @@ export const UNKNOWN_COMMITS = Symbol('UnknownCommits');
 
 export class RemoteHunk {
 	diff!: string;
-	hash?: string;
+	hash?: number;
 
 	get id(): string {
 		return hashCode(this.diff);
diff --git a/crates/gitbutler-core/Cargo.toml b/crates/gitbutler-core/Cargo.toml
index ca6b75459..711b4f35a 100644
--- a/crates/gitbutler-core/Cargo.toml
+++ b/crates/gitbutler-core/Cargo.toml
@@ -26,6 +26,7 @@ git2-hooks = "0.3"
 itertools = "0.12"
 lazy_static = "1.4.0"
 md5 = "0.7.0"
+siphasher = "1.0.1"
 r2d2 = "0.8.10"
 r2d2_sqlite = "0.22.0"
 rand = "0.8.5"
diff --git a/crates/gitbutler-core/src/assets.rs b/crates/gitbutler-core/src/assets.rs
index 510d330e8..276cd078b 100644
--- a/crates/gitbutler-core/src/assets.rs
+++ b/crates/gitbutler-core/src/assets.rs
@@ -135,7 +135,7 @@ impl Proxy {
         }
     }
 
-    // takes a url of a remote assets, downloads it into cache and returns a url that points to the cached file
+    // takes the of a remote assets, downloads it into cache and returns the url that points to the cached file
     pub async fn proxy(&self, src: &Url) -> Result<Url> {
         #[cfg(unix)]
         if src.scheme() == "asset" {
@@ -146,6 +146,8 @@ impl Proxy {
             return Ok(src.clone());
         }
 
+        // TODO(ST): when the remote asset handling is overhauled to be (more) lazy (see #3512),
+        //           then `md5` should be replaced with `blake3` so the `md5` dependency can be removed.
         let hash = md5::compute(src.to_string());
         let path = path::Path::new(src.path());
         let ext = path
diff --git a/crates/gitbutler-core/src/reader.rs b/crates/gitbutler-core/src/reader.rs
index b02f48089..e841975fb 100644
--- a/crates/gitbutler-core/src/reader.rs
+++ b/crates/gitbutler-core/src/reader.rs
@@ -18,8 +18,11 @@ pub enum Error {
     Io(Arc<io::Error>),
     #[error(transparent)]
     From(FromError),
-    #[error("failed to parse virtual_branches.toml: {0}")]
-    ParseError(String),
+    #[error("failed to parse {}", path.display())]
+    ParseError {
+        path: PathBuf,
+        source: toml::de::Error,
+    },
 }
 
 impl From<io::Error> for Error {
diff --git a/crates/gitbutler-core/src/virtual_branches/branch.rs b/crates/gitbutler-core/src/virtual_branches/branch.rs
index f826c6e6d..b3f946d0c 100644
--- a/crates/gitbutler-core/src/virtual_branches/branch.rs
+++ b/crates/gitbutler-core/src/virtual_branches/branch.rs
@@ -4,7 +4,7 @@ mod ownership;
 
 use anyhow::Result;
 pub use file_ownership::OwnershipClaim;
-pub use hunk::Hunk;
+pub use hunk::{Hunk, HunkHash};
 pub use ownership::{reconcile_claims, BranchOwnershipClaims};
 use serde::{Deserialize, Serialize};
 
diff --git a/crates/gitbutler-core/src/virtual_branches/branch/hunk.rs b/crates/gitbutler-core/src/virtual_branches/branch/hunk.rs
index e95aac2be..2b650cf71 100644
--- a/crates/gitbutler-core/src/virtual_branches/branch/hunk.rs
+++ b/crates/gitbutler-core/src/virtual_branches/branch/hunk.rs
@@ -1,13 +1,17 @@
+use std::hash::Hasher;
 use std::{fmt::Display, ops::RangeInclusive, str::FromStr};
 
 use anyhow::{anyhow, Context, Result};
 use bstr::{BStr, ByteSlice};
+use siphasher::sip::SipHasher;
 
 use crate::git::diff;
 
+pub type HunkHash = u64;
+
 #[derive(Debug, Eq, Clone)]
 pub struct Hunk {
-    pub hash: Option<String>,
+    pub hash: Option<HunkHash>,
     pub timestamp_ms: Option<u128>,
     pub start: u32,
     pub end: u32,
@@ -70,7 +74,7 @@ impl FromStr for Hunk {
             if raw_hash.is_empty() {
                 None
             } else {
-                Some(raw_hash.to_string())
+                Some(raw_hash.parse()?)
             }
         } else {
             None
@@ -106,7 +110,7 @@ impl Hunk {
     pub fn new(
         start: u32,
         end: u32,
-        hash: Option<String>,
+        hash: Option<HunkHash>,
         timestamp_ms: Option<u128>,
     ) -> Result<Self> {
         if start > end {
@@ -121,8 +125,8 @@ impl Hunk {
         }
     }
 
-    pub fn with_hash(mut self, hash: &str) -> Self {
-        self.hash = Some(hash.to_owned());
+    pub fn with_hash(mut self, hash: HunkHash) -> Self {
+        self.hash = Some(hash);
         self
     }
 
@@ -150,12 +154,13 @@ impl Hunk {
         self.start == other.new_start && self.end == other.new_start + other.new_lines
     }
 
-    // TODO(perf): keep the hash as digest to avoid allocation.
-    pub fn hash(diff: &BStr) -> String {
-        let mut ctx = md5::Context::new();
+    pub fn hash(diff: &BStr) -> HunkHash {
+        let mut hasher = SipHasher::new();
         diff.lines()
             .skip(1) // skip the first line which is the diff header.
-            .for_each(|line| ctx.consume(line));
-        format!("{:x}", ctx.compute())
+            .for_each(|line| {
+                hasher.write(line);
+            });
+        hasher.finish()
     }
 }
diff --git a/crates/gitbutler-core/src/virtual_branches/state.rs b/crates/gitbutler-core/src/virtual_branches/state.rs
index 93a19db47..1709ab6e4 100644
--- a/crates/gitbutler-core/src/virtual_branches/state.rs
+++ b/crates/gitbutler-core/src/virtual_branches/state.rs
@@ -136,8 +136,11 @@ impl VirtualBranchesHandle {
         let mut file: File = File::open(self.file_path.as_path())?;
         let mut contents = String::new();
         file.read_to_string(&mut contents)?;
-        let virtual_branches: VirtualBranches = toml::from_str(&contents)
-            .map_err(|e| crate::reader::Error::ParseError(e.to_string()))?;
+        let virtual_branches: VirtualBranches =
+            toml::from_str(&contents).map_err(|e| crate::reader::Error::ParseError {
+                path: self.file_path.clone(),
+                source: e,
+            })?;
         Ok(virtual_branches)
     }
 
diff --git a/crates/gitbutler-core/src/virtual_branches/virtual.rs b/crates/gitbutler-core/src/virtual_branches/virtual.rs
index 2c493acc6..8ced7b1dc 100644
--- a/crates/gitbutler-core/src/virtual_branches/virtual.rs
+++ b/crates/gitbutler-core/src/virtual_branches/virtual.rs
@@ -20,6 +20,7 @@ use super::{
     },
     branch_to_remote_branch, errors, target, RemoteBranch, VirtualBranchesHandle,
 };
+use crate::virtual_branches::branch::HunkHash;
 use crate::{
     askpass::AskpassBroker,
     dedup::{dedup, dedup_fmt},
@@ -131,7 +132,7 @@ pub struct VirtualBranchHunk {
     pub diff: BString,
     pub modified_at: u128,
     pub file_path: PathBuf,
-    pub hash: String,
+    pub hash: HunkHash,
     pub old_start: u32,
     pub start: u32,
     pub end: u32,
@@ -1840,7 +1841,7 @@ fn get_applied_status(
                                     claimed_hunk
                                         .clone()
                                         .with_timestamp(timestamp)
-                                        .with_hash(hash.as_str()),
+                                        .with_hash(hash),
                                 );
                             } else if claimed_hunk.intersects(git_diff_hunk) {
                                 diffs_by_branch
@@ -1907,7 +1908,7 @@ fn get_applied_status(
                     file_path: filepath.clone(),
                     hunks: vec![Hunk::from(&hunk)
                         .with_timestamp(get_mtime(&mut mtimes, &filepath))
-                        .with_hash(Hunk::hash(hunk.diff.as_ref()).as_str())],
+                        .with_hash(Hunk::hash(hunk.diff.as_ref()))],
                 });
 
             diffs_by_branch
diff --git a/crates/gitbutler-core/tests/virtual_branches/branch/hunk.rs b/crates/gitbutler-core/tests/virtual_branches/branch/hunk.rs
index 99049af31..c11eb2678 100644
--- a/crates/gitbutler-core/tests/virtual_branches/branch/hunk.rs
+++ b/crates/gitbutler-core/tests/virtual_branches/branch/hunk.rs
@@ -13,9 +13,10 @@ fn parse_invalid() {
 
 #[test]
 fn parse_with_hash() {
+    let hash = "123";
     assert_eq!(
-        "2-3-hash".parse::<Hunk>().unwrap(),
-        Hunk::new(2, 3, Some("hash".to_string()), None).unwrap()
+        format!("2-3-{hash}").parse::<Hunk>().unwrap(),
+        Hunk::new(2, 3, Some(hash.parse().unwrap()), None).unwrap()
     );
 }
 
@@ -54,33 +55,33 @@ fn eq() {
             false,
         ),
         (
-            "1-2-abc".parse::<Hunk>().unwrap(),
-            "1-2-abc".parse::<Hunk>().unwrap(),
+            "1-2-123".parse::<Hunk>().unwrap(),
+            "1-2-123".parse::<Hunk>().unwrap(),
             true,
         ),
         (
-            "1-2-abc".parse::<Hunk>().unwrap(),
-            "2-3-abc".parse::<Hunk>().unwrap(),
+            "1-2-123".parse::<Hunk>().unwrap(),
+            "2-3-123".parse::<Hunk>().unwrap(),
             false,
         ),
         (
             "1-2".parse::<Hunk>().unwrap(),
-            "1-2-abc".parse::<Hunk>().unwrap(),
+            "1-2-123".parse::<Hunk>().unwrap(),
             true,
         ),
         (
-            "1-2-abc".parse::<Hunk>().unwrap(),
+            "1-2-123".parse::<Hunk>().unwrap(),
             "1-2".parse::<Hunk>().unwrap(),
             true,
         ),
         (
-            "1-2-abc".parse::<Hunk>().unwrap(),
-            "1-2-bcd".parse::<Hunk>().unwrap(),
+            "1-2-123".parse::<Hunk>().unwrap(),
+            "1-2-456".parse::<Hunk>().unwrap(),
             false,
         ),
         (
-            "1-2-abc".parse::<Hunk>().unwrap(),
-            "2-3-bcd".parse::<Hunk>().unwrap(),
+            "1-2-123".parse::<Hunk>().unwrap(),
+            "2-3-234".parse::<Hunk>().unwrap(),
             false,
         ),
     ] {
diff --git a/crates/gitbutler-core/tests/virtual_branches/branch/ownership.rs b/crates/gitbutler-core/tests/virtual_branches/branch/ownership.rs
index d50f888ad..4a57eb845 100644
--- a/crates/gitbutler-core/tests/virtual_branches/branch/ownership.rs
+++ b/crates/gitbutler-core/tests/virtual_branches/branch/ownership.rs
@@ -16,13 +16,13 @@ fn reconcile_ownership_simple() {
                     Hunk {
                         start: 1,
                         end: 3,
-                        hash: Some("1,3".to_string()),
+                        hash: Some(Hunk::hash("1,3".into())),
                         timestamp_ms: None,
                     },
                     Hunk {
                         start: 4,
                         end: 6,
-                        hash: Some("4,6".to_string()),
+                        hash: Some(Hunk::hash("4,6".into())),
                         timestamp_ms: None,
                     },
                 ],
@@ -39,7 +39,7 @@ fn reconcile_ownership_simple() {
                 hunks: vec![Hunk {
                     start: 7,
                     end: 9,
-                    hash: Some("7,9".to_string()),
+                    hash: Some(Hunk::hash("7,9".into())),
                     timestamp_ms: None,
                 }],
             }],
@@ -54,13 +54,13 @@ fn reconcile_ownership_simple() {
             Hunk {
                 start: 4,
                 end: 6,
-                hash: Some("4,6".to_string()),
+                hash: Some(Hunk::hash("4,6".into())),
                 timestamp_ms: None,
             },
             Hunk {
                 start: 7,
                 end: 9,
-                hash: Some("9,7".to_string()),
+                hash: Some(Hunk::hash("9,7".into())),
                 timestamp_ms: None,
             },
         ],
@@ -78,7 +78,7 @@ fn reconcile_ownership_simple() {
                 hunks: vec![Hunk {
                     start: 1,
                     end: 3,
-                    hash: Some("1,3".to_string()),
+                    hash: Some(Hunk::hash("1,3".into())),
                     timestamp_ms: None,
                 },],
             }],
@@ -94,13 +94,13 @@ fn reconcile_ownership_simple() {
                     Hunk {
                         start: 4,
                         end: 6,
-                        hash: Some("4,6".to_string()),
+                        hash: Some(Hunk::hash("4,6".into())),
                         timestamp_ms: None,
                     },
                     Hunk {
                         start: 7,
                         end: 9,
-                        hash: Some("9,7".to_string()),
+                        hash: Some(Hunk::hash("9,7".into())),
                         timestamp_ms: None,
                     },
                 ],
```

</details>

### DIFF
--- a/crates/gitbutler-core/src/deltas/operations.rs
+++ b/crates/gitbutler-core/src/deltas/operations.rs
@@ -8,6 +8,7 @@ use similar::{ChangeTag, TextDiff};
 #[serde(rename_all = "camelCase")]
 pub enum Operation {
     // corresponds to YText.insert(index, chunk)
+    // TODO(ST): Should probably be BString, but it's related to delta-code.
     Insert((usize, String)),
     // corresponds to YText.remove_range(index, len)
     Delete((usize, usize)),
--- a/crates/gitbutler-core/src/git/diff.rs
+++ b/crates/gitbutler-core/src/git/diff.rs
@@ -9,6 +9,9 @@ use tracing::instrument;
 
 use super::Repository;
 use crate::git;
+use crate::virtual_branches::BranchStatus;
+
+pub type DiffByPathMap = HashMap<PathBuf, FileDiff>;
 
 /// The type of change
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
@@ -107,10 +110,7 @@ pub struct FileDiff {
 }
 
 #[instrument(skip(repository))]
-pub fn workdir(
-    repository: &Repository,
-    commit_oid: &git::Oid,
-) -> Result<HashMap<PathBuf, FileDiff>> {
+pub fn workdir(repository: &Repository, commit_oid: &git::Oid) -> Result<DiffByPathMap> {
     let commit = repository
         .find_commit(*commit_oid)
         .context("failed to find commit")?;
@@ -143,7 +143,7 @@ pub fn trees(
     repository: &Repository,
     old_tree: &git::Tree,
     new_tree: &git::Tree,
-) -> Result<HashMap<PathBuf, FileDiff>> {
+) -> Result<DiffByPathMap> {
     let mut diff_opts = git2::DiffOptions::new();
     diff_opts
         .recurse_untracked_dirs(true)
@@ -163,8 +163,8 @@ pub fn without_large_files(
     size_limit_bytes: u64,
     diff: &git2::Diff,
     mut diff_opts: git2::DiffOptions,
-) -> (git2::DiffOptions, HashMap<PathBuf, FileDiff>) {
-    let mut skipped_files: HashMap<PathBuf, FileDiff> = HashMap::new();
+) -> (git2::DiffOptions, DiffByPathMap) {
+    let mut skipped_files = HashMap::new();
     for delta in diff.deltas() {
         if delta.new_file().size() > size_limit_bytes {
             if let Some(path) = delta.new_file().path() {
@@ -199,16 +199,13 @@ pub fn without_large_files(
 /// `repository` should be `None` if there is no reason to access the workdir, which it will do to
 /// keep the binary data in the object database, which otherwise would be lost to the system
 /// (it's not reconstructable from the delta, or it's not attempted).
-fn hunks_by_filepath(
-    repo: Option<&Repository>,
-    diff: &git2::Diff,
-) -> Result<HashMap<PathBuf, FileDiff>> {
+fn hunks_by_filepath(repo: Option<&Repository>, diff: &git2::Diff) -> Result<DiffByPathMap> {
     enum LineOrHexHash<'a> {
         Line(Cow<'a, BStr>),
         HexHashOfBinaryBlob(String),
     }
     // find all the hunks
-    let mut diff_files: HashMap<PathBuf, FileDiff> = HashMap::new();
+    let mut diff_files = HashMap::new();
 
     diff.print(
         git2::DiffFormat::Patch,
@@ -411,14 +408,8 @@ pub fn reverse_hunk(hunk: &GitHunk) -> Option<GitHunk> {
     }
 }
 
-pub fn diff_files_to_hunks(
-    files: &HashMap<PathBuf, FileDiff>,
-) -> HashMap<PathBuf, Vec<git::diff::GitHunk>> {
-    let mut file_hunks: HashMap<PathBuf, Vec<git::diff::GitHunk>> = HashMap::new();
-    for (file_path, diff_file) in files {
-        if !diff_file.skipped {
-            file_hunks.insert(file_path.clone(), diff_file.hunks.clone());
-        }
-    }
-    file_hunks
+// TODO(ST): turning this into an iterator will trigger a cascade of changes that
+//           mean less unnecessary copies. It also leads to `virtual.rs` - 4k SLOC!
+pub fn diff_files_into_hunks(files: DiffByPathMap) -> BranchStatus {
+    HashMap::from_iter(files.into_iter().map(|(path, file)| (path, file.hunks)))
 }
--- a/crates/gitbutler-core/src/git/diff.rs
+++ b/crates/gitbutler-core/src/git/diff.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::path::PathBuf;
 use std::{collections::HashMap, str};
 
@@ -52,6 +53,37 @@ pub struct GitHunk {
     pub change_type: ChangeType,
 }
 
+/// Lifecycle
+impl GitHunk {
+    /// A special hunk that signals a binary file whose complete content is a blob under `hex_id` in Git.
+    /// `changetype` is tells us what happened with the file.
+    fn binary_marker(hex_id: String, change_type: ChangeType) -> Self {
+        GitHunk {
+            old_start: 0,
+            old_lines: 0,
+            new_start: 0,
+            new_lines: 0,
+            diff_lines: hex_id.into(),
+            binary: true,
+            change_type,
+        }
+    }
+
+    /// Return a hunk that represents a new file by convention.
+    fn generic_new_file() -> Self {
+        Self {
+            old_start: 0,
+            old_lines: 0,
+            new_start: 0,
+            new_lines: 0,
+            diff_lines: Default::default(),
+            binary: false,
+            change_type: ChangeType::Modified,
+        }
+    }
+}
+
+/// Access
 impl GitHunk {
     pub fn contains(&self, line: u32) -> bool {
         self.new_start <= line && self.new_start + self.new_lines >= line
@@ -63,8 +95,12 @@ impl GitHunk {
 pub struct FileDiff {
     pub old_path: Option<PathBuf>,
     pub new_path: Option<PathBuf>,
-    pub hunks: Option<Vec<GitHunk>>,
+    /// Hunks might be empty if nothing about the files content is known, which happens
+    /// if the content is skipped due to it being a large file.
+    pub hunks: Vec<GitHunk>,
     pub skipped: bool,
+    /// This is `true` if this is a file with undiffable content. Then, `hunks` might be a single
+    /// hunk that is the hash of the binary blob in Git.
     pub binary: bool,
     pub old_size_bytes: u64,
     pub new_size_bytes: u64,
@@ -94,7 +130,7 @@ pub fn workdir(
     if !skipped_files.is_empty() {
         diff = repository.diff_tree_to_workdir(Some(&tree), Some(&mut diff_opts))?;
     }
-    let diff_files = hunks_by_filepath(repository, &diff);
+    let diff_files = hunks_by_filepath(Some(repository), &diff);
     diff_files.map(|mut df| {
         for (key, value) in skipped_files {
             df.insert(key, value);
@@ -120,7 +156,7 @@ pub fn trees(
     let diff =
         repository.diff_tree_to_tree(Some(old_tree), Some(new_tree), Some(&mut diff_opts))?;
 
-    hunks_by_filepath(repository, &diff)
+    hunks_by_filepath(None, &diff)
 }
 
 pub fn without_large_files(
@@ -137,7 +173,7 @@ pub fn without_large_files(
                     FileDiff {
                         old_path: delta.old_file().path().map(ToOwned::to_owned),
                         new_path: delta.new_file().path().map(ToOwned::to_owned),
-                        hunks: None,
+                        hunks: Vec::new(),
                         skipped: true,
                         binary: true,
                         old_size_bytes: delta.old_file().size(),
@@ -156,12 +192,22 @@ pub fn without_large_files(
     (diff_opts, skipped_files)
 }
 
+/// Transform `diff` into a mapping of `worktree-relative path -> FileDiff`, where `FileDiff` is
+/// all the diff-related information one could ask for. This is mainly to workaround `git2`
+/// which doesn't provide a format that is easy to use or hunk-based, but it's line-by-line only.
+///
+/// `repository` should be `None` if there is no reason to access the workdir, which it will do to
+/// keep the binary data in the object database, which otherwise would be lost to the system
+/// (it's not reconstructable from the delta, or it's not attempted).
 fn hunks_by_filepath(
-    repository: &Repository,
+    repo: Option<&Repository>,
     diff: &git2::Diff,
 ) -> Result<HashMap<PathBuf, FileDiff>> {
+    enum LineOrHexHash<'a> {
+        Line(Cow<'a, BStr>),
+        HexHashOfBinaryBlob(String),
+    }
     // find all the hunks
-    let mut hunks_by_filepath: HashMap<PathBuf, Vec<GitHunk>> = HashMap::new();
     let mut diff_files: HashMap<PathBuf, FileDiff> = HashMap::new();
 
     diff.print(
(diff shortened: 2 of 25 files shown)


crates/gitbutler-core/src/deltas/operations.rs
```diff
@@ -8,6 +8,7 @@ use similar::{ChangeTag, TextDiff};
 #[serde(rename_all = "camelCase")]
 pub enum Operation {
     // corresponds to YText.insert(index, chunk)
+    // TODO(ST): Should probably be BString, but it's related to delta-code.
```
Byron, line 11: This comment will be removed with the entire feature once it's time to do that, or the string can be turned into a `BString` if the feature evolves.

crates/gitbutler-core/src/git/diff.rs
```diff
@@ -42,7 +43,8 @@ pub struct GitHunk {
     pub old_lines: u32,
     pub new_start: u32,
     pub new_lines: u32,
-    pub diff: String,
+    #[serde(serialize_with = "crate::serde::as_string_lossy")]
```
Byron: Initially I thought the translation layer to the frontend is the return value of `tauri` commands, but in fact, it's `serde`. Thus, all type conversions should rather be done with custom serializers. In this case, `BString` doesn't have `serde` support enabled so one cannot forget to add this custom attribute.

The UI will fail with the default serialization of BString, which is `[1,2,3]` an array of 8 bit numbers.

crates/gitbutler-core/src/git/diff.rs
```diff
@@ -277,6 +273,12 @@ fn hunks_by_filepath(
         .map(|(k, v)| {
             if let Some(binary_hunk) = v.iter().find(|hunk| hunk.binary) {
                 if v.len() > 1 {
+                    // TODO(ST): Would it be possible here to permanently discard lines because
```
Byron: I would like to defer this question to the reviewer. In theory, one could try to construct a test that runs into the concern described here, and since it means data loss, the time should *probably* be invested to find out.
But I wanted to double-check.
*(Note that no tests runs into this branch)*

Byron (commented):
This PR is really close to being ready for review, but I found something else that needs to be discussed separately.

crates/gitbutler-core/src/virtual_branches/branch/hunk.rs
```diff
@@ -157,12 +154,35 @@ impl Hunk {
...
+        let mut hashed_something = false;
+        diff.lines()
+            .skip(1) // skip the first line which is the diff header.
+            .for_each(|line| {
+                hashed_something = true;
+                ctx.consume(line)
+            });
+        assert!(
+            hashed_something,
+            "BUG: input must have a first line, but nothing was hashed here (len={}): {diff}",
```
Byron: When putting it to the test, I noticed that the hash function can be called with empty `diff` input, something that shouldn't be possible.
After all, this hash function has requirements to its input and only then will it produce the expected hash.
Now I am not sure at all if this is called consistently, and since assertion is hit during tests something about these hashes seems fishy. The question now is if they are used consistently enough for the system to work. Maybe related to #3451 .

Byron: And what makes it worse is that these hashes are used quite interchangeably: like in [`Hunk::from(GitHunk)`](https://github.com/gitbutlerapp/gitbutler/blob/7c2e25a9e4c63581144faf403ebd0424109831af/crates/gitbutler-core/src/virtual_branches/branch/hunk.rs#L23).

crates/gitbutler-core/src/virtual_branches/branch/hunk.rs
```diff
@@ -17,7 +20,7 @@ impl From<&diff::GitHunk> for Hunk {
         Hunk {
             start: hunk.new_start,
             end: hunk.new_start + hunk.new_lines,
-            hash: Some(Hunk::hash(&hunk.diff)),
+            hash: Some(Hunk::hash(hunk.diff_lines.as_ref())),
```
Byron: Here a `hash` that is supposed to be created by `Hunk::hash_diff` is set with one that is just a normal hash. That *should* be fine as `diff_lines` are only lines, without a header, BUT it's not the same as the line-endings are included here. Even if `hash_diff()` would include line endings as well, previously [these have been synthetic](https://github.com/gitbutlerapp/gitbutler/blob/3005c067298edd7c6670fbec99f36d8b778296f2/crates/gitbutler-core/src/virtual_branches/branch/hunk.rs#L165) as well.
So no matter what, I don't think these hashes will be very comparable unless great care is taken regarding the way these lines are tokenized, and how that token is (or isn't) reproduced in the hash.

Byron: Hmm, after fiddling with it more it seems that possible values for `diff_lines` aren't as diverse as I thought and most of them do contains the `@@` header. Probably this hasn't been a cause of problems.

crates/gitbutler-core/src/virtual_branches/branch/hunk.rs
Byron: This was my mistake - it's probably OK to have empty hunks, as long as they indeed have a `@@` header line.

crates/gitbutler-core/src/git/diff.rs
Byron: Maybe no test runs into this branch because `git2` doesn't actually provide more diff-lines for binary files, but just a single marker.

crates/gitbutler-core/src/deltas/operations.rs
krlvi: its reasonable to leave for now

krlvi (approved):

PR #3566, opened 2024-04-20
Use blame for hunk locking

mtsgrd:
Based on the findings from my last refactor of this code I realised we were trying to recreate `blame` when checking if any new hunk intersects with by range with anything already committed. To keep things efficient we limit the blames to match the hunk ranges and to only go as far back as the target base.

- [x] replace commit walking with git blame
- [x] create something equivalent of `integration_commit` that includes the commits (for blame)
- [x] enable hunk locking to multiple commits
- [ ] figure out what to do if hunk is blamed on two non-intersecting diffs in different lanes


crates/gitbutler-core/src/virtual_branches/branch/hunk.rs
```diff
@@ -2,14 +2,17 @@ use std::{fmt::Display, ops::RangeInclusive, str::FromStr};
...
+
+use super::Branch;
 
 #[derive(Debug, Eq, Clone)]
 pub struct Hunk {
     pub hash: Option<String>,
     pub timestamp_ms: Option<u128>,
     pub start: u32,
     pub end: u32,
+    pub locked_to: Option<Vec<Id<Branch>>>,
```
krlvi: Consider dropping the Option here - since you have a Vec, it can be empty

mtsgrd: Ah, thanks, that was added by mistake.

crates/gitbutler-core/src/project_repository/repository.rs
```diff
@@ -611,6 +612,19 @@ impl Repository {
...
+        path: &path::Path,
+        min_line: u32,
+        max_line: u32,
+        oldest_commit: &git::Oid,
+        newest_commit: &git::Oid,
+    ) -> Result<Blame, RemoteError> {
+        Ok(self
+            .git_repository
+            .blame(path, min_line, max_line, oldest_commit, newest_commit)?)
+    }
```
krlvi: We can skip having this extra layer (the same function on the project repository) - the git repository is already accessible and this just adds a layer of indirection

krlvi (approved):
nice! i hope this puts an end to the hunk locking saga, this feels intuitively the right solution. 

we can probably keep an eye if it adds any noticeable performance reduction with the blame calls, and if it's a problem swap out the blame for the one in the oxide library ;)

crates/gitbutler-core/src/virtual_branches/integration.rs
```diff
@@ -16,9 +16,112 @@ lazy_static! {
...
+        Ok(None)
+    }
+}
+
+fn write_integration_file(head: &git::Reference, path: PathBuf) -> Result<()> {
+    let sha = head.target().unwrap().to_string();
+    std::fs::write(path, format!(":{}", sha))?;
+    Ok(())
+}
+
```
mtsgrd: Should this live in a different file?

crates/gitbutler-core/src/virtual_branches/virtual.rs
```diff
@@ -1761,70 +1773,85 @@ fn get_applied_status(
...
+                            let hash = Hunk::hash_diff(hunk.diff_lines.as_ref());
+                            locked_hunk_map
+                                .entry(hash)
+                                .and_modify(|commits| {
+                                    commits.push(commit);
+                                })
+                                .or_insert(vec![commit]);
                         }
                     }
                 }
```
mtsgrd, line 1796: This is the main part.

crates/gitbutler-core/src/virtual_branches/virtual.rs
```diff
@@ -1916,11 +1941,16 @@ fn get_applied_status(
...
+    // process the remaining ones.
     for (filepath, hunks) in base_diffs {
         for hunk in hunks {
             let hash = Hunk::hash_diff(hunk.diff_lines.as_ref());
-            let vbranch_pos = if let Some(locked_to) = git_hunk_map.get(&hash) {
-                let p = virtual_branches.iter().position(|vb| vb.id == *locked_to);
+            let locked_to = locked_hunk_map.get(&hash);
+
+            let vbranch_pos = if let Some(locked_to) = locked_to {
+                let branch_id = commit_to_branch.get(&locked_to[0]).unwrap();
```
mtsgrd, line 1905: This unwrap should never fail, but we should probably fail gracefully and warn that it happened.

Issue #2672, opened 2024-02-14
Can not add project with git-ignored unowned files/folders


NobbZ:
I have a project that contains a `docker-compose.yaml` which again uses volumes in the source tree.

Due to the nature of how those work, the owning UID/GID are not known to the host system and the current user does not have any permissions to read files or recurse folders.

The (sanitized) output of git-butler for that operation is the following:

```
2024-02-14T11:03:39.791046Z ERROR add_project: gitbutler-app/src/projects/commands.rs:75: failed to add project error=failed to open repository

Caused by:
    0: failed to run initial flush
    1: failed to build working directory tree
    2: failed to build wd index
    3: failed to working directory list files in /home/nmelzer/Projects/anonymous
    4: IO error for operation on /home/nmelzer/Projects/anonymous/data/base_version: Permission denied (os error 13)
    5: Permission denied (os error 13) path="/home/nmelzer/Projects/anonymous"
```

But:

```
$ cd /home/nmelzer/Projects/anonymous
$ rg data .gitignore
39:/data/
```

So git butler should not care for the folder or its contents. And it should probably ignore anything else that is git-ignored.

alexgherman:
can confirm, have the same issue

Mekcyed:
Having the same Problem with my Dev-Container.

jibon0070:
Can confirm, same issue

jibon0070:
any update on it

Byron:
It seems that the error message originates [in this function](https://github.com/gitbutlerapp/gitbutler/blob/26c39f2a3fd9be887c3d7f2c808db5ae680a6771/crates/gitbutler-core/src/fs.rs#L14), which does indeed not respect `.gitignore` files.

From what I could gather, it is trying to build a git tree from a working tree (`build_wd_tree_from_repo()`), which probably should respect ignore files right away. It does try [to do that later](https://github.com/gitbutlerapp/gitbutler/blob/7a4dc32ba7bbb27b754f7183b6ee04409c3df3ec/crates/gitbutler-core/src/gb_repository/repository.rs#L678), but that is too late.

This could be fixed by using a [`dirwalk_iter()`](https://docs.rs/gix/latest/gix/struct.Repository.html#method.dirwalk_iter) instead, at least in this case where a working tree is listed.

krlvi:
This change should now be part of the `0.11.4` release

Byron:
[...]

PR #3568, opened 2024-04-20
respect ignore files when listing worktree directories

Byron:
This introduces `gix` to the codebase and uses it selectively to traverse a worktree while respecting git excludes.

Related to #2672.

### Tasks

* [x] function to list worktree files specifically
* [x] use new function where appropriate, i.e. were worktrees are used.

### Notes for the Reviewer

* I changed the way worktree files are iterated to being unsorted, as sorting wasn't required for the correct functioning here while forcing all paths to be held in memory. Now paths are iterated, while additionally doing that in the background with buffering so while the consumer is busy creating a git tree, the traversal keeps happening in the background.

#### `ignore` crate

This crate can do the same, even though it might be less correct as it might not pick up all .gitignore files, which might be configured in git configuration files.
However, since going `gitoxide` is desired, this seems like a good incentive to add it to the codebase and allow its usage.

#### Use `git2` to check ignores

It would be as possible to use `git2` during the traversal to filter out directories which are ignored, as well as files. However, since `is_path_ignored()` is known to be very slow, getting slower the bigger the repository, I'd want to use it less rather than more.
In a way, the current implementation already uses it, and using it to filter some directories probably isn't going to make it much worse, but I guess `is_path_ignored()` is something I feel strongly about.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -106,6 +106,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "arc-swap"
+version = "1.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "69f7f8c3906b62b754cd5326047894316021dcfe5a194c8ea52bdd94934a3457"
+
+[[package]]
 name = "arrayvec"
 version = "0.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


krlvi (approved):
looked over the code - such a great introduction of gix to the repo. Lets merge this and further evaluate

Issue #2608, opened 2024-02-12
Error while opening Project with unreadable file in worktree


iammohan01:
When I tried to open a currently working git repository. I got error like this

<img width="1208" alt="image" src="https://github.com/gitbutlerapp/gitbutler/assets/85432311/ade2fb81-7d5d-483a-b5a5-df9d99989e1d">

Then I cloned this repo and Tried to see what's causing Issue.
In there I found that, One file don't have a read permission in that directory.
That's causing issue.

When i add that is in .gitignore, It's working fine.
```

2024-02-12T18:13:14.208850Z ERROR add_project: gitbutler-app/src/projects/commands.rs:75: failed to add project error=failed to open repository

Caused by:
    0: failed to run initial flush
    1: failed to build working directory tree
    2: failed to build wd index
    3: failed to add working directory path .../liveStreaming.csv
    4: failed open - '/Users/..../liveStreaming.csv' is locked: Permission denied; class=Os (2) path="/Users/..."
```
<img width="1728" alt="SCR-20240212-ullo-2" src="https://github.com/gitbutlerapp/gitbutler/assets/85432311/e7301ad2-c34d-4676-a586-2b5812eaabf4">.

Byron:
By now as of Version 0.11.3 (20240419.070253), the error looks like this, so is a little more descriptive:

<img width="326" alt="Screenshot 2024-04-21 at 14 36 56" src="https://github.com/gitbutlerapp/gitbutler/assets/63622/1853d1fb-778a-442e-8269-46ee2e0ef22f">

The log still looks the same.

Even though it would be trivial to ignore errors, the question really is how GB should work under these circumstances.

For now, I settled for making the error message more indicative of what the error source is.

<img width="404" alt="Screenshot 2024-04-21 at 15 15 53" src="https://github.com/gitbutlerapp/gitbutler/assets/63622/4dee11b1-c1a2-42a0-ae11-74a270ae274a">


<details>

```
2024-04-21T12:36:49.688354Z ERROR add_project: crates/gitbutler-tauri/src/projects.rs:27: error=Error(failed to open repository

Caused by:
    0: failed to run initial flush
    1: failed to build working directory tree
    2: failed to build wd index
    3: failed to add working directory path forbidden
    4: failed open - '/Users/byron/dev/git.inept.dev/~doyle/rgit/forbidden' is locked: Permission denied; class=Os (2)) path="/Users/byron/dev/git.inept.dev/~doyle/rgit"
[...]

PR #3573, opened 2024-04-21
provide better errors when adding a project with unreadable files (#2608)

Byron:
This improves the error message peopel see when a worktree contains files that can't be read.

Improves on #2608.

#### Before

Screenshot 2024-04-21 at 14 36 56

#### After

Screenshot 2024-04-21 at 15 15 53


krlvi (approved):
thanks for this improvement!

Issue #2904, opened 2024-02-24
Gitea support


FederAndInk:
Feature request for [gitea](https://about.gitea.com/) Pull Requests support.

I understand it's still very early, but at least it can be tracked here :)

himchawla:
Hey, are you able to use gitea with gitbutler. Using my private hosted gitea whenever i tried to push it gave me something went wrong, I'll attach a screenshot and maybe logs (if required/ if i can find) tomorrow.

> Feature request for [gitea](https://about.gitea.com/) Pull Requests support.
> 
> I understand it's still very early, but at least it can be tracked here :)

techknowlogick:
This would be great! I’m a part of the Gitea project and would love to help from our side of things (code, debugging, or otherwise). I’m also trialing a new bounty platform, and I think this would be a good candidate. Gitbutler maintainers: would you be ok if I posted a bounty from algora.io?

ryanm101:
Did you check your repo to see if the push did happen? I get that sometimes on Github, but when i look the commit has actually been pushed

questmaster:
Yes, I checked. It didn't work. Nothing pushed.

techknowlogick:
> This would be great! I’m a part of the Gitea project and would love to help from our side of things (code, debugging, or otherwise). I’m also trialling a new bounty platform, and I think this would be a good candidate. Gitbutler maintainers: would you be ok if I posted a bounty from algora.io?

With @schacon using [algora.io for zed](https://twitter.com/chacon/status/1770005036170375594), I hope posting one here is fine (if not, please delete the post from the algora bot and I can delete it from my side too).

Rutik7066:
@FederAndInk I would like to work on this issue. Could you please assign this.

himchawla:
Using this new option works like a charm,

![image](https://github.com/gitbutlerapp/gitbutler/assets/58020143/e0a310ff-07ff-427c-89a0-f37044ea4f1e)

And having test credentials option right there is such a convinience, thanks a bunch gitbutler team.

![image](https://github.com/gitbutlerapp/gitbutler/assets/58020143/17b3e069-5cd6-44de-b6a1-7f04f65e6878)

syedbarimanjan:
@Byron can i work on this.

Byron:
Thanks for offering! GitLab support was supposed to be added first, which would certainly bring some abstractions to make integrations a little more pluggable. Thus, I think it's best to wait on #2511 .

[...]

PR #3575, opened 2024-04-21
Allow SSH remotes to use any port

Caleb-T-Owens:
This PR solves the problem where a self-hosted gitea repository cannot have remote operations performed. (Pushing and fetching from the remote should now be working). This seems to have been due to poor handling of non-standard SSH ports. My testing setup had gitea's SSH happening on port 2222.

Related: https://github.com/gitbutlerapp/gitbutler/issues/2904

It should be noted that this PR does not close the above issue or claim the bounty because this doesn't add support for creating Pull requests (although making that area of the code more generic and supporting more providers is something I'm keen to do)

I've tested on GitHub and Gitea, but it would be great if another Gitea user could test it on their setup to ensure that the problem is in-fact resolved.

### DIFF
--- a/crates/gitbutler-core/src/ssh.rs
+++ b/crates/gitbutler-core/src/ssh.rs
@@ -27,7 +27,7 @@ pub fn check_known_host(remote_url: &git::Url) -> Result<(), Error> {
         return Ok(());
     };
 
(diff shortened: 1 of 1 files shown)


Qix-:
Thanks!

PR #3589, opened 2024-04-23
make git system executable the default auth flow

krlvi:


crates/gitbutler-git/src/repository.rs
```diff
@@ -82,7 +82,7 @@ where
     // TODO(qix-): suffix. The actual executables live in the parent directory.
     // TODO(qix-): Thus, we have to do this under test. It's not ideal, but
     // TODO(qix-): it works for now.
-    #[cfg(test)]
+    #[cfg(feature = "test-askpass-path")]
     let path = path.parent().unwrap();
```
Byron, line 86: This works locally no on MacOS, but not on CI just yet. I will leave it at that for now.

CI seems to be running `git test -p gitbutler-core --all-features`, which works locally and doesn't affect how the `test-*` feature is used. So probably the path isn't valid on CI.

krlvi: nah i bet the problem is that the ask pass binary hasnt been compiled. i will add a step to make sure it is

PR #3595, opened 2024-04-24
use `gix::tempfile` in place of the `tempfile`.

Byron:
This allows to unify handling of filewrites in more places.

### DIFF
--- a/crates/gitbutler-core/src/virtual_branches/state.rs
+++ b/crates/gitbutler-core/src/virtual_branches/state.rs
@@ -1,3 +1,4 @@
+use gix::tempfile::{AutoRemove, ContainingDirectory};
 use std::{
     collections::HashMap,
     fs::File,
@@ -5,6 +6,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
+use crate::storage;
 use serde::{Deserialize, Serialize};
 
 use super::{target::Target, Branch};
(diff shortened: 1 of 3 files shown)


krlvi (approved):
thanks for this!

crates/gitbutler-core/src/virtual_branches/state.rs
```diff
@@ -151,10 +153,11 @@ impl VirtualBranchesHandle {
...
 fn write<P: AsRef<Path>>(file_path: P, virtual_branches: &VirtualBranches) -> anyhow::Result<()> {
     let contents = toml::to_string(&virtual_branches)?;
-    let temp_file = tempfile::NamedTempFile::new_in(file_path.as_ref().parent().unwrap())?;
-    let (mut file, temp_path) = temp_file.keep()?;
-    file.write_all(contents.as_bytes())?;
-    drop(file);
-    std::fs::rename(temp_path, file_path.as_ref())?;
-    Ok(())
+    let mut temp_file = gix::tempfile::new(
+        file_path.as_ref().parent().unwrap(),
```
krlvi, line 157: is it safe to do "unwrap" here, @Byron?

Byron: It's best avoided, but the only reason this fails is if the path is `/` I think. So in practice, it's safe, but probably best to prefer `.expect("why")` here.

PR #3596, opened 2024-04-24
avoid to make files executable on Windows (#3329).

### LINKED ISSUE #3329: Windows gitbutler is changing my file permissions in the git from 644 to 755
Sorry no repro details right now just wanted to mention. Feel free to close if you are tracking this a different way.

Byron:
Instead, respect the existing bit if set in Git, but do not try to derive it from the filesystem which is something Git doesn't seem to be doing on Windows either.

This also makes me think that `gitoxide` should make it easy to build a tree from a file and deal with this.

Fixes #3329

### DIFF
--- a/crates/gitbutler-core/src/virtual_branches/virtual.rs
+++ b/crates/gitbutler-core/src/virtual_branches/virtual.rs
@@ -2101,9 +2101,7 @@ pub fn write_tree_onto_tree(
     let git_repository = &project_repository.git_repository;
     let mut builder = git_repository.treebuilder(Some(base_tree));
     // now update the index with content in the working directory for each file
-    for (filepath, hunks) in files {
-        // convert this string to a Path
-        let rel_path = Path::new(&filepath);
+    for (rel_path, hunks) in files {
         let full_path = project_repository.path().join(rel_path);
 
         let is_submodule = full_path.is_dir()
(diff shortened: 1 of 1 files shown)


crates/gitbutler-core/src/virtual_branches/virtual.rs
```diff
@@ -2122,12 +2120,21 @@ pub fn write_tree_onto_tree(
...
+
                 #[cfg(target_os = "windows")]
                 {
-                    // TODO(qix-): Pull from `core.filemode` config option to determine
-                    // TODO(qix-): the behavior on windows. For now, we set this to true.
-                    // TODO(qix-): It's not ideal, but it gets us to a windows build faster.
-                    filemode = git::FileMode::BlobExecutable;
+                    // NOTE: *Keep* the existing executable bit if it was present
+                    //       in the tree already, don't try to derive something from
+                    //       the FS that doesn't exist.
```
Qix-, line 2128: This looks good, but is there a way to have `filemode` (line 2120) pull from the config, too?

Byron: I don't think it's applicable as we do not get an 'executable' bit with any Rust API. If we would get that bit, `core.filemode = false` would mean we ignore it, but there is no way to get it in the first place.

Or in other words, `core.filemode` doesn't mean that everything is executable by default, but I don't have a Windows installation ready yet to validate this.

Qix-: Hmm alright. Well we'll see if anyone has an issue with this behavior then.

Qix-:
Thanks :)

crates/gitbutler-core/src/virtual_branches/virtual.rs
Byron: Yes, it's a bit strange. Interestingly I never had that problem with `gitoxide` as it can't alter the index yet and add files to it, which would have run into this as well.

Issue #3591, opened 2024-04-23
Error-Popups don't show anymore


Byron:
As of Version 0.11.4 (20240423.110032), error popups don't show anymore. The last known working version is Version 0.11.3.

### Reproduction

* add a newly initialized git repository as project

This now fails without any effect being visible in the UI. The logs will show the error though, and in debug mode, the console will show it too.

<img width="603" alt="Screenshot 2024-04-23 at 19 07 53" src="https://github.com/gitbutlerapp/gitbutler/assets/63622/9a2aed4c-eceb-4269-ba60-105eff8b963a">

Byron:
Bisection revealed that the issue was introduced in this commit: 5c9ce25cd009c1f2493f0cf81a86b69061be573a . Interestingly, there is no error in the console or anything like that.

mtsgrd:
@PavelLaptev can you take a look?

PavelLaptev:
@mtsgrd sure. I'll take a look

PavelLaptev:
@Byron @mtsgrd this PR https://github.com/gitbutlerapp/gitbutler/pull/3599 should solve the issue. There was the wrong condition check
from Rust we receive errors like just `error` and not `error.message`

PR #3599, opened 2024-04-24
fix: didn't show error comes from Rust

PavelLaptev:
fix of this issue https://github.com/gitbutlerapp/gitbutler/issues/3591

### DIFF
--- a/app/src/lib/notifications/toasts.ts
+++ b/app/src/lib/notifications/toasts.ts
@@ -22,7 +22,8 @@ export function showToast(toast: Toast) {
 }
 
 export function showError(title: string, err: any) {
-	if (err.message) showToast({ title, errorMessage: err.message, style: 'error' });
+	const errorMessage = err.message ? err.message : err;
+	showToast({ title, errorMessage: errorMessage, style: 'error' });
 }
 
 export function dismissToast(messageId: string | undefined) {


app/src/lib/notifications/toasts.ts
```diff
@@ -22,7 +22,8 @@ export function showToast(toast: Toast) {
 }
 
 export function showError(title: string, err: any) {
```
Byron, line 24: Would it be possible to strongly type this `err`? Or is that prevented by `err` somehow being `string | Error`?

From all I can tell, the backend [should always provide a map with `code` and `message`](https://github.com/gitbutlerapp/gitbutler/blob/5a9962bc5a602ab932a36445715ccae03a2016c3/crates/gitbutler-tauri/src/error.rs#L60-L68), but apparently at least in this case, that doesn't happen.

Byron: Screenshot 2024-04-24 at 17 04 40

```
2024-04-24T15:04:26.171529Z ERROR add_project: crates/gitbutler-tauri/src/projects.rs:27: error=Error(Bare repositories are unsupported

Caused by:
    bare repositories are not supported) path="/Users/byron/dev-gb/bare"
```

This looks like the error type is not a string even when it arrives in the frontend.

Do errors come in from other places than [the IPC code here](https://github.com/gitbutlerapp/gitbutler/blob/dac89e41a980d3e5acfef50f624d8539d5f0079a/app/src/lib/backend/ipc.ts#L26-L31)?

Byron (commented):
Thanks! It definitely works again.

Screenshot 2024-04-24 at 17 04 32

Please see my comments as well, I think there might be a more general issue with error handling and I'd love to fix the source of it. Ideally, the frontend can then refer to errors with a type as well *(which might be out of scope for this PR, of course)*.

PavelLaptev:
Got it! Thanks for checking this @Byron. I'll go ahead and merge the PR to fix the bug asap. 
@mtsgrd, tagging you here in case you want to review @Byron's feedback as well.

Issue #3601, opened 2024-04-24
Access denied error in latest nightly on Windows


daniilS:
Since updating to 0.5.427 (sadly not sure what my previously installed version was exactly, I only know that I installed it at some point last week), I keep getting "Access is denied. (os error 5)" errors on Windows, usually with "Failed to fetch" or "Failed load remote branches". This happens when first opening Gitbutler, or trying to switch or add a new project. Sometimes closing and reopening Gitbutler will load the project and just display the toast, other times it won't load the project at all. When a project does load, actions like creating or deleting a virtual branch sometimes have no effect, and work fine at other times. I've tried running Gitbutler with admin privileges, which didn't solve the problem.
![image](https://github.com/gitbutlerapp/gitbutler/assets/6144189/2aa45f36-290a-4111-ad98-6e958e5f38a6)
![image](https://github.com/gitbutlerapp/gitbutler/assets/6144189/e4643c61-c50a-460e-a48a-8bf55d375ef5)

Byron:
Thanks a lot for reporting!

The very low-level error is probably a result of #3582, which is intended as with help of [the logs](https://docs.gitbutler.com/development/debugging#logs) we should be able to see where it is coming from. If you could find such errors in the logs and post them here, we could identify what's happening and tune the errors to be more descriptive.

The root cause here is probably that reads and writes happen in parallel, which isn't allowed on Windows and GitButler has to get better at coordinating these tasks.

akbar-mirza:
I’m encountering a similar issue. It seems to have surfaced in the latest 2-3 nightly builds. Based on the logs, this issue first occurred with app version 0.5.428.

Latest logs (0.5.429) : 
```
Caused by:
    Access is denied. (os error 5)) [3mproject_id[0m[2m=[0md19669c3-798e-4b64-bfb3-b6458dca4b2e [3maction[0m[2m=[0mSome("auto")
2024-04-25T03:43:39.513733Z  INFO fetch_from_target: crates\gitbutler-tauri\src\virtual_branches.rs:411: close time.busy=5.10ms time.idle=53.9µs [3mproject_id[0m[2m=[0md19669c3-798e-4b64-bfb3-b6458dca4b2e [3maction[0m[2m=[0mSome("auto")
2024-04-25T03:43:39.517170Z  INFO list_projects: crates\gitbutler-tauri\src\projects.rs:51: new
[...]

PR #3603, opened 2024-04-25
windows improvements

Byron:
This PR starts a Windows experiment that reduces the parallelism of the filesystem watcher, but it will also
reduce the snappiness of the UI.

Related to #3601 .

### Notes for the Reviewer

* Run with `LOG_LEVEL=debug pnpm tauri dev --features adapt-to-windows` to try the windows adaptations.

### DIFF
--- a/crates/gitbutler-tauri/Cargo.toml
+++ b/crates/gitbutler-tauri/Cargo.toml
@@ -63,6 +63,9 @@ features = [
     "window-unmaximize"
 ]
 
+[target."cfg(windows)".features]
+gitbutler-watcher = { workspace = true, features = ["adapt-to-windows"] }
+
 [lints.clippy]
 all = "deny"
 perf = "deny"
@@ -70,6 +73,8 @@ correctness = "deny"
 
 [features]
 default = ["custom-protocol", "sentry", "devtools"]
+## A forwarding to all crates that have windows-specific adjustments for testing on non-Windows.
+adapt-to-windows = ["gitbutler-watcher/adapt-to-windows"]
 devtools = ["tauri/devtools"]
 
 # this feature is used used for production builds where `devPath` points to the filesystem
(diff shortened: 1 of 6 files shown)


crates/gitbutler-tauri/Cargo.toml
```diff
@@ -63,13 +63,18 @@ features = [
     "window-unmaximize"
 ]
 
+[target."cfg(windows)".features]
```
Qix-: Do you mean `dependencies` here?

crates/gitbutler-tauri/Cargo.toml
```diff
@@ -63,13 +63,18 @@ features = [
...
+
 [lints.clippy]
 all = "deny"
 perf = "deny"
 correctness = "deny"
 
 [features]
 default = ["custom-protocol", "sentry", "devtools"]
+## A forwarding to all crates that have windows-specific adjustments for testing on non-Windows.
+adapt-to-windows = ["gitbutler-watcher/adapt-to-windows"]
```
Qix-: Could this just be `windows`?

crates/gitbutler-tauri/Cargo.toml
Byron: That's a great catch! Actually I meant to change the default feature based on the platform, but that doesn't work (yet).
Hence I think the only way to do this is to explicitly enable the feature, which I hope I have done for `release.sh` and for CI.

crates/gitbutler-tauri/Cargo.toml
Byron: A good call :) - it's `windows` now.

Byron:
Thanks for the review! I think my changes to `release.sh` need extra scrutiny as I didn't test them at all, assuming it works.
I will also have to get a Windows test setup going to be able to actually run a build.

crates/gitbutler-tauri/Cargo.toml
Qix-: > Actually I meant to change the default feature based on the platform

Yeah 🙄 I've been waiting for Cargo to support some of this for _years_ for my OS project. It's a huge PITA.

Qix-:
Should be alright, let's give it a try.

PR #3604, opened 2024-04-25
Add test for ensuring hunks can lock across branches

mtsgrd:
While we haven't yet decided how to handle this edge case we still want to test that locking works as expected. With two hunks committed into separate branches, introducing a third that touches both creates a situation where it's locked to both.

### DIFF
--- a/crates/gitbutler-core/tests/suite/virtual_branches/create_commit.rs
+++ b/crates/gitbutler-core/tests/suite/virtual_branches/create_commit.rs
@@ -26,7 +26,7 @@ async fn should_lock_updated_hunks() {
 
     {
         // by default, hunks are not locked
-        write_file(repository, "file.txt", &vec!["content".to_string()]);
+        write_file(repository, "file.txt", &["content".to_string()]);
 
         let branch = get_virtual_branch(controller, project_id, branch_id).await;
         assert_eq!(branch.files.len(), 1);
@@ -42,7 +42,7 @@ async fn should_lock_updated_hunks() {
 
     {
         // change in the committed hunks leads to hunk locking
-        write_file(repository, "file.txt", &vec!["updated content".to_string()]);
+        write_file(repository, "file.txt", &["updated content".to_string()]);
 
         let branch = controller
             .list_virtual_branches(project_id)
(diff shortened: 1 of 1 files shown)


crates/gitbutler-core/tests/suite/virtual_branches/create_commit.rs
```diff
@@ -221,43 +182,167 @@ async fn should_double_lock() {
...
+    let files = get_virtual_branch(controller, project_id, branch_2_id)
+        .await
+        .files;
+    assert_eq!(files.len(), 1);
+}
+
+// TODO: Where should these helper functions live?
+
+fn write_file(repository: &TestProject, path: &str, lines: &Vec<String>) {
+    fs::write(repository.path().join(path), lines.join("\n")).unwrap()
```
mtsgrd, line 275: There are 197 other variations of this line where this could be used, but we'd need a function that accepts lines as a vector or as a single string.

crates/gitbutler-core/tests/suite/virtual_branches/create_commit.rs
```diff
@@ -221,43 +182,167 @@ async fn should_double_lock() {
...
+    branch_id: Id<Branch>,
+) -> VirtualBranch {
+    controller
+        .list_virtual_branches(project_id)
+        .await
+        .unwrap()
+        .0
+        .into_iter()
+        .find(|b| b.id == branch_id)
+        .unwrap()
```
mtsgrd, line 301: @krlvi I see you have a `get_branch` function in `VirtualBranchesHandle`. Could we make tests use this struct as well?

crates/gitbutler-core/tests/suite/virtual_branches/create_commit.rs
```diff
@@ -221,43 +182,185 @@ async fn should_double_lock() {
...
+        .create_virtual_branch(project_id, &branch::BranchCreateRequest::default())
+        .await
+        .unwrap();
+
+    let mut request_opts = branch::BranchCreateRequest::default();
+    request_opts.selected_for_changes = Some(true);
+    let branch_2_id = controller
+        .create_virtual_branch(project_id, &request_opts)
+        .await
+        .unwrap();
```
mtsgrd: Is there a way of inlining the `selected_for_changes` option?

PR #3606, opened 2024-04-25
Reset files into the same branch they were committted to

mtsgrd:
- regressed when switching to blame for hunk locking

### DIFF
--- a/crates/gitbutler-core/src/virtual_branches/virtual.rs
+++ b/crates/gitbutler-core/src/virtual_branches/virtual.rs
@@ -16,13 +16,14 @@ use git2_hooks::HookResult;
 use regex::Regex;
 use serde::Serialize;
 
+use super::integration::get_workspace_head;
 use super::{
     branch::{
         self, Branch, BranchCreateRequest, BranchId, BranchOwnershipClaims, Hunk, OwnershipClaim,
     },
     branch_to_remote_branch, errors, target, RemoteBranch, VirtualBranchesHandle,
 };
-use crate::git::diff::{diff_files_into_hunks, DiffByPathMap};
+use crate::git::diff::{diff_files_into_hunks, trees, DiffByPathMap};
 use crate::virtual_branches::branch::HunkHash;
 use crate::{
     askpass::AskpassBroker,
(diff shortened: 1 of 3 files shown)


crates/gitbutler-core/src/virtual_branches/virtual.rs
```diff
@@ -2042,11 +2043,51 @@ pub fn reset_branch(
...
+            .find_commit(updated_head)
+            .map_err(anyhow::Error::from)?
+            .tree()
+            .map_err(anyhow::Error::from)?,
+        &repo
+            .find_commit(old_head)
+            .map_err(anyhow::Error::from)?
+            .tree()
+            .map_err(anyhow::Error::from)?,
+    )?;
```
mtsgrd, line 2061: @Byron any chance you could help me make this a bit more trim?

Byron: Let me see….

Byron: Yes, and no. I pushed a commit that allows to use `?` in quite  straightforward way, which probably pays off quickly.

What follows *would* work if there was trait specialization, but it's a missing feature I keep falling for.

```patch
diff --git a/crates/gitbutler-core/src/virtual_branches/errors.rs b/crates/gitbutler-core/src/virtual_branches/errors.rs
index f802ae893..cf8473109 100644
--- a/crates/gitbutler-core/src/virtual_branches/errors.rs
+++ b/crates/gitbutler-core/src/virtual_branches/errors.rs
@@ -55,6 +55,17 @@ pub enum ResetBranchError {
     DefaultTargetNotSet(DefaultTargetNotSet),
     #[error(transparent)]
     Other(#[from] anyhow::Error),
+    #[error(transparent)]
+    Any(Box<dyn std::error::Error + Send + Sync>),
+}
+
+impl<E> From<E> for ResetBranchError
+where
+    E: std::error::Error + Send + Sync,
+{
+    fn from(err: E) -> Self {
+        ResetBranchError::Any(err.into())
+    }
 }
 
 impl ErrorWithContext for ResetBranchError {
diff --git a/crates/gitbutler-core/src/virtual_branches/virtual.rs b/crates/gitbutler-core/src/virtual_branches/virtual.rs
index 0caff1d5f..ccf1606a7 100644
--- a/crates/gitbutler-core/src/virtual_branches/virtual.rs
+++ b/crates/gitbutler-core/src/virtual_branches/virtual.rs
@@ -2056,16 +2056,8 @@ pub fn reset_branch(
     let repo = &project_repository.git_repository;
     let diff = trees(
         repo,
-        &repo
-            .find_commit(updated_head)
-            .map_err(anyhow::Error::from)?
-            .tree()
-            .map_err(anyhow::Error::from)?,
-        &repo
-            .find_commit(old_head)
-            .map_err(anyhow::Error::from)?
-            .tree()
-            .map_err(anyhow::Error::from)?,
+        &repo.find_commit(updated_head)?.tree()?,
+        &repo.find_commit(old_head)?.tree()?,
     )?;
 
     // Assign the new hunks to the branch we're working on.
```

Also Rust nightly doesn't allow that or have it in a working condition, unfortunately. But probably it's for the better here as what's pushed is simpler.

mtsgrd: Thanks! 🙏

PR #3620, opened 2024-04-27
skip tempfile under windows

krlvi:
possible workaround for https://github.com/gitbutlerapp/gitbutler/issues/3601


Byron:
I see, the file-sync didn't work :/. Would have been too nice.

PR #3664, opened 2024-05-02
Improve advice for npm prefix on Windows

anaisbetts:
Instead of trouncing the Program Files permissions, let's set an npm prefix instead


krlvi (approved):
thank you for documenting this!

PR #3665, opened 2024-05-02
Even Better npm advice

anaisbetts:
Fix a typo in the command, oops. Also add some hints as to where to get Perl


Qix-:
Thanks :)

PR #3683, opened 2024-05-03
Update the normalizeBranchName function to match rust

Caleb-T-Owens:
Before, uppercase letters and hashes were missing from the regex. I've now coppied the regex from the rust code so they are now identical


Caleb-T-Owens:
We could always expose the rust function in a tauri controller, though it feels a little overkill

krlvi (approved):
thanks!

PR #3688, opened 2024-05-04
Optimize list_virtual_branches

Caleb-T-Owens:
I've marked this as ready for review, I'm certain there are still areas which could do with improving, but this should already be a great step forwards and shouldn't be held back

---

I've taken a look into the list_virtual_branches function and there appears to be three main places which slow it down:
- Ordering hunks within a vbranch
- Calling `get_applied_status`
  - It appears like the slowest part of this function is the blame locking, though I don't see a particularly easy way to tackle that.
- `diffs_into_virtual_files` -> `virtual_hunks_into_virtual_files`
  - This is O(n) and doesn't have anything particularly slow, its just a hot loop.
- Something about starting is blocking the UI thread which is certainly problematic

## Ordering the hunks within a vbranch

The changes in this commit optimize the sorting of virtual branch files
by pre-computing the index positions of the ownership claims for each
file path. This improves the performance of the sorting operation by
avoiding the need to repeatedly search for the index positions during
the comparison step.

With 1900 hunks in a vbranch, I've found that this helps reduce the run time from 9 seconds down to just 5.

I considered not sorting at all if we didn't have any branches, but this didn't provide any substantial change in performance.

## Something about starting is blocking the UI thread which is certainly problematic

Turns out that this is due to flex.... 🤦.

By removing some un-needed flex classes, I've reduced the render time from 11 seconds down to under 2

Before:
Screenshot 2024-05-06 at 00 22 14

After:
image

This is still far from perfect, and some form of virtual scroll (https://github.com/v1ack/svelte-virtual-scroll-list) would be ideal, but the pre-made solutions don't seem to quite fit how we want our designs to behave.

EDIT: with https://github.com/gitbutlerapp/gitbutler/pull/3688/commits/645726ab937e758b4c90ba65f0558c387e2bd4e2 the UI no longer is frozen on the loading wheel for that 2 seconds

### DIFF
--- a/app/src/lib/components/BranchFiles.svelte
+++ b/app/src/lib/components/BranchFiles.svelte
@@ -39,8 +39,6 @@
 <style lang="postcss">
 	.branch-files {
 		flex: 1;
-		display: flex;
-		flex-direction: column;
 		background: var(--clr-bg-1);
 		border-radius: var(--radius-m) var(--radius-m) 0 0;
 		padding: 0 var(--size-14) var(--size-14);
--- a/app/src/lib/components/BranchFilesList.svelte
+++ b/app/src/lib/components/BranchFilesList.svelte
@@ -4,6 +4,7 @@
 	import { getContext } from '$lib/utils/context';
 	import { selectFilesInList } from '$lib/utils/selectFilesInList';
 	import { maybeMoveSelection } from '$lib/utils/selection';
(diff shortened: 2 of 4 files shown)


krlvi:
thanks, lemme try this out!

krlvi (approved):
seems to work, lets try this out on the nightly

PR #3692, opened 2024-05-05
Ignore .env.development and friends

anaisbetts:
We explicitly don't want to commit these, let's ignore them

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@
 .idea
 
 .DS_Store
+
+.env
+.env.*


krlvi (approved):

mderriey:
I think you'll need to explicitly remove them for this to take effect, they still seem to be present in the [`app` folder](https://github.com/gitbutlerapp/gitbutler/tree/7fdd85f5222498c42b8bb863d480e48161a98cd4/app).

PR #3696, opened 2024-05-06
build: adopt to resolve.conditions: es2015

imcotton:
This tweak will letting the vite[^1] to leverage the ES2015 export condition in case of rxjs[^2].

[^1]: https://vitejs.dev/config/shared-options#resolve-conditions
[^2]: https://rxjs.dev/guide/installation#es2015-via-npm


krlvi (approved):
thanks!

PR #3697, opened 2024-05-06
Clean up win32 menus

anaisbetts:
A few small fixups to make the top-level menus more in-line with platform expectations

## TODO:

- [x] Arrange some of the menu items better
- [x] ~~Add standard accelerators that are missing~~ (https://github.com/tauri-apps/wry/issues/451)


Qix-:
Thanks!

PR #3699, opened 2024-05-06
sort project list alphabetically

### LINKED ISSUE #3698: Projects are not sorted alphabetically
At the moment the projects are not sorted by default and probably listed based on the creation date.
Once you have a few items on the list, it would be useful to sort them **alphabetically**, especially if you need to rename them.

image

Qix-:
closes #3698 

Before:

grafik

After: 

grafik

### DIFF
--- a/crates/gitbutler-core/src/projects/storage.rs
+++ b/crates/gitbutler-core/src/projects/storage.rs
@@ -51,7 +51,7 @@ impl Storage {
         match self.inner.read(PROJECTS_FILE)? {
             Some(projects) => {
                 let all_projects: Vec<project::Project> = serde_json::from_str(&projects)?;
-                let all_projects: Vec<project::Project> = all_projects
+                let mut all_projects: Vec<project::Project> = all_projects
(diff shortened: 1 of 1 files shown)


app/src/lib/components/ProjectsPopup.svelte
```diff
@@ -27,9 +27,10 @@
 {#if !hidden}
 	<div class="popup" class:collapsed={isNavCollapsed}>
 		{#if $projects.length > 0}
+			{@const sortedProjects = $projects.sort((a, b) => a.title.localeCompare(b.title))}
```
mtsgrd: Would it maybe make more sense to sort earlier than here in the template?

Qix-: Aren't `hidden` and `isNavCollapsed` stores? Or are they normal variables? If they're stores then projects changing would cause a rerender even if the component is hidden, right?

If not though, then yeah - I'll just move it up top.

PR #3701, opened 2024-05-06
feat: add flake.nix for nix dev

ndom91:
Quick way for additional folks to get up and running and contributing to GitButler

### Changes

- Add `flake.nix` for Nix-based quick development environment
- Can be automatically sourced with [`direnv`](https://github.com/direnv/direnv) when `cd`-ing into the directory or manually triggered by running `nix develop` in the directory. The user will then have all packages and settings, etc. to be able to run `pnpm tauri dev`, for example, locally.


krlvi (approved):
thanks! i think there were a handful of nix devs who were toying with the projects, i think they will appreciate this. 

I am not sure if there is a better place to keep those (that is not the repo root) but we can always move it later

PR #3702, opened 2024-05-06
feat: add `@playwright/test` e2e test setup

ndom91:
### Description

- Adds `@playwright/test` E2E test-runner and framework
- new npm scripts
	- `test:e2e` - run "headless" tests with playwright
	- `test:e2e:run` - run webapp with tauri mocking enabled (`vite dev --mode testing`)
	- `test:e2e:watch` - run with playwright "ui mode" enabled, see screenshot below, good for development
- Adds dummy initial test to check if page renders and initial changed file from fixtures are shown

image

- Adds `e2e.yml` github action ("Playwright Test") to run tests in CI ([Example run here](https://github.com/gitbutlerapp/gitbutler/actions/runs/8968182812/job/24628008065?pr=3702))
- Triggers on:
	- `push` to `master`
	- `pull_requests` to `master`
	- Cron schedule at `6:00` daily
	- Manual (`workflow_dispatch`)
	- Obviously feel free to remove any you may or may not need
- GHA failures will generate a report (zip file) that can be downloaded from the GHA run "summary" screen. They can then be uploaded and viewed at https://trace.playwright.dev to get an interactive viewer just like the local one shown above. See demo [trace.zip](https://github.com/ndom91/gitbutler/files/15214042/trace.zip) here.

### Potential Follow-up

- GitHub reporter to comment status on PR's - https://github.com/estruyf/playwright-github-actions-reporter

### Challenges

- Mocking multiple use-cases; For example, we can set it to answer `invoke('get_base_branch_data')` with a certain fixture once, but if we want to test a different case where the underlying base branch data is different, or missing, etc. How should we do that? :thinking: 
	- Idea: maybe just split it up by projects? If we select a different project, that'll have different branches, etc. it'll request and we can filter on that and respond with the correct one for each test case.
- There's an issue with the left-most navigation panel not rendering in the mocked mode (`test:e2e:run`), but it's not erroring out or indicating anything else that may be missing
- Also wasn't sure how best to type these fixture objects. I basically wanted to do the opposite of class-transformers `plainToInstance`, but in Typescript. We can't just use the class as type here because the class properties are all typed for after the transformation and these fixture objects are the raw data as is sent from the BE. Basically we need a TS version of the rust structs for these objects :joy:
- There's definitely a better way to structure the mocked `invoke` return values (`/app/src/lib/testing/index.ts`), but this definitely covered the base use-case. A simple switch case statement is tough because a lot of these require a secondary value you're comparing against (i.e. `cmd === 'git_head' && args.id === 'abc123'`). But maybe some nested switch statements (?)


mtsgrd:
>Basically we need a TS version of the rust structs for these objects

Have you considered generating them? https://github.com/oscartbeaumont/tauri-specta

PR #3710, opened 2024-05-06
Don't reparse branch_id

Caleb-T-Owens:
This should make the situation where an error occurs even more absurd. It should also be saving some extra work which is always a bonus

This is a little bit of a continuation of @krlvi 's https://github.com/gitbutlerapp/gitbutler/pull/3708/files


Caleb-T-Owens:
Marked as a draft because I messed up using gitbutler's conflict resolution... I really want to make it do a rebase rather than a merge

PR #3723, opened 2024-05-07
verify head clean checks integration commit

krlvi:


crates/gitbutler-core/src/virtual_branches/integration.rs
```diff
@@ -404,3 +412,29 @@ fn verify_head_is_set(
...
+    commit.author().name().map_or(false, |name| {
+        name == GITBUTLER_INTEGRATION_COMMIT_AUTHOR_NAME
+    })
+}
+
+fn is_integration_commit_message(commit: &git::Commit) -> bool {
+    commit
+        .message()
+        .starts_with(b"GitButler Integration Commit")
+}
```
mtsgrd, line 440: What do you think about collapsing this into just one function? Feels like that would be easier to read.

crates/gitbutler-core/src/virtual_branches/integration.rs
```diff
@@ -314,9 +314,17 @@ fn verify_head_is_clean(
...
+        Some(integration_commit) => {
+            if !is_integration_commit(integration_commit) {
+                // There's nothing to do, dont error out?
+                return Ok(());
+            }
+        }
+        None => {
+            // There's nothing to do, dont error out?
+            return Ok(());
+        }
```
mtsgrd, line 327: I think we.. might want to return an error here? If you run any command using `verify_branch` and [we're not on the right branch](https://github.com/gitbutlerapp/gitbutler/pull/3718/files) then it would feel unwise to let the command continue.

krlvi: @mtsgrd Pavel is getting this error... in a situation i feel he shouldnt....

krlvi:
lets try the other fix on its own first

PR #3768, opened 2024-05-15
first pass at new oplog tree builder

schacon:
This PR changes the way we store the oplog snapshot data. It removes the need to look at the working directory at all, spares us from resetting the user's index, should be much faster, and preserves the commit data on virtual branches so they can be pushed in the oplog data without needing the rest of the project history.

The new format for the oplog snapshots looks like this. The commit itself has this type of data:

```
❯ git cat-file -p 42603177babb189d9d7caeb3384ae5ab78ccd902
tree 7f8d941752883e421038bc9dd5961c9bdad4053a
parent 2886017ff0a0b1d454c9fa7c97c215ae1c2be9cb
author GitButler <gitbutler@gitbutler.com> 1715798282 +0200
committer GitButler <gitbutler@gitbutler.com> 1715798282 +0200

CreateCommit

Version: 1
Operation: CreateCommit
message: Update work.html
sha: 71b75f34b293fc6bfc0e165689f463e27fb71b95
```

Most of the data is structured in trailers so we can present it in a pretty manner. Then it points at a tree. That tree of data has this basic structure:

```
. (7f8d9417528...)
├── (tree) conflicts              [data used when a working dir is in conflicted state]
├── (tree) index                  [the current index with snapshotting]
├── (tree) target_tree            [the target branch tree that the vbranches are based on]
├── (blob) virtual_branches.toml  [the current virtual branch and target state]
└── (tree) virtual_branches       [contains one entry per virtual branch to recreate state]
   └── (tree) d135191d-9481-40f2-8490-543034942881  [branch id]
      └── (tree) tree     [tree that the branch would commit next]
      └── (tree) commits  [data to recreate the commits on this vbranch]
         └── (tree) 70e368b630eb1ba7e1d2a16c43b60d20fdaef886   [the data for commit 70e368...]
           └── (blob) commit        [the actual raw commit data (cat-file -p style)]
           └── (tree) tree        [the tree that commit points to]
         └── (tree) 71b75f34b293fc6bfc0e165689f463e27fb71b95   [the data for commit 71b75f...]
           └── (blob) commit
           └── (tree) tree
```

Seems a little complicated, but it allows us to do some interesting things. There are lots of trees in here (mostly similar) that we can use to recreate the working directory, branch states and index very quickly (the main point). 

It also gives us the data to recreate all the virtual branches on another computer or checkout if we wanted to - it lets us push our commits in a way that doesn't pull the entire history along with it. We could push it to a git server that does not have the main repository, still allow us to reconstruct all the diffs, but _also_ allows us to reconstruct the actual git objects. 

For example, we can push the commit object `70e368b630` in this way, by putting it in our tree as a blob and then re-encoding it elsewhere back as a commit again, getting the same SHA:

```
❯ git cat-file -p 42603177babb189d9d7caeb3384ae5ab78ccd902^{tree}:virtual_branches/d135191d-9481-40f2-8490-543034942881/commits/70e368b630eb1ba7e1d2a16c43b60d20fdaef886/commit | git hash-object -t commit --stdin
70e368b630eb1ba7e1d2a16c43b60d20fdaef886
```


crates/gitbutler-core/src/ops/oplog.rs
```diff
@@ -73,16 +73,21 @@ impl Oplog for Project {
...
         let oplog_state = OplogHandle::new(&self.gb_dir());
         let oplog_head_commit = match oplog_state.get_oplog_head()? {
             Some(head_sha) => match repo.find_commit(git2::Oid::from_str(&head_sha)?) {
-                Ok(commit) => commit,
-                Err(_) => repo.find_commit(default_target_sha.into())?,
+                Ok(commit) => Some(commit),
+                Err(_) => None, // cant find the old one, start over
             },
-            // This is the first snapshot - use the default target as starting point
-            None => repo.find_commit(default_target_sha.into())?,
```
schacon, line 85: We don't want to do this, the way this was implemented, we would use the target sha as the first parent we build the oplog on top of. We would rather have the first oplog entry be parentless so we know when the history started.

crates/gitbutler-core/src/ops/oplog.rs
```diff
@@ -96,63 +101,91 @@ impl Oplog for Project {
...
-        // before the snapshot creation is initiated in the first place.
-        // However, there should be no conditions under which the integration commit is stale, and if there is, it should be fixed at the source.
-
-        // Reset the workdir to how it was
-        let integration_branch = repo
-            .find_branch("gitbutler/integration", git2::BranchType::Local)?
-            .get()
-            .peel_to_commit()?;
-
-        repo.reset(
```
schacon, line 152: No more resetting the repo 🎉

krlvi: thats awesome!

PR #3781, opened 2024-05-18
Use ScrollableContainer's bottomReached event to load more

Caleb-T-Owens:
Bit of a refactor over my async/await based gradual rendering.

It was still struggling if you had in the neigherbourhood of 9000 files, so this will only render as the user scrolls down.

There is still the issue of scrolling taking quite a long time, but it feels like the solution would need to be a floating commit button.

### DIFF
--- a/app/src/lib/components/BranchFilesList.svelte
+++ b/app/src/lib/components/BranchFilesList.svelte
@@ -4,7 +4,6 @@
 	import { getContext } from '$lib/utils/context';
 	import { selectFilesInList } from '$lib/utils/selectFilesInList';
 	import { maybeMoveSelection } from '$lib/utils/selection';
-	import { sleep } from '$lib/utils/sleep';
 	import { getCommitStore } from '$lib/vbranches/contexts';
 	import { FileIdSelection, fileKey } from '$lib/vbranches/fileIdSelection';
 	import { sortLikeFileTree } from '$lib/vbranches/filetree';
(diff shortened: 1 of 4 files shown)


Caleb-T-Owens:
@mtsgrd I've taken a look round at what the different methods of sending messages to children are, and in this case I thought the exported functions looked best, but I'm curious to know if you're aware of a better svelt-ey mechanism that I could use in this case.

I did consider an BehaviourSubject in a context, but we've generally tried to keep RX.js internal to our services

app/src/lib/components/BranchFilesList.svelte
```diff
@@ -27,23 +26,27 @@
...
-		for (let filesChunk of chunk(files, 100)) {
-			sortedFiles = [...sortedFiles, ...filesChunk];
-			await sleep(0);
-		}
+	// Make sure we display when the file list is reset
+	$: {
+		chunkedFiles = chunk(sortLikeFileTree(files), 100);
+		displayedFiles = chunkedFiles[0] || [];
+		currentDisplayIndex = 0;
 	}
```
mtsgrd: Not sure how you feel here, but I prefer making a function for these things (to make it evident what the change detection depends on).

mtsgrd:
A store could perhaps also work, but I like this solution, it feels like an appropriate and unsurprising way of doing it.

app/src/lib/components/BranchFilesList.svelte
Caleb-T-Owens: That makes quite a bit of sense as a pattern 👍

PR #3788, opened 2024-05-19
show local branches

schacon:


schacon:
Same as #3110, I should probably combine them.

PR #3789, opened 2024-05-19
setup for async tasks on oplog update

krlvi:
As we start building out code collaboration we wanna have the ability to push gitbutler state to, when enabled

### DIFF
--- a/crates/gitbutler-watcher/src/handler/mod.rs
+++ b/crates/gitbutler-watcher/src/handler/mod.rs
@@ -60,6 +60,11 @@ impl Handler {
                 .await
                 .context("failed to handle git file change event"),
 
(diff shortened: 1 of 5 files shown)


crates/gitbutler-watcher/src/handler/mod.rs
```diff
@@ -181,4 +186,8 @@ impl Handler {
         }
         Ok(())
     }
+    async fn gitbutler_oplog_change(&self, _project_id: ProjectId) -> Result<()> {
+        // TODO: Queue up pushing of data here, if configured
```
krlvi, line 190: @schacon we can initiate synchronization of the oplog from here

doing this in the watcher allows us it to be nice and non-blocking for the various commands.

I can look into adding some synchronization this week

PR #3790, opened 2024-05-19
Update IPC imports and remove old logic

Caleb-T-Owens:
I was taking a look into where certain invoke calls were coming from, and logging how long they took, so as part of that I made sure that all the calls went through our wrapper around `invoke`.

I've not included any of my extra logging as most of the same information can be gathered from the rust logs anyway.

I have however removed some extra code which appears to be now unused.


krlvi:
thx!!

PR #3802, opened 2024-05-21
Disallow UNC paths

anaisbetts:
Adding UNC paths (`\\some-server\share\some\path`) won't work, so let's make better messages telling users what to do


Caleb-T-Owens:
See conversation about file endings change: https://discord.com/channels/1060193121130000425/1204524097241878629/1242433110788800595

PR #3813, opened 2024-05-22
Fix timestamp compensation

Caleb-T-Owens:


crates/gitbutler-core/src/ops/oplog.rs
```diff
@@ -349,7 +349,7 @@ impl Oplog for Project {
                     lines_added,
                     lines_removed,
                     files_changed,
-                    created_at: commit.time().seconds() * 1000,
+                    created_at: commit.time().seconds(),
```
Byron, line 352: Since this structure probably ends up in the frontend, I'd think it would be even better if it was renamed to include the desired unit. 
The frontend seems to [turn the field into milliseconds](https://github.com/gitbutlerapp/gitbutler/blob/c699ad4dfaa32e710adf9dfbe23174c0be125bb4/app/src/lib/history/types.ts#L64), but the backend [thinks it's already milliseconds](https://github.com/gitbutlerapp/gitbutler/blob/93d5e61530910cfa9674b4de4eea3f60f86878e0/crates/gitbutler-core/src/ops/entry.rs#L21).

Probably it would be easiest here to fix the documentation in the backend, as the frontend typing should be safe enough.

Caleb-T-Owens: Hi @Byron, thanks for pointing out that docstring, I'll update it and have a think about renaming it too 👍

PR #3815, opened 2024-05-22
temp

Caleb-T-Owens:


Byron:
In case this was never meant to be merged here, I recommend creating a PR in your fork instead and enable CI there. That way, one will get exactly the same CI results as in the parent repository, but without triggering notifications.

Caleb-T-Owens:
Sorry @Byron. There was a question around whether the create PR button was working on the latest nightly today so I pushed up my branch to see whether it worked for me. I don't watch repositories so it didn't occur to me that it could be disruptive for others. I'll be more careful in future 👍

PR #3823, opened 2024-05-23
Refactor how we merge in commits from branch upstream

mtsgrd:
- send change_id to frontend for `RemoteCommit`
- associate upstream <-> local using change_id
- split up long "merge_virtual_branch_upstream" function
- add a couple of checks to prevent unexpected state
- rebase if force push allowed (we should let user choose merge over rebase still)

### DIFF
--- a/crates/gitbutler-core/src/git/error.rs
+++ b/crates/gitbutler-core/src/git/error.rs
@@ -1,6 +1,9 @@
 use std::str::Utf8Error;
 
-use crate::keys;
+use crate::{
+    error::{Context, ErrorWithContext},
+    keys,
+};
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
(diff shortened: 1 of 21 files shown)


crates/gitbutler-core/src/git/error.rs
```diff
@@ -49,6 +52,12 @@ impl From<git2::Error> for Error {
     }
 }
 
+impl ErrorWithContext for Error {
```
Byron, line 55: An empty implementation shouldn't be required. It's interesting you found the need anyway.

mtsgrd: Thanks for checking my work 🙌 

The reason I added that was because it felt cumbersome to manually convert context-free `git:Error` using `Error::from_err`, as mentioned in this comment.
https://github.com/gitbutlerapp/gitbutler/blob/f6421aab3c9ea36944fce6d690602ef2586280f6/crates/gitbutler-core/src/error.rs#L112-L121

Byron: Ah, I see, thank you. It's the auto-conversion you were after.
It likely that there is no way around it though, but I hope that now using `thiserror` most of the time is the solution, ultimately.
I will watch out for this when trying to refactor error handling.

PR #3826, opened 2024-05-23
upgrade `gix` to v0.63

Byron:
It comes with various bugfixes and security improvements.
Admittedly, none of these are currently relevant for GitButler.

It's still a good idea to upgrade as `cargo deny` would have problems with older versions due to the security advisories that were published yesterday.

### Tasks

* [x] upgrade `gix`
* [x] better isolate tests against developer global options (uboot)
* [x] assure boolean query works as expected

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -137,9 +137,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle-wincon"
-version = "3.0.3"
+version = "3.0.2"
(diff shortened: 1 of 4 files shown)


Byron:
@krlvi I am bold and self-merged this one as it was very simple where 'nothing could possibly go wrong'. Please let me know if you'd rather not have me do that even with simple PRs.

PR #3827, opened 2024-05-23
improve watcher performance

Byron:
Do less IO in the watcher when processing events, and speed it up by using `gitoxide`.

### Tasks

* [x] don't open Repository unless needed
* [x] use `gitoxide` for faster ignored checks (and faster opening of repositories).
* [x] multi-dir watching

### Notes for the Reviewer

It's also worth noting that the multi-dir watching isn't tested, and being able to do that will probably fail code elsewhere as generally, everything assume that there is a worktree-dir, and in there is a `.git/` directory, even though it could be a file or not be present at all (Git is (too) flexible).

So I think it's better to reject it for now, which means https://github.com/gitbutlerapp/gitbutler/pull/3827/commits/850045e9d30b1b50691ead12c48107edfd09e5d7 should be reverted.


crates/gitbutler-watcher/src/file_monitor.rs
```diff
@@ -116,28 +118,34 @@ pub fn spawn(
...
+                        .collect();
+                    if classified_file_paths
+                        .iter()
+                        .any(|(_, kind)| *kind == FileKind::Project)
+                    {
+                        if let Ok(repo) = gix::open(&worktree_path) {
+                            if let Ok(index) = repo.index_or_empty() {
+                                if let Ok(mut excludes) = repo.excludes(&index, None, gix::worktree::stack::state::ignore::Source::WorktreeThenIdMappingIfNotSkipped) {
+                                    for (file_path, kind) in classified_file_paths.iter_mut() {
+                                        if let Ok(relative_path) = file_path.strip_prefix(&worktree_path) {
```
Byron, line 155: This isn't quite as pretty as I would hope with all the nesting, but due to ownership I don't see another way. Except if one would have a dummy-loop around it to break out of, but it seems a bit hacky so I didn't go there.

crates/gitbutler-watcher/src/file_monitor.rs
```diff
@@ -116,28 +118,34 @@ pub fn spawn(
...
+                                            if excludes.at_path(relative_path, None).map(|platform| platform.is_excluded()).unwrap_or(false) {
```
Byron, line 156: Note that the `None` here means that there is no information about the file type, so it internally treats it like a file. However, some of these paths may be directories, and right now we tend to just pass them along. 

I checked, and there seems to be no quick way to filter them out, as the filesystem events don't provide type information (or I didn't find it).

krlvi (approved):

PR #3830, opened 2024-05-23
Replace integer timestamps with Duration

Caleb-T-Owens:
This PR aims to remove ambiguity around what style of timestamp might be stored in integer columns by using an unambiguous type


crates/gitbutler-core/src/git/commit.rs
```diff
@@ -55,8 +57,8 @@ impl<'repo> Commit<'repo> {
         self.commit.parent(n).map(Into::into).map_err(Into::into)
     }
 
-    pub fn time(&self) -> git2::Time {
-        self.commit.time()
+    pub fn time(&self) -> Duration {
```
Byron, line 60: The `git2::Time` is clearly specified as seconds since unix epoch in UTC and a local-time offset, even though the `time()` method here doesn't make that very obvious.

However, now there is a `time()` method that returns a `Duration` which is *implicitly* relative to the UNIX epoch and implicitly UTC, and all that without a doc-string to clarify this.

I strongly recommend to undo this so people working with `commit.time()` have more type-safety.

(An improvement here might have been to let it return `chrono` times as it's then impossible to mishandle it, but that would probably have made the downstream adaptation more difficult)

crates/gitbutler-core/src/ops/oplog.rs
```diff
@@ -349,7 +350,7 @@ impl Oplog for Project {
                     lines_added,
                     lines_removed,
                     files_changed,
-                    created_at: commit.time().seconds(),
+                    created_at: Duration::from_secs(commit.time().seconds().try_into().unwrap()),
```
Byron, line 353: What started as a strong recommendation now becomes a request as this is incorrect. `time().seconds()` is `i64` which is now cast into `u64`. It's a feature of `git2` (and `gitoxide`) to be able to represent commit-dates older than 1970 to allow the creation of 'historic' repositories (used for legal-texts apparently), and this `try_into()` will panic there.

crates/gitbutler-core/src/ops/oplog.rs
```diff
@@ -363,7 +364,7 @@ impl Oplog for Project {
                     lines_added: 0,
                     lines_removed: 0,
                     files_changed: Vec::new(), // Fix: Change 0 to an empty vector
-                    created_at: commit.time().seconds(),
+                    created_at: Duration::from_secs(commit.time().seconds().try_into().unwrap()),
```
Byron, line 367: See above.

crates/gitbutler-core/src/virtual_branches/base.rs
```diff
@@ -31,7 +34,7 @@ pub struct BaseBranch {
     pub behind: usize,
     pub upstream_commits: Vec<RemoteCommit>,
     pub recent_commits: Vec<RemoteCommit>,
-    pub last_fetched_ms: Option<u128>,
+    pub last_fetched_at: Option<Duration>,
```
Byron, line 37: Again, I don't think that a `Duration` is the right abstraction for a timestamp, even though I *feel* that it's better than a u128. Ideally it could just be a timestamp, like `std::time::Instant` or `std::time::SystemTime` (all of which are UTC which seems to be all that's needed here). Then, when translating it for use in the frontend, the necessary conversions would have to be done.

crates/gitbutler-core/src/virtual_branches/branch/mod.rs
```diff
@@ -25,16 +27,10 @@ pub struct Branch {
...
-        serialize_with = "serialize_u128",
-        deserialize_with = "deserialize_u128"
-    )]
-    pub created_timestamp_ms: u128,
-    #[serde(
-        serialize_with = "serialize_u128",
-        deserialize_with = "deserialize_u128"
-    )]
-    pub updated_timestamp_ms: u128,
+    #[serde(rename = "created_timestamp_ms", with = "duration_int_string_serde")]
```
Byron, line 30: I think it's a great idea, and can be a template. Instead of putting the unit into the field name on the Rust side, rely on strong (and correct) typing. Then during translation, bake the unit into the field so the frontend has an easier time with units.

crates/gitbutler-core/src/virtual_branches/remote.rs
```diff
@@ -129,12 +129,7 @@ pub fn branch_to_remote_branch(branch: &git::Branch) -> Result<Option<RemoteBran
                         None
                     },
                     name,
-                    last_commit_timestamp_ms: commit
-                        .time()
-                        .seconds()
-                        .try_into()
-                        .map(|t: u128| t * 1000)
-                        .ok(),
+                    last_commit_at: Some(commit.time()),
```
Byron, line 132: It looks like the desired type of `last_commit_at` is `git2::Time`, then it would match naturally at least my concern about `Duration` would be addressed.

crates/gitbutler-core/src/virtual_branches/virtual.rs
```diff
@@ -138,7 +138,7 @@ pub struct VirtualBranchHunk {
     pub id: String,
     #[serde(serialize_with = "crate::serde::as_string_lossy")]
     pub diff: BString,
-    pub modified_at: u128,
+    pub modified_at: Duration,
```
Byron, line 141: Maybe `Duration` is convenient because it naturally translates into a number, maybe seconds with sub-second precision?
Nonetheless I think the Rust-side should keep what makes sense and is lossless, and then transform during serialization.

crates/gitbutler-core/src/virtual_branches/virtual.rs
```diff
@@ -1968,7 +1968,11 @@ fn virtual_hunks_into_virtual_files(
             let conflicted =
                 conflicts::is_conflicting(project_repository, Some(&id)).unwrap_or(false);
             let binary = hunks.iter().any(|h| h.binary);
-            let modified_at = hunks.iter().map(|h| h.modified_at).max().unwrap_or(0);
+            let modified_at = hunks
+                .iter()
+                .map(|h| h.modified_at)
+                .max()
+                .unwrap_or(Duration::ZERO);
```
Byron, line 1975: I like this constant - typically I end up with `Duration::default()`, but this is more expressive and works in `const` contexts.

Byron (commented):
Thanks for giving this a shot, and I think particularly the 'hand-off' from backend to frontend is the way to do it, i.e. use strong types in the backend, then translate to fieldname-with-unit for the frontend, and just a POD.

However, I also noticed issues that I believe need fixing. Thanks for your consideration.

Caleb-T-Owens:
@Byron, thanks for going through. I would certainly like to use the same type for all timestamps, and have something that is easy to manipulate.

Duration seemed to fit that criteria, but I now realise is semantically incorrect.

It seems like the chrono DateTime is firstly, semantically correct, and also appears to handles time before time. If its alright with you, I'll change over to DateTime in the morning

crates/gitbutler-core/src/virtual_branches/branch/mod.rs
Caleb-T-Owens: This particular timestamp is getting persisted into the `virtual_branches.toml` rather than getting sent to the frontend.

I quite like the way Duration was maintaining the same attribute name (so there is less ambiguity between what you see on the frontend and the backend), but provided an object with meaningful keys.

Byron:
> It seems like the chrono DateTime is firstly, semantically correct, and also appears to handles time before time. If its alright with you, I'll change over to DateTime in the morning

I'd advise against it mostly for reasons of increased complexity that happens with Chrono - `git2::Time` is simple and easy to understand, in comparison, and it matches perfectly as long the origin of the time is Git.

If the origin is not Git, then all we (seem) to need is `SystemTime`, which can easily be converted into `git2::Time` with an offset of 0.

Having UTC ties *seems* to be OK for internal calculations, while the offset is only relevant for presentation in the UI maybe. I don't know enough about that though.

Please note that I once again checked how `chrono` is used and it's all for obtaining the current time in milliseconds, probably relative to the unix epoch. My stance here is to remove it rather than proliferate it.

PR #3832, opened 2024-05-23
experiment with jj style conflicted commits

schacon:
This PR is an experiment with performing our rebasing operation in a similar manner to how jj does it, where instead of stopping, we record the state of the conflict in the commit tree and headers, then keep going.


crates/gitbutler-core/src/git/repository.rs
```diff
@@ -117,6 +117,33 @@ impl Repository {
...
     }
 
+    // find the real tree of a commit, which is the tree of the commit if it's not in a conflicted state
+    // or the parent parent tree if it is in a conflicted state
+    pub fn find_real_tree(
+        &self,
+        commit: &git2::Commit,
+        side: Option<String>,
+    ) -> Result<git2::Tree> {
+        let tree = commit.tree()?;
```
schacon, line 127: we should instead check commit header

PR #3840, opened 2024-05-23
Use DateTime over duration

Caleb-T-Owens:


Byron (changes requested):
In [my comment](https://github.com/gitbutlerapp/gitbutler/pull/3830#issuecomment-2128867486) which certainly was a bit late I strongly advised against proliferating `chrono`.
Instead, it should be no problem to use `git2::Time`, and indeed these [can be constructed at will](https://docs.rs/git2/latest/git2/struct.Time.html#method.new).

`chrono` can be great if time manipulation is complex, but I don't think we are doing that.

Of course, I won't be in your way if you are feeling strongly about `chrono`, it's not that it wouldn't work. It's just that it's not necessary, and too complex for what we do with time.

PR #3842, opened 2024-05-23
Remove "Current" tag and show restore button on all snapshots

krlvi:
Because snapshots are now created _before_ actions, the one on the top of the list is also undo-able


krlvi:
@PavelLaptev FYI

PavelLaptev:
thanks, I'll check

PR #3846, opened 2024-05-24
Return the merge result in the case the rebase bails

schacon:
We changed the logic to where if a rebase fails it just bails, which can be because there is a merge commit in the branch. In this case, if we were even trying a rebase, it's because the merge worked. So if the `cherry_rebase()` bails, just do what we did before, which is return the merge result.


krlvi:
works

PR #3847, opened 2024-05-24
Revert "Replace integer timestamps with Duration"

Caleb-T-Owens:
Reverts gitbutlerapp/gitbutler#3830

@Byron raised some valid concerns about using Duration. It seems like the best solution would be git2::Time for most of the git-related timestamps, and SystemTime in places where we just want to measure the current time.

For now I'm going to revert back to using timestamps so we don't have any concerns about errors due to negative timestamps


crates/gitbutler-core/src/ops/entry.rs
```diff
@@ -19,8 +18,8 @@ use serde::Serialize;
 pub struct Snapshot {
     /// The sha of the commit that represents the snapshot
     pub id: String,
-    /// Snapshot creation time as a duration since the Unix epoch
-    pub created_at: Duration,
+    /// Snapshot creation time in epoch seconds
+    pub created_at: i64,
```
Byron, line 22: Even though `i64` is better than `Duration` in terms of the range of values it can express, it's not better for type-safety.
However, it's definitely an improvement and maybe one day there will be another pass that, in Rust, uses a type to get it right.

Maybe this is because some files were reverted?
What confuses me is that this field is `created_at`, but further down there was a rename from `created_at` to `create_timestamp_ms`, where `_ms` is the kind of thing I would have avoided by using actual types. A `SystemTime` for instance can express sub-second precision whereas `git2::Time` cannot - this might be a way to decide which one to use.

crates/gitbutler-core/src/virtual_branches/virtual.rs
```diff
@@ -66,7 +66,7 @@ pub struct VirtualBranch {
     pub upstream_name: Option<String>, // the upstream branch where this branch will push to on next push
     pub base_current: bool, // is this vbranch based on the current base branch? if false, this needs to be manually merged with conflicts
     pub ownership: BranchOwnershipClaims,
-    pub updated_at: Duration,
+    pub updated_at: u128,
```
Byron, line 69: I assume that these values are for dates we don't get from a commit, so one would expect people to have their clocks in the current time.

But then again, I wonder why risk it, and not use a type that can handle any time a system could use.
Maybe it's about serialization here? But if it was, couldn't that be handled in the code the serializes and deserializes?

Ah, I understand, this was also reverted to the original type.

crates/gitbutler-core/src/virtual_branches/virtual.rs
```diff
@@ -1112,8 +1112,8 @@ pub fn create_virtual_branch(
         upstream_head: None,
         tree: tree.id(),
         head: default_target.sha,
-        created_at: now,
-        updated_at: now,
+        created_timestamp_ms: now,
+        updated_timestamp_ms: now,
```
Byron, line 1116: What I liked previously was that the backend was allowed to have proper types, and thus didn't need to embed the type as suffix in the field name while resorting to a POD.
Then during serialization, the transformation can happen, including the rename of the field to something that is easier to handle in the frontend.

This seems now gone, but probably the file was reverted to what it was before.

Byron (commented):
Thanks a lot! I think the primary concern is addressed, even though the codebase as a whole isn't yet consistent in how it handles timestamps.

Probably this will happen over time.

Byron:
Thank you, that sounds very promising!

PR #3850, opened 2024-05-24
Watcher adjustments

Byron:
This PR contains the outcome of changes made when reviewing the `watcher` crate

### Tasks

* [x] add license information for copied files
* [x] refactor
* [x] add tests
* [x] remove unused `FileIdMap` implementation


Byron:
Hi @anaisbetts, I just wanted to give you a heads-up that I am trying to re-add tests - right now they all fail but I think it has a chance to work once I bring `roots` back. These have been moved to the file-id map and *seem* to be unused.
As I will be AFK for a moment, maybe that's a chance to let me know why that's a bad idea 😅. Thank you.

krlvi (approved):
Thanks! i suggest we merge this and make modifications accordingly after the fact if needed

Byron:
Agreed, I had the same in mind.

anaisbetts:
Did you test this on Windows? This looks like it took the code from master rather than the last tag of debounce that we were using - when I used this version of the debouncer from master, it completely broke the app

PR #3854, opened 2024-05-24
Improve usage of remotes that have been added by the user

Caleb-T-Owens:


app/src/lib/branches/types.ts
```diff
@@ -115,9 +109,9 @@ export class CombinedBranch {
 	}
 
 	currentState(): BranchState | undefined {
-		if (this.vbranch) return BranchState.VirtualBranch;
 		if (this.pr) return BranchState.PR;
 		if (this.remoteBranch) return BranchState.RemoteBranch;
+		if (this.vbranch) return BranchState.VirtualBranch;
```
Caleb-T-Owens, line 114: @PavelLaptev I've reordered these if statements because it seemed off to me, but I wanted to get your thoughts.

Reordering these means that if a vbranch or remoteBranch has an associated pull request, in the sidebar it will show the PR icon rather than the remoteBrach icon or vbranch icon.

Example:

This is an unapplied vbranch which has a corresponding remote branch and a corresponding  PR; previously it would display the vbranch icon; now it displays the green PR symbol as displayed below:
Screenshot 2024-05-24 at 17 13 13

app/src/lib/branches/service.ts
```diff
@@ -134,33 +128,18 @@ function mergeBranchesAndPrs(
...
 					return new CombinedBranch({ pr });
 				})
 		);
 	}
 
-	// deduplicate - don't show branches that point to the same SHA
-	// prioritize PRs, then remote branches, then local only
-	const deduped = contributions.filter(
-		(value, index, self) => self.findIndex((v) => v.sha === value.sha) === index
-	);
```
Caleb-T-Owens, line 152: We no longer need to dedup because we're doing a better job of checking which branches are already accounted for

PavelLaptev:
Yes, it make sense. Thank you 👍

app/src/lib/branches/types.ts
PavelLaptev: Yes, it make sense. Thank you 👍

PR #3859, opened 2024-05-25
minor Windows adjustments

Byron:
This PR is a side-effect of getting local VM builds to work, specifically x86-64 builds on ARM Windows VMs (running on Macs via Parallels).

Besides some cleanup, there is one commit that adjusts a script so that it's easier to use in these situations. Finally, there now is documentation for how to crosscompile.

Note that despite producing a release binary, the build still fails on the final rename from `gitbutler-tauri.exe` to `GitButler Nightly.exe`, but that's not fixed here nor considered a problem.

### Notes for the Reviewer

The script change *might* break the Windows build even though it shouldn't.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -20,6 +20,7 @@ you right. Let's get started.
     - [Building on Windows](#building-on-windows)
     - [File permissions](#file-permissions)
     - [Perl](#perl)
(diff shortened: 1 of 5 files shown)


krlvi (approved):

krlvi:
tested the build script in ci, should be fine

PR #3864, opened 2024-05-25
first stab at resurecting git sync

krlvi:


krlvi:
@schacon syncing seems to work once again. 
You gotta be an admin and have it enabled in the project settings (we can soon move this to a proper feature flag)

It's pushing all refs as well as it forcefully pushed the oplog head commit to `gitbutler/oplog/oplog`.
This is mostly legacy on the server end but `git_url` is no longer a thing, that used to be the seesions gb_repository.
`code_git_url` is the one with all the refs, including the oplog.

I tested this and seems to be pushing everything as expected. Performance wise, i think this may be slower than expected (uses the libgit push), but it is done in a non-blocking way so it should be fine. This also probably doesnt work on windows.

PR #3867, opened 2024-05-26
oplog review (part 1)

Byron:
This PR contains changes made as side-effect of the oplog review. It's also meant as a platform to discuss possible changes and why I'd make them.

#### Avoid 'fuzzy' types when a concrete one would do ✔️

* `get_opelog_head()` returns a `String` even though that String contains a hexsha. We have a type for that, `git2::Oid` which would prevent misuse.
* In the same vein, `list_snapshots()` takes `String`, which should be `&str` if it's the correct type. But it should be `git2::Oid`,
  and it seems that the frontend loose typing infected the Rust code which should always try to use the typesystem to its advantage.

#### Only make `pub` on `crate` level if the item is used in another crate ✔️

That way, unused items will be pointed out by the compiler.

#### Put tests for public items on integration level ✔️

That way, it becomes clear if there are errors in the intended public API pretty quickly. Personally, I like that this reduces the size of the code files as tests are neatly separated.

#### Avoid traits just for the sake of it✔️

A trait should be used if there are multiple implementations for it, or if it's used as extension to a type (usually not under your control). Any other use is confusing at least to me as I expect one of the above reasons.

#### Can the usage of oplog be more abstract? 

Right now it's hand-rolled prepare-commit steps everywhere, could that be more 'automatic'?
TBD❗️

#### Avoid PODs ✔️

If there is a type that describes date better than PODs, then definitely use it. Don't use `i64` if you could use `git2::Time`, or `String` if it could be `git2::Oid`. Units quickly get messy if they are not strongly typed.

#### On `core.git`

* When encountering `core::git::*` and all its wrapped types I was wondering if this is going to be a god-sent or a burden. Time will tell.
* The `git` wrapped types don't seem to be used much/consistently within `ops`, which makes it a potential burden.
* Somehow I think that it's easier to go with `git2` while it's there as long as repositories aren't kept around but opened mostly when needed. This is what I have seen. I have also seen a simple API wrapper for configuration handling (get/set) and though that such improvements could be done with so that we'd have `git2::Repository::open()?.config().easy().use_easy_api()` - this is just a general idea. Right now I can't tell, but I will keep an eye out for that.
* I chose not to replace `git2` usage with `git` except for in the API (where `git::Oid` is used instead of `git2::Oid`) until I know what's the right choice here.

#### On worktree-handling

Due to variable naming, it's not always clear what `Project::path` is. It's a worktree-directory, and whenever it's used like `project.path.join(".git/gitbutler")` it's a problem that will prevent GB from dealing with ALL the repository configurations out there as `.git` doesn't have to be a directory, nor does it have to be there at all. The only way to handle this correctly is to open a repo with `gix::open()` and then use the `git_dir()` and `worktree_dir()` methods respectively to get the actual directories. These do have to be treated separately, but it's definitely a good idea start from a working-tree directory as it matches with the single-worktree paradigm of GB.

When the is a `repo` already, definitely use its `path()` (which is its `git_dir`), instead of doing `wd.join(".git/gitbutler/...")`.

#### Avoid `sha` in variable names

As everything should be strongly typed, as opposed to stringly typed, by the type it's already clear that it's a hash. What's not usually clear is what the hash points to. If I read `target_sha` I don't know much, especially because `target` is GB lingo for `trunk to merge with`. Instead, I'd call it `target_commit`, which tells me something important that's not obvious from the type otherwise.

#### Don't predict the compiler

Sometimes code feels like it tries to anticipate what the compiler would say, even though the most direct (or intuitive) way of doing things would actually work. It's usually easiest to try first, and then make it work if needed. Sometimes, it just works.

#### Don't repeat module-names in test-functions

That way, they read well both on the command-line and in IDEs.

Screenshot 2024-05-28 at 09 45 35

From there, it's also easier to find more descriptive names.

### Tasks

* [x] instrument oplog creation so we learn about ignored errors
* [x] improve integration test structure and fix commit signing
* **detailed review of**
    * [x] `entry.rs`
    * [x] `snapshot.rs`
    * [ ] `oplog.rs`
    * [x] `reflog.rs`
* [x] structural changes: tests for pub APIs are integration level
* [x] show how a trait-free implementation looks like (it will be simpler with less code/boilerplate)
* [x] use `git2::Oid` instead of string-hex-hashes
* [x] review API in `rustdoc`
* [ ] answer remaining questions

### What I liked

* `SnapshotDetails` makes great use of types, and `version` is strongly typed as well.
* `oplog.rs` documentation - I think there should be more of that as it helps communicating the idea. Of course, docs can rot easily, but I'd pay the price of admission.
* types in `entry.rs` have documentation

### Questions and my answers

These questions came up by @krlvi 

#### "is there something that can go catastrophically wrong (e.g. invocations from the watcher)"

This question is from times where the watcher integration wasn't implemented yet if I remember correctly, and now the watcher only synchronizes if it sees an oplog change (and this is enabled and allowed).

So I think it will be fine even though it will be racy - it's likely that while a sync is ongoing, the next sync will fail, so the latest oplog state won't be available on the server. Right now that's just a guess though.

####  "the code ignores snapshot errors, is this the right way?"

It seems reasonable to call the oplog optional, so it shouldn't prevent anything else from functioning. Over time, it should ideally 'never fail'. And if it does, I will make sure that gets logged at least to allow improving it.

#### "does this implementation has logical errors / stuff that i missed or overlooked"

I can't promise I find these, but I will be extra-diligent and try to think beyond API and Rust.


Caleb-T-Owens:
> "the code ignores snapshot errors, is this the right way?"
It seems reasonable to call the oplog optional, so it shouldn't prevent anything else from functioning. Over time, it should ideally 'never fail'. And if it does, I will make sure that gets logged at least to allow improving it.

@krlvi 

If we're using snapshots to back up GitButler operations which may fail in a way that looses code, shouldn't we cancel the whole operation if taking that backup fails?

krlvi:
> > "the code ignores snapshot errors, is this the right way?"
> > It seems reasonable to call the oplog optional, so it shouldn't prevent anything else from functioning. Over time, it should ideally 'never fail'. And if it does, I will make sure that gets logged at least to allow improving it.
> 
> @krlvi
> 
> If we're using snapshots to back up GitButler operations which may fail in a way that looses code, shouldn't we cancel the whole operation if taking that backup fails?

You know, this may be a good idea. My main thinking initially here was to minimize the risk of creating regressions. Once we feel that we are happy with the snapshotting, we could indeed swap it to a failing thing, such that if we cant create a tree before the operation we don't do the operation.

Maybe this is unrelated, but something that I find interesting is the idea of adding a an "error" trailer to the snapshot commit - this way, if/when something fails we can do interesting things - if we know that something has side effects we can auto undo. And even if we dont do that, it would be nice to have it in the oplog for various debug reasons.

crates/gitbutler-core/src/ops/reflog.rs
```diff
@@ -20,7 +20,7 @@ use crate::projects::Project;
 /// <target branch head sha>                 <oplog head sha>
 ///
 /// The reflog entry is continuously updated to refer to the current target and oplog head commits.
-pub fn set_reference_to_oplog(
+pub(crate) fn set_reference_to_oplog(
```
krlvi: this is cool i wasnt familiar with this syntax. The `reflog` is not exported in ops/mod.rs - Is there a way to make it visible only within the ops module? I guess it's not important, im just curious what's possible

crates/gitbutler-core/src/ops/oplog.rs
```diff
@@ -34,60 +34,11 @@ const SNAPSHOT_FILE_LIMIT_BYTES: u64 = 32 * 1024 * 1024;
 /// │       ├── commit-message.txt
 /// │       └── tree (subtree)
 /// └── virtual_branches.toml
-pub trait Oplog {
+impl Project {
```
krlvi, line 46: woah, this is awesome, i didnt know u can do multiple, separate impl blocks

crates/gitbutler-core/src/ops/reflog.rs
Byron: Yes, there even is a syntax for that named `pub(in <module>)`. This I use quite rarely and instead go with `pub` while possibly putting it into a `mod` (private)` to prevent it to leak. Now that I say that, the `pub(crate)` portions could probably use another round, this time with `cargo doc` to help visualize what's visible to see if it *should* be visible.

crates/gitbutler-core/src/ops/reflog.rs
```diff
@@ -35,9 +38,14 @@ pub fn set_reference_to_oplog(
         .join("target");
 
     if !reflog_file_path.exists() {
-        let repo = git2::Repository::init(repo_path)?;
-        let commit = repo.find_commit(git2::Oid::from_str(target_head_sha)?)?;
-        repo.branch("gitbutler/target", &commit, false)?;
+        let mut repo = gix::open(repo_path)?;
+        repo.refs.write_reflog = gix::refs::store::WriteReflog::Always;
```
Byron, line 54: Due to [`core.logAllRefUpdates`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corelogAllRefUpdates) it's needed to enforce this setting. With `git2` I can't override git-configuration in memory or I don't know how, so I used `gix`.

crates/gitbutler-core/src/ops/reflog.rs
```diff
@@ -88,11 +102,13 @@ fn set_oplog_ref(file_path: &PathBuf, sha: &str) -> Result<()> {
...
 mod tests {
     use super::*;
+    use gix::Repository;
+    use std::str::FromStr;
     use tempfile::tempdir;
 
     #[test]
-    fn test_set_target_ref() {
-        let (dir, commit_id) = setup_repo();
+    fn test_set_target_ref() -> anyhow::Result<()> {
```
Byron: Even though I typically start with `.unwrap()` while it's unclear if things work or not, I finalize tests by moving them to `?` from `.unwrap()` so they are more readable. Most tests are 'read-only` after writing them, so it's helpful to leave them in the most readable state.

Byron (commented):
I think I found a bug in the reflog writer as it wrote reflog lines that were invalid. As the `gitoxide` parser is modelled after the one in Git it's just as unforgiving and did choke on the missing `\t`. In the worst case, this would have meant that `git gc` also wouldn't have managed to read what it needed to in order keep it alive, but I also didn't check if it does anything special or is more forgiving.

For now I hacked a fix together, but I think it will be more robust to use the `gix` parser here as well to rebuild the lines with `\t` from the beginning.

TBC

krlvi:
@Byron i found the `core::git::*` types to be a burden when doing anything in the codebase (mostly forwarding / wrapping calls). Especially when needing something that is not yet in our wrapper.

While building the oplog module I decided to skip them altogether to see how it feels, and now I am leaning towards getting rid of the proxied bits at least. We still need a good pattern for building generic functionality that is close to git2/oxide (a good example is perhaps the function `cherry_rebase`), but the proxying and duplicated types is probably something that we can get rid of (i was actually planning to axe it some time next week)

Byron:
Thanks for sharing! I wonder if there is anything in the wrapper that is worth retaining, or in other words, the reason for its existence. Ideally these parts can be kept, and if it's about convenience, they shouldn't be less convenient afterwards.
However, I'd think that it either boils down to utility functions or extension traits, to probably mirror everything it offers without wrapping everything.

So, it's good to hear these are going away :).

crates/gitbutler-core/src/ops/entry.rs
```diff
@@ -17,16 +18,18 @@ use serde::Serialize;
...
 pub struct Snapshot {
     /// The sha of the commit that represents the snapshot
-    pub id: String,
-    /// Snapshot creation time in epoch seconds
-    pub created_at: i64,
+    #[serde(with = "crate::serde::oid")]
+    pub id: git::Oid,
+    /// Snapshot creation time in seconds from Unix epoch seconds, based on a Git commit
+    #[serde(serialize_with = "crate::serde::as_time_seconds_from_unix_epoch")]
+    pub created_at: git2::Time,
```
krlvi: i really like this approach - it's clear in rust what it was, and we get the thing serialize correctly automatically :) 
Gonna copy this pattern in other places

crates/gitbutler-core/src/ops/oplog.rs
```diff
@@ -218,44 +158,50 @@ impl Oplog for Project {
...
-            let base_tree = repo.find_tree(target_tree_oid)?;
+            let mut workdir_tree_id = target_tree_id;
+            let base_tree = repo.find_tree(target_tree_id)?;
             let mut current_ours = base_tree.clone();
 
-            let head_trees_iter = head_trees.iter();
             // iterate through all head trees
-            for head_tree in head_trees_iter {
-                let current_theirs = repo.find_tree(git2::Oid::from(*head_tree))?;
+            // TODO: This needs a test. Right now only the last merged index is actually used.
```
Byron, line 178: I think what follows is incorrect as in comparison to the version of this in `integration.rs`, this one doesn't connect the previous merge index with the current iteration of the loop. That way, the last merge wins.

While the version in `integration.rs` has tests failing when the loop is messed with, I can neuter this one entirely and it's all green.

Next step here is to add a test that catches this issue.

crates/gitbutler-core/src/ops/oplog.rs
```diff
@@ -33,80 +39,27 @@ const SNAPSHOT_FILE_LIMIT_BYTES: u64 = 32 * 1024 * 1024;
...
+    pub(crate) fn prepare_snapshot(&self) -> Result<git::Oid> {
+        let worktree_dir = self.path.as_path();
+        let repo = git2::Repository::open(worktree_dir)?;
 
         let vb_state = self.virtual_branches();
 
-        // grab the target tree sha
-        let default_target_sha = vb_state.get_default_target()?.sha;
-        let default_target = repo.find_commit(default_target_sha.into())?;
-        let target_tree_oid = default_target.tree_id();
```
Byron, line 98: While having mixed `oid` and `id` in the `gitoxide` codebase as well, here it seems best to align with `tree_id()` and call it `id`. `gitoxide` also uses the `id` suffix on the highest level, i.e. in `gix`.

crates/gitbutler-core/src/ops/oplog.rs
krlvi: Here we are reconstructing the full working directory state from all of the applied virtual branches merged together. Per definition applied virtual branches are assumed to not conflict with each other (because they are applied, i.e. in the working directory).

Is your comment about the condition in which the merge is not clean or did i not understand it correctly? Maybe there is still something we can do here to code it more defensively

Byron: Screenshot 2024-05-27 at 19 00 20

The `workdir_tree_id` is completely decoupled from the input of the merge, which means that only the last pair of trees is actually written.

Now I will try to prove it by having a test that validates the produced tree.

Compared with `integration.rs`, where the previously merged worktree is used in the next merge.

Screenshot 2024-05-27 at 19 06 51

----

Something else I was thinking is that the worktree can be reproduced from state which is stored in the snapshot - only the `is_applied()` information would also be needed.

Then one could save the time to merge everything together and instead do that work only when a restore is requested. And there, maybe existing vbranch functionality can be used, after all a restore would restore the vbranches, and from there it should be possible to trigger an 'apply' of sorts. But… this is nothing more than a flash of an idea, nothing for now or even near-term.

schacon: I'm a little confused, unless I'm misreading this. Isn't `current_ours` being used the same as `workspace_tree` in the second example?

schacon:
> As everything should be strongly typed, as opposed to stringly typed, by the type it's already clear that it's a hash. What's not usually clear is what the hash points to. If I read target_sha I don't know much, especially because target is GB lingo for trunk to merge with. Instead, I'd call it target_commit, which tells me something important that's not obvious from the type otherwise.

In many cases, we need to get the sha or oid (sometimes it's different - usually I use `_sha` for a string and `_oid` for an Oid type), because then I need to get the object from the odb into another variable. So for example it's not helpful to do `target_commit` to hold an oid and then `target_commit` to hold the commit object, I would rather have the code be a little clearer in doubling up the type a little.

Also, if you're not using an editor that decorates the types (for example, GitHub review diffs), then it could be unclear in reading the code.

Anyhow, I'm not much of a Rust expert so maybe I'm wrong, but I'm just not sure that making the var names less clear assuming we can rely on type decoration is a better approach.

crates/gitbutler-core/src/ops/oplog.rs
Byron: `current_ours` is overwritten with `current_theirs` each iteration, giving it the effect of a sliding window, i.e. `merge(base, a, b)`, then `merge(base, b, c`. The problem seems to be that `merge()` returns an index which, despite being turned into a tree, is entirely ignored on all iterations except the last one.

In the second example, the procedure is different, as it's `merge(base, base, a) -> merge_result1`, then `merge(base, merge_result1, b) -> merge_result_2` and so forth, thus previous merge is actually used.

Maybe you are able to validate oplog workspace restore with multiple applied vbranches real quick? Maybe it works and I am misreading this entirely.

But in any case, I think there should be a test for that.

Byron:
> Anyhow, I'm not much of a Rust expert so maybe I'm wrong, but I'm just not sure that making the var names less clear assuming we can rely on type decoration is a better approach.

I don't think it needs a Rust expert, variable names are hard and what matters is alignment and consistency. To me it seems reasonable to merge both suggestions and go with `target_commit_id` to make it clear it's an `id` to a `commit` of the `target`, and from what I can tell the code does that already *(if it doesn't, it's an oversight)*. So `target_commit` does indeed point to the object, not the id.

The main suggestion left here from my side is to not re-invent a new name for `id` as it's used in the `git2` API (see `tree.id()` or `commit.tree_id()`, but use it if it's the same. Hence `_sha` becomes `_id`, which is also how `gitoxide` does it so it will match. As a fun fact: `gitoxide` does it like this because `git2` does it.

I am quite aware that `_sha` or `sha` is used everywhere and I could understand if it should just stay like this - my thinking here is to perform these renames or clarifications once code is touched, and not spend time doing it all at once, so it's a step-wise process towards more descriptive variable and field names, which I think is something everybody wants.

krlvi:
I think im guilty of renaming things that the API calls ids into _sha in a bunch of places in the code, and I really have no good explanation for why i've done it, but seems reasonable to not invent a new name

crates/gitbutler-core/src/ops/oplog.rs
Byron: https://github.com/gitbutlerapp/gitbutler/assets/63622/a9db410e-1366-4de9-ad58-f8988463b389


Here is a video showing that only the last two virtual branches will be merged into the workdir state. Now for a test that allows to fix that.

krlvi: okay so definitely a bug! its great that you caught this one before we launched the new release :)

Byron (commented):
This is Part 1 of the review, and the oplog is missing a test for the worktree merge (which is in progress), and from there will happen a review of everything in `oplog` along with even more tests, probably.

So far, two critical bugs found, one fixed.

I will continue with Part 2 on top of this branch, so no rush with merging.

crates/gitbutler-core/src/ops/entry.rs
```diff
@@ -3,30 +3,33 @@ use anyhow::Result;
...
 #[derive(Debug, PartialEq, Clone, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Snapshot {
-    /// The sha of the commit that represents the snapshot
-    pub id: String,
-    /// Snapshot creation time in epoch seconds
-    pub created_at: i64,
+    /// The id of the commit that represents the snapshot
+    #[serde(rename = "id", with = "crate::serde::oid")]
+    pub commit_id: git::Oid,
```
krlvi, line 22: the UI code assumes this field is called `id`, but i can fix that in a follow up pr

krlvi (approved):

crates/gitbutler-core/src/ops/entry.rs
krlvi: oh, never mind, i missed the rename

Byron: Admittedly, I should probably have renamed the UI part instead of doing it in the (for me) simpler way.
Maybe the UI also benefits from the clearer name.

PR #3873, opened 2024-05-27
skip signing of the integration commit

krlvi:


crates/gitbutler-core/src/git/repository.rs
```diff
@@ -245,6 +245,29 @@ impl Repository {
         self.0.blob(data).map(Into::into).map_err(Into::into)
     }
 
+    pub fn plain_commit(
```
Byron: What's the difference between `commit` and `plain_commit`? With growing team size, even the most sparing essence of a doc-string will help, and I think it's worth investing in knowledge transfer.

Personally, I see `core` as library that everyone should be able to grok and use, hence some form of documentation should exist. This has been done in the `oplog` and I think it's the way to go for newly added code at least.

krlvi: this was intended as a quick test a potential fix related to gpg commit signing (right now we sign the integration commit and we dont need to)

Byron:
```suggestion
    pub fn commit_without_signature(
```

This would keep it more similar to `commit()`, while making clear the difference even without docs.

krlvi: instead of making a new function i will start removing the usage of the wrapping git types that dont abstract anything but only forward calls

PR #3877, opened 2024-05-28
No snapshot if commit fails

mtsgrd:


krlvi (changes requested):
I am not sure this is the behavior we want here. We would still wanna have the snapshot, but ideally, capture the error in it (which is the part that is not ready yet).

What i was thinking we can add the error info as a trailer, and then show it in the UI as a failed operation. This can be especially valuable if we have commands that fail and leave side effects even though they failed, this way at least its visible and possible to go back

Byron:
I can clearly visualize this now (after #3867) and love the idea! It shouldn't be be a problem either. I'd probably try to implement it as `project.run_with_oplog(|| <something-fallible>)?`. The way the code is structured - everything is shareable - that won't be a problem with borrowchk.

I'd love to see this implemented - please feel free to do it while part 2 of the review is in progress, the merge-conflicts shouldn't be a problem if there are any.

PR #3879, opened 2024-05-28
Fix slightly confusing code related to merging in trunk

mtsgrd:
- easier to understand if we only set branch.tree in one place
- enable ? return for git errors


crates/gitbutler-core/src/virtual_branches/errors.rs
```diff
@@ -142,11 +142,13 @@ pub enum ApplyBranchError {
     Conflict(ProjectConflict),
     #[error("branch not found")]
     BranchNotFound(BranchNotFound),
-    #[error("branch conflicts with other branches - sorry bro.")]
```
krlvi, line 145: hahah :D

PR #3880, opened 2024-05-28
Add a "apply from remote" flow

Caleb-T-Owens:


Caleb-T-Owens:
Having talked to kiril, I'm going to change branch & pr matching over to a tuple of [owner, repo name, branch name] after this branch

crates/gitbutler-core/src/project_repository/repository.rs
```diff
@@ -611,6 +611,12 @@ impl Repository {
     pub fn remotes(&self) -> Result<Vec<String>> {
         self.git_repository.remotes().map_err(anyhow::Error::from)
     }
+
+    pub fn add_remote(&self, name: &str, url: &str) -> Result<()> {
+        self.git_repository
```
Byron, line 616: Another common way of writing this is `Ok(self.git_repository.add_remote(name, url)?)`, which is usually preferred as it turns into a one-liner. Once it breaks the line, the form used here is usually nicer to look at.

crates/gitbutler-core/src/remotes/controller.rs
```diff
@@ -0,0 +1,35 @@
...
+pub struct Controller {
+    projects: projects::Controller,
+}
+
+impl Controller {
+    pub fn new(projects: projects::Controller) -> Self {
+        Self { projects }
+    }
+
+    pub async fn remotes(&self, project_id: &ProjectId) -> Result<Vec<String>, Error> {
```
Byron, line 17: `&ProjectId` should be `ProjectId` as it's `Copy`.

crates/gitbutler-core/tests/suite/virtual_branches/create_virtual_branch_from_branch.rs
```diff
@@ -103,7 +103,7 @@ async fn integration() {
     {
         // should mark commits as integrated
         controller
-            .fetch_from_target(project_id, None)
+            .fetch_from_remotes(project_id, None)
```
Byron, line 106: I love this rename, `remote` is a word I understand. Probably `target` for GB ultimately is only that, even though I don't know if there are some (subtle) differences yet.

Byron (commented):
Hi, I left a few general and minor Rust notes.

crates/gitbutler-core/src/remotes/controller.rs
Caleb-T-Owens: Hey! Sorry, it looks like you've said `&ProjectId` twice. Could you clarify what you mean perhaps by adding a suggestion block?

crates/gitbutler-core/src/project_repository/repository.rs
Caleb-T-Owens: I did look at `Ok(<expression>?)` when writing but it felt super jarring in terms of control flow perspective. I'll keep that wrapping heuristic though and make a change tomorrow morning :+1:

crates/gitbutler-core/src/remotes/controller.rs
Byron: Apologies, I meant `Project`, without ampersand. I fixed the above as well.

PR #3884, opened 2024-05-28
move get_wd_tree to git2::Repository extension trait

krlvi:
This way we reduce drilling and allow ourselves to eventually remove core::git types


krlvi:
okay this is not all but merging this, and will morph this code in small steps to avoid conflicts

PR #3885, opened 2024-05-29
Tweaks to remotes code

Caleb-T-Owens:


crates/gitbutler-core/src/remotes/controller.rs
```diff
@@ -14,22 +14,22 @@ impl Controller {
         Self { projects }
     }
 
-    pub async fn remotes(&self, project_id: &ProjectId) -> Result<Vec<String>, Error> {
-        let project = self.projects.get(project_id)?;
+    pub async fn remotes(&self, project_id: ProjectId) -> Result<Vec<String>, Error> {
+        let project = self.projects.get(&project_id)?;
```
Caleb-T-Owens: @Byron I changed the argument to be type `ProjectId` for the functions I wrote, but it seems like `&ProjectId` is fairly deeply rooted throughout the application. Do we need to change these to just `ProjectId` at some point?
Screenshot 2024-05-29 at 09 55 01

Byron: Yes, it's definitely up for a refactor. My stance here is that ideally we don't proliferate certain issues, and fix the ones we find during feature development or reviews.

But fixing all at once seems easy enough if you feel like it :).

Caleb-T-Owens: > My stance here is that ideally we don't proliferate certain issues

100% Agree with you there

I'm still picking up rust best practices so I very much appreciate your reviews 👍

Caleb-T-Owens:
@Byron, I've gone and replaced &ProjectId with ProjectId and &BranchId with BranchId. Would you mind doing a quick check to make sure I've not done anything particularly daft?

In the tests, there is the pattern of:
```
let Test {
  project_id,
  ..
} = &Test::default();
```

which means project_id is of type `&ProjectId`. I looked at making the call just `Test::default()` (with no borrow), but that seemed to cause even more trouble, so I opted to dereference `project_id` when it was passed to a function in the test cases

crates/gitbutler-tauri/src/watcher.rs
```diff
@@ -103,10 +103,9 @@ impl Watchers {
     pub fn watch(&self, project: &projects::Project) -> Result<()> {
         let handler = handler_from_app(&self.app_handle)?;
 
-        let project_id = project.id;
```
Byron, line 106: That's a nice catch!

Byron (approved):
Wow, you have deleted a lot of `&` today - thanks so much for these fixes :)!

> I looked at making the call just `Test::default()` (with no borrow), but that seemed to cause even more trouble, so I opted to dereference `project_id` when it was passed to a function in the test cases

Yes, that's the way here. The `&Test::default()` is needed here and very deliberate, but that does mean one has to deref the `ProjectId`, effectively making a copy.

Let me merge this right away for a smaller chance of conflicts.

PR #3886, opened 2024-05-29
fix: cannot close file card from Trunk's commit history

lisiur:
CleanShot 2024-05-29 at 15 49 35

Currently, we cannot close the file card on the right side by clicking the 'Close' button or
canceling the selection of files. This PR fixes the issue.

### DIFF
--- a/app/src/routes/[projectId]/base/+page.svelte
+++ b/app/src/routes/[projectId]/base/+page.svelte
@@ -61,13 +61,14 @@
 			/>
 		</div>
 		<div class="base__right">
-			{#if selected}
+			{#if selected && $fileIdSelection.length > 0}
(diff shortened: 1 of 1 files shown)


Caleb-T-Owens:
Hey @lisiur, thanks for the contribution!! Its very much appreciated

It looks to me like we've got two objects which are trying to fulfil the a very similar role which feels very awkward. I'm going to talk to the others about why we have both, and if one of them can get removed as I'd much rather fix the root cause rather than adding more logic on top of it

lisiur:
Sorry, I'm not very familiar with svelte, so I'm afraid I won't be able to do this. I will open an issue to track and learn how to handle it.
For now, I'll close this PR.

PR #3887, opened 2024-05-29
remove git::tree

krlvi:


crates/gitbutler-core/src/git/mod.rs
```diff
@@ -42,3 +42,6 @@ pub use self::url::*;
 
 mod repository_ext;
 pub use repository_ext::*;
+
+mod tree_ext;
+pub use tree_ext::*;
```
Byron, line 47: I consider it more maintainable to not use `*`, but spell out the few types that are made available here. `*` is really just for `super::*` in test-mods in my book.
It's nothing Rusty in particular, but a convention I also know from Python.

Byron (commented):
-577 lines of code, I am loving these PRs :)!

PR #3888, opened 2024-05-29
Introduce BranchIdentifier for matching vbranches, remotes, and PRs

Caleb-T-Owens:


crates/gitbutler-core/src/virtual_branches/remote.rs
```diff
@@ -122,34 +126,57 @@ pub fn branch_to_remote_branch(branch: &git::Branch) -> Result<Option<RemoteBran
...
+
+    let Some(sha) = branch.target() else {
+        tracing::warn!(
+            "ignoring branch {:?} as target could not be found",
+            branch.name()
+        );
+        return Ok(None);
+    };
+
+    println!("{:?}", name);
```
Byron, line 151: It's probably better use `dbg!()` for this, and if not (because of the added formatting) `eprintln!()` can be used to get text to `stderr` without buffering.

Caleb-T-Owens: That is super neat!

Byron: One thing that can be confusing is the `dbg!()` moves its arguments, as opposed to `print/eprint` which reference their arguments. This is so that you can just wrap everything in `dbg!()` and it will work like it did before.

So instead of separately adding a line that says `dbg!(foo)`, you'd want to do `mycall(dbg!(foo))` in place and it will act normally.

Caleb-T-Owens: Yes! I saw that in the documentation. Super handy. I actually ended up using it this evening to peek inside some expressions.

Caleb-T-Owens:
This PR has been superseded by other PRs

PR #3889, opened 2024-05-29
remove-commit-reference

krlvi:


crates/gitbutler-core/src/git/commit_ext.rs
```diff
@@ -0,0 +1,46 @@
...
+        let cid = self.header_field_bytes("change-id").ok()?;
+        if cid.is_empty() {
+            None
+        } else {
+            // convert the Buf to a string
+            let ch_id = std::str::from_utf8(&cid).ok()?.to_owned();
+            Some(ch_id)
+        }
+    }
+    fn parents_gb(&self) -> Result<Vec<Commit<'repo>>> {
```
krlvi, line 32: fyi @Byron i struggled with naming this... It cant be parents because it clashes with the `git2::Commit.parents` method...
Is there some convention? But also, what is the point of this method.... perhaps we just use the one from git2 that returns an iteratior

crates/gitbutler-core/src/git/commit_ext.rs
```diff
@@ -0,0 +1,46 @@
...
+        for i in 0..self.parent_count() {
+            parents.push(self.parent(i)?);
+        }
+        Ok(parents)
+    }
+    fn is_signed(&self) -> bool {
+        let cid = self.header_field_bytes("gpgsig").ok();
+        cid.is_some()
+    }
+    fn tree_gb(&self) -> Result<git2::Tree<'repo>> {
```
krlvi, line 43: Same as the one above... seems like this exists to fit the error type, which is odd, but to keep the refactor simple, i left it as it was (with just a unique name now)

crates/gitbutler-core/src/git/commit_ext.rs
Byron: Yes, definitely, there is no need for this one. An iterator is better here. People who want a Vec do `commit.parents().collect()`.

crates/gitbutler-core/src/git/commit_ext.rs
Byron: Yes, I'd remove this as well.

Byron (commented):
Still ~30 lines removed, and removal of abstractions. Lovely!

PR #3899, opened 2024-05-30
Fix invalid commit range in blame operation

mtsgrd:
- fixes commit_to_branch related errors
- fixes big performance problem when force push disallowed

### DIFF
--- a/crates/gitbutler-core/src/virtual_branches/integration.rs
+++ b/crates/gitbutler-core/src/virtual_branches/integration.rs
@@ -1,4 +1,4 @@
-use std::path::PathBuf;
+use std::{path::PathBuf, vec};
 
 use anyhow::{anyhow, Context, Result};
 use bstr::ByteSlice;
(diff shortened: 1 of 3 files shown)


crates/gitbutler-core/src/virtual_branches/virtual.rs
```diff
@@ -1695,6 +1778,9 @@ fn get_applied_status(
     target_sha: &git::Oid,
     mut virtual_branches: Vec<branch::Branch>,
 ) -> Result<(AppliedStatuses, Vec<diff::FileDiff>)> {
+    dbg!(project_repository
```
krlvi: we dont wanna merge this

crates/gitbutler-core/src/virtual_branches/integration.rs
```diff
@@ -85,14 +85,28 @@ pub fn get_workspace_head(
...
+        .filter_map(Result::ok)
+        .collect();
+
+    if heads.is_empty() {
+        heads = vec![target_commit]
+    }
+
+    // TODO: Why does commit only accept a slice of commits? Feels like we
+    // could make use of AsRef with the right traits.
+    let head_refs: Vec<&git2::Commit<'_>> = heads.iter().collect();
```
Byron, line 101: This clearly is a place where `git2` has a very inconvenient API, as a slice of refs nobody has naturally, ever. It's also a strange choice given that it needs parent-ids, not commit objects.
This goes away with `gitoxide`.

crates/gitbutler-core/src/virtual_branches/virtual.rs
```diff
@@ -1687,6 +1688,84 @@ fn get_non_applied_status(
         .collect::<Result<Vec<_>>>()
 }
 
+fn compute_merge_base(
+    project_repository: &project_repository::Repository,
+    integration_commit: &git::Oid,
```
Byron, line 1693: This type is `Copy` and thus shouldn't be referenced, just like (now) `ProjectId`.

crates/gitbutler-core/src/virtual_branches/integration.rs
Byron, line 101: I think you can even write this:
```Rust
    let head_refs: Vec<&_> = heads.iter().collect();
```

Byron (commented):
I took a quick look and have some general Rust notes.

Some of the code that moved (and was altered) looked like it's more complex than it needs to be, but I since it's in `virtual.rs` its time will come eventually :).

PR #3900, opened 2024-05-30
Show PR checks progress as error if any check has failed

mtsgrd:


PavelLaptev:
@mtsgrd how does it look like?

PR #3901, opened 2024-05-30
fixes reflog timestamps

krlvi:
Time stapms are super imporatant for this (thanks @schacon) - without a correct time stamp, the trees will be GC'd

### DIFF
--- a/crates/gitbutler-core/src/ops/reflog.rs
+++ b/crates/gitbutler-core/src/ops/reflog.rs
@@ -86,7 +86,7 @@ fn standard_signature() -> gix::actor::SignatureRef<'static> {
     gix::actor::SignatureRef {
         name: GITBUTLER_INTEGRATION_COMMIT_AUTHOR_NAME.into(),
         email: GITBUTLER_INTEGRATION_COMMIT_AUTHOR_EMAIL.into(),
-        time: Default::default(),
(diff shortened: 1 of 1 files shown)


Byron:
That'a a great catch and sorry for not catching this myself!

[Over in my latest PR I have added a test for this](https://github.com/gitbutlerapp/gitbutler/pull/3898/files) and could verify that it indeed fails before this fix, but works after.

crates/gitbutler-core/src/ops/reflog.rs
```diff
@@ -86,7 +86,7 @@ fn standard_signature() -> gix::actor::SignatureRef<'static> {
     gix::actor::SignatureRef {
         name: GITBUTLER_INTEGRATION_COMMIT_AUTHOR_NAME.into(),
         email: GITBUTLER_INTEGRATION_COMMIT_AUTHOR_EMAIL.into(),
-        time: Default::default(),
+        time: gix::date::Time::now_local_or_utc(),
```
Byron, line 89: It's good that we don't care, but as a word of warning, unless the application is single-threaded or sets `time::set_soundness(Unsound)`, the timezone offset won't be available.

This is as `time` still relies on libc for this, and it isn't threadsafe. This will be an issue [until this was implemented](https://github.com/time-rs/time/issues/193).

PR #3905, opened 2024-05-30
handle merge commits

krlvi:


krlvi:
@PavelLaptev i am so sorry about this design 🙈

crates/gitbutler-core/src/virtual_branches/remote.rs
```diff
@@ -187,12 +188,20 @@ pub fn branch_to_remote_branch_data(
 }
 
 pub fn commit_to_remote_commit(commit: &git2::Commit) -> RemoteCommit {
+    let parent_ids: Vec<git::Oid> = commit
+        .parents()
+        .map(|c| {
+            let c: git::Oid = c.id().into();
```
Byron, line 194: Instead, I'd do `git::Oid::from(c.id())` and that's it :).

PR #3920, opened 2024-05-30
snapshots save errors

krlvi:


crates/gitbutler-core/src/ops/snapshot.rs
```diff
@@ -11,49 +12,63 @@ use super::entry::Trailer;
...
-        let details =
-            SnapshotDetails::new(OperationKind::ApplyBranch).with_trailers(vec![Trailer {
-                key: "name".to_string(),
-                value: branch_name,
-            }]);
-        self.create_snapshot(details)?;
+    pub(crate) fn snapshot_branch_applied(
+        &self,
+        snapshot_tree: git::Oid,
+        result: Result<&String, &Error>,
```
Byron, line 18: Even though this looks strange I see why it's done, and very much much akin to `Option<&T>` being preferred over `&Option<T>`. it's a first time for me :).

crates/gitbutler-core/src/ops/snapshot.rs
```diff
@@ -146,3 +161,33 @@ impl Project {
         Ok(())
     }
 }
+
+fn result_trailer(result: Result<Option<String>, &Error>, key: String) -> Vec<Trailer> {
+    match result {
+        Ok(v) => {
```
Byron, line 167: It will be more concise if it was matched on `Ok(Some(v))` and `Ok(None)`.

crates/gitbutler-core/src/ops/snapshot.rs
```diff
@@ -146,3 +161,33 @@ impl Project {
...
+            if let Some(v) = v {
+                vec![Trailer {
+                    key,
+                    value: v.clone(),
+                }]
+            } else {
+                vec![]
+            }
+        }
+        Err(error) => vec![Trailer {
```
Byron, line 177: I all codebases that I remember I see `err` here (instead of `error`) and I keep renaming these out of habit. However, it's also a matter of convention and we can have our own here, please let me know if I should stop renaming these to `err`.

PR #3927, opened 2024-05-31
Use notify-debouncer-full v0.3.1

anaisbetts:
The master version of notify-debouncer-full seems to have issues, let's roll back to the last public release, but keep the tests and better organization of #3850 

## TODO:

- [x] Get the app working
  - [x] Windows
  - [x] macOS
  - [x] Linux
- [x] Bring back the tests


krlvi (approved):
LFG

PR #3948, opened 2024-06-02
watcher refactor

Byron:
This PR is a follow-up on #3927 and is more or less similar to #3850.
However, this time, we'd want to be extra-sure (and extra-careful) to not end up with a performance regression.

### Tasks

* [x] remove test duplication and unused files
* [x] make 'flush-changes' visible
    - [x] bring back original code and make it work with the app
    - [x] make sure it works with the original version
* [x] remove dead code warnings by making the debouncer code a crate again
* [x] ~~double-check there are no side-effects for performance~~ - I actually never noticed these :/

### Notes

* This time around I realized why there was so much confusion around this. For me it all started with not being able to see what was imported code, and what were changes. Doing each one step at a time (something that is reproduced here) really helps to understand.
    - From there came the realization how `FileIdCache` is supposed to be used, and that the dead-code came from the code not being a crate anymore.
    - Leaving imported crates as is really helps with this, and I made this an example by vendoring the code.
* The commit https://github.com/gitbutlerapp/gitbutler/pull/3948/commits/7d91d60bc0fc4fa67fa4f6020c02c1d4f69d7044 works just like the crate would, but it's easy to have it seemingly not work as the timeout of 60s needs to be more like 1s for the original version, and also the debouncer must be kept alive in the main polling task. If one of these aren't accommodated, the app will either never respond, or only once a minute.

### Notes for the Reviewer

* Each commit goes one step towards the final result, and it should be helpful to follow them along to understand what happened here better.
* Having the `debouncer` crate in `vendored` is something I made up as it seems fitting for code mostly taken from elsewhere. If we were to stick to the status quo, it could also be living in `./crates` on the top-level of course (*it just felt it's getting too much significance that way*).
* Please do recheck if the performance issues are gone. I made one change related to rename-tracking which might affect performance on systems that actually use it. If so, that can just be removed and replaced with `NoCache` for file-ids.


Byron:
Could you also take a look, @anaisbetts?
This works for me and I think this is the refactor that it should have been the first time 😅.
Thank you.

krlvi (changes requested):
it appears that this re-introduces the debouncer overhead

Byron:
Thanks for testing this! I will undo one change that I think could be responsible (rename tracking), and if you leave some "Steps to reproduce" I could validate it myself. Thanks again.

Byron:
I have rebased onto master and in the one new commit tried out to use no `FileIdCache`, which is what would have been the case with the previous version as well.

Thus I'd hope that the performance issue is fixed now, but I couldn't validate it yet for lack of a good test-case to reproduce it, and I didn't want to try blindly but look forward to your guidance.

Byron:
Thanks @krlvi for sharing how you tested this. Based on that, my test is to merely change the project from `rgit` (tiny) to `jgit` (2700 files) in debug mode.

This is `master`
Screenshot 2024-06-16 at 10 10 05

And this is the branch here, with fix:

Screenshot 2024-06-16 at 10 10 12

Further, I checked what happens if 1600 files are changed at once by means of `git checkout @~2090 -- .` and `git reset --hard @` respectively.  The debouncer did not show up.

Screenshot 2024-06-16 at 10 14 25

However, I definitely suggest to try it yourself in case my testing still didn't quite trigger it the way it should have (even though I think that 1.6k immediate worktree changes should work for that).

Byron:
For good measure, I also tried that (in debug mode) with the GitLab repository with the code on this branch. There the logs indicate, after causing a change of 11139 files with `git checkout @~2090 -- .` that it took 1.2s to process 5.6k paths in 22k filesystem events.

```
2024-06-16T08:25:35.127075Z DEBUG crates/gitbutler-watcher/vendor/debouncer/src/lib.rs:613: Flushed 22301 events
2024-06-16T08:25:36.369613Z  INFO set_project_active:watch:file monitor:handle debounced events: crates/gitbutler-watcher/src/file_monitor.rs:121: close time.busy=1.24s time.idle=35.0µs id=b31820bc-be4e-4f8b-ac5a-137ade109c1e project_id=b31820bc-be4e-4f8b-ac5a-137ade109c1e fs_events=22301 fs_events=22301 ignored=6 ignored=6 git_noop=3 git_noop=3 git=1 git=1 project=5861 project=5861 git_dedup=1 git_dedup=1 project_dedup=5861 project_dedup=5861
```

In release mode this could be 5x to 10x faster, so I'd think that this branch is OK now.

Here is the profile of this operation, and in the 2 minutes it was busy there is no mention of the debouncer.

Screenshot 2024-06-16 at 10 33 05

Of course it's slightly strange that `git status | wc -l` yields 11139 where the numbers in the debouncer indicated 5.6k, but that's something for another day, assuming the best.

krlvi (approved):
thanks for taking care of this. i did some testing and seems like the perf regression is gone

Byron:
Awesome!

This means the regression was caused by the `FileIdMap`, which previously was in the 'type', but was never configured at runtime to work. I 'fixed' this, and thus caused the issue. Now the 'type' contains `NoCache` explicitly to make clear FileIDs aren't cached.

All in all, I feel the watcher is now much better understood.

Issue #3646, opened 2024-04-30
Please add support for VSCodium and allow for external editors


dangreene0:
I am simply requesting for the ability to select other code editors besides VSCode.

Thanks.

krlvi:
Hey @dangreene0 
Thanks to this pr https://github.com/gitbutlerapp/gitbutler/pull/3958 by **[TheGB0077](https://github.com/TheGB0077)** the app should now be able to open in vscodium. For other editors we would need to do some more work.
Could you test out the latest release and see if it works for you?

dangreene0:
**Confirmed** to work on Linux Mint 21.3, however it says `VSCode` in GitButler but will still prompt to open `VSCodium` in the OS. This is probably unintentional but is worth noting.

Byron:
Thanks for sharing! Admittedly, I never thought of changing the label as well, and the current solution definitely doesn't attempt that. If memory serves, it's handled in the backend as well so I don't know how easy it would be to do that.

From a UX perspective, I think it's less confusing to show what would be opened. However, I also think that when we go there, we should also disable the menu item if no vscode/vscodium are available at all, as we'd have to be proactively checking for available editors then.

PR #3958, opened 2024-06-02
feat: Open with VS Code also opens VS Codium if available, partial fix for #3646

TheGB0077:
I initially implemented this looking through the PATH variable. On Linux and Mac this should be good enough, but I'm not sure if it works on Windows out of the box, for that reason, VS Code remains as the Fallback if VS Codium is not found.

Suggestions to improve, if necessary, are welcome!!

Also, there are many open() shell invocations, and this PR should also take this into account, but for now, I did the basic only.

### DIFF
--- a/crates/gitbutler-tauri/src/menu.rs
+++ b/crates/gitbutler-tauri/src/menu.rs
@@ -49,13 +49,10 @@ pub fn resolve_vscode_variant() -> &'static str {
 
 fn check_if_installed(executable_name: &str) -> bool {
     match env::var_os("PATH") {
-        Some(env_path) => env::split_paths(&env_path)
-            .filter_map(|mut path| {
-                path.push(executable_name);
-                fs::metadata(path).ok()
-            })
-            .next()
-            .is_some(),
+        Some(env_path) => env::split_paths(&env_path).any(|mut path| {
+            path.push(executable_name);
+            fs::metadata(path).is_ok()
+        }),
         None => false,
     }
 }
--- a/crates/gitbutler-tauri/src/menu.rs
+++ b/crates/gitbutler-tauri/src/menu.rs
@@ -48,16 +48,16 @@ pub fn resolve_vscode_variant() -> &'static str {
 }
 
 fn check_if_installed(executable_name: &str) -> bool {
-    let paths = match env::var_os("PATH") {
-        Some(path) => env::split_paths(&path).collect::<Vec<_>>(),
-        None => vec![],
-    };
-
-    paths.into_iter().any(|mut path| {
-        path.push(executable_name);
-        //println!("{}", path.to_string_lossy());
-        fs::metadata(path).is_ok()
-    })
+    match env::var_os("PATH") {
+        Some(env_path) => env::split_paths(&env_path)
+            .filter_map(|mut path| {
+                path.push(executable_name);
+                fs::metadata(path).ok()
+            })
+            .next()
+            .is_some(),
+        None => false,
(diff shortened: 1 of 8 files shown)


crates/gitbutler-tauri/src/menu.rs
```diff
@@ -34,6 +36,30 @@ pub async fn menu_item_set_enabled(
...
+    if vscodium_installed {
+        "vscodium"
+    } else {
+        // Fallback to vscode, as it was the previous behavior
+        "vscode"
+    }
+}
+
+fn check_if_installed(executable_name: &str) -> bool {
+    let paths = match env::var_os("PATH") {
```
Byron: This could better be one of these new let-else bindings :):

```Rust
let Some(paths) = env::var_os("PATH").map(|p| env::split_paths(&p).collect::<Vec<_>>()) else { return false }
```

crates/gitbutler-tauri/src/menu.rs
```diff
@@ -34,6 +36,30 @@ pub async fn menu_item_set_enabled(
...
+        Some(path) => env::split_paths(&path).collect::<Vec<_>>(),
+        None => vec![],
```
Byron: But maybe it's prettier and more 'reasonable' to bail out early here (as opposed to going with `let-else`.
```suggestion
        None => return false,
```

crates/gitbutler-tauri/src/menu.rs
```diff
@@ -34,6 +36,30 @@ pub async fn menu_item_set_enabled(
...
+    };
+
+    paths.into_iter().any(|mut path| {
```
Byron: That's a neat way to (try) to avoid allocations and re-use what's there.

Byron (commented):
Thanks a lot for contributing! That's a great idea!

My comments aren't even suggestions, but something that could be used if you like it.

crates/gitbutler-tauri/src/menu.rs
TheGB0077: I see. Will do just that.

crates/gitbutler-tauri/src/menu.rs
```diff
@@ -34,6 +36,30 @@ pub async fn menu_item_set_enabled(
...
+}
+
+fn check_if_installed(executable_name: &str) -> bool {
+    match env::var_os("PATH") {
+        Some(env_path) => env::split_paths(&env_path)
+            .filter_map(|mut path| {
+                path.push(executable_name);
+                fs::metadata(path).ok()
+            })
+            .next()
```
Byron: Instead of `next().some()`, the `filter_map()` should be `any()`.

Byron (commented):
Thanks a lot!

With the following fixed (see below), I'd think it's ready for merge. However, I'd want to be sure @krlvi agrees with making this change.

crates/gitbutler-tauri/src/menu.rs
TheGB0077: Ok! I thought this change would make it a bit easier to parse through, but I'll revert it :)

TheGB0077:
> Thanks a lot!
> 
> With the following fixed (see below), I'd think it's ready for merge. However, I'd want to be sure @krlvi agrees with making this change.

Changes made! I also checked on Windows and saw that VSCodium is added to the PATH by default, unless the user opts out. So I guess it should be fine.

Byron (approved):
LGTM! Thanks again!

PR #3959, opened 2024-06-02
remove unnecessary checks in oplog commit

krlvi:
This little manouver costed us 150 milliseconds

image


Byron:
Here is a 'wacky' idea: `prepare_commit()` has to be blocking as it needs to take in all the state.

But since we ignore the outcome of `commit()`, it can be asynchronous. What if the whole `commit()` would be done *in the background*, including the no-op detection, and mark the no-op-yness in as part of the commit message?

That way the oplog entry is there, but the UI can just hide it.

PR #3960, opened 2024-06-02
Remove likly useless conditions

Caleb-T-Owens:
CC. @krlvi and @mtsgrd


krlvi:
this uses quite a bit of wall time to compute in some conditions... lets merge and keep an eye if anything breaks

PR #3968, opened 2024-06-03
Services: Improved base branch handling

Caleb-T-Owens:
Added a constant for the error message when no default target is set and updated the error handling to use this new message.

### DIFF
--- a/app/src/lib/vbranches/baseBranch.ts
+++ b/app/src/lib/vbranches/baseBranch.ts
@@ -73,7 +73,7 @@ export class BaseBranchService {
 				action: action || 'auto'
 			});
 		} catch (err: any) {
-			if (err.message?.includes('does not have a default target')) {
+			if (err.code == Code.DefaultTargetNotFound) {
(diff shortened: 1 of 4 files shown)


app/src/lib/vbranches/baseBranch.ts
```diff
@@ -20,6 +20,9 @@ import type { Readable } from 'svelte/store';
 
 export class NoDefaultTarget extends Error {}
 
+// Lowercase string of the error message returned from the backend when no target is set.
+const NO_TARGET_SET_ERROR_MESSAGE = 'there is no default target';
```
krlvi: i think you can use error code!

Caleb-T-Owens: It is an `errors.unknown`... let me see if I can add a new code in the rust side

Caleb-T-Owens: That was super simple to introduce!

PR #3969, opened 2024-06-03
handle sign failure

krlvi:
Okay the changes here are: 
- The app will look for a git config `gitbutler.signCommits` to determine if it should attempt to perform signing (as opposed to `gpgsign = false`)
- Adds a config to toggle the signing option on and off
- If there is an error during signing, provides a detailed toast with [a link to the docs](https://docs.gitbutler.com/features/virtual-branches/verifying-commits) explaining how to set up signing
image
Screenshot 2024-06-03 at 21 47 46

### DIFF
--- a/crates/gitbutler-core/src/git/repository_ext.rs
+++ b/crates/gitbutler-core/src/git/repository_ext.rs
@@ -1,7 +1,9 @@
-use anyhow::{bail, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use git2::{BlameOptions, Repository, Tree};
 use std::{path::Path, process::Stdio, str};
 use tracing::instrument;
+
+use crate::{config::CFG_SIGN_COMMITS, error::Code};
 
 use super::Refname;
 use std::io::Write;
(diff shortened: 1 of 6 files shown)


crates/gitbutler-core/src/git/repository_ext.rs
```diff
@@ -79,8 +79,21 @@ impl RepositoryExt for Repository {
...
-        let oid = do_commit_buffer(self, commit_buffer)?;
+        let should_sign = self.config()?.get_bool("commit.gpgSign").unwrap_or(false);
 
+        let oid = if should_sign {
+            match sign_buffer(self, &commit_buffer) {
+                Ok(signature) => self.commit_signed(&commit_buffer, &signature, None)?,
+                Err(e) => {
+                    tracing::warn!("Failed to sign commit, committing unsigned: {}", e);
+                    self.odb()?
+                        .write(git2::ObjectType::Commit, commit_buffer.as_bytes())?
```
schacon: The only problem I can see immediately is that it's not going to tell the user that it's failing or why it's failing.
It just won't sign things if it's misconfigured.

Byron:
Note that the `error.message` is `Errors.commit.signing_failed` is wrong - it's supposed to be the root cause, never the marker (i.e. `Code`). There are tests for this, but apparently something is off. Taking a look now.

PR #3974, opened 2024-06-03
fix: cleanup Toggle event handling

ndom91:
Small bug I ran into while familiarizing myself with the app

- Refactor `<Toggle />` to just handle click instead of custom `change` events. 

## Before
- Toggling when clicking on the section (as opposed to the little toggle in the top right) would execute the callback twice because it had both an `on:click` handler and was the `<label />` for the `<input type="checkbox" />`

https://github.com/gitbutlerapp/gitbutler/assets/7415984/b755ea15-08fb-4875-a14a-244370cc6945

## After

- Toggling when clicking on the section (as opposed to the little toggle in the top right) only toggles the state once now
- Also the `AnonMetrics` section `on:click` handler was calling the wrong fn

### DIFF
--- a/app/src/lib/components/AnalyticsSettings.svelte
+++ b/app/src/lib/components/AnalyticsSettings.svelte
@@ -12,16 +12,19 @@
 	const metricsEnabled = appMetricsEnabled();
 	const nonAnonMetricsEnabled = appNonAnonMetricsEnabled();
 
-	function toggleErrorReporting() {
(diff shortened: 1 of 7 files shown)


mtsgrd:
Just to check, given pr description it felt like you intended to keep the ability to click on the card to toggle, but it looks from the code that you want to make it so you have to click the toggle component?

ndom91:
@mtsgrd yeah so the card toggle still works because its the `label` for those `inputs`. The bug was because it was both the `label` and had an additional event handler on it.

PR #3978, opened 2024-06-03
support setting signCommits git config

krlvi:


crates/gitbutler-core/src/config/git.rs
```diff
@@ -0,0 +1,35 @@
...
+        let repo = git2::Repository::open(&self.path)?;
+        let config = repo.config()?;
+        match config.open_level(ConfigLevel::Local) {
+            Ok(mut local) => local.set_bool(key, val).map_err(Into::into),
+            Err(err) => Err(err.into()),
+        }
+    }
+
+    fn get_bool(&self, key: &str) -> Result<Option<bool>> {
+        let repo = git2::Repository::open(&self.path)?;
```
Byron, line 25: I think for reading and probably writing configuration in this ad-hoc manner (i.e. open a repo, then do something small), we'd want to use `gix::Repository` wherever we can.
And for configuration, I think we should as it's much faster in opening repositories.
~~I will submit a patch showing how this is done shortly.~~ The patch [is here](https://github.com/gitbutlerapp/gitbutler/pull/3985).

Byron (commented):
Unfortunately writing configuration files back through a convenient API isn't supported yet, but I put it as TODO (also on my tracking issue for features GB needs).

PR #3988, opened 2024-06-04
fix: minor showError refactor

ndom91:
- Updated usage of vanilla `svelte-french-toast` to our own `toasts.ts` ephemeral toasts so they animate in from the bottom, etc.
- All other refactorings basically required more and more changes, so we decided to cut those back and to just ship the minimal fix for now and potentially revisit refactoring the Toasts at a later date


mtsgrd:
lgtm

Issue #2590, opened 2024-02-12
Type to search/filter target branch selector


abjorck:
When opening a git project with hundreds or branches it is tedious to find the right "target" by scrolling

inickt:
Was about to open a request for this too– thanks to #2923 basically being done in #3677, unfortunately now I have 837 across the remotes I have set up for one of my repos 😅

krlvi:
Hello, thanks to @estib-vega, this should be now fixed. Would you mind trying it out with the latest release (0.12.4)?

inickt:
@krlvi This is great! Definitely a nice QOL improvement. Seems to be 1 tiny UI bug where if you select the remote with your mouse, it doesn't update the UI. Functionally it seems to work though.

https://github.com/gitbutlerapp/gitbutler/assets/4574041/0f15ea98-be81-4940-bb53-7468b59ec0f7

Also might be nice if the cursor shows up when typing (or always reset the filter) but that is such a minor nit. It looks if you have previously typed to filter (and select), and then you type again, it will append, but otherwise it will erase the filter completely (what I would expect to happen every time).

estib-vega:
Dang, nice catch! I think this should fix that: https://github.com/gitbutlerapp/gitbutler/pull/4102

It will reset always as a default

marktoman:
A major bug: If you uncheck something that filters out all branches, it makes the filter button disappear, making it impossible to revert the filter.
![Screenshot 2024-06-18 at 6 42 31](https://github.com/gitbutlerapp/gitbutler/assets/47689923/de01918c-233f-472d-a7e8-58ff5fd275c4)

Byron:
Great catch, I can confirm that.

What's worse is that it survives reloads!

https://github.com/gitbutlerapp/gitbutler/assets/63622/583beb0a-a4f7-470d-95cc-e7b58535c6dd

estib-vega:
Hey @marktoman 👋
That's a big one indeed, nice catch! But I think that bug is unrelated to this issue.

This other PR should fix that, though: https://github.com/gitbutlerapp/gitbutler/pull/4103

Byron:
That's true, putting the feedback label back, as this issue is talking about the target branch selector.

Caleb-T-Owens:
<img width="635" alt="Screenshot 2024-08-15 at 15 00 44" src="https://github.com/user-attachments/assets/e401b642-ba1b-441b-8875-bcf8686e4ae0">

[...]

PR #3992, opened 2024-06-04
Select input improvements: Arrow navigation & text filtering

estib-vega:
Related to #2590

This adds type-to-filter and up-down-arrow navigation to the Select component. 
This should enable easy & fast switching between remote branches 😊

This is how it looks when using it on my local fork of gitbutler.
Once the dropdown is open, one can navigate up and down the options and click enter to select.
Added to that, one can type text (shown in the Select component's text field) and the list will be filtered based on whether or not the typed context is included in the label displayed.

select-optimization

### Style
The way this shows what option is being 'highlighted' when navigating with the arrows, if by changing the background to
```
background-color: var(--clr-bg-3);
```

I'm not sure this is consistent with design guidelines or not. Happy to change that accordingly if not

### DIFF
--- a/app/src/lib/ai/types.ts
+++ b/app/src/lib/ai/types.ts
@@ -39,11 +39,11 @@ export interface AIClient {
 	defaultCommitTemplate: Prompt;
 }
 
-export interface UserPrompt {
+export type UserPrompt = {
 	id: string;
 	name: string;
 	prompt: Prompt;
-}
+};
 
 export interface Prompts {
 	defaultPrompt: Prompt;
--- a/app/src/lib/components/ProjectSwitcher.svelte
+++ b/app/src/lib/components/ProjectSwitcher.svelte
@@ -4,6 +4,7 @@
 	import SelectItem from './SelectItem.svelte';
 	import { ProjectService, Project } from '$lib/backend/projects';
 	import { getContext, maybeGetContext } from '$lib/utils/context';
+	import { derived } from 'svelte/store';
 	import { goto } from '$app/navigation';
 
 	const projectService = getContext(ProjectService);
@@ -14,17 +15,13 @@
 		title: string;
 	};
 
-	let mappedProjects: ProjectRecord[] = [];
-
-	projectService.projects.subscribe((projectList) => {
-		// Map the projectList to fit the ProjectRecord type
-		mappedProjects = projectList.map((project) => {
-			return {
-				id: project.id,
-				title: project.title
-			};
-		});
-	});
+	// Create a derived store that maps the projects to a simpler structure
+	const mappedProjects = derived(projectService.projects, ($projects) =>
+		$projects.map((project) => ({
+			id: project.id,
+			title: project.title
+		}))
+	);
 
 	let loading = false;
 	let select: Select<ProjectRecord>;
@@ -37,7 +34,7 @@
 		label="Switch to another project"
 		itemId="id"
 		labelId="title"
-		items={mappedProjects}
+		items={$mappedProjects}
 		placeholder="Select a project..."
 		wide
 		bind:value={selectValue}
--- a/app/src/lib/components/ProjectSwitcher.svelte
+++ b/app/src/lib/components/ProjectSwitcher.svelte
@@ -15,7 +15,6 @@
 		title: string;
 	};
 
-	// Create a derived store that maps the projects to a simpler structure
 	const mappedProjects = derived(projectService.projects, ($projects) =>
 		$projects.map((project) => ({
 			id: project.id,
--- a/app/src/lib/components/ProjectSwitcher.svelte
+++ b/app/src/lib/components/ProjectSwitcher.svelte
@@ -9,11 +9,26 @@
 	const projectService = getContext(ProjectService);
 	const project = maybeGetContext(Project);
 
-	const projects = projectService.projects;
+	type ProjectRecord = {
+		id: string;
+		title: string;
+	};
+
+	let mappedProjects: ProjectRecord[] = [];
+
+	projectService.projects.subscribe((projectList) => {
+		// Map the projectList to fit the ProjectRecord type
+		mappedProjects = projectList.map((project) => {
+			return {
+				id: project.id,
+				title: project.title
+			};
+		});
+	});
 
 	let loading = false;
-	let select: Select;
-	let selectValue = project;
+	let select: Select<ProjectRecord>;
+	let selectValue: ProjectRecord | undefined = project;
 </script>
 
 <div class="project-switcher">
@@ -22,7 +37,7 @@
 		label="Switch to another project"
 		itemId="id"
 		labelId="title"
-		items={$projects}
+		items={mappedProjects}
 		placeholder="Select a project..."
 		wide
 		bind:value={selectValue}
--- a/app/src/lib/components/Select.svelte
+++ b/app/src/lib/components/Select.svelte
@@ -1,18 +1,30 @@
-<script lang="ts">
+<script lang="ts" context="module">
+	export type SelectItemType<S extends string> = Record<S, unknown>;
+</script>
+
+<script lang="ts" generics="SelectItemType extends Record<string, unknown>">
 	import ScrollableContainer from './ScrollableContainer.svelte';
 	import TextBox from './TextBox.svelte';
 	import { clickOutside } from '$lib/clickOutside';
+	import { filterStringByKey } from '$lib/utils/filters';
+	import { KeyName } from '$lib/utils/hotkeys';
+	import { throttle } from '$lib/utils/misc';
 	import { pxToRem } from '$lib/utils/pxToRem';
+	import { isChar } from '$lib/utils/string';
 	import { createEventDispatcher } from 'svelte';
+
+	const INPUT_THROTTLE_TIME = 100;
+
+	type SelectItemKeyType = keyof SelectItemType;
 
 	export let id: undefined | string = undefined;
 	export let label = '';
 	export let disabled = false;
 	export let loading = false;
 	export let wide = false;
-	export let items: any[];
-	export let labelId = 'label';
-	export let itemId = 'value';
+	export let items: SelectItemType[];
+	export let labelId: SelectItemKeyType = 'label';
+	export let itemId: SelectItemKeyType = 'value';
 	export let value: any = undefined;
 	export let selectedItemId: any = undefined;
 	export let placeholder = '';
@@ -27,8 +39,21 @@
 	let listOpen = false;
 	let element: HTMLElement;
 	let options: HTMLDivElement;
+	let highlightIndex: number | undefined = undefined;
+	let highlightedItem: SelectItemType | undefined = undefined;
+	let filterText: string | undefined = undefined;
+	let filteredItems: SelectItemType[] = items;
 
-	function handleItemClick(item: any) {
+	$: filterText === undefined
+		? (filteredItems = items)
+		: (filteredItems = filterStringByKey(items, labelId, filterText));
+
+	// Set highlighted item based on index
+	$: highlightIndex !== undefined
+		? (highlightedItem = filteredItems[highlightIndex])
+		: (highlightedItem = undefined);
+
+	function handleItemClick(item: SelectItemType) {
 		if (item?.selectable === false) return;
 		if (value && value[itemId] === item[itemId]) return closeList();
 		selectedItemId = item[itemId];
@@ -52,6 +77,78 @@
 
 	function closeList() {
 		listOpen = false;
+		highlightIndex = undefined;
+		filterText = undefined;
+	}
+
+	function handleEnter() {
+		if (highlightIndex !== undefined) {
+			handleItemClick(filteredItems[highlightIndex]);
+		}
+		closeList();
+	}
+
+	function handleArrowUp() {
+		if (highlightIndex === undefined) {
+			highlightIndex = filteredItems.length - 1;
+		} else {
+			highlightIndex = highlightIndex === 0 ? filteredItems.length - 1 : highlightIndex - 1;
+		}
+	}
+
+	function handleArrowDown() {
+		if (highlightIndex === undefined) {
+			highlightIndex = 0;
+		} else {
+			highlightIndex = highlightIndex === filteredItems.length - 1 ? 0 : highlightIndex + 1;
+		}
+	}
+
+	const handleChar = throttle((char: string) => {
+		highlightIndex = undefined;
+		filterText ??= '';
+		filterText += char;
+	}, INPUT_THROTTLE_TIME);
+
+	const handleDelete = throttle(() => {
+		if (filterText === undefined) return;
+
+		if (filterText.length === 1) {
+			filterText = undefined;
+			return;
+		}
+
+		filterText = filterText.slice(0, -1);
+	}, INPUT_THROTTLE_TIME);
+
+	function handleKeyDown(event: CustomEvent<KeyboardEvent>) {
+		if (!listOpen) {
+			return;
+		}
+		event.detail.stopPropagation();
+		event.detail.preventDefault();
+
+		const { key } = event.detail;
+		switch (key) {
+			case KeyName.Escape:
+				closeList();
+				break;
+			case KeyName.Up:
+				handleArrowUp();
+				break;
+			case KeyName.Down:
+				handleArrowDown();
+				break;
+			case KeyName.Enter:
+				handleEnter();
+				break;
+			case KeyName.Delete:
+				handleDelete();
+				break;
+			default:
+				if (isChar(key)) handleChar(key);
+				break;
+		}
 	}
 </script>
 
@@ -67,9 +164,10 @@
 		type="select"
 		reversedDirection
 		icon="select-chevron"
-		value={value?.[labelId]}
+		value={filterText ?? value?.[labelId]}
 		disabled={disabled || loading}
 		on:mousedown={() => toggleList()}
+		on:keydown={(ev) => handleKeyDown(ev)}
 	/>
 	<div
 		class="options card"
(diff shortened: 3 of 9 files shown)


app/src/lib/utils/string.ts
```diff
@@ -11,3 +11,23 @@ export function hashCode(s: string) {
...
+}
+
+/**
+ * Checks if a value is a string.
+ *
+ * @param s - The value to check.
+ * @returns A boolean indicating whether the value is a string.
+ */
+export function isStr(s: unknown): s is string {
+	return typeof s === 'string';
```
Caleb-T-Owens: Please use `==`

app/src/lib/ai/types.ts
```diff
@@ -39,11 +39,11 @@ export interface AIClient {
 	defaultCommitTemplate: Prompt;
 }
 
-export interface UserPrompt {
+export type UserPrompt = {
 	id: string;
 	name: string;
 	prompt: Prompt;
-}
+};
```
Caleb-T-Owens, line 46: Is there a reason to prefer a type over an interface here?

app/src/lib/utils/string.ts
estib-vega: Will do in a follow-up commit, but would you be please be open to reconsider this position at a later date?

Not only there are some cases in which even using TypeScript (e.g. when using union types) you need to use strict equality, but it's also just [faster than loose equality](https://stackoverflow.com/questions/12374815/javascript-vs-operators-performance).

Given, the performance gains are not huge but they are not 0. And I'd say that that in combination with the cases in which is still needed in TS it's something that should win against cross-language stylistic choices.

Ok that's what I needed to say haha 😅

app/src/lib/ai/types.ts
estib-vega: In this specific commit, just so that the shape can be inferred into the `SelectItemType extends Record<string, unknown>` generic. Interfaces can't be inferred as something that extends `Record`s

In general, just because `type` will raise compile errors if there are any other types **with the same name** in the same scope. While `interface`s will just merge both by default, which is not always what one wants.
It's not that common so it's not necessary a huge improvement.

estib-vega: Reference 😊 https://github.com/microsoft/TypeScript/issues/15300#issuecomment-332366024

app/src/lib/utils/string.ts
Caleb-T-Owens: Hey! I'm interested, could you share an example of how a union can be problematic?

estib-vega: Gladly!
The simplest example would be a something like this:
```ts
let example: string | number[];

function initializeExample() {
  example = [1];
}

initializeExample();

if (example == "1") {
  console.log("unsafe");
} else {
  console.log("safe");
}

// Console will log: unsafe
```
The TS linter would not catch this because the value is being set outside of the immediate scope in which the `if` statement is being evaluated, so it will still treat the union type as `string | number[]` instead of `number[]`.

Another example is with `any` which just completely disables TypeScript checks.
```ts
let example: any;

function initializeExample() {
  example = [1,2];
}

initializeExample();

if (example == "1,2") {
  console.log("unsafe");
} else {
  console.log("safe");
}

// Console will log: unsafe
```

Caleb-T-Owens: Hi @estib-vega,

We've talked about it internally, and we'll be moving over to `===`, so feel free to revert the commit, or I'll be coming in later this week to convert the project.

estib-vega: That's awesome!
Thanks for hearing this out! 😬

app/src/lib/components/ProjectSwitcher.svelte
```diff
@@ -9,11 +9,26 @@
...
+
+	projectService.projects.subscribe((projectList) => {
+		// Map the projectList to fit the ProjectRecord type
+		mappedProjects = projectList.map((project) => {
+			return {
+				id: project.id,
+				title: project.title
+			};
+		});
+	});
```
mtsgrd: Thanks for contributing! Since this will never unsubscribe, could we turn it into a derived store?

estib-vega: Sure thing!
I'll create a follow-up commit later today

app/src/lib/components/ProjectSwitcher.svelte
```diff
@@ -14,17 +15,13 @@
...
-	projectService.projects.subscribe((projectList) => {
-		// Map the projectList to fit the ProjectRecord type
-		mappedProjects = projectList.map((project) => {
-			return {
-				id: project.id,
-				title: project.title
-			};
-		});
-	});
+	// Create a derived store that maps the projects to a simpler structure
```
mtsgrd: This comment does not add much value, maybe drop before we merge this pr?

app/src/lib/utils/misc.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Throttles the execution of a function.
+ * @param fn The function to be throttled.
+ * @param wait The time to wait between function invocations in milliseconds.
+ * @returns A throttled version of the original function.
```
mtsgrd: Type params are redundant with typescript, and perhaps we can omit the comment too since it's a simple function?

app/src/lib/components/Select.svelte
```diff
@@ -52,6 +77,80 @@
...
+
+		if (filterText.length === 1) {
+			filterText = undefined;
+			return;
+		}
+
+		filterText = filterText.slice(0, -1);
+	}, INPUT_THROTTLE_TIME);
+
+	function handleKeyDown(event: CustomEvent<KeyboardEvent>) {
```
mtsgrd: Can we use param name `e` for consistency with the rest of our code base? 🙏

app/src/lib/components/Select.svelte
```diff
@@ -27,8 +39,21 @@
...
 
-	function handleItemClick(item: any) {
+	$: filterText === undefined
+		? (filteredItems = items)
+		: (filteredItems = filterStringByKey(items, labelId, filterText));
+
+	// Set highlighted item based on index
+	$: highlightIndex !== undefined
+		? (highlightedItem = filteredItems[highlightIndex])
+		: (highlightedItem = undefined);
```
mtsgrd: Could you rewrite these as `$: var = expr ? val_1 : val_2`?
```js
$: highlightedItem = !highlightIndex ? filteredItems[highlightIndex] : undefined;
```

app/src/lib/components/Select.svelte
```diff
@@ -1,18 +1,30 @@
...
+	import { filterStringByKey } from '$lib/utils/filters';
+	import { KeyName } from '$lib/utils/hotkeys';
+	import { throttle } from '$lib/utils/misc';
 	import { pxToRem } from '$lib/utils/pxToRem';
+	import { isChar } from '$lib/utils/string';
 	import { createEventDispatcher } from 'svelte';
 
+	const INPUT_THROTTLE_TIME = 100;
+
+	type SelectItemKeyType = keyof SelectItemType;
```
mtsgrd: Since it's already a type, could we drop the `Type` suffix? To avoid conflict with `SelectItem.svelte` we could use e.g. `Selectable` and `SelectableKey`.

But also, since it's a keyof something that extends `Record<string, unknown>`, I don't see the value of this type. Wdyt?

mtsgrd (commented):
Again, thanks so much for contributing! Could you have a look at my comments and let me know what you think?

app/src/lib/components/ProjectSwitcher.svelte
estib-vega: Sounds good 👍

app/src/lib/utils/misc.ts
estib-vega: > Type params are redundant with typescript
>

Do you mean the `<T extends (...args: any[]) => any>`? I would disagree.
In this case it is accomplishing 3 things:
1. Narrows down the `fn` argument to be any kind of function
2. Ensures that the return type is always the exact same type of function as the one passed in as an argument.
3. Allows us to keep type-guards when accessing the parameters or return types, dynamically (see lines 12 and 14). 

Is that what you meant, or were you referring to something else?

> and perhaps we can omit the comment too since it's a simple function?
>

Will remove that 👍

app/src/lib/components/Select.svelte
estib-vega: Sure thing 👍

app/src/lib/components/Select.svelte
estib-vega: Yeah, that way is actually cleaner, will change that 😄

app/src/lib/components/Select.svelte
estib-vega: > Since it's already a type, could we drop the `Type` suffix?
>
Totally 👍 

> But also, since it's a keyof something that extends Record<string, unknown>, I don't see the value of this type.
>
It's so that we can narrow down the `labelId` and `itemId` properties (see lines 26 and 27). This way, the component dynamically checks what the item type is, and based on that we set the acceptable types for the `labelId` and `itemId`.

What this does in practice is for the `AIPromptSelect.svelte` component is that if we pass the `allPrompts: UserPrompt[]` variable as the `items` property, the component will infer that the only allowed keys are `"id"`, `"name"` or `"prompt"`

```ts
export type UserPrompt = {
	id: string;
	name: string;
	prompt: Prompt;
};
```

Which results in the auto-completion in the IDE:
auto-suggest

And if the wrong key is used or if it's updated, a TS compiler error will catch that:
type-guard

estib-vega:
@mtsgrd I've added another commit addressing the feedback given.
Let me know if some of the responses to the feedback still don't seem good enough arguments for keeping some things as they are. 
I'd be happy to change them if still needed!

app/src/lib/utils/misc.ts
mtsgrd: Sorry, no I meant the jsdoc, not the actual parameter!

app/src/lib/components/Select.svelte
```diff
@@ -27,8 +39,18 @@
...
 	let options: HTMLDivElement;
+	let highlightIndex: number | undefined = undefined;
+	let highlightedItem: Selectable | undefined = undefined;
+	let filterText: string | undefined = undefined;
+	let filteredItems: Selectable[] = items;
 
-	function handleItemClick(item: any) {
+	$: filterText === undefined
+		? (filteredItems = items)
+		: (filteredItems = filterStringByKey(items, labelId, filterText));
```
mtsgrd: This one too pls :)

app/src/lib/utils/filters.ts
```diff
@@ -1,4 +1,28 @@
...
+
+/**
+ * Filters an array of objects based on a specific string value in a given key.
+ *
+ * @template T - The type of objects in the array.
+ * @template K - The key of the object to filter by.
+ * @param {T[]} arr - The array of objects to filter.
+ * @param {K} key - The key to filter by.
+ * @param {string} value - The string value to search for in the specified key.
+ * @returns {T[]} - The filtered array of objects.
```
mtsgrd: If you agree with the other comment about jsdc then we can probably omit this too?

mtsgrd:
So not directly about this pull request, but I'm curious, would you know how to generically type the <Modal> component to give us types for the `let:item` directive?

app/src/lib/utils/misc.ts
estib-vega: Ah gotcha :). Yeah, those are redundant 👍

app/src/lib/components/Select.svelte
mtsgrd: Thanks, you're right. That's awesome.

app/src/lib/components/Select.svelte
estib-vega: Totally, yeah 👍

app/src/lib/utils/filters.ts
estib-vega: Ah yeah, sorry should have thought about that one. Will take it out too

estib-vega:
> So not directly about this pull request, but I'm curious, would you know how to generically type the component to give us types for the `let:item` directive?

It's a good question. I'm not sure, would need to try it out.
I'm still learning how "advanced" TS interacts with Svelte. Plus my understanding is that generics themselves are relatively new to Svelte components so they might still have a funky behavior.

My first guess would be to just add the type generic type in the prop declaration and see if it works as we expect.

Do you have a specific use-case in mind 👀?

app/src/lib/utils/string.ts
```diff
@@ -11,3 +11,17 @@ export function hashCode(s: string) {
 	}
 	return hash.toString();
 }
+
+/**
+ * Checks if a string is a single character.
+ */
```
mtsgrd: Same line of thought as earlier, let's drop these comments since it feels redundant.

app/src/lib/utils/filters.ts
```diff
@@ -1,4 +1,21 @@
...
+
 // If a value occurs > 1 times then all but one will fail this condition.
 export function unique(value: any, index: number, array: any[]) {
 	return array.indexOf(value) === index;
 }
+
+/**
+ * Filters an array of objects based on a specific string value in a given key.
+ */
+export function filterStringByKey<T, K extends keyof T>(arr: T[], key: K, value: string): T[] {
```
mtsgrd: I would again drop the comment since the function name is sufficient to understand what's going on. But, could we still rename the function? It looks like it would be more appropriately called `filterObjectsByKey`?

Caleb-T-Owens: At the same time, I'm curious of the value of having this function at all given that it can be rather clearly as:

```ts
declare const items: ({ foo: number, bar: string })[]
const needle = "asfd"

items.filter(({ bar }) => bar.includes(needle))
```

Caleb-T-Owens: To me writing the filter statement is clearer than calling a function called `filterObjectsBySpecifiedPropertyThatIncludesString(objects, "key", "asfd")`

app/src/lib/utils/string.ts
estib-vega: Will do 👍

app/src/lib/utils/filters.ts
estib-vega: I don't have strong feelings about either approach :).

The main 2 reasons for which I opted for a separate dedicated function:
- I think inlining functional array methods into certain expressions can be actually a bit harder to read. In this case we'd write that into the ternary operator:

```ts
	$: filteredItems =
		filterText === undefined ? items : items.filter((it) => it[labelId].includes(filterText));
```
vs

```ts
$: filteredItems =
		filterText === undefined ? items : filterObjectsByKey(items, labelId, filterText);
```

- Also, just because `for`-loops are [generally more performant](https://medium.com/@maxheadway/exploring-performance-traditional-loops-vs-functional-methods-in-javascript-76dedc749375) than functional array methods (which kind of sucks because they are more elegant in most cases). In the case that the list of e.g. target branches is huge, filtering the list will become more expensive and we should use as much performance optimizations as we can.

In practice that might not be the case most times so if you still prefer the functional method I'm ok with that 👍

estib-vega: Also, sorry if I keep pushing back on some of the comments. I appreciate both of you being so patient 😄😄

Caleb-T-Owens: > Also, sorry if I keep pushing back on some of the comments. I appreciate both of you being so patient 😄😄

I'm convinced that having discussion between multiple developers always ends up producing a much better result. Its great that you're pushing back and helping us consider the problem from your perspective. Please don't stop 😂!!

---

> Also, just because for-loops are [generally more performant](https://medium.com/@maxheadway/exploring-performance-traditional-loops-vs-functional-methods-in-javascript-76dedc749375) than functional array methods (which kind of sucks because they are more elegant in most cases). In the case that the list of e.g. target branches is huge, filtering the list will become more expensive and we should use as much performance optimizations as we can.

This is undoubtably true, but in the vast majority of situations the difference in performance is simply not relevant, even on lists of 10,000 odd items.

I think our general consensus is that code clarity should take a priority over scraping every last nanosecond of performance.

> I think inlining functional array methods into certain expressions can be actually a bit harder to read. In this case we'd write that into the ternary operator:

I'm still struggling to be convinced about your `filterObjectsByKey` because the name doesn't tell me that you're doing a string `.includes` on it. If anything `filterObjectsByKey` tells me that you've got a list of arbitrarily shaped objects and you're filtering out the objects which have one particular key.

I do see that you're point about the expression getting quite dense, so how about we define a function local to the component and say:

```ts

function filterItems(items: Selectable[], filterText: string | undefined) {
  if (filterText) {
    return items.filter((it) => it[labelId].includes(filterText));
  } else {
    return items
  }
}

$: filteredItems = filterItems(items, filterText)
```

By keeping the definition local, we're also leaving a convenient place where would could pull in a fuzzy comparison operation in the future

estib-vega: > I'm convinced that having discussion between multiple developers always ends up producing a much better result. Its great that you're pushing back and helping us consider the problem from your perspective. Please don't stop 😂!!
>
Haha I agree 😄

> By keeping the definition local, we're also leaving a convenient place where would could pull in a fuzzy comparison operation in the future
>
Seems like a good compromise. Will add it as you described :)

app/src/lib/components/Select.svelte
```diff
@@ -27,8 +38,28 @@
...
+
+		return items.filter((it) => {
+			const property = it[labelId];
+			if (!isStr(property)) return false;
+			return property.includes(filterText);
+		});
+	}
 
-	function handleItemClick(item: any) {
+	$: filteredItems = filterText === undefined ? items : filterItems(items, filterText);
```
Caleb-T-Owens: filterItems handles the case for when the filterText is undefined so this can be reduced to the following:
```suggestion
$: filteredItems = filterItems(items, filterText);
```

estib-vega: Oh right, nice catch!

mtsgrd (approved):

PR #3996, opened 2024-06-04
Memoize derive calls

Caleb-T-Owens:
If we call files() 90 odd times, it will create 90 different derived stores. This means that we're going to do an awful lot of extra work as each update to the selected ids will trigger 90 jobs.

By memoizing the store, we will only ever create one derived object and as such eliminate the dupilcate work.


mtsgrd:
Nice!

PR #3999, opened 2024-06-04
refactor-diff-workdir

krlvi:


crates/gitbutler-core/src/virtual_branches/base.rs
```diff
@@ -355,10 +354,13 @@ pub fn update_base_branch(
...
         .context("failed to get new target commit tree")?;
 
-    let old_target_tree = repo.find_commit(target.sha)?.tree().context(format!(
-        "failed to get old target commit tree {}",
-        target.sha
-    ))?;
+    let old_target_tree = repo
+        .find_commit(target.sha.into())?
+        .tree()
+        .context(format!(
```
Byron, line 360: When doing anything that is non-const, like `format!` which allocates, one would use `with_context(|| <expensive stuff>)` to have the cost only when needed.

crates/gitbutler-core/src/virtual_branches/virtual.rs
```diff
@@ -1696,8 +1696,11 @@ fn get_applied_status(
     target_sha: &git::Oid,
     mut virtual_branches: Vec<branch::Branch>,
 ) -> Result<(AppliedStatuses, Vec<diff::FileDiff>)> {
-    let base_file_diffs = diff::workdir(&project_repository.git_repository, integration_commit)
-        .context("failed to diff workdir")?;
+    let base_file_diffs = diff::workdir(
+        project_repository.repo(),
+        &integration_commit.to_owned().into(),
+    )
+    .context("failed to diff workdir")?;
```
Byron, line 1703: Here for example, `context()` is perfect as the message is statically known.

crates/gitbutler-core/tests/virtual_branches/mod.rs
```diff
@@ -1266,28 +1224,21 @@ fn detect_mergeable_branch() -> Result<()> {
...
-        .git_repository
-        .head()
-        .unwrap()
-        .target()
-        .unwrap();
-    project_repository.git_repository.reference(
-        &"refs/remotes/origin/remote_branch".parse().unwrap(),
-        up_target.into(),
+    commit_all(project_repository.repo());
+    let up_target = project_repository.repo().head().unwrap().target().unwrap();
```
Byron, line 1238: We have to be careful with these `unwrap()` calls as a panic *may* cause a timeout in the UI. When we see them, let's `?` out of there or add `context()` in case of `Option`.

I think they might be from times when `thiserror` made it harder to error out at will.

Byron (commented):
It really is getting there 👍!

crates/gitbutler-core/tests/virtual_branches/mod.rs
krlvi: i only allowed myself to do it because its testcode

Byron: 🤦‍♂️
And I double checked that it's not test-code, yet I failed. Apologies.

There is a bunch of old unwraps still outside of test-code though, and I hope we can slowly eradicate them as touch the code for other refactors. Now it should be easy thanks to `anyhow`.

PR #4005, opened 2024-06-05
fix: use `svelte:window` for managing hotkey event listeners

ndom91:
- Use `svelte:window` event listeners to handle global keybinds
	- Auto adds and disposes of the event listeners on dismount
- Loosen up `.nvmrc` requirements and rm hardcoded node version in CI


mtsgrd:
Yeah I think it'd be great to get rid of this tiny dependency. As you suggest, I think it'd be great to roll our own thing to avoid the extra noise of e.g. `preventDefault()` on each key handler?

PR #4008, opened 2024-06-05
reimplement-hunk-locking-code

Caleb-T-Owens:


krlvi (approved):

crates/gitbutler-core/src/git/diff.rs
```diff
@@ -101,6 +101,22 @@ impl GitHunk {
     }
 }
 
+/// Comparison
+impl GitHunk {
+    /// integration_intersects_unapplied is used to determine if a hunk from a diff between integration and the trunk intersects with an unapplied hunk.
+    /// We want to use the new start/end for the integraiton hunk and the old start/end for the unapplied hunk.
+    pub fn integration_intersects_unapplied(
+        integration_hunk: &GitHunk,
```
Byron, line 109: I think when the first argument is also a `GitHunk`, it should really be `&self` to allow calling it more naturally.
With an adjusted method name it should also 'sound' natural.

Byron (commented):
Just one note, please ignore if that's very deliberate which I think it probably even is.

crates/gitbutler-core/src/git/diff.rs
Caleb-T-Owens: Hi @Byron, this was indeed deliberate because I wanted to clearly name the roles of each hunk you're passing in because its very important in this case.

I could define some `struct IntegrationHunk(GitHunk)` which has the method on it but it felt a bit excessive

PR #4009, opened 2024-06-05
fix(windows): replace project.path with project.vscodepath, fixes #4003

### LINKED ISSUE #4003: "Open in VSCode" does not open the current repository
## Description

When I click on the "Open in VSCode" button , VSCode opens, but it does not open the current repository. 

## Steps to Reproduce

1. Click on the "Open in VSCode" button.
image

## Expected Behavior

VSCode should open with the current project.

## Actual Behavior

VSCode opens as if it was launched using the `code` command from the terminal.
## Environment

- **Operating System:** Windows 11
- **VSCode Version:** 1.89.1

TheGB0077:
It seems the previous pull request #3495 addressed a fix for opening VSCode on Windows, but it missed those two extra files that still had the bug


ndom91 (approved):
Good catch, thanks!

PR #4010, opened 2024-06-05
fix: simplify Modal logic

ndom91:
- Add `svelte:window` click listener to handle a Modal's `clickoutside` action
- Removes need for `svelte-outclick` dependency
- @mtsgrd this one is basically good to go imo, except same issue with the pnpm-lock file, therefore I'm keeping it on draft for now.


mtsgrd:
Hmm, can I just double check that this doesn't close the modal when clicking inside?

ndom91:
> Hmm, can I just double check that this doesn't close the modal when clicking inside?

Yeah, did it not work for you? 

The idea is basically, "if the modal contains the target element of the click, then don't close"

I just triple checked and that logic does seem to hold :+1:

mtsgrd:
@ndom91 could you take another look at this? I think it breaks at least the "delete branch" modal in the branch lane menu.

PR #4017, opened 2024-06-06
fix: use hunksection instead of individual lines

ndom91:
- Improve hunkviewer perf
	- Went from ~9s to open a test `pnpm-lock.yaml` diff to ~2-3s
- Moves hunk lines from individual comopnents into 1 `HunkSection` component
- It previously spent a lot of time setting dynamic `style:` attributes on each line
	- Now we set css vars on the parent element of all the lines only once, and those css custom properties are consumed by the line elements themselves in css


app/src/lib/components/HunkLines.svelte
```diff
@@ -118,7 +135,7 @@
 	.selectable-wrapper {
 		cursor: text;
 		display: inline-block;
-		text-indent: 4px;
-		margin-right: 4px;
+		text-indent: var(--size-4);
+		margin-right: var(--size-4);
```
mtsgrd: I think @PavelLaptev has phased out --size now in favour of automatically translating px into rem when building the app.

app/src/lib/components/HunkLines.svelte
```diff
@@ -34,42 +39,51 @@
...
-			on:click={() => selectable && dispatch('selected', !selected)}
-			class="numbers-line-count"
-			class:selected={isSelected}
-			style:min-width={minWidth + 'rem'}
-			style:cursor={draggingDisabled ? 'default' : 'grab'}
+<div
+	class="line-wrapper"
+	style="--tab-size: {tabSize}; --minwidth: {minWidth}rem; --cursor: {draggingDisabled
+		? 'default'
+		: 'grab'}"
```
mtsgrd, line 48: Hmm, minWidth should probably be expressed in px now.. but I'm wondering if inline styles will be translated? @PavelLaptev

app/src/lib/components/HunkLines.svelte
ndom91: Cleaned up :+1:

app/src/lib/components/HunkLines.svelte
ndom91: So based on our conversation, for inline styles they're either supposed to be expressed in `px` and use the helper fn or just directly in `rem`'s, right?

app/src/lib/components/HunkViewer.svelte
```diff
@@ -75,7 +73,7 @@
 		class:opacity-60={section.hunk.locked && !isFileLocked}
 	>
 		<div bind:this={contents} class="hunk__bg-stretch">
-			{#if linesModified > 1000 && !alwaysShow}
+			{#if linesModified > 5000 && !alwaysShow}
```
ndom91: @mtsgrd any opinion on what we should set this to? I had a ~4900 line test file that took ~2-3s to render now. Still seems a bit long though, so maybe we set this to 2500-3000, somewhere in that range?

app/src/lib/components/HunkLines.svelte
PavelLaptev: [pxToRem](https://github.com/cuth/postcss-pxtorem#readme) plugin can't convert `PX` that were written inline in the style attribute. In this case, if it's not a component property, you can use `REM` directly. Or if it's a property you can use a helper function like in this file https://github.com/gitbutlerapp/gitbutler/blob/cdd596f8bc4b9801671793360184f1236766f0c6/app/src/lib/components/Spacer.svelte#L11

app/src/lib/components/HunkViewer.svelte
krlvi: i guess 2500 is reasonable

krlvi:
something looks off - could it be a chromium/webkit thing?
image

ndom91:
> something looks off - could it be a chromium/webkit thing? image

Good catch, fixed :+1:

app/src/lib/components/HunkViewer.svelte
ndom91: Updated to 2500 :+1:

PR #4019, opened 2024-06-06
lets try to speed CI with a custom runner

krlvi:


krlvi:
nevermind, its not really faster

PR #4026, opened 2024-06-06
feat(project): enable new locking by default

krlvi:
@mtsgrd @Caleb-T-Owens what do u guys think? good enough to try having it as the default for the patch release? (i have been trying to see if i can break it today)


krlvi:
fuck it, lets go for it - it appears to work better!

PR #4028, opened 2024-06-06
Revert "fix: simplify Modal logic"

krlvi:
Reverts gitbutlerapp/gitbutler#4010

Seems like something is not quite right, reverting so that we can make a patch release tonight


krlvi:
@ndom91 FYI.
I guess it's probably something tiny, but i couldn't forward fix it unfortunately, so reverting here and we can get it back tomorrow

PR #4036, opened 2024-06-07
Fix a couple of small typos

LukeCarrier:
Noticed these while digging into some other problems and figured a smaller diff would make a better first contribution 😅


Byron:
Thank you very much, appreciated!
And we are of course looking forward to seeing what else you are working on :)!

PR #4047, opened 2024-06-09
set timeout of 2 sec when caching assets

krlvi:


crates/gitbutler-core/src/assets.rs
```diff
@@ -168,7 +168,11 @@ impl Proxy {
 
         tracing::debug!(url = %src, "downloading image");
 
-        let resp = reqwest::get(src.clone()).await?;
+        let client = reqwest::Client::builder()
+            .timeout(Duration::from_secs(2))
```
Byron, line 172: After having read how live on the southpole is with shoddy internet, I believe setting short timeouts is also not the right thing to do.
However, I think when this is refactored to not be in the critical path (but more of a background, on demand thing), that can be adjusted again.

Byron: PS: I also think that it would help to add something like `(#issue-number)` to the subject, as it would reveal the original motivation for the timeout.

PR #4048, opened 2024-06-09
if an oplog entry had an error, show it

krlvi:
looks something like this:
image
image


krlvi:
fyi @PavelLaptev

krlvi:
design process for this mvp 🙈:
image

PavelLaptev:
👍 Got it

PR #4054, opened 2024-06-10
Diff the correct way round

Caleb-T-Owens:


crates/gitbutler-core/src/virtual_branches/virtual.rs
```diff
@@ -1615,7 +1615,7 @@ fn new_compute_locks(
             let commit = repository.find_commit(branch.head).ok()?;
             let tree = commit.tree().ok()?;
             let diff = repository
-                .diff_tree_to_tree(Some(&tree), Some(&target_tree), Some(opts))
+                .diff_tree_to_tree(Some(&target_tree), Some(&tree), Some(opts))
```
Byron, line 1618: What frightens me a bit is that this can be done without breaking any test 🥶.

Caleb-T-Owens: I'll write some tests next week 👍

Issue #4029, opened 2024-06-07
Toast message with the new version change log contains an error field with "undefined"


krlvi:
![image](https://github.com/gitbutlerapp/gitbutler/assets/4030927/e633e3fa-7907-4bc8-a2f8-d30da9e3bc44)

mtsgrd:
Fixed.

PR #4059, opened 2024-06-10
Hide error div in toast when no error present

mtsgrd:
Slots cannot be conditionally filled with an if statement, and there is no way to check if a slot has content. So.. feels like we can either switch between two `<InfoMessage>` components, or pass error as a prop. @PavelLaptev wdyt?


mtsgrd:
Fixes #4029

mtsgrd:
Merging for now so we can have the fix in the nightly, and let's chat offline about slots vs. props? @ndom91 @Caleb-T-Owens in case you guys have any thoughts.

Issue #4060, opened 2024-06-11
When generating commit messages, use any existing message that was inputted as part of the prompt


krlvi:
... and ask the LLM to improve it with the full context

PR #4069, opened 2024-06-11
Commit message gen: Use user input as starting point

estib-vega:
Related to #4060

Flag to have the commit generation only improve or reword the given user input.

- The flag is persisted just as the the brevity and emoji use
- Add types prompt parameters and directives for improved type guarding and **dynamic meta-prompt generation**
- Add directive-specific meta-prompting for Ollama client.

improve-user-input

Also:
- Add branded types util (this might be just as well moved to the `typeguards.ts`)
- Gracefully handle Network Errors on the Ollama Client (e.g. when the Ollama server is not running)
Screenshot 2024-06-11 at 10 00 33

### DIFF
--- a/app/src/lib/ai/ollamaClient.ts
+++ b/app/src/lib/ai/ollamaClient.ts
@@ -1,7 +1,16 @@
 import { LONG_DEFAULT_BRANCH_TEMPLATE, LONG_DEFAULT_COMMIT_TEMPLATE } from '$lib/ai/prompts';
 import { MessageRole, type PromptMessage, type AIClient, type Prompt } from '$lib/ai/types';
+import { isStr } from '$lib/utils/string';
 import { isNonEmptyObject } from '$lib/utils/typeguards';
 import { fetch, Body, Response } from '@tauri-apps/api/http';
+
+const NETWORK_ERROR_MESSAGE =
+	'Failed to connect to the LLM. Please make sure it is running and the endpoint is configured correctly.';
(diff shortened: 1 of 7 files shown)


estib-vega:
A bit more on the *dynamic meta-prompt generation*

In order to get the smaller LMs to only reword/extend the given user input, they needed to be provided with specific meta-prompts again.
I tried adding those simply to the array of `PromptMessage`s, but the generation became (a bit but noticeably) slower and not as accurate, in my limited testing at least.

That's why I came up with this approach:
Depending on the directives given (generate from scratch or edit from user input), dynamically build the array of `PromptMessage`s.

This changes do not interfere with the ability to customize the prompts on the application settings.

krlvi:
hey, thanks for this contribution! I didn't tag https://github.com/gitbutlerapp/gitbutler/issues/4060 properly at the time - it's probably an interesting idea, and im curious to try it out. 
I will take a proper look / merge tomorrow, we wanna first make a bugfix release with some performance things tonight

estib-vega:
> hey, thanks for this contribution! I didn't tag #4060 properly at the time - it's probably an interesting idea, and im curious to try it out. I will take a proper look / merge tomorrow, we wanna first make a bugfix release with some performance things tonight

All good, no rush 😉

estib-vega:
I've made is a bit more strict about how this treats the starting message.
It will now attempt to leave it as is and only adding text to it.

It still seems to be very probabilistic with Llama 3 8B. Not sure how it behaves with bigger models
commit-improvement

PR #4073, opened 2024-06-11
improve oplog snapshot creation performance

krlvi:
We learned that the tree creation that happens during oplog entry creation can be slow on repos with large indexes/many files:
image
In general, GitButler performance takes a hit on repos with large indexes due to the aggressive tree writing (something for us to address). 

In this PR, i am removing the "workdir" subtree of the snapshot because that can be created on demand if/when needed.

One downside of this implementation is the fact that computing the workdir tree on the fly means potentially doing extra work when listing / getting the diff of a snapshot. There is a small cache outside the "list snapshots" loop, but that is not available between different invocations.

### DIFF
--- a/crates/gitbutler-core/src/ops/oplog.rs
+++ b/crates/gitbutler-core/src/ops/oplog.rs
@@ -264,6 +264,8 @@ impl Project {
 
         let mut snapshots = Vec::new();
 
+        let mut wd_trees_cache: HashMap<git2::Oid, git2::Oid> = HashMap::new();
+
         for commit_id in revwalk {
             if snapshots.len() == limit {
                 break;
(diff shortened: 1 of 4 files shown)


krlvi:
@Byron fyi

crates/gitbutler-core/src/ops/oplog.rs
```diff
@@ -786,3 +756,50 @@ fn deserialize_commit(
...
+
+    let target_tree_entry = snapshot_tree
+        .get_name("target_tree")
+        .context("failed to get target tree entry")?;
+    let target_tree = repo
+        .find_tree(target_tree_entry.id())
+        .context("failed to convert target tree entry to tree")?;
+
+    let vb_toml_entry = snapshot_tree
+        .get_name("virtual_branches.toml")
```
Byron, line 776: It might be less cumbersome to use `get_path(format!("target_tree/{id}/virtual_branches.toml", id = target_tree_entry.id()))` here, but I might be missing the reason that's not done as well.

Byron (commented):
Thanks for letting me know.

It's strange to see that merging multiple trees is slower than writing it.
Maybe `gitoxide` could have merge-facilities that are optimized for multi-branch merges to reuse data in-memory without having to restore to continuously producing new indices that are then written back.

PR #4086, opened 2024-06-15
feat: Add in-memory method to open temporary repository for comparing or getting diffs.

krlvi:


Byron:
That's cool, and I am already curious how performance will improve due to it on the one hand, but also how one can assure that the required in-memory objects are eventually flushed to disk on the other hand.

PR #4094, opened 2024-06-16
Use https for remotes when remote is using https

### LINKED ISSUE #4085: Adding remote uses SSH when origin uses HTTPS
Adding a remote via GitButler (from the "Apply from fork" button) sets up the remote with SSH. As I use `gh` with HTTPS I'm unable to fetch refs over SSH via `git` and GitButler also fails when its authentication is configured to "Use a Git executable" or "Use a Git credentials helper".

## Expected behavior

GitButler configures remotes with HTTPS or SSH based on the schema of `origin`.

## Workaround

Manually adding the remote with the correct schema works.

## System info

GitButler 0.12.4
Git 2.39.4 (Apple Git-150.1)

Caleb-T-Owens:
* Closes https://github.com/gitbutlerapp/gitbutler/issues/4085

### DIFF
--- a/app/src/lib/components/PullRequestPreview.svelte
+++ b/app/src/lib/components/PullRequestPreview.svelte
@@ -8,10 +8,12 @@
 	import { RemotesService } from '$lib/remotes/service';
 	import { getContext } from '$lib/utils/context';
 	import * as toasts from '$lib/utils/toasts';
+	import { remoteUrlIsHttp } from '$lib/utils/url';
 	import { BaseBranchService } from '$lib/vbranches/baseBranch';
 	import { BranchController } from '$lib/vbranches/branchController';
 	import { VirtualBranchService } from '$lib/vbranches/virtualBranch';
 	import { marked } from 'marked';
+	import { get } from 'svelte/store';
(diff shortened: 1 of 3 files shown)


app/src/lib/components/PullRequestPreview.svelte
```diff
@@ -33,8 +35,27 @@
...
+		const baseRemoteUrl = get(baseBranchService.base)?.remoteUrl;
+
+		if (!baseRemoteUrl) return;
+
+		if (remoteUrlIsHttp(baseRemoteUrl)) {
+			if (!pullrequest.repositoryHttpsUrl) return;
+			return pullrequest.repositoryHttpsUrl;
+		} else {
+			if (!pullrequest.repositorySshUrl) return;
+			return pullrequest.repositorySshUrl;
```
H1ghBre4k3r: Wouldn't it be sufficient to just return `pullrequest.repository(Http|Ssh)Url` directly? Since both can either be set or `undefined`.

Caleb-T-Owens: You're right... 🤦 

That really is quite daft of me

PR #4102, opened 2024-06-17
Select component: Minor filtering fix

estib-vega:
Filtering and then selecting with the cursor would not reset the filter. Trying to filter the list after that would only append to the remaining filter string content. This is now fixed by always reseting the filter text


estib-vega:
reported here: https://github.com/gitbutlerapp/gitbutler/issues/2590#issuecomment-2168432046

Caleb-T-Owens:
Sorry for the wait! Thank you greatly for the contribution, its really appreciated 👍

estib-vega:
No problem at all :)

PR #4103, opened 2024-06-18
Branches filtering: Don't hide the filter component when filtering

### LINKED ISSUE #4106: Branch filtering option disappeared
Hi, I unchecked every option in the filter menu for branches and now it's gone and I can't turn them back on!

image

I'm on Mac and using Version 0.12.5 (20240617.130345)

estib-vega:
If you filter out all branches, the filter component would disappear. Making it impossible to unset the filters.

As reported here: https://github.com/gitbutlerapp/gitbutler/issues/2590#issuecomment-2174993191

This change fixes that.

### DIFF
--- a/app/src/lib/components/Branches.svelte
+++ b/app/src/lib/components/Branches.svelte
@@ -109,19 +109,23 @@
 </script>
 
 <div class="branch-list">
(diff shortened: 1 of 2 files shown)


Caleb-T-Owens:
Hi @estib-vega, thank you for taking a look into this! I ended up choosing to rename the props and introduce a new `totalBranchCount` to make the logic clearer.

PR #4109, opened 2024-06-18
Update modal component branch

Caleb-T-Owens:


app/src/lib/branches/dragActions.ts
```diff
@@ -0,0 +1,53 @@
+import { DraggableCommit, DraggableHunk, DraggableFile } from '$lib/dragging/draggables';
+import { filesToOwnership } from '$lib/vbranches/ownership';
+import type { BranchController } from '$lib/vbranches/branchController';
+import type { Branch } from '$lib/vbranches/types';
+
+class BranchDragActions {
+	constructor(
+		private branchController: BranchController,
+		private branch: Branch
+	) {}
```
Caleb-T-Owens, line 10: I was thinking about whether I should have some
```ts
interface DropzoneActions {
  accepts: //...
  onDrop: //...
}
```
So we'd have one class for each dropzone actions. It would also mean that we could pass something that implements that interface directly to the Dropzone component, but I didn't quite want to go that far yet.

app/src/lib/branches/dragActions.ts
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+export class BranchDragActionsFactory {
+	constructor(private branchController: BranchController) {}
+
+	build(branch: Branch) {
+		return new BranchDragActions(this.branchController, branch);
+	}
+}
```
Caleb-T-Owens, line 53: I'm using a factory so that when we go to use the drag actions, we only need to provide the relevant state, rather than having to also provide a bunch of controllers and servicies

app/src/lib/components/BranchDropzone.svelte
```diff
@@ -31,49 +31,45 @@
...
+<div class="canvas-dropzone">
+	<Dropzone {accepts} ondrop={onDrop}>
+		{#snippet overlay({ hovered, activated })}
+			<div class="new-virtual-branch" class:activated class:hovered>
+				<div class="new-virtual-branch__content">
+					<div class="stimg">
+						<div class="stimg__hand">
+							{@html handSvg}
+						</div>
+						<div class="stimg__top-sheet">
```
Caleb-T-Owens, line 43: Because this is quite a specialised dropzone, I chose not to have some generic component for the overlay that it uses.

app/src/lib/components/BranchLanePopupMenu.svelte
```diff
@@ -159,7 +159,7 @@
 			kind="solid"
 			on:click={async () => {
 				await branchController.deleteBranch(branch.id);
-				deleteBranchModal.close();
+				close();
```
Caleb-T-Owens, line 162: 😁 a cheeky unrelated change... if only we could work in multiple branches at the same time

app/src/lib/components/CommitDragItem.svelte
```diff
@@ -1,131 +1,60 @@
...
-		return (data: DraggableCommit) => {
-			if (data.commit.isParentOf(commit)) {
-				branchController.squashBranchCommit(data.branchId, commit.id);
-			} else if (commit.isParentOf(data.commit)) {
-				branchController.squashBranchCommit(data.branchId, data.commit.id);
-			}
-		};
-	}
+		return commitDragActionsFactory.build($branch, commit);
+	});
```
Caleb-T-Owens, line 24: I was considering saying something like `$derived($branch && commitDragActionsFactory.build($branch, commit))`, but I decided that I really just want to keep everything as simple as possible, and being explicit and having that sort of guard clause to me just expressed the intent much better

app/src/lib/components/CommitDragItem.svelte
```diff
@@ -1,131 +1,60 @@
...
-		<DropzoneOverlay class="amend-dz-marker" label="Amend" />
-		<DropzoneOverlay class="squash-dz-marker" label="Squash" />
-
-		<slot />
+	<div class="commit-card-wrapper">
+		{#if actions}
+			{@render ammendDropzone()}
+		{:else}
+			{@render children()}
+		{/if}
```
Caleb-T-Owens, line 33: This component needs to gracefully fall back to not talking to the branchController because its used in the remoteBranch and trunk pages.

I thought about having the snippets inside the if block too, but it ended up looking too messy

app/src/lib/components/CommitDragItem.svelte
```diff
@@ -1,131 +1,60 @@
...
+
+{#snippet squashDropzone()}
+	<Dropzone accepts={actions!.acceptSquash.bind(actions)} ondrop={actions!.onSquash.bind(actions)}>
+		{@render children()}
+
+		{#snippet overlay({ hovered, activated })}
+			<CardOverlay {hovered} {activated} label="Squash commit" />
+		{/snippet}
+	</Dropzone>
+{/snippet}
```
Caleb-T-Owens, line 56: If anyone has n idea about how to better express the nesting, I am all ears

app/src/lib/components/CommitList.svelte
```diff
@@ -97,8 +107,31 @@
...
+	}
 </script>
 
+{#snippet reorderDropzone(dropzone: ReorderDropzone, yOffsetPx: number)}
+	<Dropzone accepts={dropzone.accepts.bind(dropzone)} ondrop={dropzone.onDrop.bind(dropzone)}>
+		{#snippet overlay({ hovered, activated })}
+			<LineOverlay {hovered} {activated} {yOffsetPx} />
+		{/snippet}
+	</Dropzone>
+{/snippet}
```
Caleb-T-Owens, line 132: I didn't think an entire component was required for the Reordering dropzone. I'm willing to be swayed on this position though.

app/src/lib/dragging/dropzone.ts
```diff
@@ -1,37 +1,123 @@
...
+		if (this.registered) {
+			this.unregister();
+		}
+
+		// Register listeners and keep references to the functions
+		this.registered = true;
+
+		this.registeredOnDrop = async (e: DragEvent) => await this.onDrop(e, data);
+		this.registeredOnDragEnter = this.onDragEnter.bind(this);
+		this.registeredOnDragLeave = this.onDragLeave.bind(this);
```
Caleb-T-Owens: I do make the assumption that the user can't drag two things at once, but that seems like a pretty sound assumption to me

app/src/lib/dragging/dropzone.ts
```diff
@@ -1,37 +1,123 @@
...
+	private onDragLeave(e: DragEvent) {
+		e.preventDefault();
+		if (!this.active) return;
+		this.configuration.onHoverEnd();
+	}
+}
+
+export const dropzoneRegistry = new Map<HTMLElement, Dropzone>();
+
+export function dropzone(node: HTMLElement, opts: DropzoneConfiguration) {
```
Caleb-T-Owens: This is now only consumed by the Dropzone, so its not relevant to support a partial configuration that contains defaults

app/src/lib/dragging/reorderDropzoneManager.ts
```diff
@@ -0,0 +1,127 @@
...
+export class ReorderDropzoneManagerFactory {
+	constructor(private branchController: BranchController) {}
+
+	build(branch: Branch, commits: Commit[]) {
+		return new ReorderDropzoneManager(this.branchController, branch, commits);
+	}
+}
+
+// Private classes used to calculate distances between commtis
+class Indexer {
```
Caleb-T-Owens, line 63: I decided to move the indexing and distance calculations into their own classes, because its now on longer required to be publicly exported (like it was when we used the old ReorderDropzone.svelte component)

app/src/lib/branches/dragActions.ts
ndom91: Yeah, makes sense. Definitely sounds like a good next step. However as you said, lets maybe save that for a more full-scale refactor of the whole drag-n-drop system :+1:

app/src/lib/components/CommitDragItem.svelte
ndom91: Yup, makes it more readable too, etc. etc. Good call :+1:

ndom91 (approved):
LGTM functionality-wise!

I have a few small niggles design-wise, but you mentioned you still need to talk to Pavel so I'll just hold off :blush:

Caleb-T-Owens:
Waiting on https://github.com/sveltejs/svelte-eslint-parser/pull/540 in order for our linting to support nested snippets.

PR #4119, opened 2024-06-19
Fix: Wrong throttling in Select components

H1ghBre4k3r:
## ☕️ Reasoning

In our `Select` component, we need to throttle the actions performed on the user input in order to prevent unneeded filtering. While filtering is usually pretty fast, it can lead to performance issues in huge repositories. 

However, previously, we throttled _all_ actions performed after the user performed some input (including storing the actual input). This lead to lost characters if the user typed fast, thus triggering the throttle. This PR changes is behaviour, moving the throttling to only the filtering process, leaving everything else unthrottled.

## 🧢 Changes

- Moved throttling from input handling to filtering

### DIFF
--- a/app/src/lib/components/Select.svelte
+++ b/app/src/lib/components/Select.svelte
@@ -42,7 +42,7 @@
 	let filterText: string | undefined = undefined;
 	let filteredItems: Selectable[] = items;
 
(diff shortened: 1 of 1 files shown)


Caleb-T-Owens:
Thank you greatly for the contribution!

PR #4121, opened 2024-06-19
Fix askpass pipe on windows

Lukylix:
## ☕️ Reasoning
Askpass on windows wasn't working due to 2 small issues :  
- The pipe prefix was using backslash probably causing an escaping issue.  
- The lib windows-named-pipe was trying to connect to the pipe with the wrong shared acces flags.
   (restricted instead of read and write as initialized by tokio named pipes)

## 🧢 Changes
- Changed pipe prefix path to use slashes instead of backslashs on windows.  
- Removed windows-named-pipe dependency.
- Created a minimal file to handle pipe connection.
- Change winapi to windows crate for gitbutler-git.


Byron:

It seems the PR-labeling is failing @ndom91 due to missing permissions. Maybe this could be deactivated as quick fix or if permissions aren't fine-grained enough. From what I have seen, the labeller probably doesn't have a killer-application here just yet.

Besides, this is probably a great PR for @anaisbetts to take a look at. With a little help I could probably also serve as tester on my trusty VM.

ndom91:
@Byron I think I know a way we can fix this, give me a min

ndom91:
@Byron okay nice, seems to work now :tada: 

As you said, there isn't really a killer app for it yet for us, but I do like using it just for organisation or filtering PR's I'm potentially not interested in (like Rust PR's for us "frontend" folks :wink:)

Lukylix:
Thanks 

Just realized that I forgot to close the handle using a drop trait.

But with the try_clone current implementation it could lead to double close therefore an illegal memory access. 
So I also need to call DuplicateHandle inside the try_clone function.

I will do this tomorrow.

Lukylix:
I also made a test for askpass, but I wasn't sure it would fit as I added some test dependencies (assert_cmd) and made the executor public.
 
Here it is:
```rust
use std::time::Duration;
use assert_cmd::Command;
use gitbutler_git::executor::AskpassServer;
use gitbutler_git::executor::GitExecutor;
use gitbutler_git::executor::Socket;
use gitbutler_git::tokio::TokioAskpassServer;

// cargo test --package gitbutler-git --test askpass -- --nocapture
#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
async fn test_askpass() {
    let secret = "super-secret-secret";
    let executor = gitbutler_git::tokio::TokioExecutor;
    let sock_server: TokioAskpassServer = unsafe { executor.create_askpass_server() }
        .await
        .expect("create_askpass_server():");
    let sock_server_string = sock_server.to_string();
    let handle = tokio::spawn(async move {
        let mut cmd = Command::cargo_bin("gitbutler-git-askpass").unwrap();
        let assert = cmd
            .env("GITBUTLER_ASKPASS_PIPE", sock_server_string)
            .env("GITBUTLER_ASKPASS_SECRET", secret)
            .arg("Please enter your password:")
            .assert();
        assert.success().stdout("super_secret_password\n");
    });

    let mut sock = sock_server
        .accept(Some(Duration::from_secs(10)))
        .await
        .expect("accept():");

    let peer_secret = sock.read_line().await.expect("read_line() peer_secret:");

    assert_eq!(peer_secret, secret);

    let prompt = sock.read_line().await.expect("read_line() prompt:");
    assert_eq!(prompt.trim(), "Please enter your password:");

    sock.write_line("super_secret_password")
        .await
        .expect("write_line() password:");
    std::thread::sleep(Duration::from_secs(1));
    handle.await.expect("Askpass command failed");
}
```

crates/gitbutler-git/src/executor/tokio/windows.rs
```diff
@@ -10,30 +10,25 @@ use tokio::{
     net::windows::named_pipe::{NamedPipeServer, ServerOptions},
     sync::Mutex,
 };
+use windows::Win32::{Foundation::HANDLE, System::Pipes::GetNamedPipeClientProcessId};
 
-const ASKPASS_PIPE_PREFIX: &str = r"\\.\pipe\gitbutler-askpass-";
+// Slashes instead of backslashes to prevent any issues with escaping.
+const ASKPASS_PIPE_PREFIX: &str = r"//./pipe/gitbutler-askpass-";
```
anaisbetts, line 16: Out of curiosity, does this have to be a named pipe? Can you capture askpass's stdout instead? Any time a pipe has an explicit name you now have to do a lot of Security Thinking about who can access this pipe, and what happens if someone Bad can mess with it

crates/gitbutler-git/src/bin/askpass/windows-pipe.rs
```diff
@@ -0,0 +1,159 @@
...
+    handle: Handle,
+}
+
+impl Pipe {
+    pub fn connect(path: &Path) -> io::Result<Pipe> {
+        let mut os_str: OsString = path.as_os_str().into();
+        os_str.push("\x00");
+        let mut wide_path: Vec<u16> = os_str.encode_wide().collect();
+
+        let pwstr_path = PWSTR(wide_path.as_mut_ptr());
```
anaisbetts, line 66: I feel like there's got to be some helpers to make these backflips less annoying!

anaisbetts, line 66: I feel like there's got to be some helpers to make these backflips less annoying!

crates/gitbutler-git/src/executor/tokio/windows.rs
Lukylix: The `nMaxInstances` parameter when `CreateNamedPipeW` is called could limit it to only one connection, but this is probably not sufficient for security, and I will need to check the Tokio internals for it.

There is also a permission mechanism in Windows to only allow a specific process handle to connect.
I don't know the details yet, but it's possible.

However, the executable of this process (gitbutler-git-askpass.exe) could also be replaced. 
So, we will need to check the signature of this executable before calling it.

I don't quite understand how to capture the standard output. I am not aware of a solution that could capture only part of the standard output, as Git will interpret any output as a password.  
Could you please elaborate on this? It could be a great solution.

As for non-named pipes, they will be more secure, but I am not sure if GitButler will still be considered the parent process since it will be spawned by Git.

I am sorry I didn't think this through initially. I was only trying to make the current implementation work. :/

crates/gitbutler-git/src/bin/askpass/windows-pipe.rs
Lukylix: Me too, but couldn't find a suitable one, as I learned Rust quite recently.

crates/gitbutler-git/src/executor/tokio/windows.rs
anaisbetts: > I don't quite understand how to capture the standard output. I am not aware of a solution that could capture only part of the standard output, as Git will interpret any output as a password.

Yeah, I don't quite know how GitButler is interacting with askpass, so this might not be workable, I was more writing generally - https://stackoverflow.com/questions/32641423/how-to-access-an-inherited-anonymous-pipe-handle-other-than-stdout-stderr-st has some useful details around anonymous pipes, though if Git is spawning -askpass it might not work, you can always play around with it. 

> I am sorry I didn't think this through initially. I was only trying to make the current implementation work. :/

No worries at all! This can always be improved in a later PR

Lukylix: Could you briefly explain how you secured the Linux side, if you don't mind? 
If this could apply to Windows, it would be a great start for me ! 

I will try to mess around with software signatures and non-named pipes, if possible, or fall back to named pipes with extra security.
Is it needed to check the git signature from various windows providers or is it out of scope and prone to errors ? 

Kinda lost if you could limit my scope this will be great !

There is plenty of stuff i dont know so i cant guarantee my success but i will definitely try in the upcoming days or maybe weeks !

Would love to contribute to the project

Lukylix: I did not refresh the page so i just saw your answer. I will have a look thanks

Byron:
> I also made a test for askpass, but I wasn't sure it would fit as I added some test dependencies (assert_cmd) and made the executor public.
> 
> Here it is:
> 
> ```rust
> use std::time::Duration;
> use assert_cmd::Command;
> use gitbutler_git::executor::AskpassServer;
> use gitbutler_git::executor::GitExecutor;
> use gitbutler_git::executor::Socket;
> use gitbutler_git::tokio::TokioAskpassServer;
> 
> // cargo test --package gitbutler-git --test askpass -- --nocapture
> #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
> async fn test_askpass() {
>     let secret = "super-secret-secret";
>     let executor = gitbutler_git::tokio::TokioExecutor;
>     let sock_server: TokioAskpassServer = unsafe { executor.create_askpass_server() }
>         .await
>         .expect("create_askpass_server():");
>     let sock_server_string = sock_server.to_string();
>     let handle = tokio::spawn(async move {
>         let mut cmd = Command::cargo_bin("gitbutler-git-askpass").unwrap();
>         let assert = cmd
>             .env("GITBUTLER_ASKPASS_PIPE", sock_server_string)
>             .env("GITBUTLER_ASKPASS_SECRET", secret)
>             .arg("Please enter your password:")
>             .assert();
>         assert.success().stdout("super_secret_password\n");
>     });
> 
>     let mut sock = sock_server
>         .accept(Some(Duration::from_secs(10)))
>         .await
>         .expect("accept():");
> 
>     let peer_secret = sock.read_line().await.expect("read_line() peer_secret:");
> 
>     assert_eq!(peer_secret, secret);
> 
>     let prompt = sock.read_line().await.expect("read_line() prompt:");
>     assert_eq!(prompt.trim(), "Please enter your password:");
> 
>     sock.write_line("super_secret_password")
>         .await
>         .expect("write_line() password:");
>     std::thread::sleep(Duration::from_secs(1));
>     handle.await.expect("Askpass command failed");
> }
> ```

I think dev-dependencies are absolutely fine. Further, you should be able to test private APIs by adding a test-module right in the module that has that API.

Feel free to apply the patch below the fold as it integrates the above test like I mentioned.

Further, I wonder if this PR is still a draft (and could be set as such), or if it is ready for (another) review.

Thanks again for your help with this, it's much appreciated!

```patch
index 73cea1430..c6fb3b867 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -163,6 +163,21 @@ version = "0.7.4"
 checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
 
+[[package]]
+name = "assert_cmd"
+version = "2.0.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ed72493ac66d5804837f480ab3766c72bdfab91a65e565fc54fa9e42db0073a8"
+dependencies = [
+ "anstyle",
+ "bstr",
+ "doc-comment",
+ "predicates",
+ "predicates-core",
+ "predicates-tree",
+ "wait-timeout",
+]
+
 [[package]]
 name = "async-broadcast"
 version = "0.5.1"
@@ -1250,6 +1265,12 @@ version = "0.1.13"
 checksum = "56254986775e3233ffa9c4d7d3faaf6d36a2c09d30b20687e9f88bc8bafc16c8"
 
+[[package]]
+name = "difflib"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6184e33543162437515c2e2b48714794e37845ec9851711914eec9d308f6ebe8"
+
 [[package]]
 name = "diffy"
 version = "0.3.0"
@@ -1339,6 +1360,12 @@ version = "0.2.0"
 checksum = "bd0c93bb4b0c6d9b77f4435b0ae98c24d17f1c45b2ff844c6151a07256ca923b"
 
+[[package]]
+name = "doc-comment"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fea41bba32d969b513997752735605054bc0dfa92b4c56bf1189f2e174be7a10"
+
 [[package]]
 name = "dtoa"
 version = "1.0.9"
@@ -2165,6 +2192,7 @@ dependencies = [
 name = "gitbutler-git"
 version = "0.0.0"
 dependencies = [
+ "assert_cmd",
  "futures",
  "nix 0.29.0",
  "rand 0.8.5",
@@ -4870,6 +4898,33 @@ version = "0.1.1"
 checksum = "925383efa346730478fb4838dbe9137d2a47675ad789c546d150a6e1dd4ab31c"
 
+[[package]]
+name = "predicates"
+version = "3.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "68b87bfd4605926cdfefc1c3b5f8fe560e3feca9d5552cf68c466d3d8236c7e8"
+dependencies = [
+ "anstyle",
+ "difflib",
+ "predicates-core",
+]
+
+[[package]]
+name = "predicates-core"
+version = "1.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b794032607612e7abeb4db69adb4e33590fa6cf1149e95fd7cb00e634b92f174"
+
+[[package]]
+name = "predicates-tree"
+version = "1.0.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "368ba315fb8c5052ab692e68a0eefec6ec57b23a36959c14496f0b0df2c0cecf"
+dependencies = [
+ "predicates-core",
+ "termtree",
+]
+
 [[package]]
 name = "pretty_assertions"
 version = "1.4.0"
@@ -6583,6 +6638,12 @@ dependencies = [
  "utf-8",
 ]
 
+[[package]]
+name = "termtree"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3369f5ac52d5eb6ab48c6b4ffdc8efbcad6b89c765749064ba298f2c68a16a76"
+
 [[package]]
 name = "thin-slice"
 version = "0.1.1"
@@ -7133,6 +7194,15 @@ dependencies = [
  "libc",
 ]
 
+[[package]]
+name = "wait-timeout"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9f200f5b12eb75f8c1ed65abd4b2db8a6e1b138a20de009dacee265a2498f3f6"
+dependencies = [
+ "libc",
+]
+
 [[package]]
 name = "waker-fn"
 version = "1.1.1"
diff --git a/crates/gitbutler-git/Cargo.toml b/crates/gitbutler-git/Cargo.toml
index 2f07a3dbc..b88b78b3d 100644
--- a/crates/gitbutler-git/Cargo.toml
+++ b/crates/gitbutler-git/Cargo.toml
@@ -40,6 +40,10 @@ rand = "0.8.5"
 futures = "0.3.30"
 sysinfo = "0.30.12"
 
+[dev-dependencies]
+assert_cmd = "2.0.14"
+tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
+
 [target."cfg(unix)".dependencies]
 nix = { version = "0.29.0", features = ["process", "socket", "user"] }
 
diff --git a/crates/gitbutler-git/src/executor/tokio/mod.rs b/crates/gitbutler-git/src/executor/tokio/mod.rs
index 980034f1b..8472162d7 100644
--- a/crates/gitbutler-git/src/executor/tokio/mod.rs
+++ b/crates/gitbutler-git/src/executor/tokio/mod.rs
@@ -138,3 +138,53 @@ unsafe impl super::GitExecutor for TokioExecutor {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use crate::executor::AskpassServer;
+    use crate::executor::GitExecutor;
+    use crate::executor::Socket;
+    use assert_cmd::Command;
+    use std::time::Duration;
+
+    // cargo test --package gitbutler-git --test askpass -- --nocapture
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_askpass() {
+        let secret = "super-secret-secret";
+        let executor = TokioExecutor;
+        #[allow(unsafe_code)]
+        let sock_server: TokioAskpassServer = unsafe { executor.create_askpass_server() }
+            .await
+            .expect("create_askpass_server():");
+        let sock_server_string = sock_server.to_string();
+        let handle = tokio::spawn(async move {
+            let mut cmd = Command::cargo_bin("gitbutler-git-askpass").unwrap();
+            let assert = cmd
+                .env("GITBUTLER_ASKPASS_PIPE", sock_server_string)
+                .env("GITBUTLER_ASKPASS_SECRET", secret)
+                .arg("Please enter your password:")
+                .assert();
+            assert.success().stdout("super_secret_password\n");
+        });
+
+        let mut sock = sock_server
+            .accept(Some(Duration::from_secs(10)))
+            .await
+            .expect("accept():");
+
+        let peer_secret = sock.read_line().await.expect("read_line() peer_secret:");
+
+        assert_eq!(peer_secret, secret);
+
+        let prompt = sock.read_line().await.expect("read_line() prompt:");
+        assert_eq!(prompt.trim(), "Please enter your password:");
+
+        sock.write_line("super_secret_password")
+            .await
+            .expect("write_line() password:");
+        std::thread::sleep(Duration::from_secs(1));
+        handle.await.expect("Askpass command failed");
+    }
+}
```

</details>

crates/gitbutler-git/src/executor/tokio/windows.rs
Lukylix: > Yeah, I don't quite know how GitButler is interacting with askpass, so this might not be workable, I was more writing generally - https://stackoverflow.com/questions/32641423/how-to-access-an-inherited-anonymous-pipe-handle-other-than-stdout-stderr-st has some useful details around anonymous pipes, though if Git is spawning -askpass it might not work, you can always play around with it.

I can't pass any arguments to gitbutler-git-askpass or even write to its standard input. However, I could take inspiration from it and pass the non-named-pipe handle as an environment variable, but I don't know the scope of environment variables when initialized inside Gitbutler for security purposes.
I'm also unsure if the handle inheritance will work (as it's started by Git, not Gitbutler).

Should I take this approach?

Or should I keep using the named pipe? However, I don't see how I could get the right process handle to limit the permission of the pipe to it (since it's started by Git, not Gitbutler).

Alternatively, should I just limit the number of connections to the named pipe to 1 ?
As gitbutler-git-askpass will probably be the first to connect to it as the pipe name at the end is randomly generated.

How did you handle this security concern of the linux side ? Maybe i could try to reproduce it.

## Facultative reading
I think there is too many way a bad actor could mess with any of the following systems.
Process injection, Manual mapping, replacing windows api functions, git, gitbutler-askpass or even a simple Keylogger and many more ... 

I also want this process to be secure as possible but even simple attack vectors will just work on any solution mentioned :/

Unfortunatly i feel like on windows there not much we can do as a simple keylogger using the standard library doesn't even require admin privilege. (This is not the case on linux)

The best we can do is assume the AV will do a proper work and trust windows functions.

But i agree the current solution need at least an extrat step. (As i dont think we limit the pipe acces strongly enough)

Sorry for this reply i just look like a depressed dev 🤠

Lukylix:
> I think dev-dependencies are absolutely fine. Further, you should be able to test private APIs by adding a test-module right in the module that has that API.
> 
> Feel free to apply the patch below the fold as it integrates the above test like I mentioned.
> 
> Further, I wonder if this PR is still a draft (and could be set as such), or if it is ready for (another) review.
> 
> Thanks again for your help with this, it's much appreciated!

Thanks i will definitely apply your patch !

Happy to help !

crates/gitbutler-git/src/executor/tokio/windows.rs
Lukylix: Hey !

I started implementing the non named pipe and i feels kind wrong to convert a pointer to string and vice versa.
This introduce some usage overhead (make sure the pointer is not a dangling pointer due to the rust borrowing rules ect). Even if for now it work (I havent checked the process inheritance when spawned by git).

I feel like it doesn't add more than limiting the pipe connection to only 1 instance.
As if a bad actor connect to the pipe first i will not have the secret passed in the scoped environment and will not receive the password as such.

Let me know if this it's still your preferred solution and i will keep working on it.

Lukylix: So i tried to use a non named pipe and i keep getting access violation errors as soon as the process is created with rust. The bypass would be to create the process directly with CreateProcess and the bInheritHandles option on windows.
But we also got no warranty that git will create the process with the same option.

So i guess i will just see how to limit the number of named pipe connection with tokio.

crates/gitbutler-git/src/bin/askpass/windows-pipe.rs
Byron: I have seen it done like this as well. Of course there would be no harm in [~~reusing some code~~ taking inspiration](https://github.com/Byron/open-rs/blob/main/src/windows.rs#L83) to make this less boilerplaty.

krlvi:
Thank you @Lukylix for this contribution! I tested it on parallels and seems to work fine. Also took the liberty of solving the merge conflicts so that we can merge this and test on the nightly

PR #4130, opened 2024-06-20
feat: Trunk search

estib-vega:
Naive search of the trunk branch by commit message.

## ☕️ Reasoning
Was thinking that there is still some areas of opportunity wrt git-archeology 🤠 and thought the best place to start would be to have a nice and simple commit search bar.

This is a very simple filter-by-commit-message contents.
But down-the-line we could start adding a way of filtering by remote/local, author, files, etc.
All that is a bit more design-heavy and so probably requires a bit more of a thorough conversation about it.

In any case, I think this already bring some value to the archeology side of GitButler.

## 🧢 Changes
- Add a search bar at the top of the trunk page, that receives the filter string
- How the resizing of the lanes behave (not sure if that's what is wanted)

## GIFS 👾

This is how searching works. Searching does not close the file viewer (that's by design but I can change it).
Also, the button to merge into common base disappears on search.
And there is an empty state with the message *"No commits found that match the current search"*
naive-trunk-search

The second change mentioned is there so that the commit lane doesn't resize when being filtered, but also I'd say that's a bit more how I'd expect it to work, even without the filtering mechanism. Happy to change that if you think otherwise.

How it was before (you can't even resize it without first opening the file viewer I think):
resizing-now

How it is implemented in this change:
resize-change

### DIFF
--- a/app/src/lib/components/BaseBranch.svelte
+++ b/app/src/lib/components/BaseBranch.svelte
@@ -9,9 +9,10 @@
 	import { getContext } from '$lib/utils/context';
 	import { tooltip } from '$lib/utils/tooltip';
 	import { BranchController } from '$lib/vbranches/branchController';
-	import type { BaseBranch } from '$lib/vbranches/types';
+	import type { BaseBranch, RemoteCommit } from '$lib/vbranches/types';
 
 	export let base: BaseBranch;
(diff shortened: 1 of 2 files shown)


estib-vega:
Closing in favor of https://github.com/gitbutlerapp/gitbutler/pull/4157

PR #4143, opened 2024-06-21
Redo line drawing

Caleb-T-Owens:
### Updating terminology

In the application we would previously use "upstream" to refer to a commit that was only available on the remote git server. These are now called "remote" commits.

In the application we would previously use "remote" to refer to a commit that was available on the git server and locally. These are now called "localAndRemote" commits.

### StoryBook

This PR introduces StoryBook as a way of developing a component in a sandbox environment, if it makes sense.

This is not meant to be a new methodology or intended to be used as a testing framework, but a handy tool if it makes sense for what you are working on.

### Commit lines

For rendering lines, we now pre-compute the relevant data in the LineManager class. This data then gets rendered by svelte. This means that we are able to use mutable state while determining where lines should be drawn.

Examples of this data structure can be seen in https://github.com/gitbutlerapp/gitbutler/blob/7a868a484a7e09a53f189c89b2d1f86bc1455bd8/app/src/stories/commitLines/DemoStaticCommitLines.stories.ts.

image


PavelLaptev:
@Caleb-T-Owens there are a few UI things:

image

1. The dash line is different. The new dashed line has wider dashes.
2. The base icon is not visible in the new version
3. There is a big gap between the left container edge and the commit line. Looks like an empty line

image

4. Avatars should be vertically centered with titles.

Let me know if you need some help here

ndom91:
Interesting.. some of mine align the avatar with the title, but not all.

But I also noticed the initial curve is a bit janky

image

ndom91:
Otherwise, functionally it LGTM and storybook works nicely!

app/package.json
```diff
@@ -23,9 +23,12 @@
 		"format": "prettier --write .",
 		"fix": "eslint --fix .",
 		"tauri": "tauri",
-		"prepare": "svelte-kit sync"
+		"prepare": "svelte-kit sync",
+		"storybook": "storybook dev -p 6006",
+		"build-storybook": "storybook build"
```
ndom91:
```suggestion
		"build:storybook": "storybook build"
```

ndom91 (dismissed):
Small nit

Caleb-T-Owens:
@PavelLaptev and @ndom91 I have corrected the alignment issues that you highlighted

PavelLaptev:
> @PavelLaptev and @ndom91 I have corrected the alignment issues that you highlighted

Thanks! Looks good 🙂 
I'm checking the rest of the UI 👀

Byron:
> *This is absolutely magical to me! Like, I have absolutely no clue how it works, yet it does :D. Fascinating!*

Anonymous Rust Developer

PavelLaptev:
@Caleb-T-Owens, I hope you don't mind if I push directly into your PR 🫣

PavelLaptev (approved):
Nice work @Caleb-T-Owens 😎

Issue #4132, opened 2024-06-20
Failed to load Pull Requests "no repo name specified"


Arilas:
### Version

0.12.5

### Operating System

Mac OS X

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

In a repo, that hosted on a GitHub I cannot create a PR or load a PRs list.

Additional details:
1. remoteUrl is a ssh string with github alias (Because I have multiple GitHub ssh keys, so I've made an alias). Example: `git@arilas.github.com:Arilas/my-repo.git`
2. `GitHubService` line 81: it's incorrectly parse this remoteUrl, so repo is undefined, and owner is `git@arilas`

![image](https://github.com/gitbutlerapp/gitbutler/assets/3540524/189fc347-905a-4c85-9e8e-3d2612efa7bb)

### How to reproduce

1. Create an alias for github in your ssh/config file 
2. Clone a repo using this alias
3. Try to open GitButler
4. Observe errors

### Expected behavior

It should ask user about repo or at least try to parse a string

Byron:
Thanks a lot for reporting!

Indeed, the code that extracts information form the URL [is here](https://github.com/gitbutlerapp/gitbutler/blob/f2da13ea83d93b473c45d601d80f71f1b601069f/app/src/lib/github/service.ts#L80-L84) and it makes more assumptions than it should.

A quick solution *should* be to use [gix_url::parse](https://docs.rs/gix-url/0.27.3/gix_url/fn.parse.html) (also accessible by `gix::url::parse()` in the backend as it should be able to correctly parse this kind of URL as well.

Maybe @ndom91 could also just fix it in the frontend, but be warned that it is a can of worms to do this correctly, thus it's best to outsource this (or one would have to reimplement [this](https://github.com/git/git/blob/master/urlmatch.c#L112)).

Arilas:
@Byron I've made PR which fixes it for me, can you check?

ndom91:
@Arilas that seems like a reasonable quick-fix, but as @Byron said, long-term leveraging a library that handles all the edge cases would be best. 

Imo we can merge the quick-fix for now, and I can bring up using the `gix` methods to parse the remote strings with some of the rust folks here next week. What do you think?

PR #4147, opened 2024-06-21
fix: improve git remoteUrl parsing

ndom91:
## ☕️ Reasoning

- We weren't parsing some different types/variants of git remote URLs
- https://github.com/gitbutlerapp/gitbutler/issues/4132

## 🧢 Changes

- New dependency `git-url-parse` (https://github.com/IonicaBizau/git-url-parse) [`~3.5kb minified + gzipped`]
	- GitHub only supports a subset of the remote URL types that `git` supports, so maybe the manual parsing we had was sufficient for the time being
	- But having less error-prone / battle tested parsing will make it easier for us to support other git services in the future and avoid other edge-cases
- Add a few initial vitest tests for the `GitHubService` to ensure that `owner`/`repo` are parsed out as expected

Out of the 22 URLs I added to test, the previous implementation only parsed 13/22 correctly.


app/src/lib/github/service.test.ts
```diff
@@ -0,0 +1,42 @@
+import { GitHubService } from './service';
+import { BehaviorSubject } from 'rxjs';
+import { expect, test, describe } from 'vitest';
+
+const exampleRemoteUrls = [
+	'ssh://user@host.xz:123/org/repo.git/',
```
Byron, line 6: I am loving this!

Byron (commented):
That's a great call, and of course the JS world has a parser already which probably covers 99% of the common cases as well by now.

So I'd say, despite my recommendation, let's use it until it breaks :).

Caleb-T-Owens:
Love this @ndom91!! 

If you're up for it, do you want to change `convertRemoteToWebUrl` to something like: `GitUrlParse("...").toString("https").replace(/.git$/, '')`?

Caleb-T-Owens (approved):

PR #4150, opened 2024-06-21
Persist FileListItem checkbox detail

mental32:
## ☕️ Reasoning

There are multiple lanes on the board and I tend to work on the left-most one hosting a lot of changed files. Eventually I would like to make a commit but maybe only 1/3rd of the files in my file list are relevant so I: "start commit", unselect the other 2/3rds and notice there's actually a file or two in a lane to the right I would also like to include in this WIP commit BUT on file 'drop' the entire file list was reselected and added to `selectedOwnership`, which is very annoying to deal with.

## 🧢 Changes

Changes are all in the FileListItem component it tracks the last state of the checkbox detail and attempts to make sure to exclude the file from the `selectedOwnership` set to keep it coherent.

### DIFF
--- a/app/src/lib/components/FileListItem.svelte
+++ b/app/src/lib/components/FileListItem.svelte
@@ -30,9 +30,10 @@
 
 	let checked = false;
 	let draggableElt: HTMLDivElement;
+	let lastCheckboxDetail = true;
 
 	$: if (file && $selectedOwnership) {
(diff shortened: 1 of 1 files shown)


krlvi (approved):
Hey @mental32, thanks for contributing this! Just tested it myself and seems to work just fine 👍 
Gonna have this in the next release :)

PavelLaptev:
@mental32, thank you for the nice improvement. Works great!

PR #4165, opened 2024-06-24
migrate plain-text secrets to keystore

Byron:
Remove plain-text secrets from various configuration files and put them into the keystore instead.

This is a similar move as the one done by the `gh` command-line tool recently, which now also stores
its secrets (like the GitHub authentication token) in the keystore.

### Tasks

* [x] migrate user-related tokens
    - [x] assure different build types (i.e. nightly, dev, production) are reasonably isolated
    - [x] assure tests are isolated
    - [x] assure `Sensitive` fields won't be serialized (but they need to be deserialized for migration)
    - [x] assure frontend still sees secrets
    - [x] for safety, delete the user if the access token isn't in the secrets store anymore
* [x] setting the password to an empty string should delete it from the keychain
     - (even though empty string is fine for MacOS at least, but that might not be the case everywhere)
* [x] upgrade secrets storage to avoid keyring popups on MacOS
    - [x] https://github.com/Byron/gitoxide/pull/1427
    - [x] use gix from main with explicit rev
* [x] migrate AI related tokens
    * [x] assure that these settings are truly global, with secrets spanning all application kinds
* [x] validate on Windows (access_token only)
* [x] validate on Ubunutu
* [x] validate on MacOS (access_token and github token)
* [x] cleanup UI

### Platform Notes

* **On MacOS**  - Dev Mode
    - it asks for access to the secret each time the binary changes. `gh` doesn't have this problem, as it appears to use the Git credential helpers (which are stable, binary-wise).
    - With the recent updates to `gitoxide` it finds git executables more thoroughly, enabling credentials to be used. I think the custom keychain should be the level-2 store, with git-credentials being the one tried first.
* **On MacOS**  - Nightly Build
    - No popup was observed, not even after a restart.
    - a github login worked fine as well (and apparently, empty passwords are allowed)
    - auto-deletion of GItHub credentials works as well
    - This is how secrets look like now, so each entry is easily identifiable.
Screenshot 2024-06-26 at 17 35 03

* **On Windows**
    - login doesn't work, it says there was an error in the browser (edge). probably something about the backend that it uses there.
    - using a dev-user.json from MacOS on Windows does work though, and it successfully creates and deletes secrets in ~~whatever secret-store Windows has~~ the Credential Manager. I.e. it can remove the secret from the plain file, access it from the secret store, and delete it from the secret store once the user is signed out.
Screenshot 2024-06-26 at 15 07 41

* **On Ubuntu 22.04**
    - The login doesn't work, similarly to Windows. It's probably the same, with :`https://auth.gitbutler.com/login/callback?error=redirect_uri_mismatch&error_description=The+redirect_uri+MUST+match+the+registered+callback+URL+for+this+application.&error_uri=https%3A%2F%2Fdocs.github.com%2Fapps%2Fmanaging-oauth-apps%2Ftroubleshooting-authorization-request-errors%2F%23redirect-uri-mismatch&state=IqB-c-28IA0egwI-dqs39UEY_7ZpZ0Zn`
    - passwords manager access works without popup, tried read, write and deletion via GitButler.
    - There is no popups, despite rebuilds and thus changed binaries. 
Screenshot 2024-06-26 at 16 39 26

### Notes for the Reviewer

* ❗️The secrets implementation on MacOS is different between dev and nightly/production builds. For proper testing, one must also build the nightly locally.
    - On other platforms, it's fine to use the dev-build for testing - they all use the system keychain.
    - On MacOS, dev build, a git-credential based keystore implementation needed to prevent nagging popups.
* ❗️Windows and Linux testing was done with an older version that stored passwords under different names, but since the keychain implementation didn't change I'd assume it works like before. However, I only logged in, so never created multiple secrets. That might be something to test as well, easiest by setting the OpenAI and Antrophic API keys.
* **Frontend**
    - The frontend still receives a struct just like before, with secrets being mere strings. For this PR I avoided changes there, but it would certainly be advisable to develop some strategy on how to handle secrets in the frontend. Right now it's easy to spill them, or to send them to remote locations (i.e. log them). In Rust, that's impossible now.
    - There is some duplication and the most naive implementation that makes this work you can imagine. My recommendation (and hope) is that the cleanup can be done by someone who has a proper dev setup for this and… is more experience in the ways of the frontend :).
    -  I could have imagined to use a real implementation of the respective service and mock the invoke call instead. But since this wasn't done and I wanted to keep my frontend time short, I just implemented it directly without creating a `SecretService` with respective mock-tests using that. However, this is probably what should be created at some point once more secrets are used.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2246,8 +2246,10 @@ dependencies = [
  "anyhow",
  "git2",
  "gitbutler-core",
+ "keyring",
  "once_cell",
  "pretty_assertions",
+ "serde_json",
  "tempfile",
 ]
 
@@ -5759,9 +5761,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.117"
+version = "1.0.118"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "455182ea6142b14f93f4bc5320a2b31c1f266b66a4a5c858b013302a5d8cbfc3"
+checksum = "d947f6b3163d8857ea16c4fa0dd4840d52f3041039a85decd46867eb1abef2e4"
 dependencies = [
  "indexmap 2.2.6",
  "itoa 1.0.11",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2172,7 +2172,7 @@ name = "gitbutler-git"
 version = "0.0.0"
 dependencies = [
  "futures",
- "gix-path",
+ "gix-path 0.10.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "nix 0.29.0",
  "rand 0.8.5",
  "serde",
@@ -2273,14 +2273,14 @@ dependencies = [
 [[package]]
 name = "gix"
 version = "0.63.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "984c5018adfa7a4536ade67990b3ebc6e11ab57b3d6cd9968de0947ca99b4b06"
+source = "git+https://github.com/Byron/gitoxide?rev=55cbc1b9d6f210298a86502a7f20f9736c7e963e#55cbc1b9d6f210298a86502a7f20f9736c7e963e"
 dependencies = [
  "gix-actor",
  "gix-attributes",
  "gix-command",
  "gix-commitgraph",
  "gix-config",
+ "gix-credentials",
  "gix-date",
  "gix-diff",
  "gix-dir",
@@ -2295,11 +2295,13 @@ dependencies = [
  "gix-index",
  "gix-lock",
  "gix-macros",
+ "gix-negotiate",
  "gix-object",
  "gix-odb",
  "gix-pack",
- "gix-path",
+ "gix-path 0.10.8 (git+https://github.com/Byron/gitoxide?rev=55cbc1b9d6f210298a86502a7f20f9736c7e963e)",
  "gix-pathspec",
+ "gix-prompt",
  "gix-ref",
  "gix-refspec",
  "gix-revision",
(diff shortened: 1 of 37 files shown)


Byron:
I declare it ready for review despite the Frontend-based CI failures. @mtsgrd offered help in turning the naive current implementation into something maintainable much faster and better than I could do it. Thus, I will stop force-pushing here and will at most push fixes on top if @krlvi finds issues during testing.

mtsgrd:
@Byron thanks, I've pushed a commit splitting out a separate service for the secrets. Could you have a look and make sure things still work as expected?

Byron:
Thanks a lot for cleaning up the UI parts and for doing so so swiftly!
I found and fixed a couple of small issues during testing, with the highlight being a race condition when typing into the key input field quickly (not that this was always present, I just didn't notice it before).

Further, I could validate migration as well and it seemed to all work on MacOS, from where we we can assume it will also work on other platforms due to the portability of the frontend.

Regarding CI, the GitGuardian failure is clearly a false positive, and I don't know how to fix that.

Lastly, I didn't merge `master` nor did I rebase to deal with the easily resolvable conflict yet, but would rebase once the code is approved, at the latest. If merges are fine, which I am unsure of, I can do that right away, please let me know.

mtsgrd:
Feel free to merge, we catch more bugs and earlier if we have varied workflows. ✌️ 

About GitGuardian, I think let's just change the test secret to something low entropy like `test-key`?

Byron:
Great, I will rebase then to fix the merge conflict.

Regarding GitGuardian, what's confusing me is that it points out a [`describe`](https://github.com/gitbutlerapp/gitbutler/commit/b0ababbabf1164b8e83b923c3d124b2780d743cf#diff-806b07fa44fb9bbaf589ff8acacfb1e2f53a8091b7ae716b7e0100e0c72a4a5aL186) block in service test, which does not at all resemble a secret.

Byron:
And I just noticed that there was a UI linting failure which hopefully is addressed now.
The *real* GitGuardian issue was a duplicate line `186` (or its inability to address the right one), which confused me into thinking that it must be a bug. It did try point at a quite high-entropy key which has now been replaced so it passes the check. Thanks @mtsgrd for pointing this out as your comment convinced me to look again.

As suggested, this PR will be merged once CI is green.

Next up for me will be to purge logs older than X days for simplicity, a value that could one day be hooked up to the UI as well.

PR #4168, opened 2024-06-24
Fix bug in hunk -> branch assignment

mtsgrd:
- return early on lock only if hunk matches or intersects


mtsgrd:
Fixes #4168

Caleb-T-Owens:
@mtsgrd you lovely lovely man!! Thanks for fixing this

krlvi (approved):
👍

PR #4174, opened 2024-06-25
cool with submodules

krlvi:


Byron:
When trying a repository with submodules, I am seeing errors still. When trying to add `with-submodules`, it fails. but when trying to add the submodule itself, i.e. `with-submodules/zip`, it enters an add-loop with the project being in the list, but I can't get past the 'Lets Go' button.

[with-submodules.zip](https://github.com/user-attachments/files/15984491/with-submodules.zip)

I hope the archive here does reproduce the two issues for you, @krlvi . *(I decided to post this here to not add more noise to the issue)*

PR #4192, opened 2024-06-26
feat: create separate component library subpackage

ndom91:
## ☕️ Reasoning

- Initial "dumb components" UI Library starter + storybook

## 🧢 Changes

- Create initial `packages/ui/*` structure for component library and future ui components package
- Based off of svelte-create CLI "create me a library to eventually publish" option
	- Going to leave this here: https://kit.svelte.dev/docs/packaging
	- `svelte-package` will bundle up the stuff in `src/lib` by default to consume/ship, anything in `src/routes` is designed to be used as a demo app / development app for the library contained in the `lib` dir, so long story short not needed by us atm.
- Renamed the existing app from `@gitbutler/ui` -> `@gitbutler/app`, and the new ui package is then `@gitbutler/ui`
	- `@gitbutler/app` is consuming `@gitbutler/ui`, which currently just consists of the `CommitLines/*` components
- Moved all Storybook stuff over to the `@gitbutler/ui` package
	- Copied @Caleb-T-Owens's first few stories for the CommitLines components there too
- Centralized the `.prettierrc`, `eslint.config.js`, and `.gitignore` in the root to apply to all packages/apps
- Storybook is setup to potentially be published in the future, if we wanted a hosted version of the component library either for ourselves or to share with the world maybe
- Added [turborepo](https://turbo.build) to orchestrate the monorepo tasks (i.e. UI kit has to be built before it can be consumed by the main app, etc.)

Basically, based off of our discussions the other day, the idea is to be able to put "dumb components" here and be able to package them up and publish them on their own, to be consumed else where. 

### Unrelated Changes

- Removed a few unused FE dependencies (`mm-jsr`, `nanoid`, `leven`, `crypto-js`).

### Future

- We could package and publish `@gitbutler/ui` to a package repository if we wanted. Currently we're just consuming it via pnpm's `workspace` feature.
	- If we want to publish, I'd highly recommend adding [publint](https://publint.dev) as a quick lint task after packaging, before publishing, i.e. [example](https://github.com/radix-svelte/radix-svelte/blob/develop/package.json#L9)
- Currently I'ved just copied over the 3 util files required by the components in the UI kit (can be found under `packages/ui/src/lib/utils/*`) as well as the styles (`packages/ui/src/styles/*`). Pavel mentioned already he wants to take a look at refactoring the styles.


PavelLaptev (approved):

ndom91:
Quick status update: So @Caleb-T-Owens and I spent a good amt of time last night debugging this stupid export map issue and finally figured it out :joy:. We also got [turborepo](https://turbo.build) setup to orchestrate the monorepo tasks (i.e. UI library needs to be built before app can run, stuff like that).

The only issue remaining now is that when running everything in dev mode (`pnpm tauri dev` in the root still), `turbo` runs the UI kit in watch mode to rebuild any changes immediately, it also runs the dev app in its normal watch mode, howeverrr HMR seems to be "broken" on the frontend side, it always does full reloads.

Caleb-T-Owens:
We've chosen to merge at this point and incrementally add to the setup in the future.

The most obvious avenue for improvement is just moving more components to the UI library. We should make an effort to where possible create components in a UI first manner.

The other avenue for improvement is to add separate turbo files to each package. This will allow us to have caching for those build steps. When caching is enabled, you need to manually declare all of the output files

PR #4222, opened 2024-07-02
fix: cleanup `hooks.client.ts` `handleError` fn

ndom91:
## 🧢 Changes

- Remove duplicate unhandled error handling
	- The linked issue regarding sveltejs error boundaries seems like it's assuming that the user is using _just_ svelte, and not **sveltekit**. In sveltekit the `handleError` fn in `hooks.client.ts` should handle [all unhandled errors](https://kit.svelte.dev/docs/hooks#shared-hooks-handleerror).
- Remove duplicate console.logs 
- Use Sentry's `captureException` instead of `handleErrorWithSentry` and pass the returned `errorId` along via the return from `handleError` to the FE (via `$page.error`).
- In the FE we can then `showError` with the `error.message` as body to the users.
- Long story short, this is a step toward not hiding errors from the user and passing on the `error.message` to the user.


ndom91:
@mtsgrd curious what you think of this.

It's basically just using the built-in `handleError` function from sveltekit to catch any last unhandled errors, send them to Sentry, and then pass on the sentry defined ID and `error.message` to the FE via the `$page` store. Where we can then `showError` to surface the errors to the users.

I'm a little worried this will potentially surface too many errors, but we did talk about wanting to give our users more actionable / real error messages.

Also sidenote, it might be useful to also surface that `errorId` on the user side so if people reach out they could mention it and we can then easily look that up on Sentry.

Final question, it looks like we were doing this twice previously, once via `window.onunhandledrejection` and another time via the `handleError` sveltekit functionality, is there a reason for that? :thinking:  I think the issue that was linked next to the window handler about error boundaries was specific to svelte only, not sveltekit, which does ship with this hooks file and `handleError` functionality.

mtsgrd:
This looks great! And yes, I think it would be helpful to surface the error id. Since you're likely offline for the rest of the week I might fix the checks and land this. 🙌

ndom91:
> This looks great! And yes, I think it would be helpful to surface the error id. Since you're likely offline for the rest of the week I might fix the checks and land this. 🙌

I'm back, I can continue with this. 

So far the todo's look like this:

- Expose `errorId` to user as well
- Fix GHA / lint errors

Anything else?

app/src/routes/+layout.svelte
```diff
@@ -62,6 +64,13 @@
 	$: userSettings.update((s) => ({ ...s, zoom: zoom }));
 
 	onMount(() => {
+		if ($page.error?.message) {
+			let message = $page.error.message;
+			if ($page.error.errorId) {
+				message += `\n\nError ID: ${$page.error.errorId}`;
+			}
+			showError('There was a problem', message);
+		}
```
mtsgrd, line 73: Actually, one thing. I'm unsure if this is correctly placed within `onMount`?

app/src/hooks.client.ts
```diff
@@ -1,25 +1,25 @@
...
-import type { NavigationEvent } from '@sveltejs/kit';
+import type { HandleClientError } from '@sveltejs/kit';
 
-function myErrorHandler({ error, event }: { error: any; event: NavigationEvent }) {
-	console.error(error.message + '\n' + error.stack);
-	logErrorToFile(error.toString());
-	console.error('An error occurred on the client side:', error, event);
-}
+// eslint-disable-next-line func-style
+export const handleError: HandleClientError = ({ error, status }) => {
```
mtsgrd: Change this to `export function handleError({ error: unknown, status: number }): HandleClientError`?

app/src/hooks.client.ts
```diff
@@ -1,25 +1,25 @@
...
+	let errorId: string = crypto.randomUUID();
```
mtsgrd, line 12: On Safari this crypto method [seems to only be available from March 2022](https://caniuse.com/mdn-api_crypto_randomuuid), is there a more compatible way we could generate this ID?

app/src/hooks.client.ts
ndom91: Yeah that works too, unfortunately we can't use the full function type declaration (incl. parameters) with the function declaration style, which is why I went this route ^^.

But `error` and `status` here are pretty straight forward and the return can just be `ReturnType<HandleClientError>` :+1:

app/src/hooks.client.ts
ndom91: So it was added in Safari `15.4` it looks like, right. It seems 99%+ of our users are on Safari 16+, check out on posthog.

That top 52.9% are seemingly using some useragent cloaking privacy feature of Mac OS X / Safari, but Caleb on his up-to-date Macbook had the same useragent. The rest are Windows users with Edge webview / Chromium 126 (latest version), and then Linux folks with webkitgtk that seems to advertise as Safari 17 / Safari 16.

So long story short I think we're probably good here. But, if you want to be on the safe side, this ID doesn't really need to be super collision resistant, right, plus it'll most likely be overwritten by the Sentry generated ID anyway, so I could swap it out with some `Math.random()`-based solution, just let me know :+1:

app/src/routes/+layout.svelte
ndom91: Hmm so my thinking was that after we've mounted and done the first render, the `onMount` hook would be called we'd check the error store and then pop up the `showError` alert then. Instead of having the alert be rendered with everything initially. It probably doesn't matter technically though, right. What do you wanna do? I think popping it up afterward is more attention grabbing, compared to having the toast be a part of the initial content, you know what I mean?

app/src/hooks.client.ts
mtsgrd: This might be related to the top 52.9%: https://chromestatus.com/feature/5452592194781184
> Cap the reported macOS version in the user-agent string at 10_15_7

I also had a look at the data and it seems to me that we don't actually have any version information available for macos? The strings _with_ version data all appear to be from linux platforms.

@PavelLaptev do you recall what the oldest version is we have encountered wrt to CSS bugs?

app/src/routes/+layout.svelte
mtsgrd: You're right, that sounds better.

app/src/hooks.client.ts
ndom91: Yeah interesting, so all Mac platforms seem to be wrapped up in that one capped user-agent string :thinking: 

I did a bit more research, and it seems that there's often another suffix on that user-agent string we're dealing with which includes more info (See for example: https://github.com/nielsbasjes/yauaa/blob/main/analyzer/src/main/resources/UserAgents/Apple.yaml#L1704). But that seems totally missing in our case.

Anyway, gonna to merge this PR now :+1:

PR #4226, opened 2024-07-02
fix: cleanup GHA setup for nightly builds

ndom91:
## ☕️ Reasoning

- Confusing additional 'nightly build' which doesn't use the target branch when run manually

## 🧢 Changes

- Remove `nightly.yaml` job
- Add an additional `on:` parameter to pre-existing `publish.yaml` job to run at `03:00` nightly to generate nightly build and replace functionality of `nightly.yaml`.
- When a job is triggered by anything other than `workflow_dispatch`, those input parameters are undefined. So you need to set a fallback for them which should apply for all other triggering scenarios (i.e. cron at 03:00).
	- They then default to `bump = 'patch'` and `channel = 'nightly'` when run from the cron trigger
- **To manually build nightly jobs going forward then**, go to `Publish` job and select your branch of choice and select `bump = 'patch'` and `channel = 'nightly'`. 

See: https://docs.github.com/en/actions/learn-github-actions/expressions#example


krlvi (approved):
lgtm

PR #4234, opened 2024-07-03
Start capturing branch metrics

mtsgrd:
- service keeps count
- component polls every 15 minutes
- capture only when metrics have changed
- capture once a day if no changes


krlvi (approved):
seems reasonable 👍

mtsgrd:
A'ight, I changed the variable names to reflect [best practice](https://posthog.com/product-engineers/5-ways-to-improve-analytics-data) by PostHog. Also, note that it captures the first metrics after an initial delay rather than immediately on startup.

PR #4243, opened 2024-07-04
simplify virtual branches controller deps graph

krlvi:


crates/gitbutler-core/src/virtual_branches/controller.rs
```diff
@@ -500,7 +485,7 @@ impl Controller {
         let updated_project = self
             .projects
```
krlvi, line 486: This here is the only remaining dependency that virtual branches controller has towards projects controller. Removing it in a follow up pull request since that requires a change of how that update operates

Byron: That sounds like it's the last link preventing vbranches to move into their own crate and without needing access to projects at all. Neat!

PR #4246, opened 2024-07-04
remove with_verify_branch middleware closures

krlvi:


crates/gitbutler-core/src/virtual_branches/controller.rs
```diff
@@ -43,29 +44,33 @@ impl Controller {
...
-                project_repository.project().snapshot_commit_creation(
-                    snapshot_tree,
-                    result.as_ref().err(),
-                    message.to_owned(),
-                    None,
-                )
-            });
-            result
-        })
+        let project_repository = open_with_verify(project)?;
```
Byron, line 47: I see now - doing preparation is best done without a closure as it has to be at the beginning of a method anyway. This is definitely easier to look at than the closure version, and ~40 lines removed overall definitely prove this right as well.

PR #4247, opened 2024-07-05
Change ID Database

Caleb-T-Owens:


crates/gitbutler-project-store/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+fn sleeper(attempts: i32) -> bool {
+    println!("SQLITE_BUSY, retrying after 50ms (attempt {})", attempts);
+    std::thread::sleep(std::time::Duration::from_millis(50));
+    true
+}
+
+#[cfg(test)]
+mod tests {
```
Byron, line 79: Let's have integration tests, that is tests for public APIs go into `<crate>/tests/` similar to how it's done in `core`. It's actually a good check to assure public APIs are actually usable from other crates as well.

crates/gitbutler-project-store/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+use rusqlite::Connection;
+use std::path::Path;
+
+mod changes;
+mod migrations;
+
+const DATABASE_NAME: &str = "project.sqlite";
+
+/// ProjectStore provides a light wrapper around a sqlite database
+struct ProjectStore {
```
Byron, line 12: Since `Connection` is `!Sync`, i.e. `not Sync`, the whole `ProjectStore` will have to be behind a `Mutex` to compensate for that.

If `Clone` is implemented such that it connects to the database without rerunning initialization and migration then it concurrency will be possible nonetheless, and it can be as granular as `sqlite` can make it.

crates/gitbutler-project-store/src/migrations/migrator.rs
```diff
@@ -0,0 +1,204 @@
...
+/// This should only ever be changed with great caution
+const CREATE_MIGRATIONS_TABLE: &str = "
+CREATE TABLE IF NOT EXISTS migrations (
+    id BLOB PRIMARY KEY NOT NULL, -- A V4 UUID
+    name TEXT NOT NULL
+)
+";
+
+/// Migrator is used to perform migrations on a particular database.
+pub(crate) struct Migrator<'l> {
```
Byron, line 18: When there is no name used for the lifetime then it's traditionally `'a`, which is better for readability and sticking to conventions. Then I for one don't have to wonder if `l`  means `link`, but be upset because I will never know for sure 😁.

crates/gitbutler-project-store/src/migrations/migrator.rs
```diff
@@ -0,0 +1,204 @@
...
+    #[test]
+    /// Testing the `find_applied_migrations` function.
+    /// Ensuring that it creates the `migrations` table if it doesn't exist
+    fn find_applied_migrations_creates_migrations_table_if_not_exists() {
+        let mut connection = Connection::open_in_memory().unwrap();
+
+        let migrator = Migrator::new(&mut connection);
+
+        // Try to find unapplied migrations
+        migrator.find_applied_migrations().unwrap();
```
Byron, line 128: There should probably be an assertion here to assure they are empty.

crates/gitbutler-project-store/src/migrations/migrator.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    #[test]
+    /// Testing the `find_applied_migrations` function.
+    /// Assuming the `migrations` table exists (which we create by calling
+    /// `find_applied_migrations), we expect that any values inserted into
+    /// the table will get returned when we call it again.
+    fn find_applied_migrations_lists_all_entries_in_table() {
+        let mut connection = Connection::open_in_memory().unwrap();
+
+        // Call find_applied_migrations in order to create the migrations table
```
Byron, line 144: I'd move this comment into an assertion as last argument of `assert!` for example:

```Rust
assert!(migrator.find_applied_migrations().unwrap().is_empty(), "nothing happened yet, but creates table as side-effect");
```

crates/gitbutler-project-store/src/migrations/migrator.rs
```diff
@@ -0,0 +1,204 @@
...
+            assert_eq!(results.len(), 2);
+            assert!(results.contains(&"base".to_string()));
+            assert!(results.contains(&"other".to_string()));
+        }
+    }
+
+    #[test]
+    /// Testing the `migrate` function.
+    /// When given a list of migrations to run, it will perform the migrations in order
+    fn migrate_applies_migrations_in_order() {
```
Byron, line 177: I recommend to not test `find_applied_migrations()` as it needs to know a lot of details. 
Instead, let `migrate()` return enough information to know which migrations were applied. That way there can be a 'journey' that does what happens typically, empty database, then the first migration, then the second, with idempotency checks sprinkled in.

Tests that are considered good because they don't panic are an indicator that some more useful information could be returned organically  to make what happens more observable.

crates/gitbutler-project-store/src/migrations/migrator.rs
```diff
@@ -0,0 +1,204 @@
...
+impl<'l> Migrator<'l> {
+    pub(crate) fn new(connection: &'l mut Connection) -> Migrator<'l> {
+        Migrator { connection }
+    }
+
+    /// Iterates over the list of provided migrations starting at index 0.
+    /// If the migration has already been run, it will be skipped.
+    /// Run migrations get recorded in the `migrations` table.
+    /// The `migrations` table will be created if it doesn't exist.
+    pub(crate) fn migrate(&mut self, migrations: Vec<Migration>) -> Result<()> {
```
Byron, line 31: `migrations` doesn't need to be consumed, and when I read it I wonder why that needs to be consumed. Better not 'lie' and do `&[Migration]` instead.

crates/gitbutler-project-store/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// repository
+    pub fn initialize(gitbutler_project_directory: &Path) -> Result<ProjectStore> {
+        let database_path = gitbutler_project_directory.join(DATABASE_NAME);
+        let database_path = database_path.to_str().ok_or(anyhow!(
+            "Failed to get database {}",
+            gitbutler_project_directory.display()
+        ))?;
+
+        let connection = Connection::open(database_path)?;
+
```
Byron, line 33: This whitespace puzzles me, even though I know that I had a phase where I did that too.

Byron (commented):
It's great to see `sqlite` happening :)!

I left a few comments which I hope are helpful.

mtsgrd:
@Caleb-T-Owens do you still intend to work on this pr?

Caleb-T-Owens:
@mtsgrd At some point I think it would be good to introduce a SQL database. When I started we had a product-related reason for doing it (we had data we wanted to associate to commits/changes) but we ended up representing the data differently, so its no longer a priority.

PR #4255, opened 2024-07-07
better logs

Byron:
This PR improves on application logging in various ways, see tasks.

### Tasks

* [x] make sure we don't keep all the logs (`max_log_files(14)`) seems ineffective
    - The problem is that this is  only activated on roll-over, not on initialization or other events we actually have
* [x] make log filename simpler and purge all non-conforming logs

### Notes for the Reviewer

* I performed manual testing on each commit, which worked well as there were many stray log files in my case. For safety, I compressed the `com.gitbutler.app.dev` folder to be able to restore and re-test if necessary, but didn't end up using it.
* Logs are now named `GitButler.<date>.log` consistently.
* This needs a documentation change which needs to be timed with a release, or be written so it's flexible enough.
* The docs-update [is available here](https://github.com/gitbutlerapp/docs/pull/19), and ideally it's synced with the release of this PR in stable. But even if it was released before people will probably still find their logs without problems.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -137,9 +137,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle-wincon"
-version = "3.0.3"
+version = "3.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


krlvi (approved):
👍

PR #4262, opened 2024-07-07
Move branch state to a separate crate

krlvi:
being separate from the `gitbutler-branch` create, this allows us to have other things separated out e.g. oplog in a separate create


crates/gitbutler-core/src/virtual_branches/state.rs
```diff
@@ -29,6 +29,16 @@ pub struct VirtualBranchesHandle {
     file_path: PathBuf,
 }
 
+pub trait VirtualBranchesAccess {
+    fn virtual_branches(&self) -> VirtualBranchesHandle;
+}
+
+impl VirtualBranchesAccess for Project {
```
krlvi, line 35: Im creating a trait here so that `Project` doesn't depend on it (deleted the method here https://github.com/gitbutlerapp/gitbutler/pull/4262/commits/71809fc882ea35f307cfa7c19e58275a18e19342#diff-5b60cde080fe61b162ffd4754d11be74279e9fcfb2a5189f4eb89257a7d631b1L121 )

@Byron do you know if is a common naming convention for traits like this? E.g. similar to the naming convention that is with `...Ext` for extension traits?

Byron: Thanks for reeling me in. To me this qualifies as extension trait as it's sole purpose is to make creating virtalbranches more native-looking on the `Project` type. Thus, `VirtualBranchesExt` would be a more conventional name.

PR #4263, opened 2024-07-07
refactor project Repository

krlvi:
The `project_repository::Repository` type was (still is) a bit of a monstrosity. 
It was acting as:
- a "request context" - a struct that keeps the "Project" and an opened "git2::Repository"
- a collection of functionality over the Repository (eg. committing/pushing etc)

In this PR the goal is to:
- [x] Split up the two usages - the state & the functionality using a separate trait
- [x] Reduce the places that depend on the functionality trait (especially in `gitbutler-core`
- [x] Reduce the size of the interface when possible

In a separate PR this will be further separated


crates/gitbutler-core/src/project_repository/repository.rs
```diff
@@ -15,14 +15,15 @@ use crate::{
...
-pub struct Repository {
+use super::Config;
+
+pub struct ProjectRepo {
     git_repository: git2::Repository,
     project: projects::Project,
 }
 
-impl Repository {
+impl ProjectRepo {
```
Caleb-T-Owens, line 26: sitory

PR #4265, opened 2024-07-07
move repository code to own create

krlvi:
The RepositoryExt is still missing from this refactor but some other things need to be untangled first


crates/gitbutler-repo/src/repository.rs
```diff
@@ -0,0 +1,487 @@
...
+    askpass,
+    error::Code,
+    git::{self, CommitHeadersV2},
+    projects::AuthKey,
+    ssh,
+    virtual_branches::{Branch, BranchId},
+};
+
+use gitbutler_core::project_repository::{Config, ProjectRepo};
+pub trait RepoActions {
```
Caleb-T-Owens, line 16: sitory

PR #4312, opened 2024-07-09
no-proxy

Byron:
The frontend should be better at this due to being a browser engine, which can naturally
handle slow network links and firewalls.

Caching should also be better and more consistent as cache-related headers are respected.

This should speed up display times of new branches significantly, ideally without blocking
the UI at all.

### Tasks

* [x] check performance, see if it's actually better

### Observations

* **without proxy-cache**
    - The frontend downloads everything from gravatar directly.
    - I have seen pictures loaded from disk in the same session, but not between sessions
    - The UI looks more sluggish when loading images with a non-local URL

https://github.com/gitbutlerapp/gitbutler/assets/63622/4390ff34-4df3-4817-b0e0-a40915eeb896

* **with proxy-cache**
    - There is a delay while waiting for the proxied images to be downloaded, but it's not noticable given the overall loading times here.
    - images are displayed more snappily for some reason

https://github.com/gitbutlerapp/gitbutler/assets/63622/3efc692e-2b74-49b1-9a92-3899340b770c

#### Shortcomings

* I didn't try the `--release` build - maybe it's generally faster? Even though it's the frontend and I wonder to what extend that would be optimized then.

### My Conclusions

The proxy-cache is there for a reason and I'd rather want to fix that than to remove it entirely, particularly because the image-cache of the frontend doesn't seem to persist.

My thinking is that the proxy-cache could be made as fast as a file-access to see if a file exists, while downloading everything in the background. Thus, the first time the proxy is requested, it will not be present and the frontend loads the image directly. The next time this happens, the local version of the image will be there to accelerate the frontend.

Doing that still feels kind-of hacky though, and maybe it would be better to give persistent caches to the frontend somehow. So on second thought, probably it's better to delegate improving this to the frontend, after merging this PR.


krlvi:
cc @ndom91 - i guess there are things we can do in UI-land to make this good

krlvi (approved):

Byron:
> Do you have any other thoughts here?

Complications don't sound great, and I'd also be afraid of the multi-webview aspect of this and their subtle differences. Making images a little better also doesn't seem like a sufficient motivation to introduce such a 'beast'.

Right now, there is also no usecase for making GB a true local-first application which then would have to cache as much as possible I suppose. Somehow I think that ideally, the WebView can have the usual browser state as well so caches persist between sessions, which is what basically what one would do with service-workers, but for free ideally.

I would have hoped that CSS tricks are enough to reduce the pop-in effect of images being loaded though, which probably is enough here. After all, even if the network is unavailable, that shouldn't block the browser engine or the window, it's just stuff that fails silently eventually, and that's fine.

ndom91:
Yeah okay sounds like we're on the same page in regard to serviceworker then - they seem like a bit overkill for us at the moment.

My idea then for a first pass at improving this:
- First of all, I'd like to move to using GitHub avatars first instead of gravatar. That's the only remote we support atm, and I'd bet they have many more avatars of folks available compared to gravatar where we end up loading the placeholder image for most people
- Then I can see if any CSS tricks or rel=preload-ing the GH user assets domain help improve the perceived performance here
- Finally, I'm sure both github and gravatar set a pretty high max-age cache header, but maybe there's something more we can do in regard to caching without operating our own proxy

Any other ideas? Otherwise, I'm not sure how worthwhile it is to invest much more time than that into this atm.

Byron:
This sounds like a practical approach 👍.

I don't know about the semantics of preloading, maybe it could reduce the popup effect if people scroll. Somehow I feel this could be reduced as well if there was already something, like a proxy image. Then the pop-in won't be as strong.

ndom91:
Yeah maybe a simple loading indicator like one of these

https://miro.medium.com/v2/resize:fit:1358/0*07eEx7E__7USH9Ib.gif

Or maybe a generic [blurhash](https://blurha.sh) is enough :+1:

Byron:
I am just a dev 😅, but @PavelLaptev probably has all the answers :).

PR #4315, opened 2024-07-10
Refactor GitHub integration to allow for alternatives

mtsgrd:
Refactoring the code to support other hosted git services, e.g. GitLab.

- [x] Move GitHub related code to `$lib/hostedServices`
- [x] Split GitHub services into multiple small services
- [x] Create interfaces and make injection generic
- [x] Update touched layout.svelte to runes mode (removing `$:`)
- [x] Add test for HosteGitServiceFactory
- [x] Add test for PrMonitor


mtsgrd:
@Caleb-T-Owens I'm going to add back retries to `createPr` since it can fail just after pushing a new branch, and then add tests for that. But at this point it'd be great to get some feedback, especially around the interfaces and names of files/folders.

Caleb-T-Owens:
I'll take a look through now @mtsgrd

app/src/lib/hostedServices/interface/hostedGitPrMonitor.ts
```diff
@@ -0,0 +1,17 @@
...
+	pr: Readable<DetailedPullRequest | undefined>;
+	loading?: Readable<boolean>;
+	lastFetch?: Readable<Date | undefined>;
+	merge(method: MergeMethod): Promise<void>;
+	refresh(): Promise<void>;
+}
+
+export const [getHostedGitPrMonitorStore, createHostedGitPrMonitorStore] = buildContextStore<
+	HostedGitPrMonitor | undefined
+>('prMonitor');
```
Caleb-T-Owens: @mtsgrd why does this need to be a store and not just a regular context?

app/src/lib/hostedServices/github/githubPrMonitor.ts
```diff
@@ -0,0 +1,73 @@
...
+				await sleep(1000);
+			}
+		}
+		if (!pr) throw 'Failed to fetch pull request details';
+		return pr;
+	}
+
+	async merge(method: MergeMethod) {
+		await this.prService.merge(method, this.prNumber);
+	}
```
Caleb-T-Owens: Why does the monitoring class have merging as part of its responsibilities?

app/src/lib/hostedServices/github/octokit.ts
```diff
@@ -0,0 +1,9 @@
+import { Octokit } from '@octokit/rest';
+
+export function octokitFromAccessToken(accessToken: string | undefined) {
+	return new Octokit({
+		auth: accessToken,
+		userAgent: 'GitButler Client',
+		baseUrl: 'https://api.github.com'
+	});
+}
```
Caleb-T-Owens: When do we want to pass undefined as the accesss token?

app/src/lib/hostedServices/github/githubPrMonitor.ts
mtsgrd: I did it because it has the PR number handy, but now that you point it out, something that "monitors" something shouldn't mutate the thing it's monitoring.

app/src/routes/+layout.svelte
```diff
@@ -57,11 +60,26 @@
...
 	setContext(ProjectMetrics, data.projectMetrics);
 
+	const user = data.userService.user;
+	const accessToken = $derived($user?.github_access_token);
+	const octokit = $derived(octokitFromAccessToken(accessToken));
+
+	// This store is literally only used once, on GitHub oauth, to set the
+	// gh username on the user object. Furthermore, it isn't used anywhere.
+	// TODO: Remove the gh username completely?
+	const githubUserService = $derived(octokit ? new GitHubUserService(octokit) : undefined);
```
Caleb-T-Owens: Because we're always creating a octokit even when the accessToken is undefined, this will always create teh GitHubUserService

app/src/lib/hostedServices/interface/hostedGitPrMonitor.ts
mtsgrd: Because `HostedGitPrMonitor` is an interface and our current `setContext` works only with classes.

app/src/lib/hostedServices/github/octokit.ts
mtsgrd: Never!

app/src/lib/hostedServices/github/githubService.ts
```diff
@@ -0,0 +1,35 @@
...
+		return new GitHubListingService(this.projectMetrics, this.octokit, this.repo);
+	}
+
+	prService(baseBranch: string, upstreamName: string): HostedGitPrService {
+		return new GitHubPrService(this.octokit, this.repo, baseBranch, upstreamName);
+	}
+
+	checksMonitor(sourceBranch: string): HostedGitChecksMonitor {
+		return new GitHubChecksMonitor(this.octokit, this.repo, sourceBranch);
+	}
```
Caleb-T-Owens: I don't think this makes sense

We have the DefaultHostedGitServiceFactory, which will create a specific GitHubService, which is actually a factory for these smaller services.

I don't really see the benefit of the intermediary step. The DefaultHostedGitServiceFactory would be better of being a factory for the listService, prService, and checksMonitor.

app/src/lib/hostedServices/hostedGitServiceFactory.ts
```diff
@@ -0,0 +1,27 @@
...
+		switch (repo.provider) {
+			case 'github.com':
+				if (!this.octokit) throw new Error('Octokit not available');
+				return new GitHubService(new ProjectMetrics(), this.octokit, repo);
+		}
+	}
+}
+
+export const [getHostedGitServiceFactoryStore, createHostedGitServiceFactoryStore] =
+	buildContextStore<HostedGitServiceFactory | undefined>('HostedGitServiceFactory');
```
Caleb-T-Owens: It looks like we don't make use of this and just depend on the DefaultHostedGitServiceFactory directly in the layout file.

app/src/lib/hostedServices/interface/hostedGitPrMonitor.ts
Caleb-T-Owens: So, it looks like we're using derived to create the instances, so a store makes sense; but if we weren't or could avoid that, it would be nice to have a buildContext which didn't make use of a store

app/src/lib/hostedServices/github/githubService.ts
mtsgrd: As discussed offline, we'll keep this for now as the purpose is not having to provide `RepoInfo` and `Octokit` when constructing the individual services.

app/src/lib/hostedServices/hostedGitServiceFactory.ts
mtsgrd: My thinking was that if we write a test for e.g. `BranchLane`, then we might not want to use that implementation since it would mean we can't replace any of the services with mocks.

Caleb-T-Owens: @mtsgrd but won't we have the `GithubService` in the store which is an intermediary factory?

app/src/routes/[projectId]/pull/[number]/+page.svelte
```diff
@@ -6,24 +6,20 @@
...
 	import { page } from '$app/stores';
 
-	const githubService = getContext(GitHubService);
-
-	$: pr$ = githubService.prs$?.pipe(
-		map((prs) => prs.find((b) => b.number.toString() === $page.params.number))
-	);
+	const githubService = getHostedGitListingServiceStore();
+	$: prs = $githubService?.prs;
+	$: pr = $prs?.find((b) => b.number.toString() === $page.params.number);
```
Comment (ndom91) [inline app/src/routes/[projectId]/pull/[number]/+page.svelte]: Was there a particular reason you didnt migrate these?

Comment (Caleb-T-Owens) [reply app/src/routes/[projectId]/pull/[number]/+page.svelte]: I'm actually getting rid of that page in another PR so don't worry about it

app/src/lib/hostedServices/hostedGitServiceFactory.ts
mtsgrd: Right sorry, I mixed things up. Thanks, I'll drop these lines. 👍

app/src/routes/[projectId]/pull/[number]/+page.svelte
Comment (mtsgrd) [reply app/src/routes/[projectId]/pull/[number]/+page.svelte]: Just that it wasn't necessary given my change, I migrated the other ones because I wanted to use runes mode for new code. But, I might as well. :)

ndom91 (commented):
Yeah so in general this looks good to me, makes a lot of sense. 

Unrelated, but I see a few references to `posthog-js` directly, ideally I'd like to abstract the analytics / product metrics functions out, sort of like what we did here with git providers. To be able to swap others in/out more easily in the future. But thats definitely lower priority..

mtsgrd:
>Unrelated, but I see a few references to posthog-js directly, ideally I'd like to abstract the analytics / product metrics functions out, sort of like what we did here with git providers.

Definitely.

app/src/lib/hostedServices/github/githubPrService.test.ts
```diff
@@ -0,0 +1,28 @@
...
+				name: 'test-repo',
+				owner: 'test-owner'
+			},
+			'base-branch',
+			'upstream-branch'
+		);
+		const pr = await service.get(123);
+		expect(pr?.title).equal(title);
+	});
+});
```
mtsgrd: @Caleb-T-Owens @ndom91 just fyi. I will still add a more meaty test for the pr monitor before merging  anything.

PR #4316, opened 2024-07-10
multi-window safety

### LINKED ISSUE #3861: Multiple GB instances fail to see each others changes; race against each other
I think this behaviour is best described with a video:

https://github.com/gitbutlerapp/gitbutler/assets/63622/73c67659-df0e-4c8e-814d-5557f5897648

The core of the problem *seems* to be that not all updates are picked up despite file-changed events coming in. Since information is stored in individual files, each app instances ends up seeing their own version of vbranches. This will cause races where one application wins, and the other looses vbranch changes.

It's notable that changing the project in one instance cause the other to see the vbranch changes.

Byron:
Currently, mutliple instance of GitButler will interfere (and race) with each other when opened on the same project.
This PR makes this safe by preventing this by means of filesystem locking, using locks that can't go stale.

This mechanism prevents multiple instaces (processes) or windows (within one process) to observe the same repository,
making it exclusive to the respective 'owning' window.

This PR also implements early support for multiple windows for good measure.

### Tasks

* [x] encourage more 'mordern' `tauri` commands by example
* [x] Lock per project, which follows the active project
    - there now is an issue with newly added projects being 'always on', so somehow changing back to an older project doesn't setup watchers. Ignoring, but setting up multi-window support right away.
* [x] multi-window support
    - this should force some cleanup in how events are fired, needs to touch the UI

### Tasks: practical multi-window support

Generally
 - keep state per Window to handle watchers and don't rely on a single-window to exit
 - `listProjects` provides information about the window that has a project open (could be used to focus it)

* [x] per-window state, respond to open/close
* [x] open new project window with `alt+click` on existing project
* [x] fix strange close-behaviour/make it possible to close these
* [x] make sure multiple windows actually clash if displaying the same project
* [x] UI displays which projects are currently opened, no matter which window

https://github.com/gitbutlerapp/gitbutler/assets/63622/2cc81f6c-f216-47ce-b9a3-4e58cac35415

### Out of Scope

* Differentiate between 'observing' (read-only) Windows which cannot change things.
    - Doing this would require the Window to watch all relevant state and be able to update itself automatically.
      Right now, file-watching is only implemented for particular cases.

### Notes for the Reviewer

* Sometimes, when a window is closed, the application half-crashes - all Windows close, without coming back. I'd think that this is some `tauri` issue as safe Rust simply doesn't do that.
* I validated that multi-window project handling works as well as multi-instance handling, i.e. they won't allow to open a project in more than one window at a time no matter where the window is coming from. However, that test was only on MacOS, but should act the same also on Linux and Windows.
* The UI part can certainly use some tuning to have better UX around discovering how to open new windows, and probably the entire flow around opening windows. What's here is just a bare minimum, even though I personally like the UX of it, coincidentally, as decent UX wasn't the goal.
* I have implemented the 'is open' information for the project picker in the backend, even though in theory it could be done in the frontend if it would be able to list all windows. But I couldn't figure out how, without relying on lower-level `tauri://` messages that the implementation seems to be using under the hood. What's there is neat too though, and feels quite alright.

#### Empty `ProjectSwitcher`
When opening a new Window and it fails as the same project is already opened elsewhere, then the project list will be empty. This seems to be due to it being called twice, the first time it's empty which is shown, the second time it gets projects, but that's ineffective. It's notable that the first time when it's deemed empty, there is no call to `list_projects`.

Screenshot 2024-07-11 at 10 20 07

Note that this bug now is much harder to run into, and really needs multiple instance of the application to be used as a single application will make it hard to open a project that is already opened (in the same instance).

### Known Shortcomings

* Each Window of the UI acts like it's the main window, which is notable when the updater popup shows in each of them. 
* Even though the main `ProjectSelector` (in Navigation) merges the state of other Windows to avoid trying to open those that it knows are opened, the `ProjectSwitcher` doesn't currently do that.
* The `ProjectSelector` should update its list each time it is opened, otherwise it's still quite easy to run into "the Project is locked" problem as the listing of projects isn't uptodate in terms of which one is opened.

### DIFF
--- a/app/src/routes/[projectId]/+layout.ts
+++ b/app/src/routes/[projectId]/+layout.ts
@@ -35,7 +35,7 @@ export async function load({ params, parent }) {
 	let project: Project | undefined = undefined;
 	try {
 		project = await projectService.getProject(projectId);
-		await invoke('set_project_active', { id: projectId }).then((_r) => {});
+		await invoke('set_project_active', { id: projectId });
 	} catch (err: any) {
 		throw error(400, {
 			message: err.message
--- a/app/src/routes/[projectId]/+layout.ts
+++ b/app/src/routes/[projectId]/+layout.ts
@@ -30,10 +30,12 @@ export async function load({ params, parent }) {
 	// Getting the project should be one of few, if not the only await expression in
 	// this function. It delays drawing the page, but currently the benefit from having this
 	// synchronously available are much greater than the cost.
+	// However, what's awaited here is required for proper error handling,
+	// and by now this is fast enough to not be an impediment.
 	let project: Project | undefined = undefined;
 	try {
 		project = await projectService.getProject(projectId);
-		invoke('set_project_active', { id: projectId }).then((_r) => {});
+		await invoke('set_project_active', { id: projectId }).then((_r) => {});
 	} catch (err: any) {
 		throw error(400, {
 			message: err.message
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2175,7 +2175,6 @@ name = "gitbutler-project"
 version = "0.0.0"
 dependencies = [
  "anyhow",
- "async-trait",
  "git2",
  "gitbutler-error",
  "gitbutler-id",
@@ -2292,7 +2291,6 @@ name = "gitbutler-tauri"
 version = "0.0.0"
 dependencies = [
  "anyhow",
- "async-trait",
  "backtrace",
  "console-subscriber",
  "dirs 5.0.1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1609,6 +1609,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "fslock"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "04412b8935272e3a9bae6f48c7bfff74c2911f60525404edfdd28e49884c3bfb"
+dependencies = [
+ "libc",
+ "winapi",
+]
+
+[[package]]
 name = "funty"
 version = "2.0.0"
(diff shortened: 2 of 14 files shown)


app/src/routes/[projectId]/+layout.ts
```diff
@@ -30,10 +30,12 @@ export async function load({ params, parent }) {
 	// Getting the project should be one of few, if not the only await expression in
 	// this function. It delays drawing the page, but currently the benefit from having this
 	// synchronously available are much greater than the cost.
+	// However, what's awaited here is required for proper error handling,
+	// and by now this is fast enough to not be an impediment.
 	let project: Project | undefined = undefined;
 	try {
 		project = await projectService.getProject(projectId);
-		invoke('set_project_active', { id: projectId }).then((_r) => {});
+		await invoke('set_project_active', { id: projectId }).then((_r) => {});
```
Comment (mtsgrd) [inline app/src/routes/[projectId]/+layout.ts]: You can drop the `.then()` here.

mtsgrd (commented):
Front-end parts look good to me. 👍

krlvi (approved):
looks good from my perspective too. lets merge it and iterate where needed

PavelLaptev:
@krlvi @Byron to me, this UX doesn't feel right, because I didn't see this pattern before.
In my view, the `Project selector` should go away then. Instead, we should show the project name somewhere.

image

Add a `new project` we can move to the top bar menu

image

@krlvi wdyt?

krlvi:
Let's touch up the UI in a follow up pr

Byron:
Thanks @PavelLaptev, I am glad this will be much better :). Right now it's implemented so that it can be tested, and in a way that probably nobody will run into even if this version was released. There also are many shortcomings, still, that ideally also get addressed with the next iteration.

It's an obscure feature for now, with the backend now being able to support multiple windows.

PR #4319, opened 2024-07-10
remove virtual branch controller

krlvi:
there is no more state in the virtual branches controller. Renaming the type to _Actions to signify that this is a collection of functions. 

@Byron I am thinking there is probably a nicer pattern to express this... so this is just a starting point


Byron:
Functions are the most flexible :)! And this PR is a nice incremental step as it literally leaves no state in the struct, and it would be an interesting way to signal which functions are used by other functions. Maybe good to have as intermediate state indeed.

I'd find them most native when located either directly in `gitbutler_virtual` or in a module to further categorize them when approriate, like `gitbutler_virtual::foo::<function>`.

PR #4326, opened 2024-07-10
Split branch manager into seperate traits

Caleb-T-Owens:
@Byron I'd be keen on getting your thoughts on the branch manager, and this sort of structuring with multiple smaller traits

### DIFF
--- a/crates/gitbutler-branch-actions/src/branch_manager/branch_removal.rs
+++ b/crates/gitbutler-branch-actions/src/branch_manager/branch_removal.rs
@@ -19,7 +19,9 @@ use gitbutler_repo::{RepoActions, RepositoryExt};
 use super::BranchManager;
 
 pub trait BranchRemoval {
+    /// Perminently deletes a virtual branch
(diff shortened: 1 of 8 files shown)


crates/gitbutler-branch-actions/src/branch_manager/branch_removal.rs
```diff
@@ -0,0 +1,248 @@
...
+    ) -> Result<git2::Oid>;
+    fn build_real_branch(
+        &self,
+        vbranch: &mut branch::Branch,
+        name_conflict_resolution: NameConflitResolution,
+    ) -> Result<git2::Branch<'_>>;
+}
+
+impl PrivateBranchRemoval for BranchManager<'_> {
+    fn build_real_branch(
```
krlvi, line 149: it's used only in one place above. consider making this a simple function. or feel free to try this as is 🤷

Byron:
If there aren't multiple implementations or if it's not an `Ext`ension trait, it most certainly shouldn't be a trait. Instead, I'd use modules and multiple `impl` blocks. Maybe I am missing something though, I only glanced at it.

PR #4328, opened 2024-07-10
Use impls rather than private traits

Caleb-T-Owens:


Byron:
Oh, I am always a little behind it seems :D.

PR #4335, opened 2024-07-11
fix: commit avatar loading improvement

ndom91:
## ☕️ Reasoning

- Commit Avatars would jump in "out of nowhere" after the initial commit card had already been rendered, as soon as the gravatar image was loaded from the network. With long branches / lists of commits this didn't look great

## 🧢 Changes

- Added an initial placeholder [blurhash](https://evanw.github.io/thumbhash/) consisting of a blurred version of one of our accent-colored images from the marketing site, which would then be replaced with the actual avatar as soon as its loaded. 
	- Ideally you'd calculate a blurhash placeholder beforehand for each image, but I dont think thats necessary here. As long as it blends in with the general theme, these images are so small and dynamic that that would be asking a lot for very little return imo
- I also refactored the `gravatarUrl` to be a `string`, not a `URL`. We never took advantage of it being an instance of URL (like pulling out just the pathname, host, etc. etc.). Our use-case was simply to consume the URL as an `img.src` eventually. So just keeping it a string simplifies it a bit.

### DIFF
--- a/packages/ui/src/lib/CommitLines/AvatarImage.svelte
+++ b/packages/ui/src/lib/CommitLines/AvatarImage.svelte
@@ -1,0 +1,60 @@
+<script lang="ts">
+	import { tooltip } from '$lib/utils/tooltip';
+
+	interface Props {
+		srcUrl: string;
+		tooltipText: string;
+		altText: string;
+	}
+
+	const { srcUrl, tooltipText, altText, ...rest }: Props = $props();
+
+	let isLoaded = $state(false);
+</script>
+
+<div class="image-wrapper">
+	<img
+		class="avatar"
+		alt={altText}
+		src={srcUrl}
+		onload={() => (isLoaded = true)}
+		width="100"
+		height="100"
+		class:hidden={!isLoaded}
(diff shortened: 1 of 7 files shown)


packages/ui/src/lib/CommitLines/AvatarImage.svelte
```diff
@@ -0,0 +1,59 @@
...
+		class="avatar"
+		alt={altText}
+		src={srcUrl}
+		onload={() => (isLoaded = true)}
+		width="100"
+		height="100"
+		class:hidden={!isLoaded}
+		use:tooltip={tooltipText}
+		{...rest}
+	/>
```
Caleb-T-Owens: should we have `loading="lazy"` to this?

ndom91: Yeah we can definitely try it, this seems like an appropriate place to lazy load the avatar images.

I think there's a bigger issue though when loading large lists of commits, it seems like the entire commit card is a bit janky to load when being rendered for the first time. Not just this small image part. I've been doing some more digging, but haven't found anything concrete yet

PR #4338, opened 2024-07-11
Fix double remote fetch 

mtsgrd:
- all remotes already fetched in existing iter/map


mtsgrd:
Thanks @Byron for fixing this up. 🙏

PR #4342, opened 2024-07-12
Apply from local branches

Caleb-T-Owens:


crates/gitbutler-command-context/src/repository.rs
```diff
@@ -70,4 +71,51 @@ impl ProjectRepository {
...
+                .ok_or(anyhow::anyhow!("Branch name was not utf-8"))
+                .map(String::from)?;
+
+            let remotes = repo
+                .remotes()
+                .context("Failed to get remotes")?
+                .iter()
+                .flatten()
+                .map(String::from)
+                .sorted_by_key(|reference_name| -(reference_name.len() as i32))
```
Byron: Shouldn't that be `remote_name`?

crates/gitbutler-command-context/src/repository.rs
```diff
@@ -70,4 +71,51 @@ impl ProjectRepository {
...
+                .collect::<Vec<String>>();
```
Byron: This collect isn't necessary - after sorting the search is sequential.

crates/gitbutler-command-context/src/repository.rs
```diff
@@ -70,4 +71,51 @@ impl ProjectRepository {
...
+            let longest_remote = remotes
+                .into_iter()
+                .find(|reference_name| shorthand_name.starts_with(reference_name))
+                .ok_or(anyhow::anyhow!(
+                    "Failed to find remote branch's cooresponding remote"
+                ))?;
+
+            let shorthand_name = shorthand_name
+                .chars()
+                .skip(longest_remote.len() + 1)
```
Byron: Maybe a nit, but I think it's more explicit if the `/` would be mentioned one way or another. One way to do that would be to `shorthand_name.strip_prefix(remote_name)`, and then strip once more to get rid of the `/` character or fail if for whatever reason it's not there.

crates/gitbutler-command-context/src/repository.rs
```diff
@@ -70,4 +71,51 @@ impl ProjectRepository {
     pub fn repo(&self) -> &git2::Repository {
         &self.git_repository
     }
+
+    /// Fetches a branches name without the remote name attached
+    ///
+    /// refs/heads/my-branch -> my-branch
+    /// refs/remotes/origin/my-branch -> my-branch
```
Byron, line 78: Is there a reason to not just strip the first path component from a remote tracking shorthand? For example, `origin/feat/one` or `fork/my-branch` could be shortened that way, and as far as I know, `/` is not allowed in remote names.
If there is a good reason though, there should probably be a test for that.

crates/gitbutler-branch-actions/src/remote.rs
```diff
@@ -116,34 +122,37 @@ pub fn branch_to_remote_branch(branch: &git2::Branch) -> Result<Option<RemoteBra
...
-                    last_commit_author: commit
-                        .author()
-                        .name()
-                        .map(std::string::ToString::to_string),
-                })
-            })
-            .transpose(),
-        Err(_) => Ok(None),
-    }
+    let Ok(name) = name else { return Ok(None) };
```
Byron, line 123: This could be less repetitive/more concise.
```suggestion
    let (Ok(name), Some(sha)) = (name, branch.get().target()) else { return Ok(None) };
```

crates/gitbutler-command-context/src/repository.rs
```diff
@@ -70,4 +71,51 @@ impl ProjectRepository {
...
+    ///
+    /// An ideal implementation wouldn't require us to list all the references,
+    /// but there doesn't seem to be a libgit2 solution to this.
+    pub fn given_name_for_branch(&self, branch: &git2::Branch) -> Result<String> {
```
Byron, line 83: I don't know how this is actually used, but if it is to obtain the remote name of a local tracking branch, then this wouldn't *always* work as refspecs need to be considered.
While `git2` doesn't have methods for this ([these probably lack what you are looking for](https://docs.rs/git2/latest/git2/struct.Repository.html#method.branch_remote_name)), `gix` has more but also doesn't provide a mapping from tracking branch to the ref name on the remote.

I have added this feature [to the `gitoxide` tracking issue for GitButler](https://github.com/Byron/gitoxide/issues/1287).

Byron (commented):
I think I got quite stuck on the function that deals with local tracking branches, apologies for all the comments and ideas that sparked.

crates/gitbutler-command-context/src/repository.rs
Caleb-T-Owens: Yep indeed

crates/gitbutler-command-context/src/repository.rs
Caleb-T-Owens: Oooh! I didn't know about strip_prefix. That indeed does seem preferable

crates/gitbutler-command-context/src/repository.rs
Caleb-T-Owens: @Byron you can actually have forward slashes in remote names 
image

crates/gitbutler-command-context/src/repository.rs
Caleb-T-Owens: We want to group items in the sidebar by the branch name, so we need to collect up all the remote branches with that given name, prs for branches with the given name, and local references with the given name.

GitHub's api gives you the `sourceBranch` property which is what I want, local references, by default don't have the remote name attached, so its just the remote branches that need this processing


Screenshot 2024-07-12 at 14 30 31

crates/gitbutler-command-context/src/repository.rs
Caleb-T-Owens: our `struct Refname` parsing is actually incorrect as it makes that forward slash assumption

Byron: I see! Then I'd definitely at least mention this in the examples list so it's clear why it is what it is. Of course, a test would be even better, but it's a bit cumbersome to do right now.

Byron: And I just verified that `gitoxide` sees that similarly, as [the validation](https://github.com/Byron/gitoxide/blob/f8ce3d0721b6a53713a9392f2451874f520bc44c/gix/src/remote/name.rs#L20-L26) (as taken from Git) is quite 'interesting' :D - the name is valid if it can become a valid refspec of the 'normal' form.

crates/gitbutler-branch-actions/src/remote.rs
Caleb-T-Owens: I've seen that sort of pattern elsewhere and I'm not sure if I like it. To me it looks quite cluttered having it all on one line.

If it is the preferred way of doing it though I guess I should get used it it

crates/gitbutler-command-context/src/repository.rs
Caleb-T-Owens: I was mainly wanting to split the chain up, but I've actually changed to doing all the manipulation as just `&str`s instead which made it a bit simpler, and also gave a natural break at another point for life-time reasons.

crates/gitbutler-command-context/src/repository.rs
Caleb-T-Owens: You know, I spent about 15 minutes looking for a method to that effect when writing it the first time 😆

crates/gitbutler-command-context/src/repository.rs
Caleb-T-Owens: Git is full of all of these surprises.

For fun the other day ago, I was looking at parsing odb objects for fun. Blobs are just compressed strings, commits are just compressed strings, but trees? Hah! enjoy a bisque format you need to parse 😬 https://gist.github.com/Caleb-T-Owens/e8630f2b600e99acc3c3e317efb42a61.

I spent about 2 hours reverse engineering that format 😆.

Some day I'll try to understand packfiles...

crates/gitbutler-branch-actions/src/remote.rs
Byron: The line isn't formatted, so I think it will look better after formatting.
But even then, if you don't like it, that should be enough to keep things as is.

crates/gitbutler-command-context/src/repository.rs
Byron: Hmm, I didn't try it but I would certainly be surprised if ` .sorted_by_key` can't be the last thing in a chain. But if the compiler thinks so, I rest my case - personally I avoid using `itertools`, it hides things and `sorted_by_key()` isn't actually an iterator in the sense I think of them (it allocates).

crates/gitbutler-command-context/src/repository.rs
Byron: > Some day I'll try to understand packfiles...

They are a bit more complicated than trees especially if you plan to decode objects as well. And if reading the Git sources isn't cheating, that would definitely be my recommendation (but where is the fun in that ;)).

Caleb-T-Owens: I've found a well written post that goes over how to read packfiles in haskell so I'll probably make good use of that https://stefan.saasen.me/articles/git-clone-in-haskell-from-the-bottom-up/#pack-file-format

PR #4347, opened 2024-07-12
Change: preferred action for commit integration

PavelLaptev:
Currently, if there are no integrated remote commits, the preferred action is to `Force push`, which deletes remote commits. The preferred action in this case should be `Integrate upstream`.

image

### DIFF
--- a/app/src/lib/branch/BranchFooter.svelte
+++ b/app/src/lib/branch/BranchFooter.svelte
@@ -71,7 +71,7 @@
 				wide
 				projectId={project.id}
 				requiresForce={$branch.requiresForce}
-				integrate={hasRemoteCommits}
+				canIntegrate={hasRemoteCommits}
 				{isLoading}
 				trigger={async (action) => {
 					isLoading = true;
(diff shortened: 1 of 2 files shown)


mtsgrd:
Sorry for the delay, I'm trying to figure what the appropriate event is for setting that preferred action. I don't think it's reactively to `requiresForce`, but more something like when it transitions from not having remote commits to having them. Also, this setting seems to be set at a project level, it feels like it would make sense to use branchId in the key instead?

PavelLaptev:
@mtsgrd Thank you for helping me. Can I merge it, or do you want to change anything else?

mtsgrd:
Sorry for the delay, let me change one more thing. I didn't want it setting
based on the parameter alone, but rather when it changes value from false
to true.

mtsgrd:
Hey so we talked about this offline and consensus was that we want to hold off on this change. Just as you are likely to want to integrate upstream changes, it feels like it might be even more common that you have upstream changes because of undoing a commit.

Let's close this for now and solve it differently in the near future?

PR #4366, opened 2024-07-14
Handle unhandled promise rejections

mtsgrd:
- users need this feedback in case something goes wrong

### DIFF
--- a/app/src/hooks.client.ts
+++ b/app/src/hooks.client.ts
@@ -1,30 +1,44 @@
+import { showError } from '$lib/notifications/toasts';
 import { captureException } from '@sentry/sveltekit';
 import { error as logErrorToFile } from 'tauri-plugin-log-api';
 import type { HandleClientError } from '@sveltejs/kit';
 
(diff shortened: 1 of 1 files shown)


ndom91:
Maybe I'm missing something, but it should still be shown to the user before. 

The error would be returned here from `handleError` handle and from there available in the `$page.error` store in the `layout.svelte`/`page.svelte` file and shown to the user via `showError` there

mtsgrd:
Yeah, promises are special, SvelteKit doesn't intercept them. :v:

Caleb-T-Owens:
Promises are equivalent to a lazily evaluated Result<> type.

An async function that throws an exception, IE:
```js
async function foo() {
  if (false) {
    return "sadf"
  } else {
    throw new Error("asdf")
  }
}
```

is more or less the same as saying:

```js
function foo() {
  return new Promise((resolve, reject) => {
    if (false) {
      return resolve("sadf")
    } else {
      return reject(new Error("asdf"))
    }
  }
}
```

PR #4375, opened 2024-07-14
refactor: Branch name normilize

### LINKED ISSUE #4187: The problem of setting the remote branch name to Chinese
### Version

0.5.561

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

"Set remote branch name" is set to empty in Chinese

### How to reproduce

_No response_

### Expected behavior

_No response_

### Relevant log output

_No response_

PavelLaptev:
## ☕️ Reasoning

Currently, GitButler removes all special characters from branch names, although Git and GitHub support them. GitHub has some [recommendations](https://docs.github.com/en/get-started/using-git/dealing-with-special-characters-in-branch-and-tag-names), including trimming the string and removing characters like `|\+^~<>*` from the beginning and the end of the string. This makes sense, so I have implemented the same logic here.

Here is the test name:

image

## 🧢 Changes

Updated two functions `normalizeBranchName` and `normalize_branch_name` to handle branch name normalization.

## 🎫 Affected issues

Fixes: https://github.com/gitbutlerapp/gitbutler/issues/4187


PavelLaptev:
@Caleb-T-Owens I removed the `normalizeBranchName` function from the frontend and instead used `normalize_branch_name` with the `invoke` method. I thought it would be better to use one normalization function instead of two (one in a TS file and another in an RS file).

PR #4376, opened 2024-07-14
Fix bug in GitHost factory

mtsgrd:
- you might have github enabled but not in a github repo


mtsgrd:
@Byron thanks for the bug report 🙏

PR #4384, opened 2024-07-15
Disable pr button when no gitHost available

mtsgrd:
- disabled button still prevents tooltip from displaying

### DIFF
--- a/app/src/lib/branch/BranchHeader.svelte
+++ b/app/src/lib/branch/BranchHeader.svelte
@@ -240,7 +240,8 @@
 						{#if !$pr}
 							<PullRequestButton
 								click={async ({ draft }) => await createPr({ draft })}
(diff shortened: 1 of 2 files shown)


mtsgrd:
@ndom91 @Caleb-T-Owens any thoughts on how we could work around tooltip not working for disabled input elements?

ndom91:
2 ideas:

1. Put the `onmouseenter`/`..leave` handlers on a wrapping `div` so we can still set `disabled` on the `button`.
2. A quick google seems to saw the pointer events versions do still work on disabled elements, so we could swap for `pointerenter`/`pointerleave`

mtsgrd:
I tried swapping mouseover for pointerover, but it made no difference. I'll land this pr now to disable the button, and we can follow-up on the tooltip later.

image

PR #4392, opened 2024-07-15
normilize-branch-name-update

PavelLaptev:
### Reasoning:
- Be able to handle other languages and special character exept `|\+^~<>*`
- use only one nomalize function instead of two

### Affected issue https://github.com/gitbutlerapp/gitbutler/issues/4187

image


PavelLaptev:
@Caleb-T-Owens, I checked if `git2::normalize_name` could be useful here, but it seems suitable only for specific cases, such as when you have levels like `refs/tags/`. Doesnt work in our case because e.g. if there are spaces in the string, the function will throw an error.

image

So I keep the `Regex` implementation.

PR #4401, opened 2024-07-16
Rename daft names

Caleb-T-Owens:
Hey @mtsgrd and @ndom91,

I decided that I ought to introduce the renames outside of my other branch, so I'm merging them here.

I thought I'd just thought I'd notify you guys so you're aware of the renames I've performed.

---

Both the RemoteBranch and RemoteCommit, could actually be local which is incredibly unhelpful


mtsgrd:
Thank you! Sanity at last.

PR #4426, opened 2024-07-17
Add better support for different forge web URLs, flatten BaseBranch f…

JMonsorno:
…rom repeat logic.

## ☕️ Reasoning
- Support for Azure DevOps branch and commit URLs
- Easier future support for more forge types without deep integrations

## 🧢 Changes

- Moved forge type from a collection of Boolean functions to their own type, new Azure DevOps, and a default Unknown.
- Added test cases for each forge type.
- Flattened `BaseBranch` logic to do potentially repetitive work once, most of which could be simplified further if moved to the rust side still.
- Moved some logic around to better support performing string actions once, namely `convertRemoteToWebUrl` and tests there of.

### DIFF
--- a/app/src/lib/baseBranch/baseBranch.ts
+++ b/app/src/lib/baseBranch/baseBranch.ts
@@ -1,6 +1,7 @@
-import { convertRemoteToWebUrl } from '$lib/utils/url';
-import { Commit } from '$lib/vbranches/types';
-import { Type } from 'class-transformer';
+import { ipv4Regex } from '$lib/utils/url';
+import { Commit, type ForgeType } from '$lib/vbranches/types';
+import { Expose, Transform, Type, type TransformFnParams } from 'class-transformer';
+import GitUrlParse from 'git-url-parse';
 
 export class NoDefaultTarget extends Error {}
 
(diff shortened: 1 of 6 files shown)


Caleb-T-Owens:
Hi! Thanks for contributing!!!! It means an awful lot to us. We'll take as look as soon as we can.

mtsgrd:
This looks really great, thank you! 🙌 

Haven't had a chance to view it locally yet, but I'm curious about the use of the term forge. We have been looking for a term to describe what github, gitlab etc. are, and currently use the term "git host" here: https://github.com/gitbutlerapp/gitbutler/blob/master/app/src/lib/gitHost/interface/gitHostService.ts#L6-L10

If we wanted to align the terms between this pr and that interface, which one would you recommend we keep?

JMonsorno:
@mtsgrd I have no personal stake in it and can adjust.

I was just looking for a proper term and didn't know one when I had the feature request in mind and came across "forge." I'll link the two Wiki articles in favor of forge:
https://en.wikipedia.org/wiki/Forge_(software)
https://en.wikipedia.org/wiki/Comparison_of_source-code-hosting_facilities

Since GitButler is exclusive to git and can be used on any git repository hosting rather than a proper forge it'd be reasonable to say-what-it-is and use "gitHostService" too.

Personally, I think the end goal of supporting complex features and deep integrations to each for items such as Pull Requests and links to Issues would lend more towards using Forge.

Caleb-T-Owens:
Having taken a look, forge does seem like a good term, I'm down for it

mtsgrd:
Oh nice, I did not know this term, but the wikipedia page also says "In 2023 the two most widely used collaborative development platforms are not advertised as being forges". Anecdotally I also asked some people and they weren't familiar with the term either.

If you don't mind, let's stick with "say-what-it-is" is terminology and use GitHost instead of GitForge? Also, just fyi, see #4440.

mtsgrd:
@Caleb-T-Owens didn't see you message before I responded. I don't have strong feelings here but I'm leaning towards not adopting that term, even @JMonsorno wasn't familiar with the term until he went looking for it. Wdyt?

Caleb-T-Owens:
If we were to go for forge, I'd want to call it "Forge" rather than
"GitForge".

Looks like GitLab just changed to jump on the AI band wagon...

Probably worth asking Scott's opinion

mtsgrd:
@JMonsorno I have a different approach to this in #4505. Would you mind having a look and let me know if something doesn't work for you?

mtsgrd:
Closing this since #4505 has been merged.

PR #4435, opened 2024-07-18
improved `conflict` file handling

Byron:
### Tasks

* [x] handle encoding properly (`as_encoded_bytes()`) requires `from_encoded_bytes()` for creation
* [x] avoid `String` as return value
* [x] avoid repeated writes and reads of conflict file in case there are multiple conflicts (might help with #3601)


Byron:
Merging this early as I think it's good and I really want to get the `Cargo.lock` update in, which I forgot when upgrading `gix` yesterday.

PR #4437, opened 2024-07-18
UX: Do not close the commit dialog after commit

PavelLaptev:
Fix for https://discord.com/channels/1060193121130000425/1206670506271707156/1263141470165274678


krlvi:
this is nice. I just got a thought that it would be neat to actually close it when there are no more uncommitted hunks @PavelLaptev

PavelLaptev:
@krlvi, if there are no uncommitted files or hunks, we show the empty state instead. Is that what you meant?

image

Issue #4423, opened 2024-07-17
Unintuitive UI when trying to amend a locked hunk past its dependency


niklaswimmer:
### Version

0.12.12

### Operating System

Mac OS X

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

I did some changes to a file and comitted those. I then did some additional changes and committed those again. I now have some changes I want to amend to the first commit I created. However, when I try to amend the first commit with the new changes via drag&drop, nothing happens. Amending the second commit works, but shows an error ("Target ownership not found").

The changes I made created a dependency of the file on the second commit, which I think is a key point here.

I recorded the actions I take in the video below.

https://github.com/user-attachments/assets/dd9753e5-4006-44b2-bc2f-ac54ca57db77

### How to reproduce

I was able to reproduce this exact behavior in a clean repository.

- Create a new Git repository and open it as GB project
- Add a file with one line of content, `asdf` in my case
- Commit
- Change `asdf` to `asqwerdf`
- Commit
- Add a new line of content, `qwe` in my case

The new change is locked and the UI shows that it is locked on the second commit.

- Try to amend the first commit you created, it does not work
- Try to amend the second commit you created, it does work, but shows an error ("Target ownership not found")

The log below is from this "clean" run.

### Expected behavior

As far as I know, GitButler currently has some limitations when it comes to hunk splitting, so I assume that amending the change to the first commit just does not work right now. The UI here is however quite confusing, because it indicates that amending my changes *should* work, and then nothing happens. I would have also not expected to get an error when amending my changes to the commit they depend on (especially because it ends up working just fine).

Another interesting thing I have noticed is that when I try to amend the changes to the first commit multiple times (say for example three times, because maybe I desperately hope that it will just work if I keep trying), the error that gets shown also pops up multiple times (three times in our example, it seems to be for every failed attempt, I then get an error).

Now that I have written all of this, I think this is kinda included in #3288, but I figured having a concrete example of how the current UI lacks would be helpful.

### Relevant log output

```shell
https://gist.github.com/niklaswimmer/71784b82048af86aed5b64fc85a2bb53
```

Byron:
Thanks a lot for reporting!
Indeed, one would expect the UI to either only show what it can do, or to not silently do nothing after a seemingly legal drop was performed.

mtsgrd:
I fixed this in #4438, thanks for reporting it. 💪

Byron:
That's great to hear!
[...]

PR #4438, opened 2024-07-18
Only amend specific commit if hunk locked to previous change

mtsgrd:
- the only commit that can be amended is the one the current change depends on


mtsgrd:
Fixes #4423

krlvi:
> the only commit that can be amended is the one the current change depends on

Hmm, how about subsequent commits? It would be a valid git operation, and it feels we should permit that. E.g. i can amend a change into commits that are after the one it depends on as well. We only need to prevent amending to commits that are earlier than the dependency
cc @mtsgrd

mtsgrd:
You're right, let's fix.

PR #4442, opened 2024-07-19
optimize `is_commit_integrated()`

Byron:
`is_commit_integrated()` might at least in parts responsible for the very bad performance when the upstream is ahead by many commits.

### Tasks

* [x] optimize

### Notes for the Reviewer

* I didn't actually run performance tests but refactored this solely based on the potential cost. The most expensive part is probably still finding the merge-base and performing the merge itself, which can't really be optimized with `git2`.

### DIFF
--- a/crates/gitbutler-branch-actions/src/virtual.rs
+++ b/crates/gitbutler-branch-actions/src/virtual.rs
@@ -459,6 +459,7 @@ pub fn list_virtual_branches(
 
         // find all commits on head that are not on target.sha
         let commits = ctx.log(branch.head, LogUntil::Commit(default_target.sha))?;
+        let check_commit = IsCommitIntegrated::new(ctx, &default_target)?;
         let vbranch_commits = commits
             .iter()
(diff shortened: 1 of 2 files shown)


krlvi:
This seems like a nice improvement, though it does not seems to affect the specific problematic case. I have a test with a really out of sync repo 
image
The time to load is the same with and without this change (1m 10s). 

I did some profiling on that, and seems like another function that seems quite hot is `target_to_base_branch`, and possibly there is quite a bit of IO:
image

Regardless, it seems like something that should be safe to merge, and perhaps it brings a speedup somewhere else

krlvi (approved):

Byron:
> Regardless, it seems like something that should be safe to merge, and perhaps it brings a speedup somewhere else

Thanks, even though I think this is a close call - I was making this change because it *felt* like a waste, without having a measurement to prove it, so this could be a premature optimization which does introduce some complexity.
I will be more careful about this in the future.

PR #4449, opened 2024-07-20
don't fail hard if there is no project metadata

### LINKED ISSUE #4371: Sending the `Project Data` portion of "Share Debug Info" fails
### Version

0.12.9 (20240710.081834) and n 0.5.604 (20240714.031300)

### Operating System

Mac OS X

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

Screenshot 2024-07-14 at 17 42 05

Hitting "Share With GitButler` with the settings above in the feedback dialog will cause an error.

### How to reproduce

See above

### Expected behavior

It succeeds.

### Relevant log output

```shell
The problem seems to be that the access to `<data-dir>/projects/<id>` fails as the directory doesn't exist (or doesn't always exist).
```

Byron:
This fixes #4371.

### Tasks

* [x] validate by hand that the respective issue is truly fixed

### DIFF
--- a/app/src/lib/components/ShareIssueModal.svelte
+++ b/app/src/lib/components/ShareIssueModal.svelte
@@ -82,6 +82,7 @@
 					? zip
 							.gitbutlerData({ projectId })
 							.then(async (path) => await readZipFile(path, 'data.zip'))
+							.catch(() => undefined)
 					: undefined,
 				sendProjectRepository
 					? zip


Byron:
I am merging this proactively as I *think* the change is minimal and idiomatic, but CC @ndom91 to be sure it truly is.

krlvi:
In a way this field is no longer relevant. About 4-5 months ago we had a secondary repository per project that was used as a "database" and this option was about sharing the application state from there. Now, all of the state is persisted in "virtual_branches.toml" so in a way, this is option is no longer necessary. If anything, we should probably include an option to share the toml file.
Though, this fix is very valid, because it was really unpleasant for the app to error out when this is selected

Byron:
Thanks for the clarification, I wasn't sure if this location is going to be populated with other data at some point, and thought better not to remove something potentially valuable.

> Now, all of the state is persisted in "virtual_branches.toml" so in a way, this is option is no longer necessary. If anything, we should probably include an option to share the toml file.

The `virtual_branches.toml` would be included in the `Share project repository` option, but indeed sharing only project-specific metadata might be more tenable for users. The question is if they would risk sharing any data at all if it's a private project, at least without previewing the data before sending it. An alternative question would be how useful these bare toml files would be for debugging.

PR #4452, opened 2024-07-20
fix 4389

### LINKED ISSUE #4389: git fetch fails silently unless an ssh master process is started from CLI
### Version

0.12.10

### Operating System

Mac OS X

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

Hello, I'm running git CLI over SSH using either 1password or https://github.com/FiloSottile/yubikey-agent

Because I was using GB before it supported the execution of the git executable I was still using the self-generated keys.

Since the drop of that feature (no longer self-generated keys from GB) I have very weird behaviour.

- If I try to refresh the project trunk by clicking the ↺ icon, nothing hapens, like if there are no changes to fetch
- If I try to push, I got an error
- If I try to test credentials I get the following error

image

However, it I open a shell and run `git fetch` I get the 1password prompt for my ssh keys and from that point on, until the ssh master process expires its connection, I have a working GB (push and pull work as expected).

### How to reproduce

_No response_

### Expected behavior

If fetching fails, the UI should tell us.

I also would like to be able have this working without the need of a shell to run the first fetch operation.

### Relevant log output

_No response_

Byron:
This PR fixes #4389 by causing loud errors in the backend if a remote failed to fetch, and selectively ignoring them when a background fetch is performed.
That way, the original behaviour is kept while showing errors on user action.

### Tasks

* [x] reproduce the issue
* [x] fix it
* [x] validate it's fixed and we see error messages only on explicit user action

### How to reproduce the original bug

* open a project that is cloned via SSH and setup a password protected SSH key
* trigger a fetch using the refresh button
* cancel the popup that asks for a password, possibly multiple times, until it gives up and errors

With this PR, the error is visible, without it is isn't.

### DIFF
--- a/app/src/lib/baseBranch/baseBranchService.ts
+++ b/app/src/lib/baseBranch/baseBranchService.ts
@@ -45,7 +45,7 @@ export class BaseBranchService {
 				return;
 			} else if (err.code === Code.ProjectsGitAuth) {
 				showError('Failed to authenticate', err);
-			} else {
+			} else if (action !== undefined) {
 				showError('Failed to fetch', err);
 			}
 			console.error(err);
--- a/crates/gitbutler-tauri/src/virtual_branches.rs
+++ b/crates/gitbutler-tauri/src/virtual_branches.rs
@@ -9,7 +9,7 @@ pub mod commands {
     use gitbutler_branch_actions::{RemoteBranch, RemoteBranchData};
     use gitbutler_error::error::Code;
     use gitbutler_project as projects;
-    use gitbutler_project::ProjectId;
+    use gitbutler_project::{FetchResult, ProjectId};
     use gitbutler_reference::normalize_branch_name as normalize_name;
     use gitbutler_reference::ReferenceName;
     use gitbutler_reference::{Refname, RemoteRefname};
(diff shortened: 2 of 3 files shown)


app/src/lib/baseBranch/baseBranchService.ts
```diff
@@ -45,7 +45,7 @@ export class BaseBranchService {
 				return;
 			} else if (err.code === Code.ProjectsGitAuth) {
 				showError('Failed to authenticate', err);
-			} else {
+			} else if (action !== undefined) {
```
Byron, line 48: This causes the frontend to not how errors when background fetches are performed.

crates/gitbutler-tauri/src/virtual_branches.rs
```diff
@@ -478,12 +478,16 @@ pub mod commands {
...
             .update(&projects::UpdateRequest {
                 id: project.id,
-                project_data_last_fetched: Some(project_data_last_fetched),
+                project_data_last_fetched: Some(project_data_last_fetched.clone()),
                 ..Default::default()
             })
             .await
             .context("failed to update project with last fetched timestamp")?;
 
+        if let FetchResult::Error { error, .. } = project_data_last_fetched {
```
Byron, line 487: This allows the frontend to respond to errors during fetching, which otherwise would have been swallowed here.

krlvi (approved):

krlvi:
nice, thanks for fixing this @Byron!

PR #4473, opened 2024-07-23
New branch listing API

krlvi:
We are updating the UI of the list of branches (in the sidebar). The existing API has been long overdue for replacement (especially with the fact that it was talking only about "remote branches" before which was incorrect)


Byron:
### Notes from our call (K + S)

#### Goal

Prepare data for the user interface, formatting and grouping of data, specifically for the UI. Group references and display them as items. Branch  + remote tracking + virtual branch are grouped.

Enum: Either virtual branch, or a git-branch, grouping by identity. UI should be able to clearly state where and what a branch is. Reason for `gix` - easy to get the remote name.

API and code is documented to say what it should do (even though right now it doesn't yet show the right thing).

Surprised: There was a panic, when having a symbolic ref… to be reproduced by K.

#### Tasks

* [ ] review
* [ ] a simple test maybe?
* [ ] fixup and improve

crates/gitbutler-branch-actions/src/branch.rs
```diff
@@ -0,0 +1,395 @@
...
+        })
+        .collect();
+
+    // The head commit for which we calculate statistics.
+    // If there is a virtual branch let's get it's head. Alternatively, pick the first local branch and use it's head.
+    // If there are no local branches, pick the first remote branch.
+    let head = if let Some(vbranch) = virtual_branch {
+        Some(vbranch.head)
+    } else if let Some(reference) = local_branches.first().cloned() {
+        Some(git2::Oid::from_bytes(reference.id().as_bytes())?) // We have filtered out symbolic references, which would otherwise panic
```
krlvi: @Byron this is the place where i had an issue, but now i am filtering on line 119 and 131

crates/gitbutler-branch-actions/src/branch.rs
```diff
@@ -0,0 +1,408 @@
...
+    });
+
+    let mut remotes: Vec<BString> = Vec::new();
+    for reference in remote_branches.iter() {
+        let short_name: String = reference.name().shorten().to_str_lossy().to_string();
+        let file_name: String = reference.name().file_name().to_str_lossy().to_string();
+        // remove file_name suffix from short_name and also remote the trailing '/' character
+        let remote_name = short_name
+            .trim_end_matches(&file_name)
+            .trim_end_matches('/');
```
Caleb-T-Owens: Does this account for remote names which contain forward slashes?

krlvi: @Caleb-T-Owens it turns out that "file_name" is also not what i thought it was... this will not work

Byron: Definitely not.

I don't think that there is anything built-in for the kind of name you are looking for, i.e. `foo/bar` in a ref names `refs/remotes/my/special/remote/foo/bar` where `my/special/remote` is a valid remote name.

apps/desktop/src/lib/branches/branchListing.ts
```diff
@@ -0,0 +1,82 @@
...
+			);
+			console.log(branches);
+			return branches;
+		} catch (err: any) {
+			console.error(err);
+		}
+	}
+}
+
+/// Represents a branch that exists for the repository
```
Byron, line 20: I love the documentation to communicate intent!

apps/desktop/src/lib/branches/branchListing.ts
```diff
@@ -0,0 +1,82 @@
...
+	/// followed by the local branch and then the remote branches (taking the max if there are multiple)
+	numberOfFiles!: number;
+	/// The number of commits associated with a branch
+	/// Since the virtual branch, local branch and the remote one can have different number of commits,
+	/// the value from the virtual branch (if present) takes the highest precedence,
+	/// followed by the local branch and then the remote branches (taking the max if there are multiple)
+	numberOfCommits!: number;
+	/// Timestamp in milliseconds since the branch was last updated.
+	/// This includes any commits, uncommited changes or even updates to the branch metadata (e.g. renaming).
+	updatedAt!: number;
```
Byron, line 57: Since the typing here doesn't indicate it's milliseconds, it's easy to get wrong. I think I have seen somewhere that `_<unit>` suffixes were used to make it painfully clear what is what.

apps/desktop/src/lib/branches/branchListing.ts
```diff
@@ -0,0 +1,82 @@
...
+	/// If this branch is a local branch, this list will be empty.
+	remotes!: string[];
+	/// The branch may or may not have a virtual branch associated with it
+	virtualBranch?: VirtualBranchReference | undefined;
+	/// The number of lines added within the branch
+	/// Since the virtual branch, local branch and the remote one can have different number of lines removed,
+	/// the value from the virtual branch (if present) takes the highest precedence,
+	/// followed by the local branch and then the remote branches (taking the max if there are multiple)
+	/// If this branch has a virutal branch, lines_added does NOT include the uncommitted lines.
+	linesAdded!: number;
```
Byron, line 38: This metadata is expensive. Since it's branch-related, I assume it's for all lines added over all commits in all the remotes as well, and it's all put into one number here.

Somehow I would prefer to keep this data associated more closely with the entity that it relates to, and then have to UI do the aggregation.

Byron (commented):
In my quick review, I only looked at the `BranchListing` in conjunction with the UI it's supposed to feed:

image

Generally, since the line counts are so expensive, I think they have to go async into the background after the first list of branches (paginated) was retrieved. This is probably what I'd do with `git2` for ease of use, while doing the branch listing and organization with `gix`.

My next step would be to refactor for this and hopefully bring back `gix` for some parts at least.

PR #4482, opened 2024-07-24
Convert Button component to svelte5

mtsgrd:
Tied into this commit I also refactor the `ContextMenu` use such that it doesn't need local variables. It's in the same PR because of the removal of the 2-way binding that let us acquire the `HTMLElement` of a `Button`.


mtsgrd:
As discussed offline, postponing this for a bit later.

PR #4491, opened 2024-07-25
Match pull requests on their given names rather than display names

Caleb-T-Owens:
While we would ideally like to have PRs match on owner+repo+branchname, using the given name is a step forwards as the display name doesn't always accuratly map to the branchname


Caleb-T-Owens:
@mtsgrd FYI

PR #4501, opened 2024-07-27
first trial of `git2-to-gix`

Byron:
When looking at `BranchListing` only, try to convert the usage of `git2` *related to branches* to using `gix` instead.

Do so without having while assuring that the output stays the same.

Follow-up of #4473.

### Tasks

* [ ] a test utility to support the creation of read-only fixtures with GitButler setup
* [ ] a test for the `BranchListing` function to see the status quo
* [ ] convert `git2` to `gix`.


Byron:
I put it back to `draft`, probably your UI was outdated so it allowed to merge anyway.

PR #4506, opened 2024-07-27
Added enum for context

Caleb-T-Owens:


crates/gitbutler-command-context/src/context.rs
```diff
@@ -0,0 +1,174 @@
...
+        Ok(open_workspace_context)
+    }
+}
+
+pub trait ContextProjectAccess {
+    fn project(&self) -> &Project;
+    fn set_project(&mut self, project: &Project);
+}
+
+pub trait ContextRepositoryAccess {
```
Byron, line 125: I think anything that works with this trait now could instead just work with a `&git2::Repository` - that's a great finding that would simplify the functions. Once they need more, that can be changed.

This also means the caller invokes the `.repo()` method to produce the required `git2::Repository`.

crates/gitbutler-command-context/src/context.rs
```diff
@@ -0,0 +1,174 @@
...
+        let RequestContext::OpenWorkspace(open_workspace_context) = RequestContext::open(project)?
+        else {
+            bail!("Open workspace required for this action");
+        };
+
+        Ok(open_workspace_context)
+    }
+}
+
+pub trait ContextProjectAccess {
```
Byron, line 120: I would think that all this works without any traits, as the concrete types will always be known and method calls only happen with a single implementation.

Byron (commented):
Thanks for putting this up, it helps visualising the impact this change would have. With some simplification I think this will work.

Something that's interesting is that the data both variants contain is the same. Maybe this will change, but if it doesn't it might not be the right abstraction or there might be no need for a top-level `enum` at all. Or maybe that `enum` is just useful internally to for the creation of the respective kind of context to help make clear in the typesystem what kind of worktree functions and methods are designed to work on.

Maybe from there, it's also possible to derive a 'CurrentBranchContext` that informs about what can and can't be done with it. Maybe there are things that `gitbutler/integration` is fine with, but user-controlled branches are not?

I am lacking an idea on the differences to be able to tell, but I hope sharing this helps to get a conversation going.

Caleb-T-Owens:
> Maybe from there, it's also possible to derive a 'CurrentBranchContextthat informs about what can and can't be done with it. Maybe there are things thatgitbutler/integration` is fine with, but user-controlled branches are not?

@Byron Do you mean something like this?

```rs
trait Mode {}

struct UnspecifiedWorkspaceMode;
impl Mode for UnspecifiedWorkspaceMode {}

struct OpenWorkspaceMode;
impl Mode for OpenWorkspaceMode {}

struct OutsideWorkspaceMode;
impl Mode for OutsideWorkspaceMode {}

struct NewNewContext<T: Mode> {
    mode: PhantomData<T>,
    git_repository: git2::Repository,
    project: Project,
}

// It would be sweet if we could just cast the PhantomData type so we don't need
// to re-create the git repository (as it can't be cloned)
impl TryFrom<&NewNewContext<UnspecifiedWorkspaceMode>> for NewNewContext<OpenWorkspaceMode> {
    type Error = anyhow::Error;

    fn try_from(
        value: &NewNewContext<UnspecifiedWorkspaceMode>,
    ) -> std::result::Result<Self, Self::Error> {
        // Some guard
        if true {
            bail!("Uh oh, you can't do this");
        }

        Ok(NewNewContext {
            mode: PhantomData,
            git_repository: git2::Repository::open(&value.project.path)?,
            project: value.project.clone(),
        })
    }
}

impl TryFrom<&NewNewContext<UnspecifiedWorkspaceMode>> for NewNewContext<OutsideWorkspaceMode> {
    type Error = anyhow::Error;

    fn try_from(
        value: &NewNewContext<UnspecifiedWorkspaceMode>,
    ) -> std::result::Result<Self, Self::Error> {
        // Some guard
        if true {
            bail!("Uh oh, you can't do this");
        }

        Ok(NewNewContext {
            mode: PhantomData,
            git_repository: git2::Repository::open(&value.project.path)?,
            project: value.project.clone(),
        })
    }
}

impl TryFrom<&NewNewContext<OpenWorkspaceMode>> for NewNewContext<UnspecifiedWorkspaceMode> {
    type Error = anyhow::Error;

    fn try_from(
        value: &NewNewContext<OpenWorkspaceMode>,
    ) -> std::result::Result<Self, Self::Error> {
        // This can always happen
        Ok(NewNewContext {
            mode: PhantomData,
            git_repository: git2::Repository::open(&value.project.path)?,
            project: value.project.clone(),
        })
    }
}

impl TryFrom<&NewNewContext<UnspecifiedWorkspaceMode>> for NewNewContext<UnspecifiedWorkspaceMode> {
    type Error = anyhow::Error;

    fn try_from(
        value: &NewNewContext<UnspecifiedWorkspaceMode>,
    ) -> std::result::Result<Self, Self::Error> {
        // This can always happen
        Ok(NewNewContext {
            mode: PhantomData,
            git_repository: git2::Repository::open(&value.project.path)?,
            project: value.project.clone(),
        })
    }
}

impl NewNewContext<UnspecifiedWorkspaceMode> {
    pub fn open(_project: String) -> NewNewContext<UnspecifiedWorkspaceMode> {
        NewNewContext {
            mode: PhantomData,
            git_repository: todo!(),
            project: todo!(),
        }
    }
}

pub fn foo() -> Result<()> {
    let context: NewNewContext<UnspecifiedWorkspaceMode> =
        NewNewContext::open("testeroni".to_string());
    let narrowed_context: NewNewContext<OpenWorkspaceMode> =
        NewNewContext::<OpenWorkspaceMode>::try_from(&context)?;

    Ok(())
}
```

Caleb-T-Owens:
Is it safe to use mem::transmute if the only type chat gets changed is the phantom data?

```rs
impl From<&NewNewContext<OutsideWorkspaceMode>> for &NewNewContext<UnspecifiedWorkspaceMode> {
    fn from(value: &NewNewContext<OutsideWorkspaceMode>) -> Self {
        unsafe { mem::transmute(value) }
    }
}
```

Byron:
Regarding [this question](https://github.com/gitbutlerapp/gitbutler/pull/4506#issuecomment-2254177671), I meant a struct that identifies properties of things that can and can't be done with the current branch. Based on this, functions can decide if they are able to work with it or not. This would also mean that the checks would move inside respective function. Right now, that certainly wouldn't be helpful as most functions need `gitbutler/integration`.

And regarding `mem::transmute()`, using it for *this* definitely is 'unwritable' - I am glad the wizards agree based on merit.

Caleb-T-Owens:
Hmm... I do rather like the idea of having some form of type that indicates the state, but it seems like there isn't quite a silver bullet. 

I think for now, I'll just introduce some guards that can be called from a ctx (similar to `assure_unconflicted` or whatever its called), and use them in our commands, and in some key code paths like `list_virtual_branches`.

Caleb-T-Owens:
Another reason why using types is also potentially a bad idea is that we can change state mid-run, so having a function which actively does a check of reality has less chance of being misleading too

Caleb-T-Owens:
Closing in favour of a less intrusive approach

PR #4509, opened 2024-07-27
first trial of `git2-to-gix`

Byron:
When looking at `BranchListing` only, try to convert the usage of `git2` *related to branches* to using `gix` instead.

Do so without having while assuring that the output stays the same.

Follow-up of #4473.

### Tasks

* [x] a test utility to support the creation of read-only fixtures with GitButler setup
* [x] allow it to create vbranches as well
* [x] a all relevant tests for the `BranchListing` function to see the status quo (minimal)

### Follow-Up
* Maybe: don't pretend it's `async`? Sweeping changes. 
    - [x] #4517
* more elaborate testing
    - [x] #4543 
    - [x] #4564
    - [x] #4590
* convert `git2` to `gix`.
   - [x] #4632
   - [x] #4665
   - [x] #4670

### Review Notes

* I am not so sure about the sub-command hierarchy chosen here, and invite changes that are more in-line with the terminology of the application.
* I finally realized that `tauri::command(async)` is entirely independent of `async fn`, which means there is no need to pretend anymore. It seems no matter what, [`runtime::spawn` is used](https://v2.tauri.app/develop/calling-rust/), so it should be fine. I'd rather do that instead of having fake `async` functions, as `tauri` then at least knows these functions block so they should definitely be spawned as `spawn_blocking()`. Not lying about this makes it easy to call these functions from the CLI


krlvi:
> It seems no matter what, [runtime::spawn is used](https://v2.tauri.app/develop/calling-rust/), so it should be fine.

Sounds good, just note that the link is from the Tauri 2 docs (we are still on v1). I imagine that async behaviour is the same in v1 too.

krlvi (approved):

Byron:
Yes, I'd think that too, otherwise it's more like a bug to fix. Nonetheless, I will try the non-async version to assure it still feels snappy.

PR #4514, opened 2024-07-28
build: Reanme rust-toolchain.toml.nightly to rust-toolchain.toml

PavelLaptev:
I encountered an error when I tried to run `dev`:

Error Screenshot

@krlvi suggested renaming `rust-toolchain.toml.nightly` to `rust-toolchain.toml`, which resolved the issue. @Byron, could you please verify if this is the correct way to solve the problem?


Byron:
It's very strange that this didn't happen during my testing and also wasn't caught in CI, but I am on it now.

Byron:
I think the answer is that `async fn` in traits [was stabilized in Rust 1.75](https://caniuse.rs/features/async_fn_in_trait), and you want to run `rustup update stable` once @PavelLaptev.

PavelLaptev:
Thank you @Byron

Byron:
I will submit a PR that will make the error message of the compiler much clearer.
And here it is: #4532

PR #4517, opened 2024-07-28
no async

Byron:
Only use `async` where absolutely necessary or where it's unclear if it would
work similarly without `async`.

### Tasks

* [x] remove `async`
* [x] run `cargo fmt` with nightly settings to unify imports
* [x] validate


krlvi (approved):
works as before. really nice that we can make this simplification

PR #4520, opened 2024-07-28
Fix branch reordering

mtsgrd:
- move branches into their own div
- make only one update request


mtsgrd:
@krlvi fyi - should work better, but let me know if you have any thoughts on skipping oplog for these updates?

PR #4528, opened 2024-07-29
Introduce operating modes

Caleb-T-Owens:
It should be noted that I've not replaced all instances where we check for refs/heads/gitbutler/integration as they were in contexts it didn't have access to the project or command context, or it didn't feel appropriate to create a command context.

### DIFF
--- a/crates/gitbutler-operating-modes/src/lib.rs
+++ b/crates/gitbutler-operating-modes/src/lib.rs
@@ -1,0 +1,38 @@
+use anyhow::{bail, Context, Result};
+use gitbutler_command_context::CommandContext;
+
+/// Operating Modes:
+/// Gitbutler currently has two main operating modes:
+/// - `in workspace mode`: When the app is on the gitbutler/integration branch.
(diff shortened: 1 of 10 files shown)


crates/gitbutler-operating-modes/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+
+pub fn in_outside_workspace_mode(ctx: &CommandContext) -> Result<bool> {
+    in_open_workspace_mode(ctx).map(|open_mode| !open_mode)
+}
+
+pub fn assure_outside_workspace_mode(ctx: &CommandContext) -> Result<()> {
+    if in_open_workspace_mode(ctx)? {
+        Ok(())
+    } else {
+        bail!("Unexpected state: cannot perform operation on non-integration branch")
```
Caleb-T-Owens, line 36: I'm aware this is currently incorrect. This function is unused so won't be causing an current problems, and I already have a fix in place

PR #4531, opened 2024-07-29
Use seperate folder for rust analzyer

Caleb-T-Owens:
By asking rust analzyer to use a seperate
folder, it means the cache doesn't get
invalidated when running tests or by
running the application under different 
compiler flags.

Having the cache invalidated
dramatically increases the compile times.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ storybook-static
 
 # Vercel
 .vercel
+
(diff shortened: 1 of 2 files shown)


Byron:
I could so use this for IntelliJ! It trashes the caches all the time, it's obscene :/.

PR #4534, opened 2024-07-29
Use octopus commit for creating gitbutler/integration

mtsgrd:
This way when you e.g. `git log --graph` you'll see the individual branching and their commits.

### DIFF
--- a/crates/gitbutler-branch-actions/src/integration.rs
+++ b/crates/gitbutler-branch-actions/src/integration.rs
@@ -38,19 +38,8 @@ pub(crate) fn get_workspace_head(ctx: &CommandContext) -> Result<git2::Oid> {
 
     let mut virtual_branches: Vec<Branch> = vb_state.list_branches_in_workspace()?;
 
-    let branch_heads = virtual_branches
-        .iter()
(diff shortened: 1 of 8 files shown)


mtsgrd:
@Byron fyi removing a few test cases since it is no longer possible to list branches without having a base branch set. It would throw on not finding a default target.

@Caleb-T-Owens fyi, here's the pr we talked about, let me know if you have any thoughts.

crates/gitbutler-branch-actions/src/integration.rs
```diff
@@ -330,28 +319,31 @@ fn verify_head_is_clean(ctx: &CommandContext, perm: &mut WorktreeWritePermission
...
+        c.message()
+            .is_some_and(|message| message.starts_with("GitButler Integration Commit"))
+    });
 
-    if integration_commit.is_none() {
+    if integration_index.is_none() {
         // no integration commit found
         bail!("gibButler's integration commit not found on head");
     }
+    let integration_index = integration_index.unwrap();
```
mtsgrd: todo: will remove this unwrap.

PR #4537, opened 2024-07-30
fix: skip turbo in init-node-env

ndom91:
## 🧢 Changes

- Skip turbo and use pnpm cmd directly in the correct directory


krlvi (approved):
appears to fix the issue in powershell

PR #4542, opened 2024-07-30
feat: clone repo onboarding step

ndom91:
## ☕️ Reasoning

- Add "Clone Repo" onboarding step, next to "Add Local Project"

## 🧢 Changes

- Add `/onboarding` routes to group and organize onboarding related pages / components
- Adds additional onboarding wizard route for cloning a user's repository and adding it as a project
- After either "Add local project" or "Clone project" the user is redirected to `/{projectId}/board`
- Remove `components/SegmentController/*` from app in favor of the new copies of those files in the UI library
- Updated `SegmentController` props naming

### Todo

- [x] Still needs some error handling in the `/[projectId]/*` routes to redirect folks back to `/onboarding` if they don't have a project id or some other error arises
- [x] Rust fn for `git_clone_repo(repoUrl: string, targetDirectory: string) => void`
- [x] Double check the new routing logic 
- [x] Clean up


mtsgrd:
Would it make sense if you could also clone a new repo outside the onboarding flow, the way you can currently add a new project?

apps/desktop/src/lib/components/Welcome.svelte
```diff
@@ -18,19 +21,48 @@
...
 	}
+
+	async function onCloneProject() {
+		cloneProjectLoading = true;
+		try {
+			goto('/onboarding/clone');
+		} finally {
+			cloneProjectLoading = false;
+		}
+	}
```
mtsgrd: Can there be a delay between calling `goto` and the navigation actually happening? Otherwise it feels like we don't need the `cloneProjectLoading` state here.

ndom91:
> Would it make sense if you could also clone a new repo outside the onboarding flow, the way you can currently add a new project?

Yeah actually this is a part of Pavels upcoming sidebar changes

apps/desktop/src/lib/components/Welcome.svelte
ndom91: Hmm yeah no probably not. This whole fn can most likely be simplified to just the goto 👍

PR #4543, opened 2024-07-30
branch listing in `gix`

Byron:
This PR is a follow-up to #4509 to ideally lure out more branch-specific behaviour in testing to assure the transition to `gix` doesn't break anything.

### Tasks

* [x] ability to create stable commit-graphs (reproducible hashes)
    - This can generally help testing, even though it's probably not needed here
* [x] tests for special cases of branch handling (but more to be done)

### Follow-up
* [ ] transition to `gix`

### Notes for the Reviewer

* Even though I am trying to write proper documentation for fields and structs, it might very well be that it misses the point or is too partial. Please do scrutinize and amend or change as needed.
* This PR also improves error handling in regards to vbranch names, as fly-by fix.


crates/gitbutler-branch-actions/src/branch.rs
```diff
@@ -22,82 +21,68 @@ pub fn list_branches(
     ctx: &CommandContext,
     filter: Option<BranchListingFilter>,
 ) -> Result<Vec<BranchListing>> {
+    let has_filter = filter.is_some();
+    let filter = filter.unwrap_or_default();
     let vb_handle = ctx.project().virtual_branches();
     // The definition of "own_branch" is based if the current user made the first commit on the branch
```
krlvi: On note here - the comment is already out of date unfortunately. I realized that the question that the UI is trying to answer is "what are the branches that the user is involved in" - and offer segmented control in the UI so that the user can more easily focus on their branches.
Thus the definition changed to "branches that the user has committed to" (and if they are virtual branches without commit that they created, also include that). The implementation below already reflects this definition, just not the comment here

crates/gitbutler-reference/src/refname/virtual.rs
```diff
@@ -6,7 +6,9 @@ use super::error::Error;
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct Refname {
-    // contains slug of the virtual branch name
+    /// contains slug of the virtual branch name.
+    /// Note that `Display` is implemented to put `refs/gitbutler/` in front of `branch`
+    /// so it's assume to be a valid branch name otherwise.
     pub branch: String,
```
krlvi, line 12: Just FYI, the entire `gitbutler-reference` crate can be considered on the "kill list". It's a bit of an incorrect abstraction and there are bunch of incorrect things that are modeled in it. I'd love to move away from it when we can, its just that it's used in a lot of places and it's removal is not straight forward.

(btw, this comment is not related to the docs update diff, i just needed a place to comment on the PR)

Byron: Thanks for letting me know. Unless you think it's a problem, I'd keep changing code there as well, assuming that even if the crate should be dissolved, the code will just move to other places and/or be refactored.

In a way, this is what I think I will end up doing in small steps, even though I am less concerned about where code lives.

crates/gitbutler-branch-actions/src/branch.rs
Byron: Alright, I will update the comment (or remove it, because I think that's tested now), and update the docs for the filter accordingly.

krlvi (approved):

PR #4548, opened 2024-07-30
prevent trashing when building with rustc in Jetbrains IDEs.

Byron:
For the IDE to discover what's going on, it will use hidden environment variables to perform the build with, but that also affects certain crates which then report they changed.

If this is intermixed with actual nightly builds, unnecessary rebuilds happen all the time. To prevent this, just use `nightly` by default without actually using nightly features.


krlvi:
👍

Byron:
Thanks for your support.
I am still evaluating this while working on #4543 - at the end of this, I will merge or not depending on the outcome.

Byron (commented):
This definitely works for me.

PR #4562, opened 2024-07-31
fork exec with futures instead of tokio

krlvi:
@Byron do you think its a good idea to just block for the fork exec with futures instead of starting a whole tokio thing?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2264,6 +2264,7 @@ version = "0.0.0"
 dependencies = [
  "anyhow",
  "bstr",
+ "futures",
  "git2",
  "gitbutler-branch",
  "gitbutler-command-context",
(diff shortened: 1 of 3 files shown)


krlvi:
okay probably this doesnt work

Byron (commented):
I saw that too but was thinking that maybe these futures need a tokio executor to work, which is why I didn't dare to change it.
However, it seems to work without which is a little unexpected, but very welcome, too.
Edit: It seems it was needed after all, but see below, `tokio` shouldn't be needed, nor should async be.

My feeling is that this (ssh prompt handling) can be done entirely without async and the respective complications, it feels like it's much more than it has to be, but I will need to validate that hunch (one day).

krlvi:
it worked in my manual testing of it but some tests failed in CI... im relatively confused

Byron:
Every tokio-future, i.e. every async call in `tokio` needs the tokio runtime to work. In the application, that's the case naturally , but in stripped-down test environments that's not the case, particularly now that `tokio::test` isn't used anymore.

PR #4564, opened 2024-08-01
branch-listing tests and refactors

Byron:
This PR is a follow-up to #4543 and contains a thorough review of the new branch-listing feature, with a focus on adding tests to pin-point (and understand) the current behaviour and intentions.
It's fair to use this PR to do refactorings as well.
When done, it should be trivial to use `gix` for the initial branch listing.

<details>

352271590-6604004a-4553-4b23-9675-cdf4069b3f0c

</details>

### Tasks

* [x] fix first logic bug
* [x] more testing

### Follow-up

* make `Identity` clear in the type system.

### Notes for the reviewer

* I started making the commit-graph stable (to allow making assertions on it using the commit ID), and the only thing preventing it is the `gitbutler-change-id` which currently is a UUID. It seems unused right now, but it would be good do learn what is planned there.


krlvi (approved):
lgtm lets gooo :)

PR #4590, opened 2024-08-03
make the notion of `identity` typesafe.

Byron:
As follow-up to #4509, this PRs adds a type for the `Identity` of a branch primarily to make clear its importance, and some assumptions that have to be made to make the application more pleasant to use.

Even with it, there is likely still assumptions that actually can't be made, but it's a first step towards using branches more correctly.

### DIFF
--- a/crates/gitbutler-branch-actions/src/branch.rs
+++ b/crates/gitbutler-branch-actions/src/branch.rs
@@ -19,7 +19,7 @@ use crate::VirtualBranchesExt;
 pub fn list_branches(
     ctx: &CommandContext,
     filter: Option<BranchListingFilter>,
-    filter_branch_names: Option<Vec<String>>,
+    filter_branch_names: Option<Vec<BranchIdentity>>,
 ) -> Result<Vec<BranchListing>> {
     let has_filter = filter.is_some();
     let filter = filter.unwrap_or_default();
@@ -30,7 +30,7 @@ pub fn list_branches(
         if let Some(branch_names) = &filter_branch_names {
             let has_matching_name = branch_names.iter().any(|branch_name| {
                 if let Ok(Some(name)) = branch.name() {
-                    name.ends_with(branch_name)
+                    name.ends_with(&branch_name.0)
                 } else {
                     false
                 }
@@ -92,7 +92,7 @@ fn combine_branches(
     let remotes = repo.remotes()?;
 
     // Group branches by identity
-    let mut groups: HashMap<String, Vec<GroupBranch>> = HashMap::new();
+    let mut groups: HashMap<BranchIdentity, Vec<GroupBranch>> = HashMap::new();
     for branch in group_branches {
         let Some(identity) = branch.identity(&remotes) else {
             continue;
@@ -127,7 +127,7 @@ fn combine_branches(
 
 /// Converts a group of branches with the same identity into a single branch entry
 fn branch_group_to_branch(
-    identity: &str,
+    identity: &BranchIdentity,
     group_branches: Vec<GroupBranch>,
     repo: &git2::Repository,
     remotes: &git2::string_array::StringArray,
@@ -258,7 +258,7 @@ impl GroupBranch<'_> {
     /// A name identifier for the branch. When multiple branches (e.g. virtual, local, remote) have the same identity,
     /// they are grouped together under the same `Branch` entry.
     /// `None` means an identity could not be obtained, which makes this branch odd enough to ignore.
-    fn identity(&self, remotes: &git2::string_array::StringArray) -> Option<String> {
+    fn identity(&self, remotes: &git2::string_array::StringArray) -> Option<BranchIdentity> {
         match self {
             GroupBranch::Local(branch) => branch.get().given_name(remotes).ok(),
             GroupBranch::Remote(branch) => branch.get().given_name(remotes).ok(),
@@ -272,12 +272,13 @@ impl GroupBranch<'_> {
                 Some(identity.to_string())
             }
         }
+        .map(BranchIdentity)
     }
 }
 
 /// Determines if a branch should be listed in the UI.
 /// This excludes the target branch as well as gitbutler specific branches.
-fn should_list_git_branch(identity: &str) -> bool {
+fn should_list_git_branch(identity: &BranchIdentity) -> bool {
     // Exclude gitbutler technical branches (not useful for the user)
     let is_technical = [
         "gitbutler/integration",
@@ -285,7 +286,7 @@ fn should_list_git_branch(identity: &str) -> bool {
         "gitbutler/oplog",
         "HEAD",
     ]
-    .contains(&identity);
+    .contains(&&*identity.0);
     !is_technical
 }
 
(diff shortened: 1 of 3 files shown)


Byron:
Vercel seems to work now!

crates/gitbutler-branch-actions/src/branch.rs
```diff
@@ -347,6 +348,32 @@ pub struct Author {
...
+/// The identity of a branch as to allow to group similar branches together.
+///
+/// * For *local* branches, it is what's left without the standard prefix, like `refs/heads`, e.g. `main`
+///   for `refs/heads/main` or `feat/one` for `refs/heads/feat/one`.
+/// * For *remote* branches, it is what's without the prefix and remote name, like `main` for `refs/remotes/origin/main`.
+///   or `feat/one` for `refs/remotes/my/special/remote/feat/one`.
+/// * For virtual branches, it's either the above if there is a `source_refname` or an `upstream`, or it's the normalized
+///   name of the virtual branch.
+#[derive(Debug, Clone, Serialize, PartialEq, Eq, Hash, Ord, PartialOrd)]
+pub struct BranchIdentity(String);
```
Caleb-T-Owens, line 355: Why is it preferable to use a separate struct over a TaggedString here?

Byron: That's a great point - I wasn't aware of `TaggedString`, let me try it.

Byron: I have a patch here that works but think it increases complexity and doesn't come with benefits over a minimal type as was implemented before.

```patch
commit 9b8488ad46d2f40b9781576b4ab138da762262cb
Date:   Mon Aug 5 09:38:12 2024 +0200

    Use tagged-string instead of custom type with the same purpose.
    
    Thanks Caleb for the suggestion.

index 04481683b..7a029ccd1 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2033,6 +2033,7 @@ dependencies = [
  "gitbutler-reference",
  "gitbutler-repo",
  "gitbutler-serde",
+ "gitbutler-tagged-string",
  "gitbutler-testsupport",
  "gitbutler-time",
  "gitbutler-url",
diff --git a/crates/gitbutler-branch-actions/Cargo.toml b/crates/gitbutler-branch-actions/Cargo.toml
index 82c5a5a83..d408d160b 100644
--- a/crates/gitbutler-branch-actions/Cargo.toml
+++ b/crates/gitbutler-branch-actions/Cargo.toml
@@ -24,6 +24,7 @@ gitbutler-commit.workspace = true
 gitbutler-url.workspace = true
 gitbutler-fs.workspace = true
 gitbutler-diff.workspace = true
+gitbutler-tagged-string.workspace = true
 gitbutler-operating-modes.workspace = true
 serde = { workspace = true, features = ["std"] }
 bstr = "1.10.0"
diff --git a/crates/gitbutler-branch-actions/src/branch.rs b/crates/gitbutler-branch-actions/src/branch.rs
index 1488ef49d..fb35f74ad 100644
--- a/crates/gitbutler-branch-actions/src/branch.rs
+++ b/crates/gitbutler-branch-actions/src/branch.rs
@@ -268,7 +268,7 @@ impl GroupBranch<'_> {
                 Some(identity.to_string())
             }
         }
-        .map(BranchIdentity)
+        .map(BranchIdentity::from)
     }
 }
 
@@ -282,7 +282,7 @@ fn should_list_git_branch(identity: &BranchIdentity) -> bool {
         "gitbutler/oplog",
         "HEAD",
     ]
-    .contains(&&*identity.0);
+    .contains(&&***identity);
     !is_technical
 }
 
@@ -356,23 +356,10 @@ pub struct Author {
 ///   or `feat/one` for `refs/remotes/my/special/remote/feat/one`.
 /// * For virtual branches, it's either the above if there is a `source_refname` or an `upstream`, or it's the normalized
 ///   name of the virtual branch.
-#[derive(Debug, Clone, Serialize, PartialEq, Eq, Hash, Ord, PartialOrd)]
-pub struct BranchIdentity(String);
-
-/// Facilitate obtaining this type from the UI - otherwise it would be better not to have it as it should be
-/// a particular thing, not any string.
-impl From<String> for BranchIdentity {
-    fn from(value: String) -> Self {
-        BranchIdentity(value)
-    }
-}
+pub type BranchIdentity = gitbutler_tagged_string::TaggedString<IdentityTag>;
 
-/// Also not for testing.
-impl From<&str> for BranchIdentity {
-    fn from(value: &str) -> Self {
-        BranchIdentity(value.into())
-    }
-}
+#[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub struct IdentityTag;
 
 impl From<git2::Signature<'_>> for Author {
     fn from(value: git2::Signature) -> Self {
diff --git a/crates/gitbutler-tagged-string/src/lib.rs b/crates/gitbutler-tagged-string/src/lib.rs
index 2e95136ef..4df83d0df 100644
--- a/crates/gitbutler-tagged-string/src/lib.rs
+++ b/crates/gitbutler-tagged-string/src/lib.rs
@@ -1,8 +1,8 @@
 use std::{fmt, marker::PhantomData, ops::Deref};
 
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
-
 /// Tagged string is designed to clarify the purpose of strings when used as a return type
+#[derive(Serialize, Deserialize, Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
 pub struct TaggedString<T>(String, PhantomData<T>);
 
 impl<T> From<String> for TaggedString<T> {
@@ -25,24 +25,6 @@ impl<T> Deref for TaggedString<T> {
     }
 }
 
-impl<'de, T> Deserialize<'de> for TaggedString<T> {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        String::deserialize(deserializer).map(Into::into)
-    }
-}
-
-impl<T> Serialize for TaggedString<T> {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        self.0.serialize(serializer)
-    }
-}
-
 impl<T> fmt::Display for TaggedString<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         self.0.fmt(f)
```

Something I don't fancy is to have two types now, one as a tag (which unfortunately needs to derive a bunch of traits for no good reason :/), and one that is the actual type used in the struct.

Overall, I'd hope that there would be no need for any kind of `TaggedString` or `BranchIdentity` as they don't add a guarantee that the value it contains is valid - they are merely a useful programming aid.

Please let me know if you'd like the patch applied nonetheless.

krlvi:
Oh, i somehow thought this was merged. In any case, I think with this one https://github.com/gitbutlerapp/gitbutler/pull/4557 merged we can get this one in too

Byron:
I have fixed the merge-conflicts and used the `BranchIdentity` instead of Strings - it should work like before even though I didn't test it beyond running the existing tests.

Byron:
After running the app and playing with the new branch-listing, I came to believe it's working as intended. During testing on the `cmake-rs` repository, I ran into `Failed to find Branch` error boxes, but could reproduce the same on `master`.

Thus it should be fine for me to merge this PR.

PR #4592, opened 2024-08-03
fix: add two additional tsconfig options

ndom91:
## 🧢 Changes

- 2 new `tsconfig.json` options
	- `noUncheckedIndexAccess: true` - [[Doc]](https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess) - Does what it says on the tin, no unchecked access on arrays and objects. Seems like a valuable rule to have on to potentially avoid runtime errors
	- `verbatimModuleSyntax: true` - [[Doc]](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax) - Forces us to use `import type { ..` syntax when importing types, which (a) is more correct imo and (b) causes those import statements to be transpiled out during build so we ship less code

### Todo

- [x] If you all are okay with the `noUncheckedIndexAccess` TS option, then this PR needs a bit of TS clean up before we can merge it, as we have a few instances where we violate this rule. I can do that here though


apps/desktop/tsconfig.json
```diff
@@ -7,7 +7,9 @@
 		"checkJs": true,
 		"esModuleInterop": true,
 		"forceConsistentCasingInFileNames": true,
+		"noUncheckedIndexedAccess": true,
```
Caleb-T-Owens, line 10: yess!!!

Caleb-T-Owens: noUncheckedIndexAccess should be a default option in my mind. I'm for it all the way.

I'm not quite so familiar with verbatimModuleSyntax but it seems to be the new recommended option, so I'm happy to go for it

ndom91:
Updated all the failing instances, waiting for some other PRs to be merged so I can address any new files / failing instances here.

apps/desktop/src/lib/branch/BranchLane.svelte
```diff
@@ -44,7 +44,9 @@
 	// BRANCH SERVICE
 	const prService = createGitHostPrServiceStore(undefined);
 	$effect(() =>
-		prService.set(upstreamName ? $gitHost?.prService(baseBranchName, upstreamName) : undefined)
+		prService.set(
+			upstreamName ? $gitHost?.prService(baseBranchName ?? '', upstreamName) : undefined
```
mtsgrd: Maybe do `upstreamName && baseBranchName` instead of passing empty string?

ndom91: Yeah good idea, fixed

mtsgrd:
There more type coercions here than I expected. Would it be possible to try work around the need for some of the `!` and `as` expressions?

ndom91:
> There more type coercions here than I expected. Would it be possible to try work around the need for some of the `!` and `as` expressions?

Yeah I can spend some more time on this, however, a lot of the coercions / use of `!` come from situations where we do return early if something is wrong, but TS doesn't infer the implications of this on the resulting types. For example: https://github.com/gitbutlerapp/gitbutler/pull/4592/files#diff-6b5453e3f4f31a5ea4ae9775b821bc89f83eb7c0b304205c83f61a5a37307086

image

Caleb-T-Owens:
> > There more type coercions here than I expected. Would it be possible to try work around the need for some of the `!` and `as` expressions?
> 
> Yeah I can spend some more time on this, however, a lot of the coercions / use of `!` come from situations where we do return early if something is wrong, but TS doesn't infer the implications of this on the resulting types. For example: https://github.com/gitbutlerapp/gitbutler/pull/4592/files#diff-6b5453e3f4f31a5ea4ae9775b821bc89f83eb7c0b304205c83f61a5a37307086
> 
> image
> 
> I guess in this situation `selectedBranch[0]` cuold be `undefined`, but then that initial check would also blow up, right?
> 
> **EDIT**: Yeah okay so additional more explicit checks in the early returns do get picked up by TS inference, i.e.:
> 
> image

In this example, it very much could be undefined 😆 `undefined === "" // false`

PR #4595, opened 2024-08-04
feat: add `@gitbutler/ui`, `userService`, `authService` and some minor cleanup

ndom91:
## 🧢 Changes

- Added `@gitbutler/ui` dependency from workspace, in order to import our global `main.css` (includes all our css vars as well as the standard css reset stuff)
	- Remember to use `pnpm dev:web` in the root of the repo in order to transpile and then also watch the `packages/ui/*` directory for changes
- Add Sentry (explicitly disabled currently; remove `if (false)` in `apps/web/src/lib/analytics/sentry.ts` to enable)
- Add a `userService` and an `authService` to handle the user metadata and `gb_access_token` respectively
- Extract navbar into own component
- Cleanup some minor style stuff
- Add persisted storage helper to store token in localstorage
- Cleanup `<title>`, `favicon`'s, and that sort of stuff


apps/web/src/lib/persisted.svelte.ts
```diff
@@ -0,0 +1,132 @@
...
+		console.error(`Error when writing value from persisted store "${key}" to ${storage}`, e);
+	}
+}
+
+function getStorage(): Storage | null {
+	if (typeof window === 'undefined') {
+		return null;
+	}
+
+	return localStorage;
```
Byron, line 80: Is there a way  to store secrets securely (encrypted at rest) in local storage on all platforms?
If not, I believe these should be persisted in memory only, and re-fetched on application startup as the backend will store them securely on all platforms.

apps/web/src/routes/+layout.svelte
```diff
@@ -1,52 +1,36 @@
...
+
+	$effect(() => {
 		if ($page.url.searchParams.get('gb_access_token')) {
-			let token = $page.url.searchParams.get('gb_access_token');
+			const token = $page.url.searchParams.get('gb_access_token');
 			if (token && token.length > 0) {
-				localStorage.setItem('gb_access_token', token);
-				// redirect to remove search params
-				window.location.href = '/';
+				authService.setToken(token);
```
Byron, line 21: Unless we are sure tokens are encrypted at rest on all platforms, I'd think we better store them using the backend, but see no issue to keeping them in-memory to reduce the amount of calls (it's pretty slow, after all).

Byron (commented):
I have a concern related to storing secrets in `localStorage` as I fear this storage type has no guarantee to be encrypted at rest *on all platforms*. 
If that could indeed be the case, I'd want to avoid creating any release with this change active so the gitbutler access token doesn't get stored to disk somewhere for malware to read.

CC @krlvi @mtsgrd

mtsgrd:
Thanks @Byron, I agree, it's not ideal to use localStorage. There are actually more problems with storing a token in localStorage than encryption at rest. https://stackoverflow.com/a/44209185

But let's make sure to fix this in the near term rather than right away? Our existing pages also use a token and it'll take a couple of steps to swap things out.

apps/web/src/routes/+layout.svelte
ndom91: Yeah definitely a fair point. 

If we want to persist the token though there's also the argument for using cookies instead of localstorage because of their httpOnly flag so they can't be targeted from JS. 

Right now I just cleaned up what was already there and didn't really change any fundamental functionality. I think this is very much still in the experimental/prototype stage. 

But this is definitely a conversation worth having. I'll bring it up with the folks in the office this week too 🙏

ndom91: Also sidenote, this is the web app, it's not included in/related to the tauri app, right. Just so we're on the same page. We had already stored it in localstorage previous to this PR in the web app, like I said I just moved stuff around a bit

Byron:
Thanks, also for the heads-up on XSS! I also didn't realize this is more of a refactoring than a change in how it works. 

But I did notice that the released GB doesn't store the token in local storage either, or at least it can't be found when looking for `token` in the developer storage view. Even the most recent nightly after a fresh GB-login won't show it in local storage. It is, however, placed in the secure storage provided by the backend. In practice, it seems to work as intended.

In any case, since you have it on your radar, I will get off your backs about that 😁.

krlvi:
Since we now have a monorepo, it would be cool to add some kinda tagging for the "webapp" changes

ndom91:
> Since we now have a monorepo, it would be cool to add some kinda tagging for the "webapp" changes

There was a typo in the `pr-labeler.yml` config, but it does now (after this PR :sweat_smile) correctly tag PR's with `@gitbutler/web`, for example. In addition to the pre-existing `@gitbutler/desktop` and `@gitbutler/ui` tags

PR #4640, opened 2024-08-07
Write tests for existing modes

Caleb-T-Owens:
This PR introduces some tests to ensure they are doing what they say they do on the tin, and to prevent regressions when we increase the complexity.

The tests I've added are as follows:
```
{} operating_modes
  {} open_workspace_mode
    in_open_workspace_mode_true_when_in_gitbutler_integration fn()
    in_open_workspace_mode_false_when_on_other_branches fn()
    assure_open_workspace_mode_ok_when_on_gitbutler_integration fn()
    assure_open_workspace_mode_err_when_on_other_branch fn()
  {} outside_workspace_mode
    in_outside_workspace_mode_true_when_in_gitbutler_integration fn()
    in_outside_workspace_mode_false_when_on_other_branches fn()
    assure_outside_workspace_mode_ok_when_on_gitbutler_integration fn()
    assure_outside_workspace_mode_err_when_on_other_branch fn()
```

PR #4642, opened 2024-08-07
Add in edit mode variant (without logic for transitioning)

Caleb-T-Owens:
This is a stacked PR; visit https://github.com/gitbutlerapp/gitbutler/compare/write-tests-for-existing-gitbutler-modes...add-edit-mode for a more accurate diff.

This PR introduces a third "edit" mode. This state will be used to indicate when the app has been switch to a special `gitbutler/edit` branch in order to edit a particular commit. 

I've now got an enum with a variant for each mode, and a function `operating_mode(&ctx)` which can be used to get the current context.

```rs
#[derive(Debug, Serialize, Deserialize, PartialEq)]
pub struct EditModeMetadata {
    #[serde(with = "gitbutler_serde::serde::oid")]
    pub editee_commit_sha: git2::Oid,
    pub editee_branch: String,
}

#[derive(PartialEq)]
pub enum OperatingMode {
    /// The typical app state when its on the gitbutler/integration branch
    InWorkspace,
    /// When the user has chosen to leave the gitbutler/integration branch
    OutsideWorkspace,
    /// When the app is off of gitbutler/integration and in edit mode
    Edit(EditModeMetadata),
}
```

I've also added in two methods for checking and assuring you're specifically on edit mode, and also updated the other functions to now check the enum.

```
pub fn in_edit_mode(ctx: &CommandContext) -> Result<bool> {
    match operating_mode(ctx)? {
        OperatingMode::Edit(_) => Ok(true),
        _ => Ok(false),
    }
}

pub fn assure_edit_mode(ctx: &CommandContext) -> Result<EditModeMetadata> {
    match operating_mode(ctx)? {
        OperatingMode::Edit(edit_mode_metadata) => Ok(edit_mode_metadata),
        _ => bail!("Expected to be in edit mode"),
    }
}
```

When the app is in edit mode, it expects a `.git/gitbutler/edit_mode_metadata.toml` file to be written. If we fail to find or read that file, we default the user back to OutsideWorkspace mode
```rs
pub fn operating_mode(ctx: &CommandContext) -> Result<OperatingMode> {
    let head_ref = ctx.repository().head().context("failed to get head")?;
    let head_ref_name = head_ref.name().context("failed to get head name")?;

    if head_ref_name == INTEGRATION_BRANCH_REF {
        Ok(OperatingMode::InWorkspace)
    } else if head_ref_name == EDIT_BRANCH_REF {
        let edit_mode_metadata = read_edit_mode_metadata(ctx);

        match edit_mode_metadata {
            Ok(edit_mode_metadata) => Ok(OperatingMode::Edit(edit_mode_metadata)),
            Err(error) => {
                tracing::warn!(
                    "Failed to open in edit mode, falling back to outside workspace {}",
                    error
                );
                Ok(OperatingMode::OutsideWorkspace)
            }
        }
    } else {
        Ok(OperatingMode::OutsideWorkspace)
    }
}
```

### DIFF
--- a/crates/gitbutler-operating-modes/src/lib.rs
+++ b/crates/gitbutler-operating-modes/src/lib.rs
@@ -1,38 +1,123 @@
+use std::{fs, path::PathBuf};
+
 use anyhow::{bail, Context, Result};
 use gitbutler_command_context::CommandContext;
+use gitbutler_reference::ReferenceName;
+use serde::{Deserialize, Serialize};
 
-/// Operating Modes:
-/// Gitbutler currently has two main operating modes:
-/// - `in workspace mode`: When the app is on the gitbutler/integration branch.
-///     This is when normal operations can be performed.
-/// - `outside workspace mode`: When the user has left the gitbutler/integration
-///     branch to perform regular git commands.
+/// The reference the app will checkout when the workspace is open
+pub const INTEGRATION_BRANCH_REF: &str = "refs/heads/gitbutler/integration";
+/// The reference the app will checkout when in edit mode
+pub const EDIT_BRANCH_REF: &str = "refs/heads/gitbutler/edit";
 
-const INTEGRATION_BRANCH_REF: &str = "refs/heads/gitbutler/integration";
-
-pub fn in_open_workspace_mode(ctx: &CommandContext) -> Result<bool> {
-    let head_ref = ctx.repository().head().context("failed to get head")?;
-    let head_ref_name = head_ref.name().context("failed to get head name")?;
-
-    Ok(head_ref_name == INTEGRATION_BRANCH_REF)
+fn edit_mode_metadata_path(ctx: &CommandContext) -> PathBuf {
+    ctx.project().gb_dir().join("edit_mode_metadata.toml")
 }
 
-pub fn assure_open_workspace_mode(ctx: &CommandContext) -> Result<()> {
-    if in_open_workspace_mode(ctx)? {
-        Ok(())
+#[doc(hidden)]
+pub fn read_edit_mode_metadata(ctx: &CommandContext) -> Result<EditModeMetadata> {
+    let edit_mode_metadata = fs::read_to_string(edit_mode_metadata_path(ctx).as_path())
+        .context("Failed to read edit mode metadata")?;
+
+    toml::from_str(&edit_mode_metadata).context("Failed to parse edit mode metadata")
+}
+
+#[doc(hidden)]
+pub fn write_edit_mode_metadata(
+    ctx: &CommandContext,
+    edit_mode_metadata: &EditModeMetadata,
+) -> Result<()> {
+    let serialized_edit_mode_metadata =
+        toml::to_string(edit_mode_metadata).context("Failed to serialize edit mode metadata")?;
+    gitbutler_fs::write(
+        edit_mode_metadata_path(ctx).as_path(),
+        serialized_edit_mode_metadata,
+    )
+    .context("Failed to write edit mode metadata")?;
+
+    Ok(())
+}
+
+/// Holds relevant state required to switch to and from edit mode
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct EditModeMetadata {
+    /// The sha of the commit getting edited.
+    #[serde(with = "gitbutler_serde::oid")]
+    pub editee_commit_sha: git2::Oid,
+    /// The ref of the vbranch which owns this commit.
+    pub editee_branch: ReferenceName,
+}
+
+#[derive(PartialEq)]
+pub enum OperatingMode {
+    /// The typical app state when its on the gitbutler/integration branch
+    OpenWorkspace,
+    /// When the user has chosen to leave the gitbutler/integration branch
+    OutsideWorkspace,
+    /// When the app is off of gitbutler/integration and in edit mode
+    Edit(EditModeMetadata),
+}
+
+pub fn operating_mode(ctx: &CommandContext) -> OperatingMode {
+    let Ok(head_ref) = ctx.repository().head() else {
+        return OperatingMode::OutsideWorkspace;
+    };
+
+    let Some(head_ref_name) = head_ref.name() else {
+        return OperatingMode::OutsideWorkspace;
+    };
+
+    if head_ref_name == INTEGRATION_BRANCH_REF {
+        OperatingMode::OpenWorkspace
+    } else if head_ref_name == EDIT_BRANCH_REF {
+        let edit_mode_metadata = read_edit_mode_metadata(ctx);
+
+        match edit_mode_metadata {
+            Ok(edit_mode_metadata) => OperatingMode::Edit(edit_mode_metadata),
+            Err(error) => {
+                tracing::warn!(
+                    "Failed to open in edit mode, falling back to outside workspace {}",
+                    error
+                );
+                OperatingMode::OutsideWorkspace
+            }
+        }
     } else {
-        bail!("Unexpected state: cannot perform operation on non-integration branch")
+        OperatingMode::OutsideWorkspace
     }
 }
 
-pub fn in_outside_workspace_mode(ctx: &CommandContext) -> Result<bool> {
-    in_open_workspace_mode(ctx).map(|open_mode| !open_mode)
+pub fn in_open_workspace_mode(ctx: &CommandContext) -> bool {
+    operating_mode(ctx) == OperatingMode::OpenWorkspace
+}
+
+pub fn assure_open_workspace_mode(ctx: &CommandContext) -> Result<()> {
+    if in_open_workspace_mode(ctx) {
+        Ok(())
+    } else {
+        bail!("Expected to be in open workspace mode")
+    }
+}
+
+pub fn in_edit_mode(ctx: &CommandContext) -> bool {
+    matches!(operating_mode(ctx), OperatingMode::Edit(_))
+}
+
+pub fn assure_edit_mode(ctx: &CommandContext) -> Result<EditModeMetadata> {
+    match operating_mode(ctx) {
+        OperatingMode::Edit(edit_mode_metadata) => Ok(edit_mode_metadata),
+        _ => bail!("Expected to be in edit mode"),
+    }
+}
+
+pub fn in_outside_workspace_mode(ctx: &CommandContext) -> bool {
+    operating_mode(ctx) == OperatingMode::OutsideWorkspace
 }
 
 pub fn assure_outside_workspace_mode(ctx: &CommandContext) -> Result<()> {
-    if in_outside_workspace_mode(ctx)? {
+    if in_outside_workspace_mode(ctx) {
         Ok(())
     } else {
-        bail!("Unexpected state: cannot perform operation on integration branch")
+        bail!("Expected to be in outside workspace mode")
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2275,8 +2275,13 @@ dependencies = [
  "anyhow",
  "git2",
  "gitbutler-command-context",
+ "gitbutler-fs",
+ "gitbutler-reference",
+ "gitbutler-serde",
  "gitbutler-testsupport",
  "serde",
+ "toml 0.8.15",
+ "tracing",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,8 @@ resolver = "2"
 [workspace.dependencies]
 bstr = "1.10.0"
 # Add the `tracing` or `tracing-detail` features to see more of gitoxide in the logs. Useful to see which programs it invokes.
-gix = { git = "https://github.com/Byron/gitoxide", rev = "d51f330e9d364c6f7b068116b59bf5c0160e47af", default-features = false, features = [] }
+gix = { git = "https://github.com/Byron/gitoxide", rev = "d51f330e9d364c6f7b068116b59bf5c0160e47af", default-features = false, features = [
+] }
 git2 = { version = "0.18.3", features = [
     "vendored-openssl",
     "vendored-libgit2",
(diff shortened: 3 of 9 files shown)


krlvi:
> This is a stacked PR; visit [write-tests-for-existing-gitbutler-modes...add-edit-mode](https://github.com/gitbutlerapp/gitbutler/compare/write-tests-for-existing-gitbutler-modes...add-edit-mode) for a more accurate diff.
> 
> This PR introduces a third "edit" mode. This state will be used to indicate when the app has been switch to a special `gitbutler/edit` branch in order to edit a particular commit.
> 
> I've now got an enum with a variant for each mode, and a function `operating_mode(&ctx)` which can be used to get the current context.
> 
> ```rust
> #[derive(Debug, Serialize, Deserialize, PartialEq)]
> pub struct EditModeMetadata {
>     #[serde(with = "gitbutler_serde::serde::oid")]
>     pub editee_commit_sha: git2::Oid,
>     pub editee_branch: String,
> }
> 
> #[derive(PartialEq)]
> pub enum OperatingMode {
>     /// The typical app state when its on the gitbutler/integration branch
>     InWorkspace,
>     /// When the user has chosen to leave the gitbutler/integration branch
>     OutsideWorkspace,
>     /// When the app is off of gitbutler/integration and in edit mode
>     Edit(EditModeMetadata),
> }
> ```
> 
> I've also added in two methods for checking and assuring you're specifically on edit mode, and also updated the other functions to now check the enum.
> 
> ```
> pub fn in_edit_mode(ctx: &CommandContext) -> Result<bool> {
>     match operating_mode(ctx)? {
>         OperatingMode::Edit(_) => Ok(true),
>         _ => Ok(false),
>     }
> }
> 
> pub fn assure_edit_mode(ctx: &CommandContext) -> Result<EditModeMetadata> {
>     match operating_mode(ctx)? {
>         OperatingMode::Edit(edit_mode_metadata) => Ok(edit_mode_metadata),
>         _ => bail!("Expected to be in edit mode"),
>     }
> }
> ```
> 
> When the app is in edit mode, it expects a `.git/gitbutler/edit_mode_metadata.toml` file to be written. If we fail to find or read that file, we default the user back to OutsideWorkspace mode
> 
> ```rust
> pub fn operating_mode(ctx: &CommandContext) -> Result<OperatingMode> {
>     let head_ref = ctx.repository().head().context("failed to get head")?;
>     let head_ref_name = head_ref.name().context("failed to get head name")?;
> 
>     if head_ref_name == INTEGRATION_BRANCH_REF {
>         Ok(OperatingMode::InWorkspace)
>     } else if head_ref_name == EDIT_BRANCH_REF {
>         let edit_mode_metadata = read_edit_mode_metadata(ctx);
> 
>         match edit_mode_metadata {
>             Ok(edit_mode_metadata) => Ok(OperatingMode::Edit(edit_mode_metadata)),
>             Err(error) => {
>                 tracing::warn!(
>                     "Failed to open in edit mode, falling back to outside workspace {}",
>                     error
>                 );
>                 Ok(OperatingMode::OutsideWorkspace)
>             }
>         }
>     } else {
>         Ok(OperatingMode::OutsideWorkspace)
>     }
> }
> ```

 if this is a stacked diff which is the parent PR this is stacked to?

krlvi:
Lets not take the full CommandContext when all you need is the repository

crates/gitbutler-operating-modes/src/lib.rs
```diff
@@ -1,38 +1,113 @@
...
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct EditModeMetadata {
+    #[serde(with = "gitbutler_serde::serde::oid")]
+    pub editee_commit_sha: git2::Oid,
+    pub editee_branch: String,
+}
+
+#[derive(PartialEq)]
+pub enum OperatingMode {
```
krlvi, line 44: why not enum her?

crates/gitbutler-operating-modes/src/lib.rs
```diff
@@ -1,38 +1,113 @@
...
+    /// The typical app state when its on the gitbutler/integration branch
+    InWorkspace,
+    /// When the user has chosen to leave the gitbutler/integration branch
+    OutsideWorkspace,
+    /// When the app is off of gitbutler/integration and in edit mode
+    Edit(EditModeMetadata),
```
krlvi, line 50: why do you need a separate type for this?

crates/gitbutler-operating-modes/src/lib.rs
```diff
@@ -1,38 +1,113 @@
+use std::{fs, path::PathBuf};
+
 use anyhow::{bail, Context, Result};
 use gitbutler_command_context::CommandContext;
+use serde::{Deserialize, Serialize};
+
+pub const INTEGRATION_BRANCH_REF: &str = "refs/heads/gitbutler/integration";
+pub const EDIT_BRANCH_REF: &str = "refs/heads/gitbutler/edit";
```
krlvi, line 11: what is this reference for? Also, should this be excluded when listing branches?

crates/gitbutler-operating-modes/src/lib.rs
krlvi, line 58: Can the app be in workspace and in edit mode or outside workspace and in edit mode? Also EditMode metadata is not describing to me what it represents

crates/gitbutler-operating-modes/src/lib.rs
```diff
@@ -1,38 +1,113 @@
...
+        edit_mode_metadata_path(ctx).as_path(),
+        serialized_edit_mode_metadata,
+    )
+    .context("Failed to write edit mode metadata")?;
+
+    Ok(())
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct EditModeMetadata {
```
krlvi, line 43: Some docs can help me understand what this represents

Caleb-T-Owens:
> if this is a stacked diff which is the parent PR this is stacked to?

https://github.com/gitbutlerapp/gitbutler/pull/4640

crates/gitbutler-operating-modes/src/lib.rs
```diff
@@ -1,38 +1,113 @@
...
-/// - `in workspace mode`: When the app is on the gitbutler/integration branch.
-///     This is when normal operations can be performed.
-/// - `outside workspace mode`: When the user has left the gitbutler/integration
-///     branch to perform regular git commands.
+fn edit_mode_metadata_path(ctx: &CommandContext) -> PathBuf {
+    ctx.project().gb_dir().join("edit_mode_metadata.toml")
+}
 
-const INTEGRATION_BRANCH_REF: &str = "refs/heads/gitbutler/integration";
+pub fn read_edit_mode_metadata(ctx: &CommandContext) -> Result<EditModeMetadata> {
```
krlvi, line 18: why is this public?

crates/gitbutler-operating-modes/src/lib.rs
Caleb-T-Owens: You are on `gitbutler/edit` while in edit mode. It should be excluded from listing branches

crates/gitbutler-operating-modes/src/lib.rs
```diff
@@ -1,38 +1,113 @@
...
+    let edit_mode_metadata = fs::read_to_string(edit_mode_metadata_path(ctx).as_path())
+        .context("Failed to read edit mode metadata")?;
 
-pub fn in_open_workspace_mode(ctx: &CommandContext) -> Result<bool> {
+    toml::from_str(&edit_mode_metadata).context("Failed to parse edit mode metadata")
+}
+
+pub fn write_edit_mode_metadata(
```
krlvi, line 26: why is this public?

crates/gitbutler-operating-modes/src/lib.rs
```diff
@@ -1,38 +1,113 @@
...
 pub fn assure_open_workspace_mode(ctx: &CommandContext) -> Result<()> {
     if in_open_workspace_mode(ctx)? {
         Ok(())
     } else {
-        bail!("Unexpected state: cannot perform operation on non-integration branch")
+        bail!("Expected to be in open workspace mode")
+    }
+}
+
+pub fn in_edit_mode(ctx: &CommandContext) -> Result<bool> {
```
krlvi: can't this return just a bool?

crates/gitbutler-operating-modes/src/lib.rs
```diff
@@ -1,38 +1,113 @@
...
+    match operating_mode(ctx)? {
+        OperatingMode::Edit(_) => Ok(true),
+        _ => Ok(false),
+    }
+}
+
+pub fn assure_edit_mode(ctx: &CommandContext) -> Result<EditModeMetadata> {
```
krlvi, line 106: why do you need a helper function that only does a check when you have a `in_edit_mode`? seems like noise

crates/gitbutler-operating-modes/src/lib.rs
```diff
@@ -1,38 +1,113 @@
...
+                );
+                Ok(OperatingMode::OutsideWorkspace)
+            }
+        }
+    } else {
+        Ok(OperatingMode::OutsideWorkspace)
+    }
+}
+
+pub fn in_open_workspace_mode(ctx: &CommandContext) -> Result<bool> {
```
krlvi: overall, if we have operation mode as an enum, what is the purpose of the multiple `is_xxx` functions? Are you optimizing for something that is not clear to me?

crates/gitbutler-operating-modes/src/lib.rs
```diff
@@ -1,38 +1,113 @@
+use std::{fs, path::PathBuf};
+
 use anyhow::{bail, Context, Result};
 use gitbutler_command_context::CommandContext;
+use serde::{Deserialize, Serialize};
+
+pub const INTEGRATION_BRANCH_REF: &str = "refs/heads/gitbutler/integration";
```
krlvi, line 9: why is this defined here? There are other places in the app that use this string, perhaps it should live there

crates/gitbutler-operating-modes/src/lib.rs
```diff
@@ -1,38 +1,113 @@
...
+pub enum OperatingMode {
+    /// The typical app state when its on the gitbutler/integration branch
+    InWorkspace,
+    /// When the user has chosen to leave the gitbutler/integration branch
+    OutsideWorkspace,
+    /// When the app is off of gitbutler/integration and in edit mode
+    Edit(EditModeMetadata),
+}
+
+pub fn operating_mode(ctx: &CommandContext) -> Result<OperatingMode> {
```
krlvi: from a high level perspective, it seems to me that this should be the only entry point (API method). document what it is for

krlvi: How should the user or this API treat the error scenario? E.g when the edit mode file is not found or is invalid. 
Is there a 4th state, e.g. "corrupted"?

crates/gitbutler-operating-modes/src/lib.rs
krlvi: Its not clear what these fields represent. Is the commit sha guaranteed to exist? What is it used for?  How about the branch? Is it just a label of some kind? Do you mean a reference? Why use string here?

krlvi (changes requested):

crates/gitbutler-operating-modes/src/lib.rs
Caleb-T-Owens: No, because its calling `operating_mode` which is fallible

crates/gitbutler-operating-modes/src/lib.rs
Caleb-T-Owens: We've ended up calling its brother assure_open_workspace_mode around 50 times now; its a convenient way to constrain the commands of the application

crates/gitbutler-operating-modes/src/lib.rs
Caleb-T-Owens: Modes are mutually exclusive. Think like vim modes.

You may leave "InWorkspace" and go to "Edit" and back again to the same workspace state, but it would be incorrect to say you are in both modes simultaneously.

crates/gitbutler-operating-modes/src/lib.rs
Caleb-T-Owens: Fair; the `editee_commit_sha` represents the commit that is getting edited in edit mode, and the `editee_branch` is the reference of the virtual branch getting edited.

I can't use a reference because it's lifetime is tied to a git2::Repository (making it an ass pain) and can't be serialized/deserialized.

crates/gitbutler-operating-modes/src/lib.rs
Caleb-T-Owens: I ended up writing the code for working with edit mode in a separate crate as there were dependencies required for that, which didn't make sense for `gitbutler-operating-modes` to contain. We do need at least `read_edit_mode_metadata` to exist in this crate so it can be read when calling `operating_mode`. `write_edit_mode_metadata` could be moved to the future crate and made private.

crates/gitbutler-operating-modes/src/lib.rs
Caleb-T-Owens: See above

crates/gitbutler-operating-modes/src/lib.rs
krlvi: We have a tagged string reference type, maybe you wanna use that?

krlvi (approved):
lgtm

crates/gitbutler-operating-modes/src/lib.rs
Byron: > I can't use a reference because it's lifetime is tied to a git2::Repository (making it an ass pain) and can't be serialized/deserialized.

Right, this is good to know. `gitoxide` has fully owned versions, but I don't know if the serialize (yet). If not, that can easily be added, or own serializers could be added for it (then).

My stance these days is to use the strongest type possible in the backend, possibly converting to a specialized version just for the frontend. Using [`serde(flatten)` worked well for that](https://github.com/Byron/gitbutler/blob/39099ba41cd128b13642c72994096f44089feec7/crates/gitbutler-tauri/src/projects.rs#L97) previously.

Also, generally, I document everything I know about newly added types along with their fields, just because only in that moment I still know most about it.

PR #4644, opened 2024-08-07
Add inline unified word diffs

Caleb-T-Owens:
Introducing inline unified word diffs.

This adds in an experimental feature flag which changes the behavior of word diffs. Rather than showing one line which is the all the removals and additions with the specific words emboldened. With the feature flag enabled, only one line with both the words added and removed is displayed.

Enabled:

Screenshot 2024-08-07 at 18 55 29

Disabled:

Screenshot 2024-08-07 at 18 56 12

This PR introduces a new funciton `computeInlineWordDiff(prevSection: ContentSection, nextSection: ContentSection): Row[]` which is very similar to the old `computeWordDiff` but returns a single section of tokens.

```ts
	function computeInlineWordDiff(prevSection: ContentSection, nextSection: ContentSection): Row[] {
		const numberOfLines = nextSection.lines.length;

		const rows = [];

		// Loop through every line in the section
		// We're only bothered with prev/next sections with equal # of lines changes
		for (let i = 0; i < numberOfLines; i++) {
			const oldLine = prevSection.lines[i] as Line;
			const newLine = nextSection.lines[i] as Line;

			const sectionRow = {
				beforeLineNumber: newLine.beforeLineNumber,
				afterLineNumber: newLine.afterLineNumber,
				tokens: [] as string[],
				type: nextSection.sectionType,
				size: newLine.content.length
			};

			const diff = charDiff(oldLine.content, newLine.content);

			for (const token of diff) {
				const text = token[1];
				const type = token[0];

				if (type === Operation.Equal) {
					sectionRow.tokens.push(...toTokens(text));
				} else if (type === Operation.Insert) {
					sectionRow.tokens.push(`<span data-no-drag class="token-inserted">${text}</span>`);
				} else if (type === Operation.Delete) {
					sectionRow.tokens.push(
						`<span data-no-drag class="token-deleted token-strikethrough">${text}</span>`
					);
				}
			}
			rows.push(sectionRow);
		}

		return rows;
	}
```

We loop over all the sections like before, but look at a feature flag to determine which function to call.

I did simplify the `computeWordDiff` path as it was previously iterating through all previous rows to try and find the previous sections to modify. I've changed it to some simpler indexing operations which should be more performant on larger diffs.

```diff
@@ -186,24 +230,26 @@
 				return acc;
 			}
 
-			const { prevRows, nextRows } = computeWordDiff(prevSection, nextSection);
+			if ($inlineUnifiedDiffs) {
+				const rows = computeInlineWordDiff(prevSection, nextSection);
 
-			// Insert returned row datastructures into the correct place
-			// Find and replace previous rows with tokenized version
-			prevRows.forEach((row) => {
-				const accIndex = acc.findIndex(
-					(accRow) =>
-						accRow.beforeLineNumber === row.beforeLineNumber &&
-						accRow.afterLineNumber === row.afterLineNumber
-				);
-				if (!accIndex) return;
+				acc.splice(-prevSection.lines.length);
 
-				acc[accIndex] = row;
-			});
+				acc.push(...rows);
+				return acc;
+			} else {
+				const { prevRows, nextRows } = computeWordDiff(prevSection, nextSection);
+
+				// Insert returned row datastructures into the correct place
+				// Find and replace previous rows with tokenized version
+				prevRows.forEach((row, previousRowIndex) => {
+					acc[acc.length - (prevRows.length - previousRowIndex)] = row;
+				});
 
-			acc.push(...nextRows);
+				acc.push(...nextRows);
 
-			return acc;
+				return acc;
+			}
```

### DIFF
--- a/apps/desktop/src/lib/config/uiFeatureFlags.ts
+++ b/apps/desktop/src/lib/config/uiFeatureFlags.ts
@@ -10,3 +10,8 @@ export function featureBaseBranchSwitching(): Persisted<boolean> {
 	const key = 'featureBaseBranchSwitching';
 	return persisted(false, key);
 }
+
+export function featureInlineUnifiedDiffs(): Persisted<boolean> {
+	const key = 'inlineUnifiedDiffs';
+	return persisted(false, key);
+}
--- a/apps/desktop/src/lib/hunk/HunkDiff.svelte
+++ b/apps/desktop/src/lib/hunk/HunkDiff.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
 	import { type Row, Operation, type DiffRows } from './types';
+	import { featureInlineUnifiedDiffs } from '$lib/config/uiFeatureFlags';
 	import Scrollbar from '$lib/shared/Scrollbar.svelte';
 	import { create } from '$lib/utils/codeHighlight';
 	import { maybeGetContextStore } from '$lib/utils/context';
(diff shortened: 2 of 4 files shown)


ndom91 (approved):
Nice, LGTM!

PR #4654, opened 2024-08-08
fix: cleanup `@gitbutler/ui` scripts so that `dev:desktop` watch mode works correctly

ndom91:
## ☕️ Reasoning

- Turbo watch mode still didn't update correctly when changes were made in the `@gitbutler/ui` package

## 🧢 Changes

- Merge the `package:styles` and part of the font npm script into 1 call to the `postcss` cli
- That just left the copying of the actual `.woff2` font files
- Removed `rimraf`, didn't seem to be used anywhere anymore

@PavelLaptev @Caleb-T-Owens This seemed to work on my linux box here at home - I could make changes in a UI component, i.e. `packages/ui/src/lib/inputs/Button.svelte` and the UI package scripts would be executed and then the main development app would hot-reload correctly. I'd love to see if it helps the situation on your Mac's as well :crossed_fingers:


ndom91:
@PavelLaptev updated the `dist/styles/fonts/` thing here as well :+1:

PR #4655, opened 2024-08-08
Fix: path to fonts in UI

PavelLaptev:


krlvi (approved):
approving now instead of waiting cause i released this in 0.12.17 without realising the fonts were broken

PR #4658, opened 2024-08-09
Fix complete app breakage due to BString serialization

Byron:
As a follow-up to #4657, this brings back previously reverted improvements and assures that BString will never be serialized as sequence of bytes, something the UI will not be able to interpret as string.

This PR will also put safeties in place to assure an over-eager future me does that again.

### Tasks

* [x] revert previous quick-fix
* [x] assure gix-types with internal BString are serialized correctly
* [x] assure that bare `BString` cannot be mis-interpreted again
* [x] validate that the UI is indeed working

### Notes

Previously I managed to not correctly communicate in code the `BString` serialization 'specialty' (it serializes as array of bytes), and managed to undo it in a refactoring step. Now this case is handled by its own type which would make it very hard for me or others to make the same mistake - now it's clearly communicate why the type exists.

It's also a nice marker that signals that the backend (which probably should be frontend agnostic eventually) is using frontend specific types in its API.


krlvi (approved):
fixes it, thanks

PR #4664, opened 2024-08-09
fix: handle local paths as git remote URIs

ndom91:
## ☕️ Reasoning

- `parseRemoteUrl` errored on local paths as remote URLs

## 🧢 Changes

- Return `undefined` when unable to parse


mtsgrd (approved):

ndom91:
> @ndom91 looks like we need to adjust the vercel deployment? Only files in apps/desktop were touched here.

Yeah I looked into this briefly the other day. It looks like it'll still deploy at least the first commit from every branch since it doesn't have anything to compare it against, based on its default settings. i.e.:

image

In order to determine whether to build or not, Vercel is just running the [turbo-ignore](https://npm.im/turbo-ignore) CLI which leverages the dependency graph it knows about our packages via the turbo config + git info to determine whether things need to be built. Now this is where I'm a bit out of my depth - it does offer a `--fallback` arg which you can pass a ref, do you know a ref we could always pass here to fallback to? `refs/heads/master` maybe?

mtsgrd:
Ah right, let's pass that flag! Thanks for reminding me @ndom91.

PR #4665, opened 2024-08-09
Use `gix` for branch normalization

Byron:
This promises to never create invalidate branch names, while being similar to what Git already does when adding worktrees,
it implements the same rules.

The current implementation uses Regex and creates them on the fly, which takes a lot of time, thus the new implementation promises to shave-off 5% off the runtime in some benchmarks.

### Tasks

* [x] Use latest `main` of `gix`
* [x] Use sanitization when normalizing branches
* [x] check benchmarks to assure it didn't get slower, but hopefully to show it's faster.

### Performance

This image compares `master` with this branch.

Screenshot 2024-08-09 at 21 49 49

The reason for the few regressions I simply don't understand - I wasn't holding my breath while running the benchmark, maybe I was using more resources than I thought I did.

### DIFF
--- a/crates/gitbutler-reference/tests/reference.rs
+++ b/crates/gitbutler-reference/tests/reference.rs
@@ -5,11 +5,12 @@ mod normalize_branch_name {
     fn valid_substitutions() {
         for (input, expected) in [
             ("a", "a"),
-            ("a+b", "a-b"),
+            ("a+b", "a+b"),
             ("a^b", "a-b"),
+            ("a^^^b", "a-b"),
             ("a~b", "a-b"),
-            ("a<b", "a-b"),
-            ("a>b", "a-b"),
+            ("a<b", "a<b"),
+            ("a>b", "a>b"),
             ("a\\b", "a-b"),
             ("a:b", "a-b"),
             ("a*b", "a-b"),
@@ -24,7 +25,11 @@ mod normalize_branch_name {
             ("/-a-/", "a"),
             (".a.", "a"),
         ] {
-            assert_eq!(normalize_branch_name(input).expect("valid"), expected);
+            assert_eq!(
+                normalize_branch_name(input).expect("valid"),
+                expected,
+                "{input} -> {expected}"
+            );
         }
     }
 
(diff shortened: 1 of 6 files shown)


mtsgrd:
We got some user feedback btw, saying a branch name like `Don't crash` caused an error in Jenkins. Would you say we should accommodate for such breakages, or does it make more sense to fix on that side so Jenkis can handle valid branch names?

Byron:
I presume the `'` in `Don't crash` would be the issue there. Personally I don't think GitButler (or Git) would be responsible for making sure branch names work with every possible consumer.

However, it could impose a certain 'taste' and say that `'` doesn't look good in branch names so it should rather end up looking like `Don-t-crash`.

When users notice such an issue, it might be easiest to avoid characters that cause Jenkings trouble.

crates/gitbutler-reference/tests/reference.rs
```diff
@@ -5,11 +5,12 @@ mod normalize_branch_name {
     fn valid_substitutions() {
         for (input, expected) in [
             ("a", "a"),
-            ("a+b", "a-b"),
+            ("a+b", "a+b"),
             ("a^b", "a-b"),
+            ("a^^^b", "a-b"),
             ("a~b", "a-b"),
-            ("a<b", "a-b"),
```
Byron, line 11: It's notable that I enlarged the allowed set of characters to be closer to what Git would allow, but that might be undesired.

crates/gitbutler-reference/tests/reference.rs
```diff
@@ -24,25 +25,27 @@ mod normalize_branch_name {
...
-            "Could not turn \"#[test]\" into a valid reference name"
+            "Could not turn \"-\" into a valid reference name",
+            "show the original value, not the processed one to be familiar to the user"
         );
     }
 
     #[test]
     fn complex_valid() -> anyhow::Result<()> {
         assert_eq!(normalize_branch_name("feature/branch")?, "feature/branch");
+        assert_eq!(normalize_branch_name("#[test]")?, "#-test]");
```
Byron, line 48: This is a good example for how it will now handle all inputs, just like Git.

krlvi (approved):

PR #4669, opened 2024-08-10
feat(CloneForm): add error tracking and improve typing

Caleb-T-Owens:
Recently we added the ability to clone within HTTP and have been tracking the percentage of users who successfully cloned, and those who try to clone with SSH. This adds in another metric so we can see how many people try to clone with HTTP but then hit a failure.

```diff
@@ -87,6 +88,8 @@
 			posthog.capture('Repository Cloned', { protocol: remoteUrl.protocol });
 			await projectService.addProject(targetDir);
 		} catch (e) {
+			Sentry.captureException(e);
+			posthog.capture('Repository Clone Failure', { error: String(e) });
 			errors.push({
 				label: String(e)
 			});
```

### DIFF
--- a/apps/desktop/src/lib/onboarding/CloneForm.svelte
+++ b/apps/desktop/src/lib/onboarding/CloneForm.svelte
@@ -9,6 +9,7 @@
 	import { parseRemoteUrl } from '$lib/url/gitUrl';
 	import { getContext } from '$lib/utils/context';
 	import Button from '@gitbutler/ui/inputs/Button.svelte';
+	import * as Sentry from '@sentry/sveltekit';
 	import { open } from '@tauri-apps/api/dialog';
 	import { documentDir } from '@tauri-apps/api/path';
 	import { join } from '@tauri-apps/api/path';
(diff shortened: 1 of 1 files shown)


krlvi (approved):

apps/desktop/src/lib/onboarding/CloneForm.svelte
```diff
@@ -87,6 +88,8 @@
 			posthog.capture('Repository Cloned', { protocol: remoteUrl.protocol });
 			await projectService.addProject(targetDir);
 		} catch (e) {
+			Sentry.captureException(e);
+			posthog.capture('Repository Clone Failure', { error: String(e) });
```
Byron, line 92: Just as a note related to [possible future changes to the cloning-backend](https://github.com/gitbutlerapp/gitbutler/commit/cda04e9b5b869edb23fd8c73d7fa03287bed0ad3#r145247361), I wanted to mention that `gitoxide` will provide more detailed errors which may contain the URL of the repository. It will naturally [strip the password in URLs](https://github.com/Byron/gitoxide/blob/f8ce3d0721b6a53713a9392f2451874f520bc44c/gix-url/src/impls.rs#L79-L91), but sometimes URLs themselves contain sensitive information.

Thus, also as a note to myself, when the backend changes there should probably be some redaction implemented here that strips URLs in their entirety.

PR #4670, opened 2024-08-10
branch details with `gix`: 2.6x faster!

Byron:
This PR is the first step towards providing  branch details with `gitoxide`.

### Tasks

* [x] `git2` performance optimizations for 'apples-to-apples' comparison.
* [x] Fill in missing details in understanding
* [x] Add a couple of tests (validate merge-base and diff information)
* [x] Add benchmarks - ideally also with larger repositories
* [x] Use `gix` for commit-walk - more than twice as fast.
       Screenshot 2024-08-12 at 11 56 56, and it can certainly get even faster.
* [x] Use `gix` for tree-tree diff Screenshot 2024-08-13 at 10 24 34 - first results are promising, **at 11 times faster** on a single core, pure diffing performance. It will be slower if the diffed objects have to be retrieved/aren't cached, so in reality on `gitlab` it's just about ~~1.6x~~ **1,85x** of `git2` (thanks to moderate multi-threading).
* [x] make it fast in the real world
* [x] improve tree-tree diff API
* [x] don't forget `origin/478475-model-registry-markdown-support-for-version-description` has 2 commit (itself) above the merge-base, but lists ` commit in the `gix`
    - This is correct, actually, as there is only one commit and `git2` for some reason counts the first parent, even though it's 'underneath' the mergebase.
* [x] Don't forget to re-enable all benchmarks!
* [x] the last performance improvement
     - **2.6x** improvement overall!
     - With that, even once `gix` is faster at merge-bases, it will probably only make a difference in memory consumption, but not in performance.
* [x] Add support for `CC` based `zlibng` to `flate2`

### Follow-Ups

* #4744 
* #4757 
* #4765

### Notes for the Reviewer

* This PR also optimizes `git2` for performance, which also levels the playing field as certain validations cost time, but aren't performed (yet) by `gix`.
* `gix` is able to be 1.6x of `git2` when it comes to tree-diffing, and it's noticeable. however, for now it's more like 1.5x as stock zlib has to be used until `cmake` is either available or `flate2` can be configured to not use `cmake` at all.

### Notes

* When having a virtual branch named `main` it becomes 'entangled' with `refs/heads/main`. However, when `refs/heads/main` receives a new commit, `main` (the vbranch) knows nothing about it and also won't offer to rebase on top of it. It's still tied to `origin/main`. It's probably OK, even though I would have expected to be able to incorporate the changes somehow.
* Eventually, we will want to make `max-performance` work on CI - shouldn't be too hard and would already work if docker wasn't used.
* I have added sorting (lexicographically, increasing) to make both versions more comparable, and also because `git branch` sorts like that (but using the full ref name. In the end, the UI should probably sort and group like it wants, then the backend doesn't have to do that anymore. Then I realized the UI is already sorting by last-modified date.

<details>

* For clean profiling, use `gitbutler-cli branch details 478475-model-registry-markdown-support-for-version-description "pedropombeiro/454313/3-deprecate-fields" "mc/overwrite-jobs-instead-of-uniqueness-enforcement" "backfill-multiple-desired-sharding-key-small-table-compliance_framework_security_policies" "jay_mccure-master-patch-68663" "work-items-widgets-migration" "ah-dedup-records-migration" "455903-limits" "477889-implement-additional-telemetry-for-added-context-code-completion" "psi-ff-summduo" "450705-policies-mvp-combined-mode" "ia-arkose-token-verification-consitent-return" "hustewart-add-build-snippet-service" "bw-add-subscribed-filter" "adjust-new-user-signup-caps-validation" "bk/473748-backfill-owasp-top-10-null" "cwoolley-gitlab-master-patch-150b" "application-settings-pattern-poc-haml" "470051-document-backup-of-object-storage-data" "j_lar/add_redis_keys_to_monitor" "477306-glql-table-of-issues" "rlehmann1-fix-lint-issues-policies" "ban-ai-duo-settings-admin-group-form" "handle-empty-ff-merge-in-from-train-ref-strategy" "bw-run-markdown-benchmark-on-tag" "477813-follow-up-from-refactor-ai-ga-policies" "gitaly-ci-jobs-52550" "443369-update-self-managed-billing" "433082-iteration-rest-api-does-not-properly-filter-out-iterations-from-groups-with-no-access" "jkhabie-add-reachability-column-sbom" "jreporter-master-patch-d091" "vmairet-add-token-associations" "vij-remove-member-default-value" "autoflow/refactor-naming" "thutterer-danger-demo-settings" "443369-update-billed-shared-group-users" "fix_audit_event_flow" "16-11-stable-ee" "477712-health-check-repeated-click" "470701-branch-rule-editing-minimum-required-approvals-per-target-branch" "ensure_prepared_worker_automation" "janis-add-pages-to-usage-quotas" "quarantine-flaky-tests-spec-lib-gitlab-import_export-project-relation_factory_spec-rb-81" "440857-glql-integration" "472352-transition-dast-site-profiles-builds" "tachyons-pat-sharding-key-full" "368542-container-registry-metadata" "design-management-widget-migration" "ag/477091-seat-controls-fe-validation" "458831-remove-required_instance_ci_template-column" "455903-placeholder-limit" "gmh-add-approver-to-mr-graphql" "quarantine-flaky-tests-spec-lib-banzai-pipeline-plain_markdown_pipeline_spec-rb-106" "463258-create-common-solution-on-how-to-count-and-track-feature-flag-enabled-value" "mk/fix-empty-replication-details-view" "rodrigo/integrate-improved-user-mapping-in-direct-transfer" "16-9-stable-ee" "jmc-16.9-backport-shm-qa" "mattkasa/refactor-query-analyzers" "application-settings-pattern-poc" "464587-support-merge-requests-as-context-for-duo-chat-responses" "449139-fe-update-the-list-and-the-counter-after-inline-role-change-2" "jivanvl-update-permissions-container-expiration-policy" "hustewart-snippets-refactor" "477547_nullify_organization_id_migration" "slashmanov/add-release-filter-to-mr-list-app" "dagron1-vulnerability-resolution-supported-scanners" "464132-work-items-prevent-adding-more-than-10-items-to-hierarchy-widget-in-one-batch" "md-add-feature-flag-headers" "form-issue-jira-setting" "ph/mrDashboardNavCount" "mc_rocha-poc-ingest-license-sbom-security-policy""462398-refactor-pb-fe" "bk/473748-modify-finder" "471726-ignore-invalid-project-CI-with-pep-override-strategy" "autoflow/issue-created-event" "466718-add-api-endpoint-for-bulk-usage_data-sending" "vi-display-closed-status-for-branches" "docs/bprescott/20240618-dbmisc" "eduardosanz-master-patch-19280" "461761-refactor-clean-widgets-to-only-use-id-iid-and-fullpath-to-fetch-work-item-in-individual" "andrey-move-gdk-build" "reintroduce-create-jira-issue-form" "seed-placeholder-users-task" "ck3g-test-xray-job-do-not-merge" "463822-create-service-to-generate-metadata-for-group-instance-level"`

stderr with `cache-efficiency-debug` enabled, showing decent object cache performance, but bad pack-cache performance. The object cache is the most important.

</details>


crates/gitbutler-branch-actions/src/branch.rs
```diff
@@ -425,58 +425,65 @@ pub fn get_branch_listing_details(
...
+        (
+            default_branch.get().peel_to_commit()?.id(),
+            default_target.sha,
+        )
+    };
 
+    let mut enriched_branches = Vec::new();
     for branch in branches {
-        let merge_base_comparison = if let Some(virtual_branch) = branch.virtual_branch {
+        let other_branch_commit_id = if let Some(virtual_branch) = branch.virtual_branch {
```
Byron: Could someone help me understand this logic?

I think what I fail to grasp is why applied virtual branches merge with and compare to what *probably* comes down to `refs/heads/main`, whereas unapplied branches compare to *probably* `refs/remotes/origin/main`.

crates/gitbutler-branch-actions/src/branch.rs
```diff
@@ -425,58 +425,65 @@ pub fn get_branch_listing_details(
...
+
+    let (default_target_upstream_commit_id, default_target_commit_id) = {
+        let default_target = ctx
+            .project()
+            .virtual_branches()
+            .get_default_target()
+            .context("failed to get default target")?;
+        let default_local_branch =
+            repo.find_branch(default_target.branch.branch(), git2::BranchType::Local)?;
+        let default_branch = default_local_branch.upstream()?;
```
Byron: This would probably return the local tracking branch, `refs/remotes/origin/main`, right?

Byron (commented):
@krlvi Maybe you could help answer a couple of probably minor questions - they will probably resolves themselves as well once tests are added.

crates/gitbutler-branch-actions/src/branch.rs
krlvi: I think there is a difference in intent that is poorly expressed. 
When we are previewing branches made by others, a user would generally expect the diff to be the same diff as what a github pull request would show, which is the state of the origin branch.
On the other hand when working locally, you may not have updated the workspace base yet (because you are not ready for it). In that context it may make sense to compare to that local main. 
Let me know if this explanation doesnt quite make sense still

crates/gitbutler-branch-actions/src/branch.rs
krlvi: i think this should give you `refs/heads/main`

crates/gitbutler-branch-actions/src/branch.rs
Byron: Thank you, this does make sense but only if I think in terms of local branch and local tracking branch.

And I think once there are some tests it will quickly become clear what is what, and maybe then I can add the documentation that I would need to remember.

crates/gitbutler-branch-actions/src/branch.rs
Byron: That is a bit unexpected as I thought `default_target.branch.branch()` is `main` and `upstream()` should then be something else - the only upstream branch I can think of is the local tracking branch. But once there is a test, I think this should become clear.
We are probably looking at the difference between diffing against what was last fetched (with the local tracking branch `refs/remotes/origin/main` updated), and the local state that we would rebase against, `refs/heads/main`.

Byron: `upstream()` does return the local tracking branch, e.g. typically `refs/remotes/origin/main`. And interestingly, the `sha` field seems to be the merge-base between the local tracking branch and HEAD at the time the target is computed. 
This strikes me like an error source as if that drifts, but maybe that's not a problem as long as it recomputes it when switching to the integration branch.

krlvi: > i think this should give you refs/heads/main

Okay i am actually not sure of my own answer. I will take a look tomorrow to see i can give some more helpful feedback.

crates/gitbutler-branch-actions/tests/virtual_branches/list_details.rs
```diff
@@ -0,0 +1,113 @@
...
+        }
+    );
+    assert_eq!(
+        list[2],
+        BranchListingDetails {
+            name: "non-virtual-feature".into(),
+            lines_added: 55,
+            lines_removed: 0,
+            number_of_files: 1,
+            number_of_commits: 50 + 5,
```
Byron: This is essentially a merge-base test and it's a bit unexpected, as I was trying to make the merge-base logic observable. Sometimes, depending on order of operations, it creates virtual branches from git branches implicitly, even though here I managed to avoid that.

It's all about the local tracking branch `origin/main`, which is unchanged. I would have thought it's possible to make it diff against the local `main` though, just to see the logic in action - so far it always diffs against `origin/main`.

crates/gitbutler-branch-actions/src/branch.rs
Byron: By now I have made sense of it all, and the actual `upstream()` (i.e. local tracking branch) is used for virtual branches that aren't in the workspace, or other branches. Otherwise, the last-seen (and stored) location of the local tracking branch is used, as applied virtual branches are always assumed to be applied on top of that local tracking branch state.

Running `update` on it would then re-apply them on top of the latest location of the local tracking branch.

Byron:
> the tree to tree diffing being 11 times faster is really impressive! looking forward to using that in other hot paths of the app :)

Unfortunately it's really only like that in the given microbenchmark, which just diffs a single-line changes of the same two blobs, 10k times. As `gitoxide` has a cache for this, it doesn't do much work after extracting both objects, whereas `git2` doesn't seem to have this optimization (an optimization really coming from implementing rename-tracking which is disabled here).

In practice, on a real GitLab repo it still clocks in at 1.6x faster, which is something I can feel when using the UI.

krlvi (approved):
awesome! cant wait to try it

PR #4671, opened 2024-08-10
fix: clicking 'Cancel' on new project page does nothing

### LINKED ISSUE #2539: Can't back out of adding new project
On macOS, after clicking “Add new project” and selecting a directory, you are prompted to select a target branch. There is a “Back” button next to “Let's go!”, but it does not work. Even after restarting the app, I can't leave the screen.

Resetting the app with `rm -rf ~/Library/Application\ Support/com.gitbutler.app` works.

Comnir:
## ☕️ Reasoning
- Clicking 'Add new project' immediately adds the new repo to the projects list
- Clicking 'Cancel' on the project setup page redirects to the app root page
- If the projects list contains a project, the root page redirects to the project page. When the project setup is incomplete, the project setup is displayed

## 🧢 Changes
- To avoid this redirect, clicking cancel now deletes the project
- To keep the `BackButton` generic, added a callback (currently named `beforeOnMouseDown`) to be executed before the generic implementation of back.

## 🎫 Affected issues
Fixes: https://github.com/gitbutlerapp/gitbutler/issues/2539

## 📌 Todos
- Find a better name for `beforeOnMouseDown` or replace with a different way to execute the deletion
- Is this the correct way to delete a project? In some place `reload` is called only if `deleteProject` didn't throw, but in other places it's called anyway.


apps/desktop/src/lib/components/BackButton.svelte
```diff
@@ -1,17 +1,30 @@
...
+					history.back();
+				} else {
+					goto('/');
+				}
+			},
+			(err) => {
+				console.log('Failed to execute the pre-mouse-down action');
+				console.log(err);
+			}
+		);
```
mtsgrd, line 27: What is the intent with introducing a "tick" here?

apps/desktop/src/lib/components/ProjectSetupTarget.svelte
```diff
@@ -76,6 +76,16 @@
...
+
+	const projectService = getContext(ProjectService);
+	async function deleteProject() {
+		try {
+			await projectService.deleteProject(project.id);
+			await projectService.reload();
+		} catch (err: any) {
+			console.error(err);
+		}
+	}
```
mtsgrd: Could we perhaps drop the try/catch here, and rely on the uncaught exception handler to display a toast and log the error? Since it's only two lines I would also prefer if we inline click handler.

apps/desktop/src/lib/components/ProjectSetupTarget.svelte
```diff
@@ -248,7 +258,7 @@
 		</SetupFeature>
 	</div>
 	<div class="action-buttons">
-		<BackButton>Cancel</BackButton>
+		<BackButton preMouseDown={deleteProject}>Cancel</BackButton>
```
mtsgrd: Do you think we could just introduce a `<Button>Cancel</Button>` here instead of using the `<BackButton>`? That way we don't need to introduce the `preMouseDown` handler.

Comnir: I'm not entirely sure what you mean...

Do you mean to use the generic `Button` and set `onmousedown` to delete the project and add the existing implementation of `BackButton`'s `onmousedown`?
Or do you mean a to create a `CancelButton` with the delete project behavior?

mtsgrd: Yeah sorry, I mean a generic button with `onmousedown` to delete the project!

Comnir: I replaced `BackButton` with `Button`. Is the inline-ing still relevant?

mtsgrd:
Thanks, this looks good. Would you mind fixing those lint errors and I'll merge this pr?

Comnir:
Sure, sorry!

mtsgrd (approved):

apps/desktop/src/lib/components/ProjectSetupTarget.svelte
mtsgrd: No need to inline, looks good now.

Comnir:
@mtsgrd is there any other action required from side before you merge?

mtsgrd:
Sorry for the delay, no not at all. Let me rebase it and merge!

PR #4674, opened 2024-08-11
Have selected filter be specific to a project

Caleb-T-Owens:
Previously it was scoped globally but that doesn't make sense as some projects have the PR tab and others don't.

I've also made it correctly default to index 0 if the stored value in LS is not found.


krlvi (approved):
lgtm

PR #4682, opened 2024-08-13
Provide current mode to frontend

Caleb-T-Owens:
## Context

This PR is part of a larger series of PRs.

- [x] Introduce Modes
- [x] Add edit mode
- Provide modes to the frontend (this PR)
- [ ] Switching to and from edit mode
- [ ] Writing and reading conflicted commits
- [ ] Rebase conflicted commits

---

I've modified `headService` to now be the `modeService` which now provides both the mode and current head, as the two are still closely related. With time we might want to separate them, but at the minute one infers the other, so it seams reasonable.


krlvi (approved):

apps/desktop/src/lib/modes/service.ts
```diff
@@ -0,0 +1,36 @@
...
+export class ModeService {
+	readonly head = writable<string | undefined>(undefined);
+	readonly mode = writable<Mode | undefined>(undefined);
+
+	readonly gbBranchActive = derived(this.head, (head) => head === 'gitbutler/integration');
+
+	unsubscribe?: () => Promise<void>;
+
+	constructor(private projectId: string) {
+		this.unsubscribe = subscribeToHead(projectId, (head, mode) => {
```
mtsgrd, line 15: Sorry, got in a bit late here. Could we subscribe/unsubscribe using the store like we do in e.g. headService.ts?

PR #4684, opened 2024-08-13
UI lib: text utility classes

PavelLaptev:
Simplify text classes:
- Shorter names
- Replace full text class `class="text-body-size"`. Now they are two classes — `class="text-size text-body"`


mtsgrd (approved):
Awesome.

PR #4692, opened 2024-08-15
faster E2E runs

Byron:
For E2E, there isn't much data processing happening which makes release builds less useful. Instead, build in debug mode **to cut the compile time from 4m to 2.5m**, and see the e2e test start faster, and complete sooner.

It's notable that it's still much slower than it has any right to be, probably because the `--debug` of a tauri-build boils down to slightly different settings that prevent it from reusing the caches that have been created when building `gitbutler-cli`.

### Tasks

* [x] make the change
* [ ] figure out if the test is now flaky as the backend is too slow?

Seem to not work, and the saving isn't great anyway.

### DIFF
--- a/apps/desktop/wdio.conf.ts
+++ b/apps/desktop/wdio.conf.ts
@@ -15,7 +15,7 @@ export const config: Options.WebdriverIO = {
 			// @ts-expect-error custom tauri capabilities
 			maxInstances: 1,
 			'tauri:options': {
-				application: '../../target/release/git-butler-dev'
+				application: '../../target/debug/git-butler-dev'
 			}
 		}
 	],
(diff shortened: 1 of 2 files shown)


ndom91:
Nice! :partying_face:

ndom91:
Edit: Just saw that it didnt work out :( haha

mtsgrd:
Looks to me like we'd just need to build `gitbutler-git` as well? Everything else looks normal to me in that test run.

ndom91:
I'm going to spend a couple minutes following this up today because if we can cut the build time from 4m -> 2.5m for every test run (i.e. every push atm), that'd be pretty big actually imo

PR #4695, opened 2024-08-15
Add edit mode actions

Caleb-T-Owens:
Introduces some basic actions for entering and leaving edit mode. Its all hidden behind a feature flag in the experimental settings, with a warning not to use it :D.

The key logic is introduced here: https://github.com/gitbutlerapp/gitbutler/blob/4a9114f4f3a02136d12c7d13cd6fb631ddc5dca1/crates/gitbutler-edit-mode/src/lib.rs

### DIFF
--- a/apps/desktop/src/lib/modes/service.ts
+++ b/apps/desktop/src/lib/modes/service.ts
@@ -2,8 +2,8 @@ import { invoke, listen } from '$lib/backend/ipc';
 import { derived, writable } from 'svelte/store';
 
 export interface EditModeMetadata {
-	editeeCommitSha: string;
-	editeeBranch: string;
+	commitOid: string;
+	branchReference: string;
 }
 
 type Mode =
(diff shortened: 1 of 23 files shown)


apps/desktop/src/lib/modes/service.ts
```diff
@@ -29,6 +40,20 @@ export class ModeService {
 
 		this.headAndMode.set({ head, operatingMode });
 	}
+
+	async enterEditMode(editeeCommitId: string, editeeBranchRef: string) {
+		await invoke('enter_edit_mode', {
+			projectId: this.projectId,
+			editee: editeeCommitId,
+			editeeBranch: editeeBranchRef
```
mtsgrd: At first glance I react to this variable name, I don't think it's a real word. What exactly does _editee_ refer to?

apps/desktop/src/routes/[projectId]/+layout.svelte
```diff
@@ -160,6 +162,8 @@
 	onDestroy(() => {
 		clearFetchInterval();
 	});
+
+	$inspect($mode);
```
Comment (mtsgrd) [inline apps/desktop/src/routes/[projectId]/+layout.svelte]: 👀

apps/desktop/src/lib/modes/service.ts
Caleb-T-Owens: Screenshot 2024-08-19 at 12 01 54


I was calling the commit that was being edited the editee. I guess
```rs
struct EditMode {
  subject_commit: Oid,
  subject_branch: String
}
```
works. WDYT?

mtsgrd: Yeah I think that's better, but let me still ask, is a prefix necessary?
```rs
struct EditMode {
  commitId: Oid,
  branchId: String
}
```

krlvi (approved):
lgtm

PR #4700, opened 2024-08-15
Include packages in gh actions node filter

mtsgrd:


Caleb-T-Owens (approved):
🤔 Should it be `apps/**` too given that we presumably will write tests for apps/web as well?

mtsgrd:
Let's do that, but also, what would it take to not have to run web tests when changing desktop files?

PR #4703, opened 2024-08-16
fix: tooltips when used in elevated contexts

ndom91:
## Changes

- Tooltips used in elevated contexts, like contextMenu's would be rendered behind that context. With `calc(infinity)` nothing should be able to be on top of it, other than the newer `top-layer` stuff

### Before

image

### After

image


PavelLaptev (approved):
I updated z-index variables

PR #4707, opened 2024-08-16
Fix branch re-ordering

mtsgrd:
- lets svelte re-order visually rather than manipulating the dom

### DIFF
--- a/apps/desktop/src/lib/components/Board.svelte
+++ b/apps/desktop/src/lib/components/Board.svelte
@@ -10,10 +10,12 @@
 	import { getGitHost } from '$lib/gitHost/interface/gitHost';
 	import { persisted } from '$lib/persisted/persisted';
 	import { getContext, getContextStore } from '$lib/utils/context';
+	import { throttle } from '$lib/utils/misc';
 	import { BranchController } from '$lib/vbranches/branchController';
 	import { VirtualBranchService } from '$lib/vbranches/virtualBranch';
 	import Icon from '@gitbutler/ui/Icon.svelte';
 	import { open } from '@tauri-apps/api/shell';
+	import { flip } from 'svelte/animate';
 
 	const vbranchService = getContext(VirtualBranchService);
 	const branchController = getContext(BranchController);
@@ -26,8 +28,6 @@
 
 	let dragged: HTMLDivElement | undefined;
 	let dropZone: HTMLDivElement;
-	let priorPosition = 0;
-	let dropPosition = 0;
 
 	let dragHandle: any;
 	let clone: any;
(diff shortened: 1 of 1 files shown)


apps/desktop/src/lib/components/Board.svelte
```diff
@@ -46,48 +45,49 @@
...
-					const updates = $branches.map((b, i) => {
-						return { id: b.id, order: i };
-					});
-					branchController.updateBranchOrder(updates);
+
+				// Update sorted branch array manually.
+				if (currentPosition !== dropPosition) {
+					const el = sortedBranches.splice(currentPosition, 1);
+					sortedBranches.splice(dropPosition, 0, ...el);
+					sortedBranches = sortedBranches; // Redraws #each loop.
```
Caleb-T-Owens: I'm a little surprised that this works, and the prior line doesn't.

In vue3, you'd have to say something like `sortedBranches = [...sortedBranches]` because it would only react to a change in object reference. Svelte however seems to behave much better with methods like `push` which is in part what surprises me about `splice` not working.

Caleb-T-Owens: Looking it up. I seem to be wrong, and what you did there is reasonable svelte code.

mtsgrd: A'ight. I but also, I put this PR back to draft. I want to fix one more thing, a once and for all, we've had too many issues with branch re-ordering.

mtsgrd:
@PavelLaptev I added `animate:flip` to the branches, I know you've been wanting something like that. There's some problem though, the animation looks great when dragging left, but there's something jittery about dragging right. Any chance you know how to fix that?

https://github.com/user-attachments/assets/2c708fe8-8ad8-45e4-bdc4-32ba145bb330

PavelLaptev:
Nice! Thanks @mtsgrd 
I'll check this out today

PavelLaptev:
Trying to debug the thing. So far, I noticed that the flickering bug in this function: 

```js
const handleDragOver = throttle((e: MouseEvent & { currentTarget: HTMLDivElement }) => {
  // code in between
}, 50);
```

I increased `50` to `1000` and also increase the `flip` animation duration and when I move the lane to the right, it first goes to the very end and then goes to the right position.

https://github.com/user-attachments/assets/d98170cd-fc57-455a-8eac-01d21eb07d6d

https://github.com/user-attachments/assets/3b288cb6-6370-474b-b051-fc34672a3caf

If I remove the `throttle`, flickering but not always.

https://github.com/user-attachments/assets/e0617cb1-09d4-4f1b-b470-eb2351f61178

PavelLaptev:
Just tried this library [SVELTE DND ACTION](https://github.com/isaacHagoel/svelte-dnd-action?tab=readme-ov-file) for reordering. [REPL](https://svelte.dev/repl/b66ec221150444cda0f5bc59a2ca4e23?version=4.2.17)

It works, but I'm not sure that we want to add this library because in other places we're still using our custom methods.

https://github.com/user-attachments/assets/9fe01339-d778-4bab-8a97-6496d565c714

mtsgrd:
Thanks @PavelLaptev, let me have another go at this, I have some clues based on what you found above. About `svelte-dnd-action`, it's actually the library we started out with, but we came to the conclusion it was making things a bit harder to understand than just doing vanilla js drag & drop.

mtsgrd:
I think I sorted it out now, the drop position depends on whether you are dragging left or right.


https://github.com/user-attachments/assets/dea66c28-1757-4210-b095-2b761160404b

Issue #2651, opened 2024-02-14
🚀 Feature: Support Azure DevOps 


cftad:
Like the existing GitHub integration, support for viewing and raising PRs for Azure DevOps would be great. 😄 

https://azure.microsoft.com/en-gb/products/devops/repos/

cddigi:
So far, I have not had any issue using DevOps with GitButler for Windows, I was able to just add my exiting repo to GB and it worked. It picks up the RSA key on my machine that I've added to my DevOps profile and is able to create and branch and push to remote. However, when trying to open commits that were pushed to Azure DevOps, it is using the wrong URL format. 

What GitButler produces:
`https://ssh.dev.azure.com/v3/<organization>/<project-name>/<repo-name>/commit/<hash>`

The actual link to the commit:
`https://dev.azure.com/<organization>/<project-name>/_git/<repo-name>/commit/<hash>`

This is the repo that DevOps has me use for SSH:
`git@ssh.dev.azure.com:v3/<organization>/<project-name>/<repo-name>`

DylanSnel:
@cddigi How do you create a PR for devops with it? Am i missing a button?

cddigi:
I never got a button to create the PR from within GitButler like I would for GitHub. It only offers a button to view the commit on DevOps, it's that button which is giving me the invalid URL.

Caleb-T-Owens:
Marked as good first issue, git host support can be added incrementally; for example, listing pull requests can be added without continuous integration reporting. Feel free to do piece by piece

Byron:
As a `gitoxide` related note, once `gitoxide` performs clones for added performance with big repositories, it would need [this issue](https://github.com/Byron/gitoxide/issues/1025) to be fixed, or GB needs to special case this and use `git` while it's not fixed.

PR #4708, opened 2024-08-16
Use gix for cloning repositories

Caleb-T-Owens:
Looking through my emails this morning I saw this comment from @Byron on cloning azure devops https://github.com/gitbutlerapp/gitbutler/issues/2651#issuecomment-2293123466 (source gix issue: https://github.com/Byron/gitoxide/issues/1025) repositories not working through gix. It however made me think: Currently in GitButler, both cloning via SSH URLs and cloning private repositories doesn't work. It also got me wondering if switching to gix, despite the aforementioned issue, could be a quick and easy win that would resolve a large majority of the issues with the current setup. After trying it out, it works pretty much perfectly.

```diff
 
     #[tauri::command(async)]
     pub fn git_clone_repository(repository_url: &str, target_dir: &Path) -> Result<(), Error> {
-        git2::Repository::clone(repository_url, target_dir).context("Cloning failed")?;
+        let url =
+            gix::url::parse(repository_url.into()).context("Failed to parse repository URL")?;
+        let should_interrupt = AtomicBool::new(false);
+        let mut prepared_clone =
+            gix::prepare_clone(url, target_dir).context("Failed to prepare clone")?;
+        let (mut prepared_checkout, _) = prepared_clone
+            .fetch_then_checkout(Discard, &should_interrupt)
+            .context("Failed to fetch")?;
+        let should_interrupt = AtomicBool::new(false);
+        prepared_checkout
+            .main_worktree(Discard, &should_interrupt)
+            .context("Failed to checkout main worktree")?;
         Ok(())
     }
 }
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2588,6 +2588,7 @@ dependencies = [
  "gix-path 0.10.9 (git+https://github.com/Byron/gitoxide?rev=d51f330e9d364c6f7b068116b59bf5c0160e47af)",
  "gix-pathspec",
  "gix-prompt",
+ "gix-protocol",
  "gix-ref 0.45.0",
  "gix-refspec",
  "gix-revision",
@@ -2601,6 +2602,7 @@ dependencies = [
  "gix-utils 0.1.12 (git+https://github.com/Byron/gitoxide?rev=d51f330e9d364c6f7b068116b59bf5c0160e47af)",
  "gix-validate 0.8.5 (git+https://github.com/Byron/gitoxide?rev=d51f330e9d364c6f7b068116b59bf5c0160e47af)",
  "gix-worktree 0.34.1 (git+https://github.com/Byron/gitoxide?rev=d51f330e9d364c6f7b068116b59bf5c0160e47af)",
+ "gix-worktree-state",
  "once_cell",
  "smallvec",
  "thiserror",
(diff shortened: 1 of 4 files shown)


Byron:
> After trying it out, it works pretty much perfectly.

It sure lures you into thinking that 😅. However, you actually have to configure something to be sure it truly works, particularly on Windows. The variable of interest [here is `git_installation`](https://github.com/Byron/gitoxide/blob/main/src/plumbing/main.rs#L98) which should be `true`, that way it will obtain git-installation specific configuration that usually contains all the credential helper invocations, particularly on Windows. As this has to invoke the Git binary, it can be slow (on Windows, at least), so it's not done by default, also because of the security implications.

To get clone to work, it must be done like this though.

In future, `should_interrupt` should probably be hooked up as well to be turned on if the user tries to abort the operation.

That should be all the gotchas I can think of right now.

Caleb-T-Owens:
> It sure lures you into thinking that 😅. However, you actually have to configure something to be sure it truly works, particularly on Windows. The variable of interest [here is git_installation](https://github.com/Byron/gitoxide/blob/main/src/plumbing/main.rs#L98) which should be true, that way it will obtain git-installation specific configuration that usually contains all the credential helper invocations, particularly on Windows. As this has to invoke the Git binary, it can be slow (on Windows, at least), so it's not done by default, also because of the security implications.

@Byron Could you show me how to provide that argument? I'm finding the gix documentation quite hard to dig through.

Byron:
Actually, I was wrong, and since cloning *always* needs the git binary installation configuration, it will automatically use it.
No change is actually needed.

Also sorry for confusing opening a repository with cloning it - the APIs are very different and it's not obvious how to affect the open-options (but it's possible with `gix::clone::PrepareFetch(…)` should it be needed at some point).

Maybe the backend could check for azure-devops HTTPS URLs and use `git2` for these - this would be an example URL: `https://dev.azure.com/bittrance/public/_git/gitoxide`, and one whose clone should currently fail with `gix`.

krlvi (approved):
lgtm

PR #4709, opened 2024-08-16
Fix branch name deduplication

mtsgrd:
- do not include name of branch being updated when deduping


Caleb-T-Owens (approved):
I'm suprised that this hasn't been a huge problem

PR #4710, opened 2024-08-16
chore: migrate BranchHeader files to runes

ndom91:


apps/desktop/src/lib/branch/BranchLaneContextMenu.svelte
```diff
@@ -15,10 +15,14 @@
 	import Button from '@gitbutler/ui/inputs/Button.svelte';
 	import Modal from '@gitbutler/ui/modal/Modal.svelte';
 
-	export let contextMenuEl: ContextMenu;
-	export let target: HTMLElement;
-	export let onCollapse: () => void;
-	export let onGenerateBranchName: () => void;
+	interface Props {
+		contextMenuEl?: ContextMenu;
```
mtsgrd, line 19: Since the component is only used in one place, do we need to make this optional?

mtsgrd (approved):

apps/desktop/src/lib/branch/BranchLaneContextMenu.svelte
ndom91: Yeah I think we do unfortunately because the `contextMenu` variable from the parent (`BranchHeader`) is `undefined` initially and only gets bound once the child (`ContextMenu`) of the `BranchLaneContextMenu` is rendered / mounted, right?

PR #4712, opened 2024-08-16
chore: migrate BranchLabel to runes

ndom91:
- Migrate `BranchLane.svelte` to runes
- Replace `ResizeObserver` method with simple `{number}ch` resizing 
  - since we're not using a monospace font, the `ch` sizing strays a _little_ bit once you get to larger sizes, i.e. the input is a bit bigger than the characters in it since it'll usually have some thin characters. So no worries if that trade-off for simplicity (no resizeobserver) isn't something you want to make, I can revert that part of this


PavelLaptev (dismissed):
@ndom91 I think the way to resize the field using the `resize observer` is better because it has fewer UI issues. It also doesn't slow things down much because it only checks the size when you edit the field.

The current way (`ch` method) doesn't work well for fonts that aren't the same width everywhere (like you said), it doesn't always get the size right, and it calculates the width incorrectly.👇

https://github.com/user-attachments/assets/9eaae6fc-b4f5-4564-a31b-495cc3fe17d2

Let's switch back to using the `resizeObserver` if you don't mind :-)

ndom91:
@PavelLaptev Yeah fair enough, reverted :+1:

PavelLaptev (approved):
Looks good! Thank you @ndom91

Issue #2650, opened 2024-02-14
feature request : Cloud Login link from app should support copy/paste


bkcsfi:
After launching the application, when I click on Account and then "Sign Up or Log In", it would be very helpful if I could copy the link that's mentioned in the text "click here".

*If you were not redirected automatically, you can  click here*

I need to be able to copy the link so that I can paste it into a browser that is already running and authenticated to "work".

Unfortunately when I click it, Firefox launches with the profile selector. I am unable to choose the work profile because its already running and in-use.

Instead I have to choose a throw-away profile, let the browser start, copy the URL (hoping it doesn't redirect quickly), then paste the url into my "work browser" instance.

Its a pain..

ndom91:
> @ndom91 @PavelLaptev  Perhaps we could have a special external link button which always can be right clicked to copy the URL? 
> 
> I've marked this as "good first issue" as a much simpler fix is to just add in a context menu.

Hmm yeah so for buttons I don't mind that. 

For links, i.e. anchor tags, what do you all think of having one of those "copy to clipboard" btns appear on hover?

PavelLaptev:
Yes, I think I can add that

PavelLaptev:
The PR is merged. Should be in the next release :-)

![image](https://github.com/user-attachments/assets/d4372965-82cb-48ea-957f-742b21c95b9c)

bkcsfi:
great, thanks for looking into this!

PR #4716, opened 2024-08-18
Login flow update and UI updated

PavelLaptev:
## ☕️ Reasoning

- Updated UI components structuring. In order to simplify paths to components removed component groups like `inputs\...`. Path to UI components should be simple unless it's a complex ciomponent like `SegmentController` which consist of two components but you can't use them separatly.  Other components should be in the root folder.
- Added an option to copy the login link.

image

## 🧢 Changes

- Many small changes for paths
- New component `LinkButton`
- Updated `login` functions. Added new function `loginAndCopyLink`

## 🎫 Affected issues

Issue https://github.com/gitbutlerapp/gitbutler/issues/2650


apps/desktop/src/lib/utils/clipboard.ts
```diff
@@ -4,13 +4,15 @@ export function copyToClipboard(text: string) {
 	if (!navigator.clipboard) {
 		toasts.error('Clipboard API not available');
 	} else {
+		console.log('Copying to clipboard:', navigator.clipboard);
```
ndom91: Left over?

ndom91:
Seems like there are some import-order eslint issues and both `/commitLines` and `/CommitLines` are both still present. Same deal with `segmentControl`/`SegmentControl`

apps/desktop/src/lib/utils/clipboard.ts
PavelLaptev: yes, thanks for noticing :-)

packages/ui/src/stories/modal/DemoModal.svelte
```diff
@@ -14,11 +14,16 @@
 	let modal = $state<Modal>();
 </script>
 
-<Button on:click={() => modal?.show()}>Show</Button>
+<Button
+	onclick={() => {
+		console.log('show');
```
mtsgrd: Just scanning this PR, noticed this should probably be removed.

PavelLaptev: Done :-)

PavelLaptev:
> Seems like there are some import-order eslint issues and both `/commitLines` and `/CommitLines` are both still present. Same deal with `segmentControl`/`SegmentControl`. But otherwise looks alright 👍

There was a casing issue. Fixed that

mtsgrd (approved):

PR #4718, opened 2024-08-20
branch stacking first stab

krlvi:
building an mvp for stacking. this requires multiple references associated with the same branch

- [x] Add the custom branch references as a field / new type to the persisted branch data
- [x] Add functions for creating / managing references
- [x] Add tests
- [x] Ensure references are kept up to date during the following operations
   - [x] `cherry_rebase_group`
   - [x] `update_commit_message`
   - [x] `squash_branch_commit`
   - [x] `reorder_commit`
   - [x] `undo_commit`
   - [x] `move_commit_file` 
   - [x] `insert_blank_commit`
- [x] Basic user interface (behind a feature flag)
   - [x] Create refs
   - [x] Show refs
   - [x] Push refs

NB: `move_commit` (which moves the commit from one lane to another) will not move the reference to the new lane. That should be performed via updating the reference


crates/gitbutler-testsupport/src/lib.rs
```diff
@@ -60,40 +60,40 @@ pub fn init_opts_bare() -> git2::RepositoryInitOptions {
...
+    use crate::DRIVER;
+    use gitbutler_command_context::CommandContext;
+    use gitbutler_project::{Project, ProjectId};
 
+    pub fn fixture(script_name: &str, project_directory: &str) -> anyhow::Result<CommandContext> {
+        let project = fixture_project(script_name, project_directory)?;
+        CommandContext::open(&project)
+    }
+    pub fn fixture_project(script_name: &str, project_directory: &str) -> anyhow::Result<Project> {
+        let root = gix_testtools::scripted_fixture_writable_with_args(
```
Byron, line 81: This function returns a tempdir, which is not returned and thus dropped.
Retaining it and returning it next to `Project` should do the trick.

crates/gitbutler-testsupport/src/lib.rs
```diff
@@ -60,40 +60,40 @@ pub fn init_opts_bare() -> git2::RepositoryInitOptions {
...
+            script_name,
+            Some(DRIVER.display().to_string()),
+            gix_testtools::Creation::ExecuteScript,
```
Byron, line 84: There is an alternative creation mode, `CopyFromReadOnly` which might work in some or all cases here. If it does work, it should be preferred as it will be faster to just copy the read-only location into the temporary directory.

Byron (commented):
I hope these suggestions make it work.

crates/gitbutler-testsupport/src/lib.rs
krlvi: this solves my issue, thank you!!

crates/gitbutler-branch-actions/src/virtual.rs
```diff
@@ -2050,6 +2050,9 @@ pub(crate) fn move_commit(
         vb_state.set_branch(destination_branch.clone())?;
     }
 
+    // move the
+    {}
```
Byron: A leftover?

```suggestion
```

crates/gitbutler-branch/src/reference.rs
```diff
@@ -0,0 +1,20 @@
...
+/// Represent a deployable / reviewable part of a virtual branch that can be pushed to a remote
+/// and have a "Pull Request" created for it.
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+pub struct ChangeReference {
+    /// Branch id of the virtual branch this reference belongs to
+    /// Multiple references may belong to the same virtual branch, representing separate deployable / reviewable parts of the vbranch.
+    pub branch_id: BranchId,
+    /// Fully qualified reference name.
+    /// The reference must be a remote reference.
+    pub name: ReferenceName,
```
Byron, line 17: Didn't we have a type that embodies remote references? I remember something that had a remote name and a branch name, a local tracking branch effectively.

Or am I misreading this and it's actually the name of a reference on the remote?

crates/gitbutler-repo/src/change_reference.rs
```diff
@@ -0,0 +1,213 @@
...
+    let handle = VirtualBranchesHandle::new(ctx.project().gb_dir());
+
+    // The branch must exist
+    let mut vbranch = handle.get_branch(branch_id)?;
+
+    // Enusre that the commit acutally exists
+    // let commit = ctx
+    //     .repository()
+    //     .find_commit(commit_id)
+    //     .context(anyhow!("Commit {} does not exist", commit_id))?;
```
Byron: Probably a leftover.

```suggestion
```

crates/gitbutler-repo/src/change_reference.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+    // Ensure the change is not already referenced
+    if all_references.iter().any(|r| r.change_id == change_id) {
+        return Err(anyhow!(
+            "A reference for change {} already exists",
+            change_id
+        ));
+    }
+    validate_commit(&vbranch, commit.id(), ctx, &handle)?;
+    vbranch.references.push(branch_reference.clone());
```
Byron, line 79: Interesting, it looks like these references are fully 'virtual', no Git reference exists for it.
Doing that certainly has the advantage that the user can't alter them, nor will they see them outside of GB.

crates/gitbutler-repo/src/change_reference.rs
```diff
@@ -0,0 +1,213 @@
...
+    credentials: &Helper,
+) -> Result<()> {
+    let handle = VirtualBranchesHandle::new(ctx.project().gb_dir());
+    let vbranch = handle.get_branch(branch_id)?;
+    let reference = vbranch
+        .references
+        .iter()
+        .find(|r| r.name == name)
+        .ok_or_else(|| anyhow!("Reference {} not found", name))?;
+    let upstream_refname = gitbutler_reference::RemoteRefname::from_str(&reference.name)
```
Byron, line 144: I see, so the `reference.name` is something like `refs/remotes/<remote-without-slash>/<name>`.
Thanks to this constructor, I also think it won't work with multi-slash remotes. Fortunately they are rare, but in the back of my head I want to retire it and replace it with something that gets a list of available remotes as well.

While writing this, I also realize that storing remote names like this will make it difficult to properly deal with renamed remotes. But that's yet another topic.

crates/gitbutler-repo/src/change_reference.rs
```diff
@@ -0,0 +1,213 @@
...
+    vbranch: &Branch,
+    handle: &VirtualBranchesHandle,
+    change_id: String,
+) -> Result<git2::Commit<'a>> {
+    let target = handle.get_default_target()?;
+    let branch_commits = ctx
+        .log(vbranch.head, LogUntil::Commit(target.sha))?
+        .iter()
+        .map(|c| c.id())
+        .collect_vec();
```
Byron: There should be no need to collect here.
Not collecting means the search can stop early as well, assuming that `ctx.log()` is a true iteration, which I will make sure one day it will be.

crates/gitbutler-repo/src/change_reference.rs
```diff
@@ -0,0 +1,213 @@
...
+    // Find the commit with the change id
+    let commit = branch_commits
+        .iter()
+        .find(|c| {
+            let commit = ctx.repository().find_commit(**c).expect("Commit not found");
+            commit.change_id() == Some(change_id.clone())
```
Byron: Never clone for comparisons.

```suggestion
            commit.change_id().as_deref() == Some(&change_id)
```

crates/gitbutler-repo/src/change_reference.rs
```diff
@@ -0,0 +1,213 @@
...
+        })
+        .map(|c| ctx.repository().find_commit(*c).expect("Commit not found"))
```
Byron: Let's avoid panics, the UI will not respond kindly.

```suggestion
        .and_then(|c| ctx.repository().find_commit(*c).ok())
```

crates/gitbutler-repo/src/change_reference.rs
```diff
@@ -0,0 +1,213 @@
...
+/// - The commit is between the branch base and the branch head
+fn validate_commit(
+    vbranch: &Branch,
+    commit_id: git2::Oid,
+    ctx: &CommandContext,
+    handle: &VirtualBranchesHandle,
+) -> Result<()> {
+    let target = handle.get_default_target()?;
+    let branch_commits = ctx
+        .log(vbranch.head, LogUntil::Commit(target.sha))?
```
Byron, line 193: Once `log()` returns an iterator, this can be rewritten without collection and thus stop the search once the commit is found. Right now it always has the worst-case performance, like if a commit isn't found.

crates/gitbutler-repo/src/rebase.rs
```diff
@@ -71,9 +71,14 @@ pub fn cherry_rebase_group(
                     if !ctx.project().succeeding_rebases {
                         return Err(anyhow!("failed to rebase")).context(Marker::BranchConflict);
                     }
-                    commit_conflicted_cherry_result(ctx, head, to_rebase, cherrypick_index)
+                    commit_conflicted_cherry_result(ctx, head.clone(), to_rebase, cherrypick_index)
```
Byron:
```suggestion
                    commit_conflicted_cherry_result(ctx, head, to_rebase, cherrypick_index)
```

crates/gitbutler-repo/src/rebase.rs
```diff
@@ -71,9 +71,14 @@ pub fn cherry_rebase_group(
...
                 } else {
-                    commit_unconflicted_cherry_result(ctx, head, to_rebase, cherrypick_index)
+                    commit_unconflicted_cherry_result(
+                        ctx,
+                        head.clone(),
```
Byron:
```suggestion
                        head,
```

crates/gitbutler-repo/tests/change_reference.rs
```diff
@@ -0,0 +1,254 @@
...
+    let (ctx, _temp_dir) = command_ctx("multiple-commits")?;
+    let test_ctx = test_ctx(&ctx)?;
+    let wrong_change = test_ctx.other_commits.first().unwrap().change_id().unwrap();
+    let result = create_change_reference(
+        &ctx,
+        test_ctx.branch.id,
+        "refs/remotes/origin/asdf".into(),
+        wrong_change.clone(),
+    );
+    assert!(result.is_err());
```
Byron, line 140: One *could* even use `unwrap_err().to_string()` and assert on the error message.
Doing so of course increases the likelihood of test failures once the message changes, but also indicates that it is an important text for the user if it's expected to happen a lot.

For now, I'd also think that these errors aren't expected to be common though, so probably better to not assert on their text.

crates/gitbutler-repo/tests/change_reference.rs
```diff
@@ -0,0 +1,254 @@
...
+        reference.name,
+        false,
+        &Helper::default(),
+    );
+    assert!(result.is_ok());
+    Ok(())
+}
+
+fn command_ctx(name: &str) -> Result<(CommandContext, TempDir)> {
+    gitbutler_testsupport::writable::fixture("stacking.sh", name)
```
Byron, line 232: I wonder if it's easier or faster to use this over, say, using the `git2` based test-project initialization.
Never tested it.

In theory, this can be faster as it will just copy files around, even though this is configured to execute the script which then might be slower thanks to the program invocation overhead.

Or in other words: even though I am using these for read-only fixtures, for writable fixtures using the standard test-support might be faster.

crates/gitbutler-repo/tests/fixtures/stacking.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+set -eu -o pipefail
+CLI=${1:?The first argument is the GitButler CLI}
+
+function tick () {
```
Byron, line 5: I'd probably remove this if time isn't asserted.

Byron (commented):
Thanks for inviting me to do a review.
I refrained from pushing into this branch, and instead left a few suggestions.

crates/gitbutler-branch/src/reference.rs
krlvi: the one that we had was kinda broken in subtle ways, so avoiding it here. perhaps this is a place where we could use gix types

crates/gitbutler-repo/src/change_reference.rs
krlvi: yep - these are also references to a change id which provides some stability (eg. when a message is reworded etc)

crates/gitbutler-repo/tests/change_reference.rs
krlvi: i see. i kinda wanted to test this out, and ended getting myself into the situation of needing to expose the scripted fixtures in a writable form. 
Having to write bash can be annoying, so perhaps the other approach is also more ergonomic

Caleb-T-Owens:
Fuck yeah!!!!!!

crates/gitbutler-branch/src/reference.rs
Byron: `gix` 'only' knows `gix::refs::FullName` which is a validated full name of any reference. From there, one can get the short-name as well, but I can't tell how practical it would be here.
I wouldn't fix what ain't broken, but will keep in mind that `ReferenceName` might also not be the most desirable type here.

crates/gitbutler-repo/tests/change_reference.rs
Byron: Yes, `bash` is great only when there are no nicer APIs around, and it's really nice for read-only fixtures with unbeatable performance.
But for everything writable, it's probably not very desirable unless the `Copy` mode can be used so the script isn't executed every time.

PR #4720, opened 2024-08-20
fix: allow github.com in img-src CSP

ndom91:
- Allow `https://github.com` in `img-src` CSP in order for most PR Description images to load correctly


Caleb-T-Owens (approved):
Nice 👍

PR #4721, opened 2024-08-20
Simplify and test updater service

mtsgrd:
- reverts to old UI
- moves more business logic into .ts file
- add tests for a few scenarios
- add a component test to ensure `AppUpdater` displays download/restart buttons

### DIFF
--- a/apps/desktop/src/lib/backend/updater.ts
+++ b/apps/desktop/src/lib/backend/updater.ts
@@ -1,33 +1,29 @@
-import { listen } from './ipc';
+import { Tauri } from './tauri';
 import { showToast } from '$lib/notifications/toasts';
-import { getVersion } from '@tauri-apps/api/app';
 import { relaunch } from '@tauri-apps/api/process';
 import {
-	checkUpdate,
 	installUpdate,
-	onUpdaterEvent,
 	type UpdateResult,
 	type UpdateManifest,
 	type UpdateStatus
 } from '@tauri-apps/api/updater';
 import posthog from 'posthog-js';
-import { derived, writable, type Readable } from 'svelte/store';
+import { derived, readable, writable } from 'svelte/store';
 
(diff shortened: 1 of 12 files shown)


apps/desktop/src/lib/backend/ipc.ts
```diff
@@ -64,3 +71,22 @@ export function listen<T>(event: EventName, handle: EventCallback<T>) {
...
+	}
+
+	async onUpdaterEvent(callback: (status: UpdateStatusResult) => void): Promise<UnlistenFn> {
+		return await onUpdaterEvent(callback);
+	}
+
+	async currentVersion(): Promise<string> {
+		return await getVersion();
+	}
+}
```
mtsgrd: It seems difficult to mock these functions without having an object, I tried several ways with no luck.

Will also move this into its own class and just reference the `ipc.ts` methods instead of duplicating them. Eventually we should move all services over to using an object that can be mocked for testing.

apps/desktop/src/lib/backend/updater.ts
```diff
@@ -1,124 +1,161 @@
...
+const TIMEOUT_SECONDS = 30;
+
+/**
+ * Note that the Tauri API `checkUpdate` hangs indefinitely in dev mode, build
+ * a nightly if you want to test the updater manually.
+ *
+ * export TAURI_PRIVATE_KEY=doesnot
+ * export TAURI_KEY_PASSWORD=matter
+ * ./scripts/release.sh --channel nightly --version "0.5.678"
+ */
```
mtsgrd, line 25: @PavelLaptev you should be able to build a nightly with a specific version using this command.

ndom91 (approved):
LGTM :ok_hand:

PavelLaptev (approved):
Thank you @mtsgrd
Should be good now. I just made a few adjustments to some conditions.

mtsgrd:
Awesome, thank you!

PR #4723, opened 2024-08-20
fix: skip bundling AppImage updater tarball

ndom91:
- Reduce AppImage artifact size

### 👷 Changes

- Because we set the AppImage passphrase env var, `tauri build` generates a "raw" `git-butler.AppImage`, as well as the updater artifacts like a tarball of the identical AppImage (`git-butler.AppImage.tar.gz`) and a signature of that tarball (`git-butler.AppImage.tar.gz.sig`).
- The tarball and it's signature are deisgned to be used by the Tauri updater
- However, we currently don't support updating on Linux / with AppImages ([Error](https://docs.rs/tauri/latest/tauri/updater/enum.Error.html#variant.UnsupportedLinuxPackage))
- So I figure, until that changes, we can remove those two files (the tarball and its signature) from the resulting artifact that is uploaded to GitHub. 
- **This cuts the AppImage artifact size almost in half**

What do you all think? Fine to cut out for now?


ndom91:
Think I found out how to update appimage, closing this

PR #4729, opened 2024-08-21
Rename gitbutler/integration -> gitbutler/workspace

mtsgrd:
It makes it easier to communicate the meaning of the branch, a merge of all the commits in the virtual branches in your workspace.

warning: this pr will transition you from integration -> workspace, but unapplying it will not undo that transition. Clear your workspace first.

### DIFF
--- a/crates/gitbutler-branch-actions/src/integration.rs
+++ b/crates/gitbutler-branch-actions/src/integration.rs
@@ -4,12 +4,13 @@ use anyhow::{anyhow, Context, Result};
 use bstr::ByteSlice;
 use gitbutler_branch::{
     self, Branch, BranchCreateRequest, SignaturePurpose, VirtualBranchesHandle,
-    GITBUTLER_INTEGRATION_REFERENCE,
+    GITBUTLER_WORKSPACE_REFERENCE,
 };
 use gitbutler_cherry_pick::RepositoryExt as _;
 use gitbutler_command_context::CommandContext;
 use gitbutler_commit::commit_ext::CommitExt;
 use gitbutler_error::error::Marker;
+use gitbutler_operating_modes::OPEN_WORKSPACE_REFS;
 use gitbutler_project::access::WorktreeWritePermission;
 use gitbutler_repo::{LogUntil, RepoActionsExt, RepositoryExt};
 use tracing::instrument;
@@ -94,9 +95,9 @@ pub(crate) fn get_workspace_head(ctx: &CommandContext) -> Result<git2::Oid> {
     Ok(workspace_head_id)
 }
 
-// Before switching the user to our gitbutler integration branch we save
+// Before switching the user to our gitbutler workspace branch we save
 // the current branch into a text file. It is used in generating the commit
-// message for integration branch, as a helpful hint about how to get back
+// message for workspace branch, as a helpful hint about how to get back
 // to where you were.
 struct PreviousHead {
     head: String,
@@ -138,8 +139,8 @@ pub fn update_gitbutler_integration(
     let integration_filepath = repo.path().join("integration");
     let mut prev_branch = read_integration_file(&integration_filepath)?;
     if let Some(branch) = &prev_branch {
-        if branch.head != GITBUTLER_INTEGRATION_REFERENCE.to_string() {
-            // we are moving from a regular branch to our gitbutler integration branch, write a file to
+        if branch.head != GITBUTLER_WORKSPACE_REFERENCE.to_string() {
+            // we are moving from a regular branch to our gitbutler workspace branch, write a file to
             // .git/integration with the previous head and name
             write_integration_file(&head_ref, integration_filepath)?;
             prev_branch = Some(PreviousHead {
@@ -218,14 +219,14 @@ pub fn update_gitbutler_integration(
         parents.iter().collect::<Vec<_>>().as_slice(),
     )?;
 
-    // Create or replace the integration branch reference, then set as HEAD.
+    // Create or replace the workspace branch reference, then set as HEAD.
     repo.reference(
-        &GITBUTLER_INTEGRATION_REFERENCE.clone().to_string(),
+        &GITBUTLER_WORKSPACE_REFERENCE.clone().to_string(),
         final_commit,
         true,
         "updated integration commit",
     )?;
-    repo.set_head(&GITBUTLER_INTEGRATION_REFERENCE.clone().to_string())?;
+    repo.set_head(&GITBUTLER_WORKSPACE_REFERENCE.clone().to_string())?;
 
     let mut index = repo.index()?;
     index.read_tree(&workspace_tree)?;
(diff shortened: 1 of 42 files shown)


mtsgrd:
@krlvi this feels pretty nice, what do you think?

image

krlvi:
lets ship it

krlvi (approved):

mtsgrd:
Right on. I'm traveling atm but will rebase and land within.. about an hour.

mtsgrd:
@Caleb-T-Owens would you mind giving this a proper read through? I should be changing integration -> workspace everywhere, except renaming the "integration" crate.

There's a couple of places in which we still want to recognise both the old and new name since otherwise it wouldn't smoothly transition from old to new.

crates/gitbutler-branch-actions/src/integration.rs
```diff
@@ -152,34 +154,39 @@ pub fn update_gitbutler_integration(
...
 
     // message that says how to get back to where they were
-    let mut message = GITBUTLER_INTEGRATION_COMMIT_TITLE.to_string();
+    let mut message = GITBUTLER_WORKSPACE_COMMIT_TITLE.to_string();
     message.push_str("\n\n");
-    message.push_str(
-        "This is an integration commit for the virtual branches that GitButler is tracking.\n\n",
-    );
+    if !virtual_branches.is_empty() {
+        message.push_str("This is an merge commit the virtual branches in your workspace.\n\n");
```
Caleb-T-Owens: `an merge` -> `a merge`

crates/gitbutler-diff/src/diff.rs
```diff
@@ -96,17 +96,17 @@ impl GitHunk {
 
 /// Comparison
 impl GitHunk {
-    /// integration_intersects_unapplied is used to determine if a hunk from a diff between integration and the trunk intersects with an unapplied hunk.
+    /// workspace_intersects_unapplied is used to determine if a hunk from a diff between workspace and the trunk intersects with an unapplied hunk.
     /// We want to use the new start/end for the integraiton hunk and the old start/end for the unapplied hunk.
-    pub fn integration_intersects_unapplied(
-        integration_hunk: &GitHunk,
+    pub fn workspace_intersects_unapplied(
```
Caleb-T-Owens: The naming here feels a little janky. (I know I wrote this originally).

What do we want to call something that is localized to a branch.

Workspace-scoped or just "workspace_x" feels reasonable for things in the combined scope, but what do we want for things that are branch-scoped?

Also, you missed a miss-spelling of integration in the comment 😅

Caleb-T-Owens:
Where does the code which actually moves you over live, and I would love to see some tests which exercised the code path before we merge in.

crates/gitbutler-branch-actions/tests/fixtures/for-workspace-migration.sh
```diff
@@ -0,0 +1,18 @@
...
+)
+
+export GITBUTLER_CLI_DATA_DIR=../user/gitbutler/app-data
+git clone remote workspace-migration
+(cd workspace-migration
+  $CLI project add --switch-to-workspace "$(git rev-parse --symbolic-full-name @{u})"
+  $CLI branch create virtual
+  # Start the test on the old integration branch.
+  git checkout -b gitbutler/integration
+)
```
mtsgrd, line 18: @Byron I'm adding this file to follow the pattern set by other tests, but I'm a bit unsure how to think about the output in `generated-do-not-edit`. The associated test fails if run twice without removing the output. Is there something I should do in either this file, or in the rust code, such that I can run the tests repeatedly?

crates/gitbutler-diff/src/diff.rs
mtsgrd: Thanks for spotting spelling mistakes!

About the name, by unapplied_hunk, do we really mean uncommitted_hunk? I also see "unstaged_hunk", but that to me feels like a misnomer since we do not engage in staging of changes.

mtsgrd:
>I'm also not sure about calling the merge commit we make on the workspace branch the "workspace commit". Perhaps its better to stick with a family of things are workspace related.

Can you elaborate?

crates/gitbutler-branch-actions/tests/virtual_branches/workspace_migration.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    // Should not throw verification error until migration is complete.
+    let result = verify_branch(&ctx, perm);
+    assert!(result.is_ok());
+
+    // Updating workspace commit should put us on the workspace branch.
+    update_workspace_commit(&VirtualBranchesHandle::new(ctx.project().gb_dir()), &ctx)?;
+    assert_eq!(ctx.repository().head()?.name(), Some(WORKSPACE_BRANCH_REF));
+    Ok(())
+}
```
mtsgrd: >Where does the code which actually moves you over live, and I would love to see some tests which exercised the code path before we merge in.

@Caleb-T-Owens it's a call to update_workspace_commit that actually moves you over, so what we need to prevent is `verify_branch` throwing an error while you're still on `gitbutler/integration`. Can you think of additional steps we should be taking?

crates/gitbutler-diff/src/diff.rs
Caleb-T-Owens: Hmmmm...

Taking a look at the usage.

`integration_hunk` is hunk which has been created by diffing the head of an individual virtual branch against the base commit. (of note, its not the integration commit that is getting diffed, which may cause line number differences if the same file has had additions in two different branches)
`unapplied_hunk` is an uncommitted hunk that has been created by diffing all uncommitted changes against the integration commit.


renaming `unapplied_hunk` to `uncommitted_hunk` makes sense. I think `integration_hunk` might best be renamed to `branch_hunk`

crates/gitbutler-branch-actions/tests/virtual_branches/workspace_migration.rs
Caleb-T-Owens: @mtsgrd you should call `assure_open_workspace_mode` to only do the move whilst on `gitbutler/integration`

mtsgrd: @Caleb-T-Owens that function does not have any side-effects and makes the test fail. What did you have in mind?

image

mtsgrd: Done.

crates/gitbutler-diff/src/diff.rs
mtsgrd: I had a look, a few things need renaming in the surrounding code. Let's land this and follow-up with a separate pr for that?

crates/gitbutler-branch-actions/tests/virtual_branches/workspace_migration.rs
mtsgrd: Btw, we should probably use the verb _ensure_ rather than _assure_. Ensure means “make certain that something will occur.” Assure means “convince” or “give confidence to.”

crates/gitbutler-branch-actions/tests/fixtures/for-workspace-migration.sh
Byron: I think @krlvi fell for this as well, as the tests are designated to be read-only. Many of the even read-only sounding operations actually write, which destroys the fixture.

If these are executed through Rust, I think @krlvi added ways to get a writable fixture, i.e. one that is created into a temp-dir.
Does this help? (I didn't look at the code here specifically)

crates/gitbutler-branch-actions/tests/virtual_branches/workspace_migration.rs
Byron: I think here is a problem, as a read-only fixture is used for something that clearly changes it.
There should be `::writable::fixture()` now that can be used instead.

@krlvi and I were also contemplating that using these scripts isn't an advantage when they are *not* read-only as they will be slower to produce than just doing everything in Rust like has been done before.

Of course, maybe there is other good reasons to use it that I have overlooked.

PR #4730, opened 2024-08-21
Fix applied -> in_workspace performance issue

mtsgrd:
- loop was taking > 1s for me
- my 8 branches (2 in workspace) were constantly being saved


krlvi (approved):
oh perhaps this is why the unapplying did not help with a performance issue that one person on discord had. nice find!!

Caleb-T-Owens:
@mtsgrd and @krlvi we can probably get rid of this function and the `applied` property now as the switch over point was reasonably far back in time

PR #4731, opened 2024-08-21
Delete entry from .toml file when branch is integrated

mtsgrd:
- virtual_branches.toml file was full of old entries

### DIFF
--- a/crates/gitbutler-branch-actions/src/base.rs
+++ b/crates/gitbutler-branch-actions/src/base.rs
@@ -388,7 +388,7 @@ pub(crate) fn update_base_branch(
                 if non_commited_files.is_empty() {
                     // if there are no commited files, then the branch is fully merged,
                     // and we can delete it.
-                    vb_state.mark_as_not_in_workspace(branch.id)?;
+                    vb_state.delete_branch_entry(&branch.id)?;
                     ctx.delete_branch_reference(&branch)?;
                     Ok(None)
                 } else {


mtsgrd:
@krlvi @Caleb-T-Owens I think we might need some retroactive cleanup if I'm correct about this change?

Caleb-T-Owens:
@mtsgrd the reason we don't just delete the entries from virtual_branches.toml is so that we can re-use the fancy name when the branch gets re-applied.

If we want to "garbage collect" entries, we ought to do it when there are no local or remote reference which relate to a given name

Caleb-T-Owens:
(or we have to accept that the fancy names will be lost)

mtsgrd:
> If we want to "garbage collect" entries, we ought to do it when there are no local or remote reference which relate to a given name

This PR changes the code to delete the entry when we detect upstream has been integrated, and after we have deleted the local branch. In this case surely it doesn't make sense to retain the fancy name? In our case the remote branch is also automatically deleted after merging to master, so unless I'm mistaken, the only reference left is in fact in that toml file.

Caleb-T-Owens (approved):

mtsgrd:
@ndom91 I just noticed vercel deployed here based on a change in an .rs file. What options did we have for fixing this?

ndom91:
> @ndom91 I just noticed vercel deployed here based on a change in an .rs file. What options did we have for fixing this?

It seems to always deploy once for the first commit of a branch, since it doesn't know which commit to compare that against. We can pass the `turbo-ignore` cli a `--fallback` ref though for those situations. 

See my comment here: https://github.com/gitbutlerapp/gitbutler/pull/4664#issuecomment-2281203216

PR #4735, opened 2024-08-21
fix: zoom shortcut codes for international keyboard users

### LINKED ISSUE #4733: Wrong shortcuts for zoom on non-US keyboards
### Version

0.12.20

### Operating System

Mac OS X

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

When trying to zoom on a non-US keyboard via `CMD +/-`, nothing changes (at least not the desired behaviour, most of the times). 
On a German keyboard, however, `CMD ß` zooms out and `CMD ´` zooms in. On a Swedish keyboard, `CMD +` zooms out.

### How to reproduce

Use a non-US keyboard and try to zoom. Nothing will happen. 

### Expected behavior

`CMD +` and `CMD -` should zoom in and out, respectively.

### Relevant log output

_No response_

ndom91:
- Fix zoom shortcuts for international keyboard users
- use `event.key` instead of `event.code`

Fixes: #4733


apps/desktop/src/routes/+layout.svelte
```diff
@@ -92,13 +92,13 @@
 	});
 
 	const handleKeyDown = createKeybind({
-		'$mod+Equal': () => {
```
H1ghBre4k3r, line 95: I guess, we should use `+` instead of `=` for this. At least on German keyboards, they are on different keys (which was the reason for using `Equal` here).

H1ghBre4k3r (changes requested):

apps/desktop/src/routes/+layout.svelte
ndom91: Hmm yeah so that seems to break the shortcut for English language keyboard users though because then it technically has to be `$mod+Shift++` :thinking: 

I guess that's technically more correct, but never works that way anywhere else

ndom91: How about just having both shortcuts? One for english keyboard users and one for German and other keyboards. I can't immediately see anythign wrong with that :thinking: We're not using `$mod+=` for anything else. i.e.

```ts
'$mod++': () => {
	zoom = Math.min(zoom + 0.0625, 3);
},
'$mod+=': () => {
	zoom = Math.min(zoom + 0.0625, 3);
},
```

Just tested it, it does seem to work for English keyboards without any side effects since tehcnically `$mod++` is never possible for us to hit, since to get `+` you also have to hold `Shift`

H1ghBre4k3r: Ah - I wasn't aware of the issues in the English keyboard. To tackle that, we could use the feature to use regular expressions in the shortcuts. E.g., 

```ts
'$mod+([+=])': () => {
	zoom = Math.min(zoom + 0.0625, 3);
},
``` 

Or would you prefer to have two separate shortcuts?

ndom91: Oddly that regex one doesn't work for me on the english keyboard. I'll push the double one for now :+1:

H1ghBre4k3r (approved):
LGTM. 👍

mtsgrd (approved):

PR #4736, opened 2024-08-22
feat: use PR template when available

### LINKED ISSUE #3313: 'Create pull request' should use PR template from Github if present
When creating a pull request, GitButler doesn't use the Github PR template.

ndom91:
### 👷 Changes

- Use a repositories default `.github/PULL_REQUEST_TEMPLATE.md` when available and no other `body` has been passed
	- Add optional `fetchPrTemplate` method to `GitHostPrService` class 
	- Implement this method in the GitHub PR Service
- GitHost's arguments were beginning to get too long, therefore refactored them into an object
	- Updated all relevant tests
- Harmonized various GitHost's property names and added shared type for constructor args

### 📓 Notes

Fixes: #3313

### DIFF
--- a/apps/desktop/src/lib/gitHost/github/githubPrService.ts
+++ b/apps/desktop/src/lib/gitHost/github/githubPrService.ts
@@ -3,8 +3,8 @@ import { DEFAULT_HEADERS } from './headers';
 import { ghResponseToInstance, parseGitHubDetailedPullRequest } from './types';
 import { sleep } from '$lib/utils/sleep';
 import posthog from 'posthog-js';
-import { get, writable, type Readable } from 'svelte/store';
-import type { Settings } from '$lib/settings/userSettings';
+import { get, writable } from 'svelte/store';
+import type { Persisted } from '$lib/persisted/persisted';
 import type { RepoInfo } from '$lib/url/gitUrl';
 import type { GitHostPrService } from '../interface/gitHostPrService';
 import type { DetailedPullRequest, MergeMethod, PullRequest } from '../interface/types';
@@ -20,7 +20,7 @@ export class GitHubPrService implements GitHostPrService {
 		private repo: RepoInfo,
 		private baseBranch: string,
 		private upstreamName: string,
-		private userSettings?: Readable<Settings>
+		private usePullRequestTemplate?: Persisted<boolean>
 	) {}
 
 	async createPr(title: string, body: string, draft: boolean): Promise<PullRequest> {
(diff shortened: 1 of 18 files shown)


apps/desktop/src/lib/gitHost/github/githubPrService.ts
```diff
@@ -19,16 +19,21 @@ export class GitHubPrService implements GitHostPrService {
 		private upstreamName: string
 	) {}
 
-	async createPr(title: string, body: string, draft: boolean): Promise<PullRequest> {
+	async createPr(
+		title: string,
+		body: string,
+		draft: boolean,
+		templatePath: string = '.github/PULL_REQUEST_TEMPLATE.md'
```
mtsgrd: Let's use a constant instead of parameter until a parameter is needed? If you see this signature you'll assume there is e.g.  more than one caller.

ndom91: Yeah fair enough, I'll refactor this a bit 👍

Caleb-T-Owens:
Hmm.

I'm not sure that its the responsibility of `createPR` to find and use the template like that.

It feels like we should instead have some other class which handles the business logic of formatting the message.

mtsgrd:
Good point, and the path will be different depending on the host. @ndom91 what if you added it under `lib/gitHost` as e.g. `GitHostTemplates`?

ndom91:
I think @Caleb-T-Owens meant formatting as in building the body of the PR. 

Just playing devils advocate a bit here, each git host will bring with it a different way to fetch the PR template body (i.e. via `octokit` for GitHub, maybe a plain `fetch()` call for GitLab, maybe the BitBucket SDK allows doing it via a flag in their `createPr` method, etc.), so I'm having trouble seeing an early abstraction here as worthwhile.

What about just a separate method in the PrService of each host? Putting it directly in the `createPr()` function maybe wasn't ideal, but a separate `[gitHost]PrService.fetchPrTemplate()` sounds alright to me. 

Thoughts?

mtsgrd:
I was thinking a toggle would be nice, I wouldn't necessarily want that fetch to happen every time if there is no template. But, I also feels like this begs at a larger question.. We should probably run some one-time check on adding a project so we could suggest "we see you have a template, would you like to use it?".

mtsgrd (approved):

apps/desktop/src/lib/gitHost/github/githubPrService.ts
```diff
@@ -1,34 +1,39 @@
...
 import posthog from 'posthog-js';
-import { writable } from 'svelte/store';
+import { get, writable } from 'svelte/store';
+import type { Persisted } from '$lib/persisted/persisted';
 import type { RepoInfo } from '$lib/url/gitUrl';
 import type { GitHostPrService } from '../interface/gitHostPrService';
 import type { DetailedPullRequest, MergeMethod, PullRequest } from '../interface/types';
 import type { Octokit } from '@octokit/rest';
 
+const DEFAULT_PULL_REQUEST_TEMPLATE_PATH = '.github/PULL_REQUEST_TEMPLATE.md';
```
mtsgrd, line 14: @ndom91 I just realized the gh docs say this filename should be lowercase. https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

image

PR #4741, opened 2024-08-22
fix: add linux dependencies to tauri.conf

ndom91:
- ~Testing~ Tested whether this will result in these dependencies automaticlaly being installed on the respective platforms via their package managers (apt/dpkg, rpm, etc.)
	- Update: It doesn't look like the two package managers I tested will _automatically_ install the dependencies, but it does warn everywhere you'd expect it to 

Ubuntu complains when you try to uninstall stuff we depend on:
> image

It also complains when installing `git-butler-nightly` that pkgs are potentially missing:
>image

And same goes for Fedora (rpm):
> image

Should result in a smoother experience overall for our linux users :pray:

### DIFF
--- a/crates/gitbutler-tauri/tauri.conf.nightly.json
+++ b/crates/gitbutler-tauri/tauri.conf.nightly.json
@@ -26,6 +26,17 @@
 			],
 			"windows": {
 				"certificateThumbprint": null
+			},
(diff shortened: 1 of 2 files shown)


krlvi (approved):
lgtm

PR #4743, opened 2024-08-23
Fix wdio and tsconfig errors

Caleb-T-Owens:
I was getting two errors. In the `wdio.conf.ts` I was getting errors about `process.stdout` and `process.stderr` having bad types. I was also having an error in my `apps/desktop/tsconfig.json` complaining that it couldn't find a package called "node".

As it so turns out, the `types` field in the `tsconfig.json` is for globally importing packages for their types. (packages called `@types/*` have their types globally imported by default). We never had a package called `node` so all hell was breaking loose.

```diff
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -81,7 +81,8 @@
 		"tinykeys": "^2.1.0",
 		"ts-node": "^10.9.2",
 		"vite": "catalog:",
-		"vitest": "^2.0.5"
+		"vitest": "^2.0.5",
+		"@types/node": "^22.3.0"
 	},
 	"dependencies": {
 		"openai": "^4.47.3"
```

```diff
--- a/apps/desktop/tsconfig.json
+++ b/apps/desktop/tsconfig.json
@@ -14,7 +14,7 @@
 		"sourceMap": true,
 		"strict": true,
 		"experimentalDecorators": true,
-		"types": ["node", "@wdio/globals/types", "@wdio/mocha-framework", "vitest/importMeta"]
+		"types": ["@wdio/globals/types", "@wdio/mocha-framework", "vitest/importMeta"]
 	},
 	"include": [
 		"vitest-setup.js",
```

### DIFF
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -81,7 +81,8 @@
 		"tinykeys": "^2.1.0",
 		"ts-node": "^10.9.2",
 		"vite": "catalog:",
-		"vitest": "^2.0.5"
+		"vitest": "^2.0.5",
+		"@types/node": "^22.3.0"
 	},
 	"dependencies": {
 		"openai": "^4.47.3"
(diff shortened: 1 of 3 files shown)


ndom91 (approved):
Hmm, nice find!

PR #4744, opened 2024-08-23
better rev-walk in branch listing

Byron:
Follow-up of #4670

### Tasks

* [x] Use improved rev-walk API for convenience similar to `git2`
* [x] validate that rev-walk won't accidentally 'overshoot' and traverse the entire history
    - This was an issue before setting the cutoff by hand, but fortunately the new API covers this automatically.
* [x] validated visually on GitLab repository that the traversal order is the same

### DIFF
--- a/crates/gitbutler-repo/src/repository.rs
+++ b/crates/gitbutler-repo/src/repository.rs
@@ -449,10 +449,17 @@ impl RepoActionsExt for CommandContext {
 
 type OidFilter = dyn Fn(&git2::Commit) -> Result<bool>;
 
+/// Generally, all traversals will use no particular ordering, it's implementation defined in `git2`.
 pub enum LogUntil {
+    /// Traverse until one sees (or gets commits older than) the given commit.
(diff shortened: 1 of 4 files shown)


crates/gitbutler-repo/src/repository.rs
```diff
@@ -449,10 +449,17 @@ impl RepoActionsExt for CommandContext {
 
 type OidFilter = dyn Fn(&git2::Commit) -> Result<bool>;
 
+/// Generally, all traversals will use no particular ordering, it's implementation defined in `git2`.
```
Byron, line 452: I looked into replacing `l`  with `gitoxide`, but only ended up documenting this type as I though it's not really worth replacing something this trivial as it would need to on-demand open a `gix::Repository` again.

Once `merge_base` is also available, I think I might change the `CommandContext` to also open a `gix::Repository` lazily.

krlvi (approved):

PR #4748, opened 2024-08-23
fix: use our own `open-rs` implementation instead of relying on tauri's "shell-open"

ndom91:
## ☕️ Reasoning

- Using Tauri's `open-shell` feature meant that it was bundling in the `xdg-open` program from the build environment. This lead to two issues
	- `xdg-open` is just a bash script which hard-codes paths to the many coreutils it requires based on the platform it's on
	- Tauri would also pass all environment variables on to this child process of course, but in the case of the AppImage build, this included modified versions of many important paths for looking up libraries for dynamically linked binaries. This lead to various errors on different platforms like libraries not being found, version mismatches, etc.
    - For more info, see the tauri folks briefly discussing this recently: https://github.com/tauri-apps/tauri/issues/10078

## 🧢 Changes

- Drop usage of Tauri feature `open-shell`
- Add our own "open external URL" fn based on `open-rs` crate (same underlying crate used by Tauri)
	- Cleans up the env vars passed to the child proc by removing any paths in the configured env vars with the strings `appimage-run` and `/tmp/.mount`
	- Spawns a new thread to execute the open cmd in as some platforms apparently block here
	- `open-rs` still calls the correct binary on other platforms (like `open` on darwin), so that it still works on all platforms and for protocols other than http/https as well (i.e. `vscode://`).
- Expose that fn to frontend as `open_url`
- Ensure all frontend external URL invocations are using this instead of just being anchor tags with `href` attributes, for example
- In our rendered markdown, we just inserted an anchor tag where links appeared, however, this wouldn't work on linux either. We need to be able to render our own custom components (i.e. our own pre-existing `<Link />` in this case). So the markdown rendering also got a good bit of changes to it here


crates/gitbutler-tauri/src/open.rs
```diff
@@ -0,0 +1,24 @@
+use crate::error::Error;
+use tracing::instrument;
+
+pub fn open_that(path: String) {
+    // `open` required to run in separate thread, to avoid blocking on some
+    // platforms (eg Fedora38 blocks)
+    std::thread::spawn(|| {
+        for mut cmd in open::commands(path) {
```
Byron: It's very interesting that this is needed as I always thought that `tauri` is already using `open-rs` under the hood. Thus I would have hoped it works just as well there as it can work here.

ndom91: Yeah it is, but if you use the "open-shell" feature from tauri, it also packs the `xdg-open` binary from the build env, which I'm trying to avoid.

This is still very much WIP, might need to whitelist `xdg-open` as a binary we can call from the users system as well

crates/gitbutler-tauri/src/open.rs
```diff
@@ -1,24 +1,28 @@
 use crate::error::Error;
+use anyhow::anyhow;
+use regex::Regex;
 use tracing::instrument;
 
-pub fn open_that(path: String) {
-    // `open` required to run in separate thread, to avoid blocking on some
-    // platforms (eg Fedora38 blocks)
+pub fn open_that(path: String) -> Result<(), Error> {
+    let re = Regex::new(r"^((https://)|(http://)|(mailto:)|(vscode://)|(vscodium://)).+").unwrap();
```
Byron: I'd use the `url` crate here, parse the URL, and check what protocol/schema it's using. Then a match can be used which I'd find more maintainable than a long one-line regex.
Alternatively, I think `regex` might support multi-line regex.

ndom91: Yeah good idea. This is all still a bit of a PoC, but I'll definitely implement that if we decide to go ahead with this :ok_hand:

crates/gitbutler-tauri/src/open.rs
```diff
@@ -0,0 +1,124 @@
+use crate::error::Error;
+use anyhow::anyhow;
+use regex::Regex;
+use tracing::instrument;
+
+// fix_env_variables
+use std::ffi::{OsStr, OsString};
+use std::io;
+use std::os::unix::prelude::OsStrExt;
```
Byron: This will probably prevent it from compiling on Windows. There is [`gix::path::os_string_into_bstring()`](https://docs.rs/gix-path/latest/gix_path/fn.os_string_into_bstring.html) which is cross-platform.

But it turns out that it's probably better to work with `PathBuf` here.

crates/gitbutler-tauri/src/open.rs
```diff
@@ -0,0 +1,124 @@
...
+    command.env_remove("APPIMAGE");
+    command.env_remove("APPDIR");
+
+    // it was set by AppRun
+    command.env_remove("PYTHONHOME");
+}
+
+#[tauri::command(async)]
+#[instrument(err(Debug))]
+pub fn open_url(url: &str) -> Result<(), Error> {
```
Byron, line 73: Tauri can easily inject the AppHandle here, and then it can pass the appdir as parameter.

crates/gitbutler-tauri/src/open.rs
```diff
@@ -0,0 +1,124 @@
...
+                if cmd.status().is_ok() {
+                    break;
+                };
+            }
+        }
+    });
+    Ok(())
+}
+
+static APPDIR: ArcSwapOption<OsString> = ArcSwapOption::const_empty();
```
Byron: No :).

Please see below for how to do that without globals.

crates/gitbutler-tauri/src/open.rs
Byron, line 73: This should work and allow to extract the app-dir when `open_that()` is called, to be passed as parameter.

```suggestion
pub fn open_url(app_handle: tauri::AppHandle, url: &str) -> Result<(), Error> {
```

crates/gitbutler-tauri/src/open.rs
```diff
@@ -0,0 +1,124 @@
...
+        if command.get_envs().any(|(name, _)| name == var_name) {
+            continue; // do not change manually specified variables
+        }
+        // in this case, the variable will be inherited from the environment
+
+        let Some(current) = std::env::var_os(var_name) else {
+            continue; // the variable is not set; nothing to do
+        };
+
+        let current_bytes = current.as_bytes();
```
Byron: I think it *could* be better to turn it into a `PathBuf` (just with `.into()`).
But I also admit that I don't really know what it's trying to do (or how these paths look like). It seems like `<app_dir>:something/else`, but can that be?

Byron (commented):
It's getting there!

Also, I have some more notes, predominantly on the global variable. Something that would have helped me is a full example of the kind of paths it's fixing there. In any case, opening something seems very involved and it surprised me a little.

crates/gitbutler-tauri/src/open.rs
ndom91: Haha thanks for all the feedback, I copied this fix-env-var from elsewhere to see if it fixes those Linux library issues I mentioned to you. Will continue investigating today

ndom91:
> It's getting there!
> 
> Also, I have some more notes, predominantly on the global variable. Something that would have helped me is a full example of the kind of paths it's fixing there. In any case, opening something seems very involved and it surprised me a little.
> 

Haha thanks for all the feedback, I copied this fix-env-var from elsewhere to see if it fixes those Linux library issues I mentioned to you. Will continue investigating today. 

But yeah definitely surprising me too how non-straightforward this is turning out to be 😂

ndom91:
@Byron found an issue (from 7yr ago 😅) that perfectly describes what I'm running into as well. Seems to be an appimage specific thing, and looks like all that might be required is unsetting `XDG_DATA_DIRS`: https://github.com/AppImage/AppImageKit/issues/124#issuecomment-341929806

crates/gitbutler-tauri/src/open.rs
ndom91: Okay so I'm going back through your comments now, and regarding this one - these are all either actual URL protocols (`http`,`https`,`mailto`) or custom protocols that should be handled like URLs by the OS, no? 

For example, these can all be passed on to `xdg-open` on Linux or `open` on MacOS and the OS then needs to query its list of protocol handlers for the right app, for example for `vscodium://`.

I think we'd be doing unnecessary work trying to find the right binary to pass it to at this point

crates/gitbutler-tauri/src/open.rs
ndom91: I think the rest of these are no longer relevant as I've removed the majority of that `fix_env_variables` fn from before.

ndom91:
Built a test nightly, would appreciate Mac OS folks checking this out and seeing if you can still open external links, like those in the "Help" menu and links to branches / PR's in the branch headers :pray: 

https://github.com/gitbutlerapp/gitbutler/actions/runs/10600508432

crates/gitbutler-tauri/src/open.rs
ndom91: Happy about any other rust feedback you may have on this though :pray:

ndom91:
`0.5.704` seems good :pray: :crossed_fingers: 

- Linux AppImage: https://releases.gitbutler.com/releases/nightly/0.5.704-1210/linux/x86_64/git-butler-nightly_0.5.704_amd64.AppImage.tar.gz
- Mac OS Apple Silicon: https://releases.gitbutler.com/releases/nightly/0.5.704-1210/macos/aarch64/GitButler_Nightly_0.5.704_aarch64.dmg

crates/gitbutler-tauri/src/menu.rs
```diff
@@ -303,17 +304,17 @@ pub fn handle_event<R: Runtime>(event: &WindowMenuEvent<R>) {
...
             "help/report-issue" => {
-                open::that("https://github.com/gitbutlerapp/gitbutler/issues/new")
+                open::open_url("https://github.com/gitbutlerapp/gitbutler/issues/new")
             }
-            "help/discord" => open::that("https://discord.com/invite/MmFkmaJ42D"),
-            "help/youtube" => open::that("https://www.youtube.com/@gitbutlerapp"),
-            "help/x" => open::that("https://x.com/gitbutler"),
+            "help/discord" => open::open_url("https://discord.com/invite/MmFkmaJ42D"),
+            "help/youtube" => open::open_url("https://www.youtube.com/@gitbutlerapp"),
+            "help/x" => open::open_url("https://x.com/gitbutler"),
```
Byron: Usually I avoid duplicating parts of the name, so `open::url()` would probably be all it needs here.

crates/gitbutler-tauri/src/open.rs
Byron: If you are saying that it should be fine to let `open()` handle the URL entirely without pre-filtering it, then I agree.

crates/gitbutler-tauri/src/open.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    fn clean_env_vars(var_names: &[&str]) {
+        for var_name in var_names {
+            if let Ok(var_value) = env::var(var_name) {
+                let cleaned_value = var_value
+                    .split(':')
+                    .filter(|path| !path.contains("appimage-run"))
+                    .collect::<Vec<_>>()
+                    .join(":");
+                env::set_var(var_name, cleaned_value);
```
Byron: The environment of the running process should not be affected - it's an unexpected side-effect.
It's certainly OK to read, but the new values should be written to the 'about-to-be-started' process only.

Byron (changes requested):
Thanks for the invite, and sorry for the delay.

The reason for requesting changes is that the function alters the environment of the entire process, but it should be straightforward to make it to only affect the process it's about to start.

crates/gitbutler-tauri/src/open.rs
ndom91: Hmm okay yeah I get that this is not a good programming practice, but this is also the crux of this bug fix. See the appimage is basically a zip file that packages in all the code and libraries required to run gitbutler. Including all the `lib*.so` C libraries. 

To make those work, it also prepends a bunch of paths to important env vars like `LD_LIBRARY_PATH` and `XDG_DATA_DIRS`. So that, for example, `XDG_DATA_DIRS` is no longer `XDG_DATA_DIRS=/usr/share/gnome:/usr/local/share:/usr/share:/var/lib/snapd/desktop`, but inside the appimage is then `XDG_DATA_DIRS=/run/appimage-run/abc123/usr/lib:/usr/share/gnome:/usr/local/share:/usr/share:/var/lib/snapd/desktop`.

So in the case of `xdg-open`, it executes that bash script which is on the host system, with the env vars modified by the appimage, and then subprocesses of `xdg-open` are looking for their libraries in incorrect places. 

In my local system's case specifically, `xdg-open` calls `gdbus` to open a lib and `gdbus` finds one of its dynamically linked libraries in the `../appimage-run/*` path instead of the host's normal library path, there's a version mismatch, and it fails due to not being able to find some function. I ran into very similar errors in Ubuntu and Fedora. [This guy](https://github.com/vrc-get/vrc-get/pull/1260#issuecomment-2242153215) has the same thing with openssl when `xdg-open` uses `kde-open` on KDE

ndom91: So long story short, I guess we could place these `appimage-run` paths at the end of the env var's, but either way, they need to be modified to fix this bug :/

ndom91: Ahh wait sorry I just reread your comment - yes, the idea is to only modify it for the process it's about to start haha. I thought that was the case here since we're modifying the `cmd.envs`, but I see where I messed up with this filter fn. 

I'll give this another look over this weekend 👌

crates/gitbutler-tauri/src/menu.rs
ndom91: Hmm so I'd like to make it `open::url`, that does look and feel nice. However then the fn is also just called `url` from the point of view of `invoke` on the Svelte side which feels a bit too abstract. 

So I just imported the fn directly and referred to it as `open_url` instead of `open::open_url`. What do you think?

crates/gitbutler-tauri/src/open.rs
ndom91: Okie doke, I think I've got it squared away finally :pray:

crates/gitbutler-tauri/src/menu.rs
Byron: Now it's definitely looking better, and clearly, it's a nit, and often it comes down to the question if everything should be uniquely named if imported directly, or if things should be referred to by module and thus each piece of the path can be short.

Gitoxide definitely does the latter, whereas `git2` does the former, and I don't think I have an answer as to what's *always* the best.

crates/gitbutler-tauri/src/open.rs
```diff
@@ -0,0 +1,69 @@
...
+pub fn open_that(path: &str) -> Result<(), Error> {
+    let re = Regex::new(r"^((https://)|(http://)|(mailto:)|(vscode://)|(vscodium://)).+").unwrap();
+    if !re.is_match(path) {
+        return Err(anyhow!("Invalid path format").into());
+    }
+
+    fn clean_env_vars(var_names: &[&str]) -> Vec<(String, String)> {
+        let mut cleaned_vars = Vec::new();
+        for var_name in var_names {
+            if let Ok(var_value) = env::var(var_name) {
```
Byron: Just as a note: ideally one will always work with `env::var_os()` when paths are involved as UTF-8 can't always represent what's representable on the filesystem. It makes everything harder, too, so I wouldn't bother here also because we know these paths are probably just ASCII :D.
However, when dealing with paths in the belly of the application, one cannot resort to this.

crates/gitbutler-tauri/src/open.rs
```diff
@@ -0,0 +1,69 @@
...
+                    .filter(|path| !path.contains("appimage-run"))
+                    .collect::<Vec<_>>()
+                    .join(":");
+                cleaned_vars.push((var_name.to_string(), cleaned_value));
+            }
+        }
+        cleaned_vars
+    }
+
+    std::thread::spawn({
```
Byron: I would remove this behaviour - the function should block and be able to report an error if all commands failed to open the URL.

The non-blocking portion is handled when called through Tauri - it automatically launches it on a thread, and the UI can always deal with that using its own promises.

Correct error handling is definitely preferable here.

Byron (changes requested):
Thanks a lot for going the extra mile with me!

Now that I wasn't 'taken away' by the `std::env::set_var()` I found another issue that I think should be fixed. But it's not a strong as the previous one and I am happy to just select 'approve' if you would ask me to.

crates/gitbutler-tauri/src/open.rs
ndom91: Sounds good, will clean this up too. 

Just out of curiosity, Is it always the case that tauri handles the non-blocking stuff, or only when annotated with the async keyword in the tauri command macro?

Byron: It's correct, it's only with the `async` keyword in the command macro. Otherwise it would be blocking.
And now that you are asking, I don't see that macro anywhere and wasn't aware that any function can be exposed like apparently is happening here 😅.

Another thing I apparently missed - I think my lesson here is to not do armchair reviews, I need the IDE to really understand the whole context of the code.

ndom91: Haha isn't it at the bottom of the file here or am I missing something? 

I vaguely remember reading (was it you and kiril?) someone talk about the async keyword in the macro == nonblocking, so I wanted to be sure, becuase it's currently missing there. 

I'll add the async keyword and remove this threading 👌

Byron: Now that I have opened this PR in my IDE it's clear that… the macro exists but is indeed missing the `async` keyword.

I am surprised about the `unwrap()` there though.

crates/gitbutler-tauri/src/open.rs
Byron: Isn't this regex something that should be removed entirely, or did I misunderstand something?
It's surprising that this was deemed necessary as we control all the URLs that are passed.

Byron (commented):
Now that I opened it in my IDE I am tempted to push minor refactors, and hope that's alright as it's much easier to do that way.

crates/gitbutler-tauri/src/open.rs
ndom91: Hmm yeah so we had it previously as an additional layer of protection in the `shell-open` configuration. 

That reminds me, we still have 1 place actually that users can pass links and that's via PR descriptions / into our markdown renderer. It's currenlty still just creating an anchor tag, which won't work on linux/appimage, looking into that now

Byron: Thanks for elaborating! 
Alright, then it seems like the check should remain. Personally I find regex too flimsy and would definitely prefer 'harder' parsers here, as provided by the `url` crate.

Byron (commented):
I think the `regex` or `url` question shouldn't be a blocker here, so only the error-handling remains and avoiding `thread::spawn()` and adding `#[tauri::command(async)]`.

crates/gitbutler-tauri/src/open.rs
```diff
@@ -0,0 +1,71 @@
...
+            "PERLLIB",
+            "PYTHONHOME",
+            "PYTHONPATH",
+            "QT_PLUGIN_PATH",
+            "XDG_DATA_DIRS",
+        ]);
+
+        cmd.envs(cleaned_vars);
+        cmd.current_dir(env::temp_dir());
+        if cmd.status().is_ok() {
```
Byron, line 59: It should keep track of the last seen (or even all) errors and return that if no command succeeded.

crates/gitbutler-tauri/src/open.rs
```diff
@@ -0,0 +1,71 @@
...
+use anyhow::anyhow;
+use std::env;
+use tracing::instrument;
+use url::Url;
+
+pub(crate) fn open_that(path: &str) -> Result<(), Error> {
+    let target_url = match Url::parse(path) {
+        Ok(url) => url,
+        Err(_) => return Err(anyhow!("Invalid path format").into()),
+    };
```
Byron:
```suggestion
    let target_url = Url::parse(path).with_context(|| format!("Invalid path format: '{path}'")?;
```

crates/gitbutler-tauri/src/open.rs
```diff
@@ -0,0 +1,71 @@
...
+    if !["http", "https", "mailto", "vscode", "vscodium"].contains(&target_url.scheme()) {
+        return Err(anyhow!("Invalid path scheme").into());
```
Byron:
```suggestion
        anyhow::bail!("Invalid path scheme: {}", target_url.scheme());
```

Byron (commented):
Thanks! I think it's nearly there.

The suggestions are best not applied directly as I wrote them in the browser, they might not compile. There is also the answer to the error-handling question, I hope it makes sense.

crates/gitbutler-tauri/src/open.rs
ndom91: I tried it initially with the question mark syntax, but it complained about the potnential error not matching some frontend error type, is that avoided by adding some context here?

Byron: Right, that reminds me of another issue regarding the error types used here.
I have pushed a fix for my recent suggestions, leaving only the error handling part.

ndom91: Great, thanks!

crates/gitbutler-tauri/src/open.rs
```diff
@@ -0,0 +1,75 @@
...
+            "QT_PLUGIN_PATH",
+            "XDG_DATA_DIRS",
+        ]);
+
+        cmd.envs(cleaned_vars);
+        cmd.current_dir(env::temp_dir());
+        if cmd.status().is_ok() {
+            return Ok(());
+        } else {
+            cmd_errors.push(anyhow::anyhow!("Failed to execute command {:?}", cmd));
```
Byron, line 62: Good call, I think that will help to reproduce issues on systems that fail for all possible commands.

Byron (approved):
Thanks a lot, and sorry for holding it for so long!

Let's merge this :).

PR #4749, opened 2024-08-24
update git2-hooks to 0.3.3 to use default login shell instead just bash

### LINKED ISSUE #4569: Use default login shell when running git-hooks
We use git2-hooks (gitui) [here](https://github.com/gitbutlerapp/gitbutler/blob/master/crates/gitbutler-branch-actions/src/virtual.rs#L877) which in turn [runs hooks with bash](https://github.com/extrawurst/gitui/blob/4ef633a75a3581936d0da9aade7af8ed7aad98ab/git2-hooks/src/hookspath.rs#L116-L117). This is an issue for most mac users where the default shell is `zsh`. The hooks will run, but sourcing `.bashrc` rather than `.zshrc`.

yerke:
## ☕️ Reasoning
Fixes https://github.com/gitbutlerapp/gitbutler/issues/4569

## 🧢 Changes
Update git2-hooks to 0.3.3. That version includes https://github.com/extrawurst/gitui/pull/2343.

## 🎫 Affected issues

Fixes: 4569

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2080,9 +2080,9 @@ dependencies = [
 
 [[package]]
 name = "git2-hooks"
-version = "0.3.1"
+version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9dbf7f731ce1bcf0c6578738cb6cad3232eca69c0508da3d084b5cf95e3a38bf"
+checksum = "3e010ee9804b082020b3aa9e39a32b11595e7b46185b55b509948217e847203f"
 dependencies = [
  "git2",
  "log",


Cargo.lock
```diff
@@ -6798,7 +6798,7 @@ version = "3.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "da03fa3b94cc19e3ebfc88c4229c49d8f08cdbd1228870a45f0ffdf84988e14b"
 dependencies = [
- "dirs 5.0.1",
+ "dirs 4.0.0",
```
mtsgrd: Hmm, this downgrade looks a bit weird? `gitui` depends on 5.0.

yerke: Hmm, I am not sure why cargo decided to resolve to that version when I ran `cargo update git2-hooks`. `shellexpand` has `dirs = { version = ">=4, <6", optional = true }`, and dirs 4 in Cargo.lock comes from `resolve-path`.
Removed this change.

krlvi (approved):
thanks for fixing this upstream in git2_hooks <3

PR #4760, opened 2024-08-26
Show files in edit mode

Caleb-T-Owens:
This PR jiggles some code about, and adds a list of files in edit mode.


crates/gitbutler-branch-actions/src/branch.rs
```diff
@@ -21,6 +23,33 @@ use std::{
...
+    _permission: &WorktreeReadPermission,
+) -> Result<Vec<RemoteBranchFile>> {
+    let repository = context.repository();
+    let head_commit = repository
+        .head()
+        .context("Failed to get head")?
+        .peel_to_commit()
+        .context("Failed to get head commit")?;
+
+    let files = workdir(repository, &head_commit.id())
```
Byron: I think this really should remain as `gitbutler_diff::workdir()` so it's clear(er) that it's a diff.

Byron (commented):
Very interesting work - I took a cursory look just to see what eventually has to be done by `gix` :D.

apps/desktop/src/lib/components/EditMode.svelte
```diff
@@ -14,16 +18,95 @@
 	const { editModeMetadata }: Props = $props();
 
 	const project = getContext(Project);
+	const uncommitedFileWatcher = getContext(UncommitedFilesWatcher);
 	const modeService = getContext(ModeService);
 
+	const uncommitedFiles = uncommitedFileWatcher.uncommitedFiles;
+
+	let modeServiceAborting = $state<'inert' | 'loading' | 'completed'>('inert');
```
ndom91, line 26: Nit pick: I'd love these to be extracted to an enum / as const object, same with the `outputMap.status` field, instead of "magic" strings of `'D', 'M', 'A'`, ya know

ndom91 (approved):
In general LGTM. I like the `modified`, etc. badges on the changed files while a user is in the edit mode cover view. 

Maybe as a future follow up, we could also allow folks to click on that file to open a diff, just like we do in the pre-existing filelist hunk view.

PR #4762, opened 2024-08-26
Update app updater button copy

mtsgrd:


mtsgrd:
@PavelLaptev this would be more accurate, what do you think? Download and install happen together, so currently, if you abort after download then the app has already been updated.
image

PavelLaptev:
Looks good to me 🙂

PavelLaptev:
@mtsgrd sorry for the late thoughts, but how about using 'Update to 0.5.692' to keep the copy shorter?

mtsgrd:
I like it.

PavelLaptev (approved):

PavelLaptev:
> I like it.

Cool :-) I'll leave it to you then

Issue #4734, opened 2024-08-21
New Keyboard Shortcuts


ndom91:
We're looking for input on some more keyboard shortcuts. We want to align with other similar apps and implement keyboard shortcuts that will be familiar to users. We're currently considering adding the following:

- [ ] `ctrl + a` to select all files in filelist in commit mode
- [ ] `space` to toggle select/unselect on focused file
- [ ] `esc` to close an opened file diff preview
- [ ] `ctrl + ?` to open a keyboard shortcut overview menu

Also potentially `ctrl + z/y` to travel back and forth in the oplog aka undo stack. 

Please add any other ideas below :grin:

freak4pc:
Just wanted to report that using up/down keys to navigate between files doesn't work - not sure if it's planned to not work or it's a bug, but that's the behavior in other Git UI tools :)

ndom91:
Thanks for chiming in! Yeah those sound sensible as well 👌

freak4pc:
> > Just wanted to report that using the up/down keys to navigate between files doesn't work. Not sure if it's intentional or a bug, but that's how it works in other Git UI tools. :)
> 
> @freak4pc This is indeed a bug. The functionality should be working, so I'm wondering why it isn't on your end. If possible, could you please create a separate issue for this? You can do so using [this link](https://github.com/gitbutlerapp/gitbutler/issues/new?assignees=&labels=bug&projects=&template=1_bug.yml).
> 
>  Screen.Recording.2024-08-26.at.15.22.52.mov

Thanks, I actually meant when being focused on a specific commit, not on the staged workspace changes

PavelLaptev:
> Thanks, I actually meant when being focused on a specific commit, not on the staged workspace changes

@freak4pc, thank you for clarifying this. Yes, it doesn’t work at the moment. It's something we need to fix.

H1ghBre4k3r:
I am currently thinking if hitting `esc` while in the settings (both, project and app settings) should bring you back to the previous screen from where you opened the settings (similar to the small arrow on the upper left hand side).

ndom91:
Note for ourselves for the future: Navigating `up`/`down` with the keyboard arrows in the Filelist sometimes behaves very strangely. For example, skipping items, stopping at "incorrect" items, etc.

PavelLaptev:
[...]

PR #4764, opened 2024-08-26
Fix files selection

PavelLaptev:
## ☕️ Reasoning

- The selection using arrow keys didn't work on commits. It's because we didn't provide the `commit id` and there wasn't any matches.
- Allow to select multiple files, except `remote` commits
 - disable `draggable` icon for unapplied commits

Result:

https://github.com/user-attachments/assets/3bfe924b-3ca7-41a7-9f51-767f977bed87

Related issue — https://github.com/gitbutlerapp/gitbutler/issues/4734#issuecomment-2307668787

### DIFF
--- a/apps/desktop/src/lib/utils/selection.ts
+++ b/apps/desktop/src/lib/utils/selection.ts
@@ -8,8 +8,8 @@ import type { AnyFile } from '$lib/vbranches/types';
 
 export function getNextFile(files: AnyFile[], currentId: string): AnyFile | undefined {
 	const fileIndex = files.findIndex((f) => f.id === currentId);
-	const nextFile =
-		fileIndex !== -1 && fileIndex + 1 < files.length ? files[fileIndex + 1] : undefined;
+	const nextFile = fileIndex > 0 && fileIndex + 1 < files.length ? files[fileIndex + 1] : undefined;
+	console.log('nextFile', nextFile);
 	return nextFile;
 }
 
--- a/apps/desktop/src/lib/utils/selection.ts
+++ b/apps/desktop/src/lib/utils/selection.ts
@@ -8,12 +8,15 @@ import type { AnyFile } from '$lib/vbranches/types';
 
 export function getNextFile(files: AnyFile[], currentId: string): AnyFile | undefined {
 	const fileIndex = files.findIndex((f) => f.id === currentId);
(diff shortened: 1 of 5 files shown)


krlvi (approved):

apps/desktop/src/lib/utils/selection.ts
```diff
@@ -8,12 +8,15 @@ import type { AnyFile } from '$lib/vbranches/types';
 
 export function getNextFile(files: AnyFile[], currentId: string): AnyFile | undefined {
 	const fileIndex = files.findIndex((f) => f.id === currentId);
-	return fileIndex !== -1 && fileIndex + 1 < files.length ? files[fileIndex + 1] : undefined;
+	const nextFile =
+		fileIndex !== -1 && fileIndex + 1 < files.length ? files[fileIndex + 1] : undefined;
+	return nextFile;
```
mtsgrd: It looks like this is a no-op change?

PavelLaptev: I'm checking if fileIndex is greater than 0 and within the bounds of the files array. If both conditions are true, the code returns the next file in the array. Otherwise, it returns undefined.

Can you suggest a better solution here?

mtsgrd: The only change you're making here is declaring the variable `nextFile` and then immediately removing it. Perhaps you meant to modify the expression?
image

apps/desktop/src/lib/utils/selection.ts
```diff
@@ -8,12 +8,15 @@ import type { AnyFile } from '$lib/vbranches/types';
...
 }
 
 export function getPreviousFile(files: AnyFile[], currentId: string): AnyFile | undefined {
 	const fileIndex = files.findIndex((f) => f.id === currentId);
-	return fileIndex > 0 ? files[fileIndex - 1] : undefined;
+	const previousFile = fileIndex > 0 ? files[fileIndex - 1] : undefined;
+	return previousFile;
```
mtsgrd: Same thing here!

PavelLaptev: Fixed :-)

apps/desktop/src/lib/utils/selection.ts
PavelLaptev: Ah, I see. Sure, I'll fixed that

PR #4765, opened 2024-08-27
merge-base performance

Byron:
Follow-up to #4757.

Get `merge-base` computation performance to be en-par with the cache-less variant in `git2`.
Prevously when there is no commit-graph cache, it was 7% slower than `git2` (which is just as fast as Git itself),
which is likely due to Git2 leveraging already parsed objects between multiple merge-base queries.

`gix` has to learn that to be en-par or faster, and I think it's not going to take long.

### Tasks

* [x] move rev-walking to the thread that deals with commits walks to leverage caches (from 6% slower to 4% slower)
* [x] performance comparison compared to the previous version that used `git2` for merge-bases. (from 6% slower to 8% faster)
* [x] use [latest version of `gix`](https://github.com/Byron/gitoxide/pull/1564) (that can re-use the existing commit-graph cache (with parsed objects)

### Performance

This was a win, as `gix` is now about 4% faster than Git (in one particular workload without commit-graph), and *8% faster* than `git2` thanks to greatly improved caching.
Memory consumption is still lower at 575MB compared to 617MB.

When a commit-graph is present (via `git commit-graph write --reachable`), then `gix` is 26% faster. Now it probably hits the boundary where diffing trees is the slowest operation, whereas merge-base computation is very quickly done (it's at least 10x faster then). Those who have repos big enough to use commit-graphs will certainly appreciate it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2629,32 +2629,32 @@ dependencies = [
 [[package]]
 name = "gix"
 version = "0.66.0"
-source = "git+https://github.com/Byron/gitoxide?rev=649f5882cbebadf1133fa5f310e09b4aab77217e#649f5882cbebadf1133fa5f310e09b4aab77217e"
+source = "git+https://github.com/Byron/gitoxide?rev=0fe5133598c6f843fb3172a4e0c4f58932405647#0fe5133598c6f843fb3172a4e0c4f58932405647"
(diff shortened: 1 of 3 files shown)


krlvi (approved):
nice improvements indeed :)

PR #4770, opened 2024-08-27
fix: dont hide "discard file" contextMenu option from binary files

ndom91:
## ☕️ Reasoning

- Right click "Discard Item" was disabled for Binary files. I can't see any reason why that should be, but maybe I'm missing something? CC: @krlvi 

## 🧢 Changes

- Always allow discarding items from the file list


krlvi:
@ndom91 it used to not function. does it work now?

ndom91:
> @ndom91 it used to not function. does it work now?

Yeah seems to work without issue now :+1:

krlvi (approved):

PR #4773, opened 2024-08-27
build: add a .vercelignore to ignore all files in crates

krlvi:


ndom91:
Just trying to avoid another dotfile in the root of the repository? :joy:

Otherwise, I feel like it should go there with content like this: `/crates/*`

krlvi:
i am just trying rid us of the vercel spam... technically, it shouldn't get triggered on rust files, according to the config we got but doesnt seem to be workign

ndom91:
> i am just trying rid us of the vercel spam... technically, it shouldn't get triggered on rust files, according to the config we got but doesnt seem to be workign

Yeah Mattias and I have talked about this briefly, it'll attempt to deploy every branch once for the first commit of each branch. That's because when running the `turbo-ignore` CLI on the first commit, it doesn't have a commit to compare to apparently according to its output. It needs a fallback ref. I'll add that now :+1:

PR #4775, opened 2024-08-27
Set gitHostFactory for gitlab self hosted instances

### LINKED ISSUE #4558: Branch button no longer opens branch in browser
### Version

0.12.16

### Operating System

Mac OS X

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

Usually the button `origin/<branch-name>` would open a new tab on my browser to my branch inside of my repository. Now it seems to no longer work.

Clicking on the button doesn't do anything. Maybe coming from the last update or the one before that. Only notice it this week.

Screenshot 2024-07-31 at 15 00 49

### How to reproduce

Apply branch that has an remote branch and try to click the button.

### Expected behavior

Clicking on the button should open a web page to the repository in the correct branch

### Relevant log output

_No response_

fgLouro:
## ☕️ Reasoning
The `DefaultGitHostFactory` class assumes that gitlab domain is always `gitlab.com`, which is not correct for self hosted instances of gitlab, and leads to branch link buttons to be unusable. 

## 🧢 Changes
- Adds `resource` to `RepoInfo` type to be used for adding the correct `baseUrl`
- Checks `resource` for gitlab subdomain to initialize the `gitHostFactory` with a Gitlab class

## 🎫 Affected issues

Fixes: https://github.com/gitbutlerapp/gitbutler/issues/4558#issuecomment-2293337474

## Things to consider
Self hosted instance might not have `gitlab` in their subdomain, which will result in the same issue.

### DIFF
--- a/apps/desktop/src/lib/gitHost/gitHostFactory.ts
+++ b/apps/desktop/src/lib/gitHost/gitHostFactory.ts
@@ -23,11 +23,10 @@ export class DefaultGitHostFactory implements GitHostFactory {
 		fork?: RepoInfo,
 		usePullRequestTemplate?: Persisted<boolean>
 	) {
-		const source = repo.source;
-		const resource = repo.resource;
+		const domain = repo.domain;
 		const forkStr = fork ? `${fork.owner}:${fork.name}` : undefined;
 
-		if (source.includes(GITHUB_DOMAIN)) {
+		if (domain.includes(GITHUB_DOMAIN)) {
 			return new GitHub({
 				repo,
 				baseBranch,
@@ -37,13 +36,13 @@ export class DefaultGitHostFactory implements GitHostFactory {
 				usePullRequestTemplate
 			});
 		}
-		if (source === GITLAB_DOMAIN || resource.startsWith(GITLAB_SUB_DOMAIN + '.')) {
+		if (domain === GITLAB_DOMAIN || domain.startsWith(GITLAB_SUB_DOMAIN + '.')) {
 			return new GitLab({ repo, baseBranch, forkStr });
 		}
-		if (source.includes(BITBUCKET_DOMAIN)) {
+		if (domain.includes(BITBUCKET_DOMAIN)) {
 			return new BitBucket({ repo, baseBranch, forkStr });
 		}
-		if (source.includes(AZURE_DOMAIN)) {
+		if (domain.includes(AZURE_DOMAIN)) {
 			return new AzureDevOps({ repo, baseBranch, forkStr });
 		}
 	}
--- a/apps/desktop/src/lib/gitHost/gitHostFactory.ts
+++ b/apps/desktop/src/lib/gitHost/gitHostFactory.ts
@@ -37,7 +37,7 @@ export class DefaultGitHostFactory implements GitHostFactory {
 				usePullRequestTemplate
 			});
 		}
-		if (source === GITLAB_DOMAIN || resource.startsWith(GITLAB_SUB_DOMAIN + ".")) {
+		if (source === GITLAB_DOMAIN || resource.startsWith(GITLAB_SUB_DOMAIN + '.')) {
 			return new GitLab({ repo, baseBranch, forkStr });
 		}
 		if (source.includes(BITBUCKET_DOMAIN)) {
--- a/apps/desktop/src/lib/gitHost/gitHostFactory.ts
+++ b/apps/desktop/src/lib/gitHost/gitHostFactory.ts
@@ -1,7 +1,7 @@
 import { AZURE_DOMAIN, AzureDevOps } from './azure/azure';
 import { BitBucket, BITBUCKET_DOMAIN } from './bitbucket/bitbucket';
 import { GitHub, GITHUB_DOMAIN } from './github/github';
-import { GitLab, GITLAB_DOMAIN } from './gitlab/gitlab';
+import { GitLab, GITLAB_DOMAIN, GITLAB_SUB_DOMAIN } from './gitlab/gitlab';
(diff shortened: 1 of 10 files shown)


apps/desktop/src/lib/gitHost/gitHostFactory.ts
```diff
@@ -36,7 +37,7 @@ export class DefaultGitHostFactory implements GitHostFactory {
 				usePullRequestTemplate
 			});
 		}
-		if (source.includes(GITLAB_DOMAIN)) {
+		if (source.includes(GITLAB_DOMAIN) || resource.includes(GITLAB_SUB_DOMAIN)) {
```
mtsgrd: Thanks for contributing! :)

Would it also be a good idea to make it stricter like this?
```ts
if (source === GITLAB_DOMAIN || resource.startsWith(GITLAB_SUB_DOMAIN + ".")) {
```
The current code is incorrect in that e.g. `agitlab.com` would be recognised as GitLab.

fgLouro: Makes sense

apps/desktop/src/lib/gitHost/gitlab/gitlab.ts
```diff
@@ -1,13 +1,14 @@
-import { GitLabBranch } from './gitlabBranch';
 import type { RepoInfo } from '$lib/url/gitUrl';
 import type { GitHost } from '../interface/gitHost';
 import type { DetailedPullRequest, GitHostArguments } from '../interface/types';
+import { GitLabBranch } from './gitlabBranch';
```
ndom91, line 4: Well this now assumes that folks are putting their Gitlab instance at `gitlab.company.com` always, right?

What do you think about makign this a user setting? @mtsgrd

apps/desktop/src/lib/gitHost/gitlab/gitlab.ts
```diff
@@ -1,13 +1,14 @@
...
 
 export type PrAction = 'creating_pr';
 export type PrState = { busy: boolean; branchId: string; action?: PrAction };
 export type PrCacheKey = { value: DetailedPullRequest | undefined; fetchedAt: Date };
 
 export const GITLAB_DOMAIN = 'gitlab.com';
+export const GITLAB_SUB_DOMAIN = 'gitlab'; // For self hosted instance of Gitlab
```
ndom91, line 11: Well this now assumes that folks are putting their Gitlab instance at `gitlab.company.com` always, right?

What do you think about makign this a user setting? @mtsgrd

mtsgrd: I was thinking this is a change in the right direction, even if we want to make it explicitly configurable? Aside from that, I do think we should start prioritising integration work, both building out our GH support, as well as building real integrations for the rest.

mtsgrd (approved):

apps/desktop/src/lib/gitHost/gitlab/gitlab.ts
fgLouro: > Well this now assumes that folks are putting their Gitlab instance at gitlab.company.com always, right?

Yes, this is one issue. The better solution would be to have some setting where the user could define this in case the git url is different.

apps/desktop/src/lib/gitHost/github/githubBranch.test.ts
```diff
@@ -8,7 +8,8 @@ describe.concurrent('GitHubBranch', () => {
 	const repo = {
 		source: 'github.com',
 		name: 'test-repo',
-		owner: 'test-owner'
+		owner: 'test-owner',
+		resource: 'test-resource'
```
mtsgrd: Let's not add these unless necessary for the test to pass?

apps/desktop/src/lib/url/gitUrl.ts
```diff
@@ -4,20 +4,22 @@ export type RepoInfo = {
 	source: string;
 	name: string;
 	owner: string;
+	resource: string;
```
mtsgrd: And make this optional?

fgLouro: Wouldn't it be misleading to use an optional attribute for the `resource`? 

The `gitUrlParse` function seems to always return a `resource` attribute when the url is valid, even if it's an empty 'resource' it's not undefined.

mtsgrd: Sorry, I forgot what it represents, and sorry for the back and forth here. I think.. we should only have one field for url, instead of calling it source and resource. If you agree, would you mind creating a type and replacing the one we borrow from the parsing lib?

fgLouro: Back and forth is how good software is made 😄 

You mean removing the `resource` and `source` from the `RepoInfo` type and have only a `url` attribute which will be the value of `resource` (the full url)?

I think the `source` is only used in `DefaultGitHostFactory` class, so should be easy to replace it.

mtsgrd: Yes, exactly! I would probably go with `domain` over `url` since it is likely to have additional things like scheme, path, etc. Also, my previous message was probably confusing because I forgot we have our own `RepoInfo` type, I thought we were using the one returned from `git-url-parse`.

fgLouro: @mtsgrd Done!

Removed all references of `source` and `resource` in favour or using domain.

mtsgrd (approved):
A'ight, awesome. Let's ship it!

PR #4781, opened 2024-08-28
Fix snapshot details transformer

Caleb-T-Owens:


krlvi:
i think the e2e tests are flakey

krlvi (approved):

Issue #4694, opened 2024-08-15
Migrate to Tauri 2.0


krlvi:

ndom91:
@mtsgrd yeah checking it out now :+1:

mtsgrd:
Cool cool, so the `NotImplemented` error only happens for the Actions API on webkit2gtk-4.0 and not 4.1. The former is really starting to grind my gears.

Issue #4128, opened 2024-06-20
Distribute as VSCode Extension


Byron:
There are many open questions about this form of distribution.

### Open Questions

- How do language servers deal with installing/managing their binaries on the various platforms?
- Is this already a known path that Tauri apps can take? Are there examples of applications that already use this distribution mechanism? Prior art?

merlinseela:
issue origin: #2605

PR #4789, opened 2024-08-29
Tauri v1 -> v2

mtsgrd:
- rewrite of app updater because of new api
- updated persisted store because of new api
- webkit2gtk 4.0 -> 4.1 upgrade, will now work for appimage on Ubuntu 24.04
- webdriver actions API now working


honnip:
Part of #4694

mtsgrd:
@ndom91 I think this works pretty well, would you mind giving it a test-run today?

The only thing currently on my mind are the json files under `crates/gitbutler-tauri/capabilities`. One of them was automatically created by the migration script, we should probably move its contents over into the other one.

ndom91:
@mtsgrd yeah so I was able to fire it up in dev mode no problem after applyign that small fix from my commit ^^ :+1: 

Regarding the capabilities, yeah so that migration one seems to just add a `core:default` permission. I couldn't find _exactly_ what that enables/allows in their docs, but it seems like, well, a core one :sweat_smile: that we probably can just move over to the `main.json` as you mentioned :+1:

ndom91:
A quick note on the tests - they started up locally for me without a problem, and after remembering to upgrade to `tauri-driver@0.1.5` they also ran without issue with your updated `findAndClick()` :partying_face:

crates/gitbutler-tauri/tauri.conf.nightly.json
```diff
@@ -18,10 +18,10 @@
 		},
 		"linux": {
 			"rpm": {
-				"depends": ["webkit2gtk4.0-devel"]
+				"depends": ["webkit2gtk4.1-devel"]
 			},
 			"deb": {
-				"depends": ["libwebkit2gtk-4.0-dev", "libgtk-3-dev"]
+				"depends": ["libwebkit2gtk-4.1-dev", "libgtk-3-dev"]
```
mtsgrd: Perfect, thank you.

crates/gitbutler-tauri/capabilities/main.json
```diff
@@ -1,14 +1,16 @@
 {
 	"$schema": "../gen/schemas/desktop-schema.json",
-	"identifier": "main-capability",
-	"description": "Figuring things out",
+	"identifier": "main",
+	"description": "permissions for gitbutler tauri",
```
mtsgrd, line 4: 👍

crates/gitbutler-tauri/tauri.conf.nightly.json
```diff
@@ -27,8 +27,6 @@
 	},
 	"plugins": {
 		"updater": {
-			"active": true,
-			"dialog": false,
```
mtsgrd: Does this disable the updater for nightly builds?

ndom91: No these fields were just removed apparently, see: https://v2.tauri.app/start/migrate/from-tauri-1/#tauri-configuration

.github/workflows/publish.yaml
```diff
@@ -1,10 +1,10 @@
-name: "Publish"
+name: 'Publish'
```
mtsgrd: I will undo these formatting changes.

apps/desktop/src/lib/secrets/secretsService.ts
```diff
@@ -1,13 +1,7 @@
-import { AISecretHandle } from '$lib/ai/service';
 import { invoke } from '$lib/backend/ipc';
 import { buildContext } from '$lib/utils/context';
 import type { GitConfigService } from '$lib/backend/gitConfigService';
 
-const MIGRATION_HANDLES = [
-	AISecretHandle.AnthropicKey.toString(),
-	AISecretHandle.OpenAIKey.toString()
-];
-
```
mtsgrd, line 10: These removals do not belong in this pr, will remove.

crates/gitbutler-tauri/Cargo.toml
```diff
@@ -69,19 +77,10 @@ gitbutler-edit-mode.workspace = true
 open = "5"
 
 [dependencies.tauri]
-version = "1.7.0"
+version = "2.0.0-rc"
```
mtsgrd: Looks like this should be `2.0.0-rc.10`.

apps/desktop/src/lib/ai/service.test.ts
```diff
@@ -1,15 +1,15 @@
-import { AnthropicAIClient } from '$lib/ai/anthropicClient';
-import { ButlerAIClient } from '$lib/ai/butlerClient';
-import { OpenAIClient } from '$lib/ai/openAIClient';
-import { SHORT_DEFAULT_BRANCH_TEMPLATE, SHORT_DEFAULT_COMMIT_TEMPLATE } from '$lib/ai/prompts';
-import { AISecretHandle, AIService, GitAIConfigKey, KeyOption, buildDiff } from '$lib/ai/service';
```
Caleb-T-Owens: Can we discuss the use of relative imports on Monday? I'm very much not a fan of them.

crates/gitbutler-tauri/Cargo.toml
honnip: I believe it's a valid semver

mtsgrd: It is, but atm we prefer pinned versions.

honnip: I think it should be `=2.0.0-rc.10` right?

mtsgrd: Are you saying the rc versions are treated as patches and therefore need the added equal sign to force exact matching?

apps/desktop/src/lib/ai/service.test.ts
mtsgrd: The only exception I make, and subjectively so, is that relative imports are ok for the same package, i.e. `./file.ts`. When it comes to any other import I prefer absolute. What's your view?

Caleb-T-Owens: I would always prefer to use absolute imports. I find it clearer to see what is getting included in files. Always using the absolute imports has a second advantage of making it easier to find and update all the places where a given file is imported. For example, after a rename or large move with the LSP failed to track (like when moving things to the UI library).

crates/gitbutler-tauri/Cargo.toml
honnip: yes, otherwise it will include `2.0.0-rc.11` and even `2.0.0`

Byron:
I just noticed that #4128 (Distribute as VSCode extension) has a stunning amount of upvotes, and I thought I'd mention it here in case this PR in any way facilitates getting there.

ndom91:
> I just noticed that #4128 (Distribute as VSCode extension) has a stunning amount of upvotes, and I thought I'd mention it here in case this PR in any way facilitates getting there.

While there doesn't seem to be any official tauri-supported way to build and package a tauri code base into a VS Code extension. We could, hwoever, continue with the Tauri-based CLI and package that up with a VS Code extension. A VS Code extension could then just be a "frontend" to the CLI

Byron:
> We could, hwoever, continue with the Tauri-based CLI and package that up with a VS Code extension. Hypothetically, the VS Code extension could then just be a "frontend" to the CLI

The CLI (as it exists currently) doesn't have any `tauri` in it, which probably helps with portability. Indeed, it sounds like currently one would have to rewrite the frontend for VSCode, which seems prohibitive.
However, from a technical standpoint I think `tauri` would be in the best position to enable vscode support. After all, it 'just' wouldn't have to open its own window. But that's for the future then, and maybe something to talk about next time Daniel is in town.

Caleb-T-Owens:
> While there doesn't seem to be any official tauri-supported way to build and package a tauri code base into a VS Code extension. We could, hwoever, continue with the Tauri-based CLI and package that up with a VS Code extension. Hypothetically, the VS Code extension could then just be a "frontend" to the CLI

> The CLI (as it exists currently) doesn't have any tauri in it, which probably helps with portability. Indeed, it sounds like currently one would have to rewrite the frontend for VSCode, which seems prohibitive.
However, from a technical standpoint I think tauri would be in the best position to enable vscode support. After all, it 'just' wouldn't have to open its own window. But that's for the future then, and maybe something to talk about next time Daniel is in town.

A while back we had some conversations about wanting to support running a gitbutler server which could then talk to a gitbutler client over web socket or HTTPS. I've not read too much into the context of this conversation, but in my mind, I would want to approach anything that could create more lock-in to tauri's IPC or similar systems with caution.

crates/gitbutler-tauri/tauri.conf.release.json
```diff
@@ -1,55 +1,55 @@
 {
+	"productName": "GitButler",
+	"identifier": "com.gitbutler.app",
 	"build": {
-		"beforeBuildCommand": "[ $CI = true ] || pnpm build:desktop -- --mode production && cargo build --release -p gitbutler-git && bash ./gitbutler-tauri/inject-git-binaries.sh"
+		"beforeBuildCommand": "[ $CI = true ] || pnpm build:desktop -- --mode production && cargo build --release -p gitbutler-git && bash ./gitbutler-tauri/inject-git-binaries.sh && bash ./crates/gitbutler-tauri/inject-relauncher.sh GitButler"
```
TrueBrain:
```suggestion
		"beforeBuildCommand": "[ $CI = true ] || pnpm build:desktop -- --mode production && cargo build --release -p gitbutler-git && bash ./creates/gitbutler-tauri/inject-git-binaries.sh && bash ./crates/gitbutler-tauri/inject-relauncher.sh GitButler"
```

I ran into this by accident. The `nightly` variant does have this change, but the `release` doesn't. In result it failed to build locally for me. Your Miles May Vary.

mtsgrd: Hey, thanks for spotting that. I actually noticed it, and there's a couple more changes I need to get in now that Tauri v2 has been released. We should be merging in this PR in the coming week!

crates/gitbutler-tauri/desktopTemplate.hbs
```diff
@@ -0,0 +1,8 @@
+[Desktop Entry]
```
ndom91: This extra desktop entry template file (and the line pointing to it in the various `tauri.conf.json`'s) is no longer necessary. 

This was only used as a temporary workaround to force the app to use x11, as opposed to wayland, because there was a bug with tauri when run in a gnome+wayland environment in which it was impossible to resize the window and click+drag the window on its title bar. 

However this underlying bug seems to have been fixed in the latest `tao` version, released as of ~2 days ago. 

See: https://github.com/tauri-apps/tauri/issues/10686#issuecomment-2392720250

apps/desktop/src/lib/components/ProjectSetup.svelte
```diff
@@ -27,7 +27,7 @@
 		loading = true;
 		try {
 			// TODO: Refactor temporary solution to forcing Windows to use system executable
-			if ($platformName === 'win32') {
+			if (platformName === 'windows') {
```
krlvi, line 30: have we tested that this works?

apps/desktop/src/lib/config/appSettings.ts
```diff
@@ -2,99 +2,87 @@
...
+	 * Persisted confirmation that user has confirmed their analytics settings.
+	 */
+	readonly appAnalyticsConfirmed = this.persisted(false, 'appAnalyticsConfirmed');
+
+	/**
+	 * Provides a writable store for obtaining or setting the current state of application metrics.
+	 * The application metrics can be enabled or disabled by setting the value of the store to true or false.
+	 * @returns A writable store with the appMetricsEnabled config.
+	 */
+	readonly appMetricsEnabled = this.persisted(true, 'appMetricsEnabled');
```
krlvi, line 29: these still work correctly, right?

apps/desktop/src/lib/secrets/secretsService.ts
krlvi: it's still here

crates/gitbutler-tauri/Cargo.toml
```diff
@@ -37,11 +38,19 @@ once_cell = "1.20"
 reqwest = { version = "0.12.8", features = ["json"] }
 serde.workspace = true
 serde_json = { version = "1.0", features = ["std", "arbitrary_precision"] }
-tauri-plugin-context-menu = { git = "https://github.com/c2r0b/tauri-plugin-context-menu", branch = "main" }
-tauri-plugin-single-instance = { git = "https://github.com/tauri-apps/plugins-workspace", branch = "v1" }
-tauri-plugin-window-state = { git = "https://github.com/tauri-apps/plugins-workspace", branch = "v1" }
-tauri-plugin-store = { git = "https://github.com/tauri-apps/plugins-workspace", branch = "v1" }
-tauri-plugin-log = { git = "https://github.com/tauri-apps/plugins-workspace", branch = "v1" }
+tauri = { version = "^2.0.6", features = ["unstable"] }
+tauri-plugin-clipboard-manager = "2.0.1"
```
krlvi: where is the clipboard-manager used?

crates/gitbutler-tauri/tauri.conf.nightly.json
```diff
@@ -1,51 +1,22 @@
 {
+	"productName": "GitButler Nightly",
+	"identifier": "com.gitbutler.app.nightly",
 	"build": {
-		"beforeBuildCommand": "[ $CI = true ] || pnpm build:desktop -- --mode nightly && cargo build --release -p gitbutler-git && bash ./gitbutler-tauri/inject-git-binaries.sh"
+		"beforeBuildCommand": "[ $CI = true ] || pnpm build:desktop -- --mode nightly && cargo build --release -p gitbutler-git && bash ./crates/gitbutler-tauri/inject-git-binaries.sh"
```
krlvi, line 5: why the change? is this ran from a different place?

krlvi: also - why different from release?

scripts/release.sh
```diff
@@ -111,27 +111,22 @@ done
...
 	if [ "$OS" = "macos" ]; then
 		[ -z "${APPLE_CERTIFICATE-}" ] && error "$APPLE_CERTIFICATE is not set"
 		[ -z "${APPLE_CERTIFICATE_PASSWORD-}" ] && error "$APPLE_CERTIFICATE_PASSWORD is not set"
-		[ -z "${APPLE_SIGNING_IDENTITY-}" ] && error "$APPLE_SIGNING_IDENTITY is not set"
 		[ -z "${APPLE_ID-}" ] && error "$APPLE_ID is not set"
 		[ -z "${APPLE_TEAM_ID-}" ] && error "$APPLE_TEAM_ID is not set"
 		[ -z "${APPLE_PASSWORD-}" ] && error "$APPLE_PASSWORD is not set"
 		export APPLE_CERTIFICATE="$APPLE_CERTIFICATE"
 		export APPLE_CERTIFICATE_PASSWORD="$APPLE_CERTIFICATE_PASSWORD"
-		export APPLE_SIGNING_IDENTITY="$APPLE_SIGNING_IDENTITY"
```
krlvi, line 134: we dont need this anymore?

mtsgrd: I remember nightly failed to build unless I removed it, but I think we should investigate further.

https://github.com/tauri-apps/tauri/blob/a8105eccb2122f003b4a36bbfb747998f53c5f10/crates/tauri-cli/src/interface/rust.rs#L1299-L1306

apps/desktop/src/lib/config/appSettings.ts
mtsgrd: yes, but let me also test this manually before landing these changes.

apps/desktop/src/lib/components/ProjectSetup.svelte
mtsgrd: They are enumerated so typescript complains if you are using an incorrect value.

crates/gitbutler-tauri/Cargo.toml
mtsgrd: nice catch. looks like it's not used anywhere, removing it.

apps/desktop/src/lib/config/appSettings.ts
mtsgrd: I need to investigate this before we merge.

krlvi (approved):

PR #4771, opened 2024-08-27
list virtual branches performance

Byron:
An early PR to capture the research and first steps towards making `list_virtual_branches` faster by introducing `gix`.
Refactoring might happen here as well.

### Tasks

* [x] `list_virtual_branches` for CLI
* [x] figure out the best target for `list_virtual_branches` by profiling: **gitbutler_diff:workdir()**
* [x] Can `diff::workdir()` be implemented without writing objects (but also, without skipping files probably?) in `git2`?
    - Screenshot 2024-08-28 at 08 10 15
    - It kind-of works as it's a bit faster, doesn't write objects. Also there is no control over big files, but that's more of a diff-API issue.
* [x] make 'big files check' faster (**1.9x**)
* [x] remove duplicate `diff::workdir()` call when responding to filesystem events (reuse the data instead) - makes it **800ms** faster
* [x] fix failing tests
* [x] bring big-file support back to read-only `diff::workdir()`
* [x] fix E2E test - maybe it's a sign that `get_workspace_head()` is required?
    - it was flaky

### Notes for the Reviewer

* ~~`recalculate_everything()` is about twice as fast in this PR as it is on Master.~~ I think I compared debug with release or something. This PR is merely equally fast for responses to filesystem changes, but will write much less to the ODB in the process as it avoids `index.add_all(*)`.
* `diff::workdir()` is read-only, but a little slower as it now has to find big files to skip first (in order to be mergeable). Overall it's still an improvement, but wants `gix` (in follow-up)

### Follow-Ups

* ❌`apply_branch` speedups
   - https://github.com/gitbutlerapp/gitbutler/pull/5899
   - Actually, everything changed, let's keep it as is.
* ✅`gix` for `get_worktree_head()` - merge-trees
    - https://github.com/Byron/gitoxide/pull/1566
    - https://github.com/Byron/gitoxide/pull/1585
    - https://github.com/Byron/gitoxide/pull/1611
    - https://github.com/Byron/gitoxide/pull/1612
    - https://github.com/GitoxideLabs/gitoxide/pull/1618
    - https://github.com/gitbutlerapp/gitbutler/pull/5411
    - https://github.com/gitbutlerapp/gitbutler/pull/5416
    - https://github.com/GitoxideLabs/gitoxide/pull/1658
    - https://github.com/GitoxideLabs/gitoxide/pull/1659
    - https://github.com/gitbutlerapp/gitbutler/pull/5490
    - https://github.com/GitoxideLabs/gitoxide/pull/1661
[...]

PR #4793, opened 2024-08-29
performance improvements and bug fixes

### LINKED ISSUE #4795: Creating a virtual branch from a local tracking branch that isn't from GitButler duplicates commits
### Version

0.12.21

### Operating System

Mac OS X

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

When creating a virtual branch from a local tracking branch that has no GitButler change-ids, it creates new local commits for each commit in the local tracking branch, and fails to see that the same patch exists in the remote branch already.

### How to reproduce

In the GitLab repository: https://gitlab.com/gitlab-org/gitlab, do as shown in the video.

https://github.com/user-attachments/assets/f4a169f8-5232-4c6a-b15e-7a66af6622fe

It should work with any branch.

### Expected behavior

It should probably look like this:

Screenshot 2024-08-29 at 19 18 19

### Relevant log output

_No response_

Byron:
This PR is planned as follow-up to #4771, even though it chooses to pick other low-hanging fruit first.

This means, adding more instrumentation to find long-running functions or functions that run more than once in the same operation.
From there, one can avoid duplication, and optimize specific cases while keeping `git2`, at least at first.

Probably this will lead to this PR being merged early as it aims to deliver more small improvements while guiding the way towards
bigger improvements that would require `gix`, along with some upgrades to it which may take more time.

### Tasks

* [x] 2x performance (according to profiler) for `list_remote_branches()` (now `list_local_branches()`)
* [x] Fix #4795 
   From 
   Screenshot 2024-08-29 at 22 34 04
   To 
   Screenshot 2024-08-29 at 22 34 53

### Notes for the reviewer

* It's not actually a perfect fix, as I saw this while testing:
  Screenshot 2024-08-29 at 22 50 43
   And actually, it turned out that this is a similar looking, but different, commit, so it's correct after all.
* After unapplying, I saw that there are two duplicate commits which seem exactly the same (but have different patches). It was strange that the UI didn't show them combined anymore, maybe intentionally? Or maybe it's another implementation? It seems like it, so I didn't touch it.
   Screenshot 2024-08-29 at 22 52 34

### Follow-Up

* accelerate `is_commit_integrated()` - octopus merge with trees as results
    Screenshot 2024-08-29 at 18 48 46

### Notes

* When applying a branch in GitLab, it runs into the write-lock and operations block each other, with the last recalculation taking 20.4s (most of the time is waiting). Overall, this cascade of events seems very suboptimal.
    - I feel a lot of compute could be saved if the operation itself would emit whatever the UI needs, instead of relying on the watcher. The watcher could be deactivated/told to ignore what happens while the alteration is running. However, this also seems very complicated so it's probably better for now to make everything faster.
* It seems like `list_local_branches()` (formerly `list_remote_branches()`) is used to build a combined branches view, similar what the new `list_branches` does. It's hard for me to see how it's used, and it am surprised this information is still used anywhere as the side-bar definitely uses the new `branch_listing` now.
* It's clear that my tests of creating a vbranch from a local branch cause particularly bad performance when listing virtual branches due to their integration-commit checks.

### Issues

These issues where created while processing this PR.

* #4794
* #4795

### References

* https://www.figma.com/design/FbeLt0yjY9RiNn8MXUXsYs/Client-design?node-id=2907-112917&node-type=CANVAS - branch states

### DIFF
--- a/apps/desktop/src/lib/stores/remoteBranches.ts
+++ b/apps/desktop/src/lib/stores/remoteBranches.ts
@@ -21,7 +21,7 @@ export class RemoteBranchService {
 		try {
 			const remoteBranches = plainToInstance(
 				Branch,
-				await invoke<any[]>('list_remote_branches', { projectId: this.projectId })
+				await invoke<any[]>('list_local_branches', { projectId: this.projectId })
 			);
 			this.projectMetrics?.setMetric('normal_branch_count', remoteBranches.length);
 			this.branches.set(remoteBranches);
--- a/apps/desktop/src/lib/vbranches/types.ts
+++ b/apps/desktop/src/lib/vbranches/types.ts
@@ -179,13 +179,19 @@ export class DetailedCommit {
 	conflicted!: boolean;
 	// Set if a GitButler branch reference pointing to this commit exists. In the format of "refs/remotes/origin/my-branch"
 	remoteRef?: string | undefined;
+	// Identifies the remote commit id from which this local commit was copied. The backend figured this out by comparing
+	// author, commit and message.
+	copiedFromRemoteId?: string;
 
 	prev?: DetailedCommit;
 	next?: DetailedCommit;
 
 	get status(): CommitStatus {
 		if (this.isIntegrated) return 'integrated';
-		if (this.isRemote && (!this.relatedTo || this.id === this.relatedTo.id))
+		if (
+			(this.isRemote && (!this.relatedTo || this.id === this.relatedTo.id)) ||
+			(this.copiedFromRemoteId && this.relatedTo && this.copiedFromRemoteId === this.relatedTo.id)
+		)
 			return 'localAndRemote';
 		return 'local';
 	}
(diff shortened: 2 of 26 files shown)


krlvi (approved):

krlvi (approved):

mtsgrd:
@Byron this seems to break amending of commits. If I have a virtual branch with an upstream and I amend the only commit it stays _local and remote_ instead of becoming _local_ (which can be force pushed).

Byron:
Thanks for letting me know! I think this means that a patch-id also has to be computed to detect such changes. And that would make this operation quite expensive even though I think it can easily be accelerated by using all cores, i.e. compute patches in parallel.
I'd be glad to do that if you think that fixing the initial shortcoming, i.e. creating a virtual branch from a remote and having its commits displayed as "Local and remote" instead of separately, is worth it. If not, a revert would be in order.
Please let me know and I will submit a PR with the respective kind of fix.

mtsgrd:
Hey thanks for responding so quickly. Iiuc we would want to display the list as if the branch was rebased, but it's difficult since the remote id's don't necessarily have change id's?

It feels like checking author, committer, and message is the right way to accomplish this, so I'm thinking the problem is in the front-end code. Instead of identifying the commit existing on both local and remote, it should be identified as local, but having been rebased. Let me see if I can fix this!

mtsgrd:
See #4807!

PR #4798, opened 2024-08-30
fix: do not allow pressing branch delete btn multiple times

### LINKED ISSUE #4794: `Delete Branch` can be triggered multiple times
### Version

0.12.21

### Operating System

Mac OS X

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

  The `Delete` button for virtual branch deletion receives clicks and doesn't indicate it's doing anything, even though each click counts and triggers a delete action (visible in the log).
This is most noticeable in big repositories, and probably doesn't matter in smaller ones.

### How to reproduce

* Create a virtual branch from an existing on in the GitLab (https://gitlab.com/gitlab-org/gitlab) repository
* Delete the branch so the popup shows: 
  Screenshot 2024-08-29 at 16 36 07
* Click the `Delete` button - it will receive any click while the popup is visible, and trigger a delete operation each time.

### Expected behavior

The `Delete` button is greyed out and can't be clicked until the operation completes. If the operation fails, the popup should probably remain visible.

### Relevant log output

_No response_

ndom91:
## ☕️ Reasoning

- Delete branch actoin can be triggered multiple times on the Rust side

## 🧢 Changes

- Add a `loading` attribute to the delete fn, which when true will disable the btn, among other things.
- Fix `loading={isDeleting}` attribute in the `BranchPreviewHeader.svelte` as well.

Fixes: #4794

### DIFF
--- a/apps/desktop/src/lib/branch/BranchLaneContextMenu.svelte
+++ b/apps/desktop/src/lib/branch/BranchLaneContextMenu.svelte
@@ -38,6 +38,7 @@
 	let aiConfigurationValid = $state(false);
 	let newRemoteName = $state('');
 	let allowRebasing = $state<boolean>();
+	let loadingDelete = $state(false);
 
 	const branch = $derived($branchStore);
 	const commits = $derived(branch.commits);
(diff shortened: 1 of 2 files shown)


Byron (commented):
Thank you, that looks like it should do the trick!

I am just commenting here as I don't think I am qualified to arm-chair (browser-only) review frontend code 😅.

Byron (approved):
Actually, let me approve it so it's not blocked :).

PR #4804, opened 2024-08-30
Tooltip-refactoring-+-new-component

PavelLaptev:
the benefit with this approach is to use svelte components and transitions + isolate tooltip styles.
This is also a preparation for another tooltip variant - a mini-card tooltip for more complex content inside. In this case we should be able to pass svelte snippets


packages/ui/src/lib/Tooltip.svelte
```diff
@@ -0,0 +1,142 @@
...
+				use:portal={'body'}
+				class="tooltip-container text-11 text-body"
+				transition:flyScale={{
+					position: position
+				}}
+			>
+				<span>{text}</span>
+			</div>
+		{/if}
+	</span>
```
mtsgrd, line 114: In the past we had a tooltip component that surrounded the target with an element this way, but I felt like it introduced problems in various circumstances. Do you feel confident this works e.g. when you have an element inside an `overflow: none` container?

PavelLaptev: It shouldn’t be a problem. I also don’t like it when components like this are rendered inside their targets or triggers. The component will be rendered outside the element in the ‘body’ using the ‘portal’ hook.

PavelLaptev: I didn't provide this in the PR description: the benefit with this approach is to use svelte components and transitions + isolate tooltip styles.

mtsgrd: A'ight, but let me know if you run into any problems in the future. I personally don't like the wrapping container for anything with a tooltip, but it's a moot point if there aren't any negative consequences.

The alternative solution I could think of is using a directive still, but having single component responsible for rendering, like ToastController.

mtsgrd (approved):

packages/ui/src/lib/Tooltip.svelte
PavelLaptev: Thanks! Let's see how it goes. There will be another similar component — a mini-card tooltip that appears on hover— where I should be able to pass Svelte components like links, images, etc. This change is an experiment to see if we can create a tooltip as a Svelte component.

image

PR #4807, opened 2024-09-01
Fix `copiedFromRemoteId` bug

mtsgrd:
- correctly shows commits as rebased when changeId is missing
- setting relatedTo is sufficient, no need to touch `get status()`

For testing I applied this branch by Pavel:
image

### DIFF
--- a/apps/desktop/src/lib/vbranches/types.ts
+++ b/apps/desktop/src/lib/vbranches/types.ts
@@ -188,10 +188,7 @@ export class DetailedCommit {
 
 	get status(): CommitStatus {
 		if (this.isIntegrated) return 'integrated';
-		if (
-			(this.isRemote && (!this.relatedTo || this.id === this.relatedTo.id)) ||
-			(this.copiedFromRemoteId && this.relatedTo && this.copiedFromRemoteId === this.relatedTo.id)
-		)
+		if (this.isRemote && (!this.relatedTo || this.id === this.relatedTo.id))
 			return 'localAndRemote';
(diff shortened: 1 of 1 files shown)


krlvi (approved):

apps/desktop/src/lib/vbranches/types.ts
```diff
@@ -188,10 +188,7 @@ export class DetailedCommit {
 
 	get status(): CommitStatus {
 		if (this.isIntegrated) return 'integrated';
-		if (
-			(this.isRemote && (!this.relatedTo || this.id === this.relatedTo.id)) ||
-			(this.copiedFromRemoteId && this.relatedTo && this.copiedFromRemoteId === this.relatedTo.id)
-		)
+		if (this.isRemote && (!this.relatedTo || this.id === this.relatedTo.id))
```
Byron, line 191: Wonderful, less really is more!

I also tried my use-case again, which is to create a vbranch from another branch that has no change-ids in commits, and then undid one commit, and it also just works.

Screenshot 2024-09-01 at 18 57 19

I think what I wanted to accomplish is to have it show as `Local and remote` just as if change-ids were equal (or it's the same actual commit), but probably I misunderstood the state (or status) that such a copied commit is actually in.

And somewhat magically, it is all working even when the commit message is amended to the local commit.

krlvi:
I like this very much. Just want to note that this fixes the visuals on the workspace (1) but in the branch preview page (2) it's still borked.

(1)
image

(2)
image

PR #4813, opened 2024-09-02
Can remove repository if it's missing, from the error boundary page

estib-vega:
## ☕️ Reasoning
If you delete or move a repository tracked by GitButler, you get the nice _Something Went Wrong_ page and are offered to change the repository.

At that point, there is no way of deleting that repository from the list of tracked repos. Normally, to delete a repo from the application, one needs to go into the settings of the repository itself and click "Remove repository". This is not possible once the repository is missing.

This change allows users to delete the missing repository from the list of tracked repos from within the _Something Went Wrong_ view.

## 🧢 Changes

- Create a centralized errors module to be able to identify certain "known" errors (in this case, the missing repository error).
- Create a section in the _Something Went Wrong_ that displays some optional actions to perform, in this case, deleting the missing repository from the list.
- Adapt the `RemoveProjectButton` so that it can be used without showing the confirmation modal

## 📺 Visuals
Suuuuper rough designs. 

This is how it looks if for some reason, the application fails to even delete the repo from the list of tracked repos. (Should usually not happen)
https://github.com/user-attachments/assets/f2a37d88-e3c0-4f05-a683-801aeef34af9

This is how it looks when it succeeds in deleting the repo from the list.
https://github.com/user-attachments/assets/7eee00c7-88d3-4c90-b1d1-58ddb301273b


Caleb-T-Owens:
Hey!

I'm curious about some things:

Is there a reason why we can't make use of the error codes that are part of our rust `anyhow` setup https://github.com/gitbutlerapp/gitbutler/blob/444b03ff388fda41c73b362f3234a426dbd415a5/crates/gitbutler-error/src/error.rs#L125 and then mapped in `ipc.ts` https://github.com/gitbutlerapp/gitbutler/blob/444b03ff388fda41c73b362f3234a426dbd415a5/apps/desktop/src/lib/backend/ipc.ts#L5 ?

Is there any benefit to having a prompt to the user about this error in order to resolve the situation? If the project has been removed or moved, there is no way to resolve the situation other than removing the project from gitbutler. Would we be better off on the rust side, whenever we detect a project has a missing dependency automatically removing the zombie project and providing the user a toast saying that we've tidied it up for them?

estib-vega:
> Is there a reason why we can't make use of the error codes that are part of our rust anyhow setup
>

@Caleb-T-Owens I saw those, but I'm not sure we can access the codes from Svelte. Hmm I so those, but wasn't sure if those were accessible by Svelte since I didn't find it in the obj type, but just saw that we extend those ourselves, so maybe.

I'll take a look 👍 

> Is there any benefit to having a prompt to the user about this error in order to resolve the situation? If the project has been removed or moved, there is no way to resolve the situation other than removing the project from gitbutler. Would we be better off on the rust side, whenever we detect a project has a missing dependency automatically removing the zombie project and providing the user a toast saying that we've tidied it up for them?
>

It's a great question 🤔.
I agree that maybe an extra button is a bit of an overkill for an action that most probably will be performed, but I think still makes sense to raise this to the user at least as a check-box.

Just to accommodate edge-cases like, "oh I moved a bunch of files around and mistakenly moved this repo".

I'd stand more on the side of raising this kinds of actions to the user before they're performed, seems a bit friendlier imo.

What do you think @Caleb-T-Owens?

Caleb-T-Owens:
FYI: Estib and I talked in person and I have been convinced of having the buttons

krlvi:
@estib-vega why is it that the `delete_project` endpoint not work for this scenario?

estib-vega:
> @estib-vega why is it that the `delete_project` endpoint not work for this scenario?

I needed the project ID for that and I couldn't access it from the error.
I talked to @Caleb-T-Owens and he pointed me in a better direction to actually find the project ID first and use `delete_project`

estib-vega:
Screenshot 2024-09-02 at 18 01 39

Alright @PavelLaptev, I think it's ready from my side :)

crates/gitbutler-project/src/controller.rs
```diff
@@ -120,16 +120,22 @@ impl Controller {
...
 
     /// Like [`Self::get()`], but will assure the project still exists and is valid by
     /// opening a git repository. This should only be done for critical points in time.
     pub fn get_validated(&self, id: ProjectId) -> Result<Project> {
-        self.get_inner(id, true)
+        self.get_inner(id, true, false)
     }
 
-    fn get_inner(&self, id: ProjectId, validate: bool) -> Result<Project> {
+    fn get_inner(&self, id: ProjectId, validate: bool, skip_dir_creation: bool) -> Result<Project> {
```
estib-vega: Also, I'm not sure whether or not adding a third way of getting the project is the best idea here.

Maybe we can just skip the creation of the directory if `validate` is set to false, just wasn't sure if there are some cases in which we need to create the project directory even if there is no validation done

apps/desktop/src/lib/components/errorBoundaryActions/InexistentRepo.svelte
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	async function locate() {
+		const id = projectService.getLastOpenedProject();
+		if (id === undefined) {
+			return;
+		}
+
+		await projectService.relocateProject(id);
+	}
```
estib-vega: I don't love this code, but to me it made a bit more sense to wait on design input before cleaning it up

apps/desktop/src/lib/components/errorBoundaryActions/InexistentRepo.svelte
```diff
@@ -0,0 +1,73 @@
...
+	{/if}
+
+	{#if deleteSucceeded === true}
+		<p class="text-12 text-body text-bold">Repo removed successfully</p>
+	{/if}
+
+	{#if deleteSucceeded === false}
+		<p class="text-12 text-body text-bold">Failed to remove repo</p>
+	{/if}
+</div>
```
estib-vega: Same as the comment above. This should be cleaned up once design has vetted/improved this change :)

PavelLaptev:
@estib-vega I'll create a separate error screen for this case. I think this will be better, because there will be many small copy and layout changes, and it's easier to handle them all at once as one component instead of using many `if/else` statements in `SomethingWentWrong`.

But I need a few things. Is there a way to get the project path and name if it's not found?

image

I see in the code we can get the project Id using `projectService.getLastOpenedProject()`, but I couldn't find how to get the path and name.

estib-vega:
@PavelLaptev Sounds good!

I think we'd still need if/else statements, the question is at what level. 
It makes sense to move it a level above if we want a dedicated view for this error 👍

Yes, we can access project name and path. I'll set it up following your design and update the PR :)

PavelLaptev:
Cool. What We can do, I can update the design and leave these two - the path and the name as placeholders, so you can replace it after 🙂

estib-vega:
You've got a deal 🤝

apps/desktop/src/lib/components/errorBoundaryActions/InexistentRepo.svelte
mtsgrd: While inexistent is technically a word, it's rarely used. can we name this `NonExsistentRepo` instead? 😏

estib-vega: > While inexistent is technically a word
>
😂

Sure thing!

crates/gitbutler-project/src/controller.rs
estib-vega: Talked to @Caleb-T-Owens, he suggested we just implement a completely separate method that strips out any other validations

PavelLaptev:
I just pushed my changes. Feel free to update or modify the code I added or removed. I did this to achieve the design I wanted.

image

PavelLaptev:
---

I noticed that no projects are showing up in the project selector. I'm not sure if you can see this issue on your end.

image

estib-vega:
>I noticed that no projects are showing up in the project selector. I'm not sure if you can see this issue on your end.
>
I see that too

estib-vega:
> I noticed that no projects are showing up in the project selector. I'm not sure if you can see this issue on your end.
> 

Seems to be unrelated to this changes, so I'm addressing that in a separate PR: https://github.com/gitbutlerapp/gitbutler/pull/4821#pullrequestreview-2277744722

PavelLaptev (approved):
lgtm, thanks @estib-vega

PR #4814, opened 2024-09-02
feat: add support for custom pull request template paths

ndom91:
## ☕️ Reasoning

- After a user comment on discord and some research, it turns out that folks put their PR templates at both `.github/PULL_REQUEST_TEMPLATE.md` and `.github/pull_request_template.md` about 50/50. There is also technically an option for folks to put their template anywhere under `.github/PULL_REQUEST_TEMPLATE/*.md`
- Therefore, adding a settings path for folks to customize which PR template they want to be used. 

## 🧢 Changes

- The PR template enable boolean and PR template path (new) are now in the project preferences.
- This section is hidden if the user is not logged into GitHub

## 📺 Screenshots

image

### DIFF
--- a/apps/desktop/src/lib/config/config.ts
+++ b/apps/desktop/src/lib/config/config.ts
@@ -53,3 +53,7 @@ export function persistedCommitMessage(projectId: string, branchId: string): Per
 export function gitHostUsePullRequestTemplate(): Persisted<boolean> {
 	return persisted(false, 'gitHostUsePullRequestTemplate');
 }
+
+export function gitHostPullRequestTemplatePath(): Persisted<string> {
+	return persisted('', 'gitHostPullRequestTemplatePath');
+}
--- a/apps/desktop/src/lib/gitHost/gitHostFactory.ts
+++ b/apps/desktop/src/lib/gitHost/gitHostFactory.ts
@@ -21,7 +21,8 @@ export class DefaultGitHostFactory implements GitHostFactory {
 		repo: RepoInfo,
 		baseBranch: string,
 		fork?: RepoInfo,
-		usePullRequestTemplate?: Persisted<boolean>
+		usePullRequestTemplate?: Persisted<boolean>,
+		pullRequestTemplatePath?: Persisted<string>
 	) {
 		const domain = repo.domain;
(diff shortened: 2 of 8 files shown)


mtsgrd:
About the pre-styled message, look for `showToast(...)` instead of using `showError` if you want to render it differently!

ndom91:
> About the pre-styled message, look for `showToast(...)` instead of using `showError` if you want to render it differently!

Yeah i tried that [initially](https://github.com/gitbutlerapp/gitbutler/pull/4814/commits/65482dac29dacd9798a84b21a4084fe055b54457), but the `message` still seemed to get rendered into what was displayed as like a `<pre />` element. 

Ahh I see, since the `type: 'error'` it'll always wrap the content in a `<code />`. Maybe I'll just make this `style: 'neutral'` then.

mtsgrd:
Actually, I was the one confused. I see we use `<code>` for errors, and markdown formatted content for other content.

ndom91:
Much better!

image

ndom91:
@PavelLaptev pinging you for the UI of this preferences section :blush:

What do you think? This is in the project preferences

image

estib-vega:
The code looks good to me 👍

estib-vega (approved):

PavelLaptev (approved):
@ndom91 thanks! lgtm

koppor:
> * Therefore, adding a settings path for folks to customize which PR template they want to be used.

i would have expected that Gitbutler auto detects the existance of the template. The paths are pre-determined by GitHub. I think, it are three (times two because of casing). See: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

I agree, users can specify more in a directory `github/PULL_REQUEST_TEMPLATE/`. For that, the configuration is nice. I would have expected a drop down in this case.

Reason: GitButler is interacting with GitHub directly, not with arbitrary platforms. Thus, GitButler should make it convenient to interact with GitHub.

ndom91:
@koppor thanks for the suggestion, I think if we check the few case-variants and then potentially list hte contents of a `.github/PULL_REQUEST_TEMPLATE/` dir (if it exists), then we could easily offer this as a dropdown. I agree that'd be a ncier UX :+1:

PR #4816, opened 2024-09-02
fix: Zen PC hover animation

PavelLaptev:
## ☕️ Reasoning

There was a seam between two states — when the Zen PC is not moving and when the animation starts to play.

https://github.com/user-attachments/assets/d693c54b-d526-4570-82e9-ef8270834b31


krlvi (approved):
lgtm

Issue #4137, opened 2024-06-20
UX: make it clear "discard changes" on a file will affect all lanes


daniilS:
### Version

0.12.5

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

"Discard changes" is accessed from the context menu inside one particular lane, but will discard any changes to the file in _all_ lanes - this doesn't feel like obvious behaviour to me, and since discarding already gives a modal to confirm, I think it would be good to add a check if the selected file also has changes in any other lanes, and point it out in the modal if yes.

Byron:
Thanks for reporting, I also see how this behaviour can be confusing without pointing it out specifically.

Just in case anything got lost, one should be able to undo the discard as (roughly) indicated in the screenshot (it's the clock symbol in the bottom-left, right next to settings). 

<img width="1288" alt="Screenshot 2024-06-20 at 22 02 15" src="https://github.com/gitbutlerapp/gitbutler/assets/63622/4981ddd3-8c42-4de5-a2a3-b9a254dfbb1b">

daniilS:
Nothing was lost, don't worry - think I actually found this while testing for #4136 😂
But yes I'm a big fan of the undo button!

mtsgrd:
Would you mind if we instead make it discard only changes that belong to the particular branch?

daniilS:
@mtsgrd I think that would work well! I assumed there was some reason for the current behaviour hence my suggestion for the warning, but personally I can only think of use cases for discarding only changes in the current lane.

daniilS:
@mtsgrd wondering if there has been any update on this? Have come across a few times where I've wanted to discard all changes to a file in a branch without affecting other branches

mtsgrd:
Hey, thanks for bringing this up again. I will try and prioritise it and get it done next week. 🙏

krlvi:
It appears that this https://github.com/gitbutlerapp/gitbutler/pull/4820 resolves this problem. I'm making a release in a few minutes that will contain the fix so that we can have it tested

krlvi:
This was fixed by https://github.com/gitbutlerapp/gitbutler/pull/4820 and shipped in release `0.12.24`. Discarding a file from a lane now only discards that lane's changes. Closing this as resolved.

PR #4820, opened 2024-09-03
Only discard the files changes inside the lane

estib-vega:
### Problem
When using the file context menu inside a given lane to discard the changes, all changes to that file across all lanes are discarded.
This is either weird or unexpected behavior.

### Solution
Only discard the changes inside the lane that the context menu was opened from.

### Issues
This fixes issue https://github.com/gitbutlerapp/gitbutler/issues/4137

### DIFF
--- a/apps/desktop/src/lib/vbranches/branchController.ts
+++ b/apps/desktop/src/lib/vbranches/branchController.ts
@@ -226,11 +226,12 @@ export class BranchController {
 		}
 	}
 
-	async unapplyFiles(files: LocalFile[]) {
+	async unapplyFiles(branchId: string, files: LocalFile[]) {
 		try {
 			await invoke<void>('reset_files', {
 				projectId: this.projectId,
-				files: files.flatMap((f) => f.path).join('\n')
+				branchId,
+				files: files.flatMap((f) => f.path)
 			});
 		} catch (err) {
 			showError('Failed to unapply file changes', err);
--- a/crates/gitbutler-branch-actions/src/virtual.rs
+++ b/crates/gitbutler-branch-actions/src/virtual.rs
@@ -32,12 +32,7 @@ use gitbutler_repo::{
 use gitbutler_time::time::now_since_unix_epoch_ms;
 use serde::Serialize;
 use std::collections::HashSet;
-use std::{
-    borrow::Cow,
-    collections::HashMap,
-    path::{Path, PathBuf},
-    vec,
-};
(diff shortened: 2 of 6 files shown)


crates/gitbutler-branch-actions/src/virtual.rs
```diff
@@ -183,28 +178,44 @@ pub fn unapply_ownership(
...
+                .iter()
+                .filter(|o| files.contains(&o.file_path.to_string_lossy().to_string()))
+                .cloned()
+                .collect::<Vec<gitbutler_branch::OwnershipClaim>>();
+
+            BranchOwnershipClaims {
+                claims: filtered_ownership_claims,
+            }
+        })
+        .context("failed to determin ownership to unapply")?;
```
mtsgrd: typo!

estib-vega: Oops!
Fixed :)

apps/desktop/src/lib/vbranches/branchController.ts
```diff
@@ -226,10 +226,11 @@ export class BranchController {
 		}
 	}
 
-	async unapplyFiles(files: LocalFile[]) {
+	async unapplyFiles(branchId: string, files: LocalFile[]) {
 		try {
 			await invoke<void>('reset_files', {
 				projectId: this.projectId,
+				branchId,
 				files: files.flatMap((f) => f.path).join('\n')
```
Byron: Can't this be a list of paths right away? My edits now assume this works, and I hope you can validate that.

Byron (commented):
Thanks so much for all your recent work here (*making missing projects deletable was great!*).

I took the liberty to edit the rust-side of things so that it's idiomatic and less wasteful. However, I also tried something new which is to pass `Vec<PathBuf>` from the frontend to the backend directly. In the frontend, this means that a string-array should be passed and I think it should convert to `Vec<PathBuf>` automatically. However, I didn't try it as I didn't adjust the frontend code.

It would be great if it could work like this, but if not, `Vec<String>` will be fine, and a conversion to `Vec<PathBuf>` in the backend.

Thanks again!

estib-vega:
Hey @Byron! Your improvements are more than welcome haha.
I try to adjust the FE code accordingly and give it a go.

Thanks!

mtsgrd:
I added an e2e testing label here as a reminder this might be a good thing to cover.

ndom91 (approved):
LGTM :partying_face:

PR #4821, opened 2024-09-03
fix: Update the select options correctly

estib-vega:
The Select component options are a prop that initializes the state variable `filteredOptions`. 
The issue was, that if the prop is updated, the state variable doesn't seem to be updated as well.
This fixes that by manually updating the state variable whenever the prop updates.


apps/desktop/src/lib/select/Select.svelte
```diff
@@ -55,6 +55,10 @@
 	let listOpen = $state(false);
 	let inputBoundingRect = $state<DOMRect>();
 
+	$effect(() => {
+		filteredOptions = options;
+	});
+
```
estib-vega, line 61: @Caleb-T-Owens @ndom91 Not sure if this is actually anti-pattern in Svelte or not.
What do you guys think?

ndom91: Yeah so ideally we'd not use $effect to sync state, they recommend using a callback if available. 

Check out this video if you're curious haha - https://youtu.be/7N4maEDhy4w

Would a $derived do the trick here?

estib-vega: > Yeah so ideally we'd not use $effect to sync state,
>

Yeah that makes sense.

 Thanks for the video!

> they recommend using a callback if available.
>

A callback :) ?

> Would a $derived do the trick here?
>

I expected that but it seems like `$derived` variables are not re-assignable.
I also expected that `$state` variables would be re-set to the prop that initialized it, but seem like not :/

estib-vega: Oh wait, I think the issue is somewhere deeper.

estib-vega: Talked to @ndom91, and there seems to be no other way to handling this. We agreed to leave it like that for the time being

Caleb-T-Owens (approved):
LGTM. It would be nice to re-think this at some point, but its good for now.

estib-vega:
> LGTM. It would be nice to re-think this at some point, but its good for now.

I agree. Might also be possible that the Svelte folks improve this behavior or add a dedicated rune for this in the meantime.
Been looking at this too: https://github.com/sveltejs/svelte/issues/13101

Caleb-T-Owens:
@estib-vega something like:

```ts
let searchTerm = $state<string>();

const filteredOptions = $derived(options.filter((option) => option.whateverthekeyis.includes(searchTerm))) // Or some alternative searching (we've got fuse.js used elsewhere in the project)
```

estib-vega:
> @estib-vega something like:
> 
> ```ts
> let searchTerm = $state<string>();
> 
> const filteredOptions = $derived(options.filter((option) => option.whateverthekeyis.includes(searchTerm))) // Or some alternative searching (we've got fuse.js used elsewhere in the project)
> ```
> 
> is a possible approach

Yeah, something like that will need to be done. But in order to get that we need to remove the way that the `Select` and the `SearchItem` component are setup to abstract the filtering part.

`$derived` variables can't be reassigned and we need to reassign it because the child component updates the `filteredOptions`

estib-vega:
I'd even consider the fact that a prop update doesn't update a state variable a Svelte bug, tbh.

Issue #4790, opened 2024-08-29
Enter at beginning of a line should move text


koppor:
### Version

0.12.21

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

<kbd>Enter</kbd> should move text

### How to reproduce

1. Add something in the first line of commit message
2. Press <kbd>Pos1</kbd>
3. Press <kbd>Enter</kbd>\
    ![image](https://github.com/user-attachments/assets/841b9adf-6d6e-46f8-a295-ef84b5e86902)
4. Cursor is moved down.

### Expected behavior

Text and cursor should be moved down - and first line empty

Byron:
Thanks for reporting!

I can confirm that the cursor only moves down on enter if the commit text field isn't empty.

I couldn't reproduce the behaviour with `Pos1` (as this key doesn't exist on my keyboard), nor could I reproduce it by moving the cursor to the beginning of the first line by pressing `Ctrl + a` - Enter would work as expected.

Maybe it can be considered a feature not to allow leading with a newline and to disallow moving the cursor down with `Enter` if the textbox is entirely empty. Probably this makes the described behaviour specific to Windows, so I at best partially reproduced it.

koppor:
> I can confirm that the cursor only moves down on enter if the commit text field isn't empty.

Normal text editors move the letters following the cursor to the next line. Before, they move all subsequent lines down by one line. - GitButler just moves the cursor.

mtsgrd:
cc @PavelLaptev

PavelLaptev:
@estib-vega if you have time, could you please take a look :-)

estib-vega:
Merged the fix in https://github.com/gitbutlerapp/gitbutler/pull/4826

koppor:
I confirm that it works.

Now, <kbd>Del</kbd> at the beginning of the line (no issue filed yet; will do as soon as I start missing it) and <kbd>Cursor down</kbd> (https://github.com/gitbutlerapp/gitbutler/issues/5095) should work.

PR #4826, opened 2024-09-04
Improve the commit text content experience

estib-vega:
### As seen on issue 
https://github.com/gitbutlerapp/gitbutler/issues/4790

### Problem
Pressing the 'Enter' key on the commit message text area will only focus the description text area, but not move the text down.

### Expected behavior 
1. Move everything right of the cursor into the description text area
2. Delete all of the selected text

### Solved
Pressing enter from the title will behave in the expected way, pushing down the content into the description box and deleting any selected text

### Visuals
text-area-update


PavelLaptev (approved):
lgtm :-) thanks!

PR #4829, opened 2024-09-04
UI: Info Button component

estib-vega:
Add an "Info button" that displays a custom content on click

### Design
Following this design specs: https://www.figma.com/design/Jir22AjzvPokdbMbe8yt5k/%F0%9F%92%8E-Core?node-id=4360-9832&node-type=FRAME&t=5K62h7aFKk2HH5bt-0

Screenshot 2024-09-05 at 10 22 28


PavelLaptev:
@estib-vega thanks for the PR!

A few changes I’d like to suggest:

- We should separate the `Tooltip` from the `Info Button`. The `Info Button` shouldn’t be treated as a `custom tooltip`. Instead, both `Tooltip` and `Info Button` can share a common component, such as `TooltipWrapper`. However, I’m not in favor of allowing the `Tooltip` to be modified arbitrarily via `customTooltip`.
- The `Info Button`'s tooltip should have a `title` (optional) and a `description` (required), not just `children`.
- The tooltip should appear on hover, not on click.
- There’s an issue with arrow positioning. Since the `Info Button` uses the same code from the `Tooltip` component, the tooltip is positioned relative to available space on the left or right, but the arrow remains static and is always centered (see image). Let's adjust the `Info Button` tooltip so that it always remains centered relative to the button.

image

PavelLaptev (changes requested):
@estib-vega thanks for the PR!

A few changes I’d like to suggest:

- I want to separate the `Tooltip` from the `Info Button`. The `Info Button` shouldn't be part of the `Tooltip` as a `custom tooltip`. There could be a shared component like `TooltipWrapper`, but I don’t like the idea of being able to change the `Tooltip` using `customTooltip`. They don’t share enough code to be one component, and I don’t want one component to depend on the other. It's not a requirement it's how I see these components from the design POV.
- The `Info Button`'s tooltip should have a `title` (optional) and a `description` (required).
- The tooltip should appear on hover, not on click.
- There’s an issue with arrow positioning. Since the `Info Button` uses the same code from the `Tooltip` component, the tooltip is positioned relative to available space on the left or right, but the arrow remains static and is always centered (see image). Let's adjust the `Info Button` tooltip so that it always remains centered relative to the button.

image

PavelLaptev:
> @PavelLaptev Makes sense, will do!

thanks 🙏

PavelLaptev (approved):

estib-vega:
@PavelLaptev Looks great! Great job :)

PavelLaptev:
> @PavelLaptev Looks great! Great job :)

Cool, I'll merge it then :-)

PR #4832, opened 2024-09-05
fix: Disable the PR button if there is no GitHub integration

estib-vega:
If there is no GitHub integration, correctly disable the create PR button

### Issue
Even if there was no GitHub integration configured by the user, the button would still be enabled.
Clicking 'Create PR' would then push the branch and then fail to create the PR.

### Changes
Check whether there is both a GitHost **and** a PR service available, before enabling the PR creation button
**No design changes included in this PR**

### DIFF
--- a/apps/desktop/src/lib/branch/BranchHeader.svelte
+++ b/apps/desktop/src/lib/branch/BranchHeader.svelte
@@ -241,8 +241,10 @@
 						{#if !$pr}
 							<PullRequestButton
 								click={async ({ draft }) => await createPr({ draft })}
-								disabled={branch.commits.length === 0 || !$gitHost}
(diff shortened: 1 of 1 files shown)


apps/desktop/src/lib/branch/BranchHeader.svelte
```diff
@@ -241,8 +241,10 @@
 						{#if !$pr}
```
ndom91, line 241: Should we disable the button or just not render it at all maybe?

mtsgrd: Yeah let's disable it, it's how it was intended, but apparently not what came about.

krlvi: You guys forgot to approve the PR ;)

krlvi (approved):

PR #4834, opened 2024-09-05
Clean-up: Remove anti-pattern from the Select component

estib-vega:
Instead of updating the state through an effect, factor out the `searchValue` state variable into the parent component, and only derive the filtered options from it

I ended up following the approach suggested by @Caleb-T-Owens [here](https://github.com/gitbutlerapp/gitbutler/pull/4821#issuecomment-2328924429)

The only "compromise" is that now the search value state variable is now a bindable prop managed by the parent, and so concerns are not "fully" separated. But it's a fair price for removing an anti-pattern


krlvi (approved):
LGTM

PR #4842, opened 2024-09-06
Update the base branch correctly

estib-vega:
### Issue
The base branch is not correctly updated in the app when clicking on the sync button.

### Repro steps
1. Open a normal repository with at least two commits in GitButler
2. Using normal git in the terminal, checkout whatever your base branch is
3. Delete the last commit by doing a hard reset to the second to last commit
4. Force push that
5. Open GitButler, and go back into the `gitbutler/workspace`
6. Open the base branch view and refresh

The same commit that was deleted from the remote still appears in GitButler

### Changes
Clicking on the sync-button, will first invoke an update to the base branch.
This ensures that in the case that someone has force-pushed to the main branch, the application is able to correctly display the new state of the base branch


estib-vega:
FYI: 

@krlvi This is a bit related to what you were talking about, regarding trying make the FE "less smart". 
Maybe it makes more sense to move the call to `update_base_branch` to Rust.

I'll do that

Byron:
> FYI:
> 
> @krlvi This is a bit related to what you were talking about, regarding trying make the FE "less smart". Maybe it makes more sense to move the call to `update_base_branch` to Rust.
> 
> I'll do that

The key-takeaway is also that the backend should always leave a consistent state without requiring a certain call order or other intricate knowledge from the frontend. With that, and a couple of other changes that make state-acquisition easier for the frontend, I think the app will run much more smoothly and burn less CPU.
Exciting times!

estib-vega:
> [...] the backend should always leave a consistent state without requiring a certain call order or other intricate knowledge from the frontend.
>
 @Byron Fully agree! Exciting times indeed :)

crates/gitbutler-branch-actions/src/actions.rs
```diff
@@ -133,6 +133,7 @@ pub fn delete_local_branch(project: &Project, refname: &Refname, given_name: Str
 #[instrument(skip(project), err(Debug))]
 pub fn get_base_branch_data(project: &Project) -> Result<BaseBranch> {
     let ctx = CommandContext::open(project)?;
+    update_base_branch(project)?;
     base::get_base_branch_data(&ctx)
 }
```
Caleb-T-Owens, line 138: Is this safe?

Doesn't this mean that we could be accidentally rebasing peoples branches when they don't expect it?

Caleb-T-Owens (changes requested):

crates/gitbutler-branch-actions/src/actions.rs
estib-vega: Oh nice catch! I'm not sure 🤔.
I'll check

Caleb-T-Owens: Yeah, this means any time you refresh the app, or perform an action that re-fetches the sate of the base branch we're going to perform that update action.

We absolutely can't merge this in.

estib-vega: Gotcha 👍 

I guess then the fix should go into displaying the the update workspace update button for the situation I outlined in the PR description

Caleb-T-Owens: The issue here, is that we have no way of representing `master` (or your target commit) is ahead of `origin/master` in GitButler. We currently don't have mechanisms for merging branches into your local `master` either, so the only way to get into this state is by doing some fiddling via the git CLI.

We probably need to discuss how we want to handle having local `master` ahead of `origin/master` CC. @krlvi 

One possible solution would be to have a button which when clicked fixes any divergence between local `master` and `origin/master`. We ought to have extra warning so the user knows what is happening, because "Update" is for sure not what we'd be doing in that scenario.

Another, would be to add in a "push" button to push up that extra commit on local `master`

(or perhaps both)

estib-vega: I see, yeah that makes sense.

Byron: @estib-vega A great rule of physics of *this* universe is also that functions that indicate they are read-only, should never, ever also mutate state on disk. Right now, there is a couple of even massively important functions (I am looking at you, `list_virtual_branches()`) which violate physics essentially, and it's definitely something to roll-back and fix as well. Until then, it will let the cat out of the box occasionally :D.

In Rust, violating this rule isn't easily possible (even though it can be side-stepped), but since all of the code here is based on state that is read from disk, Rust isn't here to safe the day, unfortunately. In `list_virtual_branches()` you will also find the `perm:` parameter which acts as a marker to indicate if a method is allowed to read or write. Behind that is an actual file-based lock so no function marked that way can execute if another one is already running. However, due to this lock being quite costly, it's only acquired for heavier functions, not for those which essentially just read a file on disk (or write it).

I hope that helps.

estib-vega: It does! Thanks for the pointers :)

Issue #4791, opened 2024-08-29
How to hard reset in current default branch


LovesAsuna:
### Version

0.12.21

### Operating System

Mac OS X

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

<img width="386" alt="image" src="https://github.com/user-attachments/assets/806bd464-1196-49fe-a3c4-9e073eec9629">
How do I force push (hard reset) after undoing in the same virtual branch?

Byron:
I don't think that force-pushes from virtual branches are currently supported, but am also not entirely sure I am not missing something.
Maybe @krlvi could fill in the blanks.

LovesAsuna:
Is there a good solution for this situation? It seems like I can only force push from traditional branches.

H1ghBre4k3r:
Hey @LovesAsuna, does GitButler show you the classic `Force Push` button below the commits? 

Furthermore, did you explicitly disable force pushing for your repository within the settings?

LovesAsuna:
For example, I created a file named "a" and a file named "b," and committed them with the messages "add file a" and "add file b," respectively. At this point, I no longer want the commit "add file b," so I want to perform a hard reset. Note that these two commits are on the same commit chain. The force push button you mentioned will appear in cases where I have edited the commit message or squashed commits, causing a conflict between the local and remote commits.

estib-vega:
@LovesAsuna It's a good question. I don't see any reason not to allow this.
I'll take a look

PR #4845, opened 2024-09-06
Commit List: Allow force push when removing commits

estib-vega:
### Issue
As described in
https://github.com/gitbutlerapp/gitbutler/issues/4791

It seems that we only display the force-push button when local-only commits are present.
This is a bit annoying if you just want to e.g. get rid of the last commit of the branch by undoing it.
That would result in a list of local-and-remote commits and a remote-only commit,.

### Changes
Display the force-push action button under local-and-remote section if there are any remote-only commits.
This would mean that there is a set of commits that are only remotely available, and so the user has the option to force-push that

Screenshot 2024-09-06 at 17 26 08

### DIFF
--- a/apps/desktop/src/lib/commit/CommitList.svelte
+++ b/apps/desktop/src/lib/commit/CommitList.svelte
@@ -137,8 +137,6 @@
 			$listingService?.refresh();
 			$prMonitor?.refresh();
 			$checksMonitor?.update();
-		} catch (e) {
-			console.error(e);
 		} finally {
 			isPushingCommits = false;
 		}
--- a/apps/desktop/src/lib/commit/CommitList.svelte
+++ b/apps/desktop/src/lib/commit/CommitList.svelte
@@ -49,20 +49,30 @@
 	const reorderDropzoneManagerFactory = getContext(ReorderDropzoneManagerFactory);
 	const gitHost = getGitHost();
 
+	enum LineSpacer {
+		Remote = 'remote-spacer',
+		Local = 'local-spacer',
+		LocalAndRemote = 'local-and-remote-spacer'
+	}
+
 	$: mappedRemoteCommits =
 		$remoteCommits.length > 0
-			? [...$remoteCommits.map(transformAnyCommit), { id: 'remote-spacer' }]
+			? [...$remoteCommits.map(transformAnyCommit), { id: LineSpacer.Remote }]
 			: [];
 	$: mappedLocalCommits =
 		$localCommits.length > 0
-			? [...$localCommits.map(transformAnyCommit), { id: 'local-spacer' }]
+			? [...$localCommits.map(transformAnyCommit), { id: LineSpacer.Local }]
+			: [];
+	$: mappedLocalAndRemoteCommits =
+		$localAndRemoteCommits.length > 0
+			? [...$localAndRemoteCommits.map(transformAnyCommit), { id: LineSpacer.LocalAndRemote }]
 			: [];
 
 	$: lineManager = lineManagerFactory.build(
 		{
 			remoteCommits: mappedRemoteCommits,
 			localCommits: mappedLocalCommits,
-			localAndRemoteCommits: $localAndRemoteCommits.map(transformAnyCommit),
+			localAndRemoteCommits: mappedLocalAndRemoteCommits,
 			integratedCommits: $integratedCommits.map(transformAnyCommit)
 		},
 		!isRebased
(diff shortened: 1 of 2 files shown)


apps/desktop/src/lib/commit/CommitList.svelte
```diff
@@ -49,20 +49,30 @@
 	const reorderDropzoneManagerFactory = getContext(ReorderDropzoneManagerFactory);
 	const gitHost = getGitHost();
 
+	enum LineSpacer {
+		Remote = 'remote-spacer',
+		Local = 'local-spacer',
+		LocalAndRemote = 'local-and-remote-spacer'
+	}
```
Caleb-T-Owens, line 56: If an enum is warented, where should this enum belong?

apps/desktop/src/lib/commit/CommitList.svelte
```diff
@@ -217,9 +242,9 @@
 					/>
 				{/each}
 
-				<CommitAction bottomBorder={hasRemoteCommits}>
+				<CommitAction bottomBorder={hasRemoteCommits} hideAction={$remoteCommits.length > 0}>
 					{#snippet lines()}
```
Caleb-T-Owens: If we want to hide the action, does it make more sense to not render the commit action at all, or to ask it to hide part of itself?

Caleb-T-Owens (changes requested):
I think we might need to more carefully think about how we want to represent this state.

At the minute this appears to be saying "push these already pushed changes", and could even be surprising when the upstream changes get lost. Should we instead have the "integrate upstream" button be a multi-state button which instead has a "overwrite" option which does a force push?

Caleb-T-Owens:
The current logic also looks like it strongly prefers to hide the local push button, which doesn't seem desirable as the idea behind the design is that the button is positioned next to what it is acting upon.

```
We've got two inputs:

L := list of local commits
RL := list of local and remote commits

We've got two outputs:

Push1 := push button positioned below local commits
Push2 := push button positioned below local and remote commits

And we're saying

If L length > 0:
 - Display Push1 unless RL length > 0
If RL length > 0:
 - Display Push2
```

apps/desktop/src/lib/commit/CommitList.svelte
estib-vega: I'd say it is, just to mitigate the risk of typo-bugs since the `CommitData` id's are free-form strings. The compiler wouldn't catch typos otherwise.

Alternatively I could just turn this into simple string constants, but I initially chose enum for some added context.

> where should this enum belong?
>

Good question, I didn't immediately find a right place for it other than this module as it's only specific to the `CommitList` component. What do you think?

apps/desktop/src/lib/commit/CommitList.svelte
estib-vega: It does 👍 

I initially chose this, in order to easily keep the virtual branch lines and right padding while hiding the button, but then we have two different levels at which we're hiding components, which is indeed not great.

I'll change this

estib-vega:
> I think we might need to more carefully think about how we want to represent this state.
> 
> At the minute this appears to be saying "push these already pushed changes", and could even be surprising when the upstream changes get lost. Should we instead have the "integrate upstream" button be a multi-state button which instead has a "overwrite" option which does a force push?

That sounds like an interesting alternative. Also I agree, it's not immediately clear that force pushing will get rid of that remote commit. I'll check also with @PavelLaptev to see what he thinks

estib-vega:
> The current logic also looks like it strongly prefers to hide the local push button, which doesn't seem desirable as the idea behind the design is that the button is positioned next to what it is acting upon.
> 
> ```
> We've got two inputs:
> 
> L := list of local commits
> RL := list of local and remote commits
> 
> We've got two outputs:
> 
> Push1 := push button positioned below local commits
> Push2 := push button positioned below local and remote commits
> 
> And we're saying
> 
> If L length > 0:
>  - Display Push1 unless RL length > 0
> If RL length > 0:
>  - Display Push2
> ```

This is by design, but open to changing that.
The motivation behind choosing not to display the local force-push button is so that we only display one button at a time, since both buttons will perform the same action.

estib-vega:
Talked to @Caleb-T-Owens, will change this so that we prefer the local force-push button instead of the local-and-remote one

estib-vega:
Screenshot 2024-09-09 at 16 34 20

@Caleb-T-Owens done :)

mtsgrd (approved):

Caleb-T-Owens (approved):
Danke!

PR #4849, opened 2024-09-07
refactor clone for clarity

Byron:
This PR offers an alternative implementation for `git_clone_repository` which is closer to the original vision on how the API should be used.

The main issue was the need to turn everything into a frontend `Error` which could only convert from `anyhow::Error`. This is for good reason, 
but in this case it was in the way resolved with an `UnmarkedError` which can convert from any other error, while serializing such that the frontend can handle it.

### Motivation

I used this the `git2` version of the code along with the `gitoxide` version as an example in my [upcoming GitMerge talk](https://github.com/Byron/gitoxide/pull/1588) for how much simpler `git2` can be, in the real world, but couldn't resist to try to turn it into the 'best possible version of itself'.

### DIFF
--- a/crates/gitbutler-tauri/src/error.rs
+++ b/crates/gitbutler-tauri/src/error.rs
@@ -13,13 +13,46 @@
 //!
 //! The values in these fields are controlled by attaching context, please [see the `error` docs](gitbutler_error::error))
 //! on how to do this.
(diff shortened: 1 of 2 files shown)


Caleb-T-Owens (approved):
Cool! Thanks for this. (Looking forwards to your talk BTW)

PR #4850, opened 2024-09-08
fix: refactor pr template path input to pre-filled `Select` instead of `TextBox`

ndom91:
## ☕️ Reasoning

- Based on some [customer feedback](https://github.com/gitbutlerapp/gitbutler/pull/4814#issuecomment-2335829658), it seemed smarter to just check the few places where each git host (GitHub in this instance), allows PR Templates to be and show those options in a dropdown. Definitely feels like a much nicer UX and is much less error prone to a user typing in a path themselves.

## 🧢 Changes

- Add tauri rust cmd to read the `${project.path}/.github/` dir recursively
	- Then check those contents against the predefined filenames / directories which constitute valid PR templates
	- If found, add them to a vec and return that to the FE
- Show only valid available template options in a `<Select />`
- Modified the saving of these options from global `Persisted` variables to being saved on a project level in the `projects.json`
- Instead of passing the PR Template settings into the githost at initialisation, they're passed in as arguments to the `createPr()` fn now. This allows reading the current setting value at time of btn click (better) and logically seems more appropriate this way.


crates/gitbutler-project/src/project.rs
```diff
@@ -96,6 +96,18 @@ pub struct Project {
...
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone, Default)]
+pub struct GitHostSettings {
+    #[serde(default)]
+    pub host_type: Option<String>,
+    #[serde(default)]
+    pub use_pull_request_template: Option<bool>,
+    #[serde(default)]
+    pub pull_request_template_path: Option<String>,
```
mtsgrd: Could you put a `#[serde(rename_all = "camelCase")]` right after the `#[derive(...)]`?

crates/gitbutler-project/src/storage.rs
```diff
@@ -28,6 +28,7 @@ pub struct UpdateRequest {
     pub use_diff_context: Option<bool>,
     pub snapshot_lines_threshold: Option<usize>,
     pub succeeding_rebases: Option<bool>,
+    pub git_host: Option<GitHostSettings>,
```
mtsgrd, line 31: This struct should ideally also have the camelCase transformation, but outside the scope of this pr.

apps/desktop/src/lib/gitHost/github/githubPrService.ts
```diff
@@ -20,21 +24,25 @@ export class GitHubPrService implements GitHostPrService {
...
+	async createPr({
+		title,
+		body,
+		draft,
+		useTemplate,
+		templatePath
+	}: CreatePullRequestArguments): Promise<PullRequest> {
 		this.loading.set(true);
-		const request = async (pullRequestTemplate: string | undefined = '') => {
+		const request = async (prBody: string | undefined = '') => {
```
mtsgrd: I'm looking at the default value and I'm wondering why an empty string is better than undefined?

apps/desktop/src/lib/gitHost/github/githubPrService.ts
```diff
@@ -67,11 +74,7 @@ export class GitHubPrService implements GitHostPrService {
 		throw lastError;
 	}
 
-	async fetchPrTemplate() {
-		const path = this.pullRequestTemplatePath
-			? get(this.pullRequestTemplatePath)
-			: DEFAULT_PULL_REQUEST_TEMPLATE_PATH;
-
+	async fetchPrTemplate(path = DEFAULT_PULL_REQUEST_TEMPLATE_PATH) {
```
mtsgrd: Since this default value is never used, how about we remove it?

crates/gitbutler-project/src/project.rs
ndom91: Brilliant, applied :+1:

apps/desktop/src/lib/backend/github.ts
```diff
@@ -5,10 +5,33 @@ export type Verification = {
...
+
+	const availableTemplates: PullRequestTemplatePaths[] | undefined = await invoke(
+		'get_available_github_pr_templates',
+		{
+			path: targetPath
+		}
+	);
+
+	return availableTemplates;
+}
```
mtsgrd: Could this live in the gitHost code somewhere? In here we wouldn't be able to mock it.

apps/desktop/src/lib/gitHost/github/githubPrService.ts
ndom91: Hmm yeah actually the octokit `body` arg can take `undefined` too, so no need for a default here

apps/desktop/src/lib/backend/github.ts
```diff
@@ -5,10 +5,33 @@ export type Verification = {
...
+	// TODO: Find a workaround to avoid this dynamic import
+	// https://github.com/sveltejs/kit/issues/905
+	const path = await import('@tauri-apps/api/path');
+	const targetPath = await path.join(projectPath, '.github');
+
+	const availableTemplates: PullRequestTemplatePaths[] | undefined = await invoke(
+		'get_available_github_pr_templates',
+		{
+			path: targetPath
+		}
```
mtsgrd: Maybe reformat as `{ path: targetPath }`?

ndom91: Applied :+1:

apps/desktop/src/lib/backend/github.ts
ndom91: Moved it into the gitHosts themselves. Not 100% sure about this though, lets take a look together later if you have a few minutes :+1:

crates/gitbutler-tauri/src/projects.rs
```diff
@@ -96,6 +97,19 @@ pub mod commands {
...
+    #[instrument(skip(projects))]
+    pub fn get_pr_template_contents(
+        projects: State<'_, Controller>,
+        relative_path: &path::Path,
+        project_id: ProjectId,
+    ) -> Result<String, Error> {
+        let project = projects.get(project_id)?;
+        let template_path = project.path.join(relative_path);
+
+        Ok(read_file_from_workspace(template_path.as_path())?)
```
Caleb-T-Owens, line 111: Talking to Nico, we've acknowledged that a path could back out of the directory using `../../../`, but due to the size of the PR, we'd rather merge in now, and improve the security in a follow up PR.

Caleb-T-Owens (approved):

estib-vega (approved):
👍

PR #4852, opened 2024-09-09
feat: Display binary image and size

### LINKED ISSUE #2752: Display images
Rather than 'Binary content not shown', an image preview would be nice.

### LINKED ISSUE #3093: Show file size for binary content
It feels like we could show more information about binary files in this UI. Any thoughts?
image

Zamoca42:
Hello team! First, I want to express my gratitude for this fantastic app. 
I've been using it regularly, and it's been incredibly helpful in my workflow.

I've prepared this pull request after reviewing issues #2752 and #3093.

## ☕️ Reasoning

The current implementation doesn't provide image previews and file sizes for binary files, especially image files.

This PR aims to implement image previews for binary files, particularly images, in the FileDiff component and display file sizes for binary files.

## 🧢 Changes

- Binary file handling and information improvement:
   - Implemented image display for image files in the FileDiff component 
   - Implemented `getBinFileSize` to retrieve binary file sizes and updated the FileDiff component accordingly
   - Modified Tauri configuration to allow file access in user's home directory

### Image Preview

스크린샷 2024-09-08 오후 10 05 19

### Binary File Size

file-size

## 🎫 Affected issues

Fixes: #2752
Fixes: #3093

### DIFF
--- a/apps/desktop/src/lib/file/FileDiff.svelte
+++ b/apps/desktop/src/lib/file/FileDiff.svelte
@@ -4,11 +4,13 @@
 	import LargeDiffMessage from '$lib/shared/LargeDiffMessage.svelte';
 	import { getContext } from '$lib/utils/context';
 	import { getFileExtension } from '$lib/utils/filePath';
+	import { getBinFileSize } from '$lib/utils/fileSize';
 	import { computeAddedRemovedByHunk } from '$lib/utils/metrics';
 	import { getLocalCommits, getLocalAndRemoteCommits } from '$lib/vbranches/contexts';
 	import { getLockText } from '$lib/vbranches/tooltip';
 	import Icon from '@gitbutler/ui/Icon.svelte';
 	import Tooltip from '@gitbutler/ui/Tooltip.svelte';
+	import { join } from '@tauri-apps/api/path';
 	import { convertFileSrc } from '@tauri-apps/api/tauri';
 	import { onMount } from 'svelte';
 	import type { HunkSection, ContentSection } from '$lib/utils/fileSections';
@@ -37,17 +39,21 @@
 		readonly = false
 	}: Props = $props();
 
+	let fileSize = $state('...');
 	let imageSrc = $state('');
 	const project = getContext(Project);
 
 	onMount(async () => {
-		if (isImage) {
-			const fullPath = `${project.path}/${filePath}`;
-			try {
-				imageSrc = convertFileSrc(fullPath);
-			} catch (error) {
-				console.error('Failed to convert image path:', error);
-				console.error('Attempted image source:', fullPath);
+		if (isBinary || !isLarge) {
+			const fullPath = await join(project.path, filePath);
+			fileSize = await getBinFileSize(fullPath);
+			if (isImage) {
+				try {
+					imageSrc = convertFileSrc(fullPath);
+				} catch (error) {
+					console.error('Failed to convert image path:', error);
+					console.error('Attempted image source:', fullPath);
+				}
 			}
 		}
 	});
@@ -77,11 +83,14 @@
 
 <div class="hunks">
 	{#if isBinary}
-		{#if isImage}
-			<img src={imageSrc} alt="Binary file" class="binary-image" />
-		{:else}
-			Binary content not shown
-		{/if}
+		<div class="binary-info">
+			{#if isImage}
+				<img src={imageSrc} alt="Binary file" class="binary-image" />
+			{:else}
+				<p>Binary content not shown</p>
+			{/if}
+			<p>file size: {fileSize}</p>
+		</div>
 	{:else if isLarge}
 		Diff too large to be shown
 	{:else if sections.length > 50 && !alwaysShow}
@@ -152,6 +161,13 @@
 	.binary-image {
 		max-width: 100%;
 		height: auto;
+	}
+
+	.binary-info {
+		display: flex;
+		flex-direction: column;
+		align-items: center;
+		gap: 10px;
 	}
 
 	.added-removed {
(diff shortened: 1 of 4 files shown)


apps/desktop/src/lib/file/FileDiff.svelte
```diff
@@ -1,11 +1,18 @@
...
+	import { getContext } from '$lib/utils/context';
+	import { getFileExtension } from '$lib/utils/filePath';
+	import { getBinFileSize } from '$lib/utils/fileSize';
 	import { computeAddedRemovedByHunk } from '$lib/utils/metrics';
 	import { getLocalCommits, getLocalAndRemoteCommits } from '$lib/vbranches/contexts';
 	import { getLockText } from '$lib/vbranches/tooltip';
 	import Icon from '@gitbutler/ui/Icon.svelte';
 	import Tooltip from '@gitbutler/ui/Tooltip.svelte';
+	import { join } from '@tauri-apps/api/path';
+	import { convertFileSrc } from '@tauri-apps/api/tauri';
```
krlvi, line 14: Thank you so much for this contribution! I had no idea that this `convertFileSrc` API existed, and it does exactly what we need!

crates/gitbutler-tauri/tauri.conf.json
```diff
@@ -13,7 +13,7 @@
 		"allowlist": {
 			"fs": {
 				"readFile": true,
-				"scope": ["$APPCACHE/archives/*", "$RESOURCE/_up_/scripts/*"]
+				"scope": ["$APPCACHE/archives/*", "$RESOURCE/_up_/scripts/*", "$HOME/**"]
```
krlvi, line 16: Hmm, I wonder if there is a way to not give the frontend javascript code access to `$HOME`, from a security standpoint

apps/desktop/src/lib/file/FileDiff.svelte
Zamoca42: Thank you for your review! At first, I attempted to access local images using file://, but I encountered the ‘Not allowed to load local resource’ error. After that, I found and referred to https://github.com/tauri-apps/tauri/discussions/1438#discussioncomment-10424231.

crates/gitbutler-tauri/tauri.conf.json
Zamoca42: Following up on the above, to fetch local images using `convertFileSrc`, the linked issue suggests using "**". Since I couldn’t find a better reference, I referred to [tauri docs](https://tauri.app/v1/api/js/fs/#security) and decided to use the `$HOME` directory near the project’s path.
Currently, the Tauri configuration only allows directory read access, which seemed acceptable at the time.

I suggest implementing protection against directory traversal attacks and unintended path access by validating paths through the validateFilePath function. If you have any better alternatives or suggestions, I would appreciate it if you could let me know.

**fileDiff.svelte**

```ts 

import { getFileExtension, validateFilePath } from '$lib/utils/filePath';

onMount(async () => {
    if (isBinary || !isLarge) {
      const validatedPath = await validateFilePath(project.path, filePath);
      if (validatedPath) {
        fileSize = await getBinFileSize(validatedPath);
        if (isImage) {
          try {
            imageSrc = convertFileSrc(validatedPath);
          } catch (error) {
            console.error('Failed to convert image path:', error);
            console.error('Attempted image source:', validatedPath);
          }
        }
      } else {
        console.error('Invalid file path:', filePath);
        fileSize = 'Invalid path';
      }
    }
  });
```

**utils/filePath.ts**

```ts
import { normalize, isAbsolute, join, homeDir } from '@tauri-apps/api/path';

//...

export async function validateFilePath(projectPath: string, filePath: string): Promise<string | false> {
  try {
    const normalizedProjectPath = await normalize(projectPath);
    const fullPath = await join(normalizedProjectPath, filePath);
    const normalizedFullPath = await normalize(fullPath);
    const userHomeDir = await homeDir();

    // Check if it's an absolute path
    if (!(await isAbsolute(normalizedFullPath))) {
      return false;
    }

    // Check if the path is within the project directory
    if (!normalizedFullPath.startsWith(normalizedProjectPath) && !normalizedFullPath.startsWith(userHomeDir))  {
      return false;
    }

    // Check file extension
    const allowedExtensions = ['txt', 'pdf', 'jpg', 'png', 'gif', 'svg', 'webp', 'ttf'];
    const extension = getFileExtension(filePath);
    if (extension && !allowedExtensions.includes(extension)) {
      return false;
    }

    return normalizedFullPath; // Return the normalized full path if valid
  } catch (error) {
    console.error('Error validating file path:', error);
    return false;
  }
}
```

mtsgrd: Yeah from a security standpoint we can't allow javascript access to the $HOME directory, but I think @ndom91 is adding an api for fetching files from the workspace?

ndom91: Hey yeah so we have a very similar use-case that we're working on in [this PR](https://github.com/gitbutlerapp/gitbutler/pull/4850). We want to get the contents of every pull request template file from the local repo / project dir. The current implementation of the `read_workspace_file` fn in that PR is not safe enough and we plan on refactoring the implementation over the next few days. 

The plan is to update the function we'll expose to the FE to `invoke` to accept a relative path. Relative to the project (repositories) root. Well then use libgit2 to `cat` the target file. This way the FE can't access anything outside of the project / repo root directory and by leveraging the git cat-file, we'll also effectively disallow access to the `.git` dir and also any `.gitignore`d files. 

After talking it over with the team a bit, we figured this is a pretty safe way to expose a readFile fn to the FE. 

I can ping again here once that's complete and the read file fn is merged to master.

Byron: I love the outcome!
Just to share my thoughts, I think this will work as GB effectively adds everything that it sees in the working tree, minus large files, to the repository right away to create virtual-branch trees, which makes everything available as Git blobs as well which are referenced from these trees.

Personally, I hope that one day this can be changed to only do this when absolutely necessary or even never *implicitly* do it at all. By that time, there probably must be a way to fetch files directly from the worktree, and as Git has to protect itself from similar abuse, there is [a utility](https://docs.rs/gix-validate/latest/gix_validate/path/fn.component.html) to prevent. It's quite exhaustive as well.

PR #4855, opened 2024-09-09
fix: add crates to pnpm-workspace to avoid Vercel CI

ndom91:
## ☕️ Reasoning

- Based on Vercel support request, they said that changes to `crates/*` (for example), which are not a  part of our `pnpm-workspace.yml` pcakages, are "conservatively assumed to be a global change that could impact anything, and so we rebuild all applications."
- The recommended fix, although a bit unconventional, would be to add a `package.json` (only needs a `name` field) to `crates/package.json` and then add `crates/` to the `pnpm-workspace.yml`, as is done here.
- Are we cool with putting a `package.json` in `crates/`?

## 🧢 Changes

- Add `crates/package.json` with `name` field
- Add `crates/` to `pnpm-workspace.yml`


ndom91:
Note, this PR didn't even show the Vercel comment until I did a merge commit from `master`, so seems promising.

crates/package.json
```diff
@@ -0,0 +1,3 @@
+{
+	"name": "crates"
+}
```
mtsgrd: Can we add a comment here explaining why we have a package.json in `crates`?

ndom91: Good idea, added :+1:

ndom91:
This PR is on hold for now. Waiting on further support responses and/or decision on deploying via GHA instead

ndom91:
@krlvi so vercel answered and said that in the case of adding or modifying a file not within one of the pnpm workspaces, like for example the root `Cargo.toml`, that woudl trigger a build and subsequent comment on the PR from the Vercel bot.

I think realistically though, how often are we changing files that are not in one of the pnpm workspaces (including `/crates` if we were to add it via this PR)? (i.e. outside of `packages/*`, `apps/*` and `crates/*`).

Therefore, if you guys are okay with having a small `package.json` under `crates/package.json`, then I'd recommend we go ahead with this PR now to reduce the amount of comments and builds already significantly on all the rust-only branches / PRs.

crates/gitbutler-tauri/tauri.conf.json
```diff
@@ -1,7 +1,7 @@
 {
 	"build": {
-		"beforeDevCommand": "pnpm dev:internal-tauri",
-		"beforeBuildCommand": "[ $CI = true ] || pnpm build:desktop -- --mode development",
+		"beforeDevCommand": "cd ../ && pnpm dev:internal-tauri",
+		"beforeBuildCommand": "[ $CI = true ] || cd ../ && pnpm build:desktop -- --mode development",
```
Caleb-T-Owens, line 4: Why is this change required?

ndom91: Yeah good question, I don't remember exaclty but I assume with the additional `package.json` in the `crates/` dir, it ends up looking for the npm scripts there and fails to find them. Instead of looking one more dir up as it normally would when triyng to find the nearest `package.json`. Double checking now though :+1:

Caleb-T-Owens: Fair enough;

Would it make sense to use say `pushd ../ && pnpm <command>; popd`?

ndom91: Yeha looks like that's it :ok_hand:

Caleb-T-Owens (approved):
Approved, but I'd appreciate it if we switched from `cd` to `pushd` and `popd` in case it ends up switching a directory that another script or user doesn't expect?

ndom91:
> Approved, but I'd appreciate it if we switched from `cd` to `pushd` and `popd` in case it ends up switching a directory that another script or user doesn't expect?

Hmm so as far as I know these npm scripts are run as subprocesses, right, so this would only be helpful if someone is modifying the npm script itself, at which point they'd clearly see that we're `cd`-ing at the beginning and realize the change of `cwd` in this context. So therefore, doing `npm run .. && [new stuff here]` would still execute the second part in the same `cwd` as expected, right? Idk if `pushd`/`popd` would really be of any value here. This one npm script doesn't do anything after the one action after the directory change, so if we were to apply it here right now, we'd just have a `pushd` and no `popd`, which feels like unnecessarily complicating things, no?

PR #4857, opened 2024-09-09
IpcNameNormalizationService: Add timed cache

estib-vega:
Cache the branch name normalization responses with a stale time of 5 minutes

Motivation behind it is that I think the returned values are simple and consistent enough to deserve some front-end caching. What do you think?

### DIFF
--- a/apps/desktop/src/lib/branches/nameNormalizationService.ts
+++ b/apps/desktop/src/lib/branches/nameNormalizationService.ts
@@ -1,15 +1,39 @@
 import { invoke as ipcInvoke } from '$lib/backend/ipc';
 import { buildContext } from '$lib/utils/context';
+import { isStr } from '$lib/utils/string';
+
+const CACHE_STALE_TIME = 1000 * 60 * 5;
 
 export interface NameNormalizationService {
 	normalize(branchName: string): Promise<string>;
 }
 
+interface TimestampedBranchName {
+	name: string;
+	timestamp: number;
+}
+
 export class IpcNameNormalizationService implements NameNormalizationService {
-	constructor(private invoke: typeof ipcInvoke) {}
+	private cache: Map<string, TimestampedBranchName>;
+
+	constructor(private invoke: typeof ipcInvoke) {
+		this.cache = new Map();
+	}
 
 	async normalize(branchName: string): Promise<string> {
-		return await this.invoke('normalize_branch_name', { name: branchName });
+		const now = Date.now();
+		const cached = this.cache.get(branchName);
+		if (cached && now - cached.timestamp < CACHE_STALE_TIME) {
+			return cached.name;
+		}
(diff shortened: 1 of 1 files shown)


Caleb-T-Owens (changes requested):
Could we re-use the `lscache` library used elsewhere in the app rather than building our own cache implementation?

Caleb-T-Owens:
Also, how much of a performance differences does this make? I was under the impression the rust call was basically negligible.

estib-vega:
> Also, how much of a performance differences does this make? I was under the impression the rust call was basically negligible.

Don't think is a significant gain really, just thought it might be a nice simple improvement so that we start saving up on some backend calls.

Not really a huge improvement though, so if you feel strongly against it, I can dismiss this :)

Caleb-T-Owens:
> Not really a huge improvement though, so if you feel strongly against it, I can dismiss this :)

I'm just cautious that we're increasing the complexity of the function without any material gain. @mtsgrd what do you think?

estib-vega:
> > Not really a huge improvement though, so if you feel strongly against it, I can dismiss this :)
> 
> I'm just cautious that we're increasing the complexity of the function without any material gain. @mtsgrd what do you think?

That's a fair point 👍

Byron:
Somehow I really like the idea of memoization/caching of expensive calls, but I also feel that the backend should be in a better position to determine what that should be to avoid staleness.

Another take on this is that whenever the UI is slow or slowing due to the backend, then ideally the backend can be made so it's faster, or other inefficiencies are removed to fix the slowdown.

The frontend is calling this function a lot, but it's as fast as it gets (now) I'd be surprised if anything measurable (or meaningful to the user) comes out of that.

If you want to have a better idea where it might be worth to make things faster, try this:

```
GITBUTLER_PERFORMANCE_LOG=1 LOG_LEVEL=debug pnpm tauri dev --release
```

I am always using it to find candidates for optimization, even before profiling with Instruments.

estib-vega:
> Somehow I really like the idea of memoization/caching of expensive calls, but I also feel that the backend should be in a better position to determine what that should be to avoid staleness.
> 
> 
> 
> Another take on this is that whenever the UI is slow or slowing due to the backend, then ideally the backend can be made so it's faster, or other inefficiencies are removed to fix the slowdown.
> 
> 
> 
> The frontend is calling this function a lot, but it's as fast as it gets (now) I'd be surprised if anything measurable (or meaningful to the user) comes out of that.
> 
> 
> 
> If you want to have a better idea where it might be worth to make things faster, try this:
> 
> 
> 
> ```
> 
> GITBUTLER_PERFORMANCE_LOG=1 LOG_LEVEL=debug pnpm tauri dev --release
> 
> ```
> 
> 
> 
> I am always using it to find candidates for optimization, even before profiling with Instruments.

That's really useful!! Also good points. I'll try that out 👍

Also I'll close this PR

PR #4858, opened 2024-09-09
PR service: Initialize it only with the base branch

estib-vega:
Update: 
Initialize the PR service only with a base branch, allowing it to set the upstream branch when actually creating the PR.

~In order to initialize the PR service, use the normalized branch display name as shown in the branch service~

This fixes the issue in which creating a new branch would not enable the PR button

### More context
A recent change of mine, made it so that in order to enable the PR button, the PR service would need to be present.
The issue was, that the PR service is not present if there is no `upstream` branch set.

~This change fixes this, in that we'll generate a normalized branch name in order to initialize the PR service.
The user can still override this at any point by setting the remote name (as is normal) if so desired.~

This change fixes this, in that pushing the virtual branch will return the upstream name, that can then be directly used as a parameter when creating the PR

### DIFF
--- a/apps/desktop/src/lib/branch/BranchLane.svelte
+++ b/apps/desktop/src/lib/branch/BranchLane.svelte
@@ -1,7 +1,6 @@
 <script lang="ts">
 	import BranchCard from './BranchCard.svelte';
 	import { Project } from '$lib/backend/projects';
-	import { BaseBranch } from '$lib/baseBranch/baseBranch';
 	import { projectLaneCollapsed } from '$lib/config/config';
 	import FileCard from '$lib/file/FileCard.svelte';
 	import { getGitHost } from '$lib/gitHost/interface/gitHost';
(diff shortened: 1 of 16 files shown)


ndom91 (approved):
LGTM :+1:

Caleb-T-Owens:
@ndom91 I requested some changes in person yesterday.

apps/desktop/src/lib/branch/BranchLane.svelte
```diff
@@ -40,14 +41,33 @@
...
+	$effect(() => {
+		if (!baseBranchName) {
+			return;
+		}
+
+		if (!upstreamName) {
+			nameNormalizationService
+				.normalize(branch.displayName)
+				.then((normalizedName) => {
+					console.log('normalizing branch name', normalizedName);
```
mtsgrd: 👀

PR #4867, opened 2024-09-10
Update gix-path to version 0.10.11

krlvi:


Byron:
I was a bit late for the party, and probably should have thought about it before some tool yells at you.
Speaking of tool that yells at you, `cargo audit` might be something to adopt to get early warning for cases like these, i.e. when an advisory is out for a dependency.
Even though technically it makes CI fail occasionally, it's typically easily fixable or overridable, depending on the kind of issue.

PR #4873, opened 2024-09-11
Issues

Caleb-T-Owens:
This is a highly experimental feature based on a conversation with kiril about wanting to create notes and share them, without the clunk of creating an issue, and potentially how we could incorporate features like chat into the app.

`$mod+i` opens the issue modal
`$mod+k` opens the topic modal

Screenshot 2024-09-11 at 10 26 02

This is hidden behind an experimental feature flag


krlvi:
is the shortcut key also active only if the feature is enabled?

Caleb-T-Owens:
> is the shortcut key also active only if the feature is enabled?

Yes @krlvi

krlvi (approved):
:shipit:

PR #4876, opened 2024-09-11
fix: File key navigation and selection

estib-vega:
## ☕️ Reasoning
Partially mentioned here: https://github.com/gitbutlerapp/gitbutler/issues/4734#issuecomment-2307668787

This fixes the file list navigation to work as expected

## 🧢 Changes
- Navigating with the up or down arrow keys works as expected
- Holding down shift while navigating also works as expected
- If selecting multiply files, up-arrow will de-select everything except the top file
- If selecting multiply files, down-arrow will de-select everything except the bottom file


PavelLaptev (approved):
Works good 👍 thanks @estib-vega

PR #4878, opened 2024-09-11
fix: `createPr` arguments order

ndom91:
## ☕️ Reasoning

- baseBranch and upstreamBranchName arguments were swapped by accident

## 🧢 Changes

- ~~Swap args back into correct order~~
- Refactor createPr args to options object


estib-vega (approved):
Looks good!

PR #4883, opened 2024-09-11
fix: improve deeply nested token rendering

ndom91:
## ☕️ Reasoning

- Rendering other items inside nested items (such as `List` -> `ListItem` -> `Link`) wasn't working

### Before

> image

### After

> image


estib-vega (approved):
Looks good to me!

PR #4884, opened 2024-09-11
Add some nice keyboard shortcuts for staging and selecting

estib-vega:
## ☕️ Reasoning
As seen in https://github.com/gitbutlerapp/gitbutler/issues/4734

**DISCLAIMER**
Not all items in that issue are implemented in this PR.

It would be great if we and stage/un-stage files using only the keys!

## 🧢 Changes
- Clicking on `'Escape'` will close the commit message input field (both the modal and the lane field)
- If **at least one file list item is selected**,  clicking `Ctrl or cmd + 'a'` will select all files in the list
- If **at least one file list item is selected**, clicking the `space-bar` will open the commit message input from the lane.
- If **at least one file list item is selected** and the **commit message input is open**,  clicking the `space-bar` will toggle the staging of the selected files.


apps/desktop/src/lib/file/BranchFilesList.svelte
```diff
@@ -22,23 +25,79 @@
...
+	function handleSpace() {
+		if (commitDialogExpanded === undefined) return;
+
+		// Start commit
+		if (!$commitDialogExpanded) {
+			$commitDialogExpanded = true;
+			return;
+		}
+
+		// Stage/unstage files
```
mtsgrd: Maybe remove these comments as they are not super helpful.

apps/desktop/src/lib/vbranches/ownership.ts
```diff
@@ -203,3 +205,62 @@ export class SelectedOwnership {
...
+		.map((id) => unstringifyFileKey(id))
+		.map((id) => params.files.find((f) => f.id === id))
+		.filter((f): f is AnyFile => !!f);
+
+	if (selectedFiles.length === 0) return;
+
+	// Only one file selected
+	//
+	// Select all hunks if none are selected,
+	// otherwise ignore all hunks
```
mtsgrd: Any chance you could format and punctuate this a bit better? I see one comma, and two questionable line breaks 😂

mtsgrd (approved):

apps/desktop/src/lib/vbranches/ownership.ts
estib-vega: Don't, you, like

.them?

estib-vega: Haha, I'll just remove them 👍

PR #4885, opened 2024-09-11
fix discarding of deleted and created files

krlvi:
This fixes 3 issues:
- new files that are empty are incorrectly shown as modified (should be added)
- discarding a newly added file did not work
- discarding file deletions did not work

### DIFF
--- a/crates/gitbutler-diff/src/write.rs
+++ b/crates/gitbutler-diff/src/write.rs
@@ -1,6 +1,6 @@
 #[cfg(target_family = "unix")]
 use std::os::unix::prelude::PermissionsExt;
-use std::{borrow::Borrow, path::PathBuf};
+use std::{borrow::Borrow, fs, path::PathBuf};
 
 use anyhow::{anyhow, Context, Result};
 use bstr::{BString, ByteSlice, ByteVec};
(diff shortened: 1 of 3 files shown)


crates/gitbutler-diff/src/write.rs
```diff
@@ -62,7 +63,13 @@ where
             && hunks[0].diff_lines.contains_str(b"Subproject commit");
 
         // if file exists
-        if full_path.exists() {
+        if full_path.exists() || allow_new_file {
+            if hunks.len() == 1 && hunks[0].change_type == crate::ChangeType::Deleted {
+                // File was created but now that hunk is being discarded with an inversed hunk
+                builder.remove(rel_path);
+                fs::remove_file(full_path.clone())?;
```
Byron: Can we be sure here that the file actually exists on disk? As far as I am concerned, hunks can be arbitrary and `allow_new_file` could be `true`.

crates/gitbutler-diff/src/write.rs
```diff
@@ -62,7 +63,13 @@ where
             && hunks[0].diff_lines.contains_str(b"Subproject commit");
 
         // if file exists
-        if full_path.exists() {
+        if full_path.exists() || allow_new_file {
+            if hunks.len() == 1 && hunks[0].change_type == crate::ChangeType::Deleted {
```
Byron: It feels odd to me that the single-hunk case is so special. I am sure there is a good reason for it, and I'd hope it could be explained as a comment in code. Ideally of course, there would be a test for it (and I am saying this knowing that eventually I will have to touch all of this).

(see also my comment below, I think I get it now)

crates/gitbutler-diff/src/write.rs
```diff
@@ -178,6 +185,22 @@ where
                 let new_blob_oid = git_repository.blob(blob_contents.as_bytes())?;
                 // upsert into the builder
                 builder.upsert(rel_path, new_blob_oid, filemode);
+            } else if !full_path.exists()
+                && hunks.len() == 1
+                && hunks[0].change_type == crate::ChangeType::Added
+            {
+                // File was deleted but now that hunk is being discarded with an inversed hunk
```
Byron, line 199: Ah, I see now that the single-hunk mode is happening when the UI discards things.

Byron (commented):
Thanks for the invite for review - I could use it after working on merging and feeling like I know nothing while I bite myself into it 😅.

There is only one question left, otherwise it looks good (except that there are no tests for the added portions, but it's late and they will be added eventually I am sure).

Byron (approved):

PR #4886, opened 2024-09-11
UI fixes: Hunk preview and checkboxes

PavelLaptev:
Changes:
- Remove duplicated hunk info and replace bare text and icons with tooltips with infomessages
- There was a shift issue with borders on scroll
- Fix the hunk header paddings
- Replaced `box-shadow` with borders. Looks like it's not longer an issue
- Draggable icon update
- Selected hunk colors update

Before:

https://github.com/user-attachments/assets/e947621b-c352-4fd3-8dfd-125d0997b2f6

After:

https://github.com/user-attachments/assets/2d30826b-d4ef-42f9-ad31-690bab8ad3ed

### DIFF
--- a/apps/desktop/src/lib/file/FileDiff.svelte
+++ b/apps/desktop/src/lib/file/FileDiff.svelte
@@ -5,8 +5,6 @@
 	import { computeAddedRemovedByHunk } from '$lib/utils/metrics';
 	import { getLocalCommits, getLocalAndRemoteCommits } from '$lib/vbranches/contexts';
 	import { getLockText } from '$lib/vbranches/tooltip';
-	import Icon from '@gitbutler/ui/Icon.svelte';
-	import Tooltip from '@gitbutler/ui/Tooltip.svelte';
 	import type { HunkSection, ContentSection } from '$lib/utils/fileSections';
 
 	interface Props {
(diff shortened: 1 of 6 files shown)


estib-vega:
@PavelLaptev Thanks for the fixes! 

One thing though, I see something familiar on the after video haha.
I was fighting that weird table borders behavior in that scrolling would move the content but not all borders. 

I think I know how to fix that. 

Would you be ok if I add it as a commit on top?

PavelLaptev:
Yes, let's try. Is it too hacky? Borders was the reason why I gave up  and revert the old implementation, didn't want to over complicate things.

estib-vega:
> Is it too hacky?
>

Some parts are just about re-adding the inset box-shadows and other about having absolutely positioned divs.
The latter is the part that requires dynamically set values.

What do you think?

estib-vega:
https://github.com/user-attachments/assets/beee0a80-34f5-4c5f-89fc-14f6898336de

What about this?

estib-vega (approved):
Approved in order to get it on the release :)

PavelLaptev:
> Approved in order to get it on the release :)

@estib-vega Thanks! I'll check your implementation later.

PR #4892, opened 2024-09-12
UI for stacked branches

mtsgrd:
Note that this is work in progress (e.g. some buttons don't even work), and hidden behind a feature toggle. It changes the way we render commits based on earlier work that enables branch creation for individual commits.
- [x] transform commit list to grouped list by reference
- [x] render pr card for groups that have a pr

### DIFF
--- a/apps/desktop/src/routes/settings/experimental/+page.svelte
+++ b/apps/desktop/src/routes/settings/experimental/+page.svelte
@@ -3,7 +3,7 @@
 	import {
 		featureBaseBranchSwitching,
 		featureInlineUnifiedDiffs,
-		featureBranchStacking,
+		stackingFeature,
 		featureTopics
 	} from '$lib/config/uiFeatureFlags';
 	import SettingsPage from '$lib/layout/SettingsPage.svelte';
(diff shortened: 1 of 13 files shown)


apps/desktop/src/routes/settings/experimental/+page.svelte
```diff
@@ -3,15 +3,14 @@
 	import {
 		featureBaseBranchSwitching,
 		featureInlineUnifiedDiffs,
-		featureBranchStacking,
+		stackingFeature,
```
ndom91, line 6: Just out of curiosity, why did you decide to go against the grain with this name change here? :sweat_smile:

mtsgrd: Thanks, not sure that rename was intentional, let me have another think.

The thing I _do_ want to change is that the feature toggles should not be functions that return stores, because each time the function is called it gets a new store. So I changed one of these feature toggles so that I can flip back and forth without having to reload the page.

ndom91:
Just a heads up, this badge needs some dark mode variant styling: 
image

ndom91 (approved):
LGTM for feature flagged stuff :+1:

PR #4896, opened 2024-09-12
fix: move markdown styling to Markdown.svelte component

ndom91:
## ☕️ Reasoning

- Colocate the markdown css with the only component rendering it (`Markdown.svelte`), instead of having it live in a different css file in the UI library
- @PavelLaptev unfortunately styles can't go in the `MarkdownContent.svelte` without the `:global()` because they don't seem to apply to the dynamic components there


PavelLaptev:
@ndom91 If this doesn't work without `:global`, then I don't think there's much difference. Let's close the PR for now, and I'll check later if there's another workaround.

PR #4897, opened 2024-09-12
Fix HunkDiff borders

estib-vega:
The borders move with the hunk diffs

## ☕️ Reasoning
The borders int the hunks don't always move with the hunk itself when e.g. scrolling it.

## 🧢 Changes
Update the style, so that the borders do move accordingly

### DIFF
--- a/apps/desktop/src/lib/hunk/HunkDiff.svelte
+++ b/apps/desktop/src/lib/hunk/HunkDiff.svelte
@@ -534,11 +534,11 @@
 		text-wrap: nowrap;
 		color: var(--clr-text-2);
 		display: flex;
-		justify-content: start;
+		align-items: center;
 		border-bottom: 1px solid var(--clr-border-2);
 		border-right: 1px solid var(--clr-border-2);
 		border-top: 1px solid var(--clr-border-2);
-		border-top-right-radius: var(--radius-s);
+		border-top-right-radius: var(--radius-m);
 	}
 
 	.table__numberColumn {
--- a/apps/desktop/src/lib/hunk/HunkDiff.svelte
+++ b/apps/desktop/src/lib/hunk/HunkDiff.svelte
@@ -14,6 +14,7 @@
 	import { type Hunk } from '$lib/vbranches/types';
 	import Checkbox from '@gitbutler/ui/Checkbox.svelte';
 	import Icon from '@gitbutler/ui/Icon.svelte';
+	import { pxToRem } from '@gitbutler/ui/utils/pxToRem';
 	import diff_match_patch from 'diff-match-patch';
 	import type { Writable } from 'svelte/store';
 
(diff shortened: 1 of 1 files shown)


apps/desktop/src/lib/hunk/HunkDiff.svelte
```diff
@@ -440,17 +457,15 @@
 	.table__checkbox-container {
 		z-index: var(--z-lifted);
 
-		border-right: 1px solid var(--clr-border-2);
-		border-bottom: 1px solid var(--clr-border-2);
+		box-shadow: inset 0 0 0 1px var(--clr-border-2);
```
PavelLaptev: is it possible not to use `box-shadow`?

apps/desktop/src/lib/hunk/HunkDiff.svelte
```diff
@@ -485,14 +500,36 @@
 			transform 0.2s;
 	}
 
+	.table__right-box {
+		pointer-events: none;
+		position: absolute;
```
PavelLaptev, line 514: I don't really like this approach with absolute positioning the whole header. I know you did it to prevent scrolling, but it's not a big deal.
Another issue with this approach is shifting when you zoom the app. It's because there are px untis instead of rem
image

estib-vega:
> and another issue when you zoom in and scroll a hunk horizontally image

That seems to be the case in the released app as well

PavelLaptev (changes requested):

apps/desktop/src/lib/hunk/HunkDiff.svelte
estib-vega: > I don't really like this approach

That's fair, and if you feel strongly against it we can dismiss this approach completely.

I like it because it can look like the designs, even-though the price is this complicated structure of dynamically positioned elements.

> Another issue with this approach is shifting when you zoom the app. It's because there are px untis instead of rem
>

You're right! Using pxToRem seems to have solved it for me, though

https://github.com/user-attachments/assets/eb35fc2b-d268-4816-bc12-d980fc8c37bb

apps/desktop/src/lib/hunk/HunkDiff.svelte
estib-vega: Seems we can, yes 👍

PavelLaptev:
@estib-vega thanks for the fixes. Looks good, hopefully we covered all corner cases 🤞
 one small thing is left. When I press "Start commit" I can also see checkboxes on commits where I shouldn't see it. Can you please check this before merging?

image

PavelLaptev (approved):

estib-vega:
> When I press "Start commit" I can also see checkboxes on commits where I shouldn't see it. Can you please check this before merging?
> 

I see that, nice catch!
Seems to be unrelated to this PR so I opened another one with the fix: https://github.com/gitbutlerapp/gitbutler/pull/4916

PR #4899, opened 2024-09-12
Base branch: Safely handle divergence

estib-vega:
## ☕️ Reasoning
The application doesn't display any kind of information related to base branch divergence.
It should detect this, raise this to the user and present alternative actions to address this.

### Scenario
1. I open a repository on GitButler.
2. A team-mate of mine makes some changes, and mistakenly pushes a commit with sensitive information.
3. I merge the changes into the local base branch of GitButler. I can see the mistaken commit.
4. They decide to force-push the commit away.
5. I reload the application, but the state is the same as in step 3.

Expected behavior would be to have some information about this and the ability to restore the application to the latest version of the target upstream.

## 🧢 Changes
1. Detect whether the base-branch has diverged with its upstream.
2. Display the information to the user in a way that makes it clear if and which changes will be lost, if any action is taken.
3. Allow the user to decide how to resolve said conflict through the **Integrate Upstream Changes** modal.
4. Detect whether the base-branch has conflicts (e.g. after rebasing your local base-branch changes into the upstream) and display that.

### Allow the user to use edit mode on a base-branch conflict
Since we're allowing the user to rebase or merge changes into the base-branch, we can end up with conflicted commits there.
We already display and signal it, but we need to enable edit mode on base-branch conflicts.

### DIFF
--- a/apps/desktop/src/lib/baseBranch/baseBranchService.ts
+++ b/apps/desktop/src/lib/baseBranch/baseBranchService.ts
@@ -67,6 +67,27 @@ export class BaseBranchService {
 		});
 		await this.fetchFromRemotes();
 	}
+
+	async push(withForce?: boolean) {
+		this.loading.set(true);
+		try {
+			await invoke<void>('push_base_branch', {
+				projectId: this.projectId,
+				withForce
+			});
+		} catch (err: any) {
+			if (err.code === Code.DefaultTargetNotFound) {
+				// Swallow this error since user should be taken to project setup page
+				return;
+			} else if (err.code === Code.ProjectsGitAuth) {
+				showError('Failed to authenticate', err);
+			} else {
+				showError('Failed to push', err);
+			}
+			console.error(err);
+		}
+		await this.fetchFromRemotes();
+	}
 }
 
 export async function getRemoteBranches(
--- a/apps/desktop/src/lib/components/BaseBranch.svelte
+++ b/apps/desktop/src/lib/components/BaseBranch.svelte
@@ -66,6 +66,7 @@
 	);
 
 	let baseBranchIsUpdating = $state<boolean>(false);
+	const baseBranchConflicted = $derived(base.conflicted);
 	let updateTargetModal = $state<Modal>();
 	let resetBaseStrategy = $state<ResetBaseStrategy | undefined>(undefined);
 	let confirmResetModal = $state<Modal>();
@@ -73,6 +74,12 @@
 	let integrateUpstreamModal = $state<IntegrateUpstreamModal>();
 	const integrateUpstreamModalOpen = $derived(!!integrateUpstreamModal?.imports.open);
 	let mergeUpstreamWarningDismissedCheckbox = $state<boolean>(false);
+
+	const pushButtonTooltip = $derived.by(() => {
+		if (onlyLocalAhead) return 'Push your local changes to upstream';
+		if (base.conflicted) return 'Cannot push while there are conflicts';
+		return resetBaseTo.local.tooltip;
+	});
 
 	const multiple = $derived(
 		base ? base.upstreamCommits.length > 1 || base.upstreamCommits.length === 0 : false
@@ -296,13 +303,12 @@
 						style={resetBaseTo.local.color}
 						icon={onlyLocalAhead ? undefined : 'warning'}
 						kind="solid"
-						tooltip={onlyLocalAhead
-							? 'Push your local changes to upstream'
-							: resetBaseTo.local.tooltip}
+						tooltip={pushButtonTooltip}
 						loading={baseBranchIsUpdating || confirmResetModalOpen}
 						disabled={$mode?.type !== 'OpenWorkspace' ||
 							baseBranchIsUpdating ||
-							confirmResetModalOpen}
+							confirmResetModalOpen ||
+							baseBranchConflicted}
 						onclick={resetBaseTo.local.handler}
 					>
 						{onlyLocalAhead ? 'Push' : resetBaseTo.local.title}
--- a/apps/desktop/src/lib/components/BaseBranch.svelte
+++ b/apps/desktop/src/lib/components/BaseBranch.svelte
@@ -21,34 +21,37 @@
 	import { tick } from 'svelte';
 	import type { BaseBranch } from '$lib/baseBranch/baseBranch';
 
-	const resetBranchTo = {
-		remote: {
-			title: 'Reset to remote',
-			header:
-				'You are about to reset the target branch to the remote branch. This will lose all the changes ahead of the remote branch.',
-			tooltip:
-				'Resets the target branch to the upstream branch. Will lose all the changes ahead of the upstream branch.',
-			color: 'warning',
-			action: updateBaseBranch
-		},
-		local: {
-			title: 'Reset to local',
-			header:
-				'You are about to reset the target branch to the local branch. This will lose all the remote changes not in the local branch.',
-			tooltip:
-				'Resets the upstream branch to the local target branch. Will lose all the remote changes not in the local branch.',
-			color: 'pop',
-			action: async () => await pushBaseBranch()
-		}
-	} as const;
-
-	type ResetActionType = keyof typeof resetBranchTo;
-
 	interface Props {
 		base: BaseBranch;
 	}
 
 	const { base }: Props = $props();
+
+	const resetBaseTo = {
+		local: {
+			title: 'Push local changes',
+			header:
+				'You are about to reset the upstream target branch to the local branch. This will lose all the remote changes not in the local branch.',
+			content: 'You will force-push your local changes to the remote branch.',
+			tooltip:
+				'Resets the upstream branch to the local target branch. Will lose all the remote changes not in the local branch.',
+			color: 'pop',
+			handler: handlePushLocalChanges,
+			action: pushBaseBranch
+		},
+		remote: {
+			title: 'Discard local changes',
+			header:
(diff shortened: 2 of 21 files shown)


krlvi:
Lets separate out the update of base branch (because we can in the future show exactly what is gonna happen to branches)

crates/gitbutler-branch-actions/src/base.rs
```diff
@@ -604,10 +647,45 @@ pub(crate) fn target_to_base_branch(ctx: &CommandContext, target: &Target) -> Re
...
+    let target_commit = ctx
+        .repository()
+        .find_commit(
+            base.diverged_fork_point
+                .context("failed to find fork point")?,
+        )
+        .context("failed to find target commit")?;
+
+    set_base_branch_to_commit(ctx, target_commit, perm)
+}
```
Caleb-T-Owens: What is the difference between update_base_branch and realign_base_branch? Can they be made the same function?

apps/desktop/src/lib/baseBranch/baseBranchService.ts
```diff
@@ -67,6 +68,41 @@ export class BaseBranchService {
...
+			const stashedConflicting = await invoke<string[]>('realign_base_branch', {
+				projectId: this.projectId
+			});
+			await this.fetchFromRemotes();
+			return getStashedConflictingMessage(stashedConflicting);
+		} catch (err: any) {
+			showError('Failed to realign base branch', err);
+			console.error(err);
+		}
+	}
```
Caleb-T-Owens: Do we even call this from the frontend? Doing a quick text search I'm not seeing it.

crates/gitbutler-branch-actions/src/base.rs
estib-vega: `update_base_branch`: 
This resets the local base branch to whatever the tip of its upstream is.

`realign_base_branch`:
This resets the local base branch to the fork-point it has to its upstream.
This drops any ahead changes in the local base branch.
So it's "aligned" in the sense that it no longer diverts from upstream.

apps/desktop/src/lib/components/BaseBranch.svelte
```diff
@@ -1,57 +1,201 @@
...
+		if (infoText) {
+			showInfo('Stashed conflicting branches', infoText);
+		}
+		await tick();
+		baseBranchIsUpdating = false;
+	}
+
+	async function realignBaseBranch() {
+		baseBranchIsUpdating = true;
+		const infoText = await baseBranchService.realignBaseBranch();
```
estib-vega: @Caleb-T-Owens here :)

apps/desktop/src/lib/baseBranch/baseBranchService.ts
estib-vega: I think it's a large diff that wasn't rendered. At least for me

estib-vega:
FYI: This PR needs some rework around the way that we want the resolution to work.
We decided to use the new update with rebase/merge UI in order to deal with this kind of situations

crates/gitbutler-branch-actions/src/base.rs
```diff
@@ -550,6 +536,34 @@ pub(crate) fn update_base_branch(
...
+    perm: &mut WorktreeWritePermission,
+) -> anyhow::Result<Vec<ReferenceName>> {
+    ctx.assure_resolved()?;
+
+    // look up the target and see if there is a new oid
+    let target = default_target(&ctx.project().gb_dir())?;
+    let repo = ctx.repository();
+    let target_branch = repo
+        .find_branch_by_refname(&target.branch.clone().into())
+        .context(format!("failed to find branch {}", target.branch))?;
```
Byron: Just as a not: Whenever something is non-static thanks to formatting, using `context()` will always pay for the `format!()` even though it's not used.
Ideally, get used to delay the work for when it's actually needed using `with_context(|| format!(…))`.
Of course, it's fair to optimise for readability which is when one would rather do it like this, and performance doesn't matter outside of hot-loops and when it's measured.

estib-vega: I see, thanks for the advice!
Out of curiosity: Is there any downside to only using `with_context`, aside from the small readability penalty?

Byron: I don't think so, even though admittedly I'd use it if the string is static.
And again, I'd keep it this way, i.e. with `context()`, as I'd optimize for readability most of the time. When I write this I will always write `.with_context(|| …)` probably simply because… I just feel I have to write more efficient code if the trade-off seems acceptable. But that's a me *problem* I think 😅.

estib-vega: Fair enough :)

Yeah, to me it makes sense to get as many performance gains if the price is that small, really.
Will keep it in mind!

Byron: I think this is my point - there will be no gain by doing this (`with_context()`), nobody will ever notice. That's why what's done here, the simple thing, is probably best. It's just that I can't do it because of being a bit inflexible.

estib-vega: 😂

crates/gitbutler-branch-actions/src/base.rs
```diff
@@ -558,6 +564,34 @@ pub(crate) fn target_to_base_branch(ctx: &CommandContext, target: &Target) -> Re
...
+    ))?;
+
+    let diverged = fork_point != target.sha;
+
+    let diverged_ahead = if diverged {
+        repo.log(target.sha, LogUntil::Commit(fork_point))
+            .context("failed to get diverged ahead commits")?
+            .iter()
+            .map(|commit| commit.id())
+            .collect::<Vec<git2::Oid>>()
```
Byron:
```suggestion
            .collect::<Vec<_>>()
```

Maybe it even works without the `::<>` part, but probably not.

crates/gitbutler-branch-actions/src/base.rs
```diff
@@ -558,6 +564,34 @@ pub(crate) fn target_to_base_branch(ctx: &CommandContext, target: &Target) -> Re
...
+    } else {
+        Vec::new()
+    };
+
+    let diverged_behind = if diverged {
+        repo.log(oid, LogUntil::Commit(fork_point))
+            .context("failed to get diverged behind commits")?
+            .iter()
+            .map(|commit| commit.id())
+            .collect::<Vec<git2::Oid>>()
```
Byron:
```suggestion
            .collect::<Vec<_>>()
```

crates/gitbutler-branch-actions/src/base.rs
```diff
@@ -558,6 +564,34 @@ pub(crate) fn target_to_base_branch(ctx: &CommandContext, target: &Target) -> Re
     let commit = branch.get().peel_to_commit()?;
     let oid = commit.id();
 
+    // determined if the base branch is behind it's upstream
+    let fork_point = repo.merge_base(target.sha, oid).context(format!(
```
Byron: Maybe this method would be faster: https://docs.rs/git2/latest/git2/struct.Repository.html#method.graph_ahead_behind ?

Byron (commented):
Thanks for tackling this!

I only took a quick look so can't tell if the logic is correct everywhere, also there seems to be no test. Setting it up is probably quite cumbersome right now, but I think there are examples around the existing test-suite that shows how to get into the desired state. Getting really good at testing with whatever test-suite is available will be key to make this work over time - and I say this as someone who often skips tests here 😅.

Besides that, I could only find a possible performance improvement, along with a nit. Please note that I only ever look at the Rust code.

estib-vega:
> Thanks for tackling this!
> 
> I only took a quick look so can't tell if the logic is correct everywhere, also there seems to be no test. Setting it up is probably quite cumbersome right now, but I think there are examples around the existing test-suite that shows how to get into the desired state. Getting really good at testing with whatever test-suite is available will be key to make this work over time - and I say this as someone who often skips tests here 😅.
> 
> Besides that, I could only find a possible performance improvement, along with a nit. Please note that I only ever look at the Rust code.

@Byron It's a good point, I'll add some tests 👌

crates/gitbutler-branch-actions/src/upstream_integration.rs
```diff
@@ -595,6 +666,64 @@ mod test {
...
+    #[test]
+    fn test_conflicted_head_branch_resolve_divergence_hard_reset() {
+        let test_repository = TestingRepository::open();
+        let initial_commit = test_repository.commit_tree(None, &[("foo.txt", "bar")]);
+        let old_target = test_repository.commit_tree(Some(&initial_commit), &[("foo.txt", "baz")]);
+        let branch_head = test_repository.commit_tree(Some(&old_target), &[("foo.txt", "fux")]);
+        // new target diverged from old target
+        let new_target = test_repository.commit_tree(Some(&initial_commit), &[("foo.txt", "qux")]);
+
+        let branch = make_branch(branch_head.id(), branch_head.tree_id());
```
estib-vega: @Byron 👋
From here down I've added some tests for what I consider it's the main parts that should be tested:
1. That the **base-branch divergence resolution approach** correctly updates the state of the **branches' upstream integration statuses**
2. That the **computed resolutions** yield the expected results depending on the chosen **base-branch divergence resolution approach**

estib-vega (commented):
@Byron @Caleb-T-Owens let me know what you think or if there are some other meaningful things you think I should add tests for :)

Caleb-T-Owens (approved):
Thank you very much!

PR #4903, opened 2024-09-13
Base avatar bg color on srcUrl instead of tooltip

mtsgrd:
- consistent per author
- shuffle colors on page refresh

### DIFF
--- a/packages/ui/src/lib/utils/stringToColor.ts
+++ b/packages/ui/src/lib/utils/stringToColor.ts
@@ -11,11 +11,6 @@ const colors = [
 	'#5FD2B0'
 ];
 
-// We use a random seed to avoid colors being static with respect to a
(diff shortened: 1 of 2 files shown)


packages/ui/src/lib/utils/stringToColor.ts
```diff
@@ -11,12 +11,17 @@ const colors = [
 	'#5FD2B0'
 ];
 
+// We use a random seed to avoid colors being static with respect to a
```
PavelLaptev: I like how you fixed the randomness in `Avatar.svelte` by replacing `tooltip` with `src`. The idea of having consistent backgrounds for each avatar. The avatars should remain the same on each refresh, so I think we can drop this change

PavelLaptev (changes requested):

PavelLaptev (approved):

PR #4906, opened 2024-09-13
feat: use new Tabs components to organise Preferences page

ndom91:
## ☕️ Reasoning

- Preferences page got too long

## 🧢 Changes

- Add `Tabs`/`TabContent`/`TabTrigger`/`TabList` components
- Use tabs in user preferences page to organise and group the various sections
	- Commit Signing section and experimental feature flag toggles were in 1 component together (`PreferencesForm.svelte`), so this is split out into `CommitSigningForm` / `PreferencesForm` here.
- Move all form components to `../settings/userPreferences/*`
- Add built-in "SegmentControl" tabs style, while still allowing custom tabs style for other use-cases

### 🚧 Todo

- Move `Tab*.svelte` components to `packages/ui/` UI library
	- Add storybook stories
- Double check in which tab each section should go
- Design

## 📺 Current status:

See comments below


ndom91:
@PavelLaptev okay so here's a very basic prototype. All the functionality is there.

- I added a set of Tab components; Don't worry, I plan on moving these to the UI lib and adding stories haha
- There is only the most basic of design implemented here though. So yeah if you want to give me a few notes or a quick Figma mockup when you have some time I can take it from there :+1: 

No rush on this though ofc!

PavelLaptev:
@ndom91 I'll make my changes in your branch if you don't mind. And I'll drop the new "tabs" component, I'll replace it with segmented control

ndom91:
> @ndom91 I'll make my changes in your branch if you don't mind. And I'll drop the new "tabs" component, I'll replace it with segmented control.

Haha okay yeah no problem :+1:

PavelLaptev:
@ndom91 could you please replace the tabs component with the segmented controller. I thought it will be easy to replace, but I see you have also `TabContent` and `TabList`, it's better if you do it :-) 

image

Here is a new design — https://www.figma.com/design/FbeLt0yjY9RiNn8MXUXsYs/Client-design?node-id=1513-57296&t=6UIBc43Py1LAErBo-1

PavelLaptev (changes requested):
@ndom91 could you please replace the tabs component with the segmented controller. I thought it will be easy to replace, but I see you have also `TabContent` and `TabList`, it's better if you do it :-) 

image

Here is a new design — https://www.figma.com/design/FbeLt0yjY9RiNn8MXUXsYs/Client-design?node-id=1513-57296&t=6UIBc43Py1LAErBo-1

ndom91:
@PavelLaptev so I can definitely change it, but don't you think we'll have a need for a normal tab component eventually anyway? I guess at the latest in v3 fat tabs on the main workspace page.

ndom91:
> Yes, on V3 we will have tabs, but design will be different. You might want make it as a wrapper where we can pass any children.
> 
> And one more thing, we don't use page transitions anywhere in the app, and I don't this we should start use it here. Could you please make content switch without transitions.

Okay yeah I'll remove the transitions, we do use them a good bti though, no? I copied them from elsewhere in the app.

Regarding the tab's themselves, they already behave that way here where they'll render any children so you can totally customize them :+1:

ndom91:
@PavelLaptev I'm tryign to use the `SegmentControl` here, but tbh I think it really isn't the right choice. The Segment control to me is simply a form input component where you can select one of the `Segment`'s and that's the selected option, right. Sort of like a radio select

There's no support for providing a content area which syncs with the selected `Segment`/`Tab`. 

The current tab implementation that I have here has a simple API, and it supports completely custom Tab's (`TabTrigger`) and tab content (`TabContent`). 

What do you think about if I continue to use these Tabs under the hood, but make it look closer to your design?

ndom91:
Current state of the tabs style:

image

PavelLaptev:
Thank you @ndom91, feel free to check to merge when it's ready.
I moved "Segment" styles into a separate file in order to prevent code duplication and fix a few layout issues.

PavelLaptev (approved):

PR #4909, opened 2024-09-13
Improve applying branches

Caleb-T-Owens:
Applying branches now uses the never-conflicting logic if fearless rebasing is enabled.


krlvi (approved):
there are 6 failing tests, but other than that - LFG

PR #4915, opened 2024-09-16
feat: update nix config

gkze:
## ☕️ Reasoning
The Tauri-recommended Flake uses flake-utils, which has outgrown its original purpose a bit and handles outputs less cleanly than [flake-parts](https://flake.parts/). I realize that there's the nascent [Flakelight](https://github.com/nix-community/flakelight) out there but for now my sense is that flake-parts the optimal option.

Also add [numtide/devshell](https://github.com/numtide/devshell), which handles native C/C++ library dependencies a bit more elegantly.

## 🧢 Changes
* Swap out flake-utils for flake parts
* Add [numtide/devshell](https://github.com/numtide/devshell)
* Add `.envrc` to help with Nix-related changes a bit

Was able to build (not AppImage but doesn't build on master anyways - separate issue) and run the dev build on my primary daily driver NixOS system.

Would be interesting to see why it's not producing the AppImage on master on NixOS, just haven't dedicated enough time.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,7 @@
+# vi: ft=sh
+# shellcheck disable=all
+dotenv_if_exists
+use flake
+watch_file $(fd -tf -enix)
+
+export PATH="${PWD}/node_modules/.bin:${PATH}"
--- a/flake.lock
+++ b/flake.lock
@@ -1,78 +1,96 @@
 {
   "nodes": {
-    "flake-utils": {
+    "devshell": {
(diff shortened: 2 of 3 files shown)


ndom91:
Also, the font rendering seems a bit off, see this comparison screenshot. Before is in the background and after (with your `devShell`) is in the foreground outlined in red:

image

Seems that we really might need to include pango/cairo as runtime dependencies?

PR #4919, opened 2024-09-16
fix: rm `effect.pre` used only to avoid previous bug

ndom91:
- Remove `$effect.pre` usage in `+layout.svelte` where we were using it to avoid a previous bug


estib-vega (approved):
Looks good to me 👍

PR #4921, opened 2024-09-16
feat: create 'Apperance' settings page and add highlight branchlane setting

ndom91:
Wish I had stacking for this one :joy:

## ☕️ Reasoning

- Based on some user feedback, highlight the full branch lane name contents when clicking into it.

image

## 🧢 Changes

- Talked about it briefly in the office, and we were split on if this is how one would expect this field to work or not, so I added a feautre flag toggle.
- Added new "Appearance" settings page

### After

image

image


ndom91:
@PavelLaptev what do you think of making a `<Section />` here to organise the UI-related settings instead of having them interleaved in the "Profile" page without any distinction?

### Before

image


### After

image

apps/desktop/src/lib/config/uiFeatureFlags.ts
```diff
@@ -22,3 +22,8 @@ export function featureTopics(): Persisted<boolean> {
 	const key = 'feature--topics';
 	return persisted(false, key);
 }
+
+export function highlightBranchLaneContentsFeature(): Persisted<boolean> {
```
mtsgrd: Can we make this new addition an exported constant instead of a function? Also, I'm wondering if we can improve the name. It feels more like auto selecting input field contents rather than highlighting branch lane contents.

ndom91: yup, no problem. What do you think of `autoSelectBranchLaneContentsFeature`? It's a bit long, but gets the whole idea across imo

mtsgrd: Yeah that totally works for me, and I'll still revert the name change of the stacking flag so it follows that convention. However, the pedantic part of me is still objecting to the "branch lane contents" part. Imho it's neither the right subject (branch contents) nor is highlighting what we're doing. It's about _text selection_ on input field values isn't it?

ndom91: It is, it's about text selection on focusing an input field. But its very specific to this BranchLane input only, right. It's not a generic "do this on all inputs"

mtsgrd: Ah right - sorry, I misread something and thought it would extend to more than one field.

estib-vega (approved):
Looks good to me 👍

Issue #4847, opened 2024-09-07
Setup: Initial remote should match "Current target branch"


koppor:
### Version

0.12.26

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

I have a repository with multiple upstreams:

```
$ git remote -v
origin  git@github.com:koppor/jabref.git (fetch)
origin  git@github.com:koppor/jabref.git (push)
upstream        git@github.com:JabRef/jabref.git (fetch)
upstream        git@github.com:JabRef/jabref.git (push)
```

I click "Add local repository" in the UI

![image](https://github.com/user-attachments/assets/75a388b6-5361-4b71-ba1d-1f22e9c73af8)

I select the directory of my project.

Then, GitButler select `upstream/main` as "Current target branch" and `origin` as "remote".

![image](https://github.com/user-attachments/assets/d963368f-0b5c-4b6e-bac3-f83941bcfedc)

I think, the remote when starting the window should be the remote matching the "Current target branch".

Byron:
Thanks a lot for reporting!

> You have branches from multiple remotes. If you want to specify a remote for creating branches
that is different from the remote that your target branch is on, change it here.

And I agree, I think the text indicates that by default, it should match the target branch, so it can be changed if needed.

Maybe also interesting for @estib-vega .

PR #4923, opened 2024-09-16
Project target setup: Use default remote

estib-vega:
### Issue
As seen in https://github.com/gitbutlerapp/gitbutler/issues/4847

### Changes
The project target setup will select a remote default that matches the selected branch. The remote default can be overridden if needed.


ndom91 (approved):
LGTM :+1:

PR #4926, opened 2024-09-16
feat: Add a new dad joke to dad-jokes.txt.

krlvi:


mtsgrd (changes requested):
can I suggest `s/how/what`?

Issue #4904, opened 2024-09-13
Error "head invalid" when creating PR


ukarlsson:
### Version

0.12.25

### Operating System

Mac OS X

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

Error when creating PR. It works when I click "Create PR" again, but often I get this error the first time

GitHub validation failed
It seems there was a problem validating the request.

Please see our 
[documentation](https://docs.gitbutler.com/)
for additional help.
head invalid

### How to reproduce

Click "Create PR"

### Expected behavior

![Screenshot 2024-09-13 at 11 38 20](https://github.com/user-attachments/assets/9f29fe8d-7f9a-4119-8e49-079058f03fe3)
Screenshot of error

Byron:
Thanks a lot for reporting!

This is an issue where I have no idea how to approach it, at least assuming that a foundational feature like creating a PR isn't broken for everyone. And that I could validate, as I was able to create a PR without a problem in a private repository.

Can you provide more information that might help to reproduce the issue?

ukarlsson:
Ok many thanks for the response. Any logs I can supply?

It is pretty consistent now that first time I press create PR I get this error, and then the second time it works

Byron:
I think for this interaction, the developer console might have additional logs. You can open it with `Cmd + Opt + i`.
The `head invalid` message is probably created by the API, which means the first time around, the `head` is invalid, the second time it's working. Like a race condition, or something processed out of order.

ukarlsson:
Ok, got info from network console now

Error is

```
{
    "message": "Validation Failed",
    "errors": [
        {
            "resource": "PullRequest",
            "field": "head",
            "code": "invalid"
        }
    ],
    "documentation_url": "https://docs.github.com/rest/pulls/pulls#create-a-pull-request",
    "status": "422"
}
```

Summary
```
URL: https://api.github.com/repos/sanalabs/sierra-platform/pulls
Status: 422
Source: Network
Address: 140.82.121.6:443
Initiator: 
fetch.js:54
```

Request data
```
{"head":"refactor-namespace-processing","base":"master","title":"Update namespace handling in analytics cache refresh","body":"","draft":false}
```

ukarlsson:
When I press create PR again the request body is slightly different, and head is indeed different

```
{"head":"uk/refactor-namespace-processing","base":"master","title":"Update namespace handling in analytics cache refresh","body":"","draft":false}
```

[...]

PR #4927, opened 2024-09-16
fix: Extract the name from the remote ref correctly

estib-vega:
### Issue
From issue https://github.com/gitbutlerapp/gitbutler/issues/4904

Correctly extract the name of the remote branch, supporting branch names with nested paths


krlvi (approved):
lgtm

PR #4932, opened 2024-09-17
fix: Svelte navigation on dev

estib-vega:
`svelte-routing` needs to be excluded from the optimized dependencies for it to work properly.

https://github.com/vitejs/vite/issues/2760#issuecomment-809552053


estib-vega:
Nevermind, it stopped working

PR #4937, opened 2024-09-17
feat: bump tauri to 1.8.0

ndom91:
## ☕️ Reasoning

- The tauri folks have merged a few PR's prepping the updater for moving from tauri v1 to tauri v2, see: https://github.com/tauri-apps/tauri/pull/10938#issuecomment-2351551535
- They explicitly mentioned in their Discord, however, that we should definitely upgrade to `1.8.0` first and then to `2.0+`. 

## 🧢 Changes

- Cargo: Bump `tauri@1.8.0`
- Node: Bump `@tauri-apps/cli@1.6.2`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7345,9 +7345,9 @@ checksum = "61c41af27dd6d1e27b1b16b489db798443478cef1f06a660c96db617ba5de3b1"
 
 [[package]]
 name = "tauri"
-version = "1.7.2"
+version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


amrbashir (commented):
Hey, apologies but I forgot to mention that you should also update the tauri CLI, it is a must if you are using NSIS installer, but even if not, you probably should still update it for good measure

ndom91:
@amrbashir great, thanks for reaching out!

krlvi (approved):
lgtm

krlvi:
merging cause i would like to test this on the nightly now

PR #4938, opened 2024-09-17
fix: File section max line number & simplify hunk header calculation

estib-vega:
- Fix a bug in which the max lines would not be taken into account when checking what the biggest line number in a hunk section.
- Simplify the calculation of the header width for the HunkDiff component


PavelLaptev (approved):

krlvi:
force pushed a lint fix

PR #4941, opened 2024-09-18
fix: Calculate the right file tree when resolving

estib-vega:
Get the correct file tree when resolving a conflict

### DIFF
--- a/crates/gitbutler-edit-mode/src/lib.rs
+++ b/crates/gitbutler-edit-mode/src/lib.rs
@@ -267,7 +267,11 @@ pub(crate) fn save_and_return_to_workspace(
     let parents = commit.parents().collect::<Vec<_>>();
 
     // Recommit commit
-    let tree = repository.create_wd_tree()?;
+    let mut index = repository.index()?;
+    index.add_all(["*"], git2::IndexAddOption::DEFAULT, None)?;
+    let tree = index.write_tree()?;
+    let tree = repository.find_tree(tree)?;
+
     let commit_headers = commit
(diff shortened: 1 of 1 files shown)


Caleb-T-Owens (approved):

crates/gitbutler-edit-mode/src/lib.rs
```diff
@@ -267,7 +267,11 @@ pub(crate) fn save_and_return_to_workspace(
     let parents = commit.parents().collect::<Vec<_>>();
 
     // Recommit commit
-    let tree = repository.create_wd_tree()?;
+    let mut index = repository.index()?;
```
Byron, line 270: Oh, I think we have to look at `create_wd_tree()` then, and if this is a problem, then other usages of `create_wd_tree()` might also be.

The reason is that I changed it to bypass the index and instead work off a `git status` essentially. If this doesn't work or something is missing, then it would be good to either make work or revert the changes to `create_wd_tree()`.

Sorry for the hassle, I was convinced the implementation was equivalent.

CC @krlvi

krlvi: oh interesting... i remember struggling to create a workdir tree too for the oplog use case. it would be great if we could do that, without index add all

Byron: I agree. To resolve this particular issue, I think I'd want to understand what `create_wd_tree()` didn't do anymore so hopefully that can be fixed. Maybe it's related to conflicts, and add-all automatically resolves them?

PR #4943, opened 2024-09-18
fix: protect read file from path traversal

ndom91:
## ☕️ Reasoning

- Ensure that the new `get_pr_template_contents` fn can't read anything outside of the project's root directory


crates/gitbutler-tauri/src/projects.rs
```diff
@@ -106,9 +106,20 @@ pub mod commands {
...
+            .join(relative_path)
+            .as_path()
+            .canonicalize()
+            .context("invalid template path")?;
 
-        Ok(read_file_from_workspace(template_path.as_path())?)
+        // Ensure that the merged path is a child of the project's root path
+        // in order to avoid path traversal issues, for example.
+        if template_path.starts_with(&project.path) {
+            Ok(read_file_from_workspace(template_path.as_path())?)
```
mtsgrd: If I could make one suggestion, make the `read_file_from_workspace` function take a project path as well as a relative path? Right now that function is misnamed, it'll read a file from anywhere while the name suggests it would only read it from a workspace.

ndom91: Yeah good call, I moved the checking logic to `read_file_from_workspace` and had it take the base path and relative file path :+1:

ndom91:
@mtsgrd some small rust question:

- There's probably no need for `.context()` on each method that potentially fails, right? 
- Is it okay to add the entire `git2` crate to `gitbutler-fs`'s `Cargo.toml`? Can I add something like `feature = { Repository }` to only add what I need here?

crates/gitbutler-fs/Cargo.toml
```diff
@@ -12,3 +12,4 @@ anyhow = "1.0.86"
 gix = { workspace = true, features = ["dirwalk", "credentials", "parallel"] }
 walkdir = "2.5.0"
 toml.workspace = true
+git2 = { version = "0.18.3" }
```
krlvi: here you wanna do 
`git2.workspace = true`

Byron:
> * Is it okay to add the entire `git2` crate to `gitbutler-fs`'s `Cargo.toml`? Can I add something like `feature = { Repository }` to only add what I need here?

Features are predefined in the crate and they usually only provide features if it grates off other dependencies, not just code. When I see the example above, it reminds me of typescript imports :).

crates/gitbutler-fs/Cargo.toml
ndom91: After talking with Kiril, moved it to `gitbutler-repo` instead

crates/gitbutler-repo/src/commands.rs
```diff
@@ -47,4 +49,26 @@ impl RepoCommands for Project {
...
+            let tree = ctx.repository().head()?.peel_to_tree()?;
+            let entry = tree.get_path(relative_path)?;
+            let blob = ctx.repository().find_blob(entry.id())?;
+            let content = std::str::from_utf8(blob.content())?;
+
+            Ok(content.to_string())
+        } else {
+            anyhow::bail!("Invalid workspace file");
+        }
+    }
```
Caleb-T-Owens: Small nit, the `path()` function may return the repository path rather than the `.git` folder if the repository is bare. If we had a repository that was bare, that would mean that it could access files outside the repository.

```suggestion
    fn read_file_from_workspace(&self, relative_path: &Path) -> Result<String> {
        let ctx = CommandContext::open(self)?;
        let repository = ctx.repository();

        // If a repostiory is bare, repository.path will return the repository
        // directory rather than the `.git` repository.
        // Ref: https://docs.rs/git2/latest/git2/struct.Repository.html#method.path
        let base_path = if repository.is_bare() {
            repository.path()
        } else {
            repository
                .path()
                .parent()
                .ok_or(anyhow::anyhow!("Could not find repository base path"))?
        };

        let canonicalized_file_path = base_path.join(relative_path).canonicalize()?;

        if canonicalized_file_path.as_path().starts_with(base_path) {
            let tree = repository.head()?.peel_to_tree()?;
            let entry = tree.get_path(relative_path)?;
            let blob = repository.find_blob(entry.id())?;
            let content = std::str::from_utf8(blob.content())?;

            Ok(content.to_string())
        } else {
            anyhow::bail!("Invalid workspace file");
        }
    }
```


I wonder if we should name the function something like `read_commited_file` or something like that, to make it clear that its not going to look at uncommitted changes.

Caleb-T-Owens: https://docs.rs/git2/latest/git2/struct.Repository.html#method.path

Caleb-T-Owens: Actually, I've overcomplicating it, we have the project, so we should be able to just use `self.path` with no faffing about 😆

Byron: That's a great catch, @Caleb-T-Owens.
In general, there should be no assumption that `.git` is even in the worktree, independently of whether the repository is bare or not.
The only 'safe' way is to ask for the worktree-dir, and often the GitButler native types (like `Project` have a method for that.

krlvi: probably also worth noting that the function is trying to parse the file contents as utf8 - which of course wont work for files. I like the new blob reading implementation though, we can merge it and iterate from master

krlvi (approved):

crates/gitbutler-repo/src/commands.rs
ndom91: Used `self.path` instead and double checked if the file is binary before tryign to read it as utf-8. 

I'll merge this then and like you said we can iterate from there :+1:

PR #4945, opened 2024-09-18
Fix wrong files displaying in edit mode

Caleb-T-Owens:


ndom91 (approved):
Looks like it works as I would expect now! :pray: :+1:

PR #4947, opened 2024-09-18
The edit branch should not be listed in the side bar

estib-vega:
The `gitbutler/edit` branch should not be listed in the side bar of the application


Caleb-T-Owens (approved):
LGTM

PR #4948, opened 2024-09-18
Display the author and commit title in the edit mode

estib-vega:
Display the title of the commit and the author name in the commit instead of the placeholders

Screenshot 2024-09-18 at 15 59 14

### DIFF
--- a/apps/desktop/src/lib/commits/service.ts
+++ b/apps/desktop/src/lib/commits/service.ts
@@ -1,0 +1,20 @@
+import { invoke } from '$lib/backend/ipc';
+import { Commit } from '$lib/vbranches/types';
+import { plainToInstance } from 'class-transformer';
+
+export class CommitService {
+	constructor(private projectId: string) {}
+
+	async find(commitOid: string): Promise<Commit> {
+		const maybeCommit = await this.findMaybe(commitOid);
+		if (maybeCommit === undefined) {
+			throw new Error('Commit not found');
+		}
+		return maybeCommit;
+	}
+
+	async findMaybe(commitOid: string): Promise<Commit | undefined> {
+		const commit = await invoke<unknown>('find_commit', { projectId: this.projectId, commitOid });
+		return plainToInstance(Commit, commit);
+	}
+}
--- a/apps/desktop/src/lib/components/EditMode.svelte
+++ b/apps/desktop/src/lib/components/EditMode.svelte
@@ -1,14 +1,14 @@
 <script lang="ts">
-	// import { Project } from '$lib/backend/projects';
+	import { CommitService } from '$lib/commits/service';
 	import ActionView from '$lib/layout/ActionView.svelte';
 	import { ModeService, type EditModeMetadata } from '$lib/modes/service';
 	import { UncommitedFilesWatcher } from '$lib/uncommitedFiles/watcher';
 	import { getContext } from '$lib/utils/context';
+	import { Commit, type RemoteFile } from '$lib/vbranches/types';
 	import Button from '@gitbutler/ui/Button.svelte';
 	import InfoButton from '@gitbutler/ui/InfoButton.svelte';
 	import Avatar from '@gitbutler/ui/avatar/Avatar.svelte';
 	import FileListItem from '@gitbutler/ui/file/FileListItem.svelte';
-	import type { RemoteFile } from '$lib/vbranches/types';
 	import type { FileStatus } from '@gitbutler/ui/file/types';
 
 	interface Props {
@@ -17,7 +17,7 @@
 
 	const { editModeMetadata }: Props = $props();
 
-	// const project = getContext(Project);
+	const remoteCommitService = getContext(CommitService);
 	const uncommitedFileWatcher = getContext(UncommitedFilesWatcher);
 	const modeService = getContext(ModeService);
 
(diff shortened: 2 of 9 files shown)


Caleb-T-Owens:
Hey, I'm really not keen on this approach as it means we're effectively introducing another structure that is playing the role of a Commit.

I was hoping to have a service which can return a `struct RemoteCommit` for any given commit, so we can look up the commit details from the client side via the Oid

Caleb-T-Owens (changes requested):
comment above

crates/gitbutler-branch-actions/src/remote.rs
```diff
@@ -110,6 +110,15 @@ pub(crate) fn get_branch_data(ctx: &CommandContext, refname: &Refname) -> Result
...
 }
 
+pub(crate) fn get_commit_data(ctx: &CommandContext, sha: git2::Oid) -> Result<RemoteCommit> {
+    let commit = ctx
+        .repository()
+        .find_commit(sha)
+        .context("failed to find commit")?;
+
+    Ok(commit_to_remote_commit(&commit))
+}
```
Caleb-T-Owens:
```suggestion
pub(crate) fn get_commit_data(
    ctx: &CommandContext,
    sha: git2::Oid,
) -> Result<Option<RemoteCommit>> {
    let commit = match ctx.repository().find_commit(sha) {
        Ok(commit) => commit,
        Err(error) => {
            if error.code() == git2::ErrorCode::NotFound {
                return Ok(None);
            } else {
                bail!(error);
            }
        }
    };

    Ok(Some(commit_to_remote_commit(&commit)))
}
```

apps/desktop/src/lib/commits/service.ts
```diff
@@ -0,0 +1,15 @@
...
+	async find(projectId: string, commitOid: string): Promise<Commit | undefined> {
+		try {
+			const commit = await invoke<unknown>('find_commit', { projectId, commitOid });
+			return plainToInstance(Commit, commit);
+		} catch (error) {
+			console.error(error);
+			return undefined;
+		}
+	}
+}
```
Caleb-T-Owens: Something in my mind here is that we want to move away from using `console.error` and silently consuming errors. I figure here your intention was to represent a commit not being found, but I think with a little bit of jiggery pokery we can make it a little nicer.

```suggestion
import { invoke } from '$lib/backend/ipc';
import { Commit } from '$lib/vbranches/types';
import { plainToInstance } from 'class-transformer';
import type { Project } from '$lib/backend/projects';

export class CommitService {
	constructor(private project: Project) {}

	async find(commitOid: string): Promise<Commit> {
		const commit = await this.findMaybe(commitOid);

		if (!commit) {
			throw new Error(`Commit ${commitOid} not found`);
		}

		return commit;
	}

	async findMaybe(commitOid: string): Promise<Commit | undefined> {
		const commit = await invoke<unknown | undefined>('find_commit', { projectId: this.project.id, commitOid });

		if (!commit) return;

		return plainToInstance(Commit, commit);
	}
}
```

Caleb-T-Owens: We also know that the name "RemoteCommitSerice" is inaccurate, could we please call it "CommitService"

Caleb-T-Owens: I'd be keen to hear your thoughts on this, but in my mind it makes sense to try and pass as many dependencies in the constructor as possible. I would be keen to pass the project through the constructor until we had a good reason as to why that was a bad pattern, as it just adds more boilerplate to the consumers.

apps/desktop/src/lib/components/EditMode.svelte
```diff
@@ -27,13 +29,20 @@
...
 			initialFiles = files;
 		});
 	});
 
+	$effect(() => {
+		remoteCommitService.find(project.id, editModeMetadata.commitOid).then((maybeCommit) => {
+			commit = maybeCommit;
+		});
+	});
+
```
Caleb-T-Owens: Here we should always have a commit so I'd stick with `find` over `findMaybe`

Caleb-T-Owens (changes requested):
Thanks for iterating on this so quickly! Especially on a day when we're hosting a conference.

Very much in a good direction. I just think that we could represent error cases vs missing commits a little bit better

crates/gitbutler-branch-actions/src/remote.rs
Byron: Whenever you see this pattern, that an `err` is conditionally turned into an OK or is kept as the original (or altered) error, there is `.or_else(|err| …)` for you.

apps/desktop/src/lib/commits/service.ts
estib-vega: Will do the changes 👍 

> in my mind it makes sense to try and pass as many dependencies in the constructor as possible
>

This goes into a nice conversation I'd like to have about at some point going in the other direction :).

Short answer: Yes, I'll change it to make the CommitService project-scoped. I think that will follow the pattern that we have in other services more closely.

Long-and-to-be-had-in-person-over-a-coffee answer: I think as we keep adding functionalities we might want to walk away from OOP patterns (where we can). 
In this case, for example, this could have been a simple function, that would be easily testable and modular.
I think in the long run we'll benefit from making the FE simpler where we can.
But would be interested to hear what you think :)

Caleb-T-Owens: > Long-and-to-be-had-in-person-over-a-coffee answer: I think as we keep adding functionalities we might want to walk away from OOP patterns (where we can).
In this case, for example, this could have been a simple function, that would be easily testable and modular.
I think in the long run we'll benefit from making the FE simpler where we can.
But would be interested to hear what you think :)

I'm rather keen to have this conversation. As someone who spent three months of his life learning Haskell, I'm partial to a bit of functional programming.

estib-vega:
Thanks for the feedback @Caleb-T-Owens 
I made the agreed changes 👍

Caleb-T-Owens (approved):
LGTM

PR #4952, opened 2024-09-22
Git Changesets

Caleb-T-Owens:
In this PR I'm outlining functionality of git changesets. This is by no means in its final state. I just wanted to start pushing my changes onto GitHub so I don't accidentally use a vim binding to replace every character with 0s again 😆 .

### DIFF
--- a/crates/git-changeset/LICENSE.md
+++ b/crates/git-changeset/LICENSE.md
@@ -1,0 +1,10 @@
+This license only applies to the `crates/git-changeset` directory, and any files that are contained within. Anything outside of this directory falls under the license outlined in the root `LICENSE.md`, unless otherwise specified.
+
+
+Copyright 2023-2024 GitButler Inc
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the “Software”), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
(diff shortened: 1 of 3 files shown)


crates/git-changeset/LICENSE.md
Caleb-T-Owens, line 1: We probably need to have a standard way of clearly marking a directory under a different license. I'm unsure exactly what legal waffle may or may not be required to correctly state the scope of a particular license.

Caleb-T-Owens: CC @krlvi

H1ghBre4k3r: After some quick research, I found that it's apparently best practise to have a root license (which applies to the whole project) and extra licenses in sub folders of this project (if they should be licensed differently):

- https://softwareengineering.stackexchange.com/questions/264968/can-a-single-github-repository-include-source-code-released-under-multiple-licen
- https://opensource.stackexchange.com/questions/13172/having-one-sub-folder-in-a-github-repo-under-a-different-license-than-the-whole

So, if we go with adding an information about that in the main license file and add the respective "new" licenses in sub folders (if needed), we should be one the safe side, I think.

Caleb-T-Owens: Thanks @H1ghBre4k3r, I'll take a read through tomorrow and share with the team so we all have a shared understanding.

crates/git-changeset/README.md
```diff
@@ -0,0 +1,177 @@
...
+If any interesting parties has feedback or suggestions, I encourage you to reach out to me [caleb@gitbutler.com](mailto:caleb@gitbutler.com) or join our [Discord server](https://discord.com/invite/MmFkmaJ42D).
+
+## Requirements
+
+- A user should be able to find the latest revision of a commit for a given commit.
+- A user should be able to list the different revisions for a given changeset.
+- Changesets should be able to be worked on by two different users remotely, with ways of merging together two diverged versions of a particular changeset.
+- Changesets should have a unique identifier that tools can use to refer to them by.
+- Refer to an arbitrary commit by a changeset, without rewriting it.
+- Changesets need to be easily GCed.
```
schacon, line 22: What does this mean?

Caleb-T-Owens: We need to have some way of clearing up old and unnecessary change sets so we don't end up flooding the ODB

Caleb-T-Owens:
We've had an in person talk and we've decided to change it quite a bit

PR #4956, opened 2024-09-23
refactor: replace custom entries with Object.entries

estib-vega:
Remove custom entries utility and replace its usage with 
Object.entries across hotkeys and branch utility files.


Caleb-T-Owens (approved):
LGTM, thanks :+1:

PR #4959, opened 2024-09-23
Add basic views into stacks

schacon:
This adds a way to view the timeline, branches and stacks, create stacks and more. None of this is pretty, but we're just exercising the server APIs.

/cc @Caleb-T-Owens

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -23,6 +23,9 @@
 		"vite": "catalog:"
 	},
 	"dependencies": {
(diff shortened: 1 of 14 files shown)


Caleb-T-Owens (approved):
LGTM. Never seen `any`thing so beautiful

schacon:
OK, well, nevermind. The linter is mad about like 74 things. You'll have to use it as a reference I guess.

PR #4962, opened 2024-09-23
Fix state for file selections not propagating

### LINKED ISSUE #4934: Moving files between lanes does not work
Repro:
- Make change
- Drag change into seperate lane

It should be noted that dragging hunks does work. This indicates the the selection code is breaking down sepcifically for whole files.

Caleb-T-Owens:
Please could the reviewer pay extra attention to the changes made in the files `apps/desktop/src/lib/file/FileListItem.svelte` and `apps/desktop/src/lib/vbranches/fileIdSelection.ts`.

This pr closes the issue about dragging some hunks and then the remaining file, but there seems to still be some problems related to the file selection which I'd like to resolve tomorrow in a separate PR.
* Closes https://github.com/gitbutlerapp/gitbutler/issues/4934

I've noticed that there is a strange behavior where:
- If you select a file/s
- Drag the file/s to another lane
- Drag the file/s back

The files will retain their selected state. Ideally this state should be cleared after a drag as its holding a selected state for files that no longer exist in that lane. This potentially could cause a problematic state, if you dragged two files to the other branch, and then only dragged on back.

### DIFF
--- a/apps/desktop/src/lib/branch/BranchLane.svelte
+++ b/apps/desktop/src/lib/branch/BranchLane.svelte
@@ -84,7 +84,9 @@
 
 	const project = getContext(Project);
 	const fileIdSelection = new FileIdSelection(project.id, branchFiles);
-	const selectedFile = $derived(fileIdSelection.selectedFile);
+	const selectedFile = fileIdSelection.selectedFile;
+	const commitId = $derived($selectedFile?.[0]);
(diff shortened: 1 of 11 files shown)


krlvi (commented):
lgtm

krlvi (approved):

krlvi:
merging so we can test

PR #4963, opened 2024-09-24
IntegrateUpstreamModal: Factor out the modal

estib-vega:
Factor out the modal for upstream integration into a dedicated component so that it can be shared easily

### Other changes
- Expose the state of whether the modal is being showed or not: 
Some more info about this pattern. https://github.com/sveltejs/svelte/issues/11974#issuecomment-2157837477

### DIFF
--- a/apps/desktop/src/lib/components/IntegrateUpstreamModal.svelte
+++ b/apps/desktop/src/lib/components/IntegrateUpstreamModal.svelte
@@ -85,6 +85,12 @@
 		await tick();
 		modal?.show();
 	}
+
+	export const imports = {
+		get open() {
+			return modal?.imports.open;
+		}
+	};
 </script>
 
 <Modal bind:this={modal} title="Integrate upstream changes" {onClose} width="small" noPadding>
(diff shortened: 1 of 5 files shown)


apps/desktop/src/lib/components/IntegrateUpstreamModal.svelte
```diff
@@ -0,0 +1,239 @@
...
+		integratingUpstream = 'inert';
+		expanded = false;
+		branchStatuses = upstreamIntegrationService.upstreamStatuses();
+		await tick();
+		if (branchStatuses) {
+			modal?.show();
+			return;
+		}
+		await integrate();
+	}
```
estib-vega: This triggers the integration of the upstream changes.
If there are no upstream statuses returned, i.e there are no foreseeable conflicts, the changes are integrated.
If there are foreseeable conflicts, display the modal.

What do we think about this approach?
I like the fact that this treats the `IntegrateUpstreamModal` component as a class with methods as is intended in Svelte, but I'm not sure if this is something we like as a pattern for this specific case. 
Let me know :)

Caleb-T-Owens (approved):

PR #4964, opened 2024-09-24
`gitoxide` for commit with headers

Byron:
## ☕️ Motivation

During the [GitMerge](https://git-merge.com) @schacon talked about "Abusing Git for GitButler". One particular 'hack' stuck out for me, which is the 'dance' that is required to create signed commits with multiple headers.
This shouldn't even be necessary, and with `gitoxide` it should be very straightforward already.
Let's put it to the test.

## Tasks

* [x] some unrelated fixes to get the test-suite to pass
* [x] use `gix` for signed commit creation.
* [x] double-check the signatures are considered valid
* [x] cleanup

## Notes for the Reviewer

* This is still suboptimal as ideally `gix` provides commits with headers out of the box, and ideally signs them as well if that is configured. But I will outscope this for now in order to focus on merging.
* Validation: here is a commit generated with the new code, after asking to show the signature:
```
commit 15457fedf8693eebb77263d21b7a8b55b1fd82d2
gpg: Signature made Tue Sep 24 13:06:41 2024 CEST
gpg:                using RSA key EAF9CCFCD0876408F297C60A9CB5EE7895E8268B
gpg: Good signature from "Sebastian Thiel (YubiKey USB-C) <email>" [ultimate]
gpg:                 aka "Sebastian Thiel (In Rust I trust) <email>" [ultimate]
Date:   Tue Sep 24 13:05:13 2024 +0200

    test commit f

diff --git a/that b/that
new file mode 100644
index 0000000..e69de29
diff --git a/this b/this
new file mode 100644
index 0000000..45b983b
--- /dev/null
+++ b/this
@@ -0,0 +1 @@
+hi
```


krlvi (approved):
way nicer than the header injecting we did before!

PR #4971, opened 2024-09-24
Make the Select component generic

estib-vega:
Auto-detect the type of the value the select options can have


H1ghBre4k3r (approved):

Caleb-T-Owens (approved):
Looks reasonable.

PR #4974, opened 2024-09-24
Temporarily replace create_wd_tree with index operations

Caleb-T-Owens:
@Byron, We've had some other strange behavior that might be related to create_wd_tree on @krlvi's computer that we're looking into. We've decided to prepare a release with the old index operations in place as a hot fix in case further investigation shows that this is indeed the problem.

I'll be digging into this a little more tomorrow where we can hopefully either rule it out as a problem, or create a fix.

### DIFF
--- a/crates/gitbutler-repo/src/repository_ext.rs
+++ b/crates/gitbutler-repo/src/repository_ext.rs
@@ -158,71 +158,76 @@ impl RepositoryExt for git2::Repository {
     /// Note that right now, it doesn't skip big files.
     #[instrument(level = tracing::Level::DEBUG, skip(self), err(Debug))]
     fn create_wd_tree(&self) -> Result<Tree> {
-        let repo = gix::open(self.path())?;
-        let workdir = repo.work_dir().context("Need non-bare repository")?;
-        let mut head_tree_editor = repo.edit_tree(repo.head_tree_id()?)?;
-        let status_changes = repo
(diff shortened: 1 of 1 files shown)


krlvi (approved):

krlvi:
Hmm, I am not sure what to make of the failing test. But also at the same time, the issue that i hit earlier today happens even when running from this branch, so in that particular case `get_wd_tree` was not the issue

Caleb-T-Owens:
> Hmm, I am not sure what to make of the failing test. But also at the same time, the issue that i hit earlier today happens even when running from this branch, so in that particular case get_wd_tree was not the issue.
..And the app feels quite sluggish


Oooh, I didn't realise that... that is rather curious

Byron:
It seems the issue is more complex then.

And I think it's a hint at how to do this conversion: only one step at a time. After all, test-coverage isn't complete and the real-world is complex, so ultimately we will probably find most issues through manual testing. Then hopefully we'd know what conversion step to undo to see if it was the cause.

Alternatively, and if the problem was known to be working before, one could probably just bisect and know which commit introduced the issue.

When `merge_trees` can be done by `gitoxide`, I think I will focus on the highest value target, i.e. the one that is slowest, and then pace the introduction of using `merge_trees` in more places lower the risk of multiple things breaking at the same time.

PR #4972, opened 2024-09-24
feat: new stacking ui commit lines

ndom91:
## ☕️ Reasoning

- More initial stacking UI changes including:
	- Commit Lines
	- Commit Card
	- Commit List

## 🧢 Changes

- Since we want to maintain both functionalities for now, I created a second version of `packages/ui/commitLines` in `packages/ui/commitLinesStacked/*` as well as second copies of the `CommitList` and `CommitCard`. This will should be relatively easy to switch over once the time comes.
- Still toggle-able via the feature flag in the global settings UI as well as via the `stack` keyboard shortcut

## 📺 Current Screenshots

<table>
<tr>
<td>

image

</td>
<td>

image

</td>
</tr>
</table>

PR #4978, opened 2024-09-25
feat: stacking lanes and new branch headers

ndom91:
## ☕️ Reasoning

- Requires and follows: https://github.com/gitbutlerapp/gitbutler/pull/4972
- Adds:
	- New components for stacking branch, commit, header, "new series in stack" card and others
	- Individual series PR card
	- `StackingStatusIcon.svelte`
	- `LaneController.ts`
	- `apps/desktop/src/lib/lane/*` Components

## 🧢 Changes

<table>
<tr>
<td>

image

</td>
<td>

image

</td>
</tr>
</table>

## 🚧 New Vocab

The terminology we've decided on for now at least is as follows:

1. Stack (i.e. Lane)
2. _has many_ Series (i.e. Branches)
3. _has many_ Patches (i.e. stable Commit)
4. _has many_ historical commits (i.e. for diffing based on `changeId`)


ndom91:
A few things left to do here:
- [ ] Ensure multiple branches are passed to the `StackingLaneBranches` component (`apps/desktop/src/lib/branch/StackingLane.svelte:219`)
- [ ] "Upstream Commits" accordion

UI stuff that still needs to be done:
- [ ] Get branch status to match branch header line colors/icon
- [ ] Add connecting line in bg between branches
- [ ] Deal with correct `color`/`backgroundColor` for `StackingStatusIcon.svelte`
- [ ] Sticky "Push" btn
- [ ] Probably a few other minor things, will double check on Monday

image

ndom91:
## Todo

- [ ] Cleanup "Add Branch to Stack"
- [ ] Create "Push" (whole stack) logic
- [ ] Link up other actions like delete series (branch)

apps/desktop/src/lib/branch/StackingBranchHeader.svelte
```diff
@@ -0,0 +1,228 @@
...
+	import { BranchController } from '$lib/vbranches/branchController';
+	import { VirtualBranch } from '$lib/vbranches/types';
+	import Button from '@gitbutler/ui/Button.svelte';
+	import type { PullRequest } from '$lib/gitHost/interface/types';
+
+	// interface Props {
+	// 	upstreamName: string | undefined;
+	// }
+	//
+	// const { upstreamName }: Props = $props();
```
mtsgrd: Can we remove this? *edit: or is it needed to get the branch status perhaps.

apps/desktop/src/lib/branch/stackingUtils.ts
```diff
@@ -0,0 +1,10 @@
+export type BranchColor = 'neutral' | 'integrated';
+
+export function getColorFromBranchType(type: BranchColor) {
+	const colorMap = {
+		neutral: 'var(--clr-scale-ntrl-80)',
+		integrated: 'var(--clr-commit-integrated)'
+	};
+
+	return colorMap[type];
+}
```
mtsgrd: First, should this file be located under `$lib/stack`, and second, should the function be more accurately called `getColorFromCommitStatus`? The const map might also be better placed outside the function.

apps/desktop/src/lib/commit/StackingUpstreamCommitsAccordion.svelte
```diff
@@ -0,0 +1,116 @@
...
+
+	interface Props {
+		children: Snippet;
+		action: Snippet;
+		remoteCommitCount: number;
+	}
+
+	let isOpen = $state(false);
+
+	const { children, action, remoteCommitCount }: Props = $props();
```
mtsgrd: Just for good form, let's move this right underneath the `Props` interface declaration.

apps/desktop/src/lib/commit/StackingUpstreamCommitsAccordion.svelte
```diff
@@ -0,0 +1,116 @@
...
+				display: block;
+				color: var(--clr-text-1);
+				width: 100%;
+			}
+		}
+
+		&:focus {
+			outline: none;
+		}
+	}
```
mtsgrd: I feel like we could flatten this a bit since component css is isolated and won't interfere with other components. In other words, I don't think we should nest `.accordion-row` under `.accordion`.

apps/desktop/src/lib/branch/StackingBranchHeader.svelte
ndom91: Yeah that's still not implemented and I thought it might be useful, but we already have the branch store here that also has the usptream name. Lets get rid of it for now :+1:

apps/desktop/src/lib/branch/stackingUtils.ts
ndom91: Yeah I'm not very happy with this thing, I'd love spend some time to figure out a better way to calculate the color of these icon things.

To asnwer your question specifically, it currently controls both the line and icon colors for these two things:

image

Which are the located in the branches (series) (n-times per stack) and the "new branch (series)" header, of which there is only 1x per stack, so this felt like a logical parent. Happy to hear any better ideas though :+1:

apps/desktop/src/lib/stack/types.ts
```diff
@@ -0,0 +1,17 @@
+import { DetailedCommit, VirtualBranch } from '$lib/vbranches/types';
+import { Type } from 'class-transformer';
+
+export class Series {
```
mtsgrd: It's now called `PatchSeries` on the rust side, let's mirror that name here?

apps/desktop/src/lib/commit/StackingUpstreamCommitsAccordion.svelte
ndom91: Makes sense :+1:

apps/desktop/src/lib/stack/types.ts
ndom91: Yeah good idea. Do we want to update our vocabulary in general to replace `Series` with `PatchSeries` or only use that in the code? :sweat_smile:

mtsgrd: Actually - great question, I'm not sure. @krlvi didn't object to me calling it `PatchSeries` in the rust code, and I did that because `Series` on its own is ambiguous, just like `Sequence` or `Group` would be.

ndom91:
@mtsgrd These are the open UI things that I had on my list, if you want to help knock any of these out that'd be great:

- [ ]  Sticky "Push" btn white bg design
- [ ] Add connecting line on stack bg between the individual series (branches)
- [ ] Improve `StackingBranchHeader` and `StackingStatusIcon` color function
  - [ ] Make branch header color match color of top commit
  - [ ] StatusIcon bg/foreground color is inverted only in its use in the BranchHeader
- [ ] Add hardcoded 3-dot SVG pavel just shared in #design in the closed accordion top row

mtsgrd:
Thanks for sharing the todo list. Before I work on that, I noticed new refs don't render correctly so I'll start by debugging that. See screenshot for a commit that has "push ref" but isn't grouped into its own patch series.
image

ndom91:
Current status, most of the UI things from above are implemented:

<table>
<tr>
<td>

image

</td>
<td>

image

</td>
</tr>
</table>

ndom91:
> Thanks for sharing the todo list. Before I work on that, I noticed new refs don't render correctly so I'll start by debugging that. See screenshot for a commit that has "push ref" but isn't grouped into its own patch series. image

Hmm, I was able to both reproduce and not reproduce this :sweat_smile:

image

apps/desktop/src/lib/pr/StackingPullRequestCard.svelte
```diff
@@ -34,39 +37,44 @@
...
 	const timeAgo = $derived($lastFetch ? createTimeAgoStore($lastFetch) : undefined);
 
 	const mrLoading = $derived(prMonitor?.loading);
-	const checksLoading = $derived($checksMonitor?.loading);
+	const checksLoading = $derived(checksMonitor?.loading);
 
-	const checksError = $derived($checksMonitor?.error);
+	$effect(() => {
+		console.log($checksLoading);
+	});
```
ndom91, line 69: Just wanted to note that there's an `$inspect` rune for this now (see: https://svelte-5-preview.vercel.app/docs/runes#$inspect) @mtsgrd

PavelLaptev (approved):

PR #4981, opened 2024-09-26
Add helper function for matching trees, and start testing rebasing.

Caleb-T-Owens:
Basic rebasing tests


krlvi (approved):
best code I've ever seen!

crates/gitbutler-repo/src/rebase.rs
```diff
@@ -419,11 +421,237 @@ fn resolve_index(
 
 #[cfg(test)]
 mod test {
+    use std::path::Path;
```
Byron, line 424: I highly recommend to put tests for the public API into integration-`./tests` next to `./src`. That way, the source file won't be cluttered up and one is assured to test like one would as consumer of the library.

PR #4985, opened 2024-09-28
fix: Test for integrateUpstreamModal being undefined

estib-vega:
Don't try to read a property from the integration modal if it's undefined

Weirdly this wasn't caught by eslint at all


Caleb-T-Owens (approved):
LGTM; I'm a little bit puzzled how this didn't get caught by the type system.

estib-vega:
> LGTM; I'm a little bit puzzled how this didn't get caught by the type system.

@Caleb-T-Owens Me neither. That's a bit scary, actually

PR #4987, opened 2024-09-28
Add tests for reordering commits

Caleb-T-Owens:
## ☕️ Reasoning

This moves all the logic for reordering commits in to a `reordering_commits.rs` file. This re-implements reordering with an `inner` function which operates using only libgit2 primitives. As such it should be reasonably easy to re-implement in gix.

It also adds a `checkout_branch_trees` function which merges all the trees of the applied branches and checks them out. This is currently untested as the tests depend on `create_wd_tree` behaving correctly.

### DIFF
--- a/crates/gitbutler-branch-actions/src/reorder_commits.rs
+++ b/crates/gitbutler-branch-actions/src/reorder_commits.rs
@@ -8,6 +8,10 @@ use gitbutler_repo::{rebase::cherry_rebase_group, LogUntil, RepositoryExt as _};
 
 use crate::{branch_trees::checkout_branch_trees, VirtualBranchesExt as _};
 
+/// Moves a commit up or down a stack by a certain offset.
+///
+/// After the commit is moved, the combined branch trees are checked out.
+/// A stack must have at least two commits in it. A 0 offset is a no-op.
 ///
 /// Presume we had the stack:
 ///
@@ -23,25 +27,12 @@ use crate::{branch_trees::checkout_branch_trees, VirtualBranchesExt as _};
 /// B
 /// D
 ///
-/// Presume we had the stack:
-///
-/// A
-/// B
-/// C
-/// D
-///
-/// If B was the subject, and the offset was 1, we would expect:
+/// Or, if B was the subject, and the offset was 1, we would expect:
 ///
 /// A
 /// C
 /// B
 /// D
-///
-
-// move a given commit in a branch up one or down one
-// if the offset is positive, move the commit down one
-// if the offset is negative, move the commit up one
-// rewrites the branch head to the new head commit
 pub(crate) fn reorder_commit(
     ctx: &CommandContext,
     branch_id: BranchId,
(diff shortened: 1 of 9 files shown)


crates/gitbutler-branch-actions/src/reorder_commits.rs
```diff
@@ -0,0 +1,639 @@
...
+        .find_commit(subject_commit_oid)
+        .context("failed to find commit")?;
+
+    let ReorderResult { tree, head } = inner_reorder_commit(
+        repository,
+        &repository.find_commit(merge_base)?,
+        &subject_commit,
+        offset,
+        &repository
+            .log(branch.head, LogUntil::Commit(merge_base))?
```
Byron: Are the actual `Commit` objects needed here, or would `Oid` also do the trick?

Byron (commented):
I really like that one is greeted with documentation to help explain how it works. Even the tests have additional explanation, lovely!

*(Not a lot of substance in my comment, but that's all I could say after the glimpse I took)*

crates/gitbutler-branch-actions/src/reorder_commits.rs
Caleb-T-Owens: No, I often start with the Commit objects because they're more strongly typed and it may reduce excessive lookups, but they're certainly not needed here

Byron: I think it's also related to `log()` returning commits by default. What's hidden in `git2` will be obvious with `gix`, and this method (`log`) will definitely start returning `ids` or graph walk items (once backed by `gix`), which allow access to parent-ids and commit-times, as free by-product of the iteration.
Everything else would then be re-obtained on demand only which will still be close-to-free as the objects are usually cached.
If you see no future need for these `Commit` objects, or the need for information on `Commit` only for a few select ones, then it would be beneficial to not use `Commit` objects here in preparation for the future.

Caleb-T-Owens: We have the horrendously named `l` function which returns just the `Oid`s for us.

Btw, when I said `No, ...` I meant, "No, they're no longer needed", and I've updated the functions to only pass the objects between themselves if they're actually needed.

It would be nice if we could have typed Oids because it can be clear looking at a function signature to tell what kind it wants (most of time time its a commit Oid, but our `branch.tree` property is actually the Oid of a tree 😆.

Byron: > It would be nice if we could have typed Oids because it can be clear looking at a function signature to tell what kind it wants (most of time time its a commit Oid, but our `branch.tree` property is actually the Oid of a tree 😆.

I see, that would indeed be good to have. It would be somewhat expensive to do in `gix` as it would have to truly assure the object is of the right type which is a `find_header()` call at least - it's fast, but far from free.
Meanwhile, I think descriptive variable names can go a long way, also knowing that once one tries to get a tree with an Oid of a commit, it will fail as loud as one wishes.

krlvi (approved):
feature flagged. you have been working on this. not much to review here, lgtm :)

PR #4991, opened 2024-09-29
Custom font support for displaying diffs

PavelLaptev:
## ☕️ Reasoning
Allowing users to choose a font tailored to their preferences improves code clarity, reduces eye strain, and offers a more accessible and personalized diff view.

image

### DIFF
--- a/apps/desktop/src/lib/hunk/HunkDiff.svelte
+++ b/apps/desktop/src/lib/hunk/HunkDiff.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
 	import { type Row, Operation, type DiffRows } from './types';
-	import { featureInlineUnifiedDiffs } from '$lib/config/uiFeatureFlags';
 	import ScrollableContainer from '$lib/scroll/ScrollableContainer.svelte';
 	import { create } from '$lib/utils/codeHighlight';
 	import { maybeGetContextStore } from '$lib/utils/context';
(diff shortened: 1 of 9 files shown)


krlvi:
This is cool. By the way i noticed today the ligatures. I think the _default_ font should *not* include ligatures as they are considered non-standard and more ambiguous (e.g --> is more clear that it is 3 characters than the ligature glyph equivalent)

PavelLaptev:
> I think the _default_ font should _not_ include ligatures as they are considered non-standard and more ambiguous (e.g --> is more clear that it is 3 characters than the ligature glyph equivalent)

I'll add the "disable ligatures" option to this PR. It will be "true" by default.

PavelLaptev:
@krlvi updated

image

apps/desktop/src/lib/hunk/HunkDiff.svelte
```diff
@@ -49,13 +51,17 @@
...
+		diffFont,
+		diffLigatures,
 		minWidth,
 		draggingDisabled = false,
 		onclick,
 		handleSelected,
 		handleLineContextMenu
 	}: Props = $props();
 
+	console.log(minWidth);
```
estib-vega: I think you missed this :)

estib-vega (approved):
Approved, I think we can get rid of the extra console.log now or after 👍

estib-vega:
@PavelLaptev done :)

Screenshot 2024-09-30 at 12 41 56

PavelLaptev:
> @PavelLaptev done :)

@estib-vega looks good, thank!

PR #4993, opened 2024-09-29
Use new git2 implementation

Caleb-T-Owens:


krlvi (approved):
the tests are glorious

crates/gitbutler-repo/src/repository_ext.rs
```diff
@@ -634,3 +629,429 @@ impl GixRepositoryExt for gix::Repository {
         Ok(self)
     }
 }
+
+#[cfg(test)]
+mod test {
+    mod create_wd_tree {
```
Byron, line 635: All tests here should move to integration test level, as they are testing a public method.

crates/gitbutler-repo/src/repository_ext.rs
```diff
@@ -156,73 +157,67 @@ impl RepositoryExt for git2::Repository {
...
+
+        Ok(self.find_tree(tree_oid)?)
+    }
+
+    /// Returns the path of the working directory of the repository.
+    fn worktree_path(&self) -> PathBuf {
+        if self.is_bare() {
+            self.path().to_owned()
+        } else {
+            self.path().join("..")
```
Byron, line 219: This is not correct - the git repository can't be assumed to be a `.git` directory in the worktree.
Instead, [workdir()](https://docs.rs/git2/latest/git2/struct.Repository.html#method.workdir) should be used for that purpose, and it's fine to fail if there is none.
Lastly, if the repository is bare, we wouldn't want to treat the content of a git repository as worktree.

crates/gitbutler-repo/src/repository_ext.rs
```diff
@@ -156,73 +157,67 @@ impl RepositoryExt for git2::Repository {
...
+            if status.is_index_new() && status.is_wt_deleted() {
+                // This is a no-op
+            } else if (status.is_index_deleted() && !status.is_wt_new()) || status.is_wt_deleted() {
+                tree_update_builder.remove(path);
+            } else {
+                let file_path = worktree_path.join(path);
+                let file = std::fs::read(file_path)?;
+                let blob = self.blob(&file)?;
+
+                tree_update_builder.upsert(path, blob, git2::FileMode::Blob);
```
Byron, line 204: There are various [file-modes](https://docs.rs/git2/latest/git2/enum.FileMode.html#) possible here and all but `Tree` and `Unreadable` (and probably, `BlobGroupWritable`) should be handled. 

`Link` is particularly important, `BlobExecutable` as well. These cases were handled in the deleted code, so could be brought back from there. Submodules also weren't handled though, maybe it would have failed, which seems OK since submodules aren't supported yet.

crates/gitbutler-repo/src/repository_ext.rs
```diff
@@ -634,3 +629,429 @@ impl GixRepositoryExt for gix::Repository {
...
+        /// | add                | modify            | upsert |
+        /// | modify             | modify            | upsert |
+        #[cfg(test)]
+        mod head_upsert_truthtable {
+            use super::*;
+
+            // | add                | delete            | no-action |
+            #[test]
+            fn index_new_worktree_delete() {
+                let test_repository = TestingRepository::open();
```
Byron, line 662: Here it would be better to work with `git_testsupport::readonly::*` as it allows to create persistent read-only repositories that are faster to run on, while also being easy to introspect. I.e. one can enter the directory and manually validate that the repository is indeed in the state that it should be in, maybe even with a `git status --porcelain` to validate in the test itself to serve as 'show and validation of pre-condition'.

crates/gitbutler-branch-actions/src/branch_trees.rs
```diff
@@ -60,3 +60,103 @@ pub(crate) fn checkout_branch_trees<'a>(
...
+                .force()
+                .remove_untracked()
+                .checkout()
+                .unwrap();
+        }
+
+        // Assert tree is indeed empty
+        {
+            let tree: git2::Tree = test_project.local_repository.create_wd_tree().unwrap();
+            dbg!(tree
```
Byron, line 134: `dbg!()` should be removed.

Byron (commented):
All this is great in principle, as is what the implementation should have been from the beginning. I was too hung-up on thinking 'additive only' is enough, even though it doesn't represent what `create_wd_tree` is actually about - sorry for that :/!

However, as this PR stands some fixes are necessary before this it is released to stable, and a new nightly once the fixes have landed.

crates/gitbutler-repo/src/repository_ext.rs
Caleb-T-Owens: I'm quite a fan of having tests in the same file as the implementation as it makes them very easy to find, but they do make up half the file now, so I'll move them out.

Byron: I mean, at the end of the day, one can also create a module file from the `tests` module, to get it out of the way. Keeping test-code in the `src/` tree also isn't as bad if one doesn't have to optimize their size for release.

In the end, it's just me wanting to remove one source of API errors where one accidentally tests private parts - it's a kind of mistake that usually is discovered when just trying to use the API from any other of the various crates, so after all it's probably never an issue in practice.

PR #4997, opened 2024-09-30
ai: Methods for PR summarization

estib-vega:
## ☕️ Reasoning
This is the first out of 2 PRs for the actual feature.
I split it for easier reviewability (so looking forward to stacking!)

This pull request introduces new functionality to our AI service that improves the process of generating pull request descriptions. By implementing a method that formulates a description based on given context (commit messages + current title and description + optional user prompt). 

This changes also introduce the ability to stream the tokens into the UI, for a better UX.

## 🧢 Changes
- Added a method to the AI service that generates PR description bodies from provided context.
- Integrated the Anthropic SDK to facilitate streaming.
- Implemented an asynchronous transformer to handle multiple requests more efficiently.
- Enabled streaming capabilities for LLM responses, improving real-time feedback in the UI.

### DIFF
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -48,6 +48,7 @@
 		"@types/git-url-parse": "^9.0.3",
 		"@types/lscache": "^1.3.4",
 		"@types/marked": "^5.0.2",
+		"@types/node": "^22.3.0",
 		"@types/postcss-pxtorem": "^6.0.3",
 		"@vitest/ui": "^2.0.5",
 		"@wdio/cli": "^8.39.1",
@@ -81,10 +82,10 @@
 		"tinykeys": "^2.1.0",
 		"ts-node": "^10.9.2",
 		"vite": "catalog:",
-		"vitest": "^2.0.5",
-		"@types/node": "^22.3.0"
+		"vitest": "^2.0.5"
 	},
 	"dependencies": {
+		"@anthropic-ai/sdk": "^0.27.3",
 		"openai": "^4.47.3"
 	}
 }
--- a/apps/desktop/src/lib/ai/anthropicClient.ts
+++ b/apps/desktop/src/lib/ai/anthropicClient.ts
@@ -3,48 +3,77 @@ import {
 	SHORT_DEFAULT_BRANCH_TEMPLATE,
 	SHORT_DEFAULT_PR_TEMPLATE
 } from '$lib/ai/prompts';
-import { type AIClient, type AnthropicModelName, type Prompt } from '$lib/ai/types';
-import { buildFailureFromAny, ok, type Result } from '$lib/result';
-import { fetch, Body } from '@tauri-apps/api/http';
+import {
+	type AIClient,
+	type AIEvalOptions,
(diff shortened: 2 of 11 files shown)


apps/desktop/src/lib/ai/anthropicClient.ts
```diff
@@ -1,45 +1,79 @@
...
-			);
-		}
+		return await andThenAsync(responseResult, async (response) => {
+			const buffer: string[] = [];
+			for await (const event of response) {
+				if (event.type === 'content_block_delta' && event.delta.type === 'text_delta') {
+					const token = event.delta.text;
+					options?.onToken?.(token);
+					buffer.push(token);
+				}
```
estib-vega, line 74: Taken from https://github.com/anthropics/anthropic-sdk-typescript/blob/b7e21131de5cf9d741e3722393285b35eff26c1e/examples/raw-streaming.ts#L20-L23

apps/desktop/package.json
```diff
@@ -81,10 +82,10 @@
 		"tinykeys": "^2.1.0",
 		"ts-node": "^10.9.2",
 		"vite": "catalog:",
-		"vitest": "^2.0.5",
-		"@types/node": "^22.3.0"
+		"vitest": "^2.0.5"
 	},
 	"dependencies": {
+		"@anthropic-ai/sdk": "^0.27.3",
 		"openai": "^4.47.3"
```
mtsgrd, line 89: I have to admit, it's a bit unclear what the difference is between our use of `dependencies` vs. `devDependencies`. For consistency, perhaps we should move both of these? @ndom91 in case you know why we're doing it this way.

apps/desktop/src/lib/ai/anthropicClient.ts
```diff
@@ -1,45 +1,79 @@
...
-type AnthropicAPIResponse = {
-	content: { text: string }[];
-	error: { type: string; message: string };
-};
+const DEFAULT_MAX_TOKENS = 1024;
+
+function splitPromptMessages(prompt: Prompt): [MessageParam[], string | undefined] {
+	const messages: MessageParam[] = [];
+	let system: string | undefined = undefined;
+	for (const m of prompt) {
```
mtsgrd: is `const m` short for `const message`? I might be wrong, but in our code base I think we've so far avoided single letter variables.

mtsgrd (approved):

apps/desktop/package.json
estib-vega: I think the standard way of ordering the NPM dependencies is to add anything that gets bundled into the distributable under `dependencies` and everything not distributed (all the local dev dependencies, test dependencies, linters, etc.) under `devDependencies`.
So I think this change should be conformant.

https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file#skip-to-content

apps/desktop/src/lib/ai/anthropicClient.ts
estib-vega: I'll change that 👍

PR #4998, opened 2024-09-30
Create wd tree improvements

Caleb-T-Owens:


krlvi (approved):

crates/gitbutler-repo/src/repository_ext.rs
```diff
@@ -197,11 +196,28 @@ impl RepositoryExt for git2::Repository {
...
+                        .to_str()
+                        .context("Failed to convert path to str")?;
+
+                    let blob = self.blob(path_str.as_bytes())?;
+                    tree_update_builder.upsert(path, blob, git2::FileMode::Link);
+                } else {
+                    let file = std::fs::read(&file_path)?;
+                    let blob = self.blob(&file)?;
+
+                    let file_type = if is_executable(&file_path.metadata()?) {
```
Byron, line 213: nit: I'd get `file_path.metadata()` first, and do the file-kind checks through that. `file_path.is_symlink()` fetches its own under the hood and I don't think it's cached. Right now, this means two `stat` calls for non-symlinks, instead of one. It's a nit because this isn't a hot loop.

crates/gitbutler-repo/src/repository_ext.rs
```diff
@@ -197,11 +196,28 @@ impl RepositoryExt for git2::Repository {
...
+                if file_path.is_symlink() {
+                    let resolved_path = file_path.read_link()?;
+                    let path_str = resolved_path
+                        .to_str()
+                        .context("Failed to convert path to str")?;
+
+                    let blob = self.blob(path_str.as_bytes())?;
+                    tree_update_builder.upsert(path, blob, git2::FileMode::Link);
+                } else {
+                    let file = std::fs::read(&file_path)?;
```
Byron, line 210: Instead of reading the file into memory and than hashing it, the hash can be computed [in a streaming manner](https://docs.rs/git2/latest/git2/struct.Repository.html#method.blob_path).

crates/gitbutler-testsupport/src/testing_repository.rs
```diff
@@ -94,16 +108,60 @@ pub fn assert_tree_matches<'a>(
...
+    Blob,
+    Executable,
+}
+
+pub fn assert_tree_matches_with_mode(
+    repository: &git2::Repository,
+    tree: git2::Oid,
+    files: &[(&str, &[u8], &[EntryAttribute])],
+) {
+    let gix_repository = gix::open(repository.path()).unwrap();
```
Byron, line 135: It's better for tests (and faster) to open like this: `gix::open_opts(path, gix::open::Options::isolated())`.

Byron (commented):
Thanks a lot for the fixes! Now it should do the trick for the most part.

----
On another note, I believe to now realize that partial checkouts are also not handled here, even though similarly to submodules, we probably won't open such repos in the first place.
This is more of a note to myself when implementing this for `gitoxide` as it should at least fail gracefully (instead of creating partial trees).

Caleb-T-Owens:
Hi @Byron, thanks for the feedback, I'll add those comments to my TODO list as there are some other things that I'd like to focus on before them

PR #5002, opened 2024-09-30
fix: Correctly detect conflicted files

estib-vega:
Correctly detect and display whether the commit file is conflicted in the edit mode page


Caleb-T-Owens (approved):
LGTM :+1:

Thanks for this

PR #5004, opened 2024-10-01
Fix applying mergable branch

Caleb-T-Owens:
Testing this is a massive pain in the ass. I have tried two different approaches and both are painful.


krlvi (approved):
read through the implementation. seems to make sense. hard to know with 100% without a test, however, lets merge and evaluate

PR #5005, opened 2024-10-01
Add patch series to vbranch API struct

mtsgrd:


krlvi (approved):
the docs on it suck, but we can land it and update them later

PR #5016, opened 2024-10-02
Factor out the `move_commit` function

estib-vega:
### Motivation
Make more maintainable and testable the logic behind the `move_commit`.

### Changes
- Move the `move_commit` function to a separate module.
- Prefer using the `cherry_rebase_group` instead of calculating the commit out of the hunks to be moved on top of the destination branch.
- ⚠️ Bug fix: Stop the commit transferal if locked to any ancestor commit.
- Add some unit tests for testing whether the commit is locked.
- Add some more integration tests (rust-end only) for more scenarios of moving commits in between virtual branches


estib-vega:
cc @Pranav2612000

Caleb-T-Owens (approved):
LGTM

PR #5017, opened 2024-10-02
Edit mode: Add a max height and scrollable container

estib-vega:
In order to be able to display large edit mode files list without having to scroll the whole page to the bottom, add a scrollable container to the commits file list

### Changes
- Scrollable area
- Number of files is displayed in header (to make it clear that there are potentially more files to scroll to)
- Open context menu on click as well as right-click (seems to be a better UX imo)


PavelLaptev (approved):
@estib-vega Looks great! Thanks for the PR.
- I removed one unused component
- got rid of the scrollbar on the main view
- and replaced the text with a files counter badge.
- I also removed the left line with the avatar since the new commit cards will look different. Not fully finish this part, but I can fix that later.

PR #5023, opened 2024-10-02
fix: diff example "added/deleted" lines

PavelLaptev:
Before:
image

After:
image


Caleb-T-Owens (approved):
It was impolite of me to point it out while you were presenting, so I apologies.

Thank you for the fix though.

PavelLaptev:
> It was impolite of me to point it out while you were presenting, so I apologies.
> 
> Thank you for the fix though.

@Caleb-T-Owens not at all. Thanks for that :-)

PR #5028, opened 2024-10-03
fix: Update the type of the drop down button

estib-vega:
Fix up the type of variable used to refernce the drop down button in the PR button component


ndom91 (approved):
Nice catch!

PR #5041, opened 2024-10-04
feat: series header description field

ndom91:
## ☕️ Reasoning

- Allow both adding and removing "description" field to series via kebab menu "Add/Remove description" option on Series header (`StackingBranchHeader.svelte`)
- Extract `InputLabel.svelte` into own component and allow two variants - `input` and `textarea` for single and multi-line versions

## 🧢 Changes

- Extracted existing `BranchLabel.svelte` combination label/input field into `InputLabel.svelte` which then is used by `BranchLabel`/`StackingBranchDescription`(new)


apps/desktop/src/lib/shared/LabelInput.svelte
```diff
@@ -0,0 +1,172 @@
...
+	}, 100)}
+	class="label-input-measure-el"
+	class:text-12={multiline}
+	class:text-14={!multiline}
+	class:wrap={multiline}
+	style:width={multiline ? textAreaWidth : 'inherit'}
+>
+	{value}
+</span>
+{#if multiline}
```
ndom91: The previous `svelte:component` dynamic component mechanism is deprecated and the new dynamic component stuff didnt work either, i.e. something like

```tsx
InputComponent = multiline ? new HTMLTextAreaElement() : new HTMLInputElement()
``` 

and neither did that with `document.createElement()` 

_I think_ it only works with Svelte components, not vanilla html elements. But I'd love to simplify this some more, the `textarea`/`input` are so similar it feels like a waste having a giant `if` here, so I'm open to any ideas :+1:

Caleb-T-Owens: `is` would be wonderful, but safari doesn't support it D: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/is

Caleb-T-Owens: You could define a svelte component with is `TextInput`, which you can pass all the arguments to. It can then be like:

```svelte
<script lang="ts">
  interface Props { ... }

  const { multiline, ...args }: Props = $props()
</script>

{#if multiline}
  <textarea {...args}></textarea>
{:else}
  <input {...args}></input>
{/if}
```

Caleb-T-Owens: OOrrrrrrr, create a svelte component wrapper for `<Input />` and `<TextArea />` and then use svelte's "correct" mechanism on those.

ndom91: Yeah was thinking about a wrapper component too, but then I figured we're just adding another layer of abstraction on top of the same `if` block, but I overlooked the ability to spread all the props in now. I like this idea!

ndom91:
This PR is waiting on this ESLint issue to be fixed first: https://github.com/gitbutlerapp/gitbutler/pull/5083

ndom91:
There have been too many upstream changes. This branch / PR has been replaced via: https://github.com/gitbutlerapp/gitbutler/pull/5172

PR #5042, opened 2024-10-04
fix: use `showStackingDetails` persisted writable instead of function import

ndom91:
## ☕️ Reasoning

- Changes weren't reactive across other instances

## 🧢 Changes

- Use exported persisted writable instead of function which returns that


Caleb-T-Owens (approved):
Lovely Jubly

PR #5044, opened 2024-10-04
fix: upstream commits accordion and cards

ndom91:
## ☕️ Reasoning

- Fix Upstream-only commit lines coloring
- Fix clicking on one of those upstream `CommitCard`s. It would previously propagate the click event on so that the `toggle` would be triggered on the upstream commits accordion, instead of loading the files of the commitcard

### Before

image

### After

image


Caleb-T-Owens (approved):
LGTM 👍

PR #5048, opened 2024-10-07
[#4889] feat: allow moving any commit from a stack into another lane

### LINKED ISSUE #4889: Allow moving any commits from a stack into another lane
Currently we only allow moving the top commit of a stack between lanes. This is cumbersome, but done as historically we had no way of resolving the rebase conflicts that might have come out of the operation. Now that we can resolve these conflicts correctly, this is not an issue.

Pranav2612000:
## ☕️ Reasoning
Currently we only allow moving the top commit of a stack between lanes. This PR now also allows the user to move a non-head commit to the top of a different lane ( provided that the commit is not locked to its ancestors & children )

## 🧢 Changes
- Remove the UI check preventing the movement of a non-top commit to a different lane
- Propogate the source commit branch to the `move_commit` func. This is now required as there is no way to determine which branch a commit belongs to given just the commitId
- Perform the actual commit move
    - Add a check to ensure that none of the commits which occur after the source commit are locked to it
    - Apart from rebasing the source commit onto the new lane, now we also need to rebase the commits above the source commit to the parent of the source commit
    - Update the head of source branch ( destination branch head is already updated )
- To make the above change easier, we also update the `check_source_lock_to_ancestors` func to `check_source_lock_to_commits` and make it work without taking ownership of the commitIds to check for.

## 🎫 Affected issues

Fixes: #4889

## 🔬 Tests
_WIP_

### DIFF
--- a/apps/desktop/src/lib/branches/dragActions.ts
+++ b/apps/desktop/src/lib/branches/dragActions.ts
@@ -11,15 +11,12 @@ class BranchDragActions {
 
 	acceptMoveCommit(data: any) {
 		return (
-			data instanceof DraggableCommit &&
-			data.branchId !== this.branch.id &&
-			data.isHeadCommit &&
-			!data.commit.conflicted
+			data instanceof DraggableCommit && data.branchId !== this.branch.id && !data.commit.conflicted
 		);
 	}
 
 	onMoveCommit(data: DraggableCommit) {
-		this.branchController.moveCommit(this.branch.id, data.commit.id);
+		this.branchController.moveCommit(this.branch.id, data.commit.id, data.commit.branchId);
 	}
 
 	acceptBranchDrop(data: any) {
--- a/crates/gitbutler-branch-actions/src/move_commits.rs
+++ b/crates/gitbutler-branch-actions/src/move_commits.rs
@@ -83,7 +83,7 @@ pub(crate) fn move_commit(
     let ancestor_commits = ancestor_commits;
 
     let is_ancestor_locked =
-        check_source_lock_to_ancestors(ctx.repository(), ancestor_commits, &branch_head_diff);
+        check_source_lock_to_commits(ctx.repository(), &ancestor_commits, &branch_head_diff);
 
     if is_source_locked {
         bail!("the source branch contains hunks locked to the target commit")
@@ -160,29 +160,29 @@ fn check_source_lock(
     is_source_locked
 }
 
-/// determines if the source commit is locked to any of its ancestors
-fn check_source_lock_to_ancestors(
+/// determines if the source commit is locked to any commits
+fn check_source_lock_to_commits(
     repository: &git2::Repository,
-    ancestor_commits: Vec<git2::Commit>,
+    commits: &Vec<git2::Commit>,
     source_commit_diff: &HashMap<std::path::PathBuf, Vec<gitbutler_diff::GitHunk>>,
 ) -> bool {
-    let mut previous: Option<git2::Commit> = None;
+    let mut previous: Option<&git2::Commit> = None;
 
-    for ancestor_commit in ancestor_commits {
+    for commit in commits {
         if previous.is_none() {
-            previous = Some(ancestor_commit);
+            previous = Some(commit);
             continue;
         }
 
         let previous_commit = previous.take().unwrap();
 
-        let old_tree = ancestor_commit.tree().unwrap();
+        let old_tree = commit.tree().unwrap();
         let new_tree = previous_commit.tree().unwrap();
 
         let diff = gitbutler_diff::trees(repository, &old_tree, &new_tree);
 
         if diff.is_err() {
-            previous = Some(ancestor_commit);
+            previous = Some(commit);
             continue;
         }
 
@@ -210,7 +210,7 @@ fn check_source_lock_to_ancestors(
             return true;
         }
 
-        previous = Some(ancestor_commit);
+        previous = Some(commit);
     }
 
     false
--- a/crates/gitbutler-branch-actions/src/move_commits.rs
+++ b/crates/gitbutler-branch-actions/src/move_commits.rs
@@ -17,6 +17,7 @@ pub(crate) fn move_commit(
     target_branch_id: StackId,
     commit_id: git2::Oid,
     perm: &mut WorktreeWritePermission,
+    source_branch_id: StackId,
 ) -> Result<()> {
     ctx.assure_resolved()?;
     let vb_state = ctx.project().virtual_branches();
@@ -33,15 +34,16 @@ pub(crate) fn move_commit(
 
     let (ref mut source_branch, source_status) = applied_statuses
         .iter_mut()
-        .find(|(b, _)| b.head() == commit_id)
-        .ok_or_else(|| anyhow!("commit {commit_id} to be moved could not be found"))?;
+        .find(|(b, _)| b.id == source_branch_id)
+        .ok_or_else(|| anyhow!("the source branch could not be found"))?;
 
+    let is_head_commit = commit_id == source_branch.head();
     let source_branch_non_comitted_files = source_status;
 
     let source_commit = ctx
         .repository()
         .find_commit(commit_id)
-        .context("failed to find commit")?;
+        .with_context(|| format!("commit {commit_id} to be moved could not be found"))?;
 
     if source_commit.is_conflicted() {
         bail!("Can not move conflicted commits");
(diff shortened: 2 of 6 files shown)


apps/desktop/src/lib/branches/dragActions.ts
```diff
@@ -13,13 +13,13 @@ class BranchDragActions {
 		return (
 			data instanceof DraggableCommit &&
 			data.branchId !== this.branch.id &&
-			data.isHeadCommit &&
+			// data.isHeadCommit &&
```
estib-vega: This will need to just be deleted once it becomes a non-WIP PR. It's ok for now

crates/gitbutler-branch-actions/src/move_commits.rs
```diff
@@ -46,6 +47,14 @@ pub(crate) fn move_commit(
...
     }
 
+    let is_head_commit = commit_id == source_branch.head;
+    if !is_head_commit {
+        let ids_to_rebase = ctx.repository().l(source_branch.head, LogUntil::Commit(commit_id)).unwrap();
+
+        let _ = cherry_rebase_group(ctx.repository(), source_commit.parent_id(0).unwrap(), &ids_to_rebase, false);
+    }
+
+
```
estib-vega: This operation makes sense in principle, but you're missing some things:

1. This needs to be executed only after checking that the commit that's being moved (`source_commit`) is not locked to it's descendants (`ids_to_rebase`). 
In order to do that, you can just rename the `check_source_lock_to_ancestors` function to something more generic like `check_source_lock_to_commits`, and pass the corresponding commits derived from the `ids_to_rebase`.

2. `cherry_rebase_group` returns a commit ID that then needs to be set to the `source_branch.head`.
Almost at the end though, just like we do for the `destination_branch.head` (line 126):

```rs
 let new_destination_head_oid = cherry_rebase_group(
        ctx.repository(),
        destination_branch.head,
        &[source_commit.id()],
        true,
    )?;

    // @Pranav, here's the place you should add the rebase operation.
    // The returned commit ID should be set to the source_branch.head

    // reset the source branch to the parent commit
    // and update the destination branch head
    source_branch.head = source_branch_head_parent.id();
    vb_state.set_branch(source_branch.clone())?;

    destination_branch.head = new_destination_head_oid;
    vb_state.set_branch(destination_branch.clone())?;
```

3. Let's try to avoid to `unwrap`. In order to get the parent commit, you can already use the following variable (line 58)
```rs
    let source_branch_head_parent = source_commit
        .parent(0)
        .context("failed to get parent commit")?;
```

estib-vega (changes requested):
It's a great place to start. Let's just iterate on this a bit.

crates/gitbutler-branch-actions/src/move_commits.rs
```diff
@@ -32,7 +33,7 @@ pub(crate) fn move_commit(
 
     let (ref mut source_branch, source_status) = applied_statuses
         .iter_mut()
-        .find(|(b, _)| b.head == commit_id)
+        .find(|(b, _)| b.id == source_branch_id)
         .ok_or_else(|| anyhow!("commit {commit_id} to be moved could not be found"))?;
```
estib-vega: Let's update this error in order to convey that actually the source branch is the thing that could not be found

crates/gitbutler-branch-actions/src/move_commits.rs
```diff
@@ -92,6 +93,18 @@ pub(crate) fn move_commit(
         bail!("the source branch contains hunks locked to the target commit ancestors")
     }
 
+    let mut commits_to_rebase = vec![];
+    let is_head_commit = commit_id == source_branch.head;
+    if is_head_commit {
```
estib-vega: This should be `if !is_head_commit {` :)

crates/gitbutler-branch-actions/src/move_commits.rs
```diff
@@ -163,12 +183,12 @@ fn check_source_lock(
 }
 
 /// determines if the source commit is locked to any of its ancestors
-fn check_source_lock_to_ancestors(
+fn check_source_lock_to_commits(
     repository: &git2::Repository,
-    ancestor_commits: Vec<git2::Commit>,
+    ancestor_commits: &Vec<git2::Commit>,
     source_commit_diff: &HashMap<std::path::PathBuf, Vec<gitbutler_diff::GitHunk>>,
```
estib-vega: Nice!
Now let's get rid of all references to the word "ancestor", please.
Since this function is now 'generic' we can refer to `commits` instead of `ancestor_commits`.

estib-vega (changes requested):
We're getting closer!

I noticed there are some linter errors as well.
A small dev tip: We got a couple of commands that you can run to check locally whether the formatting is right or not:

- `pnpm lint` for the frontend changes
- `cargo fmt --check --all` for the rust formatting

One rust test seems to be also failing, but I can help out with that after the next batch of changes

crates/gitbutler-branch-actions/src/move_commits.rs
```diff
@@ -92,6 +93,22 @@ pub(crate) fn move_commit(
...
+        commits_to_rebase = ctx
+            .repository()
+            .log(source_branch.head(), LogUntil::Commit(commit_id))
+            .unwrap();
+
+        let is_parent_locked =
+            check_source_lock_to_commits(ctx.repository(), &commits_to_rebase, &branch_head_diff);
+
+        if is_parent_locked {
+            bail!("the source branch contains hunks locked to the target commit parents")
```
Pranav2612000: @estib-vega Do you think this error message is correct? Or should it be something like
`the source branch contains hunks locked to the target commit children`?

estib-vega:
Sorry for the big delay, taking a look at this again

estib-vega:
It's working great! :)

crates/gitbutler-branch-actions/src/move_commits.rs
```diff
@@ -70,32 +72,50 @@ pub(crate) fn move_commit(
...
 
     if is_ancestor_locked {
-        bail!("the source branch contains hunks locked to the target commit ancestors")
+        bail!("the target commit contains hunks locked to its ancestors")
+    }
+
+    if let Some(commits_to_check) = descendant_commits.as_mut() {
+        // we append the source commit so that we can create the diff between
+        // the source commit and its first descendant
+        commits_to_check.push(source_commit.clone());
```
estib-vega, line 107: This is a little fix I did on top.
We need to append the source commit to the commit vector we pass into the `check_source_lock_to_commits` in because that's how the diffs are calculated in that function.

That's not immediately clear nor explained anywhere, so that's my bad!

The `check_source_lock_to_commits` function takes in the diff of the commit we're attempting to move and its parent.
It then calculates all of the diffs of the passed in commits.

So if we have a branch with commits `A, B, C, D` (`A` being the top-most commit) and want to move `C`, to check if it's locked to it's descendants (`A` and `B`), we need to pass in the diff between `C` and `D` (`source_commit_diff`) and the commits `A, B, C` (`commits`).

The function will calculate, in order, the diff between `A` - `B` and `B` - `C`. And compare each diff to see if it overlaps with the `source_commit_diff` lines

estib-vega (approved):

krlvi:
https://github.com/user-attachments/assets/0158f895-e3fa-441d-9bf9-1914dc251b4d

It appears that the behavior here is incorrect

Pranav2612000:
Sorry for missing this @krlvi 
Looking into this immediately.

Do we know if the `log()` function here includes the commit passed to it as argument? ( Ref: https://github.com/Pranav2612000/gitbutler/blob/c9f2f479e525973e19794e63e3d9bb591fa41da9/crates/gitbutler-branch-actions/src/move_commits.rs#L89 )

estib-vega:
> Sorry for missing this @krlvi 
> 
> Looking into this immediately.
> 
> 
> 
> Do we know if the `log()` function here includes the commit passed to it as argument? ( Ref: https://github.com/Pranav2612000/gitbutler/blob/c9f2f479e525973e19794e63e3d9bb591fa41da9/crates/gitbutler-branch-actions/src/move_commits.rs#L89 )

@Pranav2612000 Don't worry, it was my bad 😅

I submitted a fix here:

https://github.com/gitbutlerapp/gitbutler/pull/5146

Pranav2612000:
Thank you very much @estib-vega !

So was the problem that since we weren't cloning `commits_to_check` before, the `source_commit` was being added to the list of commits to be rebased later?

Now that we're pushing to the `source_commit` to the clone, the original `descendant_commits` remains unchanged?

estib-vega:
> Thank you very much @estib-vega !
> 
> So was the problem that since we weren't cloning `commits_to_check` before, the `source_commit` was being added to the list of commits to be rebased later?
> 
> Now that we're pushing to the `source_commit` to the clone, the original `descendant_commits` remains unchanged?

@Pranav2612000 That's exactly right.

`descendant_commits` excludes the source commit, but we need it before checking for descendant lock.

And when rebasing the commits, the source commit should not be in there.

PR #5050, opened 2024-10-07
Fix upstream integration when there are no new branches

Caleb-T-Owens:
This is important as it will otherwise look like you've got the inverse diff in your uncommited changes.


estib-vega (approved):
LGTM 👍

PR #5051, opened 2024-10-07
improvement: Allow for string error toasts

estib-vega:
Passing a string as the error to be displayed in the toast will display it as is


Caleb-T-Owens (approved):
LGTM. I'm a little  that we're actually getting string errors because I thought we had a lint error that forced you to throw error objects.

estib-vega:
> LGTM. I'm a little that we're actually getting string errors because I thought we had a lint error that forced you to throw error objects.

@Caleb-T-Owens It seems like not everywhere at least.
I can't find lint error either :/

PR #5058, opened 2024-10-08
New Branch/Stack creation initializes the Stack

krlvi:


krlvi:
@Caleb-T-Owens is there a nicer way to represent this? I wanna have the initializing "constructor" be the only one used. Unfortunately, there is the issue with cyclic dependencies, if I want to keep the `head` field private. 
I thought it's a better trade off to have `new` and `new_initialized` methods than having the `head` field public

PR #5061, opened 2024-10-09
Move dummy UI and minor code refactor

PavelLaptev:
## Description

This pull request includes various changes related to UI components:
- Updated Storybook styles
- Fixed an issue with the Link button prop
- Moved the borderless textarea to the UI folder
- Added a Storybook story for the BorderlessTextarea component
- Removed the unused `autoHeight` property
- Simplified the names of "Props"
- Migrated the EmptyState component to Svelte 5
- Fixed the autofocus behavior when opening the branches search


packages/ui/src/lib/EmptyStatePlaceholder.svelte
```diff
@@ -1,10 +1,27 @@
+<script lang="ts" module>
+	export interface Props {
```
ndom91, line 2: Are you using this import somewhere? If not, lets move this into the same `<script lang="ts">` block, i.e. down on line ~15

ndom91 (approved):
1 nit, and 2 small ESLint issues in the `*CommitCard.svelte` files, but otherwise LGTM.

packages/ui/src/lib/EmptyStatePlaceholder.svelte
PavelLaptev: Now it's unfortunately needed for Storybook

Here is an example:
image

PR #5066, opened 2024-10-09
fix: stackingCommitList reorder-dropzone and "add empty commit" btn rendering

ndom91:
## ☕️ Reasoning

- `StackingCommitList` "Insert Empty Commit" buttons were misplaced
- Dropzone line was rendering in the wrong place
- Converted `InsertEmptyCommitAction` to runes syntax

## 🧢 Changes

image

image

### DIFF
--- a/apps/desktop/src/lib/commit/StackingCommitList.svelte
+++ b/apps/desktop/src/lib/commit/StackingCommitList.svelte
@@ -284,7 +284,6 @@
 		flex-direction: column;
 		background-color: var(--clr-bg-2);
 		border-radius: 0 0 var(--radius-m) var(--radius-m);
-		overflow: hidden;
 	}
 
 	.commits-group {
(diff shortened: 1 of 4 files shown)


apps/desktop/src/lib/commit/StackingCommitList.svelte
```diff
@@ -108,7 +108,7 @@
 		isMiddle?: boolean;
 		isLast?: boolean;
 	}) {
-		if (isFirst) return 12;
+		if (isFirst) return 0;
```
Caleb-T-Owens: Does this behavior need to be dependent on the stacking flag?

krlvi (approved):

apps/desktop/src/lib/commit/StackingCommitList.svelte
ndom91: No, it's in the `StackingCommitList` component here which is only used when the feature flag is on. But actually now that you pointed it out, that entire calculating of `yOffset` can be simplified here

PR #5069, opened 2024-10-09
User: Make email optional

estib-vega:
When authenticating with some IdPs, sharing the email is optional.
In that case, trying to authenticate would just throw a Rust-end error to the privacy-minded user.
This fixes that


Caleb-T-Owens (approved):
Looks good to me :+1:

PR #5076, opened 2024-10-09
LIst conflicted files in commit files list

Caleb-T-Owens:


apps/desktop/src/lib/commit/CommitCard.svelte
```diff
@@ -48,6 +48,8 @@
 	const commitStore = createCommitStore(commit);
 	$: commitStore.set(commit);
 
+	$: console.log(commit);
```
krlvi: you put this here  to test if i read it right?

krlvi (approved):

PR #5079, opened 2024-10-09
Add a shared package

Caleb-T-Owens:
This adds a `@gitbutler/shared` library which can share services and pages between the Web and Desktop apps.

New dependency graph:
```mermaid
flowchart TD
    Desktop --> Ui
    Desktop --> Shared
    Shared --> Ui
    Web --> Ui
    Web --> Shared
```

For sure, in time we will want to have differences between the Web and Desktop experience, but it seems reasonable as a starting point to have them unified.


ndom91 (approved):
In general LGTM, but I think the `terrazzo` config is unnecessary here. The design tokens and the CSS that it generates are located in `@gitbutler/ui` already, we can just import them from there instead of recreating here imo.

PR #5085, opened 2024-10-10
fix: status icon for series header

ndom91:
## ☕️ Reasoning

- Update series header icons depending on branchType


PavelLaptev (approved):
thanks!

PR #5086, opened 2024-10-10
fix: hide upstream related btns if stack/series hasn't been pushed yet

ndom91:
## ☕️ Reasoning

- Hide upstream related btns on series header (i.e. "open in browser" and "create PR") if the stack hasn't been pushed yet.

### DIFF
--- a/apps/desktop/src/lib/branch/StackingBranchHeader.svelte
+++ b/apps/desktop/src/lib/branch/StackingBranchHeader.svelte
@@ -95,17 +95,19 @@
 		<div class="text-14 text-bold branch-info__name">
 			<span class="remote-name">{$baseBranch.remoteName ?? 'origin'}/</span>
 			<BranchLabel {name} onChange={(name) => editTitle(name)} />
-			<Button
-				size="tag"
-				icon="open-link"
-				style="ghost"
-				onclick={(e: MouseEvent) => {
-					const url = gitHostBranch?.url;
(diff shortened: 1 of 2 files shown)


apps/desktop/src/lib/branch/StackingBranchHeader.svelte
```diff
@@ -135,33 +137,43 @@
...
 		</div>
-	</div>
+	{/if}
+	<PrDetailsModal
+		bind:this={prDetailsModal}
+		type="preview-series"
+		{upstreamName}
+		{name}
+		{commits}
+	/>
```
estib-vega, line 164: It's probably not an actual problem, but I think the modal should be a outside the `div`

ndom91: Yeah so it was originally, but I had to move it here to get the DOM structure to work for the css `:last-child`, for hiding the bottom border when appropriate. Putting the modal here shouldn't have any effect on its rendering, etc.

PavelLaptev (approved):

apps/desktop/src/lib/branch/StackingBranchHeader.svelte
estib-vega: Sounds good!

estib-vega (approved):
LGTM

PR #5089, opened 2024-10-10
feat: Parse & display binary files

### LINKED ISSUE #2752: Display images

### LINKED ISSUE #3093: Show file size for binary content

### LINKED ISSUE #4957: Req: Show binary files (like images) in file preview

### LINKED ISSUE #5096: "Failed to read pull request template"
### Version

0.13.0

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

I selected a PR template, but I get

grafik

I checked in Explorer - the file exists:

grafik

---

Think, this refs https://github.com/gitbutlerapp/gitbutler/pull/4814

### How to reproduce

1. Clone JabRef repo
2. Enable pull request templates\
    grafik
3. Close settings
4. See that error pop up that PR template is not found

### Expected behavior

PR template is used.

### Relevant log output

_No response_

Zamoca42:
Hi team! I’m happy to continue working on this task.
I implemented binary file parsing in the backend and continued working on #4852 to view the results.

#4852 had too many changes to apply, making rebase burdensome, so I applied those changes in this PR.

## ☕️ Reasoning

The current implementation doesn't provide image previews and file sizes for binary files, especially image files.

In this PR, I implemented the parsing of binary files from `read_file_from_workspace` to display the size and image files in the fileDiff component.

## 🧢 Changes

- Add size display for binary files in the fileDiff component.
- Add image file preview in the fileDiff component.
- Implemented binary file parsing in `read_file_from_workspace`.

### Preview

https://github.com/user-attachments/assets/362548a6-6a6f-440f-9058-73a55e8d5e6f

## 🎫 Affected issues

Fixes: #2752 
Fixes: #3093
Fixes: #4957
Closes: #4852


crates/gitbutler-repo/src/commands.rs
```diff
@@ -60,12 +61,47 @@ impl RepoCommands for Project {
             let entry = tree.get_path(relative_path)?;
             let blob = ctx.repository().find_blob(entry.id())?;
```
Byron: I am aware that these semantics were present already, but I am confused that the lookup is in the tree first. My expectation would be the following order:

* worktree
* index
* `HEAD^{tree}`

Byron:
Hi @estib-vega, maybe this PR would be interesting for you to review as it spans frontend and backend, and particularly in the frontend I think you'd do a much better job than me.

crates/gitbutler-repo/src/commands.rs
Zamoca42: That’s why it seems that untracked binary files cannot be previewed.

Byron: Just now when working on #5111 I also touched the method, and a 'blame' later I found out that I think it's best to ask @ndom91 what should be done here. Thanks in advance for your help.

ndom91: Yeah I wrote the initial get repo content for a specific use case, it would need to be modified and/or built on top of for this. 

I love the motivation though! Happy to pick this up with you on Monday 👌

Zamoca42:
Thank you for the thoughtful work on #5111. Building on that, I was able to modify `read_file_from_workspace` to check `status_file` and successfully read untracked binary files.

### Preview

https://github.com/user-attachments/assets/f29bb6a3-0c3a-449b-9d53-ae71cf76996f

crates/gitbutler-repo/src/commands.rs
```diff
@@ -59,51 +74,61 @@ impl RepoCommands for Project {
             );
         }
 
+        let status = ctx.repository().status_file(relative_path)?;
+
+        if status.is_wt_new() {
```
Byron: GitButler may show worktree files, which may be 

* untracked
* modified
* deleted in the worktree
* deleted in the worktree and index

Semantically this boils down to trying to read, in order from top to bottom:

* worktree
* index
* `HEAD^{tree}

It may also show files that are already committed, which is when they will always be contained in trees. The UI will only display added, deleted or modified ones

* deleted files have to be fetched from the 'previous' tree
* added files are found in the 'current' tree
* modified files can also be found in the `current` tree

In this case, we'd not want to look at the worktree or the index at all, but the UI would probably have to decide which tree to get the relative path from.

The semantics of `read_file_from_workspace()` fit better to the `tree` version (the latter example), and could rather be `read_file_from_tree(tree_id, relative_path)`, while a new function could handle the former case.

CC @Caleb-T-Owens and @ndom91 .

Zamoca42: Thank you for the clear explanation. Based on my understanding, it seems like it would be better to separate the part where `read_file_from_workspace` reads `HEAD^{tree}` into a function like `read_file_from_tree(tree_id, relative_path)`.

Additionally, I will work on handling the logic for modified or deleted files in the worktree or `HEAD^{tree}` within `read_file_from_workspace`.

Zamoca42: Based on the review, I have applied a few changes:

- In `read_file_from_workspace`, I attempted to handle file lookups in the order of worktree, index, and `HEAD^{tree}`.
- Separated the logic for tree lookups in `read_file_from_workspace` into a new function `read_file_from_tree(tree_id, relative_path)`.
- Added conditions to handle the untracked, modified, and deleted states that GitButler displays.

### Modified

https://github.com/user-attachments/assets/4d6f1df7-766e-4807-9e58-9330dd8b4120

### Removed

https://github.com/user-attachments/assets/476d358d-6f00-447a-85e5-3745a4cd20b6


P.S.: In the UI worktree, the file states (delete, add) are clearly visible, but in the committed tree, only the modified state appears. This can be a bit confusing.

krlvi (approved):
lgtm. @Byron if you have no more feedback here i will merge this one

Byron:
Thanks for taking a look. I plan to review/fixup and will merge when done. Should get to it till lunch.

Byron:
Since I am very delayed here, let's merge, if there are fixes I'd like to make I will open a new PR for that.

PR #5090, opened 2024-10-10
feat: add global ctrl+z keyboard shortcut

ndom91:
## ☕️ Reasoning

- Add a `$history - 1` keyboard shortcut

## 🧢 Changes

- Update `Board.svelte` to runes syntax


ndom91:
> hmm, so the behavior of this may be surprising...

What do you mean exactly? Because of the reloading of the whole UI?

krlvi (approved):

ndom91:
Replaced through:

https://github.com/gitbutlerapp/gitbutler/pull/5094

PR #5094, opened 2024-10-10
fix: refactor `showHistoryView` and add `ctrl+z` shortcut alias

ndom91:
## ☕️ Reasoning

- Cleanup `showHistoryView` and add alias shortcut

## 🧢 Changes

- Refactor persisted `$showHistoryView` a bit
- add `ctrl + z` shortcut alias in addition to the pre-existing one, `ctrl + shift + H`

PR #5097, opened 2024-10-10
Extract persisted into a shared library

Caleb-T-Owens:


apps/desktop/src/lib/components/Board.svelte
```diff
@@ -65,14 +66,8 @@
...
 	const handleKeyDown = createKeybind({
-		's t a c k': () => {
+		's t a c k': async () => {
 			$stackingFeature = !$stackingFeature;
-		},
-		'$mod+Shift+H': () => {
-			$showHistoryView = !$showHistoryView;
-		},
-		'$mod+z': () => {
-			$showHistoryView = !$showHistoryView;
```
estib-vega: Is this intentional? I think then change being un-done was introduced by @ndom91 in here: https://github.com/gitbutlerapp/gitbutler/pull/5094/files#diff-3a92322cb1896e0e3dd44e9b66fe4c2536081848ab1755f9b30a7c47f61c3e42R68-R75

apps/desktop/src/lib/stack/StackHeader.svelte
```diff
@@ -122,7 +122,6 @@
 						<Button
 							bind:el={meatballButtonEl}
 							style="ghost"
-							size="tag"
```
estib-vega: Also these little changes, are they intentional quick fixes? Just want to double check

apps/desktop/src/lib/stack/StackHeader.svelte
```diff
@@ -205,7 +205,7 @@
 	.header__info-wrapper {
 		display: flex;
 		gap: 2px;
-		padding: 12px;
+		padding: 10px;
```
estib-vega: See comment above ☝️

apps/desktop/src/lib/stack/StackHeader.svelte
```diff
@@ -218,7 +218,7 @@
 	.button-group {
 		display: flex;
 		align-items: center;
-		gap: 6px;
+		gap: 10px;
```
estib-vega: See comment above ☝️

apps/desktop/src/lib/components/Board.svelte
ndom91: Yeah seems like you might be missing some latest changes from master

Caleb-T-Owens:
Interesting, I think vscode stores cached versions of files which can cause it to crap out when you rename too much stuff

estib-vega (approved):
Looks good!

PR #5104, opened 2024-10-11
fix: update stacking CommitCard tooltip

ndom91:
## ☕️ Reasoning

- Teach user about commitList icon colors / types

## 🧢 Changes

- Update tooltip content from commit sha / author name (both of which are available on the commit card right next to the tooltip) to the label of the commit type, i.e. what the color means
- Add a `package` npm script to the root `package.json` to `package` both the `packages/ui` and `packages/shared` packages.


estib-vega (approved):
MGMT
https://www.youtube.com/watch?v=MmZexg8sxyk

PR #5110, opened 2024-10-11
fix: History lazy loading

estib-vega:
Lazy loading the history side bar list items wouldn't paginate correctly.
There seems to have been two issues that caused this:
1. The intersection observer would was not initially set correctly
2. The "all loaded" flag was incorrectly set to true when clearing the history snapshots store. It should have been set to false.

Will fix the weird loading animation on another PR

### DIFF
--- a/apps/desktop/src/lib/shared/LazyloadContainer.svelte
+++ b/apps/desktop/src/lib/shared/LazyloadContainer.svelte
@@ -12,13 +12,14 @@
 
 	let { children, minTriggerCount, role, ontrigger, onkeydown }: Props = $props();
 
-	let lazyContainerEl: HTMLDivElement;
+	let lazyContainerEl: HTMLDivElement | undefined;
 
 	const mutuationObserver = new MutationObserver(attachIntersectionObserver);
 	$effect(() => {
 		if (lazyContainerEl) {
 			mutuationObserver.disconnect();
 			mutuationObserver.observe(lazyContainerEl, { childList: true });
+			attachIntersectionObserver();
 		}
 	});
 	const intersectionObserver = new IntersectionObserver((entries) => {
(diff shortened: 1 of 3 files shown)


apps/desktop/src/lib/shared/LazyloadContainer.svelte
```diff
@@ -37,10 +37,14 @@
...
 
-		intersectionObserver.observe(containerChildren[containerChildren.length - 1] as HTMLElement);
+		const lastChild = containerChildren[containerChildren.length - 1];
+		if (!lastChild) return;
+
+		intersectionObserver.observe(lastChild);
 	}
 
 	onMount(() => {
+		attachIntersectionObserver();
```
Caleb-T-Owens: Why is it required that we call `attachIntersectionObserver` here when we have the $effect() block on line 18 through 23 which should be starting the observation?

estib-vega: It seems that it doesn't always get called, weirdly. 

The history board was mounted and everything was setup, but it appears that the container didn't mutate and hence the intersection observer didn't start its observation

estib-vega: Maybe there's a better way of fixing this, though

Caleb-T-Owens: I see. `attachIntersectionObserver` expects `lazyContainerEl` to be present (the type signature of `lazyContainerEl` ought to be `HTMLDivElement | undefined`).

We could quite reasonably call `attachIntersectionObserver` after line 20.

estib-vega: That works too 👍 
Will update the PR

Caleb-T-Owens (approved):
Thanks for those changes, LGTM

PR #5111, opened 2024-10-11
Probably fix #5096

### LINKED ISSUE #5096: "Failed to read pull request template"

Byron:
When testing paths for prefix-matches it's important they are all normalized in the same fashion.

`canonicalize()` is very particular about canonicalizing Windows paths, which makes it easy for these paths to not be compatible to other absolute-looking paths.

The difficulty here is to get the right trade-off between performance and safety, e.g. we wouldn't want these canonicalized Windows paths to be used anywhere as they are very uncommon (and don't even work everywhere).

### Notes for the Reviewer

* I didn't actually validate this on Windows, but am quite confident that this will fix the issue.
* This PR touches `read_file_from_workspace()` which needs work, I just did the bare minimum here. More context is [here](https://github.com/gitbutlerapp/gitbutler/pull/5089#discussion_r1797379253).

Related to #5096 .

### DIFF
--- a/crates/gitbutler-project/src/controller.rs
+++ b/crates/gitbutler-project/src/controller.rs
@@ -72,7 +72,7 @@ impl Controller {
         let project = Project {
             id: ProjectId::generate(),
             title,
-            path: path.to_path_buf(),
+            path: gix::path::realpath(path)?,
             api: None,
             ..Default::default()
         };
--- a/crates/gitbutler-repo/src/commands.rs
+++ b/crates/gitbutler-repo/src/commands.rs
@@ -49,25 +49,24 @@ impl RepoCommands for Project {
 
     fn read_file_from_workspace(&self, relative_path: &Path) -> Result<String> {
         let ctx = CommandContext::open(self)?;
-        if self
-            .path
-            .join(relative_path)
-            .canonicalize()?
-            .as_path()
-            .starts_with(self.path.clone())
-        {
-            let tree = ctx.repository().head()?.peel_to_tree()?;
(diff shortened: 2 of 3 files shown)


ndom91:
So what was the underlying issue here? The prefix match didn't work on windows due to canonicalize?

Byron:
Yes, it's very likely. Note that `gix_path::*` is used as it does what `canonicalize()` does, but without butchering Windows paths, so its output is usable also for presentation to the user. The lack of this trait is the reason that `Project::path` doesn't use `canonicalize()`, which would have been an alternative solution.

Caleb-T-Owens (approved):
LGTM.

I'm still not in love with the name. IIRC there was some business logic reason as to why we wanted to get the committed version of the file over the uncommitted one, but that feels like something we can bike-shed another time

Byron:
Maybe the linked PR should rather create a new function that acts like it need it to act, while this one could be renamed to something more specific. If the case, the `canonicalize()/realpath()` check also won't be needed anymore.

PR #5125, opened 2024-10-14
fix: show diverged commitId in diverged state

ndom91:
## ☕️ Reasoning

- Diverged commit line node (dual icon) should show "Diverged" + the diverged remote commit id. The local commit Id is already on th ecommit card right next to the icon


estib-vega (approved):
LGTM

PR #5127, opened 2024-10-14
fix: change stacking header line alignment

ndom91:
## ☕️ Reasoning

- Align all commit lines better by using widht + margin instead of borders


PavelLaptev:
Thank you @ndom91
Looks good :-)

Just the "upstream" commits still shifted a bit

image

ndom91:
> Thank you @ndom91 Looks good :-)
> 
> Just the "upstream" commits still shifted a bit
> 

Good catch, updated!

PavelLaptev (approved):
Looks good :-) Thank you!

PR #5128, opened 2024-10-14
fix: divergent commit shorten to 7 char sha

ndom91:
## ☕️ Reasoning

- Shorten divergent sha to 7 chars

### DIFF
--- a/packages/ui/src/lib/commitLinesStacking/CommitNode.svelte
+++ b/packages/ui/src/lib/commitLinesStacking/CommitNode.svelte
@@ -29,7 +29,7 @@
 		</Tooltip>
 	{:else if type === 'localAndShadow'}
 		<div class="local-shadow-commit-dot">
-			<Tooltip text={commitNode.commit?.remoteCommitId ?? 'Diverged'}>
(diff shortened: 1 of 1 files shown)


Caleb-T-Owens (approved):
LGTM

packages/ui/src/lib/commitLinesStacking/CommitNode.svelte
```diff
@@ -29,7 +29,7 @@
 		</Tooltip>
 	{:else if type === 'localAndShadow'}
 		<div class="local-shadow-commit-dot">
-			<Tooltip text={commitNode.commit?.remoteCommitId ?? 'Diverged'}>
+			<Tooltip text={commitNode.commit?.remoteCommitId?.substring(0, 7) ?? 'Diverged'}>
```
Byron, line 32: It will be surprising to hear that this is a bug, *sometimes* :D.
In bigger repositories, a short hash like this may be ambiguous, and if this is presented to users they might not be able to easily use it.

If it should be done correctly, [`gix::Id::shorten()`](https://docs.rs/gix/latest/gix/struct.Id.html#method.shorten) must be used. As it might be expensive to call individually, one might consider to provide shortened versions of hashes for presentation in the UI by default.
Also CC @krlvi as this may be a general issue.

PR #5129, opened 2024-10-14
fix: dont allow changing series name once pushed

ndom91:
## ☕️ Reasoning

- Series/branches that have been pushed can no longer be renamed, therefore we shouldn't even allow the user to input here any more once thats happened


PavelLaptev (approved):
👍

PR #5136, opened 2024-10-14
Update how auth tokens are passed about

Caleb-T-Owens:


apps/desktop/src/lib/stores/user.ts
```diff
@@ -15,6 +15,19 @@ export type LoginToken = {
...
+
+	get token() {
+		return this.#token;
+	}
+
+	setToken(data: string | undefined) {
+		this.#token.set(data);
+	}
+}
+
```
Caleb-T-Owens: I've had to move the token into its own little store in the Desktop app because it needs to be used by both the HttpService and the UserService, while the UserService also depends on the HttpService

```mermaid
flowchart TD
  UserService --> HttpService
  HttpService --> TokenMemoryService
  UserService --> TokenMemoryService
```

apps/web/src/lib/auth/authService.ts
```diff
@@ -1,19 +1,18 @@
-import { get, writable } from 'svelte/store';
+import { persisted } from '@gitbutler/shared/persisted';
+import { type Readable } from 'svelte/store';
 
 export class AuthService {
-	token = writable<string | undefined>(undefined);
+	#token = persisted<string | undefined>(undefined, 'AuthService--token');
```
Caleb-T-Owens, line 5: As discussed, it's not ideal that we store the token in local storage, but it is probably the best option until we're able to move to a HttpOnly cookie with expiring tokens.

apps/desktop/src/lib/ai/service.ts
```diff
@@ -209,16 +212,19 @@ export class AIService {
...
+			// TODO(CTO): Once @estib has landed the new auth, it would be good to
+			// about a good way of checking whether the user is authenticated.
+			if (!get(this.tokenMemoryService.token)) {
 				return buildFailureFromAny(
 					"When using GitButler's API to summarize code, you must be logged in"
 				);
 			}
-			return ok(new ButlerAIClient(this.cloud, userToken, modelKind));
+
+			return ok(new ButlerAIClient(this.cloud, modelKind));
```
Caleb-T-Owens, line 227: It would be nice to have something on a shared UserService that we can call to determine whether the user is authenticated, or even what level of permissions the user has, but that didn't fit into the scope of this PR

estib-vega (approved):
LGTM

PR #5141, opened 2024-10-14
Stacking flow/ UI handles already pushed vbranches

krlvi:
Background:

In the case of (v)branches that have a remote reference (e.g. have already been pushed or have been applied from an existing remote branch), the stacking UI did not show the correct remote info (i.e. remote commits or Pull Request info)
This is of course to be expected since the new flow operates on an entierely different namespace. However this is also undesirable UX during the migration phase - chances are, users will have pull requests in flight when they update the app.

Solution:

Whenever a Stack is initialized, or if it has already been initialized then whenever it is loaded - set the reference name of the stack to match that of the legacy virtual branch.

### DIFF
--- a/crates/gitbutler-branch-actions/src/virtual.rs
+++ b/crates/gitbutler-branch-actions/src/virtual.rs
@@ -292,7 +292,7 @@ pub fn list_virtual_branches_cached(
 
     let branches_span =
         tracing::debug_span!("handle branches", num_branches = status.branches.len()).entered();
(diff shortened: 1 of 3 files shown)


Caleb-T-Owens (approved):
I'm not quite familiar enough with the code to visualise the changes, but all looks reasonable and has tests

krlvi:
> I'm not quite familiar enough with the code to visualise the changes, but all looks reasonable and has tests

british for lgtm :D

PR #5143, opened 2024-10-15
Revert dialog modal implementation

PavelLaptev:
## ☕️ Reasoning

The native `<dialog>` element has several critical bugs, making it less reliable. This update reverts to a more stable implementation.

## 🧢 Changes

- Reverted `<dialog>` to `<div>`
- Added a custom backdrop layer using CSS variables
- Added `Esc` key support to close the modal
- Autofocus now targets the first focusable element in the modal
- Implemented a focus trap hook for improved accessibility
- Added an optional `focusOnElement` option in the `focusTrap` hook to allow specifying the initial focus element
- Modified `clickoutside` behavior to only trigger when clicking on the backdrop, preventing issues with components like `Select`

### DIFF
--- a/packages/ui/src/lib/Modal.svelte
+++ b/packages/ui/src/lib/Modal.svelte
@@ -34,7 +34,6 @@
 
 	let open = $state(false);
 	let item = $state<any>();
-	let dialogElement = $state<HTMLDivElement>();
 	let isClosing = $state(false);
 
 	function handleKeyDown(event: KeyboardEvent) {
@@ -51,6 +50,7 @@
 	export function show(newItem?: any) {
 		item = newItem;
 		open = true;
+
 		window.addEventListener('keydown', handleKeyDown);
 	}
 
@@ -72,12 +72,7 @@
 </script>
 
 {#if open}
-	<div
-		use:portal={'body'}
-		class="modal-container {isClosing ? 'closing' : 'open'}"
-		class:open
-		bind:this={dialogElement}
-	>
+	<div use:portal={'body'} class="modal-container {isClosing ? 'closing' : 'open'}" class:open>
 		<div
 			use:focusTrap
 			use:clickOutside={{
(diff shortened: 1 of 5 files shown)


packages/ui/src/lib/Modal.svelte
```diff
@@ -53,110 +70,106 @@
 	};
 </script>
 
-<!-- svelte-ignore a11y_no_noninteractive_tabindex -->
-<!-- svelte-ignore a11y_no_noninteractive_element_interactions -->
```
estib-vega, line 57: I think we might still need this. Haven't checked out the branch locally, but Svelte should have complained about having a key listener on a div. 

Weird that the check passed, though 🤔

packages/ui/src/lib/Modal.svelte
```diff
@@ -53,110 +70,106 @@
...
-				</div>
-			{/if}
+{#if open}
+	<div
+		role="presentation"
+		use:portal={'body'}
+		class="modal-container {isClosing ? 'closing' : 'open'}"
+		class:open
+		onclick={(e) => {
+			console.log(e.target);
```
estib-vega: Missed this console log :)

packages/ui/src/lib/Modal.svelte
```diff
@@ -53,110 +70,106 @@
...
+			e.stopPropagation();
+
+			if (e.target === e.currentTarget) {
+				close();
+			}
 
-			<div class="modal__body custom-scrollbar text-13 text-body" class:no-padding={noPadding}>
-				{@render children(item, close)}
-			</div>
+			// close();
```
estib-vega: Missed this comment :)

packages/ui/src/lib/Modal.svelte
PavelLaptev: It depends on the element's role. There's no need to add an ignore comment for elements with the "presentation" role. In this case, it makes sense because the backdrop element is interactive, but it's neither a button nor an input.

estib-vega: I see, makes sense 👍

estib-vega (approved):
Looks good!

ndom91 (approved):
LGTM, finally esc works again too :joy:

PR #5145, opened 2024-10-15
Fix button type for modals

PavelLaptev:
**PR Description:**

In modals any button was a submit button, because it was wrapped into `<form>` element


ndom91 (approved):
LGTM!

PR #5146, opened 2024-10-15
fix: Move any commit from branch to branch 

estib-vega:
Move any commit won't duplicate the commit in both virtual branches.
Also:
- Also add tests to catch this


krlvi (approved):
nice!

PR #5152, opened 2024-10-15
fix: show only stack push btn when necessary

ndom91:
## ☕️ Reasoning

- Don't always show "Push" / "Force Push" btn at bottom of stack


estib-vega (approved):
Looks good to me 👍

PR #5156, opened 2024-10-15
fix: add additional case for when to show push/force push btn

ndom91:
## ☕️ Reasoning

- Add the case when at least one of your commits in the stack is a diverged (`localAndShadow`)-type commit.


Caleb-T-Owens (approved):
LGTM

PR #5162, opened 2024-10-16
Implement 'integrate upstream changes' for the stacked flow

krlvi:
Since now the commits being integrated are not necessarely going to the stack head but to the series head which may be in the middle of the stack, this action needed to work differently

### DIFF
--- a/crates/gitbutler-branch-actions/src/branch_upstream_integration.rs
+++ b/crates/gitbutler-branch-actions/src/branch_upstream_integration.rs
@@ -348,6 +348,75 @@ mod test {
                     .unwrap(),
                 vec![remote_y.id(), remote_x.id(), local_b.id(), local_a.id()],
             );
+
+            // Also ensure the series implementation does the same thing
+            let (BranchHeadAndTree { head, tree: _tree }, _) = ctx
+                .inner_integrate_upstream_commits_for_series(local_b.id())
+                .unwrap();
+            assert_eq!(
+                test_repository
+                    .repository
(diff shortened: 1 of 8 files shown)


crates/gitbutler-branch-actions/src/branch_upstream_integration.rs
```diff
@@ -15,6 +15,108 @@ use crate::{
...
+                }
+            } else {
+                // If merge is preferred, create a merge commit between the remote and the series head and add that
+                let series_head_commit = commit_by_oid_or_change_id(
+                    &series.head.target,
+                    ctx,
+                    branch.head(),
+                    upstream_to_local_merge_base,
+                )?;
+                let merge_commit = gitbutler_merge_commits(
```
krlvi: now that i think about it - i shouldnt be doing this at all. i think i have an idea how to approach this.

In a nutshell: if this is a stack of one (or the top of the stack series), allow making the merge commit and update the head (the head of the series will equal the head of the stack in this case).

Otherwise refuse to do this and tell the user that if they are doing stacking rebasing is a must

krlvi:
@Caleb-T-Owens i reimplemented the functionality as a method on `IntegrateUpstreamContext` so that it's easier to assert the old and new implementations do the same thing.
Also i redid the way it did the rebasing - instead of iterating commits, i did a two step rebase - first up to the series head to add the new commits and then a second one to add the rest

Caleb-T-Owens (approved):
Apart from the unfortunate spelling mistakes, looks good to me :+1:

PR #5165, opened 2024-10-16
Some improvements on top of #5089

Byron:
Follow-up of #5089

The PR was merged earlier to not hold it after a passed review, yet I personally feel strongly enough about the backend portions to give it another round of polish.

### Tasks

* [x] document the current behaviour of the backend methods and refactor
* ~~turn the 'variable type' `FileInfo` into an enum to be type-safe in the frontend as well~~ - probably that would make the frontend code more complicated, at least I wouldn't want to do it.

### Notes for the Reviewer

* I think the frontend should always use relative path, and never pass or store absolute paths. Even though the backend is defensive towards absolute paths, I can imagine issues to show up once repositories are renamed on disk.
* I did test it with a PNG file and went through all the states with it, i.e. in worktree, deleted, and deleted from index, and it all worked as one would expect.

### DIFF
--- a/apps/desktop/src/lib/file/FileDiff.svelte
+++ b/apps/desktop/src/lib/file/FileDiff.svelte
@@ -15,7 +15,6 @@
 		name?: string;
 		mimeType?: string;
 		size?: number;
-		status: 'normal' | 'deleted';
 	}
 
 	interface Props {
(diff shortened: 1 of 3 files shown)


krlvi (approved):
thanks! lgtm

Issue #5113, opened 2024-10-12
"There is no default target"


navtoj:
### Version

0.13.0 (20241010.135849)

### Operating System

Mac OS X

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

![Screenshot 2024-10-11 at 7 47 09 PM](https://github.com/user-attachments/assets/afedc3ce-16e2-4ad9-8855-d327b282b142)

### How to reproduce

1. Go to `Clone a repository` page
2. Enter SSH url of Github repo (e.g. `git@github.com:navtoj/NotchBar.git`)
3. Press `Clone`

### Expected behavior

This error should not show up.

### Relevant source code

https://github.com/gitbutlerapp/gitbutler/blob/44a6125c3cec5ed312506d1be9a183791e8eb4b6/crates/gitbutler-stack/src/state.rs#L87-L95

#### ui-logs.log
```shell
[2024-10-12][02:57:48][ERROR][Rt@tauri://localhost/_app/immutable/entry/app.OhJdX_6z.js:2:3069] 2705b3b08f1847d9b51155b4947142d9: There is no default target

a@tauri://localhost/_app/immutable/chunks/Icon.BAACgJ7C.js:1:940
fromError@tauri://localhost/_app/immutable/chunks/Icon.BAACgJ7C.js:1:1109
@tauri://localhost/_app/immutable/chunks/Icon.BAACgJ7C.js:1:1319

[2024-10-12][02:57:48][ERROR][Rt@tauri://localhost/_app/immutable/entry/app.OhJdX_6z.js:2:3069] 2705b3b08f1847d9b51155b4947142d9: There is no default target

a@tauri://localhost/_app/immutable/chunks/Icon.BAACgJ7C.js:1:940
fromError@tauri://localhost/_app/immutable/chunks/Icon.BAACgJ7C.js:1:1109
@tauri://localhost/_app/immutable/chunks/Icon.BAACgJ7C.js:1:1319

[2024-10-12][02:57:48][ERROR][Rt@tauri://localhost/_app/immutable/entry/app.OhJdX_6z.js:2:3069] a0e1028c22374985b6da2027b12ab480: There is no default target

a@tauri://localhost/_app/immutable/chunks/Icon.BAACgJ7C.js:1:940
fromError@tauri://localhost/_app/immutable/chunks/Icon.BAACgJ7C.js:1:1109
[...]

PR #5167, opened 2024-10-16
fix: Swallow the default target not found

estib-vega:
Don't surface the default target not found. This already is caught at the top level and the user is taken to the 'target not found' page.

### Issue
As seen in https://github.com/gitbutlerapp/gitbutler/issues/5113

### DIFF
--- a/apps/desktop/src/lib/branches/branchListing.ts
+++ b/apps/desktop/src/lib/branches/branchListing.ts
@@ -1,7 +1,7 @@
 // Class transformers will bust a gut if this isn't imported first
 import 'reflect-metadata';
 
-import { invoke } from '$lib/backend/ipc';
+import { Code, invoke } from '$lib/backend/ipc';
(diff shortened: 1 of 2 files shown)


ndom91 (approved):
This part LGTM!

Where is the logic that should bump a user out to the project setup page though? :thinking:

estib-vega:
> This part LGTM!
> 
> Where is the logic that should bump a user out to the project setup page though? 🤔

@ndom91 That'd be here: https://github.com/gitbutlerapp/gitbutler/blob/eaa785e3bb4968a5c1a3b6971a3d83de399c814e/apps/desktop/src/routes/%5BprojectId%5D/%2Blayout.svelte#L187-L188

It looks a bit brittle, and we probably shouldn't keep just swallowing the errors, imo.
Maybe something to take into the conversation of v3/Svelte-Pattern-Definition

PR #5170, opened 2024-10-16
fix: Get the right PR source branch

estib-vega:
Proposed PR Description:

## ☕️ Reasoning

This PR fixes an issue where the PR source branch was being set incorrectly.

## 🧢 Changes

The commit in this PR ensures that the PR source branch is set correctly, instead of setting the base branch (e.g. main or master) as the source branch.


krlvi (approved):

Caleb-T-Owens (approved):
Looks good to me

Issue #4951, opened 2024-09-19
Cherry Pick


LovesAsuna:
### Version

0.12.25

### Operating System

Mac OS X

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

I want to cherry pick a commit from a virtual branch to another. Because my changes are based on part of one of the virtual branches. This virtual branch has not yet been integrated into the target branch, so I'd like to do a cherry pick first (not a full commit, but maybe just a part of a commit or a file).

Byron:
It should be possible to drag entire commits from one lane to another, or hunks from one lane to another as well. Is this something you tried, or is there an issue I am missing?

If so, maybe upload a video so it's clear what is attempted.
Thanks in advance for your help.

LovesAsuna:
![image](https://github.com/user-attachments/assets/cd31de90-645f-4127-9869-2c89675d33e7)

Byron:
`Invalid args: `branch` for command `create_virtual_branch`: ownership ranges cannot be empty` is the error message displayed, when trying to move a single file out of a commit into the `New branch` (empty) lane.

It's easily reproducible now, thank you!

On stderr, one will see this right after (but no 'typical' error):

```
[2024-09-21][07:22:32][ERROR][Rt@tauri://localhost/_app/immutable/entry/app.CN3g50-3.js:2:3027] 9d9abf00c0a3451684adb9a104d21eb4: lifecycle_outside_component

xn@tauri://localhost/_app/immutable/chunks/index.BRFB7DrO.js:1:7216
g@tauri://localhost/_app/immutable/chunks/index-client.DHPHoFon.js:1:521
@tauri://localhost/_app/immutable/chunks/CommitCard.Dvwviucn.js:287:6357




2024-09-21 09:22:40.042 GitButler[98579:4957218] _TIPropertyValueIsValid called with 16 on nil context!
2024-09-21 09:22:40.042 GitButler[98579:4957218] imkxpc_getApplicationProperty:reply: called with incorrect property value 16, bailing.
2024-09-21 09:22:40.042 GitButler[98579:4957218] Text input context does not respond to _valueForTIProperty:
[2024-09-21][07:22:41][ERROR][Rt@tauri://localhost/_app/immutable/entry/app.CN3g50-3.js:2:3027] 1f76d1f4220444eb8190005661803647: lifecycle_outside_component

xn@tauri://localhost/_app/immutable/chunks/index.BRFB7DrO.js:1:7216
[...]

PR #5177, opened 2024-10-16
fix: New branch dropzone only accepts uncommited files

estib-vega:
When creating a brand new branch, don't let committed files or hunks create new branches

### Issues
As seen in https://github.com/gitbutlerapp/gitbutler/issues/4951

### DIFF
--- a/apps/desktop/src/lib/branch/BranchDropzone.svelte
+++ b/apps/desktop/src/lib/branch/BranchDropzone.svelte
@@ -15,8 +15,9 @@
 	const branchController = getContext(BranchController);
 
 	function accepts(data: any) {
-		if (data instanceof DraggableFile) return !data.files.some((f) => f.locked);
-		if (data instanceof DraggableHunk) return !data.hunk.locked;
+		if (data instanceof DraggableFile)
+			return !(data.isCommitted || data.files.some((f) => f.locked));
+		if (data instanceof DraggableHunk) return !(data.isCommitted || data.hunk.locked);
 		return false;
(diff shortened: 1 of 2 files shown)


apps/desktop/src/lib/branch/BranchDropzone.svelte
```diff
@@ -15,8 +15,9 @@
 	const branchController = getContext(BranchController);
 
 	function accepts(data: any) {
-		if (data instanceof DraggableFile) return !data.files.some((f) => f.locked);
-		if (data instanceof DraggableHunk) return !data.hunk.locked;
+		if (data instanceof DraggableFile)
+			return !data.isCommitted && !data.files.some((f) => f.locked);
+		if (data instanceof DraggableHunk) return !data.isCommitted && !data.hunk.locked;
```
Caleb-T-Owens: Can we de'morgan this and say

```ts
		if (data instanceof DraggableFile)
			return !(data.isCommitted || data.files.some((f) => f.locked));
		if (data instanceof DraggableHunk) return !(data.isCommitted || data.hunk.locked);
```

Caleb-T-Owens (approved):
Other than a very small nitpick which you can feel free to ignore, Looks good to me. Thanks for closing that one up.

apps/desktop/src/lib/branch/BranchDropzone.svelte
estib-vega: Sure thing!

PR #5184, opened 2024-10-17
fix: update copy for feature flag for enabling multi-series stacking support

ndom91:
## ☕️ Reasoning

- Some users have already come across this, just want to ensure that expectations are correctly set when testing this feature.


estib-vega (approved):
Sounds reasonable

PR #5185, opened 2024-10-17
Patch stack listing and creation

Caleb-T-Owens:


apps/desktop/src/lib/settings/userPreferences/DetailsForm.svelte
```diff
@@ -54,18 +43,28 @@
...
 			console.error(`Failed to update project sync status: ${error}`);
 			toasts.error('Failed to update project sync status');
 		}
 	}
+
+	// This is some janky bullshit, but it works well enough for now
+	onMount(async () => {
+		if (!project?.api) return;
+		if (!$user) return;
+		console.log(project);
```
estib-vega: Missed a console log :)

estib-vega (approved):
Just seems to be an extra console log in there, but looks ok 👍

PR #5189, opened 2024-10-17
feat: hide contextMenu items that cant be activated and convert ContextMenuItem to runes

ndom91:
## ☕️ Reasoning

- Hide series context menu items that would normally be disabled

## 🧢 Changes

- Convert ContextMenuItem to runes
- Update all uses of `ContextMenuItem`


krlvi:
Maybe copy could be simpler, eg. "Not undateable after pushing" - what do you think? As a native english speaker

PavelLaptev:
We shouldn't display "rename" at all if it was already pushed

PavelLaptev:
another thing. No one doing tooltips in context menus. So find another way

PavelLaptev:
@krlvi cc

ndom91:
@PavelLaptev yeah good call, so I've just hidden the options that used to be disabled. So now invalid options aren't shown when they can't be clicked. What do you think?

PavelLaptev (approved):
thanks! Looks good :-)

PR #5197, opened 2024-10-18
New crate for calculating hunk dependencies

mtsgrd:
Adds new crate with code for calculating dependencies between workspace changes and workspace commits. What we ultimately want to understand is: given an uncommitted change, do the old line numbers intersect with anything committed in the workspace?

The problem we have to overcome is that we the workspace changes are produced by diffing the working directory against the workspace commit. It means changes from one stack can offset line numbers in changes from a different stack. The most intuitive way of checking if they touch the same lines is to use regular git blame, but it suffers from two problems, 1) speed and 2) lack of --reverse flag in git2. The latter means we can't detect intersections with deleted lines.

If we don't calculate these dependencies correctly it means a user might be able to move a hunk into a stack where it cannot be committed. So the solution here is that we build up the same information we would get from blame by adding diffs together.

Follow-up PR for integration this code: https://github.com/gitbutlerapp/gitbutler/pull/5252

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2181,6 +2181,7 @@ dependencies = [
  "gitbutler-error",
  "gitbutler-fs",
  "gitbutler-git",
+ "gitbutler-hunk-dependency",
  "gitbutler-id",
  "gitbutler-operating-modes",
  "gitbutler-oplog",
@@ -2374,6 +2375,25 @@ dependencies = [
  "tokio",
  "uuid",
  "windows 0.58.0",
+]
+
+[[package]]
+name = "gitbutler-hunk-dependency"
+version = "0.0.0"
+dependencies = [
+ "anyhow",
+ "bstr",
+ "git2",
+ "gitbutler-diff",
+ "gitbutler-id",
+ "gitbutler-reference",
+ "gitbutler-serde",
+ "gitbutler-stack",
+ "gix",
+ "itertools 0.13.0",
+ "serde",
+ "tokio",
+ "uuid",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ members = [
     "crates/gitbutler-oxidize",
     "crates/gitbutler-stack",
     "crates/gitbutler-patch-reference",
+    "crates/gitbutler-hunk-dependency",
 ]
 resolver = "2"
 
@@ -93,6 +94,7 @@ gitbutler-oxidize = { path = "crates/gitbutler-oxidize" }
 gitbutler-stack = { path = "crates/gitbutler-stack" }
 gitbutler-patch-reference = { path = "crates/gitbutler-patch-reference" }
 gitbutler-forge = { path = "crates/gitbutler-forge" }
+gitbutler-hunk-dependency = { path = "crates/gitbutler-hunk-dependency" }
 
 [profile.release]
 codegen-units = 1 # Compile crates one after another so the compiler can optimize better
--- a/crates/gitbutler-branch-actions/src/branch.rs
+++ b/crates/gitbutler-branch-actions/src/branch.rs
@@ -150,7 +150,7 @@ fn matches_all(branch: &BranchListing, filter: BranchListingFilter) -> bool {
     if let Some(local) = filter.local {
         conditions.push((branch.has_local || branch.virtual_branch.is_some()) && local);
     }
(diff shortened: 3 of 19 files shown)


krlvi:
@mtsgrd i see that this is still work in progress - one quick suggestions: it would be a good idea to make your internal members/methods/functions `pub(crate)`. That way it is both easier to reason how the library is meant to be used, and also you are not leaking your internals

crates/gitbutler-hunk-dependency/src/builder.rs
```diff
@@ -0,0 +1,128 @@
...
+/// might be able to move a hunk into a stack where it cannot be committed.
+///
+/// So the solution here is that we build up the same information we would
+/// get from blame by adding diffs together.
+#[derive(Debug, Default, PartialEq, Clone)]
+pub struct HunkDependencyBuilder {
+    stacks: HashMap<StackId, DependencyStack>,
+}
+
+impl HunkDependencyBuilder {
```
krlvi: Is it a builder really? A builder would typically have methods that return the builder itself, so that they can be stringed together and then a build() that produces the thing that it builds

mtsgrd:
@krlvi yes, work in progress. Also, thanks, I'll do that. If you have more of that kind of feedback then please keep it coming. :)

crates/gitbutler-hunk-dependency/src/builder.rs
mtsgrd: You're right. I wonder what could be a better name.. 🤔

Byron:
I do not envy you for this task, but also think that it's a great idea to solve this issue for good, pinning it down with as many tests as it takes.

Since blame is mentioned, I thought I'd CC @cruessler who is working on a first version of [`gix blame`](https://github.com/GitoxideLabs/gitoxide/pull/1453). Who knows, maybe there is some cross-pollination.

krlvi:
> @krlvi yes, work in progress. Also, thanks, I'll do that. If you have more of that kind of feedback then please keep it coming. :)

Cool i can take a look and leave some notes. What is the entry point of the library as far as usage goes?

krlvi:
Since this code is still in an early stage, instead of commenting on individual things, i would like to encourage you to consider this from an API design standpoint. 

For example, instead of objects and builders, think about
- What result do you want to provide to the caller (what is the format of that result that makes sense outside of the internals of this implementation.
- What input do you need, at the highest level, to produce that result

Currently it appears that the caller would need to have a mutable `HunkDependencyBuilder` to which the user calls `.add`. It is not clear, how do i get the result?  Which part is the result? (is it `get_path`?). 

Instead a mutable struct you can consider getting the full structure as an arg. If you really want to use a mutable struct, perhaps the `.add` function can take a typed struct as an argument, so that you can document what is supposed to come in.

I hope this is helpful

crates/gitbutler-hunk-dependency/src/diff.rs
```diff
@@ -0,0 +1,76 @@
+use anyhow::{anyhow, Context};
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct Diff {
```
krlvi: Since i expect this to be an exported struct (saw it on the .add method) - we should probably name it something like DiffRange (since it does not include the diff itself)

crates/gitbutler-hunk-dependency/src/diff.rs
```diff
@@ -0,0 +1,76 @@
...
+    S: AsRef<str>,
+{
+    iter.take_while(|line| {
+        let line_ref = line.as_ref(); // Convert to &str
+        !line_ref.starts_with('-') && !line_ref.starts_with('+')
+    })
+    .fold(0i32, |acc, _| acc + 1)
+}
+
+impl TryFrom<String> for Diff {
```
krlvi: I believe you need to create these from BString - the reason being is that not all diffs are guaranteed to be utf8 serializable. For the UI we do create "diff for fronted" which is a utf8 (String) but those are lossy and are only used for display. Otherwise this may cause subtle / difficult to debug issues for some files and not others

mtsgrd:
> Instead a mutable struct you can consider getting the full structure as an arg.

Sounds obvious now that you say it, thanks.

mtsgrd:
Closing since this was a subset of https://github.com/gitbutlerapp/gitbutler/pull/5252.

PR #5203, opened 2024-10-18
Templated branch names - prefix with user initials

krlvi:


ndom91:
Works here! However, do we really just want to use the first initial of their username? Anthing against using the full username?

krlvi:
> Works here! However, do we really just want to use the first initial of their username? Anything against using the full username?

it's splitting by whitespace... so depends on how the name is configured. I guess if we were gonna keep this logic for a while... it should be configurable.

ndom91:
> > Works here! However, do we really just want to use the first initial of their username? Anything against using the full username?
> 
> it's splitting by whitespace... so depends on how the name is configured. I guess if we were gonna keep this logic for a while... it should be configurable.

Okay gotcha. Maybe we take the first two characters then though? Most people have usernames that are just 1 long "word", right, so they'll have just 1 character prefixes like I did for "ndom91". So that only gives us 26 different prefixes, meaning you could still relatively easily run into collisions.

crates/gitbutler-stack-api/src/stack_ext.rs
```diff
@@ -228,12 +228,29 @@ impl StackExt for Stack {
...
+
+        let (author, _committer) = ctx.repository().signatures()?;
+        // Get the author initials, splitting by any whitespace
+        let initials = author
+            .name()
+            .map(|s| {
+                format!(
+                    "{}-",
+                    s.split_whitespace()
+                        .map(|word| word.chars().next().unwrap_or(' '))
```
Byron, line 240: I don't know what a good solution to this would be, but this code is bound to break for everything that Unicode can do. Thinking of Chinese names, these can theoretically be used verbatim as to use, they are initials already :).

My intuition here is to see if the name is mostly latin characters maybe aftre decomposing the unicode to split accents from letters. If so, generate initials. Otherwise, see how many grapheme clusters the name is as it might already be just a couple, using Chinese as example.
However, I also don't know many things, like what happens with right-to-left languages like Arabic and if initials make any sense there.

Probably making this configurable is the easiest way to deal with it so people whose names get butchered can control it better.

krlvi: good point. ideally, we dont wanna have any generated names, but the current design of the app kinda forces us to.  Is there a good way to detect if it's not latin and just leave it blank in that case? I was thinking that `normalize_branch_name` further down takes care of that

Byron: I think the easiest way to do this is [`is_ascii_alphabetic()`](https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_alphabetic), which probably is false for things like `À` which might be precomposed (i.e. one codepoint). So first, one would have to [decompose](https://github.com/GitoxideLabs/gitoxide/blob/9e3acde9c98537c7c8ee58f632ce21fcca5b066d/gix-utils/src/str.rs#L20) the string, this separates the `A` and the tick on top, which makes the `A` ascii-uppercase.

I'd probably do it so that if the first decomposed character is latin, I'd use the first [grapheme](https://docs.rs/unicode-segmentation/latest/unicode_segmentation/struct.Graphemes.html) of each word. That way, one would definitely work correctly both on decomposed or precomposed strings with accents even for the latin alphabet.

PR #5206, opened 2024-10-18
Styles fix: new branch modal

PavelLaptev:
## ☕️ Reasoning

just a few small things

before:
image

after:
image


estib-vega (approved):
@PavelLaptev Looks good!
I noticed I missed an extra single-quote in line 60 after the `slugifiedRefName`. Could you take it out too?
If not I can take care of it after

```ts
helperText={generatedNameDiverges ? `Will be created as '${slugifiedRefName}''` : undefined}
```

Issue #4898, opened 2024-09-12
Let user choose between multiple PR templates on demand


mtsgrd:
We already support templates when creating PRs, but there is a request to support choosing between several templates just before creating the PR instead of choosing it in the settings.

https://discord.com/channels/1060193121130000425/1060193121666863156/1283260543158849618

krlvi:
This was implemented in https://github.com/gitbutlerapp/gitbutler/pull/5220 and shipped in [0.13.3](https://github.com/gitbutlerapp/gitbutler/releases/tag/release/0.13.3). The PR creation modal now lets you turn templates on and choose one from the dropdown, so I’m closing this as completed.

PR #5220, opened 2024-10-18
PR details modal: Allow user to choose the PR template

estib-vega:
## ☕️ Reasoning

Allow users to choose their preferred PR template for each individual pull request.

## 🧢 Changes

- Added a feature that enables users to select a PR template from a dropdown menu in the PR details modal.
- Improved the test coverage for the forge review crate.

### Issues
As seen in https://github.com/gitbutlerapp/gitbutler/issues/4898

## Visuals

Screenshot 2024-10-21 at 11 37 48
Screenshot 2024-10-21 at 11 37 56


PavelLaptev:
@estib-vega if it's UI related, please tag me :-)

PavelLaptev:
@estib-vega thanks for tagging me :-)
Does this mean we should remove this option from the settings? Otherwise they might interfere. 

image

estib-vega:
> @estib-vega thanks for tagging me :-) 
> 
@PavelLaptev Sure thing! :)

> Does this mean we should remove this option from the settings? Otherwise they might interfere.

We don't have to, but we can. How it would work as I implemented it, the user can pick what the default behavior is in the project settings, and then optionally for each PR use (or not user) whatever template they want.

We can make it simple and just have the option in the PR details modal, but I'm not sure if it could be "less annoying" to have the option to set a default and somehow collapse the template selector in the PR details modal

PavelLaptev:
As the first option I thought it could work like this:
- open PR modal
- toggle on "use template". The app should remember this state
- if there is only one template it will be preselected and the selector will be disabled/locked. We preselect an option and block it to indicate that there are no options available except for this one.
- if there are more options, the first option will be preselected. We preselect an option because a user agreed to use templates already clicking on the toggle. No need to leave the select blank in my view. But if a user changes the option to something else, the app should remember this and next PR creation show this as a preselected option.

---

We can also separate this feature into two places. Toggle it in the settings and select in the modal. But I think we can try to use it just in the Modal in a one place.

PavelLaptev (approved):
Thanks @estib-vega! Looks good

PR #5232, opened 2024-10-19
fixes a bug in handling the remote name

krlvi:
Seems like projects that were created long time ago do not have the push_remote_name field set up. Add correct handling for all those situations


krlvi:
thanks @stack72 for helping us catch this

Issue #5229, opened 2024-10-19
Tauri image error popup


ukarlsson:
### Version

0.13.2

### Operating System

macOS

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

This popup comes up all the time for me since recent update.

It says

```
f6d45a630a194830bbddb22ac1d72f91: The reference did not exist a@tauri://localhost/_app/immutable/chunks/Icon.FmaqJgaA.js:1:940 fromError@tauri://localhost/_app/immutable/chunks/Icon.FmaqJgaA.js:1:1109 @tauri://localhost/_app/immutable/chunks/Icon.FmaqJgaA.js:1:1319
```

<img width="616" alt="Screenshot 2024-10-19 at 22 25 35" src="https://github.com/user-attachments/assets/0433ec04-0054-42f6-8d2a-fcbcc2f488f1">

Byron:
Thanks a lot for reporting!

It's probably something about how the UI intercepts backend errors - previously it must have (maybe unintentionally) ignored some of the errors that occurred, while now it will show all of them.

This error in particular is well known and seems to happen when the sidebar (with branch information) refreshes. It comes from `gitoxide` and I think it would help if the reference it tried to fetch was mentioned.

ukarlsson:
I still get these popups all the time but they are slightly different now
<img width="941" alt="Screenshot 2024-10-22 at 12 21 44" src="https://github.com/user-attachments/assets/a2b8b37e-79b1-4e53-a9bc-d778b8b3a20d">

anirudhsama:
> Thanks a lot for reporting!
> 
> It's probably something about how the UI intercepts backend errors - previously it must have (maybe unintentionally) ignored some of the errors that occurred, while now it will show all of them.
> 
> This error in particular is well known and seems to happen when the sidebar (with branch information) refreshes. It comes from `gitoxide` and I think it would help if the reference it tried to fetch was mentioned.

For me, it tries to fetch the same branch multiple times (main in my case). See the recording: https://cln.sh/W0Wdllx7

Byron:
That's interesting! I wonder what's happening there that `master` doesn't exist, at least in that very moment.
Having looked at the code, I'd think this error is genuine and not a false positive.

[...]

PR #5233, opened 2024-10-20
mention ref-name when reference isn't found

Byron:
Now it will include the name of the reference that wasn't found, which should help with debugging.

Related to #5229 and various others.


krlvi (approved):

ukarlsson:
Screenshot 2024-10-22 at 12 21 44

This is the error I get now, many thanks

PR #5236, opened 2024-10-20
Template branch names - ensure initials generation handles utf8 chars correctly

krlvi:
Refactors the branch generation by moving it to a function and ensures that when generating initials, accents and other special characters are handled correctly. Additionally, the PR fixes an issue where the branch name needs to be in ASCII format.

### DIFF
--- a/crates/gitbutler-stack-api/Cargo.toml
+++ b/crates/gitbutler-stack-api/Cargo.toml
@@ -9,6 +9,7 @@ publish = false
 anyhow = "1.0.86"
 itertools = "0.13"
 rand = "0.8.5"
+gix-utils = "0.1.12"
 serde = { workspace = true, features = ["std"] }
 git2.workspace = true
 gix.workspace = true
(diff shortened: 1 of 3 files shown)


krlvi:
@Byron addressing this comment here https://github.com/gitbutlerapp/gitbutler/pull/5203#discussion_r1807206382 with this PR

crates/gitbutler-stack-api/src/stack_ext.rs
```diff
@@ -17,6 +17,7 @@ use gitbutler_repo::RepositoryExt;
 use gitbutler_stack::Stack;
 use gitbutler_stack::VirtualBranchesHandle;
 use gix::validate::reference::name_partial;
+use gix_utils::str::decompose;
```
Byron, line 20: This should be `gix::utils::str::decompose` - I validated that it is exported, particularly since [`gix` was updated](https://github.com/gitbutlerapp/gitbutler/pull/5233).

crates/gitbutler-stack-api/Cargo.toml
```diff
@@ -9,6 +9,7 @@ publish = false
 anyhow = "1.0.86"
 itertools = "0.13"
 rand = "0.8.5"
+gix-utils = "0.1.12"
```
Byron, line 12: This can be removed (see comment below).

Byron (commented):
Thanks a lot, this looks great, and I am glad `gix-utils` could be useful!
I did leave a comment about the dependency, which is probably worth fixing.

Issue #5235, opened 2024-10-20
GithubIntegration error (it did actually integrate though)


benjaminmodayil:
### Version

0.13.2

### Operating System

macOS

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

Appears to be a false positive error when integrating Github PR on a local repo after a fresh install (possibly optional). It did integrate via the UI ✅s, but reported a Tauri error related to the GithubIntegration in the bottom right.

<img width="1123" alt="CleanShot 2024-10-20 at 09 05 36@2x" src="https://github.com/user-attachments/assets/2078c3e8-c988-4034-9a4d-2a90e264b3aa">

### How to reproduce

1. download the app for the first time (maybe optional)
2. signed up with Github (part of the first time setup)
3. connect a local repository
4. setup the GH PR integration
5. see error (despite it connected -- UI shows all green, but Tauri notification shows otherwise)

### Expected behavior

If no actual error, don't show the prompt, but if something is actually broken, surface to the UI and don't show the ✅ as if it did actually work.

Byron:
Thanks a lot for reporting and for the details on how to reproduce the issue!

I think the error message is the same as in #5226.

PavelLaptev:
Closing this as resolved. The GitHub integration setup flow has been completely reworked since v0.13.2. The old error notification pattern that caused false-positive Tauri error toasts has been replaced with structured error handling — errors are now shown inline within the component (via `ReduxResult`) rather than as ambiguous floating notifications. The underlying issue (#5226) has also been addressed as part of this rework.

Issue #5226, opened 2024-10-19
4d26094ceebd43a0b9865516a6c737a0: null is not an object (evaluating 'o(A).close') @tauri://localhost/_app/immutable/chunks/GithubIntegration.CxZqWrly.js:111:659


andreineamtu:
### Version

0.13.2

### Operating System

macOS

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

When I added a local repository and activated github features I got this error:

![Screenshot 2024-10-19 at 18 18 29](https://github.com/user-attachments/assets/09a955f8-eef9-45d7-94a5-b5ac8bf24449)

Byron:
Thanks a lot for reporting!

There are a lot of similar issues showing up right now, but this one is special as it appears to be an error mostly caused by the frontend, probably some return value it expected that was in fact `null`. It was unexpected enough to even get past typescripts type system, i.e. nobody ever thought it could be `null` either. CC @estib-vega who might have more of an intuition.

estib-vega:
Sorry about that @andreineamtu, I'll take care of it. 

@Byron is right, it's a type-system issue.
These are a bit especially annoying to catch because the linter is not catching them at all. Will look into a better way of catching these. 

Thanks for reporting!

krlvi:
Hey @andreineamtu, sorry this fell through the cracks. This was fixed in [#5239](https://github.com/gitbutlerapp/gitbutler/pull/5239), and the fix shipped in [0.13.3](https://github.com/gitbutlerapp/gitbutler/releases/tag/release/0.13.3). Closing this as resolved.

PR #5239, opened 2024-10-21
fix: Type the modals correctly

estib-vega:
## ☕️ Reasoning
This pull request improves the handling of modals by updating their type definitions. This change allows for `undefined` types when interacting with modals, reducing the likelihood of errors during the binding process.

## 🧢 Changes
- Updated the type definitions for modals to include `undefined`, enabling safer interactions without the risk of encountering runtime exceptions.

## Issues
As seen in:
- https://github.com/gitbutlerapp/gitbutler/issues/5235
- https://github.com/gitbutlerapp/gitbutler/issues/5226


ndom91 (approved):
Great that you went through and got all these :pray:

PR #5240, opened 2024-10-21
Implement stack commit reordering

krlvi:


crates/gitbutler-tauri/src/virtual_branches.rs
```diff
@@ -397,6 +397,20 @@ pub mod commands {
         Ok(())
     }
 
+    #[tauri::command(async)]
+    #[instrument(skip(projects, windows), err(Debug))]
+    pub fn reorder_stack(
```
krlvi, line 402: fyi @ndom91 this is the new API

PR #5252, opened 2024-10-21
Integrate new hunk dependency algorithm

mtsgrd:
Re-implement "experimental hunk locking" feature flag in the back end, and call new algorithm when enabled.

Stacked on top of: https://github.com/gitbutlerapp/gitbutler/pull/5197
Delta changes: https://github.com/gitbutlerapp/gitbutler/pull/5252/files/17a14685b312b452be9a1ef6abff4cb2c62a1a1d..c08b3a6907778e0f9028a027f4c41f2a93e5d125

### DIFF
--- a/crates/gitbutler-branch-actions/src/status.rs
+++ b/crates/gitbutler-branch-actions/src/status.rs
@@ -1,10 +1,13 @@
+use std::collections::HashSet;
 use std::{collections::HashMap, path::PathBuf, vec};
 
+use crate::file::list_virtual_commit_files;
 use crate::integration::get_workspace_head;
+use crate::BranchStatus;
 use crate::{
     conflicts::RepoConflictsExt,
     file::{virtual_hunks_into_virtual_files, VirtualBranchFile},
-    hunk::{file_hunks_from_diffs, HunkLock, VirtualBranchHunk},
+    hunk::{file_hunks_from_diffs, VirtualBranchHunk},
     BranchManagerExt, VirtualBranchesExt,
 };
 use anyhow::{bail, Context, Result};
@@ -13,9 +16,15 @@ use gitbutler_branch::BranchCreateRequest;
 use gitbutler_cherry_pick::RepositoryExt as _;
 use gitbutler_command_context::CommandContext;
 use gitbutler_diff::{diff_files_into_hunks, GitHunk, Hunk, HunkHash};
+use gitbutler_hunk_dependency::{
+    compute_hunk_locks, HunkDependencyOptions, HunkLock, InputCommit, InputDiff, InputFile,
+    InputStack,
+};
 use gitbutler_operating_modes::assure_open_workspace_mode;
 use gitbutler_project::access::WorktreeWritePermission;
+use gitbutler_repo::{LogUntil, RepositoryExt as _};
 use gitbutler_stack::{BranchOwnershipClaims, OwnershipClaim, Stack, StackId};
+use itertools::Itertools;
 use tracing::instrument;
 
 /// Represents the uncommitted status of the applied virtual branches in the workspace.
@@ -47,15 +56,12 @@ pub fn get_applied_status_cached(
     worktree_changes: Option<gitbutler_diff::DiffByPathMap>,
 ) -> Result<VirtualBranchesStatus> {
     assure_open_workspace_mode(ctx).context("ng applied status requires open workspace mode")?;
+    let workspace_head = get_workspace_head(ctx)?;
     let mut virtual_branches = ctx
         .project()
         .virtual_branches()
         .list_branches_in_workspace()?;
     let base_file_diffs = worktree_changes.map(Ok).unwrap_or_else(|| {
-        // TODO(ST): Ideally, we can avoid calling `get_workspace_head()` as everyone who modifies
-        //           any of its inputs will update the intragration commit right away.
-        //           It's for another day though - right now the integration commit may be slightly stale.
-        let workspace_head = get_workspace_head(ctx)?;
         gitbutler_diff::workdir(ctx.repository(), workspace_head.to_owned())
             .context("failed to diff workdir")
     })?;
@@ -90,11 +96,20 @@ pub fn get_applied_status_cached(
             .collect();
 
     let vb_state = ctx.project().virtual_branches();
-    let base_tree = ctx
-        .repository()
-        .find_commit(vb_state.get_default_target()?.sha)?
-        .tree()?;
-    let locks = compute_locks(ctx.repository(), &base_diffs, &virtual_branches, base_tree)?;
+    let default_target = vb_state.get_default_target()?;
+
+    let locks = if ctx.project().use_new_locking {
+        compute_locks(
+            ctx,
+            &workspace_head,
+            &default_target.sha,
+            &base_diffs,
+            &virtual_branches,
+        )?
+    } else {
+        let base_tree = ctx.repository().find_commit(default_target.sha)?.tree()?;
+        compute_old_locks(ctx.repository(), &base_diffs, &virtual_branches, base_tree)?
+    };
 
     for branch in &mut virtual_branches {
         if let Err(e) = branch.initialize(ctx) {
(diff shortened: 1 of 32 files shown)


crates/gitbutler-hunk-dependency/src/diff.rs
```diff
@@ -0,0 +1,76 @@
...
+pub struct InputDiff {
+    pub old_start: u32,
+    pub old_lines: u32,
+    pub new_start: u32,
+    pub new_lines: u32,
+}
+
+impl InputDiff {
+    pub fn net_lines(&self) -> i32 {
+        self.new_lines as i32 - self.old_lines as i32
```
krlvi: You can use the do different kinds of subtraction as a method on the new_lines, and you wont have to cast.
You probably wanna do the 'checked' kind though, which returns an Option. 

Seems like net_lines is called from add_new which is called from add, which returns a Result. 
So you can do something like 
```rust
        self.net_lines()
            .checked_sub_unsigned(self.old_lines)
            .ok_or(anyhow!("Overflowed computing net lines"))
```
and do the ? operator. It's gonna be so much easier to debug in the future

crates/gitbutler-hunk-dependency/src/workspace.rs
```diff
@@ -0,0 +1,137 @@
...
+            }
+            stacks.push(stack);
+        }
+        let paths = stacks
+            .iter()
+            .flat_map(|stack| stack.unique_paths())
+            .collect::<HashSet<_>>();
+
+        Ok(WorkspaceRanges {
+            paths: HashMap::from_iter(
```
krlvi: You can skip from_iter and have the compiler infer the type like this
```rust
paths: paths
                .iter()
                .map(|path| (path.clone(), combine_path_ranges(path, &stacks)))
                .collect(),
                ```

File: crates/gitbutler-hunk-dependency/src/workspace.rs [hunk 3]
```diff
@@ -0,0 +1,137 @@
...
+    path::{Path, PathBuf},
+};
+
+use gitbutler_stack::StackId;
+use itertools::Itertools;
+
+use crate::{diff::InputDiff, hunk::HunkRange, stack::StackRanges};
+
+#[derive(Debug, Default, PartialEq, Clone)]
+pub struct WorkspaceRanges {
```
krlvi, line 11: Kinda related to our IRL conversation earlier... so there is this convention to no "stutter" in names - this is module workspace, name WorkspaceRanges. 
So if you want your API to be consumed as a hierarchy, it would be idiomatic to have `workspace::Ranges` at the consumption side.
I tend to prefer a flatter structure - exporting these types in the root level, and in that case I need the name would be `WorkspaceRanges` to be descriptive. 
This is doing both which is suboptimal. Can rename later but trying to help shape the thinking here a bit

crates/gitbutler-hunk-dependency/src/workspace.rs
```diff
@@ -0,0 +1,137 @@
...
+                let InputCommit { commit_id, files } = commit;
+                for file in files {
+                    stack.add(stack_id, commit_id, &file.path, file.diffs)?;
+                }
+            }
+            stacks.push(stack);
+        }
+        let paths = stacks
+            .iter()
+            .flat_map(|stack| stack.unique_paths())
```
krlvi: You can write these like `.flat_map(StackRanges::unique_paths)` as well

crates/gitbutler-hunk-dependency/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+pub mod diff;
```
krlvi: There is still quite a bit of panic-y access by index. You can see it by adding this clippy rule at the top of the file
```
#![warn(clippy::indexing_slicing)]
```
While annoying to replace now, it will pay dividends if/when we hit stuff to debug

crates/gitbutler-branch-actions/tests/virtual_branches/locking.rs
```diff
@@ -0,0 +1,128 @@
...
+    repository.commit_all("initial commit");
+    repository.push();
+
+    gitbutler_branch_actions::set_base_branch(
+        project,
+        &"refs/remotes/origin/master".parse().unwrap(),
+    )
+    .unwrap();
+
+    // Introduce a change that should lock the last change to the first branch.
```
krlvi, line 25: its a bit not clear to me - where do we assert locking?

crates/gitbutler-branch-actions/src/status.rs
```diff
@@ -245,6 +259,83 @@ pub fn get_applied_status_cached(
...
+        .collect::<HashSet<_>>();
+    let files_touched_by_diffs = base_diffs
+        .keys()
+        .map(|p| p.as_path())
+        .collect::<HashSet<_>>();
+
+    let touched_by_both = files_touched_by_commits
+        .intersection(&files_touched_by_diffs)
+        .map(|item| item.to_owned())
+        .map(|item| item.to_owned())
```
mtsgrd: It's unclear to me how I need to change `files_touched_by_commits` and `files_touched_by_diffs` to avoid this double `.to_owned()`.. 🤔

crates/gitbutler-hunk-dependency/src/diff.rs
mtsgrd: That function `checked_sub_unsigned` is only available on `i32`, but `new_lines` and `old_lines` are both `u32`. What can I do?

crates/gitbutler-branch-actions/src/status.rs
Byron: What about *no* `to_owned()` ;)?

```patch
diff --git a/crates/gitbutler-branch-actions/src/status.rs b/crates/gitbutler-branch-actions/src/status.rs
index 18371cd24..605582862 100644
--- a/crates/gitbutler-branch-actions/src/status.rs
+++ b/crates/gitbutler-branch-actions/src/status.rs
@@ -289,8 +289,7 @@ fn compute_locks(
 
     let touched_by_both = files_touched_by_commits
         .intersection(&files_touched_by_diffs)
-        .map(|item| item.to_owned())
-        .map(|item| item.to_owned())
+        .copied()
         .collect::<Vec<_>>();
 
     let mut stacks_input: Vec<InputStack> = vec![];
@@ -302,7 +301,7 @@ fn compute_locks(
             let commit = repo.find_commit(commit_id)?;
             let files = list_virtual_commit_files(ctx, &commit, false)?;
             for file in files {
-                if touched_by_both.contains(&file.path) {
+                if touched_by_both.contains(&file.path.as_path()) {
                     let value = InputFile {
                         path: file.path,
                         diffs: file
```

Generally, `to_owned()` shouldn't be needed unless one wants to obtain ownership to return from the function, or similar.

crates/gitbutler-hunk-dependency/src/diff.rs
Byron: I think what we'd want here is the nightly-only [`checked_signed_diff()`](https://doc.rust-lang.org/core/primitive.u32.html#method.checked_signed_diff), starting at `u32` and going to `i32` like what seems to be intended here.
Nightly might need to be updated in the toolchain file to make it available.

crates/gitbutler-branch-actions/src/status.rs
krlvi: another thing i was gonna suggest - since what you use later in the function is PathBuf - you can skip the .as_path() and git it also a .cloned()

krlvi:
The two outstanding challenges here are:
- handling of merge commits. i have an idea what we can do there though
- performance - we can test this after merging, but it's very important that we test this with a proper/realistic test with many commits etc

crates/gitbutler-hunk-dependency/src/diff.rs
krlvi: ooh this is a really cool function

krlvi (approved):

PR #5256, opened 2024-10-22
Fixes a bug where the UI may freeze due to incorrectly processed duplicate changeID

krlvi:
If a branch/stack has more than one commit with the same change ID - well that breaks the UI.
While a change ID being duplicated in the branch/stack is a bad state to be in, it is possible to be achieved. And if the app finds itself in that condition, it needs to handle it.

This PR removes that assumption that there would be only one commit with a given changeID in branch.

This issue was initially reported on Discord: https://discord.com/channels/1060193121130000425/1297929207329984542
Once this occurs, the UI is unable to handle key duplication and freezes up entirely, with no error logs entries anywhere. 

### How is it possible to end up with duplicate change IDs?
Here are at least to scenarios, curtesy of Caleb:

**Rebase on GitHub**
- Make branch A with a commit
- Push branch A
- Use GitHub to rebase branch A (because we know this preserves commit ids)
- Click "integrate upstream", which creates the commits `A -> A'`

**Using the undo history in-app**
- Make a branch A with commit which modifies file foo.txt
- Unapply branch A
- Use history to restore before the unapply (we now have A' applied)
- Drag commit to different branch B
- Modify commit such that it no longer modifies foo.txt, and now modifies bar.txt
- Drop changes and unapply branch A'
- Apply branch A
- Drag commit from branch A into branch B, or vise versa

### DIFF
--- a/crates/gitbutler-branch-actions/src/branch_upstream_integration.rs
+++ b/crates/gitbutler-branch-actions/src/branch_upstream_integration.rs
@@ -45,7 +45,8 @@ pub fn integrate_upstream_commits_for_series(
         repo,
         remote_head.id(),
         stack_merge_base,
-    )?;
+    )?
+    .head;
 
     let do_rebease = branch.allow_rebasing || Some(subject_series) != all_series.first();
     let integrate_upstream_context = IntegrateUpstreamContext {
--- a/crates/gitbutler-branch-actions/src/stack.rs
+++ b/crates/gitbutler-branch-actions/src/stack.rs
@@ -1,12 +1,15 @@
 use std::collections::HashMap;
 
+use anyhow::anyhow;
 use anyhow::{Context, Result};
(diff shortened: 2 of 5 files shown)


abkfenris:
It looks like that did the trick unblocking the UI. 

That branch is now showing 'Force Push' even after 'Force Push'-ing, and the Github UI doesn't show that it was a force push. It seems that my attempts to previously recover an old commit have now unearthed a bunch more recovered commits.

image
image
image
image

krlvi:
Thanks for testing @abkfenris! And thanks for spending all this time debugging together yesterday. Im gonna further investigate the "force push" label as a separate item

abkfenris:
You're welcome. Want me to keep that branch around to test against? Or is there a way that I can extract it's state for you to test yourselves?

PR #5276, opened 2024-10-23
fix: show commit fileList when there are only conflictedFiles

ndom91:
## ☕️ Reasoning

- Conflicted files were hidden if the commit contained **only** conflicted files

### DIFF
--- a/apps/desktop/src/lib/file/BranchFilesList.svelte
+++ b/apps/desktop/src/lib/file/BranchFilesList.svelte
@@ -126,7 +126,7 @@
 	</div>
 {/if}
 
-{#if displayedFiles.length > 0}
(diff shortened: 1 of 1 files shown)


apps/desktop/src/lib/file/BranchFilesList.svelte
```diff
@@ -126,7 +126,7 @@
 	</div>
 {/if}
 
-{#if displayedFiles.length > 0}
+{#if displayedFiles.length > 0 || (conflictedFiles?.entries.size || 0) > 0}
```
ndom91, line 129: `conflictedFiles?.entries.size` if not found is `undefined` and `undefined > 0 === false` apparently, so that would have been fine, but then the editor still complains

krlvi (approved):

PR #5289, opened 2024-10-24
fix: Don't list the target branch

estib-vega:
Don't list the target branch in the sidebar

### DIFF
--- a/crates/gitbutler-branch-actions/src/branch.rs
+++ b/crates/gitbutler-branch-actions/src/branch.rs
@@ -9,6 +9,7 @@ use gitbutler_diff::DiffByPathMap;
 use gitbutler_oxidize::{git2_to_gix_object_id, gix_to_git2_oid};
 use gitbutler_project::access::WorktreeReadPermission;
 use gitbutler_reference::normalize_branch_name;
+use gitbutler_reference::RemoteRefname;
 use gitbutler_repo::{GixRepositoryExt, RepositoryExt as _};
 use gitbutler_serde::BStringForFrontend;
 use gitbutler_stack::{Stack as GitButlerBranch, StackId, Target};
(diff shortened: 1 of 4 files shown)


ndom91 (approved):

crates/gitbutler-branch-actions/src/branch.rs
```diff
@@ -343,6 +349,17 @@ impl GroupBranch<'_> {
         }
         .map(BranchIdentity::from)
     }
+
+    /// Determines if the branch is a remote branch by ref name
+    fn is_remote_branch(&self, ref_name: &RemoteRefname) -> bool {
+        if let GroupBranch::Remote(branch) = self {
+            let remote_branch = ref_name.to_string();
```
Byron: There should never be a need to convert `to_string()`. Maybe there is an `as_ref().as_bstr()` for that?

Byron (commented):
Maybe this is related to [this comment](https://github.com/gitbutlerapp/gitbutler/issues/5229#issuecomment-2434644064)?

crates/gitbutler-branch-actions/src/branch.rs
estib-vega: Will update that 👍

estib-vega: There wasn't but maybe `format` is a bit better?
[see](https://github.com/gitbutlerapp/gitbutler/pull/5289/files#diff-beb9965bba602a6a2e20a58f347bc759f37bf01cd2f9826374f34636c4bfb44dR103-R106)
```rs

impl PartialEq<BStr> for Refname {
    fn eq(&self, other: &BStr) -> bool {
        format!("refs/remotes/{}/{}", self.remote, self.branch) == *other
    }
}
```

Byron: It still allocates, so I posted #5303 .

estib-vega: Thanks!

Byron:
Oh, right, I thought/hoped that would fix the issue it was posted in. But even if not, I think thanks to your help it's a bit closer.

PR #5298, opened 2024-10-24
fix: type all buttons

estib-vega:
Apparently buttons without a type inside a form default to `submit`.
Example: Inside the Integrate Upstream modal, in order to open the links in the commit list and not trigger the upstream to be integrated, we need to set the type of the button to `button`

More info on the button.type: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#type

### DIFF
--- a/apps/desktop/src/lib/navigation/ProjectSelector.svelte
+++ b/apps/desktop/src/lib/navigation/ProjectSelector.svelte
@@ -6,22 +6,27 @@
 	import Icon from '@gitbutler/ui/Icon.svelte';
 	import Tooltip from '@gitbutler/ui/Tooltip.svelte';
 
-	export let isNavCollapsed: boolean;
+	interface Props {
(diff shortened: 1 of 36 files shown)


PavelLaptev (approved):

apps/desktop/src/lib/navigation/ProjectSelector.svelte
```diff
@@ -6,22 +6,27 @@
...
+	}
+
+	let { isNavCollapsed }: Props = $props();
 
-	let buttonTrigger: HTMLButtonElement;
 	const project = getContext(Project);
 
-	let popup: ProjectsPopup;
+	let buttonTrigger = $state<HTMLButtonElement>();
+	let popup = $state<ReturnType<typeof ProjectsPopup>>();
```
estib-vega, line 18: Small drive-by fix

PR #5303, opened 2024-10-24
minor fixes

Byron:
This adds code-complexity (and tests) to avoid allocating when making a comparison.
The PR also adds type-safety for the comparison.

Follow-up of #5289 .


estib-vega (approved):
Looks good, thanks for the improvements!

PR #5325, opened 2024-10-28
Modal: Close is a promise + onClickOutside

estib-vega:
## ☕️ Reasoning
Enhance modal functionality by making the closing process a promise and adding an optional callback for clicking outside the modal.

### DIFF
--- a/packages/ui/src/lib/Modal.svelte
+++ b/packages/ui/src/lib/Modal.svelte
@@ -11,7 +11,17 @@
 		type?: 'info' | 'warning' | 'error' | 'success';
 		title?: string;
 		noPadding?: boolean;
+		/**
+		 * Callback to be called when the modal is closed.
+		 *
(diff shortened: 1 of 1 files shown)


packages/ui/src/lib/Modal.svelte
```diff
@@ -52,14 +63,20 @@
 		window.addEventListener('keydown', handleKeyDown);
 	}
 
-	export function close() {
+	export function close(): Promise<void> {
+		if (isClosing) return Promise.resolve();
```
mtsgrd: If this function is called twice, then it looks to me like the second call would resolve immediately while the original resolves in 100ms. Is this intended?

mtsgrd (approved):

packages/ui/src/lib/Modal.svelte
estib-vega: You're right. Makes sense to resolve the promise only until it's fully closed, regardless of if the function is called a second time

PR #5328, opened 2024-10-28
fix(nix): flake dev env environment variables

ndom91:
## ☕️ Reasoning

- Add `GLIB_EXTRA_MODULES` `glib-networking` related env var to enable tls in dev environment


Caleb-T-Owens (approved):
glib glib

estib-vega (approved):

PR #5330, opened 2024-10-28
fix: Modal closes on mouse down outside

estib-vega:
## ☕️ Reasoning
The modal was unintentionally closing when a user clicked down inside the modal and released the mouse outside. This pull request addresses this behavior to only close the modal when the user clicks down outside the modal.

## 🧢 Changes
- Updated the functionality to close the modal only on clicking down outside the modal.


Caleb-T-Owens (approved):
LGTM

PR #5331, opened 2024-10-28
Remove feature flag for new locking algorithm

mtsgrd:


krlvi:
my recommendation is to make the feature flag default to enabled and keep the ability to switch to the old implementation for one or two cycles

mtsgrd:
right, let's do that.

PR #5332, opened 2024-10-28
nit: small change to locking code

mtsgrd:
- optional return value from intersection function
- filter those values in compute locks function

### DIFF
--- a/crates/gitbutler-hunk-dependency/src/workspace.rs
+++ b/crates/gitbutler-hunk-dependency/src/workspace.rs
@@ -50,14 +50,17 @@ impl WorkspaceRanges {
     }
 
     /// Finds commits that intersect with a given path and range combination.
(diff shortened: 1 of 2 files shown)


crates/gitbutler-hunk-dependency/src/workspace.rs
```diff
@@ -202,12 +205,12 @@ mod tests {
             },
         ])?;
 
-        let dependencies_1 = workspace_ranges.intersection(&path, 2, 1);
+        let dependencies_1 = workspace_ranges.intersection(&path, 2, 1).unwrap();
```
krlvi, line 208: why is this safe to do here?

mtsgrd: Because it's a test?

krlvi: 🤦

krlvi (approved):

PR #5339, opened 2024-10-28
Default to experimental locking

mtsgrd:
- changing the name of the field since existing projects would have an explicit false value already


mtsgrd:
@krlvi in changing the name of the property the old ones that are already explicitly set to false should be dropped, and if I understand correctly this new one should default to true. Wdyt?

krlvi (approved):

krlvi:
> @krlvi in changing the name of the property the old ones that are already explicitly set to false should be dropped, and if I understand correctly this new one should default to true. Wdyt?

yeah that will make sure that the true gets deployed for everyone

PR #5344, opened 2024-10-28
Add sequence count to branches. Design variant #2

PavelLaptev:
## ☕️ Reasoning

if more than one branch

image

if less

image

---

in case if more then 10 branches

image

---

image


Caleb-T-Owens (approved):
LGTM

PR #5345, opened 2024-10-28
patch-view

Caleb-T-Owens:
Patch view


packages/shared/src/lib/cloud/patches/events.ts
```diff
@@ -0,0 +1,147 @@
...
+		this.outdated = apiChatMessage.outdated;
+		this.issue = apiChatMessage.issue;
+		this.resolved = apiChatMessage.resolved;
+		this.threadId = apiChatMessage.thread_id;
+		this.comment = apiChatMessage.comment;
+		this.diffSha = apiChatMessage.diff_sha;
+		this.range = apiChatMessage.range;
+		this.diffPatchArray = apiChatMessage.diff_patch_array;
+		this.createdAt = new Date(apiChatMessage.created_at);
+		this.updatedAt = new Date(apiChatMessage.updated_at);
```
mtsgrd, line 83: Could this be replaced with `Object.assign(this, apiChatMessage)`?

Caleb-T-Owens: no, because there are casing changes

Caleb-T-Owens: and type changes

Issue #5307, opened 2024-10-24
Error when resolving merge conflict using Gitbluter and VS Code


Ignacio9172:
### Version

0.13.6

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

While trying to perform a merge using Gitbluter, I encountered a conflict in the files. I selected the "Resolve Conflict" option, and when I clicked to open the files in VS Code, I got an error. I have tried redoing the merge and restarting both Gitbluter and VS Code, but the error persists. I am unable to proceed with resolving the conflicts.

Attached is a screenshot of the error I encountered when attempting to open the files in VS Code.

![image](https://github.com/user-attachments/assets/b1baaafb-3ffd-4b67-8f58-899b42bdcd80)

### How to reproduce

1. Attempt a merge with conflicts in Gitbluter.
2. Select "Resolve Conflict."
3. Click on "Open files in VS Code."
4. The error appears when trying to open the files.

### Expected behavior

I should have been able to open the files in VS Code to manually resolve the conflict and continue with the merge without any issues.

### Relevant log output

```shell
ipc.DuYoxua2.js:1 
 ipc->open_url: {"url":"vscode://file\\C:\\Users\\ignac\\OneDrive\\Documentos\\GitHub\\cortex\\apps\\frontend\\src\\pages\\explore-roadmaps.vue"} s: Invalid domain character
    at s.fromError (https://tauri.localhost/_app/immutable/chunks/ipc.DuYoxua2.js:1:898)
    at l (https://tauri.localhost/_app/immutable/chunks/ipc.DuYoxua2.js:1:1102)
    at async mo (https://tauri.localhost/_app/immutable/chunks/url.DW5Nbmmp.js:14:353) 
Object

app.Bqko059y.js:6  113118acba5842b0a4f8b4bac8dd3d82: Invalid domain character

Error: Invalid domain character
    at s.fromError (https://tauri.localhost/_app/immutable/chunks/ipc.DuYoxua2.js:1:898)
    at l (https://tauri.localhost/_app/immutable/chunks/ipc.DuYoxua2.js:1:1102)
    at async mo (https://tauri.localhost/_app/immutable/chunks/url.DW5Nbmmp.js:14:353)

X @ app.Bqko059y.js:6
ipc.DuYoxua2.js:1  Uncaught (in promise) s: Invalid domain character
[...]

PR #5349, opened 2024-10-29
fix: Open files on editor in Windows

estib-vega:
## ☕️ Reasoning
The changes in this pull request fix the issue related to opening files on the editor in Windows. Previously, there was an issue with the path formatting when opening files using the editor URL on Windows.

More on the formatting of VSCode URLs: https://code.visualstudio.com/docs/editor/command-line#_opening-vs-code-with-urls

## 🧢 Changes
This pull request ensures that forward slashes are used for the path when opening files with the editor URL on Windows.

## Issue
As seen in: https://github.com/gitbutlerapp/gitbutler/issues/5307

### DIFF
--- a/apps/desktop/src/lib/utils/url.ts
+++ b/apps/desktop/src/lib/utils/url.ts
@@ -3,6 +3,8 @@ import { showToast } from '$lib/notifications/toasts';
 import GitUrlParse from 'git-url-parse';
 import { posthog } from 'posthog-js';
 
+const SEPARATOR = '/';
+
 export async function openExternalUrl(href: string) {
 	try {
 		await invoke<void>('open_url', { url: href });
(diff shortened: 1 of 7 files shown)


apps/desktop/src/lib/utils/url.ts
```diff
@@ -39,3 +41,30 @@ export function remoteUrlIsHttp(url: string): boolean {
...
+
+export interface EditorPathParams {
+	schemeId: string;
+	path: string[];
+	searchParams?: Record<string, string>;
+	line?: number;
+	column?: number;
+}
+
+export function getEditorPath(params: EditorPathParams): string {
```
mtsgrd: Would this maybe be more appropriately named `getEditorUri`?

mtsgrd (approved):

apps/desktop/src/lib/utils/url.ts
estib-vega: Agreed

Issue #5317, opened 2024-10-25
Uncommitting moves the commit's changes to the default branch


nathanwhit:
### Version

0.13.7

### Operating System

macOS

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

When uncommitting a commit, GitButler puts the changes that were in the commit into the default branch, instead of the branch the commit was in.

https://github.com/user-attachments/assets/71ac27f6-f914-481f-8669-f58e650743b1

### How to reproduce

Create two virtual branches. Make a commit on branch A. Set branch B to be the default branch. Uncommit the commit on branch A. Notice that the changes that were in the commit have been staged on branch B, instead of branch A.

### Expected behavior

Uncommitting a commit should retain the branch of the commit's contents, regardless of the default branch.

Byron:
Thanks a lot for reporting, the video makes the issue abundantly clear as well. Besides that, I was able to reproduce it with ease, too.

I also think that the current behaviour is unintuitive and one would expect uncommitted changes to be associated with their lane.

estib-vega:
Hey @Byron, if you're not already looking into this, I can take a look :)

Byron:
Yes, that would be great if you did, thank you. *(I will always assign myself if I plan to pick an issue up)*

estib-vega:
The fix should be available in the next release :)

PR #5354, opened 2024-10-29
Undo commit: Files stay in the branch that owned the undone commit

estib-vega:
## ☕️ Reasoning
This pull request aims to address the issue where undoing a commit moves the files to the default branch instead of keeping them in the source branch. The changes ensure that the files stay in the branch from which the commit was undone.

## 🧢 Changes
- Updated the functionality to retain files in the source branch when undoing a commit.

## Issues
As seen in https://github.com/gitbutlerapp/gitbutler/issues/5317

### DIFF
--- a/crates/gitbutler-branch-actions/src/undo_commit.rs
+++ b/crates/gitbutler-branch-actions/src/undo_commit.rs
@@ -1,8 +1,11 @@
+use std::collections::HashMap;
+
 use anyhow::{bail, Context as _, Result};
 use gitbutler_command_context::CommandContext;
 use gitbutler_commit::commit_ext::CommitExt as _;
+use gitbutler_diff::Hunk;
 use gitbutler_repo::{rebase::cherry_rebase_group, LogUntil, RepositoryExt as _};
-use gitbutler_stack::{Stack, StackId};
+use gitbutler_stack::{OwnershipClaim, Stack, StackId};
 
 use crate::VirtualBranchesExt as _;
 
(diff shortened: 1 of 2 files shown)


Caleb-T-Owens (approved):
I'm wondering if this is entirely accurate as the ownerships are created in the space of the commit, rather than in the context of the entire project.

IE:

If I had one branch A which had uncommitted change which added 10 lines to the bottom of a file.
I then had another branch B had committed change modified lines 17 to 25.
If branch A was the default branch

If I undid the commit in branch B; I imagine it could actually end up swapping the hunks around, such that branch A contained the 15 to 25 change, and branch B had the 1 to 10 change. (or at the very least, end up with both in one of the branches). Essentially the same old branch scope vs workspace scope for hunks

---

It should be an improvement in most cases. There is a more complex approach that we could take, but this is fine for now. It should just be noted that this is more like hitting it with a magical rake.

estib-vega:
> I'm wondering if this is entirely accurate as the ownerships are created in the space of the commit, rather than in the context of the entire project.
> 
> IE:
> 
> If I had one branch A which had uncommitted change which added 10 lines to the bottom of a file. I then had another branch B had committed change modified lines 17 to 25. If branch A was the default branch
> 
> If I undid the commit in branch B; I imagine it could actually end up swapping the hunks around, such that branch A contained the 15 to 25 change, and branch B had the 1 to 10 change. (or at the very least, end up with both in one of the branches). Essentially the same old branch scope vs workspace scope for hunks
> 
> It should be an improvement in most cases. There is a more complex approach that we could take, but this is fine for now. It should just be noted that this is more like hitting it with a magical rake.

@Caleb-T-Owens 
Have a bunch of follow-up questions, but we can take it in other channels so that it's a bit faster.

Wish there were better ways to communicate while on a review. smh.

Will merge this in the meantime!

PR #5356, opened 2024-10-29
Emit virtual branches after fetch

krlvi:
This is because, in case a branch was integrated, the state is updated


Caleb-T-Owens (approved):
🤔 How on earth do we not do this already?

PR #5357, opened 2024-10-29
Persist ForgeIdentifier on series (branches)

krlvi:
Sub tasks:
- [x] Introduce a new type `ForgeIdentifier` and have it persisted
- [x] Add the ability to set the forge identifiers for a branch
- [x] Tests for setting forge ids
- [x] Expose forge ids on list virtual branches stream
- [x] Add API endpoint for setting the forge ids
- [x] Ensure that when a lane is creating from an existing branch, forge ID is set
- [x] Integration test setting forge ids + virtual branches response

### DIFF
--- a/apps/desktop/src/lib/vbranches/types.ts
+++ b/apps/desktop/src/lib/vbranches/types.ts
@@ -449,7 +449,7 @@ export class PatchSeries {
 	 * A list of identifiers for the review unit at possible forges (eg. Pull Request).
 	 * The list is empty if there is no review units, eg. no Pull Request has been created.
 	 */
-	forgeIds!: ForgeIdentifier[];
+	forgeId?: ForgeIdentifier | undefined;
 
 	get localCommits() {
 		return this.patches.filter((c) => c.status === 'local');
--- a/apps/desktop/src/lib/vbranches/types.ts
+++ b/apps/desktop/src/lib/vbranches/types.ts
@@ -445,6 +445,12 @@ export class PatchSeries {
 	@Type(() => DetailedCommit)
 	upstreamPatches!: DetailedCommit[];
 
+	/**
+	 * A list of identifiers for the review unit at possible forges (eg. Pull Request).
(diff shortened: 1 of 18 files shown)


crates/gitbutler-stack/src/stack.rs
```diff
@@ -655,6 +658,33 @@ impl Stack {
...
+    /// If the stack has not been initialized, this method will return an error.
+    pub fn set_forge_ids(
+        &mut self,
+        ctx: &CommandContext,
+        series_name: &str,
+        new_forge_ids: Vec<ForgeIdentifier>,
+    ) -> Result<()> {
+        if !self.initialized() {
+            return Err(anyhow!("Stack has not been initialized"));
+        }
```
mtsgrd, line 675: We have these three lines in 9 other places, could we maybe make something like `self.assertInitialized()?;` that either returns this error or just Ok's?

apps/desktop/src/lib/vbranches/types.ts
```diff
@@ -445,6 +445,12 @@ export class PatchSeries {
 	@Type(() => DetailedCommit)
 	upstreamPatches!: DetailedCommit[];
 
+	/**
+	 * A list of identifiers for the review unit at possible forges (eg. Pull Request).
+	 * The list is empty if there is no review units, eg. no Pull Request has been created.
+	 */
+	forgeIds!: ForgeIdentifier[];
```
mtsgrd: In the GitHub UI you can't create more than one PR for a branch. What is your reasoning here for an array of forge id's rather than just an optional one?

crates/gitbutler-stack/src/stack.rs
krlvi: Yeah good idea. the `initialized()` method itself could do that. This code is probably ready for a light refactor now, since i have only been adding to it for a couple of weeks

mtsgrd (approved):

PR #5366, opened 2024-10-30
fix: do not cutoff traversals by date for branch details (#1650).

Byron:
We already found the merge-base and know that this will naturally be the end of the traversal. Thus there is no need to additionally set a cut-off date, which is the documented default.

With such a cutoff, it's very possible that certain operations on the branches change the date of the base to lie in the future, which is when no traversal would be able to happen anymore.

Fixes #1650.

### Tasks

* [x] initial fix
* [x] safe fix

### Notes for the reviewer

* The implementation is a bit more complex to be sure that the cutoff is still used where possible. The GitLab repository for instance has such complex graphs that somehow... the first mergebase often isn't enough to cancel the traversal.

Screenshot 2024-10-30 at 16 02 28

### DIFF
--- a/crates/gitbutler-branch-actions/src/branch.rs
+++ b/crates/gitbutler-branch-actions/src/branch.rs
@@ -571,6 +571,7 @@ pub fn get_branch_listing_details(
                                 let revwalk = repo
                                     .rev_walk(Some(branch_head))
                                     .with_pruned(Some(base))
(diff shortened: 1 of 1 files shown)


krlvi (approved):
ship it :)

mtsgrd (approved):

PR #5376, opened 2024-10-30
Store forgeId after creating a new pull request

mtsgrd:
With this id persisted we do not need to match prs and branches by their names. Since pull requests will now be visible after being closed I also add a button for reopening it.


mtsgrd:
@krlvi I had a chat with @PavelLaptev about the idea of being able to create a new PR to replace a closed one, but we came to the conclusion we probably shouldn't get into that complexity unless there is a demonstrated need for it. Can you picture scenarios where it's vital we can disassociate a PR from a patch series?

krlvi:
> @krlvi I had a chat with @PavelLaptev about the idea of being able to create a new PR to replace a closed one, but we came to the conclusion we probably shouldn't get into that complexity unless there is a demonstrated need for it. Can you picture scenarios where it's vital we can disassociate a PR from a patch series?

no, I agree, lets start simple

PavelLaptev (approved):

PR #5380, opened 2024-10-31
Improve textarea and fix bugs

PavelLaptev:
## ☕️ Reasoning
- Improve calculation method
- Fix PR textarea
- Remove old component
- update commit message with the new `textarea` component


packages/ui/src/lib/Textarea.svelte
```diff
@@ -70,8 +73,24 @@
...
+
+			measureEl.style.fontFamily = textBoxElStyles.fontFamily;
+			measureEl.style.fontSize = textBoxElStyles.fontSize;
+			measureEl.style.fontWeight = textBoxElStyles.fontWeight;
+			measureEl.style.padding = textBoxElStyles.padding;
+			measureEl.style.width = textBoxElStyles.width;
+			measureEl.style.border = textBoxElStyles.border;
+			measureEl.style.width = textBoxElStyles.width;
+		}
+	});
```
Caleb-T-Owens: This ought to be an `$effect` block

packages/ui/src/lib/Textarea.svelte
```diff
@@ -84,8 +103,10 @@
 		}
 	});
 
-	let maxHeight = fontSize * 1.5 * maxRows + padding.top + padding.bottom;
-	let minHeight = fontSize * 1.5 * minRows + padding.top + padding.bottom;
+	const lineHeight = 1.6;
+
+	let maxHeight = fontSize * maxRows + padding.top + padding.bottom;
+	let minHeight = fontSize * minRows + padding.top + padding.bottom;
```
Caleb-T-Owens: These depend on reactive properties (from `$props`) so these ought to be `$derived` statements

packages/ui/src/lib/Textarea.svelte
PavelLaptev: right!

Caleb-T-Owens (changes requested):
A net loss of ~250 LOC, awesome!

It is however important that we fix these reactivity issues.

If you're dealing with **anything** that is declared in `$props` or with `$state`, you must then consume it via `$effect` or `$derived`, otherwise you will run into reactivity issues.

packages/ui/src/lib/Textarea.svelte
PavelLaptev: danke! Fixed

Caleb-T-Owens (approved):
Speedy Danke!

PavelLaptev:
> A net loss of ~250 LOC, awesome!
> 
> It is however important that we fix these reactivity issues.
> 
> If you're dealing with **anything** that is declared in `$props` or with `$state`, you must then consume it via `$effect` or `$derived`, otherwise you will run into reactivity issues.

Thanks for pointing that out @Caleb-T-Owens

ndom91 (approved):

PR #5387, opened 2024-10-31
Prevent escape from closing "create pull request" modal that contains changes

simenbrekken:
## ☕️ Reasoning

After carefully crafting a pull request description for 10 minutes, the last thing you expect is for an accidental press of escape to close the dialog and destroy your stuff.

## 🧢 Changes

If title or body has been changed, prompt the user to confirm before closing dialog.

### DIFF
--- a/apps/desktop/src/lib/pr/PrDetailsModal.svelte
+++ b/apps/desktop/src/lib/pr/PrDetailsModal.svelte
@@ -32,6 +32,7 @@
 	import { openExternalUrl } from '$lib/utils/url';
 	import { BranchController } from '$lib/vbranches/branchController';
 	import { PatchSeries, VirtualBranch } from '$lib/vbranches/types';
+	import { confirm } from '@tauri-apps/plugin-dialog';
 	import { getContext, getContextStore } from '@gitbutler/shared/context';
 	import { persisted } from '@gitbutler/shared/persisted';
 	import Button from '@gitbutler/ui/Button.svelte';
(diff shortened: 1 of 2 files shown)


Byron:
Thank you! I would think that pressing ESC was very intentionally linked, but I'd also expect it to ask to be sure the user is fine loosing changes they made.  CC @PavelLaptev as there might be other ways to prevent the user from accidentally loosing changes without completely disabling keys.

simenbrekken:
> I would think that pressing ESC was very intentionally linked, but I'd also expect it to ask to be sure the user is fine loosing changes they made. CC @PavelLaptev as there might be other ways to prevent the user from accidentally loosing changes without completely disabling keys.

I've added a confirm dialog now that makes more sense.

krlvi:
Maybe another way to solve this could be to save the PR description in progress in local storage, so that if one presses esc by accident, reopening the modal it will be pre-filled with the content before

simen-ardoq:
> Maybe another way to solve this could be to save the PR description in progress in local storage, so that if one presses esc by accident, reopening the modal it will be pre-filled with the content before

Even better!

PavelLaptev:
> Maybe another way to solve this could be to save the PR description in progress in local storage, so that if one presses esc by accident, reopening the modal it will be pre-filled with the content before

Agreed. I would be nice if we can do this.
@estib-vega what do you think, will you have time to implement this?

estib-vega:
Sorry for the long wait! I somehow missed this completely

> > Maybe another way to solve this could be to save the PR description in progress in local storage, so that if one presses esc by accident, reopening the modal it will be pre-filled with the content before
> 
> Agreed. I would be nice if we can do this. @estib-vega what do you think, will you have time to implement this?

Yes! Let's do it

simenbrekken:
Superseded by #5559, thank you @estib-vega

PR #5388, opened 2024-10-31
Fix reactivity in pr template selection

mtsgrd:
- revert to template when input empty on change rather than keystroke
- use empty string for state variables since it the input default


PavelLaptev (approved):
LGTM!

PR #5396, opened 2024-11-01
fix: Hunk dependencies - Pass the commits in application order

estib-vega:
Pass the commits in application order when calculating the hunk diffs


estib-vega:
@mtsgrd I'll add the tests in the scope of the other adaptions coming.
I want to add a suite of integration tests

mtsgrd (approved):

mtsgrd:
@krlvi fyi, this is an important fix.

krlvi:
> @krlvi fyi, this is an important fix.

Ordering is a common error that is easy to make in our code. Generally in rust, as a convention, our functions take order like
[ newest commit, oldest commit ] simply because this is how the git "linked list" is ordered and easiest to traverse.
But regardless, really important to document this on the function level cause it's easy to mix up

PR #5399, opened 2024-11-01
Introduce gitbutler-settings crate

krlvi:
This PR adds 2 things:
- A crate/struct/mechanism for accessing application settings in various commands (using tauri_plugin_store but not tied to it)
- First usage of the new settings mechanism for optionally configuring githubOauthClientId (there are rare use cases where it is helpful for people to set their own client ID)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2837,6 +2837,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "gitbutler-settings"
+version = "0.0.0"
+
+[[package]]
 name = "gitbutler-stack"
 version = "0.0.0"
 dependencies = [
@@ -2928,6 +2932,7 @@ dependencies = [
  "gitbutler-repo",
  "gitbutler-repo-actions",
  "gitbutler-secret",
+ "gitbutler-settings",
  "gitbutler-stack",
  "gitbutler-storage",
  "gitbutler-sync",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ members = [
     "crates/gitbutler-stack",
     "crates/gitbutler-patch-reference",
     "crates/gitbutler-hunk-dependency",
+    "crates/gitbutler-settings",
 ]
 resolver = "2"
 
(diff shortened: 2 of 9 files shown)


krlvi:
@Byron please critique this approach for settings propagation in the app. The need here is to have app settings that are not tied to a user or a project. 

We were already using the `tauri_plugin_store`, but from the frontend only, to set error reporting configs (The UI can set and read values like this:
https://github.com/gitbutlerapp/gitbutler/blob/ad2a37113dc7568f0537a26152d279c4e523e9fb/apps/desktop/src/lib/config/appSettings.ts#L16)
without the rust code being aware of the keys. 

This change introduces the currently present keys to a struct in rust (but technically it doesn't need to include all) to a struct on the rust side, and it also adds a new field that is not known on the UI side.
Of course, we could be parsing the `settings.json` with serde, but in the spirit of using the plugin, it is reading and writing values with it.

Byron (approved):
Thanks for letting me take a look!

I see that this plugin can be used for application settings, and the frontend and backend can easily access the same JSON file that way.

As a special feature, the store integrates with the tauri event system and auto-saves after 100ms after a change.

All looks good enough to me, except that it will be possible to mess up the JSON file [as it doesn't write it atomically](https://docs.rs/tauri-plugin-store/latest/src/tauri_plugin_store/store.rs.html#281).

Of course it will be fine most of the time so it isn't a real issue.

krlvi:
> Thanks for letting me take a look!
> 
> I see that this plugin can be used for application settings, and the frontend and backend can easily access the same JSON file that way.
> 
> As a special feature, the store integrates with the tauri event system and auto-saves after 100ms after a change.
> 
> All looks good enough to me, except that it will be possible to mess up the JSON file [as it doesn't write it atomically](https://docs.rs/tauri-plugin-store/latest/src/tauri_plugin_store/store.rs.html#281).
> 
> Of course it will be fine most of the time so it isn't a real issue.

thanks for taking a look. I imagine that from the rust side of things we will only be reading values. Im gonna do some more testing and then merge this 🙏

Issue #5163, opened 2024-10-16
GitLab repository with 8638 commits ahead can't be handled


Byron:
### Version

0.13.1

### Operating System

macOS

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

After not touching the GitLab repository for a while, it's very far behind.

<img width="925" alt="Screenshot 2024-10-16 at 13 45 06" src="https://github.com/user-attachments/assets/f9e5c53f-384a-4a04-a096-726bcbaaf519">

However, the initial spinner takes about 30 minutes to disappear so the virtual branch is shown.
I did not manage to update the workspace successfully, and the new 'incoming changes' popup seemed to be taking a lot of resources from the browser window so it was pegged at 1CPU at all times.

**I think this is a regression as even dealing with remotes being 5000 commits ahead of the local branches was absolutely possible previously**.

### Additional information

* The backend profile showing that diff-tree-to-tree takes most of the time. <img width="932" alt="Screenshot 2024-10-16 at 13 50 25" src="https://github.com/user-attachments/assets/6a138ae2-523c-45ee-897b-57e77e85d59a">
* WebContent rendering once the backend is done <img width="695" alt="Screenshot 2024-10-16 at 13 51 57" src="https://github.com/user-attachments/assets/24e6c1c0-9a95-476c-bea2-9437010982aa">
* The profiles themselves from which the screenshots were taken [macos-instruments-profiles.zip](https://github.com/user-attachments/files/17395127/macos-instruments-profiles.zip)

### How to reproduce

Use the GitLab repository (https://gitlab.com/gitlab-org/gitlab), create a workspace, and wait until the remote is ahead - it will get slower and slower presumably the more commits the remote is ahead.
It should be possible to unzip the following archive into `.git/gitbutler` for the same effect, maybe after fiddling with the local branches. I copied them here as well.

* [dot-git-gitbutler.zip](https://github.com/user-attachments/files/17395195/dot-git-gitbutler.zip)
* [dot-git-refs-and-index.zip](https://github.com/user-attachments/files/17395225/dot-git-refs-and-index.zip)

### Expected behavior

The UI be fast enough to work on a repo of that size.

nir-tsabari:
Having similar issues, very large repo, initial load takes long, 500 commits can't be handled, and sometimes even 30 commits can't be handled.
Github, Windows x64, latest version

Byron:
I hope that this weekend I have some new numbers to share here :).

[...]

PR #5411, opened 2024-11-02
use gitoxide for merging trees

Byron:
An attempt to improve performance by integrating `gix::Repository::merge_trees()`.

In a single sample that didn't actually have merge conflicts, it was 1.8x of merge-ORT.

```
❯ hyperfine -w3 -N 'git merge-tree --write-tree diff-to-suggestion-spike master' '/Users/byron/dev/github.com/Byron/gitoxide/target/release/gix merge tree diff-to-suggestion-spike master master'
Benchmark 1: git merge-tree --write-tree diff-to-suggestion-spike master
  Time (mean ± σ):     287.7 ms ±   3.8 ms    [User: 255.3 ms, System: 27.3 ms]
  Range (min … max):   282.5 ms … 294.4 ms    10 runs

Benchmark 2: /Users/byron/dev/github.com/Byron/gitoxide/target/release/gix merge tree diff-to-suggestion-spike master master
  Time (mean ± σ):     158.7 ms ±   3.1 ms    [User: 130.6 ms, System: 30.5 ms]
  Range (min … max):   153.1 ms … 164.6 ms    18 runs

Summary
  /Users/byron/dev/github.com/Byron/gitoxide/target/release/gix merge tree diff-to-suggestion-spike master master ran
    1.81 ± 0.04 times faster than git merge-tree --write-tree diff-to-suggestion-spike master
```

GitHub wrote about changing rebases from `git2` to `git` with `merge-ORT`:

> Extrapolating those numbers out to 100%, if we had done all rebases during that interval with merge-ort, it would have taken us 2,000 minutes, or about 33 hours. That same work done with libgit2 would have taken 512 hours!
Merge-ORT seems to be 15.5x faster for GitHub in average - they mention rebases, but merges certainly apply similarly.

So in theory, we could see a 15x * 1.8x = ~28x speedup in a best-case scenario.

~~However, this much improvement is probably not going to happen... let's see.~~ - Actually it did happen, 28x on the first day, but it took a little tuning.

Related to #5163. 

### Tasks

* [x] use `gitoxide` for `is_integrated()`
* [x] evaluate performance results.
* [x] try to get better.
* [x] try to get even better

### As follow-up

* ~~try to remove duplication of work in `stack_series()`.~~ - This will happen naturally as the other path is deprecated.
* [ ] use `merge_trees` in remaining places
    - but try to assure there is a test-case for it to be able to profile it a little.

### Data so far

* **86min** baseline
* **65min** the same with gitoxide
* **4min36s** by stopping immediately when the commit fails.
* **4min16s** by avoiding some more duplicate work (but it's also the first run purely through instruments.
* **3min7s** by avoiding unnecessary hashing in a tight loop

### Stuff that could still be faster

* Can the stack-series work be omitted somehow?
    - I don't understand where these commits are coming from. Are they truly 'new'? Or is it checking different integration targets? These seem potentially different, but the `IsIntegrated` state isn't adjusted.
    - Should this not be free given that my repo has no stacks at all?
* `list_virtual_commit_files()` could be using `gitoxide` to get faster tree-diffing and faster hunk generation. This needs time though as it's basically rewriting a core of GB, so should be its own PR.

#### Performance log before any change

List virtual branches takes 86 minutes!

```
INFO     ｉ [info]: system git executable for fetch/push: "git"
INFO     ｉ [info]: starting app | version: 0.13.8 | name: GitButler
2024-11-02 15:32:36.824 GitButler[55418:10356652] +[IMKClient subclass]: chose IMKClient_Modern
INFO     list_projects [ 328µs | 100.00% ]
2024-11-02 15:32:36.892 GitButler[55418:10356652] +[IMKInputSession subclass]: chose IMKInputSession_Modern
INFO     get_user [ 42.1ms | 100.00% ]
INFO     get_project [ 840µs | 100.00% ] id: a4dfef20-d58e-47d8-984d-9f78eb012c2a | no_validation: None
INFO     git_head [ 1.42ms | 100.00% ] project_id: a4dfef20-d58e-47d8-984d-9f78eb012c2a
INFO     operating_mode [ 462µs | 100.00% ] project_id: a4dfef20-d58e-47d8-984d-9f78eb012c2a
INFO     list_branches [ 216ms | 100.00% ] project_id: a4dfef20-d58e-47d8-984d-9f78eb012c2a | filter: Some(BranchListingFilter { local: None, applied: None })
INFO     get_base_branch_data [ 386ms | 0.03% / 100.00% ] project_id: a4dfef20-d58e-47d8-984d-9f78eb012c2a
INFO     ┕━ get_base_branch_data [ 386ms | 99.97% ]
INFO     list_projects [ 119µs | 100.00% ]
INFO     get_base_branch_data [ 408ms | 0.04% / 100.00% ] project_id: a4dfef20-d58e-47d8-984d-9f78eb012c2a
INFO     ┕━ get_base_branch_data [ 408ms | 99.96% ]
INFO     list_local_branches [ 129ms | 100.00% ] project_id: a4dfef20-d58e-47d8-984d-9f78eb012c2a
INFO     list_branches [ 233ms | 100.00% ] project_id: a4dfef20-d58e-47d8-984d-9f78eb012c2a | filter: Some(BranchListingFilter { local: None, applied: None })
INFO     fetch_from_remotes [ 2.95s | 85.34% / 100.00% ] project_id: a4dfef20-d58e-47d8-984d-9f78eb012c2a | action: Some("auto")
INFO     ┕━ get_base_branch_data [ 432ms | 14.66% ]
INFO     get_branch_listing_details [ 2.46s | 100.00% ] project_id: a4dfef20-d58e-47d8-984d-9f78eb012c2a | branch_names: ["Virtual-branch", "requeue-milestone-releases-backfill", "vr-aigw-service-class", "lw/488887-allow-disabling-sidekiq-workers-via-configuration"]
INFO     handle [ 150µs | 100.00% ] event: GitFileChange(a4dfef20-d58e-47d8-984d-9f78eb012c2a, FETCH_HEAD)
INFO     get_base_branch_data [ 383ms | 0.04% / 100.00% ] project_id: a4dfef20-d58e-47d8-984d-9f78eb012c2a
INFO     ┕━ get_base_branch_data [ 383ms | 99.96% ]
[..]
INFO     fetch_from_remotes [ 2.90s | 86.93% / 100.00% ] project_id: a4dfef20-d58e-47d8-984d-9f78eb012c2a | action: Some("auto")
INFO     ┕━ get_base_branch_data [ 380ms | 13.07% ]
[..]
INFO     list_virtual_branches [ 5169s | 100.00% ] project_id: a4dfef20-d58e-47d8-984d-9f78eb012c2a
INFO     list_branches [ 255ms | 100.00% ] project_id: a4dfef20-d58e-47d8-984d-9f78eb012c2a | filter: Some(BranchListingFilter { local: None, applied: None })
```

Here we see that it takes 3m15s to get the applied status, but the rest of the time is taken to do a lot of merges. In the screenshot, I stopped the trace after about an hour.

Screenshot 2024-11-02 at 16 52 58

#### Performance after the change

TBD … but one thing is for sure - it's not 28 times faster (saying this with the CLI running for 16 minutes already with everything configured correctly, I think).

It clocked in at **1h5m** which is still an improvement.

A partial profile I took while it was running reveals that the reason for the incredible increase in time consumption is also that the commit-integration check is now **run twice: Once for the stack, once in the previous context**. However, I didn't investigate if it's doing something new or if it's a repetition.

Screenshot 2024-11-02 at 18 33 36

When it comes to the time spent on the actual merges, we see that only about 1/30th is spent on diffing trees and performing tree-edits.

Screenshot 2024-11-02 at 18 39 07

The rest goes into blob-merges, where half the time is used to read blobs from the ODB, the other half is spent on diffing them. Once again, it only takes about 1/30th of the time to actually perform the merge.

Screenshot 2024-11-02 at 18 44 09

Another fun-fact: one merge operation takes about **3.5s**!

#### The best work is the work you don't have to do ;)

This time we told the merge operation to stop as soon as there is an unresolved conflict. This saves tremendous amounts of work and the total time is now down to **4m36s**!

Now we have the commit-integrated checks times 2 like before, each of them costing nearly 2m, with 1m for getting the status.

Screenshot 2024-11-02 at 18 59 31

And there we see that it's mostly hunk conversions and the computation of an MD5 hash.

Screenshot 2024-11-02 at 19 04 13

Unfortunately the trace is cut-off so, as the hunks have to be computed as well.

#### Try to re-use 'is-integrated' work

Under the assumption that in my case, the stacks should not be anything new, then the commits it finds should mostly (?) be coming from the existing branches. We now try to avoid an actual integration check by looking up the commit (non-optimized) in a list of pre-computed virtual commits.

Screenshot 2024-11-02 at 19 29 53

This brings down the time to **4m08** in total, and 1m12s for the stack series function. Note that this value also looks better because there is no tracing, and the output is faster, but it still takes 16s to debug print a huge structure.

#### Avoid hashing in favor of intersecting first

From my possibly flawed logic, it should be enough to roughly prune out hunks by intersections, and then do the expensive hashing. This cut down the time by another minute to **3m 7s**.

Screenshot 2024-11-02 at 20 25 17

Now the time to run the `is_integrated()` checks is down to just 24s, but the most time, 1m10s times 2, is spent listing virtual commit files and diffing trees. This can probably be done faster, while trying to avoid the duplication in the stack-series more.

### DIFF
--- a/crates/gitbutler-branch-actions/src/stack.rs
+++ b/crates/gitbutler-branch-actions/src/stack.rs
@@ -7,6 +7,7 @@ use gitbutler_oplog::entry::{OperationKind, SnapshotDetails};
 use gitbutler_oplog::{OplogExt, SnapshotExt};
 use gitbutler_project::Project;
 use gitbutler_reference::normalize_branch_name;
+use gitbutler_repo::GixRepositoryExt;
 use gitbutler_repo_actions::RepoActionsExt;
 use gitbutler_stack::{
     CommitOrChangeId, ForgeIdentifier, PatchReference, PatchReferenceUpdate, Series,
@@ -191,14 +192,20 @@ pub fn push_stack(project: &Project, branch_id: StackId, with_force: bool) -> Re
 
     // First fetch, because we dont want to push integrated series
     ctx.fetch(&default_target.push_remote_name(), None)?;
-    let check_commit = IsCommitIntegrated::new(ctx, &default_target)?;
+    let gix_repo = ctx
+        .gix_repository()?
+        .for_tree_diffing()?
+        .with_object_memory();
+    let cache = gix_repo.commit_graph_if_enabled()?;
+    let mut graph = gix_repo.revision_graph(cache.as_ref());
+    let mut check_commit = IsCommitIntegrated::new(ctx, &default_target, &gix_repo, &mut graph)?;
     let stack_series = stack.list_series(ctx)?;
     for series in stack_series {
         if series.head.target == merge_base {
             // Nothing to push for this one
             continue;
         }
-        if series_integrated(&check_commit, &series)? {
+        if series_integrated(&mut check_commit, &series)? {
             // Already integrated, nothing to push
             continue;
         }
@@ -214,7 +221,7 @@ pub fn push_stack(project: &Project, branch_id: StackId, with_force: bool) -> Re
     Ok(())
 }
 
-fn series_integrated(check_commit: &IsCommitIntegrated, series: &Series) -> Result<bool> {
+fn series_integrated(check_commit: &mut IsCommitIntegrated, series: &Series) -> Result<bool> {
     let mut is_integrated = false;
     for commit in series.clone().local_commits.iter().rev() {
         if !is_integrated {
@@ -226,12 +233,14 @@ fn series_integrated(check_commit: &IsCommitIntegrated, series: &Series) -> Resu
 
 /// Returns the stack series for the API.
 /// Newest first, oldest last in the list
+/// `commits` is used to accelerate the is-integrated check.
 pub(crate) fn stack_series(
     ctx: &CommandContext,
     branch: &mut Stack,
     default_target: &Target,
-    check_commit: &IsCommitIntegrated,
+    check_commit: &mut IsCommitIntegrated,
     remote_commit_data: HashMap<CommitData, git2::Oid>,
+    commits: &[VirtualBranchCommit],
 ) -> Result<(Vec<PatchSeries>, bool)> {
     let mut requires_force = false;
     let mut api_series: Vec<PatchSeries> = vec![];
@@ -248,7 +257,10 @@ pub(crate) fn stack_series(
         // Reverse first instead of later, so that we catch the first integrated commit
         for commit in series.clone().local_commits.iter().rev() {
             if !is_integrated {
-                is_integrated = check_commit.is_integrated(commit)?;
+                is_integrated = commits
+                    .iter()
+                    .find_map(|c| (c.id == commit.id()).then_some(Ok(c.is_integrated)))
+                    .unwrap_or_else(|| check_commit.is_integrated(commit))?;
             }
             let copied_from_remote_id = CommitData::try_from(commit)
                 .ok()
--- a/crates/gitbutler-branch-actions/src/status.rs
+++ b/crates/gitbutler-branch-actions/src/status.rs
@@ -131,9 +131,7 @@ pub fn get_applied_status_cached(
                     .filter_map(|claimed_hunk| {
                         // if any of the current hunks intersects with the owned hunk, we want to keep it
                         for (i, git_diff_hunk) in git_diff_hunks.iter().enumerate() {
-                            if claimed_hunk == &Hunk::from(git_diff_hunk)
-                                || claimed_hunk.intersects(git_diff_hunk)
-                            {
+                            if claimed_hunk.intersects(git_diff_hunk) {
                                 let hash = Hunk::hash_diff(&git_diff_hunk.diff_lines);
                                 if locks.contains_key(&hash) {
                                     return None; // Defer allocation to unclaimed hunks processing
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3059,6 +3059,7 @@ dependencies = [
  "gix-ignore 0.12.0",
  "gix-index 0.36.0",
  "gix-lock 15.0.0",
+ "gix-merge",
  "gix-negotiate",
  "gix-object 0.45.0",
  "gix-odb",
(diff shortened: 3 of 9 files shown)


crates/gitbutler-branch-actions/src/status.rs
```diff
@@ -131,9 +131,7 @@ pub fn get_applied_status_cached(
                     .filter_map(|claimed_hunk| {
                         // if any of the current hunks intersects with the owned hunk, we want to keep it
                         for (i, git_diff_hunk) in git_diff_hunks.iter().enumerate() {
-                            if claimed_hunk == &Hunk::from(git_diff_hunk)
-                                || claimed_hunk.intersects(git_diff_hunk)
-                            {
+                            if claimed_hunk.intersects(git_diff_hunk) {
```
Byron, line 134: Please double-check this logic. I think it's the same as before, but avoids a ton of hashing that would happen otherwise and slow things down to a crawl.

crates/gitbutler-branch-actions/src/stack.rs
```diff
@@ -248,7 +257,10 @@ pub(crate) fn stack_series(
         // Reverse first instead of later, so that we catch the first integrated commit
         for commit in series.clone().local_commits.iter().rev() {
             if !is_integrated {
-                is_integrated = check_commit.is_integrated(commit)?;
+                is_integrated = commits
```
Byron, line 260: This seems to save time and be reasonable in comparison to what it was before. However, since the stack can have its own remote I am not sure if the `check_commit` instance is configured correctly.

crates/gitbutler-branch-actions/src/stack.rs
```diff
@@ -226,12 +233,14 @@ fn series_integrated(check_commit: &IsCommitIntegrated, series: &Series) -> Resu
...
 /// Newest first, oldest last in the list
+/// `commits` is used to accelerate the is-integrated check.
 pub(crate) fn stack_series(
     ctx: &CommandContext,
     branch: &mut Stack,
     default_target: &Target,
-    check_commit: &IsCommitIntegrated,
+    check_commit: &mut IsCommitIntegrated,
     remote_commit_data: HashMap<CommitData, git2::Oid>,
+    commits: &[VirtualBranchCommit],
```
krlvi, line 243: In general this is probably fine - but the commits that are passed in here, are something deprecated that I plan to remove asap

krlvi (approved):
imo lets just merge it and evaluate

crates/gitbutler-branch-actions/src/stack.rs
Byron: I am looking forward to seeing that work only done once, and what I understood is that `stack_series` is the new version, and I was basically altering the old one. The few changes I made should be easy to port to be internal to `stack_series()` (if this is what's needed).

PR #5416, opened 2024-11-03
Replace `merge-tree` in more places

Byron:
The `gitoxide` merge-tree implementation is faster and better in quality, so it should be used where applicable. This PR attempts to find these places and make the switch.

The goal is also to provide performance metrics, even though it's not expected we will see great numbers here as merge-tree is a smaller portion of the total work to be done.

Follow-up on #5411 .

### Tasks

Use `merge-tree` in…

* [x] Apply and Unapply branch
* [x] unapply ownership
* [x] back to integration
* [x] Oplog Listing (3x faster, 3,5x for merge-trees)
* [x] checkout branch trees
* [x] upstream integration statuses
* [x] use latest `gix` for nicer API
* [x] be sure to switch `gix` to the one from `main`

### Notes for the Reviewer

* All uses that deal with `resolve_index()` have been postponed for now, as it's probably easiest to implement this directly in the `gitoxide` merge tree function, for achieving the best possible quality. In theory, it might already work doing a post-process on the conflict entries, but that would be harder to get right.
* There are a couple of other uses in `gitbutler-cherry-pick` and `gitbutler-edit-mode` that rely on the index quite a bit, and even checkout an index. But it *seems* that [it doesn't rely on the index to show merge-conflicts](https://github.com/gitbutlerapp/gitbutler/blob/068833059d6dbaad32b6aebe4d6863ad55333a38/crates/gitbutler-edit-mode/src/lib.rs#L102).

### Follow-Up

* Improve `gix` merge trees APIs, see https://github.com/GitoxideLabs/gitoxide/pull/1651
* Use `gix-merge` in all the remaining places for performance and merge-quality

#### Before

When *applying*, merge-trees takes a whopping **72ms**. It's clearly not the bottleneck, with a lot of time being spent in checkout.

Screenshot 2024-11-03 at 14 22 36

When *unapplying*, it now clocks in at **139ms**, so once again isn't going to change anything when replaced. Checking out files takes a long time, comparatively.

Screenshot 2024-11-03 at 14 24 46

#### After

Like expected, there isn't much of a relevant change, with both merges during unapply and apply clocking in to ~90ms. Yes, it *seems*  a bit faster, but is still far from flexing any muscle.

Screenshot 2024-11-03 at 16 05 34

#### Before 

A tiny portion of 66ms of the 4s runtime is related to merging trees. Since it's the same repository  (GitLab) the after should look very similar.

Screenshot 2024-11-03 at 17 38 25

#### After

It's basically the same at 64ms.

Screenshot 2024-11-03 at 17 38 59''

#### Before

**112ms**

Screenshot 2024-11-03 at 19 30 17

#### After

**93ms**

Screenshot 2024-11-03 at 19 34 21

#### Before

Listing takes 22s, and one full diff takes 7s, which is duplicated for an additional stats. Merging of trees takes **1.93s**.

Screenshot 2024-11-03 at 19 52 45

#### After replacing merge-trees

Merging trees now take **585ms**,  3.3x of what was before.

Screenshot 2024-11-03 at 20 28 31

#### After using `gitoxide` for the diffing

The whole operation is down to 7.3s, a 3x improvement! This is also due to not duplicating work.

Screenshot 2024-11-04 at 08 14 31

It's noticeable how the TOML deserialisation takes 2s, which is a lot in comparison to the total runtime. But like I mentioned on Discord, I think there is ways to speed this up by not precomputing everything, instead of parallelising.

#### Before

Merging in total takes about **300ms** with two trivial branches. Like noticed before, writing the index to a tree is very expensive.

Screenshot 2024-11-04 at 11 04 06

#### After

Now the same task takes **140ms** for merging and writing the tree. Notably, writing the tree now takes 6ms , 29x faster than having to write a whole index.

Screenshot 2024-11-04 at 12 13 51

#### Before

As often, most time is spent elsewhere (and duplicated in `stack_series()`), but that aside we are looking at **1.09s** for merging trees in that function.

Screenshot 2024-11-04 at 16 11 10

#### After

Now it clocks in at 640ms, at 1.7x.

Screenshot 2024-11-04 at 19 21 50

### DIFF
--- a/crates/gitbutler-oplog/src/oplog.rs
+++ b/crates/gitbutler-oplog/src/oplog.rs
@@ -6,24 +6,25 @@ use std::{
     time::Duration,
 };
 
-use anyhow::{anyhow, bail, Context, Result};
-use git2::{DiffOptions, FileMode};
-use gitbutler_command_context::RepositoryExtLite;
-use gitbutler_diff::{hunks_by_filepath, FileDiff};
-use gitbutler_project::{
-    access::{WorktreeReadPermission, WorktreeWritePermission},
-    Project,
-};
-use gitbutler_repo::RepositoryExt;
-use gitbutler_repo::SignaturePurpose;
-use gitbutler_stack::{Stack, VirtualBranchesHandle, VirtualBranchesState};
-use tracing::instrument;
-
 use super::{
     entry::{OperationKind, Snapshot, SnapshotDetails, Trailer},
     reflog::set_reference_to_oplog,
     state::OplogHandle,
 };
+use anyhow::{anyhow, bail, Context, Result};
+use git2::{DiffOptions, FileMode};
+use gitbutler_command_context::RepositoryExtLite;
+use gitbutler_diff::{hunks_by_filepath, FileDiff};
+use gitbutler_oxidize::{git2_to_gix_object_id, gix_to_git2_oid};
+use gitbutler_project::{
+    access::{WorktreeReadPermission, WorktreeWritePermission},
+    Project,
+};
+use gitbutler_repo::SignaturePurpose;
+use gitbutler_repo::{GixRepositoryExt, RepositoryExt};
+use gitbutler_stack::{Stack, VirtualBranchesHandle, VirtualBranchesState};
+use gix::prelude::Write;
+use tracing::instrument;
 
 const SNAPSHOT_FILE_LIMIT_BYTES: u64 = 32 * 1024 * 1024;
 
@@ -165,8 +166,9 @@ impl OplogExt for Project {
         limit: usize,
         oplog_commit_id: Option<git2::Oid>,
     ) -> Result<Vec<Snapshot>> {
-        let repo_path = self.path.as_path();
-        let repo = git2::Repository::open(repo_path)?;
+        let worktree_dir = self.path.as_path();
+        let repo = git2::Repository::open(worktree_dir)?;
+        let gix_repo = gitbutler_command_context::gix_repository_for_merging(worktree_dir)?;
 
         let traversal_root_id = match oplog_commit_id {
             Some(id) => id,
@@ -208,7 +210,7 @@ impl OplogExt for Project {
             }
 
             // Get tree id from cache or calculate it
-            let wd_tree = get_workdir_tree(&mut wd_trees_cache, commit_id, &repo)?;
+            let wd_tree = get_workdir_tree(&mut wd_trees_cache, commit_id, &repo, &gix_repo)?;
 
             let details = commit
                 .message()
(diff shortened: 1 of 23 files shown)


crates/gitbutler-oplog/src/oplog.rs
```diff
@@ -178,68 +181,86 @@ impl OplogExt for Project {
...
+            let commit_time = gix_time_to_git2(commit.time()?);
 
-            if let Ok(parent) = commit.parent(0) {
+            if let Some(parent_id) = first_parent {
                 // Get tree id from cache or calculate it
-                let parent_tree = get_workdir_tree(&mut wd_trees_cache, parent.id(), &repo)?;
-
-                let mut opts = DiffOptions::new();
-                opts.include_untracked(true);
-                opts.ignore_submodules(true);
```
Byron, line 223: `gitoxide` has no option to ignore submodules, but it would conflict on them if both change to different values. It's planned to have similar resolution facilities like Git has, but it's not implemented yet.

crates/gitbutler-oplog/src/oplog.rs
```diff
@@ -178,68 +181,86 @@ impl OplogExt for Project {
...
+                let (mut lines_added, mut lines_removed) = (0, 0);
+                let parent_tree = get_workdir_tree(&mut wd_trees_cache, parent_id, &gix_repo)?;
+                parent_tree
+                    .changes()?
+                    .options(|opts| {
+                        opts.track_rewrites(None).track_path();
+                    })
+                    .for_each_to_obtain_tree(&wd_tree, |change| -> Result<_> {
+                        match change {
+                            Change::Addition { location, .. } => {
```
Byron, line 237: I am not 100% sure that this is what was intended in the `git2` implementation, or why that is.

In the GItLab repo, I see 464 files changed in a discard hunk snapshot, but that was 4.6k if any location is used. Maybe this is why this 'add-only' filter is applied here. The UI gets quite slow displaying this.

crates/gitbutler-oplog/src/oplog.rs
```diff
@@ -178,68 +181,86 @@ impl OplogExt for Project {
...
+                                files_changed.push(gix::path::from_bstr(location).into_owned());
+                            }
+                            Change::Deletion { .. }
+                            | Change::Modification { .. }
+                            | Change::Rewrite { .. } => {}
+                        }
+                        if let Some(counts) = change
```
Byron, line 244: This is the expensive operation, of 3s, half goes into calculating line numbers. If performance ever becomes a bigger problem, these diffs should be avoided and be done on demand entirely.

crates/gitbutler-testsupport/src/testing_repository.rs
```diff
@@ -15,6 +15,21 @@ impl TestingRepository {
     pub fn open() -> Self {
         let tempdir = tempdir().unwrap();
         let repository = git2::Repository::init_opts(tempdir.path(), &init_opts()).unwrap();
+        // TODO(ST): remove this once `gix::Repository::index_or_load_from_tree_or_empty()`
```
Byron, line 18: `gitoxide` will soon get an upgrade so this won't be an issue in future. For now it's easiest to cater to it and avoid an unborn `HEAD`.

krlvi (approved):
Looks good to me! Let's merge it in a couple of hours, when we know that the newly released `0.13.9` is okay :)

PR #5418, opened 2024-11-03
fix: ensure error message text wraps

MikaelCarpenter:
## ☕️ Reasoning

The error message is not fully visible when a commit signing test fails.

## 🧢 Changes

I moved the error message into the `content` slot, because that seems to be more in-line with how the `InfoMessage` component is used throughout the codebase. I also had to make sure the errorMessage was re-set when a new signing test was initiated. And then i added some styles to ensure that a `pre` in the content slot would always have text wrapping.

If y'all would prefer a different approach, just let me know!

## 📸 Screenshots

|Before|After|
|-----|-----|
|prod_before|Screenshot 2024-11-04 at 8 24 37 AM|

### DIFF
--- a/apps/desktop/src/lib/settings/userPreferences/CommitSigningForm.svelte
+++ b/apps/desktop/src/lib/settings/userPreferences/CommitSigningForm.svelte
@@ -52,6 +52,7 @@
 	let errorMessage = '';
 
 	async function checkSigning() {
+		errorMessage = '';
 		checked = true;
 		loading = true;
 		await invoke('check_signing_settings', { id: project.id })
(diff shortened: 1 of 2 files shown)


apps/desktop/src/lib/settings/userPreferences/CommitSigningForm.svelte
```diff
@@ -159,9 +160,12 @@
 							<p>Signing is working correctly</p>
 						{:else}
 							<p>Signing is not working correctly</p>
-							<pre>{errorMessage}</pre>
 						{/if}
 					</svelte:fragment>
+
+					<svelte:fragment slot="content">
+						{errorMessage}
```
ndom91: Let's wrap this in a  `<pre>` as well. Otherwise looking good though!

ndom91 (approved):
LGTM, thanks!

PR #5421, opened 2024-11-04
Remove files from VirtualBranchCommit struct

mtsgrd:
- loaded separately in the client


mtsgrd:
@Byron I think maybe you and @krlvi discussed this over the weekend? I'm unsure if there was more related duplication we can remove, but this had been on my mind.

krlvi (approved):

PR #5422, opened 2024-11-04
Fix performance problem with file drag & drop

mtsgrd:
- fixes manual use of a svelte action
- pass store to DraggableFile instead of updating selection list
- refactor file selection


ndom91 (approved):

apps/desktop/src/lib/file/FileListItem.svelte
```diff
@@ -45,22 +46,13 @@
...
-			return [file];
-		}
-	});
+	const draggable = !readonly && !isUnapplied;
 
 	let contextMenu = $state<ReturnType<typeof FileContextMenu>>();
-
 	let draggableEl: HTMLDivElement | undefined = $state();
-	let checked = $state(false);
+	let draggableFile: DraggableFile | undefined;
```
Caleb-T-Owens: Why does it make sense to declare this here? I don't see it used outside the effect. If it is used outside the effect, should it be a `$state()`?

mtsgrd: Thanks, that was a mistake, the variable is not needed.

apps/desktop/src/lib/file/FileListItem.svelte
```diff
@@ -109,10 +116,11 @@
...
 	onDestroy(() => {
+		chips?.destroy();
 		if (draggableEl && animationEndHandler) {
 			draggableEl.removeEventListener('animationend', animationEndHandler);
 		}
-		draggableFiles.forEach((f) => {
+		$selectedFiles.forEach((f) => {
 			const lockedElement = document.getElementById(`file-${f.id}`);
 			if (lockedElement && animationEndHandler) {
 				lockedElement.removeEventListener('animationend', animationEndHandler);
```
Caleb-T-Owens, line 126: This doesn't guarantee that we remove all the event listeners. We ought to record all the elements that have the event added to them, and then iterate through that set and remove the event listeners added to those elements

apps/desktop/src/lib/vbranches/fileIdSelection.ts
```diff
@@ -26,129 +29,173 @@ export function parseFileKey(fileKeyString: string): FileKey {
...
-	clear() {
-		this.value = [];
-		this.emit();
+	add(file: AnyFile, commitId?: string) {
+		this.select_many([file], commitId);
 	}
 
-	clearExcept(fileId: string, commitId?: string) {
-		this.value = [stringifyFileKey(fileId, commitId)];
+	select_many(files: AnyFile[], commitId?: string) {
```
Caleb-T-Owens, line 139: `selectMany`

apps/desktop/src/lib/dragging/draggable.ts
```diff
@@ -175,6 +175,7 @@ function setupDragHandlers(
 		node.removeEventListener('dragstart', handleDragStart);
 		node.removeEventListener('drag', handleDrag);
 		node.removeEventListener('dragend', handleDragEnd);
+		node.removeEventListener('mousedown', handleMouseDown, { capture: false });
```
Caleb-T-Owens, line 178: Is this the main offending leaky event listener you were mentioning?

Caleb-T-Owens (changes requested):

PR #5428, opened 2024-11-04
fix: Hunk dependencies account for complete overwrite

estib-vega:
Handle the computation of dependencies when a patch overrides a hunk completely.
Also handle the cases of determining the intersection of changes when deleting complete files

### DIFF
--- a/crates/gitbutler-hunk-dependency/src/hunk.rs
+++ b/crates/gitbutler-hunk-dependency/src/hunk.rs
@@ -14,7 +14,11 @@ pub struct HunkRange {
 impl HunkRange {
     pub fn intersects(&self, start: u32, lines: u32) -> bool {
         if self.lines == 0 {
-            // Special case for checking if a point is inside a range, happens
+            if self.start == 0 {
+                // Special case when file is deleted.
+                return true;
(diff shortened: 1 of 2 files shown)


crates/gitbutler-hunk-dependency/src/hunk.rs
```diff
@@ -14,6 +16,11 @@ pub struct HunkRange {
 impl HunkRange {
     pub fn intersects(&self, start: u32, lines: u32) -> bool {
         if self.lines == 0 {
+            if self.start == 0 && self.line_shift <= 0 {
+                // File completely removed
+                return true;
+            }
+            println!("line shift: {}", self.line_shift);
```
Byron: If printing is desired and it shouldn't be `dbg!()` for formatting reasons, one should always use `eprintln()` so the output goes to stderr (unbuffered), which is typically more instantaneous.

estib-vega: @Byron Oh that shouldn't be there at all!
Thanks for pointing it out :)

mtsgrd (approved):

PR #5437, opened 2024-11-05
New stack header

PavelLaptev:
## ☕️ Reasoning

image

## 🧢 Changes
- Add new stack header
- Modified the `Selector` component to pass custom selector button and list items inside
- Small fixes


apps/desktop/src/lib/stack/CurrentSeries.svelte
```diff
@@ -39,9 +39,25 @@
 	);
 	const forgeChecksMonitorStore = createForgeChecksMonitorStore(undefined);
 	$effect(() => forgeChecksMonitorStore.set(checksMonitor));
+
+	const seriesType = currentSeries.patches[0] ? currentSeries.patches[0].status : 'local';
```
ndom91, line 44: I have a helper for this actually since we need to look it up in a few places already. See: `apps/desktop/src/lib/branch/stackingUtils.ts`

ndom91 (approved):
Small recommendation, otherwise LGTM though :partying_face:

apps/desktop/src/lib/stack/CurrentSeries.svelte
PavelLaptev: Thanks! Updated with your function

PR #5442, opened 2024-11-05
fix: add tauri localhost entries to font-src CSP rules

ndom91:
## ☕️ Reasoning

- TEST: Branch to test build on windows
- Windows fails loading Inter fonts

image


ndom91:
This does NOT fix the issue with loading fonts in the Windows client

PR #5471, opened 2024-11-06
Set default branch no wiggle animation

krlvi:
It was looking weird with the new desing, wiggling sideways?


PavelLaptev:
Sense we will remove this in v3, I wouldn't try something different, I think there could be a different animation that could work. But let's just remove it if you think it's better without it, just to save some time.

PavelLaptev (approved):

PR #5413, opened 2024-11-02
Add a stack overview section to pr descriptions

mtsgrd:
When stacking pull requests we append/update a stack list when creating a new pull request, and the total count is >= 2. On deleting a branch from a stack we close any associated pull request, and update the descriptions to exclude it.

This is part of a stack made with GitButler:
- #5413 👈 
- #5475

PR #5475, opened 2024-11-06
Replace ForgeIdentifier with string type

mtsgrd:
Simplifying to keep things simple in the front end

This is part of a stack made with GitButler:
- #5413 
- #5475 👈

### DIFF
--- a/crates/gitbutler-stack/src/patch_reference.rs
+++ b/crates/gitbutler-stack/src/patch_reference.rs
@@ -22,29 +22,13 @@ pub struct Branch {
     pub name: String,
     /// Optional description of the series. This could be markdown or anything our hearts desire.
     pub description: Option<String>,
-    /// An identifier for a review unit at a forge (eg. GitHub Pull Request number).
-    /// None if is no review unit, eg. no Pull Request has been created.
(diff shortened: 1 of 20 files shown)


crates/gitbutler-stack/src/patch_reference.rs
```diff
@@ -25,28 +25,13 @@ pub struct Branch {
     /// An identifier for a review unit at a forge (eg. GitHub Pull Request number).
     /// None if is no review unit, eg. no Pull Request has been created.
     #[serde(default)]
-    pub forge_id: Option<ForgeIdentifier>,
+    pub pull_request_id: Option<i32>,
```
krlvi: Perhaps update the docs above - is it a PR number or a PR id? The previous type had PR number. The GitHub api has both... and i always mix them up. I think the PR number needs a repo as well, while the PR ID is more globally unique or something. 

Also - you probably want `useize` not i32

krlvi (dismissed):

krlvi (approved):
nvm. i initially though about deserializing the value from the old type to the new type but just realised that since the name is different it wont be a problem. 
i should get some sleep

PR #5478, opened 2024-11-07
`hasconfig` support in `gitoxide` 

### LINKED ISSUE #5420: GitButler not using includeIf configurations
### Version

0.13.8

### Operating System

macOS

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

GitButler doesn't appear to use git configurations that are dynamically sourced in via an `includeIf`. I don't have the following configuration set it my default `~/.gitconfig`, rather, they're sourced in depending on the remote that I'm working on.

- `user.name`
- `user.email`
- `user.signingkey`
- `gpg.ssh.allowedsignersfile`

My global `~/.gitconfig` has entries like this, which then sets the respective configuration items.

```
includeif.hasconfig:remote.*.url:git@github.com:*/**.path=~/.dotfiles/git/gitconfig-personal
includeif.hasconfig:remote.*.url:git@ghe.work:*/**.path=~/.dotfiles/git/gitconfig-work
```

I get the following error when trying to commit with the above configuration.

image

However, if I do set those variables in either the global or repository configurations, then it works fine.

### How to reproduce

As a minimum, remove the following from your global configuration, and set them via an `includeIf` directive.

- `user.name`
- `user.email`

Use GitButler to make a commit to reproduce the error.

### Expected behavior

GitButler should use any values sourced in via an `includeIf`, just like `git` currently does. It allows the user to set different names, emails, signatures, etc... depending on the `includeIf` test conditions.

### Relevant log output

_No response_

Byron:
This adds support for `includeIf` directives with the `hasconfig` condition,
something commonly used for user information which affects commits in GitButler.

Fixes #5420.

### Tasks

* [x] `gitoxide` update

### Notes for the Reviewer

This was validated via `gix` test, and by testing it in GitButler specifically.
To do that, the global configuration was changed to use an `[includeIf "hasconfig:remote.*.url:<url-to-match>"]` directive whose path then points to a new git configuration file that sets the `user.name` and `user.email` to a marker value. This value can then be observed when opening a repository whose URL matches in GitButler, and creating a commit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3039,7 +3039,7 @@ dependencies = [
 [[package]]
 name = "gix"
 version = "0.67.0"
-source = "git+https://github.com/Byron/gitoxide?rev=697a6320c7664845590e3e8251015085b6cc5d81#697a6320c7664845590e3e8251015085b6cc5d81"
+source = "git+https://github.com/Byron/gitoxide?rev=c5955fc4ad1064c7e4b4c57de32a661e693fbe49#c5955fc4ad1064c7e4b4c57de32a661e693fbe49"
 dependencies = [
  "gix-actor 0.33.0",
  "gix-attributes 0.23.0",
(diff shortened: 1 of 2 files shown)


krlvi (approved):

krlvi:
thanks

PR #5485, opened 2024-11-07
Unapply stacks to a unique uuid

Caleb-T-Owens:
It's very important that we unapply to a unique ID as otherwise we risk overwriting other unapplied branches, because at the minute, they're all called "lane" or "lane-2" ect...

This is "ugly" but we can worry about making it pretty later


krlvi (approved):
thanks

PR #5490, opened 2024-11-07
use `merge_base_octopus()` 

Byron:
This PR is the result of a learning experience that ultimately led me here.
As a result, `gitoxide` now provides `Repository::merge_base_octopus()` which will be a little bit faster on its own, but can be substantially faster when used with a re-used `graph` data structure.

Even though this PR doesn't do `graph` re-use yet, the more `gitoxide` is used, the more likely it is that this becomes possible.

### Tasks

* [x] update `gitoxide` to access the latest features
* [x] remove the second to last usage of `merge_base_octopussy()`
   - The last usage is in a spot where `l()` is used, and ideally that will also have a `gitoxide` variant so the whole function can be converted as a whole.

### Notes for the Reviewer

* There is also a way to create virtual merge bases directly which allows to use all the most advanced features of `merge-ORT` here, but if octopus is used for all merge-bases, it turns out to have no relevance.
  Part of me still thinks that maybe there is a way to leverage it, but it's good for nothing if the normal merge-base computation is order-dependent.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3061,7 +3061,7 @@ dependencies = [
 [[package]]
 name = "gix"
 version = "0.67.0"
-source = "git+https://github.com/Byron/gitoxide?rev=c5955fc4ad1064c7e4b4c57de32a661e693fbe49#c5955fc4ad1064c7e4b4c57de32a661e693fbe49"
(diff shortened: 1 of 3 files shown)


Caleb-T-Owens (approved):
LGTM :+1:

PR #5494, opened 2024-11-08
fix: remove extra spacing on InfoMessage

MikaelCarpenter:
## ☕️ Reasoning

In #5418 I made some changes to this `InfoMessage` to render the error message properly.
However, I introduced a scenario where the `content` slot could be rendered to the DOM with a `pre` element, but no text content. 

Ordinarily the content element would hide itself if empty, but the `pre` element prevents that. The parent container also uses `gap` to add space between the `title` and `content` slots, resulting in asymmetric spacing

## 🧢 Changes

- Don't render the `pre` if there's no error message

## 📸 Screenshots

|BEFORE|AFTER|
|-----|-----|
|before_checked|after_checked|
|before_success|after_signed|
|before_error|after_error|

### DIFF
--- a/apps/desktop/src/lib/settings/userPreferences/CommitSigningForm.svelte
+++ b/apps/desktop/src/lib/settings/userPreferences/CommitSigningForm.svelte
@@ -164,7 +164,9 @@
 					</svelte:fragment>
 
 					<svelte:fragment slot="content">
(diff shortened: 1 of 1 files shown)


estib-vega (approved):
Thanks for the attention to detail and for the detailed PR description.
This is highly appreciated!

PR #5520, opened 2024-11-11
Chore: remove GC restriction

krlvi:
Previous the app used to disable Git GC. This is no longer needed since all git trees used by gitbutler are now referenced via the reflog hack used by the oplog. This change removes the 'prune never' config if it was previously set by the app.


Caleb-T-Owens (approved):
Looks good to me :+1:

PR #5528, opened 2024-11-12
Refactor CommandContext out of various stack functions

Caleb-T-Owens:
The rational behind this is as follows:

Passing the CommandContext around everywhere as kind of global "god" object, is a very heavy handed thing to do - and makes unit testing harder. As well as exposing a large amount of state everywhere. In the upstream integration logic, it is written without a command context, making it very easy to test. I've gone and removed the CommandContext for everywhere reasonably can in order to enable me to call the required methods from upstream_integration.

In general, we should try to avoid CommandContext where possible, in favor of smaller Contexts or by passing around the objects directly, where appropriate.

### DIFF
--- a/crates/gitbutler-stack/src/stack_context.rs
+++ b/crates/gitbutler-stack/src/stack_context.rs
@@ -1,0 +1,35 @@
+use anyhow::Result;
+use gitbutler_command_context::CommandContext;
+
+use crate::{Target, VirtualBranchesHandle};
+
+pub struct StackContext<'repositroy> {
+    repository: &'repositroy git2::Repository,
+    target: Target,
+}
+
+impl<'repository> StackContext<'repository> {
+    pub fn repository(&self) -> &'repository git2::Repository {
+        self.repository
+    }
+
+    pub fn target(&self) -> &Target {
+        &self.target
+    }
+}
+
+pub trait CommandContextExt {
+    fn to_stack_context(&self) -> Result<StackContext>;
+}
+
+impl CommandContextExt for CommandContext {
+    fn to_stack_context(&self) -> Result<StackContext> {
+        let virtual_branch_state = VirtualBranchesHandle::new(self.project().gb_dir());
+        let default_target = virtual_branch_state.get_default_target()?;
+
+        Ok(StackContext {
+            repository: self.repository(),
+            target: default_target,
+        })
+    }
+}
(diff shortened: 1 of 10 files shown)


crates/gitbutler-stack/src/stack_context.rs
```diff
@@ -0,0 +1,33 @@
...
+    pub fn repository(&self) -> &'repository git2::Repository {
+        self.repository
+    }
+
+    pub fn target(&self) -> &Target {
+        &self.target
+    }
+}
+
+impl<'repository> TryFrom<&'repository CommandContext> for StackContext<'repository> {
```
Byron: I strongly recommend to avoid converting a reference of something into something owned, as it makes the call-sites puzzling to read. It feels like it's consuming the instance, but it's really only operating on a reference.

Instead, use `fn to_stack_context(&self) -> StackContext<'_>` on `CommandContext`, which will be clear and explicitly, with `to_` signalling that it's not consuming while producing something new.

For completeness, there are the following prefixes to choose from:

* `into_…(self) -> NewThing`
* `to_...(&self) -> NewThing`
* `as_…(&self) -> &PartOfSelf`

Only when the idea is to consume the `Thing`, then `TryFrom/From` are fitting semantically.

Byron (commented):
Thanks for pursuing this!

I really like the approach of only passing what needs to be passed. The idea of having a 'big context' and turning this into subsets of the context with just the right amount of state for each area of the codebase is very suitable to facilitate testing and of course, prevent bugs due to accidentally messing with state. Less is more.

crates/gitbutler-stack/src/stack_context.rs
krlvi:
```
into_…(self) -> NewThing
to_...(&self) -> NewThing
as_…(&self) -> &PartOfSelf
```
Thank you for sharing this! I found it educational for future code that I write!

---
PS: specifically on this issue, we chatted about it on Discord about just passing the 2 dependencies instead of making a new context https://discord.com/channels/1060193121130000425/1272848546399387669/1305972659926073456

krlvi (approved):
in the interest of iterating more quickly, lets merge as is

crates/gitbutler-stack/src/stack_context.rs
Caleb-T-Owens: As an update, we've decided to keep the `StackContext` for now, and will come back to the extra refactor later

PR #5533, opened 2024-11-13
Ability to discard single lines

estib-vega:
## ☕️ Reasoning

This pull request adds the ability to discard single lines in the file hunk context menu. This feature allows users to selectively remove individual lines from the changes, providing more granular control over the commit process.

## 🧢 Changes

- Added a method to unapply ownership lines, enabling the determination of a subset of hunk lines to be discarded.
- Implemented the option to discard single lines in the file hunk context menu.

### DIFF
--- a/crates/gitbutler-diff/src/diff.rs
+++ b/crates/gitbutler-diff/src/diff.rs
@@ -385,6 +385,227 @@ fn reverse_patch(patch: &BStr) -> Option<BString> {
     Some(reversed)
 }
 
+struct PatchHeaderInfo {
+    old_start: u32,
+    old_lines: u32,
+    new_start: u32,
+    new_lines: u32,
+    rest: BString,
+}
+
+/// Write a patch header from patch header information
+fn write_patch_header(parsed_header: &PatchHeaderInfo) -> BString {
+    let PatchHeaderInfo {
+        old_start,
+        old_lines,
+        new_start,
+        new_lines,
+        rest,
+    } = parsed_header;
+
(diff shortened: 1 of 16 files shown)


ndom91 (approved):
Works on my machine, nice!

crates/gitbutler-diff/src/diff.rs
```diff
@@ -385,6 +385,227 @@ fn reverse_patch(patch: &BStr) -> Option<BString> {
...
+    Some(PatchHeaderInfo {
+        old_start,
+        old_lines,
+        new_start,
+        new_lines,
+        rest,
+    })
+}
+
+fn create_patch_line_map_key(old_line: Option<u32>, new_line: Option<u32>) -> String {
```
Byron, line 469: `(Option<u32>, Option<u32>)` will be perfectly suitable as hashmap key, without the need for allocation. Maybe there is other reasons to go with `String` though, but I think there shouldn't be.

Byron (commented):
I just took a quick look and hope the comment will be useful.

Generally, when looking at this, I think there should be no need to deal with stringly-types like patches that could be applied with `patch`, and hope to one day be able to help remove it.

crates/gitbutler-diff/src/diff.rs
estib-vega: Thanks for the comment!
Indeed there is no need for strings to be created. I'll add a follow-up PR with the improvement :)

PR #5537, opened 2024-11-13
feat: add char count to commit msg input

ndom91:
## ☕️ Reasoning

- Feature request

image

Disappears when `msg.length === 0`

image


ndom91:
> @ndom91 I made a few changes:
> 
> - I removed the character counter from the `textarea` component to keep its logic simpler, since we don’t use this anywhere else.
> - Previously, we had a warning about the commit summary length at the bottom of the dialog box that only showed if the limit was exceeded. I removed the warning and kept just the counter, as I don’t think we need both.
> 
> image
> 
> 

Sounds good. You just gotta approve it now if you think it's ready 😉

PavelLaptev (approved):

PR #5542, opened 2024-11-14
fix: PR status polling

estib-vega:
- When creating or reopening a PR, refresh the listing service
- Poll for the mergeability status if it's `unknown`

### DIFF
--- a/apps/desktop/src/lib/branch/SeriesHeader.svelte
+++ b/apps/desktop/src/lib/branch/SeriesHeader.svelte
@@ -96,9 +96,16 @@
 		}
 	});
 
+	async function handleReloadPR() {
+		await Promise.allSettled([
+			prMonitor?.refresh(),
+			checksMonitor?.update(),
+			$hostedListingServiceStore?.refresh()
+		]);
+	}
+
 	function updateStatusAndChecks() {
-		prMonitor?.refresh();
-		checksMonitor?.update();
+		handleReloadPR();
 	}
 
 	const projectService = getContext(ProjectService);
@@ -130,10 +137,6 @@
 		}
 	});
 
-	async function handleReloadPR() {
-		await Promise.allSettled([prMonitor?.refresh(), checksMonitor?.update()]);
-	}
-
 	function handleOpenPR(pushBeforeCreate: boolean = false) {
 		prDetailsModal?.show(pushBeforeCreate);
 	}
@@ -143,7 +146,7 @@
 			return;
 		}
 		await $prService?.reopen($pr?.number);
-		await Promise.allSettled([prMonitor?.refresh(), checksMonitor?.update()]);
+		await handleReloadPR();
 	}
 
 	function editTitle(title: string) {
--- a/apps/desktop/src/lib/forge/github/githubPrMonitor.ts
+++ b/apps/desktop/src/lib/forge/github/githubPrMonitor.ts
@@ -5,6 +5,7 @@ import type { GitHubPrService } from './githubPrService';
 import type { ForgePrMonitor } from '../interface/forgePrMonitor';
 
 export const PR_SERVICE_INTERVAL = 20 * 60 * 1000;
+const MAX_POLL_ATTEMPTS = 6;
 
 export class GitHubPrMonitor implements ForgePrMonitor {
 	readonly pr = writable<DetailedPullRequest | undefined>(undefined, () => {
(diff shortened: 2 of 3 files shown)


apps/desktop/src/lib/pr/PrDetailsModal.svelte
```diff
@@ -234,6 +236,9 @@
 			if (priorIds.length > 0) {
 				updatePrStackInfo($prService, priorIds.concat([pr.number]));
 			}
+
+			// Refresh store
+			$forgeListingService?.refresh();
```
estib-vega, line 241: Refresh PR store after creation in order to have it there immediately after

apps/desktop/src/lib/branch/SeriesHeader.svelte
```diff
@@ -96,9 +96,23 @@
...
+		if ($pr?.mergeableState === 'unknown') {
+			prMonitor?.pollForMergeableState();
+		}
+	});
+
+	async function handleReloadPR() {
+		await Promise.allSettled([
+			prMonitor?.refresh(),
+			checksMonitor?.update(),
+			$hostedListingServiceStore?.refresh()
```
estib-vega, line 103: Reloading PR also reloads the store, to fetch reopened PRs

apps/desktop/src/lib/branch/SeriesHeader.svelte
```diff
@@ -96,9 +96,23 @@
 		}
 	});
 
+	// Trigger refresh of pull request mergeable state when it is unknown.
+	$effect(() => {
+		if ($pr?.mergeableState === 'unknown') {
+			prMonitor?.pollForMergeableState();
+		}
```
estib-vega: If the mergeable state is still unknown, poll it

apps/desktop/src/lib/forge/github/githubPrMonitor.ts
```diff
@@ -42,6 +43,17 @@ export class GitHubPrMonitor implements ForgePrMonitor {
 		await this.fetch();
 	}
 
+	async pollForMergeableState(attempts?: number): Promise<void> {
+		const actualAttempts = attempts ?? 0;
+		if (actualAttempts >= MAX_POLL_ATTEMPTS) return;
+		const currentMergeableState = get(this.mergeableState);
```
mtsgrd: I've been discussing something related to this `get(...)` call with @Caleb-T-Owens lately. There is an unexpected consequence of reading the mergeable state, which is that the underlying stores are synchronously subscribed to, and then unsubscribed from. In the case where nothing else is subscribed to either `pr` or `mergeableState`, then the `StartStopNotifier` will run, as well as any returned teardown function.

While this code definitely works, I'm wondering if we shouldn't have a more direct way of obtaining the loaded pr, and its mergeable state?

estib-vega: By direct do you mean something like a stand-alone function that just polls for the state?
If so I guess that we can do that, but we still need to set the information to a store so that the mergeable state is propagated

I guess another way would be to have the fetch function retry until there is data (as `getPrWithRetries` already does) and until `mergeableState` is not undefined

apps/desktop/src/lib/forge/github/githubPrMonitor.ts
```diff
@@ -56,15 +57,29 @@ export class GitHubPrMonitor implements ForgePrMonitor {
...
-				return await request();
+				const detailedPR = await request();
+				// Set the PR info immediately to show
+				// the user all preliminary information.
+				this.pr.set(detailedPR);
+				if (detailedPR.mergeableState === 'unknown') {
+					await sleep(1000);
+					continue;
+				}
+				return detailedPR;
```
estib-vega, line 82: @mtsgrd, what I did here was that as soon as we have the detailed PR information, we set to the writable.
The function does not return yet, an continues polling as long as the `mergeableState` is `unknown` or the max poll attempts are crossed.
Only then the function returns and the `loading` writable is set to false.

That way we can display PR detailed info as soon as we get it, while also ensuring that the `mergeableStatus` will be fetched.

estib-vega (commented):
@mtsgrd I opted not to add another loading PR spinner separate from the existing PR checks because of 2 things:
1. The checks loading state will only initiate after we load the initial PR information
2. We already disable the 'merge PR' button as long as the `mergeableStatus` is `unknown`.

Let me know what you think :)

apps/desktop/src/lib/forge/github/githubPrMonitor.ts
mtsgrd: Perfect.

mtsgrd (approved):
lgtm!

PR #5545, opened 2024-11-14
Use user profile image instead of gravatar when available

### LINKED ISSUE #2589: Profile picture doesn't get updated at all places
New profile picture doesn't reflect next to the commit messages in virtual branches

GordonBeeming:
## ☕️ Reasoning

Users wanted to see their GitButler profile image instead of the gravatar one when they've suppled one

## 🧢 Changes

Updated locations where gravatarUrl were referenced for various git artifacts to check the users email, and if it's a match, then use the users profile

## 🎫 Affected issues

Fixes: #2589

**NOTE:** I've never worked with this tech stack, but am trying to learn... and really want to contribute to GitButler, so starting small.

If I've completely missed the mark on how to implement this change, please let me know and I'll do some more learning 😅

### DIFF
--- a/apps/desktop/src/lib/commit/CommitCard.svelte
+++ b/apps/desktop/src/lib/commit/CommitCard.svelte
@@ -9,6 +9,7 @@
 	import { DraggableCommit, nonDraggable } from '$lib/dragging/draggables';
 	import BranchFilesList from '$lib/file/BranchFilesList.svelte';
 	import { ModeService } from '$lib/modes/service';
+	import { UserService } from '$lib/stores/user';
 	import { copyToClipboard } from '$lib/utils/clipboard';
 	import { openExternalUrl } from '$lib/utils/url';
 	import { BranchController } from '$lib/vbranches/branchController';
@@ -30,6 +31,9 @@
 	import PopoverActionsItem from '@gitbutler/ui/popoverActions/PopoverActionsItem.svelte';
 	import { getTimeAgo } from '@gitbutler/ui/utils/timeAgo';
 	import { type Snippet } from 'svelte';
+
+	const userService = getContext(UserService);
+	const user = userService.user;
 
 	interface Props {
 		branch?: VirtualBranch | undefined;
@@ -133,6 +137,10 @@
 	}
 
 	const commitShortSha = commit.id.substring(0, 7);
+	const authorImgUrl =
+		commit.author.email?.toLowerCase() === $user?.email?.toLowerCase()
+			? $user?.picture
+			: commit.author.gravatarUrl;
 
 	function handleUncommit(e: MouseEvent) {
 		e.stopPropagation();
@@ -242,7 +250,7 @@
 				label: commit.descriptionTitle,
 				sha: commitShortSha,
 				date: getTimeAgo(commit.createdAt),
-				authorImgUrl: commit.author.gravatarUrl,
+				authorImgUrl: authorImgUrl,
 				commitType: type,
 				data: new DraggableCommit(commit.branchId, commit, isHeadCommit, seriesName),
 				viewportId: 'board-viewport'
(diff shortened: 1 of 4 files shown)


apps/desktop/src/lib/navigation/BranchListingSidebarEntry.svelte
```diff
@@ -92,15 +96,18 @@
 		if (ownedByUser) {
 			const name = (await gitConfigService.get('user.name')) || unknownName;
 			const email = (await gitConfigService.get('user.email')) || unknownEmail;
-			const srcUrl = await gravatarUrlFromEmail(email);
+			const srcUrl = $user?.picture ? $user?.picture : await gravatarUrlFromEmail(email);
```
estib-vega: nit: Can we make this a bit more concise, like so:
```ts
$user?.picture ?? (await gravatarUrlFromEmail(email))
```

Also, I guess here we also need to check whether the `$user` email matches the git config's `user.email`, right?

apps/desktop/src/lib/components/EditMode.svelte
```diff
@@ -54,6 +59,15 @@
...
 
+	$effect(() => {
+		if (commit) {
+			authorImgUrl =
+				commit.author.email?.toLowerCase() === $user?.email?.toLowerCase()
+					? $user?.picture
+					: commit.author.gravatarUrl;
+		}
+	});
+
```
estib-vega: Instead of setting this from inside an effect, let's create a derived value, please:
```ts
const authorImgUrl = $derived.by(() => {
    if (commit) {
	return commit.author.email?.toLowerCase() === $user?.email?.toLowerCase()
		? $user?.picture
		: commit.author.gravatarUrl;
    }
    return undefined;
});
```
We want to avoid setting state from inside effects if possible :)

apps/desktop/src/lib/commit/CommitCard.svelte
```diff
@@ -133,6 +137,10 @@
 	}
 
 	const commitShortSha = commit.id.substring(0, 7);
+	const authorImgUrl =
+		commit.author.email?.toLowerCase() === $user?.email?.toLowerCase()
+			? $user?.picture
+			: commit.author.gravatarUrl;
```
estib-vega: Ideally, this should be also a `$derived` value as well. 
Please, check out my comment below.

apps/desktop/src/lib/navigation/PullRequestSidebarEntry.svelte
```diff
@@ -15,6 +16,14 @@
...
 	const project = getContext(Project);
 
+	const userService = getContext(UserService);
+	const user = userService.user;
+
+	const authorImgUrl =
+		pullRequest.author?.email?.toLowerCase() === $user?.email?.toLowerCase()
+			? $user?.picture
+			: pullRequest.author?.gravatarUrl;
+
```
estib-vega: Same as the comment above, let's turn this into a `$derived` value, please

estib-vega (changes requested):
Thank you for your contribution!
It looks overall great, just a couple of small changes and should be good to go!

apps/desktop/src/lib/navigation/BranchListingSidebarEntry.svelte
GordonBeeming: Originally my thoughts were that because it's checking the gitconfig, at this point it's definitely "you", so use your profile image. 
However for consistency it does make sense to check the email 🙂

estib-vega: It's a good point.
At least for now, I think it makes sense to compare against the email always so that the gitbutler's account picture appears only when using a git profile that matches, and a different picture if you decide to use another git profile (in the case you use different emails for different projects, for example).

GordonBeeming: 100% agree - I have this that config locally 😅 I think my GB email is different to both my normal git emails (work and personal) 🙂

estib-vega (approved):

estib-vega:
Merging this.
Thanks so much again for the contribution (specially at late hours of the night in Australia!).
Keep those changes coming! :)

PR #5550, opened 2024-11-14
Ability to reset to remote commit

estib-vega:
We were able to hard reset to whatever was on the remote, in the case that e.g. one would amend a commit outside of GB and wanted to pull that in.
This re-adds that.

### Report
This was reported in Discord: https://discord.com/channels/1060193121130000425/1073202153163857920/1306629953848213585

### Changes
Adds a button to reset your local branch to whatever is on remote.
This works in the case in which there are divergent commits with the same change ID.`

### Visuals
Screenshot 2024-11-14 at 17 48 17

### DIFF
--- a/apps/desktop/src/lib/commits/utils.ts
+++ b/apps/desktop/src/lib/commits/utils.ts
@@ -1,0 +1,14 @@
+import type { DetailedCommit } from '$lib/vbranches/types';
+
+/**
+ * Find the last commit that diverged from the remote branch.
+ */
+export function findLastDivergentCommit(commits: DetailedCommit[]): DetailedCommit | undefined {
+	for (let i = commits.length - 1; i >= 0; i--) {
+		const commit = commits[i];
+		if (commit!.id !== commit!.remoteCommitId) {
+			return commit;
+		}
+	}
+	return undefined;
+}
--- a/apps/desktop/src/lib/commit/CommitAction.svelte
+++ b/apps/desktop/src/lib/commit/CommitAction.svelte
@@ -12,7 +12,7 @@
 	const { type, isLast, action }: Props = $props();
 </script>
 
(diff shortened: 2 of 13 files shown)


apps/desktop/src/lib/commits/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Find the last commit that diverged from the remote branch.
+ */
+export function findLastDivergentCommit(commits: DetailedCommit[]): DetailedCommit | undefined {
+	for (let i = commits.length - 1; i >= 0; i--) {
+		const commit = commits[i];
+		if (!commit) continue;
+		if (commit.id !== commit.remoteCommitId) {
+			return commit;
+		}
```
mtsgrd: I'm assuming the `if (!commit) ..` line makes eslint happy because of `noUncheckedIndexedAccess`, but imho we shouldn't need to do this. One solution could be using `const commit = commits[i] as DetailedCommit`.

estib-vega: Yeah that's why haha. I don't mind the paranoid extra check, but I can remove it if you prefer 👍.

Also, a more concise alternative:
```ts
if (commit!.id !== commit!.remoteCommitId) {
    return commit;
}
```

mtsgrd (approved):

estib-vega:
Will wait for some input from Pavel, so that we also decide which button to show in the case that:
**There are remote-only commits** AND **there are diverged commits**. 

We probably just want to show 1 button as it will be the same thing

estib-vega:
@PavelLaptev Ready for design review again :)
Screenshot 2024-11-15 at 11 55 41
Screenshot 2024-11-15 at 11 55 51

PR #5553, opened 2024-11-14
Fixes a bug where remote branches may incorrectly show more commits

krlvi:
While `default_target.sha` may be appropriate for computing what is new in workspace branches, this would be wrong for computing of what is remote branches.
To get what is the remote branch it needs to be compared to the head of the target branch.


Caleb-T-Owens (commented):
If this requires a specific value, why don't we obtain it inside the branch_to_remote_branch_data function?

krlvi:
> If this requires a specific value, why don't we obtain it inside the branch_to_remote_branch_data function?

you mean the `remote_head` method? it seems like this would be needed in other places (i suspect we have subtle bugs not using it already)

PR #5554, opened 2024-11-14
Correctly deduplicate hunk locks

estib-vega:
Deduplicate hunk text by the property values they contain

### DIFF
--- a/apps/desktop/src/lib/utils/filters.test.ts
+++ b/apps/desktop/src/lib/utils/filters.test.ts
@@ -1,0 +1,82 @@
+import * as Filters from './filters';
+import { expect, test, describe } from 'vitest';
+
+describe.concurrent('unique', () => {
+	test('When provided an array with duplicate values, it returns an array with unique values', () => {
+		const array = [1, 2, 3, 1, 2, 3, 4, 5];
+		const expected = [1, 2, 3, 4, 5];
+
+		expect(array.filter(Filters.unique)).toEqual(expected);
+	});
+
+	test('When provided an array with no duplicate values, it returns the original array', () => {
+		const array = [1, 2, 3, 4, 5];
+		const expected = [1, 2, 3, 4, 5];
+
+		expect(array.filter(Filters.unique)).toEqual(expected);
+	});
+});
+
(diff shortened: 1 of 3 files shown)


apps/desktop/src/lib/utils/filters.ts
```diff
@@ -2,3 +2,19 @@
...
+	const propertyKeys = Object.keys(value) as (keyof T)[];
+	if (propertyKeys.length > 0) {
+		return array.findIndex((v) => propertyKeys.every((key) => v?.[key] === value[key])) === index;
+	}
+
+	const objectPropertyKeys = Object.getOwnPropertyNames(value) as (keyof T)[];
+	return (
+		array.findIndex((v) => objectPropertyKeys.every((key) => v?.[key] === value[key])) === index
+	);
+}
```
mtsgrd: For the purpose of keeping things simple, it feels like we could live without this section for non-enumerable properties? https://stackoverflow.com/a/22658584

Also, perhaps deepUnique would be a fitting name?

estib-vega: We can 👍. 

Regarding the naming, I think 'deep' would imply that we compare the values recursively, while we only check one level inside. 

Maybe something like `uniqueByObjectValue`?

apps/desktop/src/lib/utils/filters.test.ts
```diff
@@ -0,0 +1,82 @@
...
+	});
+});
+
+describe.concurrent('uniqeByPropValues', () => {
+	class Test {
+		constructor(
+			public id: number,
+			public name: string
+		) {}
+		doSomwthing() {
```
ndom91: Typo? :joy:

ndom91 (approved):
LGTM. I agree with @estib-vega about the naming thing though, `deep` for me would imply we're comparing them recursively

apps/desktop/src/lib/utils/filters.test.ts
estib-vega: 😬

mtsgrd:
Yes, I agree. I thought I was using my brain when clearly I wasn't!

PR #5559, opened 2024-11-15
Persist PR content

estib-vega:
## ☕️ Reasoning

This pull request introduces a new utility function for temporarily persisting values, as well as the implementation of persisting the PR description and title for 5 minutes.

## 🧢 Changes

- Added a shared utility function for the temporary persistence of values. If retrieved after the given number of seconds, the value will be considered not present.
- Implemented the persistence of the PR description and title for 5 minutes, to mitigate the risk of accidentally closing the PR edit modal and having the work lost.

### DIFF
--- a/packages/shared/src/lib/persisted.ts
+++ b/packages/shared/src/lib/persisted.ts
@@ -1,6 +1,7 @@
 /**
  * This is simplified copy of the persisted store in square/svelte-store.
  */
+import lscache from 'lscache';
 import { writable, type Writable } from 'svelte/store';
 export type Persisted<T> = Writable<T>;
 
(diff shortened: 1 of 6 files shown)


packages/shared/src/lib/persisted.ts
```diff
@@ -52,3 +52,95 @@ export function persisted<T>(initial: T, key: string): Persisted<T> {
...
+	} catch {
+		return undefined;
+	}
+}
+
+/**
+ * Create a persisted store that expires after a certain amount of time.
+ */
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export function persistWithExpiration<T extends {}>(
```
estib-vega, line 79: Allow only non-nullish values: https://www.totaltypescript.com/the-empty-object-type-in-typescript#constraining-functions

packages/shared/src/lib/persisted.ts
```diff
@@ -52,3 +52,95 @@ export function persisted<T>(initial: T, key: string): Persisted<T> {
...
+			return undefined;
+		}
+
+		const now = Date.now();
+		if (parsed.expiration < now) {
+			window.sessionStorage.removeItem(key);
+			return undefined;
+		}
+
+		console.log('Item not expired:', key, parsed.value);
```
mtsgrd: 👀

estib-vega: Dang! Missed that

mtsgrd:
Could you have a look at where/how we use `lscache` and let me know if that's something we phase out or possibly use for ephemeral storage?

estib-vega:
> Could you have a look at where/how we use `lscache` and let me know if that's something we phase out or possibly use for ephemeral storage?


@mtsgrd Interesting:
We seem to use it for UI settings (lane widths and file widths).
No sure if we're actively phasing that out, but we indeed don't use it a lot.

For this case, though, it seems like it could be slightly better than my implementation, because this actually **flushes** the expired values:

```ts
// call on startup so we don't accumulate old items
lscache.flushExpired();
```

pnpm-lock.yaml
```diff
@@ -12455,7 +12462,7 @@ snapshots:
       why-is-node-running: 2.3.0
     optionalDependencies:
       '@types/node': 22.3.0
-      '@vitest/browser': 2.0.5(playwright@1.47.0)(typescript@5.4.5)(vitest@2.0.5)(webdriverio@8.40.2)
+      '@vitest/browser': 2.0.5(playwright@1.48.2)(typescript@5.4.5)(vitest@2.0.5)(webdriverio@8.40.2)
```
mtsgrd: unrelated perhaps?

mtsgrd (approved):

pnpm-lock.yaml
estib-vega: Indeed 👍

PR #5564, opened 2024-11-15
fix: assure we convert worktree data to Git before writing it (#5558)

### LINKED ISSUE #5558: rebase in gitbutler cause conflicts with line endings when autocrlf is set to true
### Version

0.13.14

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

I was trying to test gitbutler on an already existing repository where I have autocrlf true enabled because I work on windows and use the git-bash there.

When I now change a commit with the edit patch option I will get conflicts in all subsequent commits over the whole file because the line endings are different.

### How to reproduce

- setup a repo without gitbutler. On windows with autocrlf set to true
- Have a file with an initial commit and some text with line breaks.
- make a new branch that is not main
- make commit 1 where you add another line with text to the beginning of the file
- make commit 2 where you add another line with text to the end of the file

- open that repo with gitbutler and edit commit 1, where you change something in the line you introduced with that commit
- commit 2 should now show up as conflict.

When I check the conflicting file with an editor that shows crlf characters, the "ours" part has crlf as line nedings, while tne ancestor and "theirs" uses LF

### Expected behavior

the autocrlf setting should be respected and I should not get merge conflicts based on line endings.

### Relevant log output

_No response_

Byron:
Fixes #5558

### Tasks

* [x] first implementation
* [x] assure that on Windows it will also read the installation-configuration to definitely get all the settings

### Notes for the reviewer

* One could also use `git2::Repository::blob_writer(path)` which automatically applies the 'worktree-to-git' filtering as well. That would lead to simpler code, at least, but will have disadvantages in correctness as it doesn't support external filters, nor will it pick up all the configuration.
* These special open-options would probably have to be used in more places, *or* `gitoxide` makes these the default for good measure - right now it prefers performance as that option involves launching `git` to have it tell us where its installation configuration is located. In future, I can see that shift as more features will only work correctly in with this option enabled.

### DIFF
--- a/crates/gitbutler-repo/src/repository_ext.rs
+++ b/crates/gitbutler-repo/src/repository_ext.rs
@@ -195,7 +195,17 @@ impl RepositoryExt for git2::Repository {
     /// or if the HEAD branch has no commits
     #[instrument(level = tracing::Level::DEBUG, skip(self), err(Debug))]
     fn create_wd_tree(&self) -> Result<Tree> {
-        let gix_repo = gix::open(self.path())?;
+        let gix_repo = gix::open_opts(
+            self.path(),
+            gix::open::Options::default().permissions(gix::open::Permissions {
+                config: gix::open::permissions::Config {
+                    // Whenever we deal with worktree filters, we'd want to have the installation configuration as well.
+                    git_binary: cfg!(windows),
+                    ..Default::default()
+                },
+                ..Default::default()
+            }),
+        )?;
         let (mut pipeline, index) = gix_repo.filter_pipeline(None)?;
         let mut tree_update_builder = git2::build::TreeUpdateBuilder::new();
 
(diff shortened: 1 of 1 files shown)


crates/gitbutler-repo/src/repository_ext.rs
```diff
@@ -241,16 +254,19 @@ impl RepositoryExt for git2::Repository {
                     let blob = self.blob(path_str.as_bytes())?;
                     tree_update_builder.upsert(path, blob, git2::FileMode::Link);
                 } else {
-                    let file = std::fs::read(&file_path)?;
-                    let blob = self.blob(&file)?;
+                    let mut file_for_git =
+                        pipeline.convert_to_git(std::fs::File::open(&file_path)?, path, &index)?;
+                    buf.clear();
+                    std::io::copy(&mut file_for_git, &mut buf)?;
+                    let blob_id = self.blob(&buf)?;
```
Caleb-T-Owens: Why do we want to copy the data into the buffer and not just directly pass it to `self.blob`?

Byron: `file_for_git` isn't a buffer, it's something that implements `std::io::Read`, while `blob()` takes `&[u8]`.

This is the version of the code that is easiest to write, but also slowest as two out of three cases already have a buffer underneath which could be passed directly.

I will adjust it to be the most efficient possible.

Caleb-T-Owens (approved):
Thanks, looks good to me :+1:

PR #5566, opened 2024-11-15
Fix applying pr from fork

mtsgrd:
We were suggesting a remote name consisting of `$user/$repo`, but at the same time `RemoteRefname` does not handle names containing `/` well. With a remote name of e.g. `$user/$repo`, we have been creating a branch called `$repo/$branch` since `RemoteRefname` doesn't know at what slash to separate branch name from remote name.

- improved error messages
- use fork owner as remote name since remote refnames containing '/' aren't parsed correcty
- svelte5 upgrade for preview component

https://github.com/user-attachments/assets/ac3597c5-92c1-42b0-9caf-505df7f1af55

---
- `2.` #5566 👈 
- `1.` #5567

### DIFF
--- a/apps/desktop/src/lib/branch/SeriesHeader.svelte
+++ b/apps/desktop/src/lib/branch/SeriesHeader.svelte
@@ -82,8 +82,8 @@
 
 	// Pretty cumbersome way of getting the PR number, would be great if we can
 	// make it more concise somehow.
-	const hostedListingServiceStore = getForgeListingService();
-	const prStore = $derived($hostedListingServiceStore?.prs);
+	const forgeListing = getForgeListingService();
+	const prStore = $derived($forgeListing?.prs);
 	const prs = $derived(prStore ? $prStore : undefined);
 
 	const listedPr = $derived(prs?.find((pr) => pr.sourceBranch === upstreamName));
@@ -118,15 +118,11 @@
 	});
 
 	async function handleReloadPR() {
-		await Promise.allSettled([
-			prMonitor?.refresh(),
-			checksMonitor?.update(),
-			$hostedListingServiceStore?.refresh()
-		]);
+		await updateStatusAndChecks();
 	}
 
-	function updateStatusAndChecks() {
-		handleReloadPR();
+	async function updateStatusAndChecks() {
+		await Promise.allSettled([prMonitor?.refresh(), checksMonitor?.update()]);
 	}
 
 	const projectService = getContext(ProjectService);
@@ -176,6 +172,7 @@
 			return;
 		}
 		await $prService?.reopen($pr?.number);
+		await $forgeListing?.refresh();
 		await handleReloadPR();
 	}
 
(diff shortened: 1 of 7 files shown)


apps/desktop/src/lib/branch/SeriesHeader.svelte
```diff
@@ -97,15 +97,11 @@
 	});
 
 	async function handleReloadPR() {
-		await Promise.allSettled([
-			prMonitor?.refresh(),
-			checksMonitor?.update(),
-			$hostedListingServiceStore?.refresh()
```
mtsgrd, line 124: I don't think `HostedListingService` needs to be refreshed since its output is only relevant to displaying a pr if `pr_number` isn't available in `virtual_branches.toml`.

apps/desktop/src/lib/forge/github/githubPrMonitor.ts
```diff
@@ -19,7 +19,6 @@ export class GitHubPrMonitor implements ForgePrMonitor {
 	readonly error = writable<any>();
 
 	readonly mergeableState = derived(this.pr, (pr) => pr?.mergeableState);
-	readonly lastFetch = writable<Date | undefined>();
```
mtsgrd, line 22: This wasn't being used anywhere, but should perhaps not be part of this pr.

apps/desktop/src/lib/branch/SeriesHeader.svelte
estib-vega: I think we only needed it when reopening the PR.
So as long as that is called then (as you are doing in line 145) we can remove it safely from here 👍

estib-vega (approved):

apps/desktop/src/lib/components/PullRequestPreview.svelte
```diff
@@ -3,91 +3,87 @@
...
+	const baseBranch = $derived(baseBranchService.base);
 
-	let remoteName = structuredClone(pullrequest.repoName) || '';
+	let prRepoOwner = $derived(structuredClone(pr.repoOwner) || '');
+	$effect(() => {
+		if (prRepoOwner) {
+			inputRemoteName = prRepoOwner;
+		}
+	});
+	let inputRemoteName = $state<string>('');
```
estib-vega: Disclaimer: *This is more a of a nit-pick, since the potential bad case is probably impossible. In any case I though it would be good to mention this as a cautionary tale*

Could we add a derived value e.g. `actualInputRemoteName` in order to avoid setting state from the effect?

Something like this:
```ts
const prRepoOwner = $derived(structuredClone(pr.repoOwner) || '');
let inputRemoteName = $state<string>('');
const actualRemoteName = $derived<string>(inputRemoteName || prRepoOwner);
```

This accomplishes 2 things:
- We avoid the anti-pattern
- We prevent the rare state of the user-given input being overridden by a the `pr.repoOwner` being updated externally (something that's admittedly, probably impossible).

estib-vega (approved):
Just one comment for an optional improvement

apps/desktop/src/lib/components/PullRequestPreview.svelte
mtsgrd: Sounds good, thank you!

PR #5567, opened 2024-11-15
Check for existing fork before adding new one

mtsgrd:
Adding these checks because a user reported on Discord that a remote was added twice. The root problem was pr preview being broken, but we should nonetheless prevent and show clear error messages if it happens.

- can prevent a cascading error
- exposes remote urls to front end

---
- `2.` #5566 
- `1.` #5567 👈


mtsgrd:
@krlvi any chance you could review this? Low impact, but good hygiene imho.

krlvi (approved):
seems reasonable

PR #5565, opened 2024-11-15
Fix up some janky code mapping commit relationships

mtsgrd:
- fixes error where applied pr from fork immediately has new upstream commits

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #5573 
- <kbd>&nbsp;1&nbsp;</kbd> #5565 👈

PR #5573, opened 2024-11-16
Remove legacy branch listing api

mtsgrd:
- renamed to `find_git_branches` by name
- was only used to get branches for preview
- was unnecessarily being reloaded constantly

---
- <kbd>&nbsp;2&nbsp;</kbd> #5573 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #5565


ndom91:
What do you think about using `<kbd>` elements? Combined some `emsp` we can even use the finger on the left and have it look alright imo


---
* <kbd>&nbsp;👉&nbsp;</kbd> #5573  
* <kbd>&emsp;1&emsp;</kbd> #5565

apps/desktop/src/routes/[projectId]/branch/[...name]/+page.svelte
```diff
@@ -1,65 +1,36 @@
...
-			<BranchPreview {localBranch} {remoteBranch} {pr} />
-		{/each}
-	{/if}
-{:else}
+{#if localBranch}
+	<BranchPreview {localBranch} {pr} />
+{/if}
+{#each remoteBranches as remoteBranch}
+	<BranchPreview {remoteBranch} {pr} />
+{/each}
```
Comment (Caleb-T-Owens) [inline apps/desktop/src/routes/[projectId]/branch/[...name]/+page.svelte]: Is there a particular design wise reason we're changing thew way we display Local and Remote previews?

apps/desktop/src/lib/stores/remoteBranches.ts
```diff
@@ -1,39 +1,15 @@
-import { Code, invoke } from '$lib/backend/ipc';
+import { invoke } from '$lib/backend/ipc';
 import { Branch, BranchData } from '$lib/vbranches/types';
 import { plainToInstance } from 'class-transformer';
-import { writable } from 'svelte/store';
-import type { BranchListingService } from '$lib/branches/branchListing';
-import type { ProjectMetrics } from '$lib/metrics/projectMetrics';
 
 export class RemoteBranchService {
```
Caleb-T-Owens: Should we call this the GitBranchService?

crates/gitbutler-branch-actions/src/remote.rs
```diff
@@ -60,32 +62,50 @@ pub struct RemoteCommit {
...
-    let remotes = ctx.repository().remotes()?;
-    for (branch, _) in ctx
-        .repository()
-        .branches(None)
-        .context("failed to list remote branches")?
-        .flatten()
+    let repo = ctx.repository();
+    let mut branches = vec![];
+    let remotes_raw = repo.remotes()?;
+    let remotes = remotes_raw.iter().flatten().collect_vec();
```
Caleb-T-Owens: Why do we need to flatten the remotes array? Isn't the return type `Result<Vec<String>>`?

crates/gitbutler-branch-actions/src/remote.rs
```diff
@@ -60,32 +62,50 @@ pub struct RemoteCommit {
...
+
+    if let Some(local_branch) = repo
+        .find_branch(branch_name, BranchType::Local)
+        .ok()
+        .and_then(|branch| {
+            branch_to_remote_branch(&branch, &remotes)
+                .map_err(|err| {
+                    tracing::warn!(?err, "Ignoring branch");
+                })
+                .ok()
```
Caleb-T-Owens: This is completely unrelated and 100% just a result of historical naming, but it did make me chuckle

crates/gitbutler-branch-actions/src/remote.rs
```diff
@@ -60,32 +62,50 @@ pub struct RemoteCommit {
...
+    let remotes = remotes_raw.iter().flatten().collect_vec();
+
+    let mut all_branches: Vec<_> = remotes
+        .iter()
+        .filter_map(|remote| {
+            let branch_name = format!("{}/{}", remote, branch_name);
+            repo.find_branch(&branch_name, BranchType::Remote)
+                .ok()
+                .and_then(|branch| branch_to_remote_branch(&branch, &[remote]).ok())
+                .unwrap_or(None)
```
Caleb-T-Owens: If I'm not mistaken, could this be simplified to:
```rs
            repo.find_branch(&branch_name, BranchType::Remote)
                 .and_then(|branch| branch_to_remote_branch(&branch, &[remote]))
                 .ok()
```

crates/gitbutler-branch-actions/src/remote.rs
```diff
@@ -60,32 +62,50 @@ pub struct RemoteCommit {
...
-        let branch = match branch_to_remote_branch(&branch, &remotes) {
-            Ok(Some(b)) => b,
-            Ok(None) => continue,
-            Err(err) => {
-                tracing::warn!(?err, "Ignoring branch");
-                continue;
-            }
-        };
+        all_branches.push(local_branch);
+    }
```
Caleb-T-Owens: Could we write this one as
```rs
// Let the compiler know we're swallowing the error
let _ = repo
         .find_branch(branch_name, BranchType::Local)
         .and_then(|branch| {
             let branch = branch_to_remote_branch(&branch, &remotes)?;
             all_branches.push(branch);
             Ok(())
         });
```

crates/gitbutler-branch-actions/src/remote.rs
```diff
@@ -95,10 +115,10 @@ pub fn list_local_branches(ctx: &CommandContext) -> Result<Vec<RemoteBranch>> {
             && branch.name.branch() != Some("gitbutler/edit")
             && branch.name.branch() != Some("gitbutler/target")
         {
-            remote_branches.push(branch);
+            branches.push(branch.to_owned());
```
Caleb-T-Owens: Could we make sure that the `branches` array is filled with the owned types already so we don't need to clone them?

crates/gitbutler-branch-actions/src/virtual.rs
```diff
@@ -455,7 +456,8 @@ pub fn list_virtual_branches_cached(
         let base_current = true;
 
         let upstream = upstream_branch.and_then(|upstream_branch| {
-            let remotes = repo.remotes().ok()?;
+            let remotes_raw = repo.remotes().ok()?;
+            let remotes = remotes_raw.into_iter().flatten().collect_vec();
```
Caleb-T-Owens: @Byron do you have an opinion on itertools? In recent weeks I've seen it get used more and more, and none of the uses seem to be much of a gain over doing it without.

crates/gitbutler-cli/src/main.rs
```diff
@@ -35,7 +35,7 @@ fn main() -> Result<()> {
                 Some(vbranch::SubCommands::SetBase {
                     short_tracking_branch_name,
                 }) => command::vbranch::set_base(project, short_tracking_branch_name),
-                Some(vbranch::SubCommands::ListLocal) => command::vbranch::list_local(project),
+                Some(vbranch::SubCommands::ListLocal) => command::vbranch::list_all(project),
```
Caleb-T-Owens: It would be reasonable to deem this out of scope, but we should add a TODO here because the command name is clearly wrong

mtsgrd: Good point. I need to change this line since I remove `list_local` function. I'll update the enum to `SubCommand:List`, and update the comment on the definition.

crates/gitbutler-branch-actions/src/virtual.rs
mtsgrd: The alternative feels a bit more verbose? Either:
```
let remotes: Vec<_> = remotes_raw.into_iter().flatten().collect();
```
or
```
let remotes = remotes_raw.into_iter().flatten().collect::<Vec<_>>();
```

crates/gitbutler-branch-actions/src/remote.rs
mtsgrd: The question mark does not work inside of `.and_then(..)`. Also, `branch_to_remote_branch` returns an `Option`, but we could perhaps change that to an error.

crates/gitbutler-branch-actions/src/remote.rs
mtsgrd: It can't, because that inner function returns a result of an option. But like I say in a comment below, let's change the inner function error on the conversion failing. It does not appear to me like there should be a legitimate reason for it to fail?

crates/gitbutler-branch-actions/src/remote.rs
mtsgrd: We need to flatten it since remotes_raw is a git2::StringArray, which converts into `Vec<Option<&str>>`.

crates/gitbutler-branch-actions/src/virtual.rs
Byron: Personally I avoid it, particularly because it can make it easy to hide allocations, e.g. it makes things just work, but will allocate internally, something that is unintuitive when using iterators.

However, sometimes it can help with the readability of code.

Also, I think I vaguely remember to have removed it in cases like these if I could then also remove the crate dependency.

But since it's proliferating I suppose it helps, and that's probably what makes it worth keeping - there are definitely bigger fish to fry.

apps/desktop/src/routes/[projectId]/branch/[...name]/+page.svelte
Comment (mtsgrd) [reply apps/desktop/src/routes/[projectId]/branch/[...name]/+page.svelte]: Right this is perhaps a mistake on my part. Can you confirm we do _not_ want to show the local branch if one or more remote branches exist?

Comment (mtsgrd) [reply apps/desktop/src/routes/[projectId]/branch/[...name]/+page.svelte]: I'll add back a loading state as well.

crates/gitbutler-branch-actions/src/virtual.rs
mtsgrd: I think if they are functionally I don't mind using `.collect()`, but which of the two styles I mention above do you guys prefer?

crates/gitbutler-branch-actions/src/remote.rs
Caleb-T-Owens: I see

apps/desktop/src/routes/[projectId]/branch/[...name]/+page.svelte
Comment (Caleb-T-Owens) [reply apps/desktop/src/routes/[projectId]/branch/[...name]/+page.svelte]: > Can you confirm we do not want to show the local branch if one or more remote branches exist?
Yes. If both a local and remote branch is provided to the component, it displays the local branch commits as the branch commits, and the remote commits as upstream ones; like how it is in the workspace

crates/gitbutler-branch-actions/src/remote.rs
Caleb-T-Owens: The question mark does work inside of `and_then`. If you've got a function or closure whose return type is `Option`, then you can use `?` on any functions which return `Option` inside of it, and similar for `Result`.


I don't see any particularly good reason why branch_to_remote_branch should be `Result<Option<>>` as it a failure to convert. If it returns just a `Result` then my snippet should work.

crates/gitbutler-branch-actions/src/virtual.rs
Caleb-T-Owens: collect_vec() is the itertools, so the plain rust collect() is the preferable one.

mtsgrd: Yes but it often requires explicit types. Just wondering if there's a better option between putting it at the beginning of the line, or next to the collect_vec call, as mentioned in my previous comment.

Caleb-T-Owens: I think `let remotes = remotes_raw.into_iter().flatten().collect::<Vec<_>>();` is preferable

Byron: Sometimes the `Vec<_>` can be inferred, but probably not in this situation.

To my mind, using `itertools` here is fine as it is intuitive and thus is fast and easy to pull off. And that matters.
For `gitoxide` I can't use it, as I get angry letters sent to my house when a new dependency shows up downstream, and it could hide allocations which I rather keep explicit (e.g. `it.sorted()`).

But in application code I don't think there is a good enough reason not to use it.

crates/gitbutler-branch-actions/src/remote.rs
```diff
@@ -60,45 +61,51 @@ pub struct RemoteCommit {
...
+        .iter()
+        .filter_map(|remote_name| {
+            let branch_name = format!("{}/{}", remote_name, branch_name);
+            let blah = repo
+                .find_branch(&branch_name, BranchType::Remote)
+                .ok()
+                .and_then(|branch| branch_to_remote_branch(&branch, &[remote_name]).ok());
+            blah
+        })
+        .collect();
```
mtsgrd: @Caleb-T-Owens I think it's best if the functions return errors, since one would expect them to work if you are looking up a valid repo and branch. In the case of `get_remote_branch_data` it's natural to fail the front-end request, while here in a "find" function it feels ok to swallow them?

crates/gitbutler-branch-actions/src/remote.rs
```diff
@@ -60,45 +61,51 @@ pub struct RemoteCommit {
...
-            remote_branches.push(branch);
-        }
+        all_branches.push(local_branch);
     }
-    Ok(remote_branches)
+
+    let target_branch = &default_target(&ctx.project().gb_dir())?.branch;
+    Ok(all_branches
+        .into_iter()
+        .filter(move |branch| {
```
mtsgrd: The `move` isn't necessary, I'll remove it.

crates/gitbutler-branch-actions/src/remote.rs
Caleb-T-Owens: Can we talk about this on a call tomorrow?

mtsgrd: Sounds like a plan.

ndom91:
> @ndom91 that misalignment introduced by the emoji hurts my eyes a bit. Could you suggest that in a different pr if you are motivated about it? :) 

Hahah yeah I was playing around with all the different space types to try to get it to line up perfectly, but no dice. 

If we drop the emoji in the front though I think just the digits, `kbd` and some `nbsp`'s would be nice, for example: 

- <kbd>&nbsp;1&nbsp;</kbd>👈 blah blah blah 
- <kbd>&nbsp;2&nbsp;</kbd> foo bar baz 

I can bring it up elsewhere 👍

Caleb-T-Owens (approved):
LGTM

PR #5587, opened 2024-11-18
fix: truncate long path from front, not back

ndom91:
## ☕️ Reasoning

- In the case of long paths, truncate from the beginning and show the end.

### Before
image

### After

image


PavelLaptev (approved):
LGTM, thanks

PR #5590, opened 2024-11-18
Dependecies: Ignore merge commits

estib-vega:
## ☕️ Reasoning
Obviously don't ignore **all** merge commits. Only some.

Calculating the dependency graph of all changes in a stack, should ignore merge-commits that bring in changes already integrated in the target branch.

## 🧢 Changes

- Added an alias for the upstream integration command to be used in the integration tests.
- Implemented a change to ignore merge commits when computing the dependency graphs, ensuring that the dependency management process is more accurate and reliable.

## Reported
As seen in this Discord thread: https://discord.com/channels/1060193121130000425/1308034358820733029

### DIFF
--- a/crates/gitbutler-branch-actions/src/dependencies.rs
+++ b/crates/gitbutler-branch-actions/src/dependencies.rs
@@ -28,13 +28,7 @@ pub fn compute_workspace_dependencies(
     let mut stacks_input: Vec<InputStack> = vec![];
     for stack in stacks {
         let mut commits_input: Vec<InputCommit> = vec![];
-        // The commit id's must be in the order they get applied (parents first).
-        let commit_ids = repo
-            .l(stack.head(), LogUntil::Commit(*target_sha), false)
-            .context("failed to list commits")?
-            .into_iter()
-            .rev()
-            .collect_vec();
+        let commit_ids = get_commits_to_process(repo, stack, target_sha)?;
         for commit_id in commit_ids {
             let mut files_input: Vec<InputFile> = vec![];
             let commit = repo.find_commit(commit_id)?;
@@ -71,6 +65,41 @@ pub fn compute_workspace_dependencies(
         workdir: base_diffs,
         stacks: stacks_input,
     })
+}
+
+/// Get the commits that need to be processed for the stack dependencies caculation.
+///
+/// Commit IDs are in the order that they are applied (parent first).
+/// Merge commits to the target branch are not included.
+fn get_commits_to_process(
+    repo: &git2::Repository,
+    stack: &Stack,
+    target_sha: &git2::Oid,
+) -> Result<Vec<git2::Oid>, anyhow::Error> {
+    let commit_ids = repo
+        .l(stack.head(), LogUntil::Commit(*target_sha), true)
+        .context("failed to list commits")?
+        .into_iter()
+        .rev()
+        .filter_map(|commit_id| {
+            let commit = repo.find_commit(commit_id).ok()?;
+            if commit.parent_count() == 1 {
+                return Some(commit_id);
+            }
+
+            let has_integrated_parent = commit.parent_ids().any(|id| {
+                let (number_commits_ahead, _) = repo.graph_ahead_behind(id, *target_sha).unwrap();
+                number_commits_ahead == 0
+            });
+
+            if has_integrated_parent {
+                None
+            } else {
+                Some(commit_id)
+            }
+        })
+        .collect_vec();
+    Ok(commit_ids)
 }
 
 pub struct StackDependencies {
--- a/crates/gitbutler-branch-actions/tests/dependencies.rs
+++ b/crates/gitbutler-branch-actions/tests/dependencies.rs
@@ -650,6 +650,76 @@ fn complex_file_manipulation_multiple_hunks_with_uncommitted_changes() -> Result
     Ok(())
 }
 
+#[test]
+fn dependecies_ignore_merge_commits() -> Result<()> {
+    let ctx = command_ctx_after_updates("merge-commit")?;
+    let test_ctx = test_ctx(&ctx)?;
+    let default_target = test_ctx.virtual_branches.get_default_target()?;
+    let my_stack = &test_ctx.stack;
+
+    let file_path = Path::new("file");
+    let file_hunk_1 = GitHunk {
+        old_start: 5,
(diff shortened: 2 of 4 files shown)


crates/gitbutler-branch-actions/src/dependencies.rs
```diff
@@ -73,6 +67,37 @@ pub fn compute_workspace_dependencies(
...
+            let has_integrated_parent = commit.parent_ids().any(|id| {
+                let (number_commits_ahead, _) = repo.graph_ahead_behind(id, *target_sha).unwrap();
+                number_commits_ahead == 0
+            });
+
+            if has_integrated_parent && commit.parent_count() > 1 {
+                None
+            } else {
+                Some(commit_id)
+            }
```
estib-vega: We filter out the commits that have multiple parents (merge commits), out of which at least one is already integrated into the target branch.

Caleb-T-Owens: lets check if it's a merge commit before doing potentially expensive lookups

Caleb-T-Owens (approved):
Other tha a minor performance problem, looks good to me

crates/gitbutler-branch-actions/src/dependencies.rs
estib-vega: Makes sense 👍

crates/gitbutler-branch-actions/src/dependencies.rs
```diff
@@ -73,6 +67,41 @@ pub fn compute_workspace_dependencies(
...
+
+            let has_integrated_parent = commit.parent_ids().any(|id| {
+                let (number_commits_ahead, _) = repo.graph_ahead_behind(id, *target_sha).unwrap();
+                number_commits_ahead == 0
+            });
+
+            if has_integrated_parent {
+                None
+            } else {
+                Some(commit_id)
```
Byron, line 98: Here one *could* use `(!has_integrated_commit).then_some(commit_id)`.

crates/gitbutler-branch-actions/src/dependencies.rs
```diff
@@ -73,6 +67,41 @@ pub fn compute_workspace_dependencies(
...
+        .rev()
+        .filter_map(|commit_id| {
+            let commit = repo.find_commit(commit_id).ok()?;
+            if commit.parent_count() == 1 {
+                return Some(commit_id);
+            }
+
+            let has_integrated_parent = commit.parent_ids().any(|id| {
+                let (number_commits_ahead, _) = repo.graph_ahead_behind(id, *target_sha).unwrap();
+                number_commits_ahead == 0
```
Byron, line 92: Unwraps are strictly forbidden as they will keep the UI waiting for a reply that never happens, usually stalling it. On top of that, panics show up in the `tokio` logs only which usually aren't provided if logs are handed in by hand, making these issues extra-hard to debug remotely.

In these situations, one can go with `fallible.map_or(true, |(num_ahead, _)| num_ahead == 0)`. *(Not actually sure what one should assume if this call fails)*

crates/gitbutler-branch-actions/tests/dependencies.rs
```diff
@@ -650,6 +650,76 @@ fn complex_file_manipulation_multiple_hunks_with_uncommitted_changes() -> Result
     Ok(())
 }
 
+#[test]
+fn dependecies_ignore_merge_commits() -> Result<()> {
```
Byron, line 654: It's great to have a test for this! Pretty advanced, too.

crates/gitbutler-branch-actions/tests/dependencies.rs
```diff
@@ -650,6 +650,76 @@ fn complex_file_manipulation_multiple_hunks_with_uncommitted_changes() -> Result
...
+
+    assert_eq!(dependencies.diffs.len(), 1);
+    let file_hunk_1_hash = Hunk::hash_diff(&file_hunk_1.diff_lines);
+    let file_hunk_2_hash = Hunk::hash_diff(&file_hunk_2.diff_lines);
+
+    let hunk_1_locks = dependencies.diffs.get(&file_hunk_1_hash);
+    assert_eq!(hunk_1_locks, None);
+
+    let hunk_2_locks = dependencies.diffs.get(&file_hunk_2_hash).unwrap();
+    assert_eq!(hunk_2_locks.len(), 1);
```
Byron, line 712: Something I always recommend is to use the 'hidden third parameter' to pass more context. For instance

* `assert!(!foo)` tells me something if I know why `foo` should be false
* `assert!(!foo, "Foo is negative as bar was positive, which always happens on full moons")` tells everyone why that is an expectation.

In other words, always pass as much context as possible with the third parameter in plain English because what seems obvious now, isn't obvious in a couple of weeks (or to people like me 😅).

crates/gitbutler-branch-actions/src/dependencies.rs
```diff
@@ -73,6 +67,41 @@ pub fn compute_workspace_dependencies(
...
+            });
+
+            if has_integrated_parent {
+                None
+            } else {
+                Some(commit_id)
+            }
+        })
+        .collect_vec();
```
Byron, line 101: Since this function is merely a filter on what could be a plain iterator, there is no need to collect it.
Instead, one can do `fn filter_iter() -> impl Iterator<Item = git2::Oid> {...}` and it will work naturally.

Probably not measurable by itself, but these things accumulate. Besides, I always find it confusing when a `Vec` is used even though there isn't a need for it.
Not a big deal as it's quite common in this codebase, I just want to create awareness.

Byron (commented):
Nice testing!

I left a few more comments below and hope they help. Except for the one about `.unwrap()` there is nothing to act on.

estib-vega:
> Nice testing!
> 
> I left a few more comments below and hope they help. Except for the one about `.unwrap()` there is nothing to act on.

@Byron Thanks for the feedback! I'll open a PR to follow this up

crates/gitbutler-branch-actions/src/dependencies.rs
estib-vega: > Probably not measurable by itself, but these things accumulate.

Totally reasonable, will change that.
 
In general, I agree that we should improve everywhere we can, as long as the maintainability (and in turn, readability) cost is not too high.

PR #5591, opened 2024-11-18
feat: handle error branches in the frontend

ndom91:
## ☕️ Reasoning

- Branch in stack before me added support for not blowing up the entire `list_virtual_branches` result if one thing errored. The error boundary is at the series level, so the `series` type turned  into `(PatchSeries | Error)[]` from the point of view of the frontend.

## 🧢 Changes

- Added transformation when parsing `branch.series` from `list_virtual_branches` into `VirtualBranch`
- Added `ErrorSeries` card for when a branch is errored
- Updated `SeriesLabels` in `StackHeader` to show errored series + red color

image

---
This is **part of a stack** made with GitButler:
- #5589 
- #5591 👈

### DIFF
--- a/apps/desktop/src/lib/branch/BranchLaneContextMenu.svelte
+++ b/apps/desktop/src/lib/branch/BranchLaneContextMenu.svelte
@@ -42,7 +42,7 @@
 		allowRebasing = branch.allowRebasing;
 	});
 
-	const allPrIds = $derived(branch.series.map((series) => series.prNumber).filter(isDefined));
+	const allPrIds = $derived(branch.validSeries.map((series) => series.prNumber).filter(isDefined));
 
 	async function toggleAllowRebasing() {
 		branchController.updateBranchAllowRebasing(branch.id, !allowRebasing);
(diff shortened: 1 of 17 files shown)


krlvi:
i wonder if we can display the error in some useful way so that it can be copy-pasted to discord or something

krlvi:
I guess the useful things that we wanna put in the error message are:
- the error message itself (and a copy button)
- link to discord or the in-app send feedback
- link to github issues?

Dunno if docs are helpful in this case but perhaps

krlvi:
Also to note that the only way for this state to be displayed is a bug in the app. So this is not a legitimate state as part of the UX. The alternative today is the "Something went wrong" message 😅

ndom91:
@krlvi update:

image

krlvi:
> > i agree, it can be too many links. i think its still better to direct people to discord + github issues because it seems more actionable. It's unlikely that a user would be able to fix a problem themselves with just our docs
> 
> Hmm well the debug docs page specifically does show folks where the data files are, the `projects.json`, etc. 🤔

fair enough. lets do docs.

krlvi (approved):

PR #5593, opened 2024-11-18
fix: bump svelte and related deps

ndom91:
## ☕️ Reasoning

- Bump svelte out of svelte-5 RC to GA release


apps/desktop/src/lib/components/MarkdownContent.svelte
```diff
@@ -23,6 +23,7 @@
 
 {#if (!type || type === 'init') && 'tokens' in rest && rest.tokens}
 	{#each rest.tokens as token}
+		<!-- eslint-disable-next-line svelte/valid-compile -->
```
ndom91, line 26: `svelte:self` still works, however, they've introduced some new syntax where you're supposed to:

```jsx
import Self from './filename.svelte'

<Self ... />
```

However, that introduces a few new type issues. Everything seems to still work as expected :ok_hand:

krlvi (approved):

PR #5596, opened 2024-11-19
Disable PR button on conflict.

estib-vega:
## ☕️ Reasoning

This pull request disables the PR button when a series contains a conflicted patch. This is to prevent users from creating a pull request when there are conflicts that need to be resolved first.

## 🧢 Changes

The main change in this PR is to disable the PR button if a series contains a conflicted patch. This ensures that users cannot create a pull request until the conflicts have been resolved.


mtsgrd (approved):
Awesome. 🙌

PR #5597, opened 2024-11-19
More PR monitoring fixes

estib-vega:
## ☕️ Reasoning

1. The PR service now keeps a map of PR monitors, ensuring that if multiple consumers ask for the same PR's monitor, the same object is returned. This fixes an issue where refreshing the monitor wouldn't propagate the update to every consumer.

2. Some PR checks are only listed asynchronously, so initially it might look like all checks have passed.
 The PR checks are now double-checked, to account for false positives that the GitHub API causes.

## 🧢 Changes

- The PR service now maintains a map of PR monitors to ensure consistent updates across consumers.
- The PR check validation process has been improved to handle asynchronous fetching of the check list, reducing the chances of false positives.

### DIFF
--- a/apps/desktop/src/lib/forge/github/githubPrService.ts
+++ b/apps/desktop/src/lib/forge/github/githubPrService.ts
@@ -16,11 +16,14 @@ import type { Octokit } from '@octokit/rest';
 
 export class GitHubPrService implements ForgePrService {
 	loading = writable(false);
+	private monitors: Map<number, GitHubPrMonitor>;
 
 	constructor(
 		private octokit: Octokit,
 		private repo: RepoInfo
-	) {}
+	) {
+		this.monitors = new Map();
+	}
 
 	async createPr({
 		title,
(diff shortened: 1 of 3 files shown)


apps/desktop/src/lib/forge/github/githubPrService.ts
```diff
@@ -94,7 +97,12 @@ export class GitHubPrService implements ForgePrService {
 	}
 
 	prMonitor(prNumber: number): GitHubPrMonitor {
-		return new GitHubPrMonitor(this, prNumber);
+		if (this.monitors.has(prNumber)) {
+			return this.monitors.get(prNumber)!;
+		}
+		const monitor = new GitHubPrMonitor(this, prNumber);
+		this.monitors.set(prNumber, monitor);
+		return monitor;
```
mtsgrd, line 105: We have no mechanism with which to clean up this the map, which means they won't get garbage collected.. Let's discuss over video? I actually just removed the duplicated pr monitor yesterday.

PR #5599, opened 2024-11-19
fix: File navigation in commits

estib-vega:
Commit files were not navigatable because the component was using an old implementation of the selection update method.

Also:
- Remove old, now unused file selection update method
- Add support for `j`, `k` navigation


ndom91 (approved):
LGTM; :raised_hands: for `j` and `k` haha

PR #5602, opened 2024-11-19
Improvew is integrated check

Caleb-T-Owens:
This commit does two things to improve the is_integrated check.

Firstly, it uses passes true to the log functions. This means that all branches of merge commits will be explored in the check for upstream integration.

It also now uses the change ID to see if a commit has been integrated.

---
This is **part 2 of 4 in a stack** made with GitButler:
- `4` #5608 
- `3` #5607 
- `2` #5602 👈 
- `1` #5601

### DIFF
--- a/crates/gitbutler-branch-actions/src/virtual.rs
+++ b/crates/gitbutler-branch-actions/src/virtual.rs
@@ -34,6 +34,7 @@ use gitbutler_stack::{
     Target, VirtualBranchesHandle,
 };
 use gitbutler_time::time::now_since_unix_epoch_ms;
+use itertools::Itertools;
 use serde::Serialize;
 use std::collections::HashSet;
 use std::{collections::HashMap, path::PathBuf, vec};
@@ -1006,6 +1007,7 @@ pub(crate) struct IsCommitIntegrated<'repo, 'cache, 'graph> {
     target_commit_id: gix::ObjectId,
     upstream_tree_id: gix::ObjectId,
     upstream_commits: Vec<git2::Oid>,
+    upstream_change_ids: Vec<String>,
 }
 
 impl<'repo, 'cache, 'graph> IsCommitIntegrated<'repo, 'cache, 'graph> {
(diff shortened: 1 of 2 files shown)


crates/gitbutler-branch-actions/src/virtual.rs
```diff
@@ -1020,35 +1022,58 @@ impl<'repo, 'cache, 'graph> IsCommitIntegrated<'repo, 'cache, 'graph> {
...
+        // Ensure upstream commits are sorted for binary search
+        upstream_commits.sort();
+        let upstream_change_ids = upstream_commits
+            .iter()
+            .filter_map(|oid| {
+                let commit = repository.find_commit(*oid).ok()?;
+                commit.change_id()
+            })
+            .sorted()
+            .collect::<Vec<_>>();
```
Byron: Here a bare `collect()` might even work as the type can be inferred due to the use in the struct that happens right after.

Caleb-T-Owens: You are indeed correct. Thank you!

krlvi (approved):

crates/gitbutler-branch-actions/src/virtual.rs
```diff
@@ -1020,35 +1022,58 @@ impl<'repo, 'cache, 'graph> IsCommitIntegrated<'repo, 'cache, 'graph> {
...
+
+        let upstream_commits =
+            ctx.repository()
+                .log(remote_head.id(), LogUntil::Commit(target.sha), true)?;
+        let upstream_change_ids = upstream_commits
+            .iter()
+            .filter_map(|commit| commit.change_id())
+            .sorted()
+            .collect();
+        let upstream_commits = upstream_commits
```
mtsgrd, line 1035: For consistency, would this be better named `upstream_commit_ids`, as to not have some `commit` variables be `git2::Oid`, and some `git2::Commit`?

mtsgrd (approved):

PR #5604, opened 2024-11-19
Add component for displaying page laod errors

mtsgrd:
To see it, you can e.g. do this in `apps/desktop/src/lib/vbranches/virtualBranch.ts`:

image

image


mtsgrd:
@PavelLaptev @ndom91 would something like this make sense to add while we're also working on isolating failures to sub component levels?

krlvi:
This is really neat. We could in theory advertise here our discord for support and the "project history" for going back in time

PavelLaptev:
A design touch is definitely needed here.

ndom91:
> @PavelLaptev @ndom91 would something like this make sense to add while we're also working on isolating failures to sub component levels?

Yeah I'd say so. Currently we've just set an error boundary on the rust side at the branch/series level. We only added the ability to render that boundary nicely on the FE side so far. 

Going forward I think some kind of higher level error boundary like this on the FE makes a lot of sense, and then if we want we can still add more tightly scoped boundaries on the FE if they make sense.

PavelLaptev:
@mtsgrd updated the design a little. Feel free to merge when you're ready.

image

---

_[Design in Figma](https://www.figma.com/design/ShIiR6hI5dzH7sT5L03Jg1/Design-3.0?node-id=2676-9649&t=EAw0BHSHczuxG1y7-1)_

PavelLaptev (approved):

PR #5607, opened 2024-11-19
Fix commit line colors

Caleb-T-Owens:
One problem was that we were over eager in assigning the remote_commit_id property, and ended up with remote commits getting the remote_commit_id set to themselves.

Another problem is that the commit line logic's integrated commit list was supposed to be mutually exclusive from the other arrays. I've added some extra logic to ensure that is the case

---
This is **part 3 of 4 in a stack** made with GitButler:
- `4` #5608 
- `3` #5607 👈 
- `2` #5602 
- `1` #5601

### DIFF
--- a/apps/desktop/src/lib/stack/SeriesList.svelte
+++ b/apps/desktop/src/lib/stack/SeriesList.svelte
@@ -89,7 +89,7 @@
 
 			{#if currentSeries.upstreamPatches.length > 0 || currentSeries.patches.length > 0}
 				<CommitList
-					remoteOnlyPatches={currentSeries.upstreamPatches.filter((p) => !p.relatedTo)}
+					remoteOnlyPatches={currentSeries.upstreamPatches}
 					patches={currentSeries.patches}
 					seriesName={currentSeries.name}
 					isUnapplied={false}
--- a/apps/desktop/src/lib/vbranches/types.ts
+++ b/apps/desktop/src/lib/vbranches/types.ts
@@ -204,7 +204,7 @@ export class VirtualBranch {
 
 // Used for dependency injection
 export const BRANCH = Symbol('branch');
-export type CommitStatus = 'local' | 'localAndRemote' | 'localAndShadow' | 'integrated' | 'remote';
+export type CommitStatus = 'local' | 'localAndRemote' | 'integrated' | 'remote';
 
 export class ConflictEntries {
 	public entries: Map<string, ConflictEntryPresence> = new Map();
@@ -234,6 +234,7 @@ export class DetailedCommit {
 	@Transform((obj) => new Date(obj.value))
 	createdAt!: Date;
 	isRemote!: boolean;
+	isLocalAndRemote!: boolean;
 	isIntegrated!: boolean;
 	parentIds!: string[];
 	branchId!: string;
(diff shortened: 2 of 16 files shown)


krlvi (approved):
seems reasonable

Caleb-T-Owens:
@krlvi this has turned out to be more complicated than just this change. We've been feeding bad data from the rust side, so I've gone in and resolved the root problem

crates/gitbutler-stack/src/stack_branch.rs
```diff
@@ -140,36 +138,18 @@ impl StackBranch {
             let head_commit = repository
                 .find_reference(&self.remote_reference(&remote_name)?)?
                 .peel_to_commit()?;
-            let target_commit = repository
-                .find_reference(default_target.branch.to_string().as_str())?
-                .peel_to_commit()?;
-            let merge_base = repository.merge_base(head_commit.id(), target_commit.id())?;
             repository
-                .log(head_commit.id(), LogUntil::Commit(merge_base), false)?
+                .log(head_commit.id(), LogUntil::Commit(previous_head), false)?
```
Caleb-T-Owens, line 142: Logging to the merge base includes extra commits from previous branches which we don't want.

crates/gitbutler-stack/src/stack_branch.rs
```diff
@@ -140,36 +138,18 @@ impl StackBranch {
...
-            .into_iter()
-            .rev() // oldest commit first
-            .collect_vec();
-        let mut upstream_only = vec![];
-        for patch in remote_patches.iter() {
-            if !local_patches_including_merge.contains_by_commit_or_change_id(patch) {
-                upstream_only.push(patch.clone());
-            }
-        }
-
```
Caleb-T-Owens, line 168: This set of commits is better determined in stack.rs where we have more context.

Otherwise we end up trying to do the same filtering twice, but just a little bit inconsistently

crates/gitbutler-branch-actions/src/virtual.rs
```diff
@@ -508,6 +458,13 @@ pub fn list_virtual_branches_cached(
             requires_force = force // derive force requirement from the series
         }
 
+        let commits = series
+            .iter()
+            .cloned()
+            .filter_map(Result::ok)
+            .flat_map(|s| s.patches)
+            .collect();
+
```
Caleb-T-Owens, line 418: There is still a little bit of frontend logic, and a load of tests which rely on this list of commits, so we can't fully get rid of it. That said, pulling it out of the patches rather than calculating it twice in two slightly different ways is certainly an improvement

apps/desktop/src/lib/vbranches/types.ts
```diff
@@ -282,12 +283,8 @@ export class DetailedCommit {
 
 	get status(): CommitStatus {
 		if (this.isIntegrated) return 'integrated';
```
Caleb-T-Owens, line 285: We want integrated to take priority as a stats. The rest should be mutually exclusive

apps/desktop/src/lib/stack/SeriesList.svelte
```diff
@@ -89,7 +89,7 @@
 
 			{#if currentSeries.upstreamPatches.length > 0 || currentSeries.patches.length > 0}
 				<CommitList
-					remoteOnlyPatches={currentSeries.upstreamPatches.filter((p) => !p.relatedTo)}
+					remoteOnlyPatches={currentSeries.upstreamPatches}
```
Caleb-T-Owens, line 92: This filter doesn't seem to make any sense to me

crates/gitbutler-branch-actions/src/virtual.rs
krlvi: <3 i love this!

crates/gitbutler-stack/src/stack_branch.rs
krlvi: this seems a 🤦 moment for me

krlvi (approved):
I have tried to carefully read through the code and it all makes sense to me. Thanks for the deep dive!

PR #5609, opened 2024-11-19
feat: add flatpak output and GHA build step

ndom91:
## ☕️ Reasoning

- Test Flatpak build

## 🧢 Changes

- Add flatpak manifest
- Add flatpak xml metadata file
- Add `build:flatpak` root package.json script
- Add GHA jobs for buliding and uploading flatpak output
	- Unfortunately this can't be totally parallel, because the flatpak relies on consuming the `.deb` output.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -94,6 +94,7 @@ jobs:
     outputs:
       platform: ${{ matrix.platform }}
       channel: ${{ env.channel }}
+      version: ${{ env.version }}
 
     steps:
       # Because GitHub broke perl installations sometime in 2022 on Windows.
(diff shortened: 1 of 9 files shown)


Milkfat32:
Hello

> Unfortunately this can't be totally parallel, because the flatpak relies on consuming the .deb output.

It is possible to build the flatpak from source. I have done this on a tauri app before. You need to use the flatpak-node-generator for the node dependencies and flatpak-cargo-generator for the rust dependencies. Then you can run the build in offline mode, since the flatpak builder cannot access the internet during build. All of this can also be automated in a workflow.

ndom91:
> Hello
> 
> > Unfortunately this can't be totally parallel, because the flatpak relies on consuming the .deb output.
> 
> It is possible to build the flatpak from source. I have done this on a tauri app before, both v1 and v2. You need to use the flatpak-node-generator for the node dependencies and flatpak-cargo-generator for the rust dependencies. Then you can run the build in offline mode, since the flatpak builder cannot access the internet during build. All of this can also be automated in a workflow.
> 

Hey thanks for chiming in. I know its possible, but reusing the .deb was much quicker and simpler. 

I'm curious though, are there any other advantages to building it again separately for flatpak? If you have any links for a tauri v2 flatpak build from scratch I'd love to take a look

Milkfat32:
Hmm I believe the biggest advantage would be the applying any flatpak specific patches. I know of one Tauri flatpak that has a patch that disables the tauri updater, since updating is handled by the flatpak package manager. Other than that, it is mostly licensing and reproducability that some people care deeply about.

I don't know of any Tauri v2 flatpaks built from source, the one I maintain is still v1 for now. But I have built the dev branch of v2 as a flatpak.
The only difference I found between v1 and v2 were all the dependencies I had to pull in. For tauri v2 all the dependencies already exist in the gnome sdk runtime, so it is much easier. The process is simply writing the build commands and the cached dependencies.
You can check out the [Cinny flatpak](https://github.com/flathub/in.cinny.Cinny) or [OpenGOAL flatpak](https://github.com/flathub/dev.opengoal.OpenGOAL) manifest files

ndom91:
@Milkfat32 awesome, thanks for sharing. Yeah removing the tauri-updater here would be great for example.

ndom91:
Closing this PR. See https://github.com/ndom91/gitbutler-flathub for updated Flatpak. There should be `.flatpak` artifacts to download from the GHA in that repository. Alternatively, we plan on PR-ing the updated contents in that repo to the pre-existing third-party flatpak `com.gitbutler.gitbutler` in flathub.

PR #5610, opened 2024-11-19
fix: integrated state commits

ndom91:
## ☕️ Reasoning

- Edge case to ensure integrated commits are shown as integrated (when they're also potentially localAndRemote)


Caleb-T-Owens (changes requested):
I've got a better fix in the pipeline which prevents the mislabeling of data in the first place

PR #5611, opened 2024-11-19
Parse url for base branch and pr bases

mtsgrd:
- will help us compare pr base against target

---
- <kbd>&nbsp;2&nbsp;</kbd> #5612 
- <kbd>&nbsp;1&nbsp;</kbd> #5611 👈

### DIFF
--- a/apps/desktop/src/lib/stores/uniqueStore.ts
+++ b/apps/desktop/src/lib/stores/uniqueStore.ts
@@ -1,0 +1,75 @@
+import { writable, type Readable } from 'svelte/store';
+
+export function uniqueDerived<T>(source: Readable<T>) {
+	return new UniqueDerived<T>(source);
+}
+
+/**
+ * A derived Svelte store that only emits new values if the objects differ
+ * from each other using `shallowCompare`.
+ */
+class UniqueDerived<T> implements Readable<T> {
+	/**
+	 * Previously set value used for comparison with new values.
+	 */
+	private prev: unknown;
+
+	/**
+	 * A writable that sets up the parent subscription when it starts, and
+	 * unsubscribes when last subscriber leaves.
+	 */
+	private store = writable<T>(undefined, (set) => {
+		const unsubscribe = this.parent.subscribe((value) => {
+			if (!shallowCompare(this.prev, value)) {
+				set(value);
+			}
+			this.prev = value;
+		});
+		return () => {
+			unsubscribe();
+		};
+	});
+
+	constructor(private parent: Readable<T>) {}
+
+	/**
+	 * Subscribe call is forwarded to our writable instance.
+	 */
+	subscribe(run: (value: T) => void, invalidate?: () => void): () => void {
+		return this.store.subscribe(
+			(value) => {
+				run(value);
+			},
+			() => {
+				invalidate?.();
+			}
+		);
+	}
+}
+
+function shallowCompare(left: unknown, right: unknown): boolean {
+	if (left === right) {
+		return true;
+	}
+
+	if (typeof left !== 'object' || typeof right !== 'object' || left === null || right === null) {
+		return false;
+	}
+
+	const keys1 = Object.keys(left);
+	const keys2 = Object.keys(right);
+
+	// Compare key lengths
+	if (keys1.length !== keys2.length) {
+		return false;
+	}
+
+	// Compare values for each key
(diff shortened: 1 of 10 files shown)


mtsgrd:
@Caleb-T-Owens what do you think of this the `UniqueDerived` class? I think we discussed this last week.

krlvi (approved):
didnt properly review this but if it sucks we can fix it tomorro

apps/desktop/src/lib/stores/uniqueStore.ts
```diff
@@ -0,0 +1,75 @@
+import { writable, type Readable } from 'svelte/store';
+
+export function uniqueDerived<T>(source: Readable<T>) {
```
Caleb-T-Owens, line 3: I'm not sure the class is helpful, can't we just say:
```ts
export function shallowDeduplicate<A>(store: Readable<A>): Readable<A> {
	const deduplicatedStore = readable(get(store), (set) => {
		let lastValue = get(store);
		const unsubscribe = store.subscribe((value) => {
			if (!shallowCompare(lastValue, value)) {
				lastValue = value;
				set(value);
			}
		});
		return unsubscribe;
	});

	return deduplicatedStore;
}
```

apps/desktop/src/lib/stores/uniqueStore.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+function shallowCompare(left: unknown, right: unknown): boolean {
+	if (left === right) {
+		return true;
+	}
+
+	if (typeof left !== 'object' || typeof right !== 'object' || left === null || right === null) {
+		return false;
+	}
```
Caleb-T-Owens, line 57: undefined needs to be considered here

apps/desktop/src/lib/utils/url.test.ts
```diff
@@ -25,22 +25,6 @@ describe.concurrent('cleanUrl', () => {
...
-
-	const nonHttpRemoteUrls = [
-		'git@github.com:user/repo.git',
-		'ssh://git@github.com:22/user/repo.git',
-		'git://github.com/user/repo.git'
-	];
-
-	test.each(nonHttpRemoteUrls)('Non HTTP Remote - %s', (remoteUrl) => {
-		expect(remoteUrlIsHttp(remoteUrl)).toBe(false);
-	});
```
Caleb-T-Owens, line 43: Why don't we want these tests?

Caleb-T-Owens:
> didnt properly review this but if it sucks we can fix it tomorro

I'm bothered by this. Especially when I was asked for feedback.

apps/desktop/src/lib/stores/uniqueStore.ts
mtsgrd: Let me try that out. *edit: one immediate thing I would want to avoid is subscribing to the store when creating the derived store, which happens here because of the `get()` method.

apps/desktop/src/lib/utils/url.test.ts
mtsgrd: Because they existed for exactly 1 location in the code asking "is the protocol http?". The if condition can just as well be `protocol.startsWith('http')` rather than having a specialized and tested function.

apps/desktop/src/lib/stores/uniqueStore.ts
Caleb-T-Owens: You should be safe to set the readable initial value to undefined, and keep the result as `Readable<T>` because the start stop notifier gets run synchronously

Caleb-T-Owens: Hmm... or even write it as a derived to make that intent clearer

Caleb-T-Owens:
```ts
export function shallowDeduplicate<A>(store: Readable<A>): Readable<A> {
	let lastValue: undefined | A = undefined;
	const deduplicatedStore = derived<Readable<A>, A>(store, (value, set) => {
			if (!shallowCompare(lastValue, value)) {
				lastValue = value;
				set(value);
			}
	});

	return deduplicatedStore;
}
```

Caleb-T-Owens: AHHhhhh no... that derived version isn't safe

PR #5613, opened 2024-11-19
Use stable #each key in SeriesList component

mtsgrd:


mtsgrd:
This doesn't seem to work..

krlvi (approved):

PR #5619, opened 2024-11-19
Add option for creating new pr

mtsgrd:
- we automatically re-attach open prs, so we only allow detaching closed prs


Caleb-T-Owens (approved):
Looks good to me. `pr` probably should be capitalised as `PR`, or even spelled out in full.

PR #5620, opened 2024-11-19
Don't unapply other branches when applied branch conflicts with base

mtsgrd:
This worked for me when applying an old branch that conflicted with the current base. I'm unsure if there are other scenarios to account for?


mtsgrd:
@Caleb-T-Owens is there some way we could avoid unapplying other branches if the branch being applied interferes only with the base, not the other lanes?

Caleb-T-Owens:
Ohhh... I see.

I'm not sure. The issue is that you'd more or less have to see if it gets _more_ conflicted.

PR #5627, opened 2024-11-20
ai: Pass the correct file hunks when editing a commit message

estib-vega:
Give the commit message generation method the file changes of the commit that's being edited, instead of the uncommited files

### DIFF
--- a/apps/desktop/src/lib/commit/CommitMessageInput.svelte
+++ b/apps/desktop/src/lib/commit/CommitMessageInput.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 	import { PromptService } from '$lib/ai/promptService';
-	import { AIService } from '$lib/ai/service';
+	import { AIService, type DiffInput } from '$lib/ai/service';
 	import { Project } from '$lib/backend/projects';
 	import ContextMenuItem from '$lib/components/contextmenu/ContextMenuItem.svelte';
 	import ContextMenuSection from '$lib/components/contextmenu/ContextMenuSection.svelte';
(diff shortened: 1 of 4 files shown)


apps/desktop/src/lib/commit/CommitMessageInput.svelte
```diff
@@ -68,10 +71,30 @@
...
 	}
 
-	async function generateCommitMessage(files: LocalFile[]) {
-		const hunks = files.flatMap((f) =>
-			f.hunks.filter((h) => $selectedOwnership.isSelected(f.id, h.id))
-		);
+	async function generateCommitMessage() {
+		let diffInput: DiffInput[] | undefined = undefined;
+
+		if (existingCommit) {
```
ndom91: Little nit, but what about somethign like this instead maybe?

```ts
	const files = await listCommitFiles(project.id, existingCommit.id);
	const diffInput = existingCommit
	 ? files.flatMap((file) =>
			file.hunks.map((hunk) => ({
				filePath: file.path,
				diff: hunk.diff
			}))
		)
	: $branch.files.flatMap((f) =>
			f.hunks
				.filter((h) => $selectedOwnership.isSelected(f.id, h.id))
				.map((h) => ({
					filePath: f.path,
					diff: h.diff
				}))
		);
```

ndom91 (approved):
Small suggestion, but otherwise LGTM. Works for here :ok_hand:

apps/desktop/src/lib/commit/CommitMessageInput.svelte
estib-vega: The issue there is that I want to call the `listCommitFiles` API only if the `existingCommit` is present.

How about we split the difference?
```ts
	async function getDiffInput(): Promise<DiffInput[]> {
		if (!existingCommit) {
			return $branch.files.flatMap((f) =>
				f.hunks
					.filter((h) => $selectedOwnership.isSelected(f.id, h.id))
					.map((h) => ({
						filePath: f.path,
						diff: h.diff
					}))
			);
		}

		const files = await listCommitFiles(project.id, existingCommit.id);
		return files.flatMap((file) =>
			file.hunks.map((hunk) => ({
				filePath: file.path,
				diff: hunk.diff
			}))
		);
	}

	async function generateCommitMessage() {
		const diffInput = await getDiffInput();
         //...
```

PR #5632, opened 2024-11-20
Fix remote commit listing when bottom commit is amended

mtsgrd:
- should list what's between current branch remote sha and previous branch remote sha

**Before**

image

**After**

image

---
- <kbd>&nbsp;2&nbsp;</kbd> #5632 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #5631

### DIFF
--- a/crates/gitbutler-stack/src/stack_branch.rs
+++ b/crates/gitbutler-stack/src/stack_branch.rs
@@ -140,12 +140,30 @@ impl StackBranch {
             .map(|ref_name| ref_name.remote().to_owned())
             .unwrap_or(default_target.push_remote_name());
 
+        let predecessor_branch = stack
+            .branch_predacessor(self)
+            .map(|p| p.name.to_owned())
+            .unwrap_or(default_target.branch.branch().to_owned());
+
+        let predecessor_ref = format!("refs/remotes/{}/{}", remote, predecessor_branch);
+        let previous_branch_head = match repository.find_reference(predecessor_ref.as_str()) {
+            Ok(reference) => reference
+                .peel_to_commit()
+                .map(|r| r.id())
+                .unwrap_or(previous_head),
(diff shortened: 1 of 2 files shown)


mtsgrd:
@krlvi this appears to fix the happy path problem.

crates/gitbutler-stack/src/stack_branch.rs
```diff
@@ -140,12 +140,30 @@ impl StackBranch {
             .map(|ref_name| ref_name.remote().to_owned())
             .unwrap_or(default_target.push_remote_name());
 
+        let predecessor_branch = stack
+            .branch_predacessor(self)
+            .map(|p| p.name.to_owned())
+            .unwrap_or(default_target.branch.branch().to_owned());
+
+        let predecessor_ref = format!("refs/remotes/{}/{}", remote, predecessor_branch);
```
krlvi, line 148: Why not use https://github.com/gitbutlerapp/gitbutler/blob/ea61bf06b4636f424aef8efa1af1f539bc6878e4/crates/gitbutler-stack/src/stack_branch.rs#L86 ?

mtsgrd: Because that takes a remote name and gives back a ref for the current branch. We need the remote ref for the previous branch.

krlvi:
how does it look like when the bottom branch is merged?

mtsgrd:
I got ahead of myself here. After merging and updating base we still have the happy path problem. What this fixes is the same symptom, but it appears after you amend a commit at the bottom.

krlvi:
can you join our call?

crates/gitbutler-stack/src/stack_branch.rs
krlvi: the remote head may not exist if deleted...

PR #5636, opened 2024-11-20
When applying a branch always rebase if there are no change IDs

krlvi:
This change makes it so that when a branch is applied, if the head does not have a change ID (i.e. was not created with gitbutler), the commits will be rebased even if the base is the same as the workspace.
This will populate the change ID headers. 

This resolves issues related to stacking which relies on change IDs


Caleb-T-Owens (approved):
LGTM

PR #5637, opened 2024-11-20
gitbutler-feeds

Caleb-T-Owens:
---
- <kbd>&nbsp;2&nbsp;</kbd> #5637 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #5588


apps/desktop/src/lib/posts/Post.svelte
```diff
@@ -0,0 +1,41 @@
...
+	};
+
+	const { postId }: Props = $props();
+
+	const feedService = getContext(FeedService);
+
+	const store = useStore();
+	const postWithRepliesInterest = $derived(feedService.getPostWithRepliesInterest(postId));
+	const post = $derived(postSelectors.selectById($store, postId));
+	const postReplies = $derived(postRepliesSelectors.selectById($store, postId));
```
mtsgrd: I'm wondering in how many instances of components we might end up reacting to `$store`, and if that could ever become problematic?

Caleb-T-Owens: That seems to be the point of selectors; You're reducing the amount of updates to the data that gets passed to subsequent derive statements.

mtsgrd: I had a look in the documentation and found this: https://redux.js.org/usage/deriving-data-selectors#globalize-selectors-if-needed

Since we already have a large app it feels like we should default to localized selectors, and use globalized ones only where needed?

Caleb-T-Owens: Sounds like a good idea 👍

mtsgrd: How can we implement that here?

packages/shared/src/lib/redux/store.ts
```diff
@@ -9,7 +10,10 @@ import { configureStore } from '@reduxjs/toolkit';
  */
 export const _store = configureStore({
 	reducer: {
-		example: exampleReducer
+		example: exampleReducer,
+		post: postReducer,
```
mtsgrd: Since this state contains an array of posts, should we perhaps use plural?

I asked chatgtp:
> When naming state slices in a Redux store that hold arrays of objects, it's generally better to use plural names. This aligns with the idea that the state slice represents a collection of items. Using plural names makes it immediately clear that the state contains multiple entities of a given type.
>
>Why Plural is Preferred:
>- Clarity: Plural names suggest collections, while singular names often imply a single item.
>- Consistency: If other parts of your application follow this pattern, it reduces cognitive load.
>- Convention: The broader JavaScript and Redux community commonly follows this practice.

Caleb-T-Owens: > I asked chatgtp:

Do we have a real source?
Text generators can not be used as a reliable form of information in any shape of form.

For example, I prompted the Chat GPT the same prompt, but switched "plural" for "singular":
Screenshot 2024-11-21 at 19 13 42

Screenshot 2024-11-21 at 19 13 19

Caleb-T-Owens: https://redux.js.org/style-guide#name-state-slices-based-on-the-stored-data

Redux's style guide doesn't explicitly mention plural or singular, but does implicitly indicate to use the plural form in their examples, so we can switch over to that.

Byron: This is a very interesting outcome, I wasn't aware! GPT seems to be an amplifier for confirmation bias. This makes sense given that the program wants to give satisfying responses, to please the user.

packages/shared/src/lib/interest/intrestStore.ts
```diff
@@ -0,0 +1,68 @@
...
+				if (!subscription) {
+					subscription = {
+						args,
+						counter: 0,
+						lastCalled: 0,
+						interval: undefined
+					} as Subscription<Arguments>;
+				}
+
+				this.subscriptions.push(subscription);
```
mtsgrd: When a previous subscription was found on lines 21:23 this line will push a duplicate. Maybe move this to line 32?

packages/shared/src/lib/interest/intrestStore.ts
```diff
@@ -0,0 +1,68 @@
...
+				// Fetch data immediately on first subscription
+				if (subscription.counter === 0) {
+					if (Date.now() - subscription.lastCalled > this.frequency) {
+						subscription.lastCalled = Date.now();
+						callback();
+					}
+
+					subscription.interval = setInterval(() => {
+						subscription.lastCalled = Date.now();
+						callback();
```
mtsgrd: The interval is meant to be shared by concurrent subscribers right? The `callback` argument to this function is ignored unless `subscription.counter === 0`.

packages/shared/src/lib/interest/intrestStore.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+interface Subscription<Arguments> {
+	args: Arguments;
+	counter: number;
+	interval?: ReturnType<typeof setInterval>;
+	lastCalled: number;
+}
+
+export class InterestStore<Arguments> {
```
mtsgrd: Could we get a description for this class? I'm also curious if there might be a more descriptive name you can give this. Even after becoming familiar with the code I still get a bit confused when I see `something.createInterest(...)`.

packages/shared/src/lib/interest/RegisterInterest.svelte
```diff
@@ -0,0 +1,45 @@
...
+			return () => {
+				inView = false;
+				observer.disconnect();
+			};
+		}
+	});
+
+	$effect(() => {
+		const shouldSubscribe = !onlyInView || inView;
+		if (interest && shouldSubscribe) {
```
mtsgrd, line 38: Variable `interest` will according to types never be falsy.

packages/shared/src/lib/redux/store.ts
mtsgrd: @Caleb-T-Owens it's worth keeping in mind I did not ask a leading question, I simply asked which one it preferred, and it gave me pros and cons for both and a verdict.

packages/shared/src/lib/interest/intrestStore.ts
Caleb-T-Owens: Thanks, this is indeed a subtle problem

packages/shared/src/lib/interest/intrestStore.ts
Caleb-T-Owens: If the arguments are the same, then the callback should also be the same definition.


That is however implicit and not enforced. I'll think about how to restructure this in such a way that it's certain that we have just one callback per subscription.

packages/shared/src/lib/interest/RegisterInterest.svelte
Caleb-T-Owens: Correct. At one point in time it was possibly undefined, but that is no longer true.

packages/shared/src/lib/interest/intrestStore.ts
```diff
@@ -0,0 +1,68 @@
+import { shallowCompare } from '$lib/shallowCompare';
+
+export interface Interest {
+	_subscribe: () => () => void;
+}
+
+interface Subscription<Arguments> {
```
mtsgrd: Could we use `T` here instead of `Arguments` since it's standard practice?

Caleb-T-Owens: Is it a good standard practice?

Caleb-T-Owens: Taking a look at some of the type definitions in redux toolkit, they seem to use named type parameters in the vast majority of cases. The typescript documentation doesn't give a recommendation, but does avoid using single letter type parameter names in their examples. https://www.typescriptlang.org/docs/handbook/2/functions.html#guidelines-for-writing-good-generic-functions

mtsgrd (approved):

PR #5639, opened 2024-11-21
Weave the remote branch into the local one

estib-vega:
Add an experimental feature:

When integrating a remote branch into a local patch series, interlock the incoming commits in order to reduce the probability of conflicts on integrate.
The interlocking order is determined by matching Change IDs.

### Benefits
1. This will usually yield less conflicts
2. This will usually yield less duplicated commits
3. The commits will end-up in roughly the desired order (no need to reorder and squashing)

### Things to look-out for
This algorithm works best if all change IDs are present, and all change IDs are uniquely matched to commits.
If that's not the case, the algorithm will still work, but it will just do some guess-work.

Nothing should break. But if we catch some weird decisions in the ordering of some commits, we should try to improve it.

### Visuals
There seems to be a noticeable improvement in some cases:

Before
before

Current integration algorithm
current

New algo
zipping

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2969,6 +2969,7 @@ dependencies = [
  "gitbutler-branch",
  "gitbutler-branch-actions",
  "gitbutler-command-context",
+ "gitbutler-commit",
  "gitbutler-oxidize",
  "gitbutler-project",
  "gitbutler-reference",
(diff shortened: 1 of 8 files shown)


Caleb-T-Owens (approved):
Love this!

PR #5653, opened 2024-11-22
Fix gitlab and bitbucket branch comparison URLs

mtsgrd:
- we don't want to provide fork information to forges unless actually forked

### DIFF
--- a/apps/desktop/src/lib/baseBranch/baseBranchService.ts
+++ b/apps/desktop/src/lib/baseBranch/baseBranchService.ts
@@ -20,17 +20,11 @@ export class BaseBranchService {
 	readonly base = shallowDeduplicate(this._base);
 
 	// Deduplicated repo information.
-	readonly repo = shallowDeduplicate(
-		derived(this.base, (base) => {
-			return base ? parseRemoteUrl(base.remoteUrl) : undefined;
-		})
-	);
+	readonly repo = shallowDeduplicate(derived(this.base, (base) => parseRemoteUrl(base?.remoteUrl)));
 
 	// Deduplicated push repo information.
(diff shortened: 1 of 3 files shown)


apps/desktop/src/lib/baseBranch/baseBranchService.ts
```diff
@@ -29,7 +29,10 @@ export class BaseBranchService {
 	// Deduplicated push repo information.
 	readonly pushRepo = shallowDeduplicate(
 		derived(this.base, (base) => {
-			return base ? parseRemoteUrl(base.pushRemoteUrl) : undefined;
+			// TODO: We should not send `pushRemoteUrl` if it's the same as `remoteUrl`.
+			if (base?.pushRemoteUrl && base.pushRemoteUrl !== base.remoteUrl) {
+				return parseRemoteUrl(base.pushRemoteUrl);
```
estib-vega: Is the comment still needed?
It seems like the condition already is taking care of what your TODO describes

estib-vega (approved):
Only that question about the comment

apps/desktop/src/lib/baseBranch/baseBranchService.ts
mtsgrd: The comment is about the back end sending identical remoteUrl and pushRemoteUrl. I think we should only send pushRemoteUrl if it's different from remoteUrl, but I also think we should kick that can down the road.

mtsgrd: Let me clarify the comment!

estib-vega: Gotcha 👍

PR #5654, opened 2024-11-22
Warn the user about not deleting their branches

estib-vega:
## ☕️ Reasoning

This pull request adds a warning to branch-stacking users who are using a forge repository that does not automatically delete branches after merging them.

## 🧢 Changes

- Created a `ForgeRepoService` that retrieves information about a given repository, specifically whether it automatically deletes branches on merge.
- Added a warning modal that is displayed to users if the repository they are using does not automatically delete branches on merge.
- The warning modal is only shown once.

### Visuals 👀
Screenshot 2024-11-22 at 17 05 52

### DIFF
--- a/apps/desktop/src/lib/config/config.ts
+++ b/apps/desktop/src/lib/config/config.ts
@@ -5,6 +5,13 @@ export function projectHttpsWarningBannerDismissed(projectId: string): Persisted
 	return persisted(false, key + projectId);
 }
 
+export function projectDeleteBranchesOnMergeWarningDismissed(
+	projectId: string
+): Persisted<boolean> {
+	const key = 'projectDeleteBranchesOnMergeWarningDismissed_';
+	return persisted(false, key + projectId);
+}
+
 export function projectCommitGenerationExtraConcise(projectId: string): Persisted<boolean> {
 	const key = 'projectCommitGenerationExtraConcise_';
 	return persisted(false, key + projectId);
--- a/apps/desktop/src/lib/forge/azure/azure.ts
+++ b/apps/desktop/src/lib/forge/azure/azure.ts
@@ -1,6 +1,7 @@
 import { AzureBranch } from './azureBranch';
 import type { RepoInfo } from '$lib/url/gitUrl';
 import type { Forge, ForgeName } from '../interface/forge';
+import type { ForgeRepoService } from '../interface/forgeRepoService';
 import type { ForgeArguments } from '../interface/types';
 
 export const AZURE_DOMAIN = 'dev.azure.com';
@@ -42,6 +43,10 @@ export class AzureDevOps implements Forge {
 	}
 
 	prService() {
+		return undefined;
+	}
+
+	repoService(): ForgeRepoService | undefined {
 		return undefined;
 	}
 
--- a/apps/desktop/src/lib/forge/bitbucket/bitbucket.ts
+++ b/apps/desktop/src/lib/forge/bitbucket/bitbucket.ts
@@ -49,6 +49,10 @@ export class BitBucket implements Forge {
 		return undefined;
 	}
 
+	repoService() {
+		return undefined;
+	}
+
 	checksMonitor(_sourceBranch: string) {
 		return undefined;
 	}
--- a/apps/desktop/src/lib/forge/github/github.ts
+++ b/apps/desktop/src/lib/forge/github/github.ts
@@ -2,6 +2,7 @@ import { GitHubBranch } from './githubBranch';
 import { GitHubChecksMonitor } from './githubChecksMonitor';
 import { GitHubListingService } from './githubListingService';
 import { GitHubPrService } from './githubPrService';
+import { GitHubRepoService } from './githubRepoService';
 import { GitHubIssueService } from '$lib/forge/github/issueService';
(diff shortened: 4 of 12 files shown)


ndom91 (approved):
LGTM :ok_hand: 

At first I thoguht, "not another service" :see_no_evil: but actually makes sense here. The "global" repo settings don't really fit into any of the existing ones, right. Also `projectDeleteBranchesOnMergeWarningDismissed` :joy:

apps/desktop/src/lib/components/Board.svelte
```diff
@@ -135,6 +173,29 @@
...
+		bind:this={deleteOnMergingWarningModal}
+		title="Branches are not deleted automatically"
+		type="warning"
+		width="small"
+		onSubmit={acknowledgeWarning}
+	>
+		{#snippet children()}
+			<p class="text-13 text-body helper-text">
+				It seems that branches are <strong>not automatically deleted after being merged</strong>.
+				<br />
```
ndom91: We need this `<br />`? I think we can keep the preceding and succeeding lines together

PavelLaptev:
@estib-vega thanks. I'll check it today

estib-vega:
> LGTM 👌
> 
> At first I thoguht, "not another service" 🙈 but actually makes sense here. The "global" repo settings don't really fit into any of the existing ones, right. Also `projectDeleteBranchesOnMergeWarningDismissed` 😂

Yeah, I also don't love it. 
I opted for consistency.  I didn't want to have one 'domain' be integrated into another one, and the rest still be separate.

Hopefully we can find a nice way of merging this into one service or set of redux selectors 👀

apps/desktop/src/lib/components/Board.svelte
```diff
@@ -135,6 +173,28 @@
...
+		onSubmit={acknowledgeWarning}
+	>
+		{#snippet children()}
+			<p class="text-13 text-body helper-text">
+				It seems that branches are <strong>not automatically deleted after being merged</strong>.
+				This may lead to unexpected behavior when using stacks.
+				<br />
+				<br />
+				We <strong>strongly</strong> recommend you to configure your repository to automatically delete
+				branches after they are merged.
```
mtsgrd: I was thinking it could be good to portray this with a little less warning, something like this?

> Configuration required for stacked branches!
> 
> We have detected your forge is currently not configured to delete branches after they are merged, please consider doing so now.
> 
> Pull requests that depend on one another will not be automatically updated when the parent is merged if you continue without this configuration.

mtsgrd: Uhh, not sure I like my own suggestion. What I'm getting at is we should probably introduce the user to stacking and explain this config is needed, instead of the first experience being a stern warning?

krlvi:
I feel this is potentially the wrong timing to show this info. 
Doing so after the PR creating is likely a better timing. Why? You don't want to get interrupted after just having started. 
After creating a PR you are likely "available" to look at GitHub settings

krlvi:
Another thing - we should frame this suggestion around GitHub (since it is about GitHub), and provide some actionable information. I have some docs that ive been preparing, we can link to that.

Regarding copy, how about something like:

```
Branch deletion after merge

After a PR has been merged on GitHub, it's branch should be deleted in order for the next PR to be updated by GitHub to point to your target branch. 
You can configure GitHub to do this automatically <link> or alternatively delete the branch after merging.

[Ok]
```

estib-vega:
> I feel this is potentially the wrong timing to show this info. Doing so after the PR creating is likely a better timing. Why? You don't want to get interrupted after just having started. After creating a PR you are likely "available" to look at GitHub settings

Makes sense, but in that case it makes more sense to display it on creation, and not after IMO.
Screenshot 2024-11-24 at 11 13 36

The reasoning for not using a modal:
- A 'information' modal after creation can be easily missed/clicked away by accident.
- We already have a warning modal before creating the PR, if creating a PR in the middle of the stack.
- I agree it should not be a scary warning, but it should be really in the face of the user at least once.

That said, I defer further UI/UX judgement to whatever @PavelLaptev thinks is best :)

PavelLaptev:
I like how it's done on the "Integration" page, where users can check if GitHub is integrated into the app.

Here’s a UX question: what happens after the user configures GitHub? When they return to the app, they still see the same info message. Shouldn't there be an action like "Refetch GitHub configuration"?

For now, a simple static info message might work better for a quick implementation.

@estib-vega, I suggest moving the info message below all the fields, just above the "Create PR" button. Let’s also add a "Configure GitHub" button.

Wdyt?

image

estib-vega:
Done 👍 
Screenshot 2024-11-25 at 09 37 40


> Here’s a UX question: what happens after the user configures GitHub? When they return to the app, they still see the same info message. Shouldn't there be an action like "Refetch GitHub configuration"?

Maybe. I would think is unnecessary in this case because:
- The warning will automatically disappear the next time that the PR modal is opened if the user configured GitHub.
- It's not a blocking action, at least not at the same level of the integration of GitHub.

I don't feel strongly against it, though

PavelLaptev (approved):

PavelLaptev:
> Done 👍

thanks @estib-vega

PR #5666, opened 2024-11-25
Show warning if a stacked branch was incorrectly merged

mtsgrd:
image


PavelLaptev (approved):
@mtsgrd Thanks! I updated the design a little.  
I set the icons to default for the `LinkButton` and changed the `outline` prop for the `info-message` to `false`, since we use either filled or outline in the design, but not both at the same time.

image

PR #5667, opened 2024-11-25
Run prettier on markdown file changes

mtsgrd:


mtsgrd:
@ndom91 do you have any thoughts on how we can best run prettier on *.md changes outside of the node stuff?

ndom91:
> @ndom91 do you have any thoughts on how we can best run prettier on *.md changes outside of the node stuff?

"Outside of node stuff" meaning separately only when *.md files have been changed? I think your solution here is fine. 

Maybe an easier way to ensure we dont end up with "mis-formatted" files in `master`, would be to have a git pre-commit hook to run prettier / eslint on all staged stuff. Is there some reason we don't have that yet actually?

ndom91 (approved):

PR #5668, opened 2024-11-25
fix: PR creation order checks only unarchived series

estib-vega:
Whenever we check if all previous series have a PR associated with them, check only unarchived series.


mtsgrd:
Since `allPreviousSeriesHavePrNumber` is only used in `SeriesHeader`, perhaps we shouldn't make it part of the class but instead just have it in the svelte file or some utility ts file?

mtsgrd (approved):

estib-vega:
> Since `allPreviousSeriesHavePrNumber` is only used in `SeriesHeader`, perhaps we shouldn't make it part of the class but instead just have it in the svelte file or some utility ts file?

I'll create an utility function 👍

PR #5670, opened 2024-11-25
macos: Settings menu can be opened with `cmd` + `,`

estib-vega:


crates/gitbutler-tauri/src/menu.rs
```diff
@@ -68,11 +68,16 @@ pub fn build<R: Runtime>(handle: &AppHandle<R>) -> tauri::Result<tauri::menu::Me
...
+    let settings_menu = MenuItemBuilder::with_id("global/settings", "Settings")
+        .accelerator("CmdOrCtrl+,")
+        .build(handle)?;
+
     #[cfg(target_os = "macos")]
     let mac_menu = &SubmenuBuilder::new(handle, app_name)
         .about(Some(AboutMetadata::default()))
         .separator()
-        .text("global/settings", "Settings")
+        .item(&settings_menu)
```
Caleb-T-Owens, line 80: why does this make a difference?

Caleb-T-Owens (approved):

crates/gitbutler-tauri/src/menu.rs
estib-vega: You can't add the keyboard shortcut to the inlined `text` menu entry. At lease there's no way to do so that I could find.

PR #5672, opened 2024-11-25
FileWatcher: only ignore files if they aren't tracked.

### LINKED ISSUE #5662: git-ignored but *tracked* files don't show up if changed
### Version

0.13.17

### Operating System

macOS

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

Gitignored but tracked files don't show up as changed even if they are changed.

This is probably related to the file-watcher indiscriminately ignoring files that are ignored, as it doesn't check if they are tracked.

### How to reproduce

* create file `a` and commit it so it's *tracked*
* `echo change > a`
* `echo a > .gitignore`

`a` will not show up in the list of files changed in the worktree anymore, even though it did right after changing it. The only file visible is `.gitignore`

### Expected behavior

It should keep showing tracked files even if they are ignored.

### Relevant log output

_No response_

Byron:
This also means that tracked files should always be checked, ignored or not.

Fixes #5662.

### Tasks

* [x] re-invigorate the formatting (even though it still kind of doesn't work for `file_monitor.rs`
    - This usually happens if the nesting level is so high that `cargo fmt` gives up or has actual errors.
* [x] never ignore tracked files

### Notes for the Reviewer

* The formatting changes are mostly made in an attempt to get the formatting to work again. Thus some lines were shortened to the minimum line width (but might still be too long).
* One commit is just for formatting with nightly, which makes imports more consistent in the crate I touched. It's definitely not needed for the fix but since this crate is very stable, I thought it doesn't hurt.
* I only tested this by hand like so (before the PR)
    - change a tracked file
    - add that file to `.gitignore`
    - notice that it either shows up permanently (even after undoing the change) or never shows up as changed
    - with this PR applied, changed files always show up, ignored or not.
    - alternatively, add a file that isn't tracked to the `.gitignore` file to see it disappear/not appear in the UI

### DIFF
--- a/crates/gitbutler-watcher/src/file_monitor.rs
+++ b/crates/gitbutler-watcher/src/file_monitor.rs
@@ -150,10 +150,18 @@ pub fn spawn(
                     {
                         if let Ok(repo) = gix::open(&worktree_path) {
                             if let Ok(index) = repo.index_or_empty() {
-                                if let Ok(mut excludes) = repo.excludes(&index, None, gix::worktree::stack::state::ignore::Source::WorktreeThenIdMappingIfNotSkipped) {
+                                if let Ok(mut excludes) = repo.excludes(
+                                    &index,
+                                    None,
+                                    gix::worktree::stack::state::ignore::Source::WorktreeThenIdMappingIfNotSkipped,
+                                ) {
                                     for (file_path, kind) in classified_file_paths.iter_mut() {
                                         if let Ok(relative_path) = file_path.strip_prefix(&worktree_path) {
-                                            if excludes.at_path(relative_path, None).map(|platform| platform.is_excluded()).unwrap_or(false) {
+                                            if excludes
+                                                .at_path(relative_path, None)
+                                                .map(|platform| platform.is_excluded())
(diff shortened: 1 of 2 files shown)


Byron:
There might be a misunderstanding here, as the change will still ignore `node_modules` and the likes, unless somebody checks these in.

Previously, GitButler wouldn't show changes to these files even though `git status` would, which is surprising. I have also seen the opposite case, where a change was removed locally, but the UI still showed changes (it missed the event that changed the file).

> With that said I thought that the app already watches everything recursively under the repo root?

Yes, definitely.

Leaving it as is definitely feels wrong in comparison to Git. Maybe it's minor issue as well as most of the time, ignored files aren't tracked in the repository. However, I don't think it's normal to track (`git add -f <db-file>`) frequently changed possibly large files, and even if that's happening, people should be able to create commits with GitButler, I think.

https://github.com/user-attachments/assets/efb6deac-990f-4d10-b510-2b9998f622e4

krlvi (approved):
thanks for clarifying, indeed i misunderstood something here

Issue #5681, opened 2024-11-26
Gitbutler no longer asks for passphrase


vreima:
Hello,

Gitbutler seems to no longer ask for a passphrase, just fails with an error. This has been an issue for 2 or 3 weeks now.

![image](https://github.com/user-attachments/assets/e72f90b1-baf0-4ba1-80aa-4a72e87b7fc7)

Byron:
Thanks a lot for posting!

For all I know, this subsystem wasn't changed (which of course, might be wrong), but if it's not, maybe this behaviour changes if GitButler is launched from a terminal which has already been authenticated. This would mean that in the terminal, like `cmd.exe`, you would run a `git push` or `git fetch`, so it pops up a password query once. When GItButler is launched from there using the full path, it *could* inherit the environment and maybe won't have to ask.

Of course, this is at most a workaround, but it would be useful to see if this works.

Something else to try would be to download [a release from a couple of weeks ago](https://github.com/gitbutlerapp/gitbutler/releases), but I just noticed that binaries aren't attached, unfortunately. But if GitButler could be downgraded, it would be possible to find the lower bound of the version that introduced the issue.

estib-vega:
Hello @vreima, thanks a lot for reporting this.
I believe a fix for this should be available in version 0.14.0.
If you get the change, please let me know if that's not the case.

vreima:
Hello @estib-vega, I just updated and I can confirm that 0.14.0 works and correctly asks for the passphrase. Thank you!

estib-vega:
@vreima Awesome!
Thanks for the confirmation and please let us know if anything else misbehaves :)

PR #5686, opened 2024-11-26
Provide an action when fetching before push

estib-vega:
When pushing a stack, we need to fetch before. 
In the case that the user has a passphrase set for their SSH key, this fetch action will require input.
For that to happen, we need to pass an action.

As seen in https://github.com/gitbutlerapp/gitbutler/issues/5681


krlvi (approved):
oh interesting. So it was *me* who caused a regression here 🙈

Byron:
Maybe related to …

* #5681 ?

estib-vega:
> Maybe related to …
> 
> 
> 
> * #5681 ?

@Byron hey, yes! I thought I had linked the issue but forgot

PR #5692, opened 2024-11-27
fix: disable analytics in dev

ndom91:
## ☕️ Reasoning

- Skip unnecessary usage of analytics services
- Home DNS blocks sentry/posthog so this fixes me having to manually filter all of the failed connection attempts out of the console 😅


estib-vega (approved):
Makes sense 👍

PR #5704, opened 2024-11-28
fix: remove last usage of `vbranches.commits`

ndom91:
## ☕️ Reasoning

- Remove last usage of `vbranches.commits`

## 🧢 Changes

- Migrated to `branch.series.(patches|upstreamPatches) 
- Regarding `localCommits`/`localAndRemoteCommits`, etc. contexts - `BranchLane` used to set `commitStore`/`localCommits`/`localAndRemoteCommits` contexts for the entire lane and all its children 
	- Now, however, that cna be different for each series. So each series gets it set in `CurrentSeries`
	- The other place where this was used was in the uncommited files `BranchFiles` list. This one needs a list of _all_ patches / upstreamPatches in the lane in order to build the "I'm locked to ..." tooltip on uncommited files potentially.
- Migrated a few files to runes syntax


apps/desktop/src/lib/vbranches/types.ts
```diff
@@ -476,6 +457,22 @@ export class PatchSeries {
 	get integrated() {
 		return this.patches.length > 0 && this.patches.length === this.integratedCommits.length;
 	}
+
+	get ancestorMostConflictedCommit(): DetailedCommit | undefined {
+		const commits = [...this.upstreamPatches, ...this.patches].sort((a, b) => {
+			if (a.createdAt > b.createdAt) {
```
estib-vega: I wonder how this works with commits that were created in the middle of the series.

I believe this is only used to check if this is the 'first' conflicted from the bottom, so in order to avoid having to sort them, I think we can assume that there are no `upstreamPatches` that are also conflicted.

Maybe we can update that horrible name I gave this getter and make it even worse: `ancestorMostConflictedLocalCommit`

estib-vega (approved):
Just that comment about the sorting of commits

apps/desktop/src/lib/vbranches/types.ts
ndom91: Haha okay yeah I cleaned this up a bit. In terms of naming, I thought about it for a minute, but seeing as we want to avoid "top"/"bottom" I coudln't think of anything better :joy:

estib-vega: Haha fair enough

PR #5705, opened 2024-11-28
Capture analytics on user facing errors

mtsgrd:
- needed for product analytics


estib-vega (approved):

mtsgrd:
We should probably move towards using the preferred naming convention from: https://posthog.com/docs/product-analytics/best-practices

PR #5395, opened 2024-11-01
feat: complete svelte-5 migration

ndom91:
## ☕️ Reasoning

- Run svelte auto migrate script (`pnpm dlx sv migrate svelte-5`)
- Bump svelte and related dependencies in `pnpm-workspace.yml`
- Bump pnpm corepack version in root `package.json`
- Manually cleaned up a bunch of other things

Migration Guide: https://svelte.dev/docs/svelte/v5-migration-guide

PR #5709, opened 2024-11-28
feat: finish svelte5 migration

ndom91:
## ☕️ Reasoning

- Migrate remaining components and files to svelte-5 runes
- Attempt # 2, replacing https://github.com/gitbutlerapp/gitbutler/pull/5395


ndom91:
Should be good to go. Will test a bit more before merging however.

apps/desktop/src/lib/components/AIPromptEdit/AIPromptEdit.svelte
```diff
@@ -6,30 +6,36 @@
...
+	let prompts = $state<Prompts>();
 
 	if (promptUse === 'commits') {
 		prompts = promptService.commitPrompts;
 	} else {
 		prompts = promptService.branchPrompts;
 	}
 
-	$: userPrompts = prompts.userPrompts;
+	let userPrompts = $derived(prompts.userPrompts);
```
mtsgrd: There's too many instances to comment on each one, but let's make derived variables and props `const` where appropriate?

apps/desktop/src/lib/components/ProjectSetup.svelte
```diff
@@ -12,15 +12,19 @@
...
 
-	export let remoteBranches: RemoteBranchInfo[];
+	interface Props {
+		remoteBranches: RemoteBranchInfo[];
+	}
+
+	let { remoteBranches }: Props = $props();
 
-	const project = getContext(Project);
+	const project = $state(getContext(Project));
```
mtsgrd: This $state looks incorrect, you should be able to just remove it?

apps/desktop/src/routes/+page.svelte
```diff
@@ -2,15 +2,16 @@
 	import { ProjectsService } from '$lib/backend/projects';
 	import FullviewLoading from '$lib/components/FullviewLoading.svelte';
 	import { getContext } from '@gitbutler/shared/context';
-	import { derived } from 'svelte/store';
+	import { run } from 'svelte/legacy';
+	import { derived as derivedStore } from 'svelte/store';
```
mtsgrd: This looks unintentional.

apps/desktop/src/routes/+page.svelte
```diff
@@ -39,13 +40,13 @@
...
 
-	$: {
+	run(() => {
 		if ($redirect.type === 'redirect') {
 			goto($redirect.subject);
 		} else if ($redirect.type === 'no-projects') {
 			goto('/onboarding');
 		}
-	}
+	});
```
mtsgrd: Should we be fixing these `run`s?

apps/desktop/src/lib/components/ProjectSetup.svelte
ndom91: Yeah lemme try that, 90% of this PR was done by the migration tool btw

apps/desktop/src/routes/+page.svelte
ndom91: No so the migration tool complained about using the store derived in the same file as runes `$derived`. Let me double check though

apps/desktop/src/routes/+page.svelte
ndom91: Yeah slowly but surely going through and removing all the svelte/legacy imports 👍

apps/desktop/src/routes/+page.svelte
mtsgrd: ~~That shouldn't be a problem!~~ Right, that makes sense. I was clearly wrong.

mtsgrd (approved):

apps/desktop/src/lib/components/ProjectSetup.svelte
ndom91: Removed :ok_hand:

krlvi:
shall we land this bad boy?

PR #5722, opened 2024-12-02
finalize gix-merge integration

Byron:
Replace the remaining uses of `git2::merge_trees` with the respective `gix` function.
This was made possible thanks to two additions:

* [index support](https://github.com/GitoxideLabs/gitoxide/pull/1661)
* [forced tree-conflict resolution](https://github.com/GitoxideLabs/gitoxide/pull/1705)

Follow-up to #5416.

**Merging/Cheery-picking/Rebasing is now 3.1x faster!** I'd also expect it to be more correct, this means it should be able to resolve more unambiguous conflicts.

### Tasks

* [x] preliminary `gix` update
* [x] make API improvements for nicer integration
* [x] no test-double-run
* [x] replace all remaining usage of `git2::merge_trees()`
    - [x] cherry-pick
    - [x] edit-mode
    - [x] repo/rebase.rs
    - [x] `get_commit_index`
    - [x] test-project/merge
* [x] once again evaluate the `@~10` variant and find out what's going on there. Manual comparisons should be easier then.
     - Turns out there was one now fixed shortcoming/bug and another bug that was there already, to be addressed separately.
* [x] Enter resolve conflict mode - fix issue with diff problem
* [x] figure out why one of the conflicts isn't actually a conflict when trying to resolve it
* [x] update to latest `gix` on `main`

### Notes for the reviewer

* Thanks to (even) more exhaustive testing, merges are now of better quality particularly when a directory gets replaced with a file, or vice-versa.
* ~~There is one known niche condition in conjunction with tree-to-file or file-to-tree replacements where detected renames can throw of the resolution, leading to worse outcomes than what Git would produce. It's the only known case of this happening and it's unclear how it can be fixed. While writing this, I had an idea on how to fix it maybe, let's hope this comment doesn't have to stay long.~~ - the issue was due to rename-detection in empty blobs, which now is deactivated by default if these renames happen to individual files.
* I noticed that the `stack` tests run serially, which was due to a shortcoming in how `gix-testtools` locks its script before execution. However, as these script invocations run in their own tempdir, nothing needs to be locked, so everything can run in parallel. When testing this, it showed that running the gitbutler-cli in debug mode in parallel with 16 cores isn't great, and the runtime increased to 25s compared to 22s while running it serially, while massively increasing the used CPU. Most of the time was kernel time due to filesystem access, so it seems the filesystem doesn't like what's happening.  This won't be an issue as `gix-testtools` v0.16 is current, while we use v0.15 here. I also didn't release the patch yet, so there can't be any surprises. At some point though, one would probably try to make these tests better in creating their initial state.

### Validation

* When testing the rebasing/cherry-picking, I used `jgit`, and set the base of the target-branch in `virtual-branches.toml` to `7b955048eb86e1c12114554beacb27b329252b15`, about 500 commits in the past. Then I applied the `6.10` branch and hit the update button. This causes a lot of conflicting commits, which were then output to with `gitbutler-cli branch status > git2|gix.rom` for comparison.
    - Both look plausible, and seemingly are the same. Change-ids and dates make comparisons harder as commit-ids change strongly. The listing of paths seems different at times, and the UI can show duplicates. 
    - I also don't understand how a single commit can become about 500 after applying, as in theory it should figure out that the tag is already merged mostly.
    - So it's somewhat inconclusive, but doesn't look wrong.
* Actually, when going **only 10 commits** into the past, the baseline version shows conflicts, even lists them in the UI for a fraction of a second, and then all but one commits disappear. This disappearance doesn't happen when using the `gix-merge` version.
     - Interestingly this only happened once, and both versions seem to be exactly the same, i.e. `git2` and `gix`.

### Questions

* How are `conflicted_files` used? It's a list of ancestor, ours and their files taken directly from the index, but I can't imagine what these duplicate sets of files are good for. In theory, these wouldn't have to be stored - instead one re-does the merge without merge-strategy and then builds the conflicting index for presentation to the user.

#### Cherry-Pick

The test setup is a `jgit` repository which rebases 500 commits by using a workspace update. The branch used is `stable-6.10` and the 'base' commit (in `virtual_branches.toml`) is `7b955048eb86e1c12114554beacb27b329252b15`.

##### Before

`integrate_upstream` takes 13.80s. It's notable that the cherry-pick merge runs twice.

Screenshot 2024-12-05 at 21 10 00

##### After

`integrate_upstream` is down to 4.43s, **a *3.1x* improvement**.

Screenshot 2024-12-05 at 21 10 49

And according to the numbers with an [updated diffing engine](https://github.com/pascalkuthe/imara-diff/pull/16) we land at 3.8s, a **3.6x improvement**.

Screenshot 2024-12-09 at 07 05 18


Caleb-T-Owens:
@Byron This is amazing! I'll take a through look through tomorrow.

krlvi:
This is looking amazing - very juicy performance speed up

crates/gitbutler-oxidize/src/ext.rs
```diff
@@ -0,0 +1,123 @@
...
+        &self,
+    ) -> Result<(gix::merge::tree::Options, TreatAsUnresolved)>;
+}
+
+impl GixRepositoryExt for gix::Repository {
+    fn for_tree_diffing(mut self) -> anyhow::Result<Self> {
+        let bytes = self.compute_object_cache_size_for_tree_diffs(&***self.index_or_empty()?);
+        self.object_cache_size_if_unset(bytes);
+        Ok(self)
+    }
```
Caleb-T-Owens, line 68: As an unrelated observation, I've seen that there are various "modes" that you can put a gix repository into. I can't help but feel like it could be a little error prone in the long run, given that the states are not represented in the types.

crates/gitbutler-oxidize/src/ext.rs
```diff
@@ -0,0 +1,123 @@
...
+            .context("failed to merge trees")?;
+        Ok(!merge_outcome.has_unresolved_conflicts(conflict_kind))
+    }
+
+    fn merge_options_force_ours(&self) -> Result<Options> {
+        Ok(self
+            .tree_merge_options()?
+            .with_tree_favor(Some(gix::merge::tree::TreeFavor::Ours))
+            .with_file_favor(Some(gix::merge::tree::FileFavor::Ours)))
+    }
```
Caleb-T-Owens, line 107: :heart:

crates/gitbutler-oxidize/src/lib.rs
```diff
@@ -50,3 +53,51 @@ pub fn gix_to_git2_signature(
...
+            },
+        id,
+        flags,
+        mode,
+        ..
+    } in index.entries()
+    {
+        if flags.contains(gix::index::entry::Flags::REMOVE) {
+            continue;
+        }
```
Caleb-T-Owens, line 85: Why do we want to skip removals?

crates/gitbutler-repo/src/rebase.rs
```diff
@@ -225,11 +227,84 @@ fn commit_conflicted_cherry_result<'repository>(
...
+    });
+    for (stage, entry) in conflicting_entries {
+        let storage = match stage {
+            Stage::Unconflicted => {
+                unreachable!("BUG: filtered above to not contain unconflicted entries")
+            }
+            Stage::Base => &mut ancestor_entries,
+            Stage::Ours => &mut our_entries,
+            Stage::Theirs => &mut their_entries,
+        };
```
Caleb-T-Owens: Why filter above, when you could just put a `continue;` on line 256?

crates/gitbutler-repo/src/rebase.rs
```diff
@@ -225,11 +227,84 @@ fn commit_conflicted_cherry_result<'repository>(
...
+        ancestor_entries,
+        our_entries,
+        their_entries,
+    };
+
+    // Since we typically auto-resolve with 'ours', it maybe that conflicting entries don't have an
+    // unconflicting counterpart anymore, so they are not applied (which is also what Git does).
+    // So to have something to show for - we *must* produce a conflict, extract paths manually.
+    // TODO(ST): instead of doing this, don't pre-record the paths. Instead redo the merge without
+    //           merge-strategy so that the index entries can be used instead.
```
Caleb-T-Owens, line 269: Long term, this would be a good thing to do.

We were previously eagerly loading the diffs of all commits and other expensive information, but this is no longer the case.

crates/gitbutler-repo/src/rebase.rs
```diff
@@ -284,32 +363,18 @@ pub fn gitbutler_merge_commits<'repository>(
 
         // in case someone checks this out with vanilla Git, we should warn why it looks like this
         let readme_content =
-        b"You have checked out a GitButler Conflicted commit. You probably didn't mean to do this.";
+            b"You have checked out a GitButler Conflicted commit. You probably didn't mean to do this.";
```
Caleb-T-Owens, line 359: There have been a few small formatting changes like this. How has clippy/rustfmt not complained about them?

crates/gitbutler-repo/src/rebase.rs
```diff
@@ -355,621 +420,23 @@ impl ConflictEntries {
...
-/// Within our rebasing and merging logic, "their" is the commit that is getting
-/// cherry picked, and "our" is the commit that it is getting cherry picked on
-/// to.
-///
-/// This means that if we experience a conflict, we drop the changes that are
-/// in the commit that is getting cherry picked in favor of what came before it
-fn resolve_index(
-    repository: &git2::Repository,
-    index: &mut git2::Index,
-) -> Result<ConflictEntries, anyhow::Error> {
```
Caleb-T-Owens, line 371: It does make me happy that we no longer take care of this concern

crates/gitbutler-stack/tests/mod.rs
```diff
@@ -1,5 +1,5 @@
-pub mod file_ownership;
-pub mod ownership;
```
Caleb-T-Owens, line 2: Is there a tool we can use to detect unused pubs within a workspace?

crates/gitbutler-testsupport/src/test_project.rs
```diff
@@ -214,64 +214,60 @@ impl TestProject {
...
+            Refname::Remote(remote) => format!("refs/heads/{}", remote.branch()).parse()?,
+            _ => "INVALID".parse()?, // todo
         };
         let branch = self
             .remote_repository
-            .maybe_find_branch_by_refname(&branch_name)
-            .unwrap();
-        let branch_commit = branch.as_ref().unwrap().get().peel_to_commit().unwrap();
+            .maybe_find_branch_by_refname(&branch_name)?
+            .expect("branch exists");
```
Caleb-T-Owens, line 226: Why do we want to expect this one?

Caleb-T-Owens: oh... this is a test helper... why do we want to return a result?

Caleb-T-Owens:
> How are conflicted_files used? It's a list of ancestor, ours and their files taken directly from the index, but I can't imagine what these duplicate sets of files are good for. In theory, these wouldn't have to be stored - instead one re-does the merge without merge-strategy and then builds the conflicting index for presentation to the user.

`conflicted_files` are used on the commit card and edit mode for listing what files are conflicted (because they likely will not be in the diff), and also for informing the user _how_ a particular file is conflicted. IE ours deleted and theirs modified...

There is no good reason why we couldn't do the merge on the fly, and display the information that way. Especially because the result can be easily cached in the frontend.

Caleb-T-Owens:
Other than a few small questions, this is incredibly accurate. Amazing stuff

crates/gitbutler-oxidize/src/ext.rs
Byron: I think I get the point and thought about this as well, especially in comparison to `git2::Repository` `gix::Repository` makes you think of things more, something that will obviously be forgotten.

This setting I'd call it configuration for which type-state systems would be too much, especially since I had to work to keep generics out of `gix::Repository`, something that wasn't always the case.

Something that I always thought about doing is to simply set a default object cache size, but I couldn't yet make me because it might be as surprising if the app keeps piling up memory.

So the error is on the side of loosing performance if the object cache is forgotten for some operations.

Something that I could imagine doing… very seriously so even, is that those operations that need it will automatically setup an object cache if it isn't set yet, and then remove/reset it afterwards. But… what if one wants to do a diff with minimal memory footprint and is OK for it to take longer?
`git2` shows that it's probably alright, it's just that I also don't want to take away all the viable choices.

crates/gitbutler-oxidize/src/ext.rs
Byron: Thanks, I do try to make APIs so that they can be stable - builders are a must for this for truly add-only types.

crates/gitbutler-oxidize/src/lib.rs
Byron: tl;dr: Usually one would want the final index without entries that shouldn't be there, but sometimes it's possible to skip over marked items instead, which saves power.

Because that's a cheap way to remove entries from a list that only works if the list is rewritten anyway.
The alternative is to prune them right away in memory, but that's costly and unnecessary as long as the `git2` conversion happens.

It's notable that I changed the API of the index creation to make this behaviour extremely obvious, as I fell for this multiple times already 😅.

crates/gitbutler-repo/src/rebase.rs
Byron: A great catch, I changed it and it's so much easier to read.

crates/gitbutler-repo/src/rebase.rs
Byron: Indeed, I also think that the new sidebar-branch implementation shows how this can work. And I am looking forward to one day see all this lazy evaluation realized.

crates/gitbutler-repo/src/rebase.rs
Byron: I know that sometimes when lines get too long, `cargo fmt` flakes out entirely and simply stops formatting that file entirely. It can be that at some point I reformatted with a longer max line length, which kind of 'jiggles' files back into place even when the `cargo fmt` run is redone with the normal file length.

Also, it's my best guess for what happened, my memory about this is vague at best.

crates/gitbutler-repo/src/rebase.rs
Byron: While doing `gix merge` index creation I realized that the index file as we know it is utterly underdefined for the complexities of merge conflicts - it simply can't represent them properly. In the past, when trivial merges without rename tracking were done, that wasn't the case, but now… it's a lossy science and there is definitely room to innovate here.

crates/gitbutler-stack/tests/mod.rs
Byron: Not that I know of. For integration tests these are never required to be `pub`, so it's an easy fix.

crates/gitbutler-testsupport/src/test_project.rs
Byron: Large amounts of unwraps make code much less readable. My take on it is to…

• …not use them at all in production code
* …not use them at all in utility code for testing
* …use them while implementing a test just to…
* …finally remove them from the test in favor of `?` once the crate stabilises

As a crate stabilizes tests are unlikely to fail outside of assertions, so more readable code is preferable. Should a test start failing by returning an error, replacing `?` with unwraps temporarily is easy enough as well.

Byron:
Thanks a lot for the review - thanks to it I was able to make a change which makes the code more readable.

Byron (commented):
Thanks a lot for the review - thanks to it I was able to make a change which makes the code more readable.

crates/gitbutler-repo/src/rebase.rs
Caleb-T-Owens: Hi, @Byron. GitHub is saying that this has changed since I last viewed it, but I'm not actually seeing the changes. Is GitHub being broken or have your amendments not been pushed? Sorry for the bother.

crates/gitbutler-oxidize/src/ext.rs
Caleb-T-Owens: Sounds like you've put a lot of thought into that, and it all seems very logical. Thanks :+1:

crates/gitbutler-oxidize/src/lib.rs
Caleb-T-Owens: I'm not super familiar with the underlying representations of indexes. (if you have some resources on them, I'd love to read it).

Are you essentially saying, that the lack of an entry in an index implies a deletion?

crates/gitbutler-testsupport/src/test_project.rs
Caleb-T-Owens: Interesting. My take was that, using the results can obscure the stacktrace, and when I'm writing a test, if something other than what I'm testing is failing, it would be nice if it was painfully clear where that testing utility was going wrong.

It is however, undeniably more concise.

crates/gitbutler-stack/tests/mod.rs
Caleb-T-Owens: Fair enough :+1:

crates/gitbutler-repo/src/rebase.rs
Caleb-T-Owens: That is really cool :D

crates/gitbutler-testsupport/src/test_project.rs
Byron: > My take was that, using the results can obscure the stacktrace, and when I'm writing a test, if something other than what I'm testing is failing, it would be nice if it was painfully clear where that testing utility was going wrong.

And I am thinking the same. That's why I start tests with `.unwrap()` but migrate them towards `?` once the crate stabilises. That way the code is always most suited for the development stage one is in.

Caleb-T-Owens: Sounds reasonable

crates/gitbutler-repo/src/rebase.rs
Byron: There is definitely room to innovate in the PR workflow as well :). And I have no clue how to most easily get to the updated version.

Here is the commit that made the change, there is only one: c64b84032ea963d188ba414a1fb40900ccbb9489

crates/gitbutler-oxidize/src/ext.rs
Byron: I wish all that thought would finally allow me to make a decision :D.
Maybe the 'auto-cache' mode is something that could be enabled by default, `git2` style, but can also be disabled as needed.
Hmmm :).

crates/gitbutler-oxidize/src/lib.rs
Byron: All entries start out in `stage-0` - unconflicted. Once a path is marked for conflict, the unconflicted version of the entry is removed in any of the ways mentioned above, and replaced with up to 3 stages, base, ours and theirs.

In this case, after applying the conflicting entries, it won't have deleted the unconflicting ones, but only marked them for deletion by using an in-memory flag. Then they can be dropped at a later stage.

Caleb-T-Owens:
All looks good. We're going to merge after we release next

Byron:
Great! Let me add #5753 to the merge-queue then.

krlvi (approved):
Release is out, so let's merge this one too so that we can test it in the nightly :)

PR #5724, opened 2024-12-02
CLI: Apply branch

estib-vega:
Add the ability to apply git branches.

Behind the scenes, we look for any existing stack that was created out of the given branch.
If no stack is found, create a stack out of the given branch and apply it to the workspace.

---
- <kbd>&nbsp;2&nbsp;</kbd> #5725 
- <kbd>&nbsp;1&nbsp;</kbd> #5724 👈

### DIFF
--- a/crates/gitbutler-cli/src/args.rs
+++ b/crates/gitbutler-cli/src/args.rs
@@ -83,7 +83,14 @@ pub mod vbranch {
         },
         /// Add a branch to the workspace.
         Apply {
-            /// The name of the virtual branch to apply.
+            /// Whether it's a branch that we're applying.
+            ///
+            /// If a stack create from the given brach is not found a new stack is created.
+            #[clap(short = 'b', long, default_value_t = false)]
+            branch: bool,
+            /// The name of the stack to apply.
+            ///
+            /// If the flag `--branch` is set, this is the name of the branch to apply a stack from.
             name: String,
         },
         /// Create a new commit to named virtual branch with all changes currently in the worktree or staging area assigned to it.
--- a/crates/gitbutler-cli/src/command/vbranch.rs
+++ b/crates/gitbutler-cli/src/command/vbranch.rs
@@ -3,6 +3,7 @@ use gitbutler_branch::{BranchCreateRequest, BranchIdentity, BranchUpdateRequest}
 use gitbutler_branch_actions::{get_branch_listing_details, list_branches, BranchManagerExt};
 use gitbutler_command_context::CommandContext;
 use gitbutler_project::Project;
+use gitbutler_reference::{LocalRefname, Refname};
 use gitbutler_stack::{Stack, VirtualBranchesHandle};
 
 use crate::command::debug_print;
@@ -49,14 +50,22 @@ pub fn status(project: Project) -> Result<()> {
 }
 
 pub fn unapply(project: Project, branch_name: String) -> Result<()> {
-    let stack = branch_by_name(&project, &branch_name)?;
+    let stack = stack_by_name(&project, &branch_name)?;
     debug_print(gitbutler_branch_actions::save_and_unapply_virutal_branch(
         &project, stack.id,
     )?)
 }
 
-pub fn apply(project: Project, branch_name: String) -> Result<()> {
-    let stack = branch_by_name(&project, &branch_name)?;
+pub fn apply(project: Project, branch_name: String, from_branch: bool) -> Result<()> {
+    if from_branch {
+        apply_from_branch(project, branch_name)
+    } else {
+        apply_by_name(project, branch_name)
+    }
+}
+
+fn apply_by_name(project: Project, branch_name: String) -> Result<()> {
+    let stack = stack_by_name(&project, &branch_name)?;
     let ctx = CommandContext::open(&project)?;
     let mut guard = project.exclusive_worktree_access();
     debug_print(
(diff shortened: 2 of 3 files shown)


mtsgrd:
How come you want to "create if missing" when applying a branch, instead of creating it explicitly?

estib-vega:
> How come you want to "create if missing" when applying a branch, instead of creating it explicitly?

@mtsgrd The thought process was to imitate how it works in the application. We *apply* a branch and behind the scenes a Stack is created and added to the workspace. The Stack will have the same changes as the branch we applied.

Explicitly creating a Stack with the same name as a local branch will result on a new and *empty* stack.

crates/gitbutler-cli/src/command/vbranch.rs
```diff
@@ -55,21 +56,40 @@ pub fn unapply(project: Project, branch_name: String) -> Result<()> {
     )?)
 }
 
-pub fn apply(project: Project, branch_name: String) -> Result<()> {
-    let stack = branch_by_name(&project, &branch_name)?;
+pub fn apply(project: Project, branch_name: String, create_if_missing: bool) -> Result<()> {
+    let mut stacks = find_all_branches_by_name(&project, &branch_name)?;
+
+    if stacks.len() > 1 {
+        bail!("Found more than one virtual branch named '{branch_name}'");
```
mtsgrd: This should perhaps be "found more than one branch stack".

crates/gitbutler-cli/src/command/vbranch.rs
```diff
@@ -165,15 +185,20 @@ pub fn commit(project: Project, branch_name: String, message: String) -> Result<
...
+    let mut found = find_all_branches_by_name(project, name)?;
     if found.is_empty() {
         bail!("No virtual branch named '{name}'");
     } else if found.len() > 1 {
         bail!("Found more than one virtual branch named '{name}'");
     }
     Ok(found.pop().expect("present"))
 }
+
+fn find_all_branches_by_name(project: &Project, name: &str) -> Result<Vec<Stack>> {
```
mtsgrd: "Find all stacks by name" instead of branches?

mtsgrd:
Right.. I have observed branches can be applied into stacks, but I'm still living in the old world where each applied branch ends up in its own lane!

crates/gitbutler-cli/src/args.rs
```diff
@@ -83,7 +83,14 @@ pub mod vbranch {
...
-            /// The name of the virtual branch to apply.
+            /// Whether it's a branch that we're applying.
+            ///
+            /// If a stack create from the given brach is not found a new stack is created.
+            #[clap(short = 'b', long, default_value_t = false)]
+            branch: bool,
+            /// The name of the stack to apply.
+            ///
+            /// If the flag `--branch` is set, this is the name of the branch to apply a stack from.
             name: String,
```
estib-vega, line 94: I updated it in the following way, but I need to gather some more opinions here:

**What**
Add a flag `-b` or `--branch` to the `apply` command, to apply a git branch as a stack.
If set, we'll look for a stack that was create from the *refname* of the specified branch.
If no stack like that exists, we create one.

**Justification**
This to me is expected behavior, at least porting it from how the application flow is.
I see a branch that I want to check out, I hit apply and a new stack containing that branch (series) appears in my workspace.
The user did not have to explicitly create a stack out of the vanilla git branch before applying it.

crates/gitbutler-cli/src/command/vbranch.rs
```diff
@@ -164,16 +194,40 @@ pub fn commit(project: Project, branch_name: String, message: String) -> Result<
...
-        bail!("No virtual branch named '{name}'");
+        bail!("No stack named '{name}'");
     } else if found.len() > 1 {
-        bail!("Found more than one virtual branch named '{name}'");
+        bail!("Found more than one stack named '{name}'");
     }
     Ok(found.pop().expect("present"))
 }
+
+pub fn stack_by_refname(project: &Project, refname: &Refname) -> Result<Option<Stack>> {
```
Byron: It seems this one could be private, it's not called in `main`, right?

crates/gitbutler-cli/src/args.rs
Byron: The CLI as a dev-tool serves the GB developers, so I think it can be whatever it needs to be to facilitate using it, particularly from tests.

crates/gitbutler-cli/src/args.rs
```diff
@@ -83,7 +83,14 @@ pub mod vbranch {
         },
         /// Add a branch to the workspace.
         Apply {
-            /// The name of the virtual branch to apply.
+            /// Whether it's a branch that we're applying.
+            ///
+            /// If a stack create from the given brach is not found a new stack is created.
+            #[clap(short = 'b', long, default_value_t = false)]
```
Byron, line 89: There should be no need to set the default value here - it's implied to be `false`.

crates/gitbutler-cli/src/command/vbranch.rs
```diff
@@ -164,16 +194,40 @@ pub fn commit(project: Project, branch_name: String, message: String) -> Result<
...
+    Ok(Some(found.pop().expect("present")))
+}
+
+fn find_all_stacks_by_name(project: &Project, name: &str) -> Result<Vec<Stack>> {
+    let found: Vec<_> = VirtualBranchesHandle::new(project.gb_dir())
+        .list_all_stacks()?
+        .into_iter()
+        .filter(|b| b.name == name)
+        .collect();
+    Ok(found)
```
Byron:
```suggestion
    VirtualBranchesHandle::new(project.gb_dir())
        .list_all_stacks()?
        .into_iter()
        .filter(|b| b.name == name)
        .collect()
```

Removing `found` makes the `collect()` type perfectly implicit, which overall should be less noisy.

crates/gitbutler-cli/src/command/vbranch.rs
```diff
@@ -164,16 +194,40 @@ pub fn commit(project: Project, branch_name: String, message: String) -> Result<
...
+}
+
+fn find_all_stacks_by_refname(project: &Project, refname: &Refname) -> Result<Vec<Stack>> {
+    let found: Vec<_> = VirtualBranchesHandle::new(project.gb_dir())
+        .list_all_stacks()?
+        .into_iter()
+        .filter(|b| b.source_refname.as_ref() == Some(refname))
+        .collect();
+    Ok(found)
```
Byron:
```suggestion
    VirtualBranchesHandle::new(project.gb_dir())
        .list_all_stacks()?
        .into_iter()
        .filter(|b| b.source_refname.as_ref() == Some(refname))
        .collect()
```

Same as above.

Byron (commented):
Thanks for contributing to the CLI, this makes me think it's getting used more, and thus is more useful.

crates/gitbutler-cli/src/command/vbranch.rs
estib-vega: Correct. Will make it private 👍

mtsgrd (approved):

PR #5725, opened 2024-12-02
fix: Don't list all parents when calculating dependencies

estib-vega:
The commit interdependecy calculation should not list all parents.
Otherwise, the commits being processed might contain unexpected content that could lead to errors being thrown.

About the tests:
The way to test that this works is to create a stack that contains a merge-commit from itself to itself.
This simulates what it would look like to have parts of the base branch *leak* into the stack by not integration the upstream base.

---
- <kbd>&nbsp;2&nbsp;</kbd> #5725 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #5724

### DIFF
--- a/crates/gitbutler-branch-actions/tests/dependencies.rs
+++ b/crates/gitbutler-branch-actions/tests/dependencies.rs
@@ -731,6 +731,40 @@ fn dependecies_ignore_merge_commits() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn dependencies_handle_complex_branch_checkout() -> Result<()> {
+    // This test ensures that checking out branches with *complex* histories
+    // does not cause the dependency calculation to fail.
+    //
+    // The *complexity* of the branch is that is contains a merge commit from itself to itself,
+    // mimicking checking-out a remote branch that had PRs merged to it.
+    let ctx = command_ctx("complex-branch-checkout")?;
+    let test_ctx = test_ctx(&ctx)?;
+    let default_target = test_ctx.virtual_branches.get_default_target()?;
+    let my_stack = &test_ctx.stack;
+
+    let dependencies = compute_workspace_dependencies(
+        &ctx,
+        &default_target.sha,
+        &HashMap::new(),
+        &test_ctx.all_stacks,
+    )?;
+
+    let commit_dependencies = dependencies.commit_dependencies.get(&my_stack.id).unwrap();
+    assert_commit_map_matches_by_message(
+        commit_dependencies,
+        HashMap::from([
+            ("update a again", vec!["update a"]),
+            ("update a", vec!["add a"]),
+            ("Merge branch 'delete-b' into my_stack", vec!["add b"]),
+        ]),
+        &ctx,
+        "Commit interdependencies correctly calculated. They should only pick up the merge commit when calculating dependencies",
+    );
+
+    Ok(())
+}
+
 // Test utility functions
 // ---
 
@@ -814,12 +848,12 @@ fn extract_commit_messages(
     for (oid_key, oid_values) in actual {
         let repo = ctx.repository();
         let key_commit = repo.find_commit(*oid_key).unwrap();
-        let key_message = key_commit.message().unwrap().to_string();
+        let key_message = key_commit.message().unwrap().trim().to_string();
         let actual_values: Vec<String> = oid_values
             .iter()
             .map(|oid| {
                 let value_commit = repo.find_commit(*oid).unwrap();
-                let value_commit_message = value_commit.message().unwrap();
+                let value_commit_message = value_commit.message().unwrap().trim();
                 value_commit_message.to_string()
             })
             .collect();
--- a/crates/gitbutler-branch-actions/tests/fixtures/dependencies.sh
+++ b/crates/gitbutler-branch-actions/tests/fixtures/dependencies.sh
@@ -2,6 +2,54 @@
 set -eu -o pipefail
 CLI=${1:?The first argument is the GitButler CLI}
 
+# Add change ID to a commit.
+#
+# Example usage:
+# set_change_id adbb3234 "change-id-1" "my_stack"
+set_change_id() {
+    local commit_hash=$1
+    local change_id=$2
+    local branch_name=$3
+
+    local change_id_key="gitbutler-change-id"
+    local gitbutler_header_version_key="gitbutler-headers-version"
+    local gitbutler_header_version_value="2"
+
+    if [ -z "$commit_hash" ] || [ -z "$change_id" ] || [ -z "$branch_name" ]; then
+        echo "Usage: set_change_id <commit_hash> <change_id> <branch_name>"
+        return 1
+    fi
+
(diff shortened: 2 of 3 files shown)


crates/gitbutler-branch-actions/tests/dependencies.rs
```diff
@@ -731,6 +731,35 @@ fn dependecies_ignore_merge_commits() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn dependecies_handle_complex_branch_checkout() -> Result<()> {
```
Byron:
```suggestion
fn dependencies_handle_complex_branch_checkout() -> Result<()> {
```

crates/gitbutler-branch-actions/tests/dependencies.rs
```diff
@@ -731,6 +731,35 @@ fn dependecies_ignore_merge_commits() -> Result<()> {
...
+    let commit_dependencies = dependencies.commit_dependencies.get(&my_stack.id).unwrap();
+    assert_commit_map_matches_by_message(
+        commit_dependencies,
+        HashMap::from([
+            ("update a again", vec!["update a"]),
+            ("update a", vec!["add a"]),
+            ("Merge branch 'delete-b' into my_stack", vec!["add b"]),
+        ]),
+        &ctx,
+        "commit_dependencies",
```
Byron: If this is an arbitrary message, I'd write a very detailed description on what this is actually testing. Probably it would be something like "we traverse only the first parent, and thus see only a small subset of the commits that are actually present to avoid…".
Maybe a lot, but helps to more quickly grasp what's going on without having to check with blame, which most of the time only tells us that some "Sebastian Thiel changed a lot of whitespace or moved stuff" :D.

Byron (commented):
I really like the read-only test!

###### Random thoughts

While looking at it I was thinking that most operations aren't actually read-only as they at least add objects to the ODB, which fortunately won't break any test.
And if they at some point start doing more, then having these tests with persistent read-only state at least gives a shimmer of a chance that we'd notice it as tests could start failing locally.

crates/gitbutler-branch-actions/tests/dependencies.rs
estib-vega: Nice catch :)!

crates/gitbutler-branch-actions/tests/dependencies.rs
estib-vega: Totally fair, you pointed that out last time and just slipped my mind. 
I'll add more a descriptive message there 👍

crates/gitbutler-branch-actions/tests/fixtures/dependencies.sh
```diff
@@ -343,4 +397,45 @@ update 8
...
+
+git clone remote complex-branch-checkout
+(cd complex-branch-checkout
+  git config user.name "Author"
+  git config user.email "author@example.com"
+
+  git switch -c my_stack
+  echo "this is a" > a
+  git add a && git commit -m "add a" --trailer ""
+  set_change_id "$(git rev-parse HEAD)" "change-id-1" "my_stack"
```
estib-vega, line 404: We need to set the change ID in the local branch commits in order to maintain the 'complex' structure.

If the commits don't have a change ID, they'll be rebased so that they all get one. See https://github.com/gitbutlerapp/gitbutler/pull/5636

This process will get rid of all but the first parent of every commit and skip empty commits, effectively loosing the `"delete b"` commit, which we want to keep

crates/gitbutler-branch-actions/tests/fixtures/dependencies.sh
```diff
@@ -2,6 +2,60 @@
 set -eu -o pipefail
 CLI=${1:?The first argument is the GitButler CLI}
 
+# Function to add a custom header to a commit
```
mtsgrd: Can I haz proper punctuation plz?

crates/gitbutler-branch-actions/tests/fixtures/dependencies.sh
```diff
@@ -2,6 +2,60 @@
...
+    local change_id_key="gitbutler-change-id"
+    local gitbutler_header_version_key="gitbutler-headers-version"
+    local gitbutler_header_version_value="2"
+
+    if [ -z "$commit_hash" ] || [ -z "$change_id" ] || [ -z "$branch_name" ]; then
+        echo "Usage: set_change_id <commit_hash> <change_id> <branch_name>"
+        return 1
+    fi
+
+    # Get tree, parent, author, and committer information
```
mtsgrd: I'm a bit allergic to documentation that says "the next line will do x", and when you read the next line "it does x". Perhaps some of these strings are superfluous?

crates/gitbutler-branch-actions/tests/dependencies.rs
```diff
@@ -731,6 +731,35 @@ fn dependecies_ignore_merge_commits() -> Result<()> {
...
+    let commit_dependencies = dependencies.commit_dependencies.get(&my_stack.id).unwrap();
+    assert_commit_map_matches_by_message(
+        commit_dependencies,
+        HashMap::from([
+            ("update a again", vec!["update a"]),
+            ("update a", vec!["add a"]),
+            ("Merge branch 'delete-b' into my_stack", vec!["add b"]),
+        ]),
+        &ctx,
+        "Commit interdependencies correctly calculated. They should only pick up the merge commit when calculating dependencies",
```
mtsgrd, line 762: Should this say "incorrectly"?

crates/gitbutler-branch-actions/tests/dependencies.rs
```diff
@@ -731,6 +731,35 @@ fn dependecies_ignore_merge_commits() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn dependencies_handle_complex_branch_checkout() -> Result<()> {
```
mtsgrd, line 735: Even though I am for sure lazy with writing test descriptions, could we get a description of the test here? I've read the pr description twice, and I'm still a bit unclear on what it tests.

crates/gitbutler-branch-actions/tests/dependencies.rs
estib-vega: At least the intention is that the message states the correct behavior, so that if it fails, it shows what was supposed to happen.

crates/gitbutler-branch-actions/tests/dependencies.rs
mtsgrd: Sorry, I see Byron already asked for this.

estib-vega: That's totally fair. I'll add a comment explaining what this is testing 👍

crates/gitbutler-branch-actions/tests/fixtures/dependencies.sh
estib-vega: I'll change that description 👍

mtsgrd (approved):

PR #5731, opened 2024-12-03
feat: update PR target automatically when parent PR is integrated

ndom91:
## ☕️ Reasoning

- When the parent branch is merged and the user hits "update", we can handle updating the target branch of the PR for the user.
- Previously, they had to have the GitHub repository setting "delete branch once merged" enabled. This change means the user will no longer have to have this  enabled.


apps/desktop/src/lib/branch/SeriesHeader.svelte
```diff
@@ -165,6 +166,28 @@
 		}
 	});
 
+	/**
+	 * If the repository does not have "delete after merged" enabled, we need to manually update the base
+	 * of the bottom most branch to point to the projects baseBranch of choice after its parent has been integrated,
+	 * and  it now is the bottom-most series.
+	 */
+	$effect(() => {
```
ndom91, line 177: Not super happy with putting this here, open to any other suggestions

krlvi:
Nice one!

> Previously, they had to have the GitHub repository setting "delete branch once merged" enabled. This change means the user will no longer have to have this enabled.

As long as merging didnt happen on github itself :p

apps/desktop/src/lib/branch/SeriesHeader.svelte
```diff
@@ -163,6 +164,28 @@
...
+		const targetBase = $baseBranch.branchName.replace(`${$baseBranch.remoteName}/`, '');
+		if (
+			listedPr?.targetBranch !== targetBase &&
+			isBottomSeries &&
```
mtsgrd, line 181: Let's name this  `isBottomBranch`, rename and `isTopSeries` to `isTopBranch`?

apps/desktop/src/lib/branch/SeriesHeader.svelte
mtsgrd: I'm always a bit worried about effects that cover many different reactive elements. And.. currently I'm wondering, what would happen if the update call was rejected with a 401? It feels like we could end up with lots of retries, and worst some loop.

Is there anything we could do to mitigate that?

mtsgrd (approved):

apps/desktop/src/lib/branch/SeriesHeader.svelte
ndom91: Yeah good call, I added a `catch` to the then chain and a boolean to stop looping

Issue #5751, opened 2024-12-05
SS API errors when storing credentials in WSL environment


fev4:
### Version

0.14.2

### Operating System

Linux

### Distribution Method

AppImage (Linux)

### Describe the issue

Git Butler is failing to access the secure storage in WSL2 Ubuntu, showing "SS error: result not returned from SS API" and "No matching entry found in secure storage" errors. This appears to be a problem with the secure storage system interaction in the WSL2 environment.

### How to reproduce

1. Install and run Git Butler in WSL2 Ubuntu
2. Try to access/set secure credentials (I tried setting Anthropic API Keys)
3. Observe secure storage (SS) API errors in the logs

### Expected behavior

Git Butler should be able to properly access and store credentials in the system's secure storage.

### Relevant log output

```shell
2024-12-05T22:33:46.052193Z  INFO secret_set_global:socket reader: /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/zbus-3.15.2/src/socket_reader.rs:33: close time.busy=188µs time.idle=7.22ms handle="aiAnthropicKey" secret="<redacted>"
2024-12-05T22:33:46.053432Z ERROR secret_set_global: crates/gitbutler-tauri/src/secret.rs:15: error=Error(Couldn't access platform secure storage: SS error: result not returned from SS API

Caused by:
    SS error: result not returned from SS API) handle="aiAnthropicKey" secret="<redacted>"
2024-12-05T22:33:46.053456Z  INFO secret_set_global:socket reader: /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/zbus-3.15.2/src/socket_reader.rs:33: close time.busy=310µs time.idle=15.6ms handle="aiAnthropicKey" secret="<redacted>"
2024-12-05T22:33:46.053567Z  INFO secret_set_global: crates/gitbutler-tauri/src/secret.rs:15: close time.busy=16.7ms time.idle=95.1µs handle="aiAnthropicKey" secret="<redacted>"
[2024-12-05][22:33:46][ct@tauri://localhost/_app/immutable/entry/app.CnMUgsan.js:2:1438][ERROR] 08076a76ae364f37ac41f57bab4ac4e2: SS error: result not returned from SS API

s@tauri://localhost/_app/immutable/chunks/ipc.C01P9Nyx.js:1:898
fromError@tauri://localhost/_app/immutable/chunks/ipc.C01P9Nyx.js:1:1098
[...]

PR #5753, opened 2024-12-06
Update `keyring` to the latest version (#5751)

Byron:
Hopefully this helps with improving WSL2 support, see #5751. It's a quick shot though, I didn't check their changelog.
Also, I couldn't find any bugs reported that would mention WSL.

## Tasks

* [x] upgrade to latest API of `keyring` (this affects the git-credentials implementation only)
    - [x] Assure `git credentials` store still works
* [x] assure the 'normal' credentials store still works (needed cargo features now, it does nothing by default).
* [x] ~~fix issue with `git credentials` store credentials being URL encoded when queried (`gitoxide` issue, which takes `url::password()` directly even though it's percent-encoded)~~
     - Will be fixed with #5722

## Notes for the Reviewer

* I only tested this on MacOS this time, and would assume it still works on all other platforms as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -214,7 +214,7 @@ dependencies = [
  "wayland-backend",
  "wayland-client",
  "wayland-protocols",
- "zbus 4.4.0",
+ "zbus",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


Caleb-T-Owens (approved):
LGTM, thanks

Byron:
It should be safe to merge, but depending on the plans for the next release, this could wait just like #5722.
Thus, I let @krlvi merge both when good and ready.

krlvi (approved):

Issue #5735, opened 2024-12-04
pre-commit hook doesn't work with gitbutler


Cellou404:
### Version

0.1

### Operating System

Linux

### Distribution Method

AppImage (Linux)

### Describe the issue

Hi, 
It's been a while that I'm using **gitbutler** for my own projects. But now, I'm working for a company and there are a bit of configuration in all our starter template. In this case, I will name **pre-commit**.
When I make a commit for changes, `pre-commit` should check my code validating the commit. But when I use **gitbutler**, no checks are made and all commit pass.
I think it would be a great thing if **gitbutler** can consider developers `pre-commit` settings.

### Expected behavior

**Gitbutler** should consider developers `pre-commit` settings so their code are checked and validated before accepting the commit.

Byron:
Thanks for reporting!

I also think it should just respect hooks by default, but until this is the default, for now it needs to be activated on a per-project basis.

<img width="943" alt="Screenshot 2024-12-04 at 13 46 16" src="https://github.com/user-attachments/assets/007e69be-a361-4f0e-af66-706cfbe1c24c">

Does that work for you? It's known that some commit-hooks don't work as they should with GitButler as it forgoes the staging area, going straight to creating a new tree for the commit, which may prevent the hook to see what changed.

Ebsan:
I just ran into this problem. I installed pre-commit and have a simple hook setup to look for debugging print statements. I set "Run commit hooks" to true but it still doesn't seem to run them. Doing a `git commit` from the cli works as expected (commit is rejected because of pre-commit hook.). Are there more settings we have to enable to get it to work?

I don't know if this matters, but we're using the pre-commit project at https://pre-commit.com/.

Byron:
Enabling to run hooks per project should be all it takes.

https://github.com/gitbutlerapp/gitbutler/blob/78fb6190b7dbb8e904bef97b9aa59113a6cfe278/crates/gitbutler-branch-actions/src/virtual.rs#L721-L744

However, can it be that `pre-commit` hooks need the staging area to work? That's a typical caveat with the way GitButler works, as it bypasses the index entirely at the moment.

Cellou404:
Hi @Byron 

Activating `Run commit hooks` is now blocking the commits but not with the appropriate error messages.

**GitButler**

![Capture d’écran du 2024-12-05 10-16-15](https://github.com/user-attachments/assets/c750ccaf-84bc-45c0-8c42-5f6cf7c90737)

[...]

PR #5755, opened 2024-12-06
provide stderr in failing pre-commit hooks

Byron:
That way it should be less likely that valuable error information is swallowed.

Related to #5735.

### Tasks

* [x] use `stderr` is `stdout` is empty
* [x] validate locally

It's notable that newlines seem to be removed when showing the error, which is why I refrained from combinding stdout + stderr, in favor of using stderr if stdout is empty.

This should help understanding why precommit hooks are failing, even though the failure rate will still be too large as the index isn't used to create commits, something many
hooks rely on to work.


mtsgrd:
LGTM!

mtsgrd (approved):

PR #5769, opened 2024-12-07
fix: assure the head of edit mode is never a conflicted tree.

Byron:
Previously, the parent of a commit was unconditionally chosen if the current commit wasn't conflicted. However, the parent can still be conflicted which means it would need a 'fake' commit as well.

https://github.com/user-attachments/assets/4cf73ba0-43c4-4b9c-870c-86684c4f1ed9

### Tasks

* [ ] implement fix

### DIFF
--- a/crates/gitbutler-edit-mode/src/lib.rs
+++ b/crates/gitbutler-edit-mode/src/lib.rs
@@ -68,34 +68,41 @@ fn get_commit_index(repository: &git2::Repository, commit: &git2::Commit) -> Res
     }
 }
 
-fn checkout_edit_branch(ctx: &CommandContext, commit: &git2::Commit) -> Result<()> {
+fn checkout_edit_branch(ctx: &CommandContext, commit: git2::Commit) -> Result<()> {
     let repository = ctx.repo();
 
     let author_signature = signature(SignaturePurpose::Author)?;
     let committer_signature = signature(SignaturePurpose::Committer)?;
+    let maybe_conflicted_parent_commit = if commit.is_conflicted() {
+        Err(commit.clone())
+    } else if commit.parent(0)?.is_conflicted() {
+        Err(commit.parent(0)?)
+    } else {
+        Ok(commit.parent(0)?)
+    };
 
     // Checkout commits's parent
-    let commit_parent = if commit.is_conflicted() {
-        let base_tree = repository.find_real_tree(commit, ConflictedTreeKey::Ours)?;
-
-        let base = repository.commit(
-            None,
-            &author_signature,
-            &committer_signature,
-            "Conflict base",
-            &base_tree,
-            &[],
-        )?;
-        repository.find_commit(base)?
-    } else {
-        commit.parent(0)?
+    let commit_parent = match maybe_conflicted_parent_commit {
+        Err(conflicted) => {
+            let base_tree = repository.find_real_tree(&conflicted, ConflictedTreeKey::Ours)?;
+            let base = repository.commit(
+                None,
+                &author_signature,
+                &committer_signature,
+                "Conflict base",
+                &base_tree,
+                &[],
+            )?;
+            repository.find_commit(base)?
+        }
+        Ok(unconflicted) => unconflicted,
     };
     repository.reference(EDIT_BRANCH_REF, commit_parent.id(), true, "")?;
     repository.set_head(EDIT_BRANCH_REF)?;
     repository.checkout_head(Some(CheckoutBuilder::new().force().remove_untracked(true)))?;
 
     // Checkout the commit as unstaged changes
-    let mut index = get_commit_index(repository, commit)?;
+    let mut index = get_commit_index(repository, &commit)?;
 
     repository.checkout_index(
         Some(&mut index),
@@ -134,7 +141,7 @@ fn find_virtual_branch_by_reference(
 
 pub(crate) fn enter_edit_mode(
     ctx: &CommandContext,
-    commit: &git2::Commit,
+    commit: git2::Commit,
     branch: &git2::Reference,
     _perm: &mut WorktreeWritePermission,
 ) -> Result<EditModeMetadata> {
--- a/crates/gitbutler-edit-mode/src/lib.rs
+++ b/crates/gitbutler-edit-mode/src/lib.rs
@@ -68,35 +68,60 @@ fn get_commit_index(repository: &git2::Repository, commit: &git2::Commit) -> Res
     }
 }
 
-fn checkout_edit_branch(ctx: &CommandContext, commit: git2::Commit) -> Result<()> {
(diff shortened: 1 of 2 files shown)


crates/gitbutler-edit-mode/src/lib.rs
```diff
@@ -68,34 +68,41 @@ fn get_commit_index(repository: &git2::Repository, commit: &git2::Commit) -> Res
...
 
     let author_signature = signature(SignaturePurpose::Author)?;
     let committer_signature = signature(SignaturePurpose::Committer)?;
+    let maybe_conflicted_parent_commit = if commit.is_conflicted() {
+        Err(commit.clone())
+    } else if commit.parent(0)?.is_conflicted() {
+        Err(commit.parent(0)?)
+    } else {
+        Ok(commit.parent(0)?)
+    };
```
Caleb-T-Owens: I'm interested about this use of Result. Is this a common rust idiom? Given that rust doesn't have a built in `Either` type, it isn't entirely unreasonable, but was a little bit of an eyebrow raiser to me.

Byron: That's a very valid question. Indeed, if such use is limited, I tend to use it like `Either` with the caveat that I'd try to assign desirable cases to `Ok` and less desirable ones to `Err`, like is done here where conflicts aren't great :D.

Alternatively this could be done with an extra function if that's preferable, as I really can't tell if that's super common or not, so probably it's not common.

Caleb-T-Owens: Hmm...

In the case of `!commit.parent(0)?.is_conflicted()` rather than taking the `Ours` side, shouldn't we instead be using the `AutoResolution`, as that is what `commit` is diffed against in the client?

Byron: For now the original logic was kept, so "if it's conflicted, use `our` side". Since my intuition on what's right here isn't great, let met me try to reason about it by sharing what I know:

##### General

* when checking out some state, the index should match the tree at `HEAD`.
* the `.git/index` would be different only if the are conflicts, so without conflicts, `HEAD` points to `commit`, and `.git/index` is the state of `commit.tree()`.

##### `commit` doesn't conflict

Since `commit` is supposed to be checked out, it seems it (and not its parent) should be set to `HEAD` as `get_commit_index` will then also only get the index of `commit.tree_id()`.

##### `commit` does conflict

`get_commit_index()` will produce an `index` from the merge result that doesn't have any favor, i.e. is what Git would produce with `git merge`. The merged tree should be the one that `HEAD` points to, it's the merged state possibly with conflict markers, which is what `git merge` would do.
Please note that #5722 would make it easy to implement `get_commit_index()` in such a way that it also makes the tree available from which the conflicted index was created. 


##### Conclusion 

If that reasoning is correct, the logic here would change quite a bit, and it *seems* to make sense when `git merge` is the reference implementation.

If that also makes sense to you, I'd suggest to merge this PR as is as it fixes the bug without making the overall logic worse, and it can be followed up by the necessary changes once #5722 has landed.

What do you think?

Caleb-T-Owens: Hi! Just wanted to acknowledge this message. I'm going to respond in full tomorrow, sorry for the wait.

Caleb-T-Owens: Given I have three commits `Base <- A <- B <- C`

If the A changes such that B does not cleanly apply
Then I get `Base <- A' <- B' (Conflicted) <- C`

The trees of B look like:
- base `.conflict-base-0`: `At`
- ours `.conflict-side-0`: `A't`
- theirs `.conflict-side-1: `Bt`
- auto-resolution: The merge completed with a preference to `A't`.

The commit `C'` gets rebased on top of the auto-resolution of `B'`. In the client, we show the tree diffs against the auto-resolution as well.

---

In edit mode, we want to let the user edit the "diff" of that commit. We want the commit that they are editing to appear as uncommitted changes.

If we want to edit commit `C'`, that means we want the uncommitted changes to "be the diff" of `C't` over `B' auto-resolution`.

As such, we should make a commit of `B' auto-resolution` and hard check that out, and then soft checkout the `C't`

---

If I want to edit `B'`, we want to reconstruct the three way merge that produces `B'`, so the user can create a `B''` that cleanly applies on top of `A'`.

Because the base of the three way merge which creates the index has a base of `.conflict-base-0`, that is why we hard checkout the a commit that contains `.conflict-base-0`.

Byron: Thanks for the writeup, that's quite enlightening.

I think something very fundamental that I was missing is what `base`, `side-0` and `side-1` actually are in case of a conflicting commit.
It's also really interesting that when rebasing `B`, it's actually favouring `A` and not `B`, another case of where my intuition was horribly wrong.

I find interesting that this is the merge that happens when `A` is modified, and I have a feeling that this isn't what's happening when doing an actual cherry-pick as there is no modified commit in the first place. But… I have a feeling that this once again isn't quite right and I'd do well to study how cherry-picking actually works to understand the logic behind all this.

Assuming that the problem this PR tries to solve is well understood, do you want o push into this PR to make it proper or supersede this PR with a better solution?

Caleb-T-Owens: @Byron I'll push onto this PR

crates/gitbutler-edit-mode/src/lib.rs
```diff
@@ -68,34 +68,41 @@ fn get_commit_index(repository: &git2::Repository, commit: &git2::Commit) -> Res
...
 
     // Checkout commits's parent
-    let commit_parent = if commit.is_conflicted() {
-        let base_tree = repository.find_real_tree(commit, ConflictedTreeKey::Ours)?;
```
Caleb-T-Owens, line 79: You know... this was a bug

Caleb-T-Owens:
@Byron I've updated the logic, would you mind taking a read and seeing if it makes sense to you.

Byron (commented):
Thanks for taking a look - having a function for that is much better!

Regarding the logic, I *think* it's now correct as…

* …we set the `HEAD` to the commit that *was* our previous state, at least hypothetically.
* …we set the index *and* worktree to the index that is comprised of the merge result + the conflicting entries.

Admittedly, I think I'd have to check the source of `git merge` to be really sure, because as a matter of fact I don't really know what Git does there. It seems to just make sense to keep `HEAD` pointing to the state that initiated the merge, the previous state if you will. But… still not sure 🤔.

Funnily enough one would think I'd know having implemented `gix merge commit`, but as a plumbing-level command it does nothing with the `HEAD`, the worktree or the index :D.

Caleb-T-Owens (approved):

PR #5772, opened 2024-12-07
Fix bug preventing change in in persisted store

mtsgrd:
- affects e.g. telemetry page toggles as reported on discord

### DIFF
--- a/apps/desktop/src/lib/config/appSettings.ts
+++ b/apps/desktop/src/lib/config/appSettings.ts
@@ -8,39 +8,46 @@
 import { Store } from '@tauri-apps/plugin-store';
 import { writable, type Writable } from 'svelte/store';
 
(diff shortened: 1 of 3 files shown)


apps/desktop/src/lib/config/appSettings.ts
```diff
@@ -10,40 +10,47 @@ import { writable, type Writable } from 'svelte/store';
 
 export async function loadAppSettings() {
 	const diskStore = await Store.load('settings.json', { autoSave: true });
+	console.log(diskStore);
```
krlvi: probably drop

mtsgrd: dropped, and tests added

krlvi (approved):

PR #5775, opened 2024-12-07
Force default target directory

c-git:
## ☕️ Reasoning

The TypeScript code already assumes that the default target directory is being used and if this is not the case the build fails

## 🧢 Changes

Force the target directory to be the default value and add a note to the instructions.

<!--


Byron (approved):
Thanks a lot for your help!

I see how this can be useful if the `target-dir` setting was overridden on a higher level, which then would have scripts fail.
By re-enforcing it here, the build will also succeed in this case.

I leave the merge to @krlvi for good measure.

PR #5781, opened 2024-12-09
Disallow edit exit doublepress

### LINKED ISSUE #5770: async buttons don't prevent multi-action (e.g. edit mode cancel)
### Version

0.14.2

### Operating System

macOS

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

When in edit mode, it's possible to click the cancel button multiple times, triggering the same action twice, with the second one showing an error.

https://github.com/user-attachments/assets/567345dd-9b97-4040-96e5-e30bffb52486

I think this is a general problem so I phrased it as such, while hopingthat maybe this can be solved in general by having a reusable async-button that is properly wired with the action is triggered while disallowing multi-triggering while the action is in progress.

### How to reproduce

* Click the `Resolve Conflicts` button
* Click the `Cancel` button in the edit modal in short succession

### Expected behavior

Buttons that trigger actions with the backend should in general be 'synced' to the action they trigger.

### Relevant log output

_No response_

Caleb-T-Owens:
* Closes https://github.com/gitbutlerapp/gitbutler/issues/5770

---
- <kbd>&nbsp;2&nbsp;</kbd> #5782 
- <kbd>&nbsp;1&nbsp;</kbd> #5781 👈

### DIFF
--- a/apps/desktop/src/lib/components/EditMode.svelte
+++ b/apps/desktop/src/lib/components/EditMode.svelte
@@ -187,6 +187,7 @@
 		modeServiceAborting = 'loading';
 
 		await modeService.abortEditAndReturnToWorkspace();
+		await modeService.awaitNotEditing();
 
 		modeServiceAborting = 'completed';
 	}
@@ -195,6 +196,7 @@
 		modeServiceSaving = 'loading';
 
 		await modeService.saveEditAndReturnToWorkspace();
+		await modeService.awaitNotEditing();
 
 		modeServiceSaving = 'completed';
 	}
@@ -219,6 +221,8 @@
 	}
 
 	let isCommitListScrolled = $state(false);
+
+	const loading = $derived(modeServiceSaving === 'loading' || modeServiceAborting === 'loading');
 </script>
 
 <div class="editmode__container">
@@ -302,9 +306,7 @@
 	</p>
 
 	<div class="editmode__actions">
-		<Button style="ghost" outline onclick={abort} disabled={modeServiceAborting === 'loading'}>
-			Cancel
-		</Button>
+		<Button style="ghost" outline onclick={abort} disabled={loading} {loading}>Cancel</Button>
 		{#if conflictedFiles.length > 0}
 			<Button
 				style="neutral"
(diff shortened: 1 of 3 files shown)


apps/desktop/src/lib/components/EditMode.svelte
```diff
@@ -187,6 +187,7 @@
 		modeServiceAborting = 'loading';
 
 		await modeService.abortEditAndReturnToWorkspace();
+		await modeService.awaitNotEditing();
```
Caleb-T-Owens, line 191: The main fix is simply waiting until the mode has switched before clearing the loading state.

estib-vega (approved):

apps/desktop/src/routes/[projectId]/+layout.ts
```diff
@@ -65,7 +65,12 @@ export const load: LayoutLoad = async ({ params, parent }) => {
 	const branchListingService = new BranchListingService(projectId);
 	const gitBranchService = new GitBranchService(projectId);
 
-	const vbranchService = new VirtualBranchService(projectId, projectMetrics, branchListingService);
+	const vbranchService = new VirtualBranchService(
+		projectId,
+		projectMetrics,
+		branchListingService,
+		modeService
+	);
```
Comment (estib-vega) [inline apps/desktop/src/routes/[projectId]/+layout.ts]: I think this hunk belongs in the next PR

Comment (Caleb-T-Owens) [reply apps/desktop/src/routes/[projectId]/+layout.ts]: Whoops, yes it does

apps/desktop/src/lib/components/EditMode.svelte
```diff
@@ -195,6 +196,7 @@
 		modeServiceSaving = 'loading';
 
 		await modeService.saveEditAndReturnToWorkspace();
+		await modeService.awaitNotEditing();
```
Byron, line 199: I find these method calls are quite treacherous as they as I have a feeling they can throw an exception. Then the button could be stuck in loading state.

If one of these fails the (backend) method itself is likely to leave the working tree/repo in an invalid state, so improving the frontend is certainly only a part of what it takes to be truly 'atomic' or deterministic.

Caleb-T-Owens: `saveEditAndReturnToWorkspace` could throw. `awaitNotEditing` is just creating a promise that resolves whenever the backend emits the fact that we're no longer in edit.

I'll add a little `try ... finally` that takes you out of the loading state if `xxxEditAndReturnToWorkspace` fails.

Issue #5763, opened 2024-12-06
GitButler will not close (CMD+Q) when error pip-ups are present


krstp:
### Version

0.14.2

### Operating System

macOS

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

Unable to quit app when error pop-ups are floating in the bottom-right corner. Had to dismiss all, only then I was able to close the app.

### How to reproduce

Cause an error message. Try to CMD+Q app.

### Expected behavior

App should close gracefully terminating any active processes or possibly ask to verify the quit process if there are other active processes.

Byron:
Thanks a lot for reporting!

Could you provide more information, as maybe the type of error matters as well.

I have recorded a video where `Cmd + Q` works fine even with an error popup visible.
https://github.com/user-attachments/assets/0519470d-f8cf-4993-acf8-77b673f41870

Thanks for your help!

krstp:
@Byron Literally, try drag'n'drop a hunk from one commit to the other, similar to here (https://github.com/gitbutlerapp/gitbutler/issues/5761 ; about to upload a video).

After error shows up, CMD+Q will not activate. No point of making video as the UI will not react to CMD+Q.

I hope this helps.

![Screenshot 2024-12-07 at 11 31 41 PM](https://github.com/user-attachments/assets/d50e5978-d53f-42cc-9985-5d418164b397)

krstp:
BTW: What a bizarre thing, now it works :| Second ago I did try it and it was resulting with exact error. I am baffled.

Byron:
Turns out that *sometimes* quitting works, but sometimes it doesn't - I even have it on video but chose to only upload the part where the popup blocks quitting.

https://github.com/user-attachments/assets/3b78750c-5f04-432b-9e45-f5dfbe41bf5d

Not shown in the video is how closing the popup makes quitting possible again.

CC @estib-vega - maybe it's something about hotkey capturing in error popups?

estib-vega:
@krstp Thanks for reporting and @Byron thanks for tagging me!

We've observed this behavior in the wild, but every time we start chasing it, we can't seem to reproduce.
I'm glad you two seem to have found a reproducible condition.
I'll take a look!

krlvi:
Hey @krstp, this was fixed in https://github.com/gitbutlerapp/gitbutler/pull/5785 and shipped in `0.14.4`: https://github.com/gitbutlerapp/gitbutler/releases/tag/release/0.14.4

[...]

PR #5785, opened 2024-12-09
fix: Don't stand in the way of closing the app

estib-vega:
The BranchFileList component would mistakenly intercept the application closing key binding (cmd + q in macOS) when focusing on it.

Don't.

### Issues
As seen in https://github.com/gitbutlerapp/gitbutler/issues/5763


ndom91 (commented):
How does preventDefault on these other keys fix this? :thinking:

estib-vega:
> How does preventDefault on these other keys fix this? :thinking: 

It's moving the preventDefault into the function, so that it's only prevented on those keys, as opposed to all keys at the parent caller

mtsgrd (approved):

PR #5791, opened 2024-12-09
fix: swallow update on base branch failed for closed pr error

ndom91:
## ☕️ Reasoning

- Purposefully swallow "cannot change the base branch of a closed pull request" errors when attmepting to update base.
- Shouldn't happen anyway since we're checking `$pr.state === 'open'` before attempting this, but looks like it can happen anyway.

### DIFF
--- a/apps/desktop/src/lib/branch/SeriesHeader.svelte
+++ b/apps/desktop/src/lib/branch/SeriesHeader.svelte
@@ -192,6 +192,9 @@
 					await updateStatusAndChecks();
 				})
 				.catch((err) => {
+					if (err.message.includes('Cannot change the base branch of a closed pull request'))
+						return;
+
 					showError('Failed to update PR target base', err.message ? err.message : err);
 					targetBaseError = err;
 				});


apps/desktop/src/lib/branch/SeriesHeader.svelte
```diff
@@ -192,6 +192,9 @@
 					await updateStatusAndChecks();
 				})
 				.catch((err) => {
+					if (err.message.includes('Cannot change the base branch of a closed pull request'))
+						return;
+
```
mtsgrd, line 197: Could we also swallow that error when base branch is already set to the thing it's being updated to?

ndom91: Actually I'm thinking both this the swallowing in this PR and the swallowing of that error you're pointing out should probably not be required any more once we merge this new ["update pr base branch at the end of 'merge' event handler instead of in $effect"](https://github.com/gitbutlerapp/gitbutler/pull/5797) PR.

PR #5792, opened 2024-12-09
update-storybook-component

PavelLaptev:
Migrate from storybook stories with two files to the single file story approach.


ndom91 (approved):
So in general this LGTM, however now that these are all using the SFC syntax, we unfortunately don't get the default tests anymore. When running `vite test` in `packages/ui` it would previously include all the stories and at least make sure the components rendered. Appanrelyt the SFC syntax + `vite-experimental-test-addon` aren't compatible yet. It's being tracked here: https://github.com/storybookjs/addon-svelte-csf/issues/213

But anyway, it should be coming soon and I dont think its that big a deal if they're off again for a short while. :shipit:

PavelLaptev:
> But anyway, it should be coming soon and I dont think its that big a deal if they're off again for a short while. :shipit:

Yes, we can probably handle this in the next iteration, since I also want to revisit some prop exports.

PR #5794, opened 2024-12-10
Branch listings: Improve search

estib-vega:
- Ignore the position of the fuzzy-matched items when searching
- Search is case insensitive
- Sort by date, and then by score

### DIFF
--- a/apps/desktop/src/lib/branches/branchListing.ts
+++ b/apps/desktop/src/lib/branches/branchListing.ts
@@ -228,35 +228,40 @@ export class CombinedBranchListingService {
 	}
 
 	search(searchTerm: Readable<string | undefined>) {
-		return derived(
-			[searchTerm, this.combinedSidebarEntries],
-			([searchTerm, combinedSidebarEntries]) => {
-				if (!searchTerm) return [];
-
-				const fuse = new Fuse(combinedSidebarEntries, {
-					keys: [
-						// Subject is branch listing
-						'subject.name',
-						'subject.lastCommiter.email',
-						'subject.lastCommiter.name',
-						// Subject is pull request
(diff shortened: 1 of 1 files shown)


apps/desktop/src/lib/branches/branchListing.ts
```diff
@@ -228,35 +228,42 @@ export class CombinedBranchListingService {
...
-					threshold: 0.3, // 0 is the strictest.
-					sortFn: (a, b) => {
-						// Sort results by when the item was last modified.
-						const dateA = (a.item.modifiedAt || a.item.updatedAt) as Date | undefined;
-						const dateB = (b.item.modifiedAt || b.item.updatedAt) as Date | undefined;
-						if (dateA && dateB) {
-							return dateA < dateB ? -1 : 1;
-						}
-						return 0;
+		const engine = derived([this.combinedSidebarEntries], ([combinedSidebarEntries]) => {
```
mtsgrd: You can skip the array when there's only one store being derived from.

apps/desktop/src/lib/branches/branchListing.ts
```diff
@@ -228,35 +228,42 @@ export class CombinedBranchListingService {
...
+						return dateA < dateB ? -1 : 1;
 					}
-				});
+					// If there are no dates or they're the same, sort by score
+					return a.score < b.score ? -1 : 1;
+				}
+			});
 
-				return fuse.search(searchTerm, { limit: 100 }).map((result) => result.item);
+			return fuse;
```
mtsgrd: Can you return this immediately instead of declaring variable -> return variable?

mtsgrd (approved):

PR #5796, opened 2024-12-10
fix: swallow signing error and dont show duplicate error msg

ndom91:
## ☕️ Reasoning

- We're already showing a special error msg for commit signing failures, no need to throw and re-show error higher up

### Before

image

### After

image


mtsgrd (approved):
Awesome.

PR #5806, opened 2024-12-11
PR Details modal: Improve the way content is updated

### LINKED ISSUE #5645: Caret moves to the end of PR description on first input
### Version

0.13.16

### Operating System

macOS

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

When opening a PR with a template from Gitbutler the caret jumps to the end of the template on first keydown

### How to reproduce

So when i place the caret at the top of the template to start writing I have to first click on any key, then move the caret back to the top, and then I'm able to start writing the PR description. In the video below I place the caret at the end of `## Description` and hit `Enter` (*any key causes the same behavior*), then the caret moves to the end of the template. 

Skjermopptak 2024-11-22 kl  08 55 09

### Expected behavior

When having placed the caret inside the PR description I expect it to be placed there when I start writing

### Relevant log output

_No response_

### LINKED ISSUE #5805: Extra automatic backspace when deleting characters in the PR description textbox
### Version

0.14.3

### Operating System

macOS

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

I think there's a small bug in the text field for writing the description of the PR in the "Create a pull request" modal where the preceding space before a word is also deleted after you finish deleting each character in the word separately.

### How to reproduce

- Type something in the description text box like: "This PR does"
- Hit backspace 4 times to delete just the word "does"
- Notice it actually deletes 5 characters and goes right up against the word "PR"

This means you have to hit the spacebar again after deleting, which isn't how any other textbox I've encountered works, and led to multiple words that were missing spaces between them

### Expected behavior

Should only delete the actual characters you clicked "backspace" to delete

### Relevant log output

_No response_

estib-vega:
Factor out the logic behind the PR title and body content update.

This fixes:
- The annoying caret-jumping-to-the-end issue
- The aggressive trimming issue

### Issues
- Fixes: https://github.com/gitbutlerapp/gitbutler/issues/5805
- Fixes: https://github.com/gitbutlerapp/gitbutler/issues/5645

### DIFF
--- a/apps/desktop/src/lib/pr/prContent.svelte.ts
+++ b/apps/desktop/src/lib/pr/prContent.svelte.ts
@@ -1,0 +1,106 @@
+import { persistedPRBody, persistedPRTitle } from '$lib/config/config';
+import type { DetailedCommit } from '$lib/vbranches/types';
+
+export interface PRContent {
+	title: string;
+	body: string;
+	templateBody: string | undefined;
+}
+
+export interface PRContentParams {
+	isDisplay: boolean;
+	projectId: string;
+	seriesName: string;
+	seriesDescription: string | undefined;
+	existingTitle: string | undefined;
+	existingBody: string | undefined;
+	commits: DetailedCommit[];
+}
+
+export default function getPRContent(params: PRContentParams): PRContent {
+	let templateBody = $state<string | undefined>(undefined);
+	const defaultTitle: string = $derived.by(() => {
+		if (params.isDisplay) return params.existingTitle ?? '';
+
+		// In case of a single commit, use the commit summary for the title
+		if (params.commits.length === 1) {
+			const commit = params.commits[0];
+			return commit?.descriptionTitle ?? '';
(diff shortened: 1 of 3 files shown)


apps/desktop/src/lib/pr/prContent.svelte.ts
```diff
@@ -0,0 +1,106 @@
...
+		}
+
+		return '';
+	});
+
+	const persistedBody = persistedPRBody(params.projectId, params.seriesName);
+	const persistedTitle = persistedPRTitle(params.projectId, params.seriesName);
+
+	let inputBody = $state<string | undefined>(undefined);
+	let inputTitle = $state<string | undefined>(undefined);
```
ndom91:
```suggestion
	let inputTitle = $state<string>();
```

This will result in the same thing :+1:

ndom91 (approved):
Small comment, but generally LGTM :partying_face:

PR #5807, opened 2024-12-11
PR Details modal: Inline the use template store

estib-vega:


apps/desktop/src/lib/config/config.ts
```diff
@@ -60,8 +60,4 @@ export function persistedPRTitle(projectId: string, seriesName: string): Persist
 	return persistWithExpiration('', 'seriesCurrentPRTitle_' + projectId + '_' + seriesName, 5);
 }
```
mtsgrd: Could we do the same with `persistedPRTitle` and `persistedPRBody`? We are needlessly packaging by layer here rather than feature.

estib-vega: Done 👍

mtsgrd (approved):

PR #5811, opened 2024-12-11
PR Details Modal: Svelte TS classes

estib-vega:
Use classes for the reactive logic for handling the PR title and body

### DIFF
--- a/apps/desktop/src/lib/pr/prContent.svelte.ts
+++ b/apps/desktop/src/lib/pr/prContent.svelte.ts
@@ -1,114 +1,66 @@
-import { persistWithExpiration, type Persisted } from '@gitbutler/shared/persisted';
 import type { DetailedCommit } from '$lib/vbranches/types';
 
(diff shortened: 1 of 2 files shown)


apps/desktop/src/lib/pr/prContent.svelte.ts
```diff
@@ -1,114 +1,70 @@
...
+		return this.title ?? '';
+	}
 
-	const defaultBody: string = $derived.by(() => {
-		if (params.isDisplay) return params.existingBody ?? '';
-		if (params.seriesDescription) return params.seriesDescription;
-		if (templateBody) return templateBody;
+	set(value: string) {
+		this.title = value;
+	}
```
Caleb-T-Owens: ought this be `set current(value: string) {` ?

estib-vega: We can do that, but I thought it would be nice to have a setter method that's explicit.

I have no strong feelings against it, though

ndom91 (approved):

PR #5815, opened 2024-12-11
Settings v2

krlvi:
The broader goal here is to create a good setup for application settings that is ergonomic for users, easy to work with and not error prone. In general, what VSCode does with settings is great and probably a good target - a user can use the UI or edit the JSON, the settings are portable.

#### Requirements:
- [x] Settings file is user-editable
- [x] If the settings file contains mistakes (eg. missing commas etc) it does not break the app
- [x] The settings file should not contain entries for all defaults values in existence
- [x] Allow nesting/grouping of related configs
- [x] The settings should be editable via the UI as well and have those changes reflected in the same file (similar to VSCode)
- [x] The app should watch/listen for updates in the settings file and upon change reflect that in it's behavior (e.g. need to update the UI)
- [x] Clear /discoverable defaults
- [x]  Within Rust, settings structs fields should not be Options if there are well defined defaults (development ergonomics) 
- [x] Should be hard to mess something up when adding new config structs or fields to existing structs and their respective update APIs
- [ ] If the settings json file contains comments, the app should not mess with those upon settings changes via the UI

#### Outstanding questions:
- Where should the settings file live? `~/.config/gitbutler/settings.json`, `~/.gitbutler.json`? Somewhere else? How about Windows?
- Maybe temporarily skip the requirement for preserving comments in the json to simplify implementing?

#### TODOs:
- [ ] More tests
- [x] APIs for UI
- [x] Filesystem watching for the settings file & appropriate actions
- [x] Settings change event propagation to frontend
- [ ] Preserve comments in the json

#### Out of scope here:
- Adopt in typescript
- Migrate existing settings to the new format
- Remove `LegacySettings` and the tauri store plugin


crates/gitbutler-settings/src/persistence.rs
```diff
@@ -0,0 +1,78 @@
...
+        // Load customizations
+        let customizations = serde_json_lenient::from_str(&std::fs::read_to_string(config_path)?)?;
+
+        // Merge the customizations into the settings
+        merge_non_null_json_value_into(customizations, &mut settings);
+        Ok(serde_json::from_value(settings)?)
+    }
+
+    /// Save the updated fields of the AppSettings in the custom configuration file.
+    pub fn save(&self, config_dir: impl Into<PathBuf> + Clone) -> Result<()> {
```
Byron: The `+ Clone` is very uncommon, and I think it will suffice to have `impl Into<PathBuf>`, convert right away and then reference the value where needed.

crates/gitbutler-settings/src/persistence.rs
```diff
@@ -0,0 +1,78 @@
...
+        // TODO: This will nuke any comments in the file
+        merge_non_null_json_value_into(diff, &mut customizations);
+        gitbutler_fs::write(config_path, customizations.to_string())?;
+        Ok(())
+    }
+}
+
+mod tests {
+    #[test]
+    fn ensure_default_settings_covers_all_fields() {
```
Byron, line 55: I love this test, and I am sure it will be very useful in future!

crates/gitbutler-settings/vendor/util/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+                target_object.insert(key.clone(), value);
+            }
+        }
+    } else if !source.is_null() {
+        *target = source
+    }
+}
+
+/// Get an embedded file as a string.
+pub fn asset_str<A: Embed>(path: &str) -> Cow<'static, str> {
```
Byron: To me the `rust-embed` crate seems overkill here as `static foo: &str = include_str!(<path>)` could easily be used.

The crate documentation says:

> Rust Custom Derive Macro which loads files into the rust binary at compile time during release and loads the file from the fs during dev.

If the `loads from fs during dev` aspect is important, which I have problems to see, then I think it would be helpful to document that workflow so devs are aware (I definitely didn't know that distinction to `include_str!()`).

crates/gitbutler-settings/Cargo.toml
```diff
@@ -6,3 +6,17 @@ authors = ["GitButler <gitbutler@gitbutler.com>"]
...
+serde_json_lenient = "0.2.3"
+gitbutler-settings-util.path = "vendor/util"
+gitbutler-fs.workspace = true
+
+[[test]]
+name = "settings"
+path = "tests/mod.rs"
+
+[dev-dependencies]
+gitbutler-settings.workspace = true
```
Byron: This shouldn't be required (but can be useful if different features should be set).

crates/gitbutler-settings/vendor/util/src/lib.rs
```diff
@@ -0,0 +1,32 @@
+use std::borrow::Cow;
+
+use rust_embed::Embed;
+
+pub fn merge_non_null_json_value_into(source: serde_json::Value, target: &mut serde_json::Value) {
```
Byron: The overhead to get this amount of code seems disproportional.

My suggestion is to adapt it, and make it better. The problem is that it can't be much different for what it's doing, but I tried anyway, and added some tests:

```patch
commit 5a8c8505f768c06eeaae3cfd419fa4dfb198e51c
Date:   Thu Dec 12 08:13:39 2024 +0100

    tmp

index f3527b413..652855d09 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2833,7 +2833,6 @@ version = "0.0.0"
 dependencies = [
  "anyhow",
  "gitbutler-fs",
- "gitbutler-settings",
  "gitbutler-settings-util",
  "rust-embed",
  "serde",
diff --git a/crates/gitbutler-settings/Cargo.toml b/crates/gitbutler-settings/Cargo.toml
index 8dfa14118..7257338bd 100644
--- a/crates/gitbutler-settings/Cargo.toml
+++ b/crates/gitbutler-settings/Cargo.toml
@@ -18,5 +18,3 @@ gitbutler-fs.workspace = true
 name = "settings"
 path = "tests/mod.rs"
 
-[dev-dependencies]
-gitbutler-settings.workspace = true
diff --git a/crates/gitbutler-settings/src/persistence.rs b/crates/gitbutler-settings/src/persistence.rs
index 957580480..ac6578a71 100644
--- a/crates/gitbutler-settings/src/persistence.rs
+++ b/crates/gitbutler-settings/src/persistence.rs
@@ -3,7 +3,7 @@ use std::path::PathBuf;
 use crate::app_settings::AppSettings;
 use crate::json::json_difference;
 use anyhow::Result;
-use gitbutler_settings_util::{asset_str, merge_non_null_json_value_into};
+use gitbutler_settings_util::{asset_str, merge_non_null_json_value};
 use rust_embed::Embed;
 use serde_json::json;
 
@@ -27,7 +27,7 @@ impl AppSettings {
         let customizations = serde_json_lenient::from_str(&std::fs::read_to_string(config_path)?)?;
 
         // Merge the customizations into the settings
-        merge_non_null_json_value_into(customizations, &mut settings);
+        merge_non_null_json_value(customizations, &mut settings);
         Ok(serde_json::from_value(settings)?)
     }
 
@@ -53,7 +53,7 @@ impl AppSettings {
 
         // Merge the new customizations into the existing ones
         // TODO: This will nuke any comments in the file
-        merge_non_null_json_value_into(diff, &mut customizations);
+        merge_non_null_json_value(diff, &mut customizations);
         gitbutler_fs::write(config_path, customizations.to_string())?;
         Ok(())
     }
diff --git a/crates/gitbutler-settings/vendor/util/src/lib.rs b/crates/gitbutler-settings/vendor/util/src/lib.rs
index d4e3d02d5..30d2204f4 100644
--- a/crates/gitbutler-settings/vendor/util/src/lib.rs
+++ b/crates/gitbutler-settings/vendor/util/src/lib.rs
@@ -2,20 +2,22 @@ use std::borrow::Cow;
 
 use rust_embed::Embed;
 
-pub fn merge_non_null_json_value_into(source: serde_json::Value, target: &mut serde_json::Value) {
+/// Based on Zed `merge_non_null_json_value_into`
+/// Note: This doesn't merge arrays.
+pub fn merge_non_null_json_value(source: serde_json::Value, target: &mut serde_json::Value) {
     use serde_json::Value;
     if let Value::Object(source_object) = source {
         let target_object = if let Value::Object(target) = target {
             target
         } else {
-            *target = Value::Object(Default::default());
-            target.as_object_mut().unwrap()
+            *target = serde_json::json!({});
+            target.as_object_mut().expect("object was just set")
         };
         for (key, value) in source_object {
             if let Some(target) = target_object.get_mut(&key) {
-                merge_non_null_json_value_into(value, target);
+                merge_non_null_json_value(value, target);
             } else if !value.is_null() {
-                target_object.insert(key.clone(), value);
+                target_object.insert(key, value);
             }
         }
     } else if !source.is_null() {
@@ -23,6 +25,36 @@ pub fn merge_non_null_json_value_into(source: serde_json::Value, target: &mut se
     }
 }
 
+// TODO: put into ./tests/
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn it_does_not_merge_null_values() {
+        let source = serde_json::json!({"a": null, "b": true });
+        let mut target = serde_json::json!({});
+        merge_non_null_json_value(source, &mut target);
+        assert_eq!(target, serde_json::json!({"b": true }));
+    }
+
+    #[test]
+    fn it_does_not_merge_arrays() {
+        let source = serde_json::json!({"a": null, "b": [1,2,3]});
+        let mut target = serde_json::json!({"a": {"b": 1}, "b": [42]});
+        merge_non_null_json_value(source, &mut target);
+        assert_eq!(target, serde_json::json!({"a": {"b": 1}, "b": [1,2,3] }));
+    }
+
+    #[test]
+    fn it_merges_nested_objects_correctly() {
+        let source = serde_json::json!({"a": {"b": {"c": 42}}});
+        let mut target = serde_json::json!({});
+        merge_non_null_json_value(source.clone(), &mut target);
+        assert_eq!(target, source);
+    }
+}
+
 /// Get an embedded file as a string.
 pub fn asset_str<A: Embed>(path: &str) -> Cow<'static, str> {
     match A::get(path).unwrap().data {
```

That way the whole sub-crate could go away, which I'd appreciate.

Byron (commented):
I am really excited to see this new settings system come to life!

Given our recent conversation I felt compelled to comment early, despite the draft status and hope you don't mind.

krlvi:
> I am really excited to see this new settings system come to life!
> 
> Given our recent conversation I felt compelled to comment early, despite the draft status and hope you don't mind.

Thank you, this was really helpful. I have now implemented your feedback

Byron:
Thanks for inviting me to take another early look!

I do have a patch for you which makes the ownership situation of the `SettingsHandle` internals more clear. It's not implemented as a semi-singleton anymore, but I don't think it needs to be as the app creates only one instance, and creating more will also work but duplicate events.

```patch
diff --git a/crates/gitbutler-settings/src/watch.rs b/crates/gitbutler-settings/src/watch.rs
index a9b3bee1a..ac517c803 100644
--- a/crates/gitbutler-settings/src/watch.rs
+++ b/crates/gitbutler-settings/src/watch.rs
@@ -1,6 +1,6 @@
 use std::{
     path::PathBuf,
-    sync::{mpsc::channel, Arc, OnceLock, RwLock, RwLockReadGuard},
+    sync::{mpsc, Arc, RwLock, RwLockReadGuard},
     time::Duration,
 };
 
@@ -10,7 +10,7 @@ use notify::{event::ModifyKind, Config, Event, RecommendedWatcher, RecursiveMode
 
 pub struct SettingsHandle {
     config_path: PathBuf,
-    app_settings: &'static RwLock<AppSettings>,
+    app_settings: Arc<RwLock<AppSettings>>,
     #[allow(clippy::type_complexity)]
     send_event: Arc<dyn Fn(AppSettings) -> Result<()> + Send + Sync + 'static>,
 }
@@ -25,9 +25,7 @@ impl SettingsHandle {
         let config_path = config_dir.into().join(SETTINGS_FILE);
         let app_settings = app_settings::AppSettings::load(config_path.clone())?;
 
-        static CONFIG: OnceLock<RwLock<AppSettings>> = OnceLock::new();
-        let app_settings = CONFIG.get_or_init(|| RwLock::new(app_settings));
-
+        let app_settings = Arc::new(RwLock::new(app_settings));
         Ok(Self {
             config_path,
             app_settings,
@@ -42,8 +40,8 @@ impl SettingsHandle {
     }
 
     pub fn watch_in_background(&self) -> Result<()> {
-        let (tx, rx) = channel();
-        let settings = self.app_settings;
+        let (tx, rx) = mpsc::channel();
+        let settings = self.app_settings.clone();
         let config_path = self.config_path.clone();
         let send_event = self.send_event.clone();
 
@@ -59,14 +57,14 @@ impl SettingsHandle {
                         kind: notify::event::EventKind::Modify(ModifyKind::Data(_)),
                         ..
                     })) => {
-                        if let Ok(mut s) = settings.try_write() {
-                            if let Ok(update) = app_settings::AppSettings::load(config_path.clone())
-                            {
-                                if *s != update {
-                                    tracing::info!("settings.json modified; refreshing settings");
-                                    *s = update.clone();
-                                    send_event(update)?;
-                                }
+                        let Ok(mut last_seen_settings) = settings.try_write() else {
+                            continue;
+                        };
+                        if let Ok(update) = app_settings::AppSettings::load(config_path.clone()) {
+                            if *last_seen_settings != update {
+                                tracing::info!("settings.json modified; refreshing settings");
+                                *last_seen_settings = update.clone();
+                                send_event(update)?;
                             }
                         }
                     }
diff --git a/crates/gitbutler-tauri/src/main.rs b/crates/gitbutler-tauri/src/main.rs
index abc782c8a..dbcd546c5 100644
--- a/crates/gitbutler-tauri/src/main.rs
+++ b/crates/gitbutler-tauri/src/main.rs
@@ -105,9 +105,11 @@ fn main() {
 
                     app_handle.manage(WindowState::new(app_handle.clone()));
 
-                    let handle = app_handle.clone();
-                    let app_settings = SettingsHandle::create(config_dir, move |app_settings| {
-                        gitbutler_tauri::ChangeForFrontend::from(app_settings).send(&handle)
+                    let app_settings = SettingsHandle::create(config_dir, {
+                        let app_handle = app_handle.clone();
+                        move |app_settings| {
+                            gitbutler_tauri::ChangeForFrontend::from(app_settings).send(&app_handle)
+                        }
                     })?;
                     app_settings.watch_in_background()?;
                     app_handle.manage(app_settings);
```

The `SettingsHandle` is the source of truth for what's up-to-date app-settings, and if it changed, it sends an event to the frontend window(s). That should work even though I fear unnecessary updates when the UI triggers changes to the settings, either by writing the file directly (probably nothing we'd want) or by sending update events that ultimately trigger the settings file to be written. As the source of truth now learns about the change through a filesystem event, the UI will soon receive another update of the app-settings it just sent. That *should* be fine, and I assume other windows will also get a copy so they actually get to update themselves with the changes.

Something that isn't solved is race-conditions around multiple windows changing the configuration concurrently, but one might hope users are slow enough to not run into this. A smaller poll-interval for the watcher might then do the trick.

Caleb-T-Owens:
All these cool kids using Arc 😆

Byron (approved):
I went 'all in' on the review and hope I left it more maintainable than before. With that said, please feel free to dial it back here and there. Particularly the names might be overly verbose and could just be simpler.

Please do all the application settings real-world testing once again, I only launched the app once and it *seemed* to work, but I didn't do any app-setting specific testing.

PR #5819, opened 2024-12-12
fix: Dependencies should ignore all line-deletion only hunks

estib-vega:
## ☕️ Reasoning
Ignore line-deletions only hunks, even if those are added out the start.
Hunk intersection calculation fails gracefully.

## 🧢 Changes
- Updated the logic to ignore hunks with 0 lines added, ensuring that such modifications do not interfere with dependency calculations.
- Improved error handling in hunk operations to prevent crashes due to potential subtraction overflows by returning appropriate result types.
- Modified relevant tests to account for these updates and verify that the changes function correctly.


krlvi (approved):
Appears to do what it says on the tin. Btw, this panicless_sub can be implemented as a trait

PR #5820, opened 2024-12-12
fix: compress signed msi for download by tauri updater

ndom91:
## ☕️ Reasoning

- Compress the MSI after signing and replace the unsigned `*.msi.zip` in the S3 bucket dir for download by the Tauri updater.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -242,6 +242,8 @@ jobs:
           find release -name "*.msi" -type f -print0 | xargs -0 -n1 -I{} python3 /sign-with-evcodesignd.py "{}"
         env:
           EVCODESIGND_PSK: ${{ secrets.EVCODESIGND_PSK }}
+      - name: Compress signed msi
+        run: find release -name "*.msi" -type f -print0 | zip "release/$(find release -name "*.msi" -type f -printf '%P\n').zip" -@
       - name: Upload signed artifacts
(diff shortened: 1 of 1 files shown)


ndom91:
@schacon unfortunately wasn't able to finish this before holidays. Running this in GHA produced all sorts of weird issues like not finding `zip`, then not allowing `sudo apt install zip`, then not finding `sudo`, then running `apt install zip` couldn't resolve `packages.ubuntu.org` :joy: 

So definitely needs some more debugging, but anyway here's the `minisign-cli` GHA you mentioned: https://github.com/marketplace/actions/minisign-cli

.github/workflows/publish.yaml
```diff
@@ -236,12 +236,27 @@ jobs:
         with:
           name: '${{ needs.build-tauri.outputs.channel }}-${{ matrix.platform }}-${{ github.run_number }}'
           path: release
+      - name: Set file as a variable
+        id: set-path
+        run: |
+          VALUE=$(find release -name "*.msi" -type f -printf '%P\n')
+          echo "::set-output name=msi_file::$VALUE"
```
ndom91: Is this syntax still supported? I thought it was:

```
- name: Set output
  run: echo "{name}={value}" >> $GITHUB_OUTPUT
```

schacon: Old school.

Technically they are still supported, but yes, I will change them to the new format.

PR #5824, opened 2024-12-12
fix: rm superfluous checkMonitors check

ndom91:
## ☕️ Reasoning

- `checkMonitor` was beign initialized on _any_ branch, meaning it was attemptingt o call GitHub and check the check status for any branch whether it had a PR or not.

## 🧢 Changes

- Rm additional `checkMontiro` instantiation. The `canMergeAll()` derived seems to still be updated at the correct points in time, so that the "Merge All" button is visible on a set of branches when all their PR's checks are passing.


mtsgrd (approved):
perfect

PR #5834, opened 2024-12-13
Update markdown content types to avoid depricated svelte

Caleb-T-Owens:


ndom91 (approved):
Haven't tested it, but code-wise looks good!

Issue #5839, opened 2024-12-15
SSH: Unable to find program


TGTGamer:
### Version

0.14.4

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

Using 1password as my ssh agent. Was working until I updated to latest version: 

SSH agent path: C:\Program Files\1Password\app\8\op-ssh-sign.exe
Confirmed it exists, is working for other programs

### Expected behavior

Should sign commits. 

### Relevant log output

```shell
Program not found
user.C034edPS.js:4 
 ipc->check_signing_settings: {"id":"31d6e997-0f4b-4b08-8a9a-e666c46c1609"} s: Program not found
    at s.fromError (http://tauri.localhost/_app/immutable/chunks/ipc.DM3tJdKb.js:1:1093)
    at w (http://tauri.localhost/_app/immutable/chunks/ipc.DM3tJdKb.js:1:1297)
    at async Object.C [as onclick] (http://tauri.localhost/_app/immutable/nodes/27.0QkJqKa9.js:10:818) 
Object
user.C034edPS.js:4 
 Error checking signing: s: Program not found
    at s.fromError (http://tauri.localhost/_app/immutable/chunks/ipc.DM3tJdKb.js:1:1093)
    at w (http://tauri.localhost/_app/immutable/chunks/ipc.DM3tJdKb.js:1:1297)
    at async Object.C [as onclick] (http://tauri.localhost/_app/immutable/nodes/27.0QkJqKa9.js:10:818)
(anonymous)	@	user.C034edPS.js:4
```

Byron:
Thanks a lot for reporting!

I am pretty sure that #5756 has something to do with it.

- The problem is that on Windows, the current implementation won't use a shell which is probably required for all required environment variables to be picked up.

This I will address, hoping that it will work again then.

DS-DanielW:
Still getting this same issue on the latest nightly with the potential fix merged.

Caleb-T-Owens:
@Byron I've taken a look at the PR which seems to use [with_shell](https://docs.rs/gix/latest/gix/diff/command/struct.Prepare.html#method.with_shell) which is always uses `sh`, which most people probably don't have an environment set up for. We possibly ought to try doing `[with_shell_program](https://docs.rs/gix/latest/gix/diff/command/struct.Prepare.html#method.with_shell_program)` and pass in `$SHELL` by default, and have a user setting that lets them override it.

WDYT?

[...]

PR #5841, opened 2024-12-16
maybe fix 5839

Byron:
This PR makes a few changes to assure the shell is also used on Windows when invoking `gpg`.

That way, more of the typical environment may be picked up, which hopefully affects #5839 positively.

The strange thing is that there it worked before without a shell as well, and since on Windows it did manual argument splitting, ultimately it would make the same command-call like before, calling a pure `gpg` binary without shell involvement.

So chances are that besides unifying the operation on all platforms by using a shell, it might still not work as expected.

Maybe there are other changes that affect #5839 between v0.14.3 and v0.14.4?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2962,7 +2962,7 @@ dependencies = [
 [[package]]
 name = "gix"
 version = "0.68.0"
-source = "git+https://github.com/GitoxideLabs/gitoxide?rev=520c832cfcfb34eb7617be55ebe2719ab35595fd#520c832cfcfb34eb7617be55ebe2719ab35595fd"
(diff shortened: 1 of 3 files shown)


mtsgrd (approved):

mtsgrd:
@ndom91 is there any way to configure an allowlist user filter for the vercel integration?

ndom91:
> @ndom91 is there any way to configure an allowlist user filter for the vercel integration?

You mean like to run the actions on forked PRs? Not that I know of. There is a global switch to allow them to run on PRs from forked repos of users who aren't a part of the Vercel team though

Byron:
Let's merge this, I think it will do what it should on Windows and it's a chance it will work better than before as Windows will also run it through a shell then, just like it does on Unix. Maybe that helps it pick up more of the environment it needs.

PR #5856, opened 2024-12-19
feat: enhance context menu labels for branch lane actions

PavelLaptev:
**Reasoning:**  

Make the "Unapply" and "Unapply and drop changes" options clearer in different lane states.  

**Related issues:** [#5683](https://github.com/gitbutlerapp/gitbutler/issues/5683)  

---

_Here is how it looks:_

image

---

Let me know if you have any improvements.

### DIFF
--- a/apps/desktop/src/lib/branch/BranchLaneContextMenu.svelte
+++ b/apps/desktop/src/lib/branch/BranchLaneContextMenu.svelte
@@ -52,6 +52,10 @@
 	function saveAndUnapply() {
 		branchController.saveAndUnapply(stack.id);
 	}
+
+	const isVirginLane = $derived(
+		stack.name.toLowerCase().includes('lane') && commits.length === 0 && stack.files?.length === 0
+	);
 </script>
 
 <ContextMenu bind:this={contextMenuEl} leftClickTrigger={trigger} {ontoggle}>
@@ -67,6 +71,7 @@
 	<ContextMenuSection>
 		<ContextMenuItem
 			label="Unapply"
+			disabled={isVirginLane}
 			onclick={async () => {
 				if (commits.length === 0 && stack.files?.length === 0) {
 					await branchController.unapplyWithoutSaving(stack.id);
@@ -78,7 +83,8 @@
 		/>
 
 		<ContextMenuItem
-			label="Unapply and drop changes"
+			label={isVirginLane ? 'Remove lane' : 'Unapply & drop changes'}
+			disabled={stack.files?.length === 0 && commits.length !== 0}
 			onclick={async () => {
 				if (
 					stack.name.toLowerCase().includes('lane') &&


mtsgrd:
@krlvi would now be a good time to start changing how this works? Even with these improvements by Pavel I find both the language and the current outcomes very confusing. Iirc the consensus is to make creating a local branch an explicit "export" action, which should resolve the ambiguity of what it means to unapply.

Caleb-T-Owens:
Looking back at the original issue, I'm a little disappointed in myself that I didn't spot this, and am sorry for causing extra work.

We've now got the button text changing based on whether or not the branch is pushed or not, which I'm trying to decide whether it makes sense.

Whether a branch is pushed or not has no effect on how the two options work.

The behaviours are as follows:

#### If no local branch exists for a lane:
- Unapply & Save changes -> Writes out a local branch that contains those changes and the branch is gone from the board.
- Unapply & Drop changes -> Changes are lost and the branch is gone from the board.

#### If a local branch exists for a lane:
- Unapply & Save changes -> Updates the existing local branch to contain those changes and the branch is gone from the board.
- Unapply & Drop changes -> Changes are lost and the branch is gone from the board. The existing local branch is left alone.

apps/desktop/src/lib/branch/BranchLaneContextMenu.svelte
```diff
@@ -52,6 +52,22 @@
 	function saveAndUnapply() {
 		branchController.saveAndUnapply(branch.id);
 	}
+
+	const isVirginLane = $derived(
+		branch.name.toLowerCase().includes('lane') && commits.length === 0 && branch.files?.length === 0
```
Caleb-T-Owens: This doesn't make sense as a pushed/unpushed check, how do we tell if all of the branches in a given lane are integrated. Presumably we could use that mechanism.

PavelLaptev:
I see. I've converted this to a draft.
I don't have a strong opinion here. I just want to address the issue, because I also think that "unapply and drop changes" shouldn't exists or should be disabled if there are no changes.

_btw Instead of disabling certain labels, I hid them. But maybe we should keep showing them in a disabled state instead._

@mtsgrd @Caleb-T-Owens, do you have any concrete suggestions regarding the labels?

Caleb-T-Owens:
As we did identify in the issue, if a branch has no commits or changes, then hiding/disabling `Unapply & Save changes` makes sense because there is nothing to save.

As for the other states, I'm not sure changing between `Remove` or `Unapply` conditionally makes sense. To me adding extra inconsistency

PavelLaptev:
> As for the other states, I'm not sure changing between `Remove` or `Unapply` conditionally makes sense. To me adding extra inconsistency

Got it. Let me update this

Caleb-T-Owens:
I guess when it's an empty branch and is only showing `Unapply & Drop changes` then `Remove` might make more sense?

I do wonder if it would be preferable to keep it with the longer label, and have two tooltips that explain what the two actions do:

`Write lane to git branch and remove lane from board` & `Remove lane from board without saving` - or something like that.

Caleb-T-Owens:
> Got it. Let me update this

Thank you for your patience!!!! ❤️

PavelLaptev:
@Caleb-T-Owens No problem! 😊 

Could we also update the docs? I'm relying more on the docs with the new software rather than tooltips. It looks like the section on "applying and unapplying branches" is outdated.  

Basically:  

image  

I'm also not sure if we're still using the term "Virtual branches."  

image

PavelLaptev:
@Caleb-T-Owens updated variant 
image

Caleb-T-Owens:
Why does it still have two extra states that depend on whether the lane has been pushed or not?

PavelLaptev:
> Why does it still have two extra states that depend on whether the lane has been pushed or not?

I thought there were three conditions:  
1. **Not pushed, no changes** — Disable "Unapply" and rename "Unapply & drop changes" to "Remove lane"  
2. **Pushed, no changes** — Disable "Unapply & drop changes"  
3. **All other cases** — No disabled buttons or copy changes.

Caleb-T-Owens:
@PavelLaptev I laid out the behaviour of the two buttons in this message: https://github.com/gitbutlerapp/gitbutler/pull/5856#issuecomment-2558234411

The behaviour doesn't change if you've pushed or not.

Does the behaviour of the buttons make sense?

PavelLaptev:
@Caleb-T-Owens I’m not changing the button behavior. According to [this issue](https://github.com/gitbutlerapp/gitbutler/issues/5683), if there are no changes and no pushed/local branches, rename "Unapply and drop changes" to "Remove" and disable "Unapply" because it doesn’t make sense.

PavelLaptev:
I closed this PR because we will address this more effectively in V3

PR #5857, opened 2024-12-19
Make committed file id distinct from uncommitted file id

mtsgrd:
- It's noteworthy that the id for files is just the filepath
- We could/should consider some hash instead
- fixes "ownership not found" edge case in amending files

---
- <kbd>&nbsp;2&nbsp;</kbd> #5858 
- <kbd>&nbsp;1&nbsp;</kbd> #5857 👈


krlvi (approved):
interesting

PR #5859, opened 2024-12-19
Rename one more occurrence of data -> dropData

mtsgrd:
---
- <kbd>&nbsp;2&nbsp;</kbd> #5860 
- <kbd>&nbsp;1&nbsp;</kbd> #5859 👈

### DIFF
--- a/apps/desktop/src/lib/commit/CommitCard.svelte
+++ b/apps/desktop/src/lib/commit/CommitCard.svelte
@@ -262,7 +262,7 @@
 				date: getTimeAgo(commit.createdAt),
 				authorImgUrl: authorImgUrl,
 				commitType: type,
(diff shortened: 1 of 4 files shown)


Caleb-T-Owens:
Does the `data` property of a `Draggable` object need to explicitly state that it gets passed to the ondrop/accepts handler?

Caleb-T-Owens (approved):
Not sure about the name, but happy to accept it fixes a bug. I'm a little bemused however at how this didn't get caught?

mtsgrd:
Actually, this didn't cause any bugs. I thought it did, but these names are disconnected from the other `dropData` names.

> Does the data property of a Draggable object need to explicitly state that it gets passed to the ondrop/accepts handler?

 and.. I think you're right, let's not land this PR.

PR #5860, opened 2024-12-19
Rename some draggable things -> droppable things

mtsgrd:
Instances of classes like DraggableFile are used as drop data, not for setting up or handling the dragging.

---
- <kbd>&nbsp;2&nbsp;</kbd> #5860 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #5859


Caleb-T-Owens:
If these classes don't manage any part of the drag/drop lifecycle, then surly by your logic they should be neither `Draggable` or `Droppable`. They're just data.

mtsgrd:
> If these classes don't manage any part of the drag/drop lifecycle

This is not what I say in the description. They are not _draggable_, they are only used as drop data.

PR #5863, opened 2024-12-20
Rename some draggable things -> drop data things

mtsgrd:
Instances of classes like DraggableFile are used as drop data, not for setting up or handling the dragging.


Caleb-T-Owens (approved):
Thanks, this is nice :+1:

Issue #5594, opened 2024-11-19
Commit message character counter - include the message body in the counting


krlvi:
https://discord.com/channels/1060193121130000425/1073202153163857920/1308337642105999382

PavelLaptev:
The title counter is meant to help users keep the summary within a limit—for example, 50 characters in GitHub.
But what’s the purpose of showing a character counter for the body?

PavelLaptev:
Closing this issue. It should be included in the next release.

PR #5864, opened 2024-12-20
Character counter for message body

PavelLaptev:
## ☕️ Reasoning

Related issue https://github.com/gitbutlerapp/gitbutler/issues/5594

https://github.com/user-attachments/assets/15351694-18fd-4888-8a7e-a8dba7182b0d

### DIFF
--- a/apps/desktop/src/lib/commit/CommitMessageInput.svelte
+++ b/apps/desktop/src/lib/commit/CommitMessageInput.svelte
@@ -57,8 +57,11 @@
 
 	let titleTextArea: HTMLTextAreaElement | undefined = $state();
 	let descriptionTextArea: HTMLTextAreaElement | undefined = $state();
+	let isTitleFocused = $state(true);
+	let isDescriptionFocused = $state(false);
 
 	const { title, description } = $derived(splitMessage(commitMessage));
+
 	$effect(() => {
 		valid = !!title;
 	});
@@ -202,9 +205,15 @@
 	export function focus() {
 		titleTextArea?.focus();
 	}
-
-	const maxTitleLength = 50;
 </script>
+
+{#snippet charCounter(tooltip: string, count: number)}
+	<Tooltip text={tooltip}>
+		<span class="text-11 text-semibold text-body commit-box__textarea-char-counter-label">
+			{count}
+		</span>
+	</Tooltip>
+{/snippet}
 
 {#if isExpanded}
 	<div class="commit-box__textarea-wrapper text-input">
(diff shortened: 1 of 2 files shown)


apps/desktop/src/lib/commit/CommitMessageInput.svelte
```diff
@@ -249,16 +272,12 @@
...
-				<span class="text-11 text-semibold text-body commit-box__textarea-char-counter-label">
-					{title.length}
-				</span>
-			</Tooltip>
+		{#if title.length > 0 && isTitleFocused}
+			{@render charCounter('Summary chars', title.length)}
+		{/if}
+
+		{#if description.length > 0 && isDescriptionFocused}
+			{@render charCounter('Description chars', description.length)}
```
Caleb-T-Owens, line 280: Should this display the character count for the description line that you're currently on, rather than the entire description character count?

PavelLaptev: Understand why. However, this will be more complex because we need to dynamically determine where the line ends and the current position of the text caret, as it's a single textarea field.

PavelLaptev: In V3, we’ll have an auto-wrap option. For now, I’m not sure if it’s worth investing more time in this—unless you already have a working solution. In that case, I’d be happy to copy/paste it! 😆

Byron (commented):
Intuitively I really liked it!

Just as a note, what seems to be counted is unicode codepoints, and what's displayed as single 'character' for lack of a better word can be 'overcounted' that way.

Screenshot 2024-12-20 at 17 11 42

No need to fix it, it's probably something that will show up as an issue at some point when people with this "Emoji-Subject" convention notice this.

Screenshot 2024-12-20 at 17 15 13

Otherwise it seems there are packages that can count graphemes as well, which would fix this particular problem.

Caleb-T-Owens (approved):
Happy to approve with the later plans for improvement

PavelLaptev:
> Just as a note, what seems to be counted is unicode codepoints, and what's displayed as single 'character' for lack of a better word can be 'overcounted' that way.

I see :-) But it's only my assumption that most linters don’t care about treating some emojis as a single symbol and also "overcount" them

ryanhiebert:
> what's displayed as single 'character' for lack of a better word can be 'overcounted' that way.

Interestingly, the particular character you took a screenshot of is a double width character, and takes the space of two characters in a fixed-with environment like a terminal. So depending the purpose of you line limits, you may _want_ it to count as two characters.

PR #5865, opened 2024-12-20
Use correct js single line comment style

mtsgrd:
- e.g. vscode does not recognise /** */ and can't uncomment it

### DIFF
--- a/apps/desktop/src/lib/config/appSettingsV2.ts
+++ b/apps/desktop/src/lib/config/appSettingsV2.ts
@@ -12,8 +12,6 @@ export class SettingsService {
 	});
 
 	private async handlePayload(settings: AppSettings) {
-		// TODO: Remove this log
-		console.log(settings);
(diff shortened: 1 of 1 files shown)


Caleb-T-Owens:
the `/** .... */` style comment is a ts/jsdoc https://tsdoc.org style comment.

When used to describe a class/function/property/etc... it comes up in vscode type hints:

With tsdoc comment:
Screenshot 2024-12-20 at 17 21 21

With regular comment:
Screenshot 2024-12-20 at 17 21 40

Caleb-T-Owens:
Your editor probably doesn't uncomment tsdoc comments with your keybind because that is likely an accidental keypress. If you do have a regular block comment, it should correctly uncomment as expected.

mtsgrd:
Thanks. TIL, sorry for changing your comments @krlvi, we should clearly be using that style for documenting fields on classes.

PR #5867, opened 2024-12-21
Fix reactivity bug preventing Ollama prompt save

mtsgrd:


Caleb-T-Owens (approved):
I spent half an hour trying to figure that out 🤦.

krlvi:
Can someone ELI5 this one to me? Why does this work?

Caleb-T-Owens:
@krlvi 

Under the hood, `foo.bar = 69` roughly equates to `get(foo).bar = 69` which doesn't trigger any subscribers. Adding `foo = foo` introduces something like `foo.set(get(foo))` which then triggers the update.

It seems that with the svelte 5 update, this has gotten less reactive.

Caleb-T-Owens:
AFAIK, updating the property should work, so this might be a svelte related bug.

PR #5881, opened 2025-01-02
Introduce commit subset check (part of stack)

Caleb-T-Owens:
Introduces a function for determining if the changes introduced by a commit, are a subset of the changes introduced by another.

This is helpful for an upcoming PR which should greatly improve our integration check.

---

`get_exclusive_tree` is able to take a stack of commits IE `Base -> A -> B`. And return a tree which only contains the introductions of the `B` commit (when diffed against the Base).

We do this by merging commits `B` and `Base` against the parent of `B`. Any conflicts are resolved in favour of the `B` commit.

`is_subset` makes use of `get_exclusive_tree` to be able to compare the subset and superset candidates without worry about changes that they may have been cherry-picked on top of

---
- <kbd>&nbsp;2&nbsp;</kbd> #5893 
- <kbd>&nbsp;1&nbsp;</kbd> #5881 👈

### DIFF
--- a/crates/gitbutler-branch-actions/src/commit_ops.rs
+++ b/crates/gitbutler-branch-actions/src/commit_ops.rs
@@ -1,0 +1,238 @@
+use anyhow::{bail, Result};
+use gitbutler_oxidize::GixRepositoryExt;
+
+/// Finds the first parent of a given commit
+fn get_first_parent(commit: gix::Commit) -> Result<gix::Commit> {
+    let first_parent = commit.parent_ids().take(1).collect::<Vec<_>>();
+    let Some(first_parent) = first_parent.first() else {
+        bail!("Failed to find first parent of {}", commit.id())
+    };
+    let first_parent = first_parent.object()?.into_commit();
+    Ok(first_parent)
+}
+
+/// Gets the changes that one commit introduced compared to the base,
+/// excluding anything between the commit and the base
+fn get_exclusive_tree(
+    repository: &gix::Repository,
+    commit_id: gix::ObjectId,
+    base_id: gix::ObjectId,
+) -> Result<gix::ObjectId> {
+    let commit = repository.find_commit(commit_id)?;
+    let commit_parent = get_first_parent(commit.clone())?;
+    let base = repository.find_commit(base_id)?;
+
+    repository
+        .merge_trees(
+            commit_parent.tree_id()?,
+            commit.tree_id()?,
+            base.tree_id()?,
+            Default::default(),
+            repository.merge_options_force_ours()?,
+        )?
+        .tree
+        .write()
+        .map_err(Into::into)
+        .map(Into::into)
+}
+
+/// Takes two commits and determines if one is a subset of or equal to the other
+#[allow(dead_code)]
+fn is_subset(
+    repository: &gix::Repository,
+    superset_id: gix::ObjectId,
+    subset_id: gix::ObjectId,
+    common_base_id: gix::ObjectId,
+) -> Result<bool> {
+    let exclusive_superset = get_exclusive_tree(repository, superset_id, common_base_id)?;
+    let exclusive_subset = get_exclusive_tree(repository, subset_id, common_base_id)?;
+
+    let common_base = repository.find_commit(common_base_id)?;
+
+    let (options, _) = repository.merge_options_fail_fast()?;
+    let mut merged_exclusives = repository.merge_trees(
+        common_base.tree_id()?,
+        exclusive_superset,
+        exclusive_subset,
+        Default::default(),
+        options,
+    )?;
+
+    if merged_exclusives.conflicts.is_empty() {
+        Ok(exclusive_superset == merged_exclusives.tree.write()?)
+    } else {
+        Ok(false)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use gitbutler_testsupport::testing_repository::TestingRepository;
+    mod get_exclusive_tree {
+        use gitbutler_oxidize::OidExt;
+
+        use super::super::get_exclusive_tree;
+        use super::*;
+
+        #[test]
+        fn when_already_exclusive_returns_self() {
+            let test_repository = TestingRepository::open();
+            let base_commit: git2::Commit =
+                test_repository.commit_tree(None, &[("foo.txt", "foo"), ("bar.txt", "bar")]);
+            let second_commit: git2::Commit = test_repository.commit_tree(
+                Some(&base_commit),
+                &[("foo.txt", "bar"), ("bar.txt", "baz")],
+            );
+
+            let exclusive_tree = get_exclusive_tree(
+                &test_repository.gix_repository(),
+                second_commit.id().to_gix(),
+                base_commit.id().to_gix(),
+            )
+            .unwrap();
+
+            assert_eq!(
+                second_commit.tree_id().to_gix(),
+                exclusive_tree,
+                "The tree returned should match the second commit"
+            )
+        }
+
+        #[test]
+        fn when_on_top_of_other_commit_its_changes_are_dropped() {
+            let test_repository = TestingRepository::open();
+            let base_commit: git2::Commit =
+                test_repository.commit_tree(None, &[("foo.txt", "foo")]);
+            let second_commit: git2::Commit = test_repository.commit_tree(
+                Some(&base_commit),
+                &[("foo.txt", "bar"), ("bar.txt", "baz")],
+            );
+            let third_commit: git2::Commit = test_repository.commit_tree(
+                Some(&second_commit),
+                &[("foo.txt", "bar"), ("bar.txt", "baz"), ("qux.txt", "bax")],
+            );
+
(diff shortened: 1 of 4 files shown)


crates/gitbutler-oxidize/src/lib.rs
```diff
@@ -15,10 +15,30 @@ pub fn git2_to_gix_object_id(id: git2::Oid) -> gix::ObjectId {
     gix::ObjectId::try_from(id.as_bytes()).expect("git2 oid is always valid")
 }
 
+pub trait OidExt {
+    fn to_gix(self) -> gix::ObjectId;
+}
+
+impl OidExt for git2::Oid {
+    fn to_gix(self) -> gix::ObjectId {
```
Byron, line 23: That's wonderful!

crates/gitbutler-oxidize/src/lib.rs
```diff
@@ -101,3 +121,16 @@ pub fn gix_to_git2_index(index: &gix::index::State) -> anyhow::Result<git2::Inde
     }
     Ok(out)
 }
+
+pub fn print_tree(tree: gix::Tree<'_>) {
```
Byron, line 125: This would probably be a better fit for the `testsupport` crate.

crates/gitbutler-oxidize/src/lib.rs
```diff
@@ -101,3 +121,16 @@ pub fn gix_to_git2_index(index: &gix::index::State) -> anyhow::Result<git2::Inde
...
+    let mut recorder = gix::traverse::tree::Recorder::default();
+    tree.traverse().breadthfirst(&mut recorder).unwrap();
```
Byron, line 127: When implementing the depth-first traversal I noticed that this isn't really breadth-first as it effectively lists files first, then directories, which makes for a funky ordering that isn't too helpful when debugging.

The old version of `gix tree entries -r` shows the effects of this nicely.

The reason I never noticed is that I never used it, but went with `termtree` instead.

```Rust
pub fn visualize_tree(
    id: &gix_hash::oid,
    odb: &impl gix_object::Find,
    name_and_mode: Option<(&BStr, EntryMode)>,
) -> termtree::Tree<String> {
    fn short_id(id: &gix_hash::oid) -> String {
        id.to_string()[..7].to_string()
    }
    let entry_name = |id: &gix_hash::oid, name: Option<(&BStr, EntryMode)>| -> String {
        let mut buf = Vec::new();
        match name {
            None => short_id(id),
            Some((name, mode)) => {
                format!(
                    "{name}:{mode}{} {}",
                    short_id(id),
                    match odb.find_blob(id, &mut buf) {
                        Ok(blob) => format!("{:?}", blob.data.as_bstr()),
                        Err(_) => "".into(),
                    },
                    mode = if mode.is_tree() {
                        "".into()
                    } else {
                        format!("{:o}:", mode.0)
                    }
                )
            }
        }
    };

    let mut tree = termtree::Tree::new(entry_name(id, name_and_mode));
    let mut buf = Vec::new();
    for entry in odb.find_tree(id, &mut buf).unwrap().entries {
        if entry.mode.is_tree() {
            tree.push(visualize_tree(entry.oid, odb, Some((entry.filename, entry.mode))));
        } else {
            tree.push(entry_name(entry.oid, Some((entry.filename, entry.mode))));
        }
    }
    tree
}
```

The above (copied from the tests of the `gix-merge` crate) lists a tree as is and prints all information about it that one would want in a simple recursive function. It can certainly be rewritten to power `print_tree()`, and I think it will be for the better.

crates/gitbutler-branch-actions/src/commit_ops.rs
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    mod is_subset {
+        use gitbutler_oxidize::OidExt;
+
+        use super::super::is_subset;
+        use super::*;
+
+        #[test]
+        fn a_commit_is_a_subset_of_itself() {
```
Byron, line 160: All these tests are read-only, which is when the `read_only::fixture*` methods are faster. They create a fixture only once and make it easy to inspect on disk. They come at the disadvantage of separating the test-setup from the test itself though, but to me it's easy enough to check the script and I can understand more easily how the structure looks like by looking at it, and if in doubt the final structure is easy to inspect on disk.

Of course, these are not the pinnacle of what's possible and I always thought that it would be great to define tests more like `cargo` does, but to apply the same techniques to easily define complex Git repositories.

Byron (commented):
Thanks for the invite, I took a close look, but only from a Rust + `gix` perspective, not from a "does this really work and is it sound" point of view.

For simplicity, I just created two commits with changes I'd do, please see them as suggestions. The second one has more of a "careful, better do it this way" narrative though, but either way they can be changed, refactored or removed.

Generally, when doing checks like these, one should assure that the `Repository` is configured to *not* write objects to disk. This is fast enough to do it on the fly, but as it would require cloning the repository that also strips object caches which would hurt performance if many commits are checked in a row. Thus I added documentation to the method that I think might be public one day, but ideally the top-level caller will take care of that.

I hope this helps.

--- 

As a final something to think about: I personally find it helpful to know about lifetimes despite the added noise, and enable `#![deny(rust_2018_idioms)]` in `lib.rs` in all crates. This came up when seeing `get_first_parent(commit: gix::Commit)` and I had to wonder why it's passed like this, until I realized that it does have a reference to the `Repository` which was just hidden. With `get_first_parent(commit: gix::Commit<'_>)` it would be clear that contains references, just like the `git2` types of the same kind. When passing it as `&` which should be done with a `Clone` type, it also became obvious why it was done like that - having all lifetimes visible may have helped here, too.

crates/gitbutler-oxidize/src/lib.rs
Caleb-T-Owens: I wasn't too concerned with the ordering/bredthfirstness of the function. In some ways, it's possibly more helpful if it prints in same order as the tree entries.

I'll have a think about what a helpful debug output would look like and revisit

crates/gitbutler-oxidize/src/lib.rs
Caleb-T-Owens: I was wanting to use it for debugging inside an application function rather than inside a test case, which afaik, means it can't be inside `testsupport`

Caleb-T-Owens:
> Thanks for the invite, I took a close look, but only from a Rust + gix perspective, not from a "does this really work and is it sound" point of view.

Thank you! This is greatly appreciated

> For simplicity, I just created two commits with changes I'd do, please see them as suggestions. The second one has more of a "careful, better do it this way" narrative though, but either way they can be changed, refactored or removed.

Thanks for these! Both look good.

> Generally, when doing checks like these, one should assure that the Repository is configured to not write objects to disk. This is fast enough to do it on the fly, but as it would require cloning the repository that also strips object caches which would hurt performance if many commits are checked in a row. Thus I added documentation to the method that I think might be public one day, but ideally the top-level caller will take care of that.

The new integration check will run over a full stack rather than getting called on an individual commit. As such, it will almost always be called many times in a row. As such, it makes more sense to open the in-memory repository outside of the `is_subset` function and instead inside the new integration function.

> As a final something to think about: I personally find it helpful to know about lifetimes despite the added noise, and enable #![deny(rust_2018_idioms)] in lib.rs in all crates. This came up when seeing get_first_parent(commit: gix::Commit) and I had to wonder why it's passed like this, until I realized that it does have a reference to the Repository which was just hidden. With get_first_parent(commit: gix::Commit<'_>) it would be clear that contains references, just like the git2 types of the same kind. When passing it as & which should be done with a Clone type, it also became obvious why it was done like that - having all lifetimes visible may have helped here, too.

I'm wary of becoming a perl codebase, but would it make sense to add `#![deny(rust_2018_idioms)]` to the top of new files so we can incrementally move over to better idioms?

crates/gitbutler-oxidize/src/lib.rs
Byron: Within `#[test]` code `dev-dependencies` are available, no matter where it is.
Thus, moving it to the `testsupport` crate will allow it to be used in any test, anywhere, as long as `testsupport` is a `dev-dependency`.

crates/gitbutler-oxidize/src/lib.rs
Byron: Within `#[test]` code `dev-dependencies` are available, no matter where it is.
Thus, moving it to the `testsupport` crate will allow it to be used in any test, anywhere, as long as `testsupport` is a `dev-dependency`.

Byron:
> I'm wary of becoming a perl codebase, but would it make sense to add `#![deny(rust_2018_idioms)]` to the top of new files so we can incrementally move over to better idioms?

It absolutely can be done that way, which certainly helps to get a feel for it. Additionally, I still don't understand how crates specified as `edition = "2021"` won't even warn about the rust-2018 idioms.
The reason it's not the default might be exactly due to the added noise, even though I find it most useful to know if references are involved.

krlvi (approved):

PR #5882, opened 2025-01-02
Exit app when last window is closed

### LINKED ISSUE #5869: Can no close the app on full-screen mode
### Version

0.14.4

### Operating System

macOS

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

The application can not be closed when you are in full-screen mode. Here's the attached video about the issue

https://github.com/user-attachments/assets/9ca2e152-d7d9-4757-b3d3-edfd85f90358

### How to reproduce

Just go to full-screen mode and try to close the app by clicking the red button on the top left

### Expected behavior

_No response_

### Relevant log output

_No response_

mtsgrd:
Fixes #5869.

### DIFF
--- a/crates/gitbutler-tauri/src/main.rs
+++ b/crates/gitbutler-tauri/src/main.rs
@@ -246,11 +246,11 @@ fn main() {
                 .menu(menu::build)
                 .on_window_event(|window, event| match event {
                     #[cfg(target_os = "macos")]
-                    tauri::WindowEvent::CloseRequested { api, .. } => {
-                        if window.app_handle().windows().len() == 1 {
-                            tracing::debug!("Hiding all application windows and preventing exit");
-                            window.app_handle().hide().ok();
-                            api.prevent_close();
+                    tauri::WindowEvent::CloseRequested { .. } => {
+                        let app_handle = window.app_handle();
+                        if app_handle.windows().len() == 1 {
+                            app_handle.cleanup_before_exit();
+                            app_handle.exit(0);
                         }
                     }
                     tauri::WindowEvent::Destroyed => {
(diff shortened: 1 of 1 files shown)


crates/gitbutler-tauri/src/main.rs
```diff
@@ -246,11 +246,11 @@ fn main() {
...
-                        if window.app_handle().windows().len() == 1 {
-                            tracing::debug!("Hiding all application windows and preventing exit");
-                            window.app_handle().hide().ok();
-                            api.prevent_close();
+                    tauri::WindowEvent::CloseRequested { .. } => {
+                        let app_handle = window.app_handle();
+                        if app_handle.windows().len() == 1 {
+                            app_handle.cleanup_before_exit();
+                            app_handle.exit(0);
                         }
```
Caleb-T-Owens, line 254: Is an else clause containing `api.prevent_close()` not required?

mtsgrd: I mean, I don't think we're interested in preventing a window from closing? If you click the red button you'd expect it to close, and if it's the last window it seems reasonable that we shutdown the app too.

Caleb-T-Owens: My understanding of applications on macOs was that if there are two windows, it's still one instance of the app. As such, we want to prevent the killing of the applicaiton if there a window count that is not equal to 1?

mtsgrd: Right, that makes sense. I did, however, test this manually and with two windows open the you can close one without the app quitting.

Caleb-T-Owens: Cool. That sounds all good then. Thanks :+1:

Caleb-T-Owens (approved):

Issue #5826, opened 2024-12-12
Cannot click on conflicts in the first file of a commit


plichard:
### Version

0.14.3

### Operating System

Linux

### Distribution Method

AppImage (Linux)

### Describe the issue

Clicking on the first conflicted file in a commit just closes the commit details instead of opening the conflict handling ui.

https://github.com/user-attachments/assets/a4ed85be-6717-46ec-9fe0-c08e00f87813

Byron:
Thanks a lot for reporting!

I noticed this behaviour as well and thought it should show the conflict itself, or inform me how to `Resolve Conflicts`.
Conflict resolution is already possible, but only after the respective button was pressed.

plichard:
Oh! Right, this has nothing to do with it being the first file in the commit listing, for some reason the "Resolve conflicts" button is often just invisible to my brain, you mentioning it revealed it :sweat_smile: .
I forgot conflict resolution worked for me previously because I actually clicked "Resolve conflicts" first, and then I was able to select one of the files (which was not the first one at the time).

This report must have been a confusing read ...

Byron:
> Oh! Right, this has nothing to do with it being the first file in the commit listing, for some reason the "Resolve conflicts" button is often just invisible to my brain, you mentioning it revealed it 😅 .

This is super-interesting! I noticed that too, and magically, after #5722 was done, this issue disappeared for me.
Could you find an occasion where this happens in the latest stable, and then [try a nightly build](https://app.gitbutler.com/downloads)? Maybe that issue goes away.

Then what stays here is the UX issue that people right now are visually trained that something happens once they click a file. Maybe while that's not the case, these conflicted files could be presented differently. CC @PavelLaptev .

PavelLaptev:
@Byron I agree with you and @plichard. We should be able to at least preview the file when we click on it. I'll create a PR.

For the next iteration, we can also think about additional ways to go to the conflict resolution page.

Caleb-T-Owens:
We currently don't load that file data. It might take some work to get a file display working.

We'd need to generate the files with conflicts written in them for display, and to have consider how to display the cases where you've modified and they've deleted

PavelLaptev:
@Caleb-T-Owens got it. I thought it's only a frontend thing, just a svelte condition that prevent to click on a conflicted file row.

Caleb-T-Owens:
[...]

PR #5884, opened 2025-01-02
Create re-usable file service for reading git blobs

mtsgrd:
- decouples reading from workspace from reading from commit
- prevents `FileDiff` component from reading file for every diff


mtsgrd:
@Caleb-T-Owens could this be helpful in resolving #5826? It would currently return the full file ofc.

Caleb-T-Owens:
> @Caleb-T-Owens could this be helpful in resolving #5826? It would currently return the full file ofc.

Yes, it could be used in the cases where the two sides have both modified the same file. It would need a bit more consideration as to how we display files that have been modified & deleted though.

Caleb-T-Owens (approved):

estib-vega (approved):

PR #5889, opened 2025-01-04
use `gix-status` for `create_wd_tree()`

Byron:
Follow-up to #4912.

### Tasks

* [x] see if additional tests can or should be added
* [x] use new `gix-status` for `create_wd_tree()`
* [x] intent-to-add test
* [x] ignored files test
* [x] submodule test - maybe we can deal with it a little?
* [x] skip huge files
* [x] figure out occasional test failure - racy git probably, but why
* [x] capture performance numbers - 35% more speed
* [x] use `gix` from `main` once https://github.com/GitoxideLabs/gitoxide/pull/1746 is merged.
* [x] wait for improvements made through https://github.com/starship/starship/pull/6476

### Notes for the Reviewer

* When a type-change is detected, it currently re-reads the entire file nonetheless even though it could re-use what's already in the index. It seems acceptable performance wise and can keep things simpler.
* The `gitoxide` version uses the typesystem to enforce exhaustiveness. This is more verbose, but really helps to cover all the bases.
* The previous implementation will not manage to deal with directories that become files or vice-versa due to the behaviour of the git2 tree editor, so it's probably advisable to prefer using `gix::object::tree::Editor` when making arbitrary edits to trees. This is rare though, so probably there aren't anymore bugs like these.
* I implemented submodule support as there is no reason not to include it.
* Sparse indices/repositories aren't supported and it will fail loudly then. This is fine for now as GB also doesn't support these (nor does Git2). It shouldn't be a big deal to make this possible either, it's no my radar.

https://github.com/user-attachments/assets/bf73befa-0922-4bb9-b5d4-018675bdd371

### Performance

Measured on `gitlab` repository where `gitbutler-cli --trace branch apply -b main` was executed. `create_wd_tree()` is called then. Unfortunately, there is no way to call it individually unless one creates a new subcommand for the CLI. As the performance improvement is visible enough, I didn't spend the time on that.

#### With `git2`

*(Best out of two runs)*

```
DEBUG    ┕━ apply_branch [ 4.72s | 1.28% / 94.49% ] stack_id: 4a71e26e-e9c5-4eb3-a46c-92de03ba79a5
DEBUG       ┝━ create_wd_tree [ 439ms | 8.64% / 8.78% ]
```

#### With `gix`

*(Best out of two runs)*

```
DEBUG    ┕━ apply_branch [ 4.70s | 1.35% / 94.42% ] stack_id: 85aa5a75-133f-4596-9a88-e50dad4a4134
DEBUG       ┝━ create_wd_tree [ 317ms | 6.22% / 6.37% ]
```

**`gix` is ~35% faster.**

<details><summary>Research</summary>

### Ordering Issues

The order of entries isn't defined, which makes it hard to decide what the final state actually is, or at least needs extra logic.

Something that doesn't work in our favor is:

```
---- create_wd_tree::tracked_file_becomes_directory_in_worktree stdout ----
[/Users/byron/dev/github.com/GitoxideLabs/gitoxide/gix/src/status/iter/mod.rs:277:9] &item = IndexWorktree(
    DirectoryContents {
        entry: Entry {
            rela_path: "soon-directory/file",
            status: Untracked,
            property: None,
            disk_kind: Some(
                File,
            ),
            index_kind: None,
            pathspec_match: Some(
                Always,
            ),
        },
        collapsed_directory_status: None,
    },
)
[/Users/byron/dev/github.com/GitoxideLabs/gitoxide/gix/src/status/iter/mod.rs:277:9] &item = IndexWorktree(
    Modification {
        entry: Entry {
            stat: Stat {
                mtime: Time {
                    secs: 1736081071,
                    nsecs: 246777717,
                },
                ctime: Time {
                    secs: 1736081071,
                    nsecs: 246777717,
                },
                dev: 0,
                ino: 29004082,
                uid: 501,
                gid: 20,
                size: 37,
            },
            id: Sha1(5781af89edf37bcada102e86c4aa48e129a4ece4),
            flags: Flags(
                0x0,
            ),
            mode: Mode(
                FILE,
            ),
            path: 0..14,
        },
        entry_index: 0,
        rela_path: "soon-directory",
        status: Change(
            Removed,
        ),
    },
)
```

By default, the order is reversed which works out, but like this the second even will remove the changes of the first.
Solved by applying untracked files last.
</details>

### DIFF
--- a/crates/gitbutler-project/src/lib.rs
+++ b/crates/gitbutler-project/src/lib.rs
@@ -18,3 +18,6 @@ pub fn configure_git2() {
     // These settings are only changed from `main` of applications.
     git2::opts::strict_object_creation(false);
 }
+
+/// The maximum size of files to automatically start tracking, i.e. untracked files we pick up for tree-creation.
+pub const AUTO_TRACK_LIMIT_BYTES: u64 = 32 * 1024 * 1024;
--- a/crates/gitbutler-project/src/lib.rs
+++ b/crates/gitbutler-project/src/lib.rs
@@ -20,4 +20,5 @@ pub fn configure_git2() {
 }
 
 /// The maximum size of files to automatically start tracking, i.e. untracked files we pick up for tree-creation.
-pub const AUTO_TRACK_LIMIT_BYTES: u64 = 32 * 1024 * 1024;
+/// **Inactive for now** while it's hard to tell if it's safe *not* to pick up everything.
+pub const AUTO_TRACK_LIMIT_BYTES: u64 = 0;
--- a/crates/gitbutler-repo/src/repository_ext.rs
+++ b/crates/gitbutler-repo/src/repository_ext.rs
@@ -1,7 +1,7 @@
 use crate::Config;
 use crate::SignaturePurpose;
 use anyhow::{anyhow, bail, Context, Result};
-use bstr::BString;
+use bstr::{BStr, BString};
 use git2::Tree;
 use gitbutler_commit::commit_headers::CommitHeadersV2;
 use gitbutler_config::git::{GbConfig, GitConfig};
@@ -12,6 +12,7 @@ use gitbutler_oxidize::{
 use gitbutler_reference::{Refname, RemoteRefname};
 use gix::filter::plumbing::pipeline::convert::ToGitOutcome;
 use gix::objs::WriteTo;
+use gix::status::index_worktree;
 use std::collections::HashSet;
 #[cfg(unix)]
 use std::os::unix::fs::PermissionsExt;
@@ -42,8 +43,15 @@ pub trait RepositoryExt {
     fn sign_buffer(&self, buffer: &[u8]) -> Result<BString>;
     fn checkout_tree_builder<'a>(&'a self, tree: &'a git2::Tree<'a>) -> CheckoutTreeBuidler<'a>;
     fn maybe_find_branch_by_refname(&self, name: &Refname) -> Result<Option<git2::Branch>>;
-    /// Based on the index, add all data similar to `git add .` and create a tree from it, which is returned.
-    fn create_wd_tree(&self) -> Result<Tree>;
+    /// Add all untracked and modified files in the worktree to
+    /// the object database, and create a tree from it.
+    ///
+    /// Use `untracked_limit_in_bytes` to control the maximum file size for untracked files
+    /// before we stop tracking them automatically. Set it to 0 to disable the limit.
+    ///
+    /// It should also be noted that this will fail if run on an empty branch
+    /// or if the HEAD branch has no commits.
+    fn create_wd_tree(&self, untracked_limit_in_bytes: u64) -> Result<Tree>;
 
     /// Returns the `gitbutler/workspace` branch if the head currently points to it, or fail otherwise.
     /// Use it before any modification to the repository, or extra defensively each time the
@@ -105,15 +113,8 @@ impl RepositoryExt for git2::Repository {
         Ok(branch)
     }
 
-    /// Add all untracked and modified files in the worktree to
-    /// the object database, and create a tree from it.
-    ///
-    /// Note that right now, it doesn't skip big files.
-    ///
-    /// It should also be noted that this will fail if run on an empty branch
-    /// or if the HEAD branch has no commits.
-    #[instrument(level = tracing::Level::DEBUG, skip(self), err(Debug))]
-    fn create_wd_tree(&self) -> Result<Tree> {
+    #[instrument(level = tracing::Level::DEBUG, skip(self, untracked_limit_in_bytes), err(Debug))]
+    fn create_wd_tree(&self, untracked_limit_in_bytes: u64) -> Result<Tree> {
         use bstr::ByteSlice;
         use gix::dir::walk::EmissionMode;
         use gix::status;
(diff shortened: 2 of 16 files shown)


Caleb-T-Owens:
@Byron should create_wd_tree be a gix built-in? To me it seems like a very common operation

Byron:
I thought about it but concluded that it's not possible at this time. `gitoxide` has to facilitate playing 'by the rules' which means no commit is made without first going through the `.git/index`, which also means that trees will only be created from the index from where they are usually committed.

Further, there is no functionality in Git that I am aware of that would automatically pick up all untracked files.

@mtsgrd has picked up the task of creating commits after building a `.git/index` to assure pre-commit hooks can work which might as well lead to changes here. This would then mean we could go back to something that is more akin to `repo.index().add("*")` (again), which `gitoxide` definitely should support.

Caleb-T-Owens:
Is it safe to exclude large files from the produced tree?

In edit mode where we write out the `current_wd_tree` as the edited version of a given commit. Inferring from the tests, if there was a large file either in the previous history or even introduced, won't it now get deleted or not added?

Byron:
Thanks for reminding me! It's most definitely not safe while GitButler still performs hard resets of the working tree, something I believe it definitely has to stop doing. But until we are there, it should pick up everything to protect it, no matter how large it may be.

It's possible that there are places where a limit can be imposed, but for now I have deactivated the limit everywhere.

crates/gitbutler-repo/src/repository_ext.rs
```diff
@@ -126,86 +132,196 @@ impl RepositoryExt for git2::Repository {
...
+                // This is probably a type-change to something we can't track. Instead of keeping
+                // what's in `HEAD^{tree}` we remove the entry.
+                head_tree_editor.remove(rela_path)?;
+                return Ok(true);
+            };
+
+            head_tree_editor.upsert(rela_path, kind, id)?;
+            Ok(true)
+        };
+        let mut head_tree_editor = repo.edit_tree(repo.head_tree_id()?)?;
```
krlvi, line 188: This tree editor is a really neat abstraction from the gix side!

crates/gitbutler-project/src/lib.rs
```diff
@@ -18,3 +18,7 @@ pub fn configure_git2() {
     // These settings are only changed from `main` of applications.
     git2::opts::strict_object_creation(false);
 }
+
+/// The maximum size of files to automatically start tracking, i.e. untracked files we pick up for tree-creation.
+/// **Inactive for now** while it's hard to tell if it's safe *not* to pick up everything.
+pub const AUTO_TRACK_LIMIT_BYTES: u64 = 0;
```
krlvi, line 24: There are cases where not disregarding huge files will lock up the app. For instance the case where somebody has a databse in the repo root - not ignored, but never intend to stage and commit that. Or an accidentally moving a large file in the repo.
Perhaps we can teach the edit mode to perform hard resets excluding these files? In terms of frequency of occurrence, it is much more likely to hit the case of an accidentally trying to add a huge file to a tree than to reset it via edit mode...

crates/gitbutler-repo/src/repository_ext.rs
Caleb-T-Owens: libgit2 tree tools:
ALRARB_01_adler-2533172538

gix tree tools:
ltz-05a_z1-321783570

---

(this is an attempt at a "meme" and is intended to be humorous. Both libraries are great)

Byron: My brain had a serious issue to align what it saw with the side I ought to be seeing, but now I have recovered 😁.

> This tree editor is a really neat abstraction from the gix side!

Thanks! It's based on `git2` and pretty much the same API for the editor itself - the way it's instantiated might be a bit more natural though. As a major difference, besides being faster, it will also just do as it's told without special rules that prevent you from doing things (like turning a directory into a file or vice-versa is forbidden in `git2` for some reason).

Regarding the pictures, I think it's perfectly true and very obvious when looking at the before-after of the whole 'status' loop. The `gix` detail is excruciating, but it did help to make the whole operation very well defined.

There is huge value in the way `git2` represents state as well and I hope to find ways to support both levels of abstraction one day. `gix` easy mode :D?

crates/gitbutler-project/src/lib.rs
Byron: I couldn't agree more.

For this PR, I'd hope that maintaining the current behaviour of `create_wd_tree()` picking up everything independently of size will be acceptable, as it at least won't make anything worse.

Then, for the future, I'd hope GB can be made to…

* …use the `.git/index` when creating commits for the user (but won't use the `.git/index` when creating internal commits, like for the oplog as it's more efficient)
* …never do hard resets
* …to leave untracked files alone just like Git does
     - …while barking if there is some clash with an untracked file

It's the "good gitizen" principle so basic expectations that people have built up towards Git will be met.
Maybe there are better ways to do that even, there should be some room for innovation, too.

Caleb-T-Owens: I'm surprised that the creation of the wd tree is slow when large files are present.

In my mind, the slowness was with both generating and displaying diffs of large files. As such, we only needed to put the limits in there.

> never do hard resets

Is that value compatible with edit mode?

> to leave untracked files alone just like Git does

Could you elaborate on what cases you're concerned about here?

Byron: > I'm surprised that the creation of the wd tree is slow when large files are present.

Independently of the UI, creating objects from content of disk will always be limited to 20-30MB/s per core due to ZLIB. This is horribly inefficient especially when said large file is a database that changes all the time while compressed by nature.

> Is that value compatible with edit mode?

Git has one mechanism to deal with this, being `git stash --include-untracked`. But even that I think should only be done once it's clear the untracked files are in the way. The content of the index would need to be stashed anyway to avoid loosing it.

> Could you elaborate on what cases you're concerned about here?

It's the expectation that a Git client behaves like Git, and Git won't touch untracked files unless it's explicitly told to (or sometime less explicitly, but it's opt-in always).`jj` has the advantage of being its own thing so it can redefine what's 'normal' to great effect. To my mind, that's something that GitButler can't do, at least not by default, merely to align with the expectations brought towards it.
Going beyond the 'normal' for Git, of course, is another form on innovation which should be explored to push the envelope, but that can't ever come at the cost of the safety of the user's data.

krlvi: Okay I see. I also mixed up the code path that is being changed here with this one https://github.com/gitbutlerapp/gitbutler/blob/1e16896adbb4743112bcd6de0ad2bf59bb3e533c/crates/gitbutler-diff/src/diff.rs#L161

krlvi (approved):

PR #5890, opened 2025-01-05
Make it easier to pass diffs to list vbranch function

mtsgrd:
- cached function should not accept option
- pass reference to diffs rather than consuming them

---
- <kbd>&nbsp;3&nbsp;</kbd> #5940 
- <kbd>&nbsp;2&nbsp;</kbd> #5894 
- <kbd>&nbsp;1&nbsp;</kbd> #5890 👈

### DIFF
--- a/crates/gitbutler-branch-actions/src/actions.rs
+++ b/crates/gitbutler-branch-actions/src/actions.rs
@@ -79,7 +79,7 @@ pub fn list_virtual_branches(project: &Project) -> Result<StackListResult> {
 
 pub fn list_virtual_branches_cached(
     project: &Project,
-    worktree_changes: Option<DiffByPathMap>,
+    worktree_changes: DiffByPathMap,
 ) -> Result<StackListResult> {
     let ctx = open_with_verify(project)?;
 
@@ -89,7 +89,7 @@ pub fn list_virtual_branches_cached(
     vbranch::list_virtual_branches_cached(
         &ctx,
         project.exclusive_worktree_access().write_permission(),
-        worktree_changes,
+        &worktree_changes,
     )
     .map_err(Into::into)
 }
--- a/crates/gitbutler-branch-actions/src/virtual.rs
+++ b/crates/gitbutler-branch-actions/src/virtual.rs
@@ -309,7 +309,8 @@ pub fn list_virtual_branches(
     ctx: &CommandContext,
     perm: &mut WorktreeWritePermission,
 ) -> Result<StackListResult> {
-    list_virtual_branches_cached(ctx, perm, None)
+    let diffs = gitbutler_diff::workdir(ctx.repo(), get_workspace_head(ctx)?)?;
+    list_virtual_branches_cached(ctx, perm, &diffs)
 }
 
 /// `worktree_changes` are all changed files against the current `HEAD^{tree}` and index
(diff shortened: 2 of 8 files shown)


crates/gitbutler-branch-actions/src/virtual.rs
```diff
@@ -860,6 +848,104 @@ pub fn commit(
...
+    }
+    Ok(())
+}
+
+pub(crate) fn unstage_all(ctx: &CommandContext) -> Result<()> {
+    let repo = ctx.repo();
+    // Get the HEAD commit (current commit)
+    let head_commit = repo.head()?.peel_to_commit()?;
+    // Reset the index to match the HEAD commit
+    repo.reset(head_commit.as_object(), ResetType::Mixed, None)?;
```
Byron: At this time, GitButler doesn't differentiate between staged changes and unstaged ones, so resetting the index get get to the pre-staging state will be visually indistinguishable.

However, I think it would be more correct and more future-proof to go back to a previous state by writing the previous state held in memory, something like this.

```Rust
fn restore_index_on_error(repo, doit() -> Result) -> Result {
    let index = repo.index();
    let res = doit();
    if res.is_err() {
        index.write();
    }
    res
}
```

That way, changes to the index made by the user will be retained when restoring.

crates/gitbutler-branch-actions/src/virtual.rs
```diff
@@ -860,6 +848,104 @@ pub fn commit(
...
+                {
+                    return true;
+                }
+            }
+            false
+        })
+    });
+
+    let diff = diff_workdir_to_index(repo, path)?;
+    repo.apply(&diff, ApplyLocation::Index, Some(&mut apply_opts))?;
```
Byron: `gitoxide` still can't apply patches, but once it can this can be written so that it will write the `.git/index` only once.

Byron (commented):
It's great to see this happen, thank you!

(*I definitely learned something about `git2` again - it's very powerful when dealing with deltas in general and it's another area where `gitoxide` has to catch up.*)

krlvi:
Good work Mattias! Since this is being overhauled, it would actually be really nice to separate the run commit hooks in it's own API command. 

Separating it from the commit command will allow us to make both commands more robust & easy to test. And this way things like the `unstage_all` will only be invoked as a result of error from the hook execution.

 But what is even better is that it allows us to make a better UX - for instance, we can allow users to run a commit hook without actually committing. And even when the hook is executed just before committing, if the operation is slow, we can make it clear in the interface that it is the commit hook that is taking time.

crates/gitbutler-branch-actions/src/virtual.rs
mtsgrd: Ah, nice! Let me do it that way.

krlvi:
From our conversation IRL - lets explore using an auxiliary index (not the default one)

crates/gitbutler-branch-actions/src/actions.rs
```diff
@@ -79,7 +79,7 @@ pub fn list_virtual_branches(project: &Project) -> Result<StackListResult> {
 
 pub fn list_virtual_branches_cached(
     project: &Project,
-    worktree_changes: Option<DiffByPathMap>,
+    diffs_by_path: DiffByPathMap,
```
krlvi: The original name was more descriptive - the new name just repeats the type name

mtsgrd:
>From our conversation IRL - lets explore using an auxiliary index (not the default one)

I'll put that at the top of this stack!

krlvi:
Since this is now part of a stack, it is not clear what the purpose of this change here is. Seems trivial and I am sure I am missing something that is in the rest of the stack

Caleb-T-Owens (approved):

krlvi (approved):

PR #5893, opened 2025-01-06
Improved integration

Caleb-T-Owens:
---
- <kbd>&nbsp;2&nbsp;</kbd> #5893 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #5881

### DIFF
--- a/apps/desktop/src/lib/backend/projects.ts
+++ b/apps/desktop/src/lib/backend/projects.ts
@@ -28,6 +28,7 @@ export class Project {
 	use_diff_context: boolean | undefined;
 	snapshot_lines_threshold!: number | undefined;
 	use_new_branch_integration_algorithm: boolean | undefined;
+	use_new_integration_check!: boolean;
 	// Produced just for the frontend to determine if the project is open in any window.
 	is_open!: boolean;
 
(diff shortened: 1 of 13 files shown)


Caleb-T-Owens:
Old time:
13.4s
New time:
107ms

I don't know how.

Caleb-T-Owens:
This makes working on gitlab downright usable

Byron:
> Old time: 13.4s New time: 107ms
> 
> I don't know how.

That's crazy awesome 🙌! It's probably doing much less work than before to do it, which is great if one can get away with it. Maybe previously, it did way too much work as well.

107ms on a repo of GitLab size is probably no more than a merge or two.

Caleb-T-Owens:
fuck... it was too good to be true.

this is indeed slower, and is slower by a fair bit.

Caleb-T-Owens:
@Byron / @krlvi Shall we still merge this despite the performance issues as it's behind a feature flag anyway.

Byron:
I think there is serious value in being able to detect if commits are integrated despite squashes. Of course, I am still concerned about the high runtime, which was high before and now is even higher.

Caleb-T-Owens:
> Of course, I am still concerned about the high runtime, which was high before and now is even higher.

Indeed. I wouldn't consider removing the feature flag that it is currently under without some serious performance improvements.

Perhaps however it's just unwanted overhead with the fork in the logic. I just don't really want it to go stale

Caleb-T-Owens:
@krlvi wdyt?

Issue #2615, opened 2024-02-13
Git hooks are not executed (on the correct .git/index)


nathanael-ruf:
For me git pre-commit hooks aren't executed.
Judging from https://github.com/gitbutlerapp/gitbutler/issues/2551 it seems like they should be, just without being able to view the output?

nolith:
Same for me, I was testing a new command I added in lefthook and I noticed that gitbutler happily committed something that was supposed to make the pre-commit check fail

nolith:
I'm using https://github.com/evilmartians/lefthook and it installs a script like this one.

```
 cat .git/hooks/pre-commit
#!/bin/sh

if [ "$LEFTHOOK" = "0" ]; then
  exit 0
fi

call_lefthook()
{
  dir="$(git rev-parse --show-toplevel)"
  osArch=$(uname | tr '[:upper:]' '[:lower:]')
  cpuArch=$(uname -m | sed 's/aarch64/arm64/')

  if lefthook -h >/dev/null 2>&1
  then
    lefthook "$@"
  elif test -f "$dir/node_modules/lefthook/bin/index.js"
  then
    "$dir/node_modules/lefthook/bin/index.js" "$@"
  elif test -f "$dir/node_modules/@evilmartians/lefthook/bin/lefthook_${osArch}_${cpuArch}/lefthook"
  then
    "$dir/node_modules/@evilmartians/lefthook/bin/lefthook_${osArch}_${cpuArch}/lefthook" "$@"
  elif test -f "$dir/node_modules/@evilmartians/lefthook-installer/bin/lefthook_${osArch}_${cpuArch}/lefthook"
  then
    "$dir/node_modules/@evilmartians/lefthook-installer/bin/lefthook_${osArch}_${cpuArch}/lefthook" "$@"
  elif bundle exec lefthook -h >/dev/null 2>&1
  then
    bundle exec lefthook "$@"
  elif yarn lefthook -h >/dev/null 2>&1
  then
    yarn lefthook "$@"
  elif pnpm lefthook -h >/dev/null 2>&1
  then
    pnpm lefthook "$@"
  elif command -v npx >/dev/null 2>&1
  then
    npx lefthook "$@"
  elif swift package plugin lefthook >/dev/null 2>&1
  then
    swift package --disable-sandbox plugin lefthook "$@"
  else
    echo "Can't find lefthook in PATH"
  fi
}

call_lefthook run "pre-commit" "$@"
```

[...]

PR #5895, opened 2025-01-06
Rewrite git hooks integration

mtsgrd:
Iiuc, tools used in pre-commit hooks often rely on files being staged to e.g. only validate hunks being committed. #5735 #2615

- the typescript side drives the hook runs
- tests for hooks, including staging

*edit: this pr has dramatically changed since it was created.

---
- <kbd>&nbsp;3&nbsp;</kbd> #5895 👈 
- <kbd>&nbsp;2&nbsp;</kbd> #5894 
- <kbd>&nbsp;1&nbsp;</kbd> #5890


crates/gitbutler-diff/src/diff.rs
```diff
@@ -16,6 +16,8 @@ pub type DiffByPathMap = HashMap<PathBuf, FileDiff>;
 pub enum ChangeType {
     /// Entry does not exist in old version
     Added,
+    /// Entry is untracked item in workdir
+    Untracked,
```
mtsgrd, line 21: @estib-vega this doesn't break any of the tests for hunk locking, but can you have a look and let me know if it feels likely we need to change/add/amend anything there?

crates/gitbutler-repo/src/hooks.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+pub fn pre_commit(
+    ctx: &CommandContext,
+    selected_hunks: &Vec<(PathBuf, Vec<GitHunk>)>,
+) -> Result<HookResult> {
+    let mut index = ctx.repo().index()?;
+    staging::stage(ctx, selected_hunks)?;
+    let result = match git2_hooks::hooks_pre_commit(ctx.repo(), Some(&["../.husky"]))? {
```
mtsgrd: @krlvi @schacon do you guys know why we originally added husky here, and if we want to keep it as is? It feels a bit ad-hoc in its current form.

crates/gitbutler-repo/src/staging.rs
```diff
@@ -0,0 +1,102 @@
...
+        .context_lines(3)
+        .pathspec(path);
+    Ok(repo.diff_index_to_workdir(Some(&index), Some(&mut diff_opts))?)
+}
+
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize, Default)]
+pub struct HunkId {
+    pub start: u32,
+    pub end: u32,
+}
```
mtsgrd: @krlvi could you help me think this through. Originally we relied on `OwnershipClaim`, but it's located in `gitbutler-stack`. We could use `Hunk` from `gitbutler-diff`, but unsure if we want to involve the hunk hash in staging.

crates/gitbutler-repo/src/staging.rs
```diff
@@ -0,0 +1,102 @@
...
+                                    claimed_hunk.start == diff_hunk.new_start
+                                        && claimed_hunk.end
+                                            == diff_hunk.new_start + diff_hunk.new_lines
+                                })
+                                .cloned()
+                        })
+                        .collect(),
+                ))
+            } else {
+                Err(anyhow!("Ownership claim not found in workspace diff"))
```
mtsgrd: Need to update this error message.

crates/gitbutler-tauri/src/repo.rs
```diff
@@ -92,4 +96,61 @@ pub mod commands {
...
+                .claims
+                .iter()
+                .map(|claim| {
+                    (
+                        claim.file_path.clone(),
+                        claim.hunks.iter().map(HunkId::from).collect(),
+                    )
+                })
+                .collect(),
+        )?;
```
mtsgrd: This feels a bit awkward.. not sure what we should do. We use ownership claims for allocating hunks to branches, but also as payload for indicating what hunks to commit.. Here we want to indicate what hunks to stage before the commit hook, but `gitbutler-repo` can atm not depend on `BranchOwnershipClaims`.

crates/gitbutler-repo/src/hooks.rs
```diff
@@ -0,0 +1,86 @@
...
+
+#[derive(Serialize, PartialEq, Debug, Clone)]
+pub struct ErrorData {
+    pub error: String,
+}
+
+/// Hook result indicating either success or failure.
+#[derive(Serialize, PartialEq, Debug, Clone)]
+#[serde(tag = "status", rename_all = "lowercase")]
+pub enum HookResult {
```
krlvi, line 25: What is the difference between Message and Success here? Or Failure and NotFound? 
Can we simplify this enum?

crates/gitbutler-repo/src/hooks.rs
krlvi: According to the [docs](https://git-scm.com/docs/githooks#_pre_commit), it seems like the path is:

```
By default the hooks directory is $GIT_DIR/hooks, but that can be changed via the core.hooksPath configuration variable (see [git-config[1]](https://git-scm.com/docs/git-config)).
```

krlvi: You can use gix to read `core.hooksPath` and if it's not set, seems like you can try `$GIT_DIR/hooks`. And if not set, you can finally fallback to the husky path

crates/gitbutler-repo/src/staging.rs
```diff
@@ -0,0 +1,102 @@
...
+
+use anyhow::{anyhow, Result};
+use git2::{ApplyLocation, ApplyOptions, Repository};
+use gitbutler_command_context::CommandContext;
+use gitbutler_diff::{ChangeType, FileDiff, GitHunk, Hunk};
+use serde::{Deserialize, Serialize};
+
+fn stage_hunk(ctx: &CommandContext, path: &PathBuf, hunks: &Vec<GitHunk>) -> Result<()> {
+    let repo = ctx.repo();
+    let mut index = repo.index()?;
```
krlvi: You probably dont wanna do this and the the index.write inside the loop - instead, pass a reference/borrow to index, and do the index.write at the very end

crates/gitbutler-repo/src/hooks.rs
mtsgrd: - `NotFound` means no hook is configured
- `Failure` means hook ran, but failed
- `Message` is special for `commit-msg`, if it modifies message
- `Success` means what you think it means

I wouldn't really want to simplify, since this setup makes it easier to conditionally display small toasts saying e.g. "Pre-commit hook successful", or "Hook modified your commit message". What do you think?

crates/gitbutler-diff/src/diff.rs
estib-vega: @mtsgrd It might, but the change should be relatively simple.
I'll check with you to understand exactly what this does

crates/gitbutler-repo/src/hooks.rs
krlvi: How about renaming `NotFound` to `NotConfigured`.

Another thing that is kinda weird is that Message is part of the type but it is only relevant to the `message` method. You should probably have a separate return struct for that, otherwise it is confusing

crates/gitbutler-repo/src/staging.rs
krlvi: What is your rationale against using `Hunk` from `gitbutler-diff`? Creating this additional one here may create extra work for us down the line. Since you are doing the From conversion anyways, there is a dependency already

crates/gitbutler-repo/src/staging.rs
krlvi: you can also just write `bail!("asdf")`

crates/gitbutler-repo/tests/hooks.rs
```diff
@@ -0,0 +1,163 @@
...
+    };
+
+    use gitbutler_repo::staging::{filter_diff_by_hunk_ids, HunkId};
+    use gitbutler_testsupport::{Case, Suite};
+    use hooks::{ErrorData, HookResult, MessageData};
+
+    use super::*;
+
+    #[test]
+    fn pre_commit_hook_success() -> anyhow::Result<()> {
```
krlvi: Good. Can we also add a test case asserting that only the "selected" files are involved in the execution?

crates/gitbutler-tauri/src/repo.rs
```diff
@@ -92,4 +96,61 @@ pub mod commands {
...
+            ownership
+                .claims
+                .iter()
+                .map(|claim| {
+                    (
+                        claim.file_path.clone(),
+                        claim.hunks.iter().map(HunkId::from).collect(),
+                    )
+                })
+                .collect(),
```
krlvi: You can make a function in gitbutler-branch-actions so that you can test this and also to keep the tauri crate lean

crates/gitbutler-repo/src/hooks.rs
```diff
@@ -0,0 +1,86 @@
...
+    let result = match git2_hooks::hooks_pre_commit(ctx.repo(), Some(&["../.husky"]))? {
+        H::Ok { hook: _ } => Ok(HookResult::Success),
+        H::NoHookFound => Ok(HookResult::NotConfigured),
+        H::RunNotSuccessful { stdout, stderr, .. } => {
+            let error = join_output(stdout, stderr);
+            Ok(HookResult::Failure(ErrorData { error }))
+        }
+    };
+    // Reset to original index to unstage files.
+    index.write()?;
```
krlvi: If the code in `git2_hooks::hooks_pre_commit` panics, this function will not have the opportunity to invoke  `index.write()` and reset the index.
To combat this, it seems like you can implement the [Drop](https://doc.rust-lang.org/book/ch15-03-drop.html) trait on something.

I can think of 3 ways to achieve this here:
1. Create a new struct `Guard` for which you implement drop and have it constructed at the beginning of the call and then let it go out of scope at the end.
2. Implement `pre_commit` as a method on a Struct that implements drop
3. Perhaps use something like the crate [scopeguard](https://docs.rs/scopeguard/latest/scopeguard/) 

I have personally not done any of the 3 things outlined. Maybe @Byron has some good advice for how to approach this

crates/gitbutler-repo/tests/hooks.rs
```diff
@@ -0,0 +1,163 @@
...
+# do nothing
+";
+        git2_hooks::create_hook(ctx.repo(), git2_hooks::HOOK_PRE_COMMIT, hook);
+        assert_eq!(
+            hooks::pre_commit(ctx, &selected_hunks)?,
+            HookResult::Success
+        );
+        Ok(())
+    }
+
```
krlvi: Can we have an assertion to an existing test or a new test ensuring that the index is brought back to it's original state after `hooks::pre_commit` has returned in both the success and failure situations?

crates/gitbutler-repo/src/staging.rs
mtsgrd: We can use that one, I just don't like that it has the extra hash.. so in tests about staging you have to construct something like `Hunk { start: (), end: (), hash: None }`. Would you prefer I switch to that?

crates/gitbutler-repo/tests/hooks.rs
mtsgrd: It already does that in the reject case. It's first run with an empty selection and passes, then with the selection and fails.

crates/gitbutler-repo/src/staging.rs
mtsgrd: I've switched it over to `Hunk` now.

crates/gitbutler-repo/src/hooks.rs
mtsgrd: I've renamed it to `NotConfigured`.

crates/gitbutler-tauri/src/repo.rs
mtsgrd: Done!

crates/gitbutler-repo/src/hooks.rs
mtsgrd: I've done this using scopeguard now, thanks.

crates/gitbutler-repo/tests/hooks.rs
mtsgrd: Done!

crates/gitbutler-repo/src/hooks.rs
mtsgrd: And now there is a separate `MessageHookResult`.

krlvi (approved):

mtsgrd:
Replaced by #5940 since I incorrectly merged this into the parent branch.

PR #5904, opened 2025-01-08
Potential fix: Correctly shrink vec into a box

estib-vega:
Shrink the vec into a box correctly, so that it can be serialized and propagated correctly.

### DIFF
--- a/crates/gitbutler-branch-actions/src/hunk.rs
+++ b/crates/gitbutler-branch-actions/src/hunk.rs
@@ -53,7 +53,7 @@ impl VirtualBranchHunk {
         hunk: GitHunk,
         hash: Digest,
         mtimes: &mut MTimeCache,
-        locked_to: &[HunkLock],
+        locked_to: Box<[HunkLock]>,
     ) -> Self {
         // Get the unique branch ids (lock.branch_id) from hunk.locked_to that a hunk is locked to (if any)
         let branch_deps_count = locked_to.iter().map(|lock| lock.branch_id).unique().count();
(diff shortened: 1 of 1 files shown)


estib-vega:
@mtsgrd I was not able to consistently reproduce the locking message not showing up.
I did see it once, but I wasn't able to reproduce it a second time.

What I *did* find out was that the message doesn't show up because the file has the property `locked` set to `true` but has 0 `lockedIds` which are derived from the `hunks.lockedTo`.

My guess is that the reason for this inconsistent condition is that the `lockedTo` value is somehow being incorrectly set from the rust side, and this change fixes the most probable culprit.

crates/gitbutler-branch-actions/src/hunk.rs
```diff
@@ -53,7 +53,7 @@ impl VirtualBranchHunk {
         hunk: GitHunk,
         hash: Digest,
         mtimes: &mut MTimeCache,
-        locked_to: &[HunkLock],
+        locked_to: Box<[HunkLock]>,
```
Byron, line 56: There is a bit of desperation here and I'd be surprised if this affects the serialisation. It will affect the runtime though, so I'd hope this can be done without enforcing an extra clone.

estib-vega: @Byron Yeah, the justification for this is that it used to be this way (cloning it too) before implemented by someone else and my guess is that when I changed it I introduced a race condition. 

I do indeed don't know for certain whether this fixes that, though. 

Also, I'm not sure I can box the vec without cloning it.

estib-vega: I won't merge this until I'm sure this is the source of the condition.

Caleb-T-Owens:
I'm rather surprised that serialisation was being problematic with the slice. @Byron is there anything serde related that you might know of that could be causing this?

estib-vega:
Closing for now, can't repro this yet consitently

Issue #5022, opened 2024-10-02
gpg signing is not working correctly


slamp:
### Version

0.12.27

### Operating System

Mac OS X

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

Hello,

I'm trying to sign with GPG my commit. In project settings I got the error: "gpg: signing failed: Inappropriate ioctl for device"
This could be solved by doing a `export GPG_TTY=$TTY`
I have the export in my .zshrc configuration

<img width="581" alt="gpg-signing-issue" src="https://github.com/user-attachments/assets/0cb46df6-2d4e-432f-8643-297158fb57ea">

### How to reproduce

MacOS Apple M2
gpg (GnuPG) 2.4.5
libgcrypt 1.10.3

Go to 'Project settings', click 'Git', enable 'Sign commits'
Define the 'signing program' to /opt/homebrew/bin/gpg 
Click 'Test signing'

### Expected behavior

GPG signing is working

### Relevant log output

```shell
Signing is not working correctly
Failed to sign GPG:  [GNUPG:] KEY_CONSIDERED xxxxxxxx 2
[GNUPG:] BEGIN_SIGNING H10
[GNUPG:] PINENTRY_LAUNCHED 30528 curses 1.3.1 - - - - 502/20 0
gpg: signing failed: Inappropriate ioctl for device
[GNUPG:] FAILURE sign 83918950
gpg: signing failed: Inappropriate ioctl for device
```

Byron:
Thanks for reporting!

Could you try to launch GitButler from a terminal window, for instance with `/Applications/GitButler.app/Contents/MacOS/GitButler`?

If that works, it's clearly an issue with GitButler, as windowed application, not seeing the necessary environment variables, which is a known issue.

Thanks for your help.

slamp:
By launching GitButler in commandline, gpg asked me the Passphrase
![gpg-sign](https://github.com/user-attachments/assets/20c5eb92-b534-4ab4-8144-0c943b3c6b90)
and then I was able to sign a commit
![gitbutler-gpg-ok](https://github.com/user-attachments/assets/5ec0f34f-86b9-4e18-bfd6-c76a7bcdf854)

Byron:
That's great to hear!

Thanks also for showing a screenshot, as it makes clear that the password entry is bound to having a terminal. So even if everything was configured correctly even without having all the environment variables, `gpg` would still be unable to query the password if there is no underlying terminal.

[...]

PR #5910, opened 2025-01-09
signing with login shell

Byron:
Use a login shell for spawning the programs to sign a buffer.

Related to #5839 and #5022.

### Tasks

* [x] cleanup
* [x] use shell for any invocation
* [x] use login shell
* [x] additional fix: use `gix` for all Git configuration

### Notes for the Reviewer

* I did some testing regarding signing, and the UI picking up signing related configuration, but that probably didn't run all the codepaths that changed.
* Fetching git configuration (one key at a time like the UI does) is now about 25% to 40% slower, at ~120microseconds. Still within the range I think.

#### Environment variables in a release application

Launching a release application on MacOS (through the finder) yields the following environment variables.

```
  (
        "TMPDIR",
        "/var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T/",
    ),
    (
        "__CFBundleIdentifier",
        "com.apple.Terminal",
    ),
    (
        "XPC_FLAGS",
        "0x0",
    ),
    (
        "TERM",
        "xterm-256color",
    ),
    (
        "SSH_AUTH_SOCK",
        "/Users/byron/.gnupg/S.gpg-agent.ssh",
    ),
    (
        "XPC_SERVICE_NAME",
        "0",
    ),
    (
        "TERM_PROGRAM",
        "Apple_Terminal",
    ),
    (
        "TERM_PROGRAM_VERSION",
        "455",
    ),
    (
        "TERM_SESSION_ID",
        "<redacted>",
    ),
    (
        "SHELL",
        "/bin/zsh",
    ),
    (
        "HOME",
        "/Users/byron",
    ),
    (
        "LOGNAME",
        "byron",
    ),
    (
        "USER",
        "byron",
    ),
    (
        "PATH",
        "<redacted, but looks complete>",
    ),
    (
        "SHLVL",
        "1",
    ),
    (
        "PWD",
        "/Users/byron",
    ),
    (
        "OLDPWD",
        "/Users/byron",
    ),
    (
        "ZSH",
        "/Users/byron/.oh-my-zsh",
    ),
    (
        "PAGER",
        "less",
    ),
    (
        "LESS",
        "-R",
    ),
    (
        "LSCOLORS",
        "Gxfxcxdxbxegedabagacad",
    ),
    (
        "LS_COLORS",
        "di=1;36:ln=35:so=32:pi=33:ex=31:bd=34;46:cd=34;43:su=30;41:sg=30;46:tw=30;42:ow=30;43",
    ),
    (
        "HOMEBREW_PREFIX",
        "/opt/homebrew",
    ),
    (
        "HOMEBREW_CELLAR",
        "/opt/homebrew/Cellar",
    ),
    (
        "HOMEBREW_REPOSITORY",
        "/opt/homebrew",
    ),
    (
        "INFOPATH",
        "/opt/homebrew/share/info:/opt/homebrew/share/info:",
    ),
    (
        "GPG_TTY",
        "/dev/ttys005",
    ),
    (
        "EDITOR",
        "hx",
    ),
    (
        "BAT_THEME",
        "TwoDark",
    ),
    (
        "STARSHIP_SHELL",
        "zsh",
    ),
    (
        "STARSHIP_SESSION_KEY",
    ),
    (
        "HOMEBREW_NO_ANALYTICS",
        "1",
    ),
    (
        "LC_CTYPE",
        "UTF-8",
    ),
    (
        "_",
        "/Users/byron/dev/github.com/gitbutlerapp/gitbutler/target/debug/gitbutler-tauri",
    ),
    (
        "__CF_USER_TEXT_ENCODING",
        "0x1F5:0x0:0x0",
    ),
]

```

It looks like `$SHELL` can be used to figure out the login shell at least on MacOS. It's notable how many other additional variables are available, as if it was opened through a login shell already.

It really seems that a lot, but not all, of the login shell environment is picked up. It's unclear why

### DIFF
--- a/crates/gitbutler-repo/src/repository_ext.rs
+++ b/crates/gitbutler-repo/src/repository_ext.rs
@@ -1,7 +1,7 @@
 use crate::Config;
 use crate::SignaturePurpose;
 use anyhow::{anyhow, bail, Context, Result};
-use bstr::{BStr, BString};
+use bstr::{BStr, BString, ByteSlice};
 use git2::Tree;
 use gitbutler_commit::commit_headers::CommitHeadersV2;
 use gitbutler_config::git::{GbConfig, GitConfig};
@@ -13,11 +13,14 @@ use gitbutler_reference::{Refname, RemoteRefname};
 use gix::filter::plumbing::pipeline::convert::ToGitOutcome;
 use gix::objs::WriteTo;
 use gix::status::index_worktree;
+use std::borrow::Cow;
 use std::collections::HashSet;
+use std::ffi::OsStr;
 #[cfg(unix)]
 use std::os::unix::fs::PermissionsExt;
 #[cfg(windows)]
 use std::os::windows::process::CommandExt;
+use std::path::Path;
 use std::{io::Write, process::Stdio, str};
 use tracing::instrument;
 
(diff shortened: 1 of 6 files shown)


crates/gitbutler-repo/src/repository_ext.rs
```diff
@@ -575,6 +579,13 @@ impl RepositoryExt for git2::Repository {
     }
 }
 
+pub fn command_with_login_shell(shell_cmd: impl Into<OsString>) -> std::process::Command {
+    gix::command::prepare(shell_cmd)
+        .with_shell_disallow_manual_argument_splitting()
+        .with_shell_program(std::env::var_os("SHELL").unwrap_or("bash".into()))
```
Byron: This is how the hooks crate does it as well and it seems to work for hooks.

Byron (commented):
I definitely changed more than originally planned, as I also added `gix` to read Git configuration. As it fully implements Git configuration, it will pick up more 'special' configuration that users might have to adjust their signing keys based on the repository at hand. `git2` doesn't implement `hasconfig`.

Also note the `config.trusted_program()` calls which is `gitoxide`s way of assuring no configuration of untrusted users (i.e. all others) is picked up to define which program to execute automatically.

crates/gitbutler-repo/src/repository_ext.rs
Caleb-T-Owens: @Byron Do we need to the same windows bash.exe lookup that the git2_hooks does?

Caleb-T-Owens: https://docs.rs/git2-hooks/latest/src/git2_hooks/hookspath.rs.html#179

slamp (commented):
I didn't see any basic stuff that seemed wrong or incorrect to me.

krlvi (approved):

crates/gitbutler-repo/src/repository_ext.rs
```diff
@@ -394,127 +397,128 @@ impl RepositoryExt for git2::Repository {
...
-            let is_ssh = if let Ok(sign_format) = sign_format {
-                sign_format == "ssh"
-            } else {
-                false
-            };
+        let repo = gix::open(self.path())?;
+        let config = repo.config_snapshot();
+        let signing_key = config.string("user.signingkey");
+        let Some(signing_key) = signing_key else {
+            return Err(anyhow::anyhow!("No signing key found"));
```
Caleb-T-Owens: There is a `bail!` macro which is equivalent to `return Err(anyhow::anyhow!(...))`

Byron: Good catch, done!

PR #5915, opened 2025-01-10
fix: Only remove the the search param if needed

estib-vega:
The code was in a redirection loop because it was always always attempting to remove the token search param regardless of whether it had it.


estib-vega:
Bypassing review in order to quickly fix the loop

PR #5923, opened 2025-01-10
Fix downloads: Filter out invalid releases

estib-vega:
Parse and guard the response from the releases API to ensure the input is the one expected.

### DIFF
--- a/apps/web/src/lib/types/releases.ts
+++ b/apps/web/src/lib/types/releases.ts
@@ -1,3 +1,5 @@
+import { captureMessage } from '@sentry/sveltekit';
+
 export interface Build {
 	os: 'windows' | 'darwin' | 'linux';
 	arch: 'x86_64' | 'aarch64';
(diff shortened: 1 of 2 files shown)


estib-vega:
The `processBuilds` function was expecting an array of objects of type `Build`, which in the case of one build, it seemed to not completely match.

This is fixed now in that we'll probe the result out and filter out whatever doesn't match the output

apps/web/src/lib/types/releases.ts
```diff
@@ -6,12 +6,54 @@ export interface Build {
...
+		if (isRelease(item)) {
+			result.push(item);
+			continue;
+		}
+
+		// TODO: Add some metrics to this
+		console.warn('Invalid release:', item);
+	}
+	return result;
+}
```
Caleb-T-Owens: Should we be reporting this to Sentry?

apps/web/src/lib/types/releases.ts
```diff
@@ -6,12 +6,54 @@ export interface Build {
...
+		typeof (something as any).version === 'string' &&
+		(typeof (something as any).notes === 'string' || (something as any).notes === null) &&
+		typeof (something as any).sha === 'string' &&
+		typeof (something as any).channel === 'string' &&
+		typeof (something as any).build_version === 'string' &&
+		typeof (something as any).released_at === 'string' &&
+		Array.isArray((something as any).builds) &&
+		(something as any).builds.every(isReleaseBuild)
+	);
+}
```
Caleb-T-Owens, line 46: Is it time for zod?

Caleb-T-Owens (approved):
Happy for this to be shipped and to consider zods and sentrees later

apps/web/src/lib/types/releases.ts
estib-vega: IMO yeah, do we already do that anywhere in the app? I'll look for an example of that

apps/web/src/lib/types/releases.ts
estib-vega: I would love that

PR #5935, opened 2025-01-13
fix: use dbus secret store method for linux keyring crate

ndom91:
## 🧢 Changes

- Use `sync-secret-service` linux keyring feature to use dbus to communicate with linux system secret store and therefore allow access via flatpak, for example.


ndom91:
Unfortunately neither `sync-secret-service` nor the mixed `linux-native-sync-persistent` seemed to work on my Ubuntu 24.04 test machine in the `.deb` or the `AppImages`. 

Therefore, I'll have to patch the flatpak to use the `sync-secret-service` only for the Flatpak and leave this to the `linux-native` (`keyutils`) only.

Byron:
That's great! At least there now is a path forward to make secrets work from a Flatpak.

PR #5943, opened 2025-01-14
Use a single textbox for commit message

mtsgrd:
- allows for `cmd + a` select all
- simpler component
- works better with commit message hook

image

image


PavelLaptev:
Sorry @mtsgrd we didn't discuss this first. Let's do it tomorrow.

PR #5912, opened 2025-01-09
V3 Uncommitted Files, CLI and Filewatching

Byron:
This PR introduces a new crate for the code that will drive the V3 of the UI design, with the goal of sketching out the API that shows how to deal with **uncommited files**.

This sketch should also include an idea how to listen to changes to the filesystem to produce events that help maintain the current state.

### Tasks

* [ ] initial state of uncommitted files
* [ ] a new CLI to show uncommitted files
* [ ] a watcher file filesystem changes
    - see if fsMonitor can be used
* [ ] quickly add a `-w` flag so the CLI can keep showing changes.

PR #5945, opened 2025-01-14
V3 Uncommitted Files, CLI and Filewatching (part 2)

Byron:
This PR introduces a new crate for the code that will drive the V3 of the UI design, with the goal of sketching out the API that shows how to deal with **uncommited files**.

This sketch should also include an idea how to listen to changes to the filesystem to produce events that help maintain the current state.

Follow-up of #5912 .

### Tasks

* [x] initial state of uncommitted files
* [x] a new CLI to show uncommitted files
* [x] tests
     - [x] case-insensitive FS test
     - [x] also: see what happens if there are case-collisions (two files fold into one name on a case-insensitive fs)
* [x] **IMPORTANT**: use `gix` from `main` before merging
    - [x] executable bit + modified == modified? Doesn't matter, just be sure UI can easily show it.
    - [x] untracked must be obvious
    - ~~blend `gix::*::Change::Type` into `Change::Modification` maybe~~
        - Actually this helps the code, even though on the consumer side it's a little awkward to use
    - [x] sort changes in tests
    - [x] One set of information per path, worktree takes precedence and and overrides the index
        - ignore the index, but not untracked
    - [x] no IntentToAdd for Frontend
    - [x] Flag with TypeChange or `+x` or `-x`

### Notes for the reviewer

* I left the code in `but-core` besides `gix` there are no dependencies needed to handle the `but_core::WorktreeChanges`.
* Definitely needs `.git/index` tracking - if it changes while a commit is made, we have to reconcile (or at least assure nothing can go wrong)

### For later

* provide patch for a change
* locking information

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -668,6 +668,7 @@ dependencies = [
  "but-workspace",
  "clap",
  "gitbutler-project",
+ "gix",
  "tracing",
  "tracing-forest",
  "tracing-subscriber",
@@ -680,9 +681,9 @@ dependencies = [
  "anyhow",
  "bstr",
  "gix",
+ "gix-testtools",
+ "insta",
  "serde",
- "toml 0.8.19",
- "walkdir",
 ]
 
 [[package]]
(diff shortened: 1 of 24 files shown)


krlvi (approved):
Mostly discussed offline - thanks for this. Im gonna merge this and put together a typescript representation of the return

Byron:
So fast!

PR #5947, opened 2025-01-14
fix: Unwrap the loadable entity value

estib-vega:
Correclty merge the value of the loadable information

### DIFF
--- a/packages/shared/src/lib/network/loadable.ts
+++ b/packages/shared/src/lib/network/loadable.ts
@@ -39,19 +39,19 @@ export function loadableUpsertMany<T, Id extends EntityId>(
 		action: { payload: LoadableData<T, Id>[] }
 	) => {
 		const values = action.payload.map((payload) => {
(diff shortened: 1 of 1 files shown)


packages/shared/src/lib/network/loadable.ts
```diff
@@ -39,19 +39,19 @@ export function loadableUpsertMany<T, Id extends EntityId>(
...
+			if (!(entity.status === 'found' && payload.status === 'found')) {
 				return payload;
 			}
 
 			const newValue: LoadableData<T, Id> = {
 				status: 'found',
 				id: payload.id,
-				value: { ...value, ...payload.value }
+				value: { ...entity.value, ...payload.value }
 			};
```
estib-vega, line 55: We were incorrectly nesting the whole entity in the value property

Caleb-T-Owens (approved):
Oohhh... Yikes

PR #5948, opened 2025-01-14
Fix stack uploading

Caleb-T-Owens:


crates/gitbutler-sync/src/stack_upload.rs
```diff
@@ -142,24 +92,25 @@ fn format_stack_for_review(
...
+            .gitbutler_headers()
+            .map(|headers| headers.change_id)
+            .unwrap_or_else(|| commit.id.to_string());
         message
             .trailers
-            .push((b"Original-Commit".into(), commit_id.to_string().into()));
+            .push(("Change-Id".into(), change_id.into()));
 
-        'heads: for stack_head in heads {
+        'heads: for stack_head in heads.iter().rev() {
```
estib-vega, line 113: Unblocking question: Do we need to swap the direction?

estib-vega (approved):

crates/gitbutler-sync/src/stack_upload.rs
Caleb-T-Owens: Yes, When it was the other way around, if you had an empty branch on top of a branch with commits, the commits were assigned to the wrong branch

PR #5952, opened 2025-01-14
Handle loadable primative types in upsert

Caleb-T-Owens:


packages/shared/src/lib/network/loadable.ts
```diff
@@ -48,10 +48,25 @@ export function loadableUpsertMany<T, Id extends EntityId>(
 				return payload;
 			}
 
+			let merged: T;
+
+			const unmergableTypes = ['string', 'number', 'boolean', 'bigint', 'symbol', 'undefined'];
```
estib-vega, line 53: I like this!
Just a. It: I think it might be better to check whether the types are arrays or non-nullish objects. Less things in the array to check for

estib-vega (approved):

PR #5953, opened 2025-01-14
v3 workspace crate

krlvi:
Starting off with the api for listing the workspace stacks (just a list of ID+names).
This will subsequently hold the functionality around getting the details for a specific stack


Byron (approved):
Great! I will also change my tauri integration to not be in the `but` module.

PR #5954, opened 2025-01-14
Move all desktop components into flat $components folder

mtsgrd:
No hierarchy is better than an incorrect hierarchy.

Benefits of flat structure:
- forces unique names to avoid importing a component using an alias
- no need for a shared folder if two components in different domains share a third component
- changes to component hierarchy without moving files around
- no analysis paralysis when deciding how to organise components into sub directories

Possibly helpful reading: https://vueschool.io/articles/vuejs-tutorials/how-to-structure-a-large-scale-vue-js-application/


Byron (approved):
Sharing some thoughts that were brought up in our conversation this morning:

* every hierarchy of components is flawed as it's not 'unique', there ware various good spots for a component in a tree of components.
    - Thus every 'one' hierarchy is flawed
    - Stop doing hierarchies and put every component with a unique-enough name into a folder.
    - Don't try to impose rules or 'hierarchies' on names either

The above leads to no cognitive load when placing new components, and accepts that internally components use/refer to each other all the time and in ways that akin to a (probably acyclic) graph.

estib-vega (approved):
Oppenheimer GIF

PavelLaptev (approved):

ndom91 (approved):
👀

mtsgrd:
Thanks, I think we have sufficient consensus to land this thing.

PR #5956, opened 2025-01-15
chore: disable in-app updater for Flatpak build

ndom91:
## 🧢 Changes

- There are patches in the `ndom91/gitbutler-flathub` flatpak build repository which disables the tauri updater for Flatpak builds because they install (and update) via the flathub store. Trying to see if I can avoid using a patch and just build this conditional righ tinto the codebase


mtsgrd (approved):

mtsgrd:
This actually doesn't work, `process` is a node thing, I will revert and then let's chat tomorrow!

PR #5957, opened 2025-01-15
A few steps to a cleaner $lib

mtsgrd:
Mostly moving things around, but also splitting certain files and combining others. The goal is to have a well organized `$lib`.


Caleb-T-Owens (approved):

apps/desktop/src/lib/project/project.ts
```diff
@@ -0,0 +1,36 @@
+export type KeyType = 'gitCredentialsHelper' | 'local' | 'systemExecutable';
```
ndom91, line 1: So this looks good in terms of grouping by feature, but what do you think about renaming these `../project/project.ts` type files to `../project/index.ts` so that the import would be like:

```diff
- import { Project } from '$lib/project/project';
+ import { Project } from '$lib/project'; 
```

mtsgrd: I agree that we should improve on that. Let me still land this now and I'll continue with more stuff in a new PR.

PR #5959, opened 2025-01-15
Ability to refetch the chat channel

estib-vega:
- **Interest Store**: Modified the refetch mechanism to be awaitable, allowing for better handling of asynchronous operations.
- **Imperative Upsert**: Updated the functionality to allow the upsert function to be called imperatively and awaited, ensuring that data is refreshed correctly when required.
- **Chat Channel Functionality**: Added a dedicated method for refetching a chat channel which enhances the ability to keep channel data up to date without requiring manual intervention.

### DIFF
--- a/packages/shared/src/lib/chat/chatChannelsService.ts
+++ b/packages/shared/src/lib/chat/chatChannelsService.ts
@@ -52,6 +52,10 @@ export class ChatChannelsService {
 			.createInterest();
 	}
 
+	async refetchChatChannel(projectId: string, changeId?: string): Promise<void> {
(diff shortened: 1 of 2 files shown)


Caleb-T-Owens (approved):
I was wondering how the invalidate thing would end up looking, but this feels like a really nice extension of the API

estib-vega:
> I was wondering how the invalidate thing would end up looking, but this feels like a really nice extension of the API

@Caleb-T-Owens Thanks :) !
I was wondering whether or not it might merit it's own loadable state.
I think that requires a bit longer conversation before I start messing too much with your APIs, and currently there is no need for it

PR #5962, opened 2025-01-15
Use helper functions for routing

Caleb-T-Owens:
Rails generates all your routes for you. I generally dispise having to just write a string and hoping that I didn't screw it up (also how do you change routes if they're referred to by strings??), so I've made some helper functions so we can more formally refer to pages


mtsgrd (approved):

estib-vega (approved):
Love this!

Caleb-T-Owens:
@mtsgrd upgrading svelte fixed the issue. There was nothing explicit in the changelog, but it looks like there was something to do with the order of how reactive things were evaluated, so perhaps something in there helped.

PR #5969, opened 2025-01-16
Review: Display the change information, but make it ugly

estib-vega:
Rough layout of the patch review information, and display the actual data without too much intricate styling

### DIFF
--- a/packages/shared/src/lib/network/loadable.ts
+++ b/packages/shared/src/lib/network/loadable.ts
@@ -8,6 +8,14 @@ export function isFound<T>(loadable?: Loadable<T>): loadable is {
 	return loadable?.status === 'found';
 }
 
+export function isError<T>(loadable?: Loadable<T>): loadable is { status: 'error'; error: Error } {
+	return loadable?.status === 'error';
+}
+
+export function isNotFound<T>(loadable?: Loadable<T>): loadable is { status: 'not-found' } {
+	return loadable?.status === 'not-found';
+}
+
 export function errorToLoadable<T, Id>(error: unknown, id: Id): LoadableData<T, Id> {
 	if (error instanceof Error) {
 		if (error instanceof ApiError && error.response.status === 404) {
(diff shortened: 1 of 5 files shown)


packages/shared/src/lib/network/loadable.ts
```diff
@@ -86,3 +94,34 @@ export function and<T>(
...
+		if (state(a)) {
+			return a;
+		}
+		if (state(b)) {
+			return b;
+		}
+	}
+
+	return { status: 'loading' };
+}
```
estib-vega, line 127: Also added these two loadable util functions

packages/shared/src/lib/network/loadable.ts
```diff
@@ -86,3 +94,34 @@ export function and<T>(
...
+
+export function dig<T, R>(
+	loadable: Loadable<T> | undefined,
+	digger: (current: T) => R
+): R | undefined {
+	if (isFound(loadable)) {
+		return digger(loadable.value);
+	}
+	return undefined;
+}
```
Caleb-T-Owens, line 106: This is rather fancy

packages/shared/src/lib/network/loadable.ts
Caleb-T-Owens, line 127: Can you share an example of what this would look like in practice?

Caleb-T-Owens (approved):

packages/shared/src/lib/network/loadable.ts
estib-vega: I was thinking to use it in situations in which I want to pass two loadables into the `<Loading>` component, so that then both are available inside of the snippet once both are loaded

estib-vega:
```
	<Loading loadable={compose(patch?.current, repositoryId.current)}>
		{#snippet children([patch, repositoryId])}
			// use patch and repositoryId
		{/snippet}
	</Loading>
```

Caleb-T-Owens: cool

PR #5984, opened 2025-01-17
v3 stack branches endpoint WIP

krlvi:
This is a port of "list_virtual_branches" but operating on a single stack at a time and without dependency to the `gitbutler-branch-actions` code.

The key constraint here is to keep this compatible with the stacks implementation. 

Tasks:
- [x] Port over the implementation
- [x] CLI integration
- [x] Tauri API
- [ ] Tests

### DIFF
--- a/crates/but-workspace/src/integrated.rs
+++ b/crates/but-workspace/src/integrated.rs
@@ -1,0 +1,130 @@
+//! ### Detecting if a commit is integrated
+//!
+//! This code is a fork of the [`gitbutler_branch_actions::virtual::IsCommitIntegrated`] in crate gitbutler-branch-actions
+//! This code is a fork of the [`gitbutler_branch_actions::r#virtual::list_virtual_branches`] in crate gitbutler-branch-actions
+
+use anyhow::anyhow;
+use anyhow::{Context, Result};
+use gitbutler_command_context::CommandContext;
+use gitbutler_commit::commit_ext::CommitExt;
+use gitbutler_oxidize::{git2_to_gix_object_id, gix_to_git2_oid, GixRepositoryExt};
+use gitbutler_repo::{
+    logging::{LogUntil, RepositoryExt},
+    RepositoryExt as _,
+};
+use gitbutler_stack::Target;
+use itertools::Itertools;
+
+pub(crate) struct IsCommitIntegrated<'repo, 'cache, 'graph> {
+    gix_repo: &'repo gix::Repository,
+    graph: &'graph mut MergeBaseCommitGraph<'repo, 'cache>,
+    target_commit_id: gix::ObjectId,
+    upstream_tree_id: gix::ObjectId,
+    upstream_commits: Vec<git2::Oid>,
+    upstream_change_ids: Vec<String>,
+}
+
+impl<'repo, 'cache, 'graph> IsCommitIntegrated<'repo, 'cache, 'graph> {
+    pub(crate) fn new(
+        ctx: &'repo CommandContext,
+        target: &Target,
+        gix_repo: &'repo gix::Repository,
+        graph: &'graph mut MergeBaseCommitGraph<'repo, 'cache>,
+    ) -> anyhow::Result<Self> {
+        let remote_branch = ctx
+            .repo()
+            .maybe_find_branch_by_refname(&target.branch.clone().into())?
+            .ok_or(anyhow!("failed to get branch"))?;
+        let remote_head = remote_branch.get().peel_to_commit()?;
+        let upstream_tree_id = ctx.repo().find_commit(remote_head.id())?.tree_id();
+
+        let upstream_commits =
+            ctx.repo()
+                .log(remote_head.id(), LogUntil::Commit(target.sha), true)?;
+        let upstream_change_ids = upstream_commits
+            .iter()
+            .filter_map(|commit| commit.change_id())
+            .sorted()
+            .collect();
+        let upstream_commits = upstream_commits
+            .iter()
+            .map(|commit| commit.id())
+            .sorted()
+            .collect();
+        Ok(Self {
+            gix_repo,
+            graph,
+            target_commit_id: git2_to_gix_object_id(target.sha),
+            upstream_tree_id: git2_to_gix_object_id(upstream_tree_id),
+            upstream_commits,
+            upstream_change_ids,
+        })
+    }
+    pub(crate) fn is_integrated(&mut self, commit: &git2::Commit<'_>) -> Result<bool> {
+        if self.target_commit_id == git2_to_gix_object_id(commit.id()) {
+            // could not be integrated if heads are the same.
+            return Ok(false);
+        }
+
+        if self.upstream_commits.is_empty() {
+            // could not be integrated - there is nothing new upstream.
+            return Ok(false);
+        }
+
+        if let Some(change_id) = commit.change_id() {
+            if self.upstream_change_ids.binary_search(&change_id).is_ok() {
+                return Ok(true);
+            }
+        }
+
+        if self.upstream_commits.binary_search(&commit.id()).is_ok() {
+            return Ok(true);
+        }
+
+        let merge_base_id = self.gix_repo.merge_base_with_graph(
+            self.target_commit_id,
+            git2_to_gix_object_id(commit.id()),
+            self.graph,
+        )?;
+        if gix_to_git2_oid(merge_base_id).eq(&commit.id()) {
+            // if merge branch is the same as branch head and there are upstream commits
+            // then it's integrated
+            return Ok(true);
+        }
+
+        let merge_base_tree_id = self.gix_repo.find_commit(merge_base_id)?.tree_id()?;
+        if merge_base_tree_id == self.upstream_tree_id {
+            // if merge base is the same as upstream tree, then it's integrated
+            return Ok(true);
+        }
+
+        // try to merge our tree into the upstream tree
+        let (merge_options, conflict_kind) = self.gix_repo.merge_options_no_rewrites_fail_fast()?;
+        let mut merge_output = self
+            .gix_repo
+            .merge_trees(
+                merge_base_tree_id,
+                git2_to_gix_object_id(commit.tree_id()),
+                self.upstream_tree_id,
+                Default::default(),
+                merge_options,
+            )
+            .context("failed to merge trees")?;
+
+        if merge_output.has_unresolved_conflicts(conflict_kind) {
+            return Ok(false);
+        }
+
+        let merge_tree_id = merge_output.tree.write()?.detach();
+
+        // if the merge_tree is the same as the new_target_tree and there are no files (uncommitted changes)
+        // then the vbranch is fully merged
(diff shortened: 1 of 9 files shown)


crates/but-workspace/src/integrated.rs
```diff
@@ -0,0 +1,129 @@
+//! ### Detecting if a commit is integrated
+//!
+//! This code is a fork of the `IsCommitIntegrated` in crate gitbutler-branch-actions
```
Byron: This could be an actual doclink so we can be sure it doesn't break. CI might not be testing this ATM, but I know I run it regularly.

crates/but-workspace/src/lib.rs
```diff
@@ -43,6 +59,196 @@ pub fn stacks(gb_dir: &Path) -> Result<Vec<StackEntry>> {
         .collect())
 }
 
+/// Represents the state a commit could be in for the purposes of rendering the UI.
```
Byron: It takes me some time getting used to, but I'd avoid mentioning the UI just to try to stay 'clean'. It's fair to think of this as the business model of GitButler, and these types here would be useful throughout the `but-*` family of crates.

crates/but-workspace/src/lib.rs
```diff
@@ -43,6 +59,196 @@ pub fn stacks(gb_dir: &Path) -> Result<Vec<StackEntry>> {
...
+    /// The `remote_commit_id` may be the same as the `id` or it may be different if the local commit has been rebased or updated in another way.
+    LocalAndRemote(gix::ObjectId),
+    /// The commit is considered integrated.
+    /// This should happen when this commit or the contents of this commit is already part of the base.
+    Integrated,
+}
+
+/// Commit that is a part of a [`StackBranch`] and, as such, containing state derived in relation to the specific branch.
+#[derive(Debug, Clone)]
+pub struct Commit {
```
Byron, line 83: Should this contain more information even? Like custom header values and such?
Then the `has_conflicts` field would be derived from the header fields, making this type a 'true' representation of how GitButler sees commits.

crates/but-workspace/src/lib.rs
```diff
@@ -43,6 +59,196 @@ pub fn stacks(gb_dir: &Path) -> Result<Vec<StackEntry>> {
...
+    pub message: BString,
+    /// Whether the commit is in a conflicted state.
+    /// Conflicted state of a commit is a GitButler concept.
+    /// GitButler will perform rebasing/reordering etc without interruptions and flag commits as conflicted if needed.
+    /// Conflicts are resolved via the Edit Mode mechanism.
+    pub has_conflicts: bool,
+    /// Represents wether the the commit is considered integrated, local only,
+    /// or local and remote with respect to the branch it belongs to.
+    /// Note that remote only commits in the context of a branch are expressed with the [`UpstreamCommit`] struct instead of this.
+    pub state: CommitState,
```
Byron, line 96: I wonder if this should be computed lazily (as it's expensive).

Byron (commented):
Some intermediate comments, will continue in a moment.

crates/but-workspace/src/lib.rs
```diff
@@ -43,6 +59,196 @@ pub fn stacks(gb_dir: &Path) -> Result<Vec<StackEntry>> {
...
+    }
+}
+
+/// Replesents a branch in a [`gitbutler_stack::Stack`]. It contains commits derived from the local pseudo branch and it's respective remote
+#[derive(Debug, Clone)]
+pub struct Branch {
+    /// The name of the branch.
+    pub name: BString,
+    /// Upstream reference, e.g. `refs/remotes/origin/base-branch-improvements`
+    pub upstream_refrence: Option<BString>,
```
Byron: There is a typo, but besides that I *think* the Git name for this is typically the "remote tracking branch". 
To double-check, I used [this chat](https://chatgpt.com/share/678e52af-b954-800a-966d-af66d71aaabf).

crates/but-workspace/src/lib.rs
```diff
@@ -43,6 +59,196 @@ pub fn stacks(gb_dir: &Path) -> Result<Vec<StackEntry>> {
...
+    /// List of commits beloning to this branch. Ordered from newest to oldest.
+    /// Created from the local pseudo branch (head currently stored in the TOML file)
```
Byron: Probably could mention that it's the commits up until the fork point with the trunk/target?

crates/but-workspace/src/lib.rs
```diff
@@ -43,6 +59,196 @@ pub fn stacks(gb_dir: &Path) -> Result<Vec<StackEntry>> {
...
+    pub commits: Vec<Commit>,
+    /// List of commits that exist **only** on the upstream branch. Ordered from newest to oldest.
+    /// Created from the tip of the local tracking branch eg. refs/remotes/origin/my-branch -> refs/heads/my-branch
```
Byron: I think this also does not include anything from `commits`, right?
The local tracking branch (as per [this definition](https://chatgpt.com/share/678e52af-b954-800a-966d-af66d71aaabf)) would be `refs/heads/my-branch` , but I have have the feeling that this is a walk like ` refs/remotes/origin/my-branch..^refs/heads/my-branch`, so everything that's not in the local tracking branch, but in the remote tracking branch.

crates/but-workspace/src/lib.rs
```diff
@@ -43,6 +59,196 @@ pub fn stacks(gb_dir: &Path) -> Result<Vec<StackEntry>> {
...
+    pub upstream_commits: Vec<UpstreamCommit>,
+    /// Description of the branch.
+    /// This is provided by the UI and can include arbitrary utf8 data, eg. markdown etc.
```
Byron: Probably something like GitButler specific metadata, without mentioning the UI.

crates/but-workspace/src/lib.rs
```diff
@@ -43,6 +59,196 @@ pub fn stacks(gb_dir: &Path) -> Result<Vec<StackEntry>> {
...
+    pub description: Option<String>,
+    /// The pull(merge) request associated with the branch, or None if no such entity has not been created.
+    /// This is provided by the UI which is also responsible for creating Pull Requests.
+    pub pr_number: Option<usize>,
```
Byron, line 120: Should we predict support for multiple backends and make the `usize` an Enum variant, or maybe even just a string, but strongly type.

crates/but-workspace/src/lib.rs
```diff
@@ -43,6 +59,196 @@ pub fn stacks(gb_dir: &Path) -> Result<Vec<StackEntry>> {
...
+    /// A unique identifier for the GitButler review associated with the branch, if any.
+    pub review_id: Option<String>,
```
Byron, line 122: This is interesting - should *our* review be special, or is it just one in a list of reviews. Could I push to GitLab and GitHub at the same time? Probably not, so maybe OK to let it be special for now.

crates/but-workspace/src/lib.rs
```diff
@@ -43,6 +59,196 @@ pub fn stacks(gb_dir: &Path) -> Result<Vec<StackEntry>> {
...
+    /// Archived represents the state when series/branch has been integrated and is below the merge base of the branch.
+    /// This would occur when the branch has been merged at the remote and the workspace has been updated with that change.
+    pub archived: bool,
```
Byron: Maybe there is a way to avoid booleans for its state and make it an enum somehow. Maybe that would be an oversimplification, I don't know. It's am a bit afraid of accumulating a ton of booleans here, even though they would be fine in `tauri` types. But even there we can (now) probably be better.

crates/but-workspace/src/lib.rs
```diff
@@ -43,6 +59,196 @@ pub fn stacks(gb_dir: &Path) -> Result<Vec<StackEntry>> {
...
+        let api_commit = Commit {
+            id: commit.id().to_gix(),
+            message: commit.message_bstr().into(),
+            has_conflicts: commit.is_conflicted(),
+            state: CommitState::LocalOnly, // TODO: implement this
+        };
+        patches.push(api_commit);
+    }
+    // There should be no duplicates, but dedup because the UI cant handle duplicates
+    patches.dedup_by(|a, b| a.id == b.id);
```
Byron: I think this only works when the items to be removed are consecutive, so would need to be sorted first. It's probably better to do a `.retain()` and keep track of all seen ids in a hashset.

Byron (commented):
Thanks for having me take an early look! I really like the docs, it's exactly what we need to keep the APIs to the point digest the existing business logic into "what it wants to be".

crates/but-workspace/src/integrated.rs
krlvi: This is a good idea, but I didn't do it because I didn't want to add a dependency to the `gitbutler_branch_actions` crate. Is there a way to add a proper link without a dependency?

crates/but-workspace/src/lib.rs
krlvi: Hmm, it is very possible for this to be the case in the future, but I believe that for now this is all that is needed... Maybe it's easier to add fields on as-needed basis

crates/but-workspace/src/lib.rs
krlvi: The current CommitState is not expensive to compute (it's only trivial operations)

crates/but-workspace/src/lib.rs
krlvi: This is correct - I updated the documentation. `refs/remotes/origin/my-branch..^refs/heads/my-branch` sounds probably right, but the way it is implemented is with a silly commit comparison... I should write a test and see if using git to find the difference yields an identical output

crates/but-workspace/src/lib.rs
krlvi: This used to be an enum but it was simplified to this at some point to make the frontend life easier. The fate of this field is unclear... So perhaps we can leave it as is for now

crates/but-workspace/src/lib.rs
krlvi: I imagine this is part of the code that will be "in flux" for a little while longer as we figure things out with review.

crates/but-workspace/src/lib.rs
krlvi: I agree, this is a great point. In the archived state, a lot of the fields are not applicable. I was going to rewrite this as an enum, but just realized that archived branches are no longer needed for anything... and should simply not be added to the list

crates/but-workspace/src/lib.rs
Byron: `git2` allows to exclude/prune certain parts of the traversal, which should suffice to prevent it from walking into the local tracking branch.
Maybe there are details though that aren't covered by that.

crates/but-workspace/src/lib.rs
Byron: I think as long as we keep making tests for 'JSON-samples' we can be quite free in changing the types we pass to the frontend as they won't break unnoticed anymore.

PR #5985, opened 2025-01-17
feat: add feature toggle for v3 mode [GB-918]

mtsgrd:
- toggle under experimental settings
- keyboard shortcut 'v3'

### DIFF
--- a/apps/desktop/src/lib/config/uiFeatureFlags.ts
+++ b/apps/desktop/src/lib/config/uiFeatureFlags.ts
@@ -6,6 +6,8 @@
  */
 import { persisted } from '@gitbutler/shared/persisted';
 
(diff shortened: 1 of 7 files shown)


ndom91 (approved):
Added a small copy fix. 

Also when I tried to update the setting, my `pathname` was the v2 board URL, i.e. `abc-123-456-789/board` and therefore the regex `rest` group would include the `/board` part and append that onto the end of the new URL to which it would redirect. So that would end up as `project/abc-123-456/board` and that would 404.

Otherwise, LGTM :+1:

ndom91:
Okay LGTM on my side when toggling `v3` feature flag back n forth via keyboard :shipit:

PR #5987, opened 2025-01-17
Get-Patch

Byron:
Add a tauri-callable function to produce a patch from information that was returned by `but_core::workspace::changes()`.

Follow-up of #5945.

### Tasks

* [x] find the right abstractions - what's native diff output, what's processed for frontend
* [x] `but_core::workspace::unified_diff()
    - [x] test patch creation with overlapping context
    - [x] test symlink diffing
    - [x] deleted files must be convenient
    - [x] test rename
    - [x] diff with diff filtering
    - [x] big file
    - [x] binary file
* [x] tauri bindings
* [x] `but status` with `--patch` 
* [x] Don't forget: update `gitoxide` to use latest from `main`

### DIFF
--- a/crates/gitbutler-tauri/src/diff.rs
+++ b/crates/gitbutler-tauri/src/diff.rs
@@ -5,9 +5,10 @@ use gitbutler_serde::BStringForFrontend;
 use serde::Serialize;
 use tracing::instrument;
 
+/// The array of unified diffs matches `changes`, so that `result[n] = unified_diff_of(changes[n])`.
 #[tauri::command(async)]
 #[instrument(skip(projects, changes), err(Debug))]
-pub fn unified_diffs(
+pub fn tree_change_diffs(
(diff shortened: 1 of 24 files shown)


crates/gitbutler-tauri/src/diff.rs
```diff
@@ -0,0 +1,81 @@
+use crate::error::Error;
+use crate::worktree::TreeChange;
+use gitbutler_project::ProjectId;
+use gitbutler_serde::BStringForFrontend;
+use serde::Serialize;
+use tracing::instrument;
+
+#[tauri::command(async)]
+#[instrument(skip(projects, changes), err(Debug))]
+pub fn unified_diffs(
```
krlvi: lets do tree_change_diffs

krlvi (approved):
LGTM

PR #5990, opened 2025-01-17
When stack fully emptied, disreguard old PR information

Caleb-T-Owens:


crates/gitbutler-stack/src/stack.rs
```diff
@@ -542,6 +542,11 @@ impl Stack {
         }
 
         if self.heads.iter().all(|branch| branch.archived) {
+            // Remove previous PR information so it looks like we started a new stack
+            for head in self.heads.iter_mut() {
```
krlvi, line 546: Do you want to do this for all or only for the integrated ones?

Caleb-T-Owens: I think so. Getting rid of context too eagerly could be anoying

krlvi (approved):

PR #6003, opened 2025-01-18
Add types for consuming worktree changes on the frontend

krlvi:
This code is currently inert, but will provide type hints whenever interactions with the new "status" API are being implemented.

@Byron I skipped adding hits for `ChangeState` (the state and previous_state). I am not sure if the UI needs this information for anything. Let me know if I am missing something. Also had to make the previous_path utf8 for the frontend

### DIFF
--- a/apps/desktop/src/lib/worktree/worktree.ts
+++ b/apps/desktop/src/lib/worktree/worktree.ts
@@ -58,6 +58,6 @@ export class IgnoredChange {
 /** The status we can't handle.*/
 export type IgnoredChangeStatus =
 	/** A conflicting entry in the index. The worktree state of the entry is unclear.*/
-	| { type: 'Conflict' }
+	| 'Conflict'
 	/** A change in the `.git/index` that was overruled by a change to the same path in the *worktree*.*/
-	| { type: 'TreeIndex' };
+	| 'TreeIndex';
--- a/apps/desktop/src/lib/worktree/worktree.ts
+++ b/apps/desktop/src/lib/worktree/worktree.ts
@@ -1,0 +1,63 @@
+import { invoke } from '$lib/backend/ipc';
+
+/** Gets the current status of the worktree */
+export async function worktree_changes(projectId: string): Promise<WorktreeChanges> {
+	return await invoke<WorktreeChanges>('changes', { projectId });
+}
+
+/** Contains the changes that are in the worktree */
+export class WorktreeChanges {
+	/** Changes that could be committed. */
+	changes!: [WorktreeChange];
+	/**
+	Changes that were in the index that we can't handle.
+	The user can see them and interact with them to clear them out before a commit can be made.
+	*/
+	ignoredChanges!: [IgnoredChange];
+}
+
+/**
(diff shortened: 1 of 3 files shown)


crates/gitbutler-tauri/src/worktree/mod.rs
```diff
@@ -76,7 +75,7 @@ pub enum Status {
     },
     Rename {
         #[serde(rename = "previousPath")]
-        previous_path: BString,
+        previous_path: BStringForFrontend,
```
Byron, line 78: Thank you, this was an oversight!

The JSON samples in the tests would have caught it, but I never tested a rename 😁.

crates/gitbutler-tauri/src/worktree/tests.rs
```diff
@@ -49,7 +49,7 @@ mod flags {
 #[test]
 fn worktree_change_json_sample() {
     let actual = serde_json::to_string_pretty(&WorktreeChange {
-        path: BString::from("some/file").into(),
+        path: "some/file".into(),
```
Byron, line 52: Good catch!

apps/desktop/src/lib/worktree/worktree.ts
```diff
@@ -0,0 +1,63 @@
...
+	| { type: 'Addition'; subject: { isUntracked: boolean } }
+	/** Something was deleted.*/
+	| { type: 'Deletion' }
+	/** A tracked entry was modified, i.e. content change, type change (eg. it is now a symlink), executable bit change.*/
+	| { type: 'Modification'; subject: { flags: Flags | undefined } }
+	/**
+	An entry was renamed from `previous_path` to its current location.
+	Note that this may include a content change, as well as a change of the executable bit.
+	*/
+	| { type: 'Rename'; subject: { previousPath: string; flags: Flags | undefined } };
```
Byron: In the sample I saw that it could send `flags: null` and wonder if `flags: Flags | null` is something one can or should do?

apps/desktop/src/lib/worktree/worktree.ts
```diff
@@ -0,0 +1,63 @@
...
+	/** The worktree-relative path to the change.*/
+	path!: string;
+	/** The reason for the change being ignored.*/
+	status!: IgnoredChangeStatus;
+}
+
+/** The status we can't handle.*/
+export type IgnoredChangeStatus =
+	/** A conflicting entry in the index. The worktree state of the entry is unclear.*/
+	| { type: 'Conflict' }
```
Byron: For consistency, I think this should just be like `Flags`. I remember that initially I thought these would hold data, but they didn't in the end, and I merely forgot removing the serde configuration.

Byron (commented):
Thanks, this looks good to me even though there might be one place where an improvement probably can be made.

krlvi:
Will adjust. Btw, is it correct of me to assume that the UI has no use for the the state and previous_states? When retrieving the patch content it will need only the path, right?

krlvi:
Merging this as is, since we are still adjusting the API

PR #6005, opened 2025-01-20
fix `Repository::ignore_large_files_in_diffs()` to not do anything if `0` is given.

Byron:
Right now it would actually filter out any file that isn't empty.

It's a bit strange this was working for the oplog, maybe I am missing something.


Caleb-T-Owens (commented):
Should 0 instead represent that there is no limit?

Byron:
Exactly. Previously it would not allow any file larger than 0 bytes as `0` wasn't special cased. Now it is or… should be 😅.

krlvi (approved):
Yeah, this is the intended behavior indeed

PR #6007, opened 2025-01-20
Fork V3 feature inside existing project route

mtsgrd:
- previous fork point would lead to unnecessary code duplication


apps/desktop/src/routes/+layout.svelte
```diff
@@ -99,6 +99,7 @@
 	setContext(LatestBranchLookupService, latestBranchLookupService);
 	setContext(HooksService, data.hooksService);
 	setContext(SettingsService, data.settingsService);
+	setContext(FileService, data.fileService);
```
mtsgrd, line 100: This was incorrectly being set in `[projectId]/layout.svelte`.

apps/desktop/src/routes/+page.svelte
```diff
@@ -33,8 +33,7 @@
 				return { type: 'redirect', subject: `/${projectId}` };
```
ndom91, line 32: Won't let me comment on the part I want, but `line 3` is left over, right? Needs removed to pass CI

ndom91 (approved):
Small left over I think, otherwise LGTM. I'll do the navigation feature flag in my chrome PR :+1:

PR #6011, opened 2025-01-20
feat: v3 chrome 

ndom91:
## 🧢 Changes

- Initial Chrome including sidebar and placeholder files for the new pages
- A lot of the buttons are just placeholders atm, merging this so others are unblocked and can work on their pages though.
- Will continue adding actions to buttons, implementing design details, etc. as a follow-up

image

### DIFF
--- a/apps/desktop/src/components/Chrome.svelte
+++ b/apps/desktop/src/components/Chrome.svelte
@@ -1,24 +1,43 @@
 <script lang="ts">
+	import Header from '$components/ChromeHeader.svelte';
 	import Sidebar from '$components/ChromeSidebar.svelte';
 	import type { Snippet } from 'svelte';
 
 	const { children, projectId }: { children: Snippet; projectId: string } = $props();
 </script>
 
-<Sidebar {projectId} />
-<div class="chrome">
-	{@render children()}
+<div class="container">
+	<Header />
+	<div class="content">
+		<Sidebar {projectId} />
+		<div class="chrome">
+			{@render children()}
+		</div>
+	</div>
 </div>
 
 <style>
-	.chrome {
+	.container {
 		display: flex;
 		flex-grow: 1;
-		height: calc(100% - 32px);
-		margin: 16px;
-		align-items: self-start;
-		user-select: none;
-		position: relative;
-		background-color: var(--clr-bg-2);
+		flex-direction: column;
+		height: 100%;
+
+		.content {
+			display: flex;
+			flex-grow: 1;
+			height: 100%;
+
+			.chrome {
+				display: flex;
+				flex-grow: 1;
+				height: calc(100% - 32px);
+				margin: 16px;
+				align-items: self-start;
+				user-select: none;
+				position: relative;
+				background-color: var(--clr-bg-2);
+			}
+		}
 	}
 </style>
--- a/apps/desktop/src/components/Chrome.svelte
+++ b/apps/desktop/src/components/Chrome.svelte
@@ -1,11 +1,11 @@
 <script lang="ts">
-	import ChromeNavigation from '$components/ChromeNavigation.svelte';
+	import Sidebar from '$components/ChromeSidebar.svelte';
 	import type { Snippet } from 'svelte';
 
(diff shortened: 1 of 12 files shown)


apps/desktop/src/components/Chrome.svelte
```diff
@@ -0,0 +1,42 @@
...
+
+			.content {
+				display: flex;
+				flex-grow: 1;
+				padding: 16px 0 0 16px;
+				align-items: self-start;
+				user-select: none;
+				background-color: var(--clr-bg-2);
+			}
+		}
```
mtsgrd, line 40: For consistency, could we omit the nesting? With scoped css we don't run the risk of name collisions.

apps/desktop/src/components/ChromeHeader.svelte
```diff
@@ -0,0 +1,33 @@
...
+	import { getContext } from '@gitbutler/shared/context';
+	import Button from '@gitbutler/ui/Button.svelte';
+	import { env } from '$env/dynamic/public';
+
+	const project = getContext(Project);
+</script>
+
+<nav>
+	<div class="left">
+		{#if platformName === 'macos' || env.PUBLIC_TESTING}
```
mtsgrd, line 14: How come you added `env.PUBLIC_TESTING`?

apps/desktop/src/components/ChromeSidebar.svelte
```diff
@@ -0,0 +1,149 @@
...
+	import { goto } from '$app/navigation';
+
+	interface Props {
+		projectId: string;
+	}
+
+	const { projectId }: Props = $props();
+
+	function navigateInProject(target: string) {
+		goto(`/${projectId}/${target}`);
```
mtsgrd, line 15: Could you create a simple `routes.svelte.ts` to mirror `webRoutes.svelte.ts` in e.g. `$lib/routes`? Not sure if we need to call it desktopRoutes since it's not shared like webRoutes.

apps/desktop/src/routes/[projectId]/+layout.svelte
```diff
@@ -228,11 +230,17 @@
...
-					<History onHide={() => ($showHistoryView = false)} />
+				{#if $v3}
+					<Chrome {projectId}>
+						{@render children()}
+					</Chrome>
+				{:else}
+					<Navigation />
+					{#if $showHistoryView}
+						<History onHide={() => ($showHistoryView = false)} />
+					{/if}
```
Comment (mtsgrd) [inline apps/desktop/src/routes/[projectId]/+layout.svelte:241]: Would this work correctly if we pull it out of the $v3 if condition and put it after? Feels like it'll still render as an overlay, and we could keep it available until we have new history page.

packages/ui/src/lib/Button.svelte
```diff
@@ -11,6 +11,7 @@
 		// Layout props
 		shrinkable?: boolean;
 		reversedDirection?: boolean;
+		height?: number | string | undefined;
```
mtsgrd, line 14: How come you want to allow string, unlike width which is either undefined or a number?

mtsgrd (approved):

apps/desktop/src/components/ChromeHeader.svelte
ndom91: We used it in other places with the trafficlights, not sure if you mac folks used this. But if not I can just drop it 👍

packages/ui/src/lib/Button.svelte
ndom91: For "auto" or "100%" and the like, but maybe not necessary. I'll double check this and either pull it or make width also take it so it's consistent

apps/desktop/src/routes/[projectId]/+layout.svelte
Comment (ndom91) [reply apps/desktop/src/routes/[projectId]/+layout.svelte:241]: Hmm yeah I'll check

apps/desktop/src/components/ChromeHeader.svelte
mtsgrd: I think we can drop it!

PR #6028, opened 2025-01-21
Fix #6013

### LINKED ISSUE #6013: Failed to sign SSH on Windows
### Version

0.14.5

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

With the latest release I gave SSH signing on Windows another go, this used to show `Program not found`.

It's now getting further but I'm still hitting an issue with it now instead returning `Failed to sign SSH: Cannot open C:UsersDanielWAppDataLocalTemp.tmp1cG0m8 for signing: No such file or directory`

Looks like sh from Git likes the windows formatted path being passed in, curiously the signing key which is also a similar path does work though.

1. Open Gitbutler & Navigate to the Git settings of a Project
2. Select to Sign commits & configure for SSH signing with a valid key & signing program empty (default)
3. Click Test Signing

I should be able to sign using SSH on Windows without receiving an error.

#### From Test in Settings

```shell
2025-01-20T21:05:10.414873Z  INFO check_signing_settings: crates\gitbutler-tauri\src\repo.rs:39: new id=7f364b41-2e01-4bef-abb3-9529ca64f1c6
2025-01-20T21:05:10.468239Z ERROR check_signing_settings: crates\gitbutler-tauri\src\repo.rs:39: error=Error(Failed to sign SSH:  Cannot open C:UsersDanielWAppDataLocalTemp.tmp1id5Uw for signing: No such file or directory
) id=7f364b41-2e01-4bef-abb3-9529ca64f1c6
2025-01-20T21:05:10.468268Z  INFO check_signing_settings: crates\gitbutler-tauri\src\repo.rs:39: close time.busy=53.4ms time.idle=10.2µs id=7f364b41-2e01-4bef-abb3-9529ca64f1c6
```

#### From Committing

```shell
2025-01-20T20:56:33.865799Z  INFO commit_virtual_branch: crates\gitbutler-tauri\src\virtual_branches.rs:31: new project_id=7f364b41-2e01-4bef-abb3-9529ca64f1c6 branch=2d6b08c6-db63-4e8c-9d18-ed34e411a83a Test commit ownership=Some(BranchOwnershipClaims { claims: [OwnershipClaim { file_path: "CHANGELOG.md", hunks: [Hunk { hash: Some(75b3ae88934aac9988ec5ca701136536), start: 1, end: 7 }] }] })
2025-01-20T20:56:33.973946Z ERROR commit_virtual_branch: crates\gitbutler-tauri\src\virtual_branches.rs:31: error=Error(failed to commit

Caused by:
    0: errors.commit.signing_failed
    1: Failed to sign commit: Failed to sign SSH:  Cannot open C:UsersDanielWAppDataLocalTemp.tmpdhLpLR for signing: No such file or directory
       ) project_id=7f364b41-2e01-4bef-abb3-9529ca64f1c6 branch=2d6b08c6-db63-4e8c-9d18-ed34e411a83a Test commit ownership=Some(BranchOwnershipClaims { claims: [OwnershipClaim { file_path: "CHANGELOG.md", hunks: [Hunk { hash: Some(75b3ae88934aac9988ec5ca701136536), start: 1, end: 7 }] }] })
2025-01-20T20:56:33.973978Z  INFO commit_virtual_branch: crates\gitbutler-tauri\src\virtual_branches.rs:31: close time.busy=108ms time.idle=9.50µs project_id=7f364b41-2e01-4bef-abb3-9529ca64f1c6 branch=2d6b08c6-db63-4e8c-9d18-ed34e411a83a Test commit ownership=Some(BranchOwnershipClaims { claims: [OwnershipClaim { file_path: "CHANGELOG.md", hunks: [Hunk { hash: Some(75b3ae88934aac9988ec5ca701136536), start: 1, end: 7 }] }] })
```

### LINKED ISSUE #6082: Failed to sign SSH on Windows
### Version

0.14.6

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

I have the same problem as described in #6013

Image

and as error I get this
```
Cannot open C:UsersMaximAppDataLocalTemp.tmpkWZqVX for signing: No such file or directory
```

I tried even through terminal
```shell
git config --global user.signingkey "C:\Users\Maxim\.ssh\id_ed25519.pub"
git config --global gpg.format ssh
git config --global gitbutler.signCommits true
```

1. Open Gitbutler & Navigate to the Git settings of a Project
2. Select to Sign commits & configure for SSH signing with a valid key & signing program empty (default)
3. Click Test Signing

```shell
userService.BBMm1TBE.js:17 
 Error checking signing: s: Failed to sign SSH:  Cannot open C:UsersMaximAppDataLocalTemp.tmpkWZqVX for signing: No such file or directory

    at s.fromError (http://tauri.localhost/_app/immutable/chunks/ipc.BXwWcGhi.js:1:1095)
    at p (http://tauri.localhost/_app/immutable/chunks/ipc.BXwWcGhi.js:1:1299)
    at async Object.h [as onclick] (http://tauri.localhost/_app/immutable/nodes/27.BwNvshWK.js:10:818)
```

```
2025-01-25T10:55:13.754702Z  INFO check_signing_settings: crates\gitbutler-tauri\src\repo.rs:39: new id=ad6a2574-5f43-48db-8013-413bd9c62cc9
2025-01-25T10:55:13.823615Z ERROR check_signing_settings: crates\gitbutler-tauri\src\repo.rs:39: error=Error(Failed to sign SSH:  Cannot open C:UsersMaximAppDataLocalTemp.tmpysIlt5 for signing: No such file or directory
) id=ad6a2574-5f43-48db-8013-413bd9c62cc9
2025-01-25T10:55:13.823638Z  INFO check_signing_settings: crates\gitbutler-tauri\src\repo.rs:39: close time.busy=68.9ms time.idle=15.8µs id=ad6a2574-5f43-48db-8013-413bd9c62cc9
```

Byron:
Make sure paths are quoted when passing them to the shell that runs the signing command.

We run it in a shell, specifically the Git shell, on Windows so more programs are available.
On Unix it sholdn't hurt even though it doesn't have a specific purpose there.

The quoting is needed on Windows to prevent backslashes to mean escaping.

Fixes #6013 .

### DIFF
--- a/crates/gitbutler-repo/src/repository_ext.rs
+++ b/crates/gitbutler-repo/src/repository_ext.rs
@@ -450,13 +450,13 @@ impl RepositoryExt for git2::Repository {
                 }
 
                 let args = format!(
-                    "{} -U {}",
+                    "'{}' -U '{}'",
                     key_storage.path().to_string_lossy(),
                     buffer_file_to_sign_path_str,
                 );
                 cmd_string += &args;
             } else {
-                let args = format!("{} {}", signing_key, buffer_file_to_sign_path_str);
+                let args = format!("'{}' '{}'", signing_key, buffer_file_to_sign_path_str);
                 cmd_string += &args;
             };
             let mut signing_cmd: std::process::Command = command_with_login_shell(cmd_string);
@@ -581,11 +581,13 @@ impl RepositoryExt for git2::Repository {
 }
 
 pub fn command_with_login_shell(shell_cmd: impl Into<OsString>) -> std::process::Command {
-    gix::command::prepare(shell_cmd)
+    let cmd: std::process::Command = gix::command::prepare(shell_cmd)
         .with_shell_disallow_manual_argument_splitting()
         // On Windows, this yields the Git-bundled `sh.exe`, on Linux it uses `/bin/sh`.
         .with_shell_program(gix::path::env::shell())
-        .into()
+        .into();
+    tracing::debug!(?cmd, "command to produce commit signature");
+    cmd
 }
 
 /// Signs the buffer with the configured gpg key, returning the signature.


krlvi (approved):

EliahKagan (commented):
If it is ever possible for the strings being single-quoted to contain single quotes, then their single quotes should also be neutralized by replacing each original occurrence of `'` with `'\''`.

(This works in `sh`, other shells behaving POSIX-compatibly, and in practice in Bourne-style shells in general. It looks like it would always work here. It is not guaranteed to work in non-Bourne-style shells. But in most non-Bourne-style shells, single-quoting is not as strong, so pasting them on the outside would be insufficient for other reasons.)

---

For quoting `command` when it is, in its entirety, the path to a program, it seems like this is something gitoxide might reasonably be able to do, as an opt-in feature when using `gix_command::Prepare`. I've opened https://github.com/GitoxideLabs/gitoxide/issues/1799 about this and other aspects of ambiguity in how the command and argument strings are interpreted.

Please note that, at least as proposed, it looks like that would not cover any of the quoting done in this PR, which seems to affect only arguments. But I think existing `gix_command` functionality could, if desired, neutralize that by using `args` for everything but the command name.

(If done here, that change would reach all the way up to the `let mut cmd_string` line. But it should work just as well to neutralize existing quotes when passing quotes the outside of arguments or, if it is certain that no `'` will ever appear, to do nothing further.)

Byron:
Thanks for researching this. I'd also think that it would be better not to build command-strings here and instead pass everything as arguments officially. That way everything regarding escaping should be taken care of properly.
Even though it's not relevant here as the input is well-known (except for the path to the command), that would definitely make for simpler and less error-prone code.

EliahKagan:
> Even though it's not relevant here as the input is well-known (except for the path to the command)

The path to the command is the main part I would be worried about, since it might potentially be in an unusually named directory. (I don't know if Windows will ever automatically name a user profile directory with a `'` in it. So I don't know how plausible the specific example I gave would be.)

Byron:
While I am working on this, do you think that it's appropriate to forcefully change an ssh key to have usable permissions?

https://github.com/gitbutlerapp/gitbutler/blob/bc482aa7e8e09df24546b58d1215e91600031330/crates/gitbutler-repo/src/repository_ext.rs#L444-L450

EliahKagan:
I'm not sure and I think I would need to understand more about the context. 0o600 are the usual permissions for it. I can think of two possible concerns:

- Maybe someone set the permissions to something `ssh` won't use (probably by making them too restrictive) to prevent a key they don't want used from being used.
- Maybe someone set the permissions too broadly, so that `ssh` refuses to use them, and depending on their threat model--for example, if there are other users on the system--the appropriate solution may be to refrain from using (or stop using) that key and generate a new one that is known not to have leaked.

But I emphasize I haven't looked into the details of how this is being used, nor as of yet thought through the full implications of either of those scenarios.

Byron:
Thank you! Then I will leave it as it clearly is a convenience measure. I'd think that if an attacker can create SSH keys there, it would probably also be able to set their permissions accordingly. Probably SSH also checks ownership, which isn't changed here, and which might prevent a 'key-swap' attack of sorts.

EliahKagan:
> I'd think that if an attacker can create SSH keys there, it would probably also be able to set their permissions accordingly. Probably SSH also checks ownership, which isn't changed here, and which might prevent a 'key-swap' attack of sorts.

At least when the key is in the usual location of the user's `.ssh` directory, I would say that if an attacker can or could previously modify it or create other files there, then the attacker can already configure `ssh` for the user in such a way as to cause the user big problems, separately from being able to read or change the private key file directly.

PR #6031, opened 2025-01-21
Don't do anything when dropping a commit onto itself

### LINKED ISSUE #6029: Dragging and dropping commit in a lane to the same spot raises error
### Version

0.14.6

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

Image

Drag and drop commit on a lane without moving it anywhere. Just to the same spot.

_No response_

```shell

```

Caleb-T-Owens:
* Closes https://github.com/gitbutlerapp/gitbutler/issues/6029

### DIFF
--- a/crates/gitbutler-branch-actions/src/squash.rs
+++ b/crates/gitbutler-branch-actions/src/squash.rs
@@ -192,6 +192,13 @@ fn validate(
     source_commits: &[git2::Commit<'_>],
     destination_commit: &git2::Commit<'_>,
 ) -> Result<()> {
+    if source_commits
+        .iter()
+        .any(|s| s.id() == destination_commit.id())
+    {
+        bail!("cannot squash commit into itself")
+    }
+
     for source_commit in source_commits {
         if !branch_commit_oids.contains(&source_commit.id()) {
             bail!("commit {} not in the stack", source_commit.id());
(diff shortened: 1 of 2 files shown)


krlvi (approved):
Brilliant, thanks!

crates/gitbutler-branch-actions/src/squash.rs
```diff
@@ -192,6 +192,13 @@ fn validate(
     source_commits: &[git2::Commit<'_>],
     destination_commit: &git2::Commit<'_>,
 ) -> Result<()> {
+    if source_commits
```
Byron, line 195: If there are multiple commits A, B, C, could I not say to squash it all into `B`? This seems not unreasonable.

Caleb-T-Owens: Why should B be a source AND the destination?

Caleb-T-Owens: Should you not say you're squashing A and C into B?

Byron: Ah, right, so that source commits never contain the destination. Yes, that's fair.
Since the user is making the selections though, maybe we could auto-correct what's correctable? Either in the frontend or in the backend?

PR #6035, opened 2025-01-21
Changes between trees

Byron:
Provide changes between trees.

### Tasks

* [x] in `but-core`
    - [x] one test with all kinds of tree/tree changes
    - [x] unify Change type
    - [x] commit-id that deals with conflicting trees
* [x] in `tauri`
    - [x] path-bytes
    - [x] JSON sample with rename
* [x] CLI commit-to-commit with revspec support

### Notes for the Reviewer

* Could add `Commit::changes_from(previous_commit)` for maybe nicer API?
* Let's assure our terminology is complete in `but-core`.


crates/gitbutler-serde/src/lib.rs
```diff
@@ -3,6 +3,28 @@ use serde::Serialize;
...
+    /// serialize an object ID as hex-string.
+    pub fn serialize<S>(v: &BString, s: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        v.to_str_lossy().serialize(s)
+    }
+
+    /// deserialize an object ID from hex-string.
+    pub fn deserialize<'de, D>(d: D) -> Result<BString, D::Error>
```
krlvi: Maybe we dont want a lossy deserialization?

krlvi: I kind of needed this code so https://github.com/gitbutlerapp/gitbutler/pull/6045

Byron: A great catch!After extra-validation I could conclude that in order to do it correctly for TreeChanges, we'd have to keep the original byte buffer around to avoid the frontend having to deal with it. Tauri would then once again have a duplicate type with the extra byte buffer, just so that the plumbing types isn't 'polluted' by it.

krlvi (approved):

PR #6036, opened 2025-01-21
File watcher only emits worktree changes if the v3 feature flag is enabled

krlvi:


Byron (approved):
This looks great!

For a moment I thought one could do less clones in the watcher, but that's unfortunately not possible as it spawn a new handler each time there are events (for parallelism), so passing references isn't possible.

PR #6046, opened 2025-01-22
Add redux based service for stacks

mtsgrd:
- uses RTK Query

---
- <kbd>&nbsp;2&nbsp;</kbd> #6054 
- <kbd>&nbsp;1&nbsp;</kbd> #6046 👈


apps/desktop/src/lib/stacks/stacksService.ts
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	select(projectId: string) {
+		return this.stacksApi.endpoints.get.select({ projectId });
+	}
+
+	// eslint-disable-next-line @typescript-eslint/promise-function-async
+	new(projectId: string) {
+		return this.state.dispatch(this.stacksApi.endpoints.new.initiate({ projectId }));
+	}
```
mtsgrd: Redux returns promise like objects that we do not want to automatically await. The return value of `initiate` has additional methods like `abort`, `unsubscribe` etc. Should we remove this lint rule or disable it on a case-by-case basis?

apps/desktop/src/lib/stacks/stacksService.ts
```diff
@@ -0,0 +1,50 @@
+import { reduxApi } from '$lib/redux/api';
```
krlvi: Probably it is a better form to call this "stackService" instead of pluralizing. This, as well as all other ones where you append @mtsgrd

apps/desktop/src/lib/stacks/stacksService.ts
```diff
@@ -0,0 +1,50 @@
...
+		})
+	});
+
+	constructor(private state: DesktopRedux) {}
+
+	poll(projectId: string): Unsubscribe {
+		return this.state.dispatch(
+			this.stacksApi.endpoints.get.initiate(
+				{ projectId },
+				{ subscribe: true, subscriptionOptions: { pollingInterval: POLLING_INTERVAL } }
```
krlvi: Why polling? Consider it is probably better to never poll these because we wish the app to be deterministic

krlvi (changes requested):

apps/desktop/src/lib/stacks/stacksService.ts
mtsgrd: Just trying it out, but removed it now from this pr.

Caleb-T-Owens (approved):

apps/desktop/src/lib/stacks/stacksService.ts
krlvi: cool cool

krlvi (approved):

PR #6052, opened 2025-01-22
Sort stacks before returning them to the front end

mtsgrd:


krlvi (changes requested):
Lets change this so that the list is still sorted according to the order but without leaking `order` to the frontend in order to keep the surface area smaller

mtsgrd:
@krlvi I did not add `order` to the struct returned to the front end.

krlvi:
@mtsgrd got it. You can still a simple sort in the but-workspace crate, and this way we don't need to introduce an extra field.

``` diff
diff --git a/crates/but-workspace/src/lib.rs b/crates/but-workspace/src/lib.rs
index 6ed7306b0..ad581f553 100644
--- a/crates/but-workspace/src/lib.rs
+++ b/crates/but-workspace/src/lib.rs
@@ -25,6 +25,7 @@ use gitbutler_oxidize::OidExt;
 use gitbutler_stack::stack_context::CommandContextExt;
 use gitbutler_stack::{stack_context::StackContext, Stack, Target, VirtualBranchesHandle};
 use integrated::IsCommitIntegrated;
+use itertools::Itertools;
 use serde::Serialize;
 use std::path::Path;
 use std::str::FromStr;
@@ -53,6 +54,7 @@ pub fn stacks(gb_dir: &Path) -> Result<Vec<StackEntry>> {
     Ok(state
         .list_stacks_in_workspace()?
         .into_iter()
+        .sorted_by_key(|s| s.order)
         .map(|stack| StackEntry {
             id: stack.id,
             branch_names: stack.heads().into_iter().map(Into::into).collect(),
```

krlvi (approved):

PR #6056, opened 2025-01-22
Add interaction to v3 tabs

mtsgrd:
- return stack from back end when creating a new one
- navigate to new stack when creating it
- hover/active states for tabs

image

### DIFF
--- a/apps/desktop/src/lib/routes/routes.svelte.ts
+++ b/apps/desktop/src/lib/routes/routes.svelte.ts
@@ -48,3 +48,7 @@ export class DesktopRoutesService {
 		return `/${targetProjectId}/${targetRestPath}`;
 	}
 }
+
+export function stackPath(projectId: string, stackId: string) {
+	return `/${projectId}/workspace/${stackId}`;
+}
--- a/crates/gitbutler-tauri/src/virtual_branches.rs
+++ b/crates/gitbutler-tauri/src/virtual_branches.rs
@@ -1,5 +1,6 @@
 pub mod commands {
     use anyhow::{anyhow, Context};
+    use but_workspace::StackEntry;
     use gitbutler_branch::{BranchCreateRequest, BranchUpdateRequest};
     use gitbutler_branch_actions::branch_upstream_integration::IntegrationStrategy;
     use gitbutler_branch_actions::internal::StackListResult;
@@ -79,12 +80,12 @@ pub mod commands {
         settings: State<'_, AppSettingsWithDiskSync>,
         project_id: ProjectId,
         branch: BranchCreateRequest,
-    ) -> Result<StackId, Error> {
+    ) -> Result<StackEntry, Error> {
         let project = projects.get(project_id)?;
         let ctx = CommandContext::open(&project, settings.get()?.clone())?;
         let stack_entry = gitbutler_branch_actions::create_virtual_branch(&ctx, &branch)?;
         emit_vbranches(&windows, project_id);
-        Ok(stack_entry.id)
+        Ok(stack_entry)
     }
 
     #[tauri::command(async)]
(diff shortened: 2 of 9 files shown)


crates/gitbutler-tauri/src/virtual_branches.rs
```diff
@@ -19,6 +19,7 @@ pub mod commands {
     use tauri::State;
     use tracing::instrument;
 
+    use crate::workspace::StackEntry;
```
mtsgrd: @krlvi this dependency isn't cool, but what can we do? I'd want to return the correct type so that we can use the response in the stack service.

apps/desktop/src/lib/routes/routes.svelte.ts
```diff
@@ -48,3 +48,7 @@ export class DesktopRoutesService {
 		return `/${targetProjectId}/${targetRestPath}`;
 	}
 }
+
+export function stackPath(projectId: string, stackId: string) {
+	return `/${projectId}/workspace/${stackId}`;
+}
```
mtsgrd, line 54: @ndom91 I'm wondering if the class above is needed? There seems to be different requirements for web routes, but here the reactivity doesn't benefit from the class.. and I'm thinking this is unlikely something we'll want to mock in testing.

Caleb-T-Owens: 🤔

I'd be for making a routing class exactly like what is in the web, it seems like a very helpful way of handling things like sidebar state and breadcrumbs.

Surly we should look to finding a way of making our tests work with .svelte.ts files, rather than trying to avoid them?

crates/gitbutler-tauri/src/virtual_branches.rs
```diff
@@ -73,11 +74,14 @@ pub mod commands {
...
         branch: BranchCreateRequest,
-    ) -> Result<StackId, Error> {
+    ) -> Result<StackEntry, Error> {
         let project = projects.get(project_id)?;
         let branch_id = gitbutler_branch_actions::create_virtual_branch(&project, &branch)?;
         emit_vbranches(&windows, project_id);
-        Ok(branch_id)
+        Ok(StackEntry {
+            id: branch_id,
+            branch_names: vec![],
```
krlvi: This is actually not correct here. A stack always has at least one branch. You could return a StackEntry here https://github.com/gitbutlerapp/gitbutler/blob/48c86ef9adfce6f3ab25fb814469418a50601aeb/crates/gitbutler-branch-actions/src/actions.rs#L110 and including the branch names.

It's not a problem if `gitbutler-branch-actions` gains a dependency on `but-workspace` because one of the design goals for `but-workspace` is to **not** depend on branch actions (hence no risk of cycle). 

`gitbutler-branch-actions` is destined to be dismantled post v3

crates/gitbutler-tauri/src/virtual_branches.rs
krlvi: It's not a problem but you will likely need to do this earlier in the flow (see the other comment)

apps/desktop/src/lib/routes/routes.svelte.ts
mtsgrd: My point is, the class is currently just decorative, it appears to me it serves no purpose other than making these functions less accessible.

crates/gitbutler-tauri/src/virtual_branches.rs
mtsgrd: Right yes, I forgot when writing that branch needs to be included. The dependency, however, is not to `gitbutler-branch-actions`, it's to `workspace.rs`, a sibling to this `virtual_branches.rs`.

So my question is really, where can we move `StackEntry`? Feels like `but-workspace` is the right answer, but would be great if you can confirm.

krlvi: This should help https://github.com/gitbutlerapp/gitbutler/pull/6064

mtsgrd: Thank you!

PR #6070, opened 2025-01-23
Add service for worktree changes

mtsgrd:
Adds service and wires up basic functionality.

- new ChangeList named components will replace FileList*
- rewritten id selection mechanism
- no "ownership" support yet for committing
- make ReduxResult support generic data type
- create new component for worktree changes

image


apps/desktop/src/components/ReduxResult.svelte
```diff
@@ -1,19 +1,27 @@
...
+		error?: unknown;
+		data?: A;
+		children: Snippet<[A]>;
+		empty: Snippet;
 	};
 
-	const { status, error, children }: Props = $props();
+	// eslint-disable-next-line no-undef
+	const { status, data, error, children, empty }: Props<A> = $props();
+	$effect(() => console.log(data));
```
Caleb-T-Owens: 👀

apps/desktop/src/components/ReduxResult.svelte
```diff
@@ -1,19 +1,27 @@
...
-	type Props = {
+	type Props<A> = {
 		status: QueryStatus;
-		error: unknown;
-		children: Snippet;
+		error?: unknown;
+		data?: A;
+		children: Snippet<[A]>;
+		empty: Snippet;
 	};
```
Caleb-T-Owens, line 12: I'm wondering about this API. Is it going to get tedious always restructuring before passing to your result?

I'm just imagining if I had two or three queries in one component, I'd much rather say
```
const foo = getFoo()
const bar = getBar()
const baz = getBaz()
```

rather than

```
const { data: fooData, error: fooError, status: fooStatus } = getFoo()
const { data: barData, error: barError, status: barStatus } = getBar()
const { data: bazData, error: bazError, status: bazStatus } = getBaz()
```

If we work with the whole response object, it also means we can define functions like `map` and `combine` like we have in the web project

apps/desktop/src/lib/worktree/worktreeTypes.ts
```diff
@@ -78,12 +53,12 @@ export type DiffHunk = {
...
 export type WorktreeChanges = {
 	/** Changes that could be committed. */
-	readonly changes: TreeChange[];
+	changes: TreeChange[];
 	/**
 	 * Changes that were in the index that we can't handle.
 	 * The user can see them and interact with them to clear them out before a commit can be made.
 	 */
-	readonly ignoredChanges: IgnoredChange[];
+	ignoredChanges: IgnoredChange[];
```
Caleb-T-Owens: Why can't these be readonly?

krlvi:
Very neat, liking this. One thought, perhaps not for this PR - can we somehow also render the "ignored changes"? Maybe in a separate category and without the ability to interact with them.
I just dont know how to simulate those for testing since it's only exceptional cases that would have these

krlvi:
Another thing - the `worktree.ts` (or `worktreeTypes.ts` now) name is probably a bit out of date, because a lot of the types that were added there are now more generically applicable. TreeChange, UnifiedDiff etc are gonna be applicable outside the worktree stuff, but I didnt know where to put it on the frontend

ndom91 (approved):
Not a big deal, I can do this in my PR if you want, but the empty purgatory styling has been bonked

image

Otherwise, LGTM! Everythign seems to work

mtsgrd:
@ndom91 thanks for spotting. I removed `svg` from the reset css where it was set to display block, and vertical align middle. Those settings are probably worse than default imho, and now that illustration centers.

@krlvi I moved the types to more appropriate places, and yes let's find a way of displaying the ignored files. Landing this for now.

PR #6077, opened 2025-01-24
hunk dependencies

Byron:
Given a set of unified diffs as extracted from worktree changes, along with a set of commits, find the commits that have changed hunks which 'connect' to the changed hunks from the worktree changes.

This association allows to merge these hunks into the right commit as part of 'absorbing' them.

### Tasks - `but-hunk-dependency`

* [x] write down how it *should* work on a high-level and how it covers everything that it needs to
* [x] sketch out minimal types to do perform the task, using the new API types, port over the machinery from there.
* [x] cleanup and simplification
    - [x] make data-structures insta-compatible
* ~~does the algorithm need `StackId`?~~ Even if not (probably truly not needed), there is no value in removing it right now.
* [x] simplify data structure in plumbing
* [x] port `dependencies` tests to new types and to assure traversal is implemented correctly.
* [x] tauri bindings that does all the frontend conversions
* [x] JSON sample for hunk-dependencies
* [x] CLI bindings, maybe even to make it work with the current branch
* [ ] ~~fix the failing tests (also can't debug-print maps as they are unstable)~~ - can't do it, yet, let's pick it up again.

### Follow-up Fixes

* Existing `gix` traversals should use `git2` while there is no proper `hide()`.

### Possible Follow-up PR

Let's do these once there is a UI to test with, the CLI isn't suitable for it due to its one-shot nature.

* Make it possible to re-use a `gix::Repository` across similar calls so that object caches can be reused, something that should lead to a considerable speedup.
* Hunk-dependencies should only be calculated for the paths that actually changed - right now it computes everything.

### Notes for the reviewer

* **There is many TODOs left**, and some are related to ported tests that now seem to not pass. They are all related to 'manual hunks' that are now set as worktree changes, it's unclear what the issue is, and if the result is right or not. But the worst is 1m per test-run as it recompiles half the world in the `gitbutler-tauri` crate each time.
    - I recommend picking this up again once the UI is ready to consume it, the more work will be needed.
* The `tauri` version of the data structure as ported from `gitbutler-hunk-dependency` is lacking its structure-specific tests. This could be ported once it's clear that the UI still needs that very data structure.
* **It's unclear how the frontend deals with hunk-hashes without having right hunk to hash the lines from** - after all, it has 3 context-lines but the hash is created with zero lines

### Research Notes on `but-hunk-dependency`

* Knowledge of `StackId` shouldn't be needed
* Ideally it could traverse the commit-graph and do so step by step

### Research Notes (new implementation, on hold)

* (Digested into actual crate)
* Is it possible to three-way-subtract trees, or blobs, to undo one side of a merge? Dealing with Workspaces is troublesome.
* Actually `blame` does handle multi-parent merges just fine, so I think it's a solution to the `workspace` problem. Maybe a `blame` based solution is the way to go after all.


crates/gitbutler-tauri/src/workspace.rs
```diff
@@ -29,3 +37,160 @@ pub fn stack_branches(
...
+    // TODO: could this be a specific type? Is the mapping truly required?
+    //       Is this because `commit_dependent_diffs` use `HunkHash`?
+    // TODO: the frontend actually has no way of associating the hunks it gets with this hash as it's made
+    //       on the patch lines without any context lines, while it has context lines.
+    //       Hash must then skip the context lines if there are any.
+    pub diffs: Vec<(HunkHash, Vec<HunkLock>)>,
+    /// A map from stack id to commit dependencies.
+    /// Commit dependencies map commit id to commits it depends on.
+    // TODO: have to use strings as keys for serialization (`gix::ObjectId`)
+    #[serde(skip)]
```
krlvi: Something that comes up to my mind while reading this - a lot of this data from the v2 flow was added "because we could" but it was never used by the frontend. 
For example the inverse dependencies, we might not need to calculate / present to the fronted.

Perhaps in the spirit of keeping it simple, we could start off with just the plain hunk dependency (currently in the diffs field) and nothing else, if that makes things easier.

crates/gitbutler-tauri/src/workspace.rs
```diff
@@ -29,3 +37,160 @@ pub fn stack_branches(
...
+/// may have succeeded computing.
+#[derive(Debug, Clone, Serialize)]
+pub struct HunkDependencies {
+    /// A map from diffs to branch and commit dependencies.
+    // TODO: could this be a specific type? Is the mapping truly required?
+    //       Is this because `commit_dependent_diffs` use `HunkHash`?
+    // TODO: the frontend actually has no way of associating the hunks it gets with this hash as it's made
+    //       on the patch lines without any context lines, while it has context lines.
+    //       Hash must then skip the context lines if there are any.
+    pub diffs: Vec<(HunkHash, Vec<HunkLock>)>,
```
krlvi, line 102: Re: `HunkHash` - it is probably time for us to start thinking about how the frontend and backend can talk with each other about specific hunks. It would be good if the language can be generalize-able too. 

Here are a few more cases where UI and rust need to exchange information about which hunk is which:
- Committing (the UI is able to say commit only hunks A and B and not the others)
- Discarding hunks (the UI instructs rust to remove/reset the specific change)
- Amending / fixing-up a hunk into an existing commit
Possibly other ones, and of course the use case here, indicating that hunk A depends on commit X & Y.

It would be really neat if we could have a single type that is used in those scenarios, because this would make the integration from the UI side easier and less error prone.

Intuitively, i would think that a line number range is a good identifier but maybe this is missing something. 

---
And there is one more thing - in the use case of "committing" something that would be nice to do is have the ability  to commit part of a hunk (i.e. the `-p` option).

krlvi (approved):
Because this is a new endpoint and none of the existing flows are touched, from my perspective this is good to merge and then iterate upon

crates/gitbutler-tauri/src/workspace.rs
Byron: Thanks, your comment helped me to go all in on the culling, which helped to remove a lot of code and plenty of tests which now would show no changes at all. Or in other words, the tests were using commit-dependencies as a way to understand about hunk relationships.
This, however, is already 'pinned' in `but-hunk-dependency` using `insta`.
My thinking is that tests can be re-added once it's clear what's needed.

The edits are in https://github.com/gitbutlerapp/gitbutler/pull/6077/commits/4db1d1b62b5263ce8f29d9728f41e32aefc40097 .

crates/gitbutler-tauri/src/workspace.rs
Byron: This is very valid, thanks for bringing it to my attention. I found the listing of primary cases very helpful.

> Intuitively, i would think that a line number range is a good identifier but maybe this is missing something.

This is also what I would think, and hunk-ownership already used this system so I think it's proven. Filename + Hunk-header-info can identify a hunk perfectly.

> And there is one more thing - in the use case of "committing" something that would be nice to do is have the ability to commit part of a hunk (i.e. the `-p` option).

I think this could be expressed as an operation on top of a hunk, i.e. give it a series of splits and what to do with each side of the split. That's nice and 'additive' so each of these types has a clear purpose.

For now I'd leave it as is along with the TODO, so this can be tackled when needed.

PR #6088, opened 2025-01-27
fix: Correctly unsubscribe shortcut listeners

estib-vega:
The unsubscribe function was calling the listener callback instead of finding the index of the function to be spliced.
Now, the listeners have incrementing integer ids to be able to identify and remove them correctly.

### DIFF
--- a/apps/desktop/src/lib/shortcuts/shortcutService.svelte.ts
+++ b/apps/desktop/src/lib/shortcuts/shortcutService.svelte.ts
@@ -1,18 +1,25 @@
 import type { Tauri } from '$lib/backend/tauri';
 
+type ShortcutListener = {
+	id: number;
+	shortcut: string;
+	callback: () => void;
+};
+
 /**
  * Service class for listening to shortcut events from the back end.
  */
 export class ShortcutService {
-	private listeners: [string, CallableFunction][] = [];
+	static idCounter = 0;
+	private listeners: ShortcutListener[] = [];
 	constructor(private tauri: Tauri) {}
 
 	listen() {
 		$effect(() =>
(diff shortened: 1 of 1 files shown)


apps/desktop/src/lib/shortcuts/shortcutService.svelte.ts
```diff
@@ -1,29 +1,41 @@
...
 
 	on(shortcut: string, callback: () => void) {
+		const id = ShortcutService.idCounter;
+		ShortcutService.idCounter += 1;
+		const listener = { id, shortcut, callback };
 		$effect(() => {
-			this.listeners.push([shortcut, callback]);
+			this.listeners.push(listener);
 			return () => {
-				this.listeners.splice(this.listeners.findIndex(callback), 1);
```
estib-vega, line 26: `Array.prototype.findIndex` expects a function predicate. Passing the callback here was effectively calling the callback function for each listener

apps/desktop/src/lib/shortcuts/shortcutService.svelte.ts
```diff
@@ -1,29 +1,41 @@
...
+						listener.callback();
 					}
 				}
 			})
 		);
 	}
 
 	on(shortcut: string, callback: () => void) {
+		const id = ShortcutService.idCounter;
+		ShortcutService.idCounter += 1;
```
estib-vega, line 31: Not necessarily convinced that an integer id is optimal, open to other opinions :) @mtsgrd

ndom91:
We all ran into this shortcut calling in the office too and were super confused for a minute haha. Caleb merged a fix last night though, so not sure if this is necessary anymore

PR #6090, opened 2025-01-27
Handle commiting to workspace correctly

Caleb-T-Owens:


crates/but-debugging/src/lib.rs
```diff
@@ -0,0 +1,109 @@
+use std::{path::Path, process::Command};
```
Caleb-T-Owens, line 1: Added some debugging utils that let you get a bit of a better idea of the git graph and trees.

crates/gitbutler-tauri/src/stack.rs
```diff
@@ -134,5 +135,6 @@ pub fn push_stack_to_review(
     let ctx = CommandContext::open(&project, settings.get()?.clone())?;
     gitbutler_sync::stack_upload::push_stack_to_review(&ctx, &user, stack_id)?;
 
+    emit_vbranches(&windows, project_id);
```
Caleb-T-Owens, line 138: This is an unrelated but important change that propagates newly assigned review_ids to the frontend

PR #6100, opened 2025-01-28
Patch: Update the review type

estib-vega:
Update the review properties of patch to match the new user information returned by the API

Depends on https://github.com/gitbutlerapp/web/pull/460

### DIFF
--- a/packages/shared/src/lib/branches/types.ts
+++ b/packages/shared/src/lib/branches/types.ts
@@ -271,6 +271,10 @@ export async function getPatchViewersWithAvatars(patch: Patch) {
 	return await getUsersWithAvatars(patch.review.viewed);
 }
 
+export async function getPatchViewersAllWithAvatars(patch: Patch) {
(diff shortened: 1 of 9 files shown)


packages/shared/src/lib/branches/types.ts
```diff
@@ -135,23 +135,49 @@ export function apiToPatchStatistics(api: ApiPatchStatistics): PatchStatistics {
...
+	name: string | null;
+};
+
+export type PatchReviewUser = {
+	id: number;
+	avatarUrl: string | undefined;
+	email: string | undefined;
+	login: string | undefined;
+	name: string | undefined;
+};
```
Caleb-T-Owens: Rather than calling this a PatchReviewUser, we should properly match the SimpleUser > User > PrivateUser higherarchy from the backend

estib-vega: Done. I created a central type under the `users` directory

PR #6102, opened 2025-01-28
feat: reposition darwin traffic-lights in title bar

ndom91:
## 🧢 Changes

- Test Mac OS title bar traffic lights repositioning
- Based off of this plugin + PR for Tauri v2 (https://github.com/ItsEeleeya/tauri-plugin-trafficlights-positioner/pull/4/files)

### DIFF
--- a/crates/gitbutler-tauri/src/main.rs
+++ b/crates/gitbutler-tauri/src/main.rs
@@ -51,6 +51,20 @@ fn main() {
                     )
                     .expect("Failed to create window");
 
+                    #[cfg(target_os = "macos")]
+                    use tauri::LogicalPosition;
+                    #[cfg(target_os = "macos")]
+                    use tauri_plugin_traffic_lights::WindowExt;
+                    #[cfg(target_os = "macos")]
+                    // NOTE: Make sure you only call this ONCE per window.
+                    {
+                        if let Some(window) = tauri_app.get_window("main") {
+                            #[cfg(target_os = "macos")]
+                            // NOTE: Make sure you only call this ONCE per window.
+                            window.setup_traffic_lights_inset(LogicalPosition::new(18.0, 22.0))?;
+                        };
+                    }
+
                     // TODO(mtsgrd): Is there a better way to disable devtools in E2E tests?
                     #[cfg(debug_assertions)]
                     if tauri_app.config().product_name != Some("GitButler Test".to_string()) {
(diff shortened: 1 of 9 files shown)


Byron:
With the fix I pushed it compiles and does seem to have an effect.

Screenshot 2025-01-29 at 09 17 39

Byron:
CI still fails with terrible build errors on Linux it seems, an issue that isn't reproducible on MacOS.

Does the following reproduce the issue locally on Linux?

```
cargo check --workspace --all-targets
```

ndom91:
> CI still fails with terrible build errors on Linux it seems, an issue that isn't reproducible on MacOS.
> 
> Does the following reproduce the issue locally on Linux?
> 
> ```
> cargo check --workspace --all-targets
> ```

Yeah so locally here on my linux box I get the same error messages as in CI :sob: 

image

ndom91:
Thanks for the `Cargo.toml` fixes @Byron. @PavelLaptev feel free to merge this whenever you're happy with the positioning :+1:

PavelLaptev:
@Byron can we merge this?

Byron (approved):
Thanks, it looks good to me and I re-validated it after making some final adjustments.

crates/gitbutler-tauri/src/main.rs
```diff
@@ -132,6 +131,25 @@ fn main() {
...
+                        #[cfg(target_os = "macos")]
+                        // NOTE: Make sure you only call this ONCE per window.
+                        {
+                            if let Some(window) = tauri_app.get_window("main") {
+                                #[cfg(target_os = "macos")]
+                                // NOTE: Make sure you only call this ONCE per window.
+                                window
+                                    .setup_traffic_lights_inset(LogicalPosition::new(16.0, 25.0))?;
+                            };
+                        }
```
mtsgrd, line 149: This is only called _once_ for the app atm, and doesn't reposition new lights in new windows. Also, it applies the same positioning to V2. Let's fix before we need to make a next release!

ndom91: @Byron looks like there was a reason for the 2nd call

Byron: I think the issues @mtsgrd is describing would be true in multi-window mode and when V3 and V2 is switched while the app is running? 
Indeed it's made to only work if the app launches in V3 mode and it's assumed to only have one Window - multi-window support was never made official, and from what I remember it's currently hanging/deadlocking on Windows even. When V3 is switched to V2, the app needs a restart for the buttons to appear in the right spot.

Should there be a follow-up that makes this multi-window compatible?

> @Byron looks like there was a reason for the 2nd call

Maybe, but if it wasn't added by you or me, who would know about it?

PR #6113, opened 2025-01-29
fix-leaving-edit-mode

Caleb-T-Owens:


crates/but-debugging/src/lib.rs
```diff
@@ -129,3 +129,19 @@ pub fn git_ls_tree(path: &Path, options: &LsTreeOptions) {
     println!("{:?}", command);
     println!("{}", std::str::from_utf8(&result).unwrap());
 }
+
+/// Runs `git status` and passes the output to println!
+#[deprecated = "git_status should not be used in production code or testing infastructure"]
+pub fn git_status(path: &Path) {
```
Byron, line 135: I think executing a script with `sh` in a context would be more powerful than hand-wrapping. This could be an API like `but_debugging::shell(cwd).run("git status").run("something else")`.

crates/gitbutler-edit-mode/src/lib.rs
```diff
@@ -236,6 +237,12 @@ pub(crate) fn save_and_return_to_workspace(
     // Recommit commit
     // While we perform hard resets we should pick up everything to avoid loosing worktree state.
     let pick_up_untracked_files_of_any_size = 0;
+
+    // Make sure any conflicted entries are added to the index
+    let mut index = repository.index()?;
+    index.add_all(["*"], IndexAddOption::FORCE, None)?;
```
Byron: I think this would be too dangerous given that we are only interested in conflicted files to be added to the index (in order to forcefully resolve the conflict).

Unfortunately, the flag isn't documented but when looking [here](https://github.com/libgit2/libgit2/blob/c86842f0a518cc15d7964027d01a4ae3bf9604db/src/libgit2/index.c#L3586-L3593) and [here](https://github.com/libgit2/libgit2/blob/c86842f0a518cc15d7964027d01a4ae3bf9604db/src/libgit2/index.c#L3685-L3687) it seems it will add untracked files as well as ignored ones.

Instead I think it should find conflicting entries and add them one by one.

crates/gitbutler-edit-mode/tests/fixtures/edit_mode.sh
```diff
@@ -0,0 +1,45 @@
...
+# Where "left" and "right" contain changes which conflict with each other
+git clone repo conficted_entries_get_written_when_leaving_edit_mode
+(cd conficted_entries_get_written_when_leaving_edit_mode
+  git checkout -b left
+  echo left > conflict
+  git add . && tick && git commit -m "left"
+  git checkout main
+  git checkout -b right
+  echo right > conflict
+  git add . && tick && git commit -m "right"
```
Byron: Is `tick` required for the  test to work? I'd think that timestamps don't matter.

Byron (commented):
Thanks, I love the test!

crates/gitbutler-edit-mode/src/lib.rs
Caleb-T-Owens: Thanks

PR #6121, opened 2025-01-30
feat: add support for keyboardShortcut prop in ContextMenuItem

ndom91:
## 🧢 Changes

- Add a `keyboardShortcut` prop to `ContextMenuItem.svelte` in order to render things like this:

image

- Resulting `ContextMenuItem.svelte`:

image

### DIFF
--- a/packages/ui/src/lib/utils/url.ts
+++ b/packages/ui/src/lib/utils/url.ts
@@ -1,0 +1,13 @@
+export function determinePlatform(userAgent: string) {
+	const ua = userAgent.toLowerCase();
+
+	if (ua.includes('mac os')) {
+		return 'macos';
+	} else if (ua.includes('windows')) {
+		return 'windows';
+	} else if (ua.includes('linux')) {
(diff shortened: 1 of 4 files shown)


PavelLaptev (approved):
LGTM!

packages/ui/src/lib/utils/url.ts
```diff
@@ -0,0 +1,13 @@
...
+	if (ua.includes('mac os')) {
+		return 'macos';
+	} else if (ua.includes('windows')) {
+		return 'windows';
+	} else if (ua.includes('linux')) {
+		return 'linux';
+	}
+
+	return 'unknown';
+}
```
mtsgrd, line 13: This file looks incorrectly named. Let's change it to platform.ts?

ndom91: Yeah the thinking was that we're getting this out of the request headers, but `url` isn't even a good name for that either. Changed.

PR #6122, opened 2025-01-30
feat: add button for overflow menu in stack tab

ndom91:
## 🧢 Changes

- Create stackTab overflow ContextMenu only visible on hover/focus of tab
- Todo in follow-up PR(s):
	- Figure out which actions we want in the stack tab overflow menu
	- Decide on shortcuts / implement the handlers for them

image

### DIFF
--- a/apps/desktop/src/components/StackTab.svelte
+++ b/apps/desktop/src/components/StackTab.svelte
@@ -1,8 +1,12 @@
 <script lang="ts">
 	import { stackPath } from '$lib/routes/routes.svelte';
+	import Button from '@gitbutler/ui/Button.svelte';
+	import ContextMenuSection from '@gitbutler/ui/ContextMenuSection.svelte';
 	import Icon from '@gitbutler/ui/Icon.svelte';
 	import type { Tab } from '$lib/tabs/tab';
 	import { goto } from '$app/navigation';
+	import ContextMenu from '@gitbutler/ui/ContextMenu.svelte';
+	import ContextMenuItem from '@gitbutler/ui/ContextMenuItem.svelte';
 
 	type Props = {
 		projectId: string;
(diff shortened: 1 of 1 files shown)


apps/desktop/src/components/StackTab.svelte
```diff
@@ -36,6 +45,47 @@
 	<div class="name">
 		{tab.name}
 	</div>
+	<div
+		class={['tab__overflow-menu', isContextMenuOpen || isHovered ? 'active' : '']}
```
mtsgrd, line 49: For style reasons, could we write this line as follows?
```svelte
<div
  class:overflow-menu={isMenuOpen}
  class:active={isHovered}
  ...
/>
```

apps/desktop/src/components/StackTab.svelte
```diff
@@ -36,6 +45,47 @@
...
+	>
+		<Button
+			kind="ghost"
+			icon="kebab"
+			bind:el={kebabMenuTrigger}
+			onclick={() => {
+				contextMenuEl?.toggle();
+			}}
+			activated={isContextMenuOpen}
+		></Button>
```
mtsgrd, line 62: Let's use a self closing tag.

apps/desktop/src/components/StackTab.svelte
```diff
@@ -54,6 +104,27 @@
...
+			width: 0px;
+			pointer-events: none;
+			transition:
+				opacity 100ms ease-in,
+				width 100ms ease-in-out;
+		}
+
+		.tab__overflow-menu.active,
+		&:active .tab__overflow-menu,
+		&:hover .tab__overflow-menu,
```
mtsgrd, line 119: Why `:hover` when you are setting it manually with mouse enter/leave?

apps/desktop/src/components/StackTab.svelte
```diff
@@ -54,6 +104,27 @@
...
+		&:focus-within .tab__overflow-menu,
+		&:focus .tab__overflow-menu {
+			pointer-events: auto;
+			margin-right: 8px;
+			display: flex;
```
mtsgrd, line 124: This `display: flex` should probably go into `tab__overflow-menu`. It's unclear why it would need to change on active/hover/focus.

PR #6125, opened 2025-01-30
fix: ChromeHeader absolute positioned element bug

ndom91:
## 🧢 Changes

- Globally centered project title element was blocking mouse interaction on other btns in chrome header

### DIFF
--- a/apps/desktop/src/components/ChromeHeader.svelte
+++ b/apps/desktop/src/components/ChromeHeader.svelte
@@ -124,10 +124,15 @@
 
 	.center {
 		position: absolute;
-		left: 0;
 		width: 100%;
+		left: 0;
 		display: flex;
 		justify-content: center;
+		pointer-events: none;
+
+		& * {
+			pointer-events: auto;
+		}
 	}
 
 	.selector-series-select {


apps/desktop/src/components/ChromeHeader.svelte
```diff
@@ -124,10 +124,15 @@
...
-		left: 0;
 		width: 100%;
+		left: 0;
 		display: flex;
 		justify-content: center;
+		pointer-events: none;
+
+		& * {
+			pointer-events: auto;
+		}
```
mtsgrd, line 135: What does this fix?

ndom91: Looked closer into this one, and it turns out it was necessary actually. 

Although the direct child of that `.center` class div is a svelte component, that component then has more children defined within this `ChromeHeader.svelte` file which do get this `& * { pointer-events: auto; }` applied to them. Specifically the `.selector-series-select` div in the `customSelectButton()` snippet.

ndom91: However, looks like this manner of aboslutely centering the element works as well. The element is then only the width of the content, not the whole viewport, and uses auto margins to center itself and therefore doesn't need any of the `pointer-events` tomfoolery that was necessary when the `.center` element was full width: https://github.com/gitbutlerapp/gitbutler/pull/6234

PR #6126, opened 2025-01-30
fix universal-diff generation

Byron:
Previously it added extra newlines, let's not do this anymore.


crates/but-core/tests/core/diff/worktree_changes.rs
```diff
@@ -798,7 +798,7 @@ fn symlink_to_file_in_index() -> Result<()> {
                     old_lines: 1,
                     new_start: 1,
                     new_lines: 1,
-                    diff: "@@ -1,1 +1,1 @@\n-target\n+content\n\n",
+                    diff: "@@ -1,1 +1,1 @@-target+content\n",
```
Caleb-T-Owens, line 801: I'm not sure that this looks correct.

Byron: Thanks for helping with this! This should now be addressed by the follow up at #6129.

PR #6127, opened 2025-01-30
feat: initial getStackBranches stackService implementation

ndom91:
## 🧢 Changes

- [x] Add `stack_branches` StackService query
- [x] Add initial UI showing branch lists
- [x] Add workspace page UI building blocks

image


apps/desktop/src/components/Chrome.svelte
```diff
@@ -30,6 +30,7 @@
 		overflow: hidden;
 		flex-grow: 1;
 		height: 100%;
+		max-width: 100vw;
```
mtsgrd: Just a nit, but generally I avoid using `vw` since it doesn't take into account available space, but is absolute to the inner width of the browser window. It means e.g. if you suddenly have a scrollbar then it doesn't resize to account for that.

apps/desktop/src/components/StackDetails.svelte
```diff
@@ -0,0 +1,61 @@
...
+				{@const last = i === branches.length - 1}
+				<Branch {branch} {first} {last} />
+			{/each}
+		</div>
+
+		<div class="stack__branch-content">
+			<StackContentPlaceholder />
+		</div>
+	{/if}
+</div>
```
mtsgrd: Instead of checking if branches is defined, can you look at how I used `ReduxResult` and use that here as well?

apps/desktop/src/components/StackTab.svelte
```diff
@@ -14,14 +14,15 @@
...
+		tabOverflow?: boolean;
 	};
 
 	let kebabMenuTrigger = $state<HTMLElement>();
 	let contextMenuEl = $state<ContextMenu>();
 	let isContextMenuOpen = $state(false);
 	let isHovered = $state(false);
 
-	const { projectId, tab, first, last, selected }: Props = $props();
+	const { projectId, tab, first, last, selected, tabOverflow = false }: Props = $props();
```
mtsgrd: Since we're in a component called tab, maybe name this `overflow`.

apps/desktop/src/components/v3/Branch.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="ts">
+	import BranchCommitList from '$components/v3/BranchCommitList.svelte';
+	import BranchHeader from '$components/v3/BranchHeader.svelte';
+	import type { WorkspaceBranch } from '$lib/branches/v3';
+
+	interface Props {
+		branch: WorkspaceBranch;
+		first: boolean;
+		last: boolean;
```
mtsgrd, line 9: Looks unused since it's not declared in this file.

apps/desktop/src/components/StackContentPlaceholder.svelte
```diff
@@ -0,0 +1,201 @@
...
+
+<div class="placeholder">
+	<div class="placeholder__top">
+		<svg
+			width="300"
+			height="220"
+			viewBox="0 0 300 220"
+			fill="none"
+			xmlns="http://www.w3.org/2000/svg"
+		>
```
mtsgrd: Since this is a rather big svg, can we save it as such and use it like in `WorktreeChanges.svelte`?

apps/desktop/src/components/Chrome.svelte
ndom91: Gotcha. In this case it doesn't look like its actually necessary anymore anyway. Removed.

apps/desktop/src/components/StackTab.svelte
ndom91: Makes sense. This was used in my implementation to be able to toggle back n forth between two different tab states - we weren't sure at the time which we wanted to go with and I kept both available to be able to experiment and see how each looked. One where the tabs overflow (or underflow I guess) under a pinned-to-the-right "+" button and the left border, and another state where the content of the tabs just shrink on smaller viewports.

Therefore, since we've decided to go with the underflow option and you've cleaned that all up - this prop / option isn't necessary anymore. Removed :+1:

apps/desktop/src/components/v3/Branch.svelte
ndom91: Correct, not used yet but the v2 branch list (`SeriesList.svelte`) had a `isBottomBranch` prop we applied to empty series (not sure off the top of my head what the exact use-case was), so I kept it here as well.

I can remove it now though and just re-add it again if it turns out that we need it

apps/desktop/src/components/StackContentPlaceholder.svelte
ndom91: Yup good call

apps/desktop/src/components/StackTabNew.svelte
```diff
@@ -13,10 +13,10 @@
 
 	type Props = {
 		projectId: string;
-		selectedId: string | undefined;
+		tabOverflow?: boolean;
 	};
 
-	const { projectId }: Props = $props();
+	const { projectId, tabOverflow = false }: Props = $props();
```
mtsgrd, line 19: It feels a bit like stuttering when we repeat part of the name of the component. Can we call this prop `overflow`, like the css class it controls?

packages/ui/src/lib/link/Link.svelte
```diff
@@ -67,14 +72,18 @@
...
 		user-select: text;
+		text-decoration: none !important;
 
 		&:hover {
 			text-decoration: none;
 		}
 	}
 
+	.link.underline {
+		text-decoration: underline !important;
```
mtsgrd, line 84: Just leaving a note here, let's avoid `!important` unless absolutely necessary.

apps/desktop/src/components/StackTabNew.svelte
ndom91: Merged :+1:

packages/ui/src/lib/link/Link.svelte
ndom91: Merged fix :+1:

PR #6136, opened 2025-01-31
colocate UI types and conversions

Byron:
Instead of dumping all UI related types and conversions into the `tauri` crate, let's put them into `ui` modules in their respective crates, close to the types they are mirroring.

This makes the `tauri` crate easier to work with and, as it turns out, also fixes the 'rebuild issue' as some dependencies could be removed.

### Tasks

* [x] move hunk-dependency
* [x] move diff


mtsgrd:
I heard back from Daniel (Tauri), he also recommends adding a `.taurignore` file, if you care to try it out? https://tauri.app/develop/#reacting-to-source-code-changes

mtsgrd (approved):

Byron:
I haven't been using the `tauri dev` mode at all recently, but can say that with this PR, building tests for the `gitbutler-tauri` crate (and building it in general) will not rebuild unnecessary dependencies anymore.

From my point of view this fixes the issue.

Using `.tauriignore` maybe useful to prevent spurious rebuilds that you have been encountering, and I will submit one in a follow-up PR (it's easy enough to setup).

Byron:
@krlvi When looking at the module paths for the UI types I have a good feeling, and am thinking it's the way to go. For comparison:

* `but_core::TreeChange`
* `but_core::ui::TreeChange`

The latter is used in the `gitbutler-tauri` crate and it's easy enough to get to the plumbing type by removing `ui::`. This property would be missing if `but-ui` would have been used.

krlvi:
> @krlvi When looking at the module paths for the UI types I have a good feeling, and am thinking it's the way to go. For comparison:
> 
>     * `but_core::TreeChange`
> 
>     * `but_core::ui::TreeChange`
> 
> 
> The latter is used in the `gitbutler-tauri` crate and it's easy enough to get to the plumbing type by removing `ui::`. This property would be missing if `but-ui` would have been used.

yes this is exactly what i meant. Btw feel free to assign me as reviewer next time so that I can see a notification (otherwise i miss it)

PR #6146, opened 2025-02-01
Change default context menu positioning

PavelLaptev:
All context menus had wrong positioning.

before:

image

after:

image


PavelLaptev:
@ndom91 I've updated the default context menu position; it should now appear at the bottom by default.

PR #6147, opened 2025-02-01
add `.taurignore` to ignore everything *but* the crates.

Byron:
The dev-mode of tauri is responsible for watching for changes to the backend code to rebuild and restart the binary.

Let's guide it to avoid spurious rebuilds if something unrelated changes.

Follow-up of https://github.com/gitbutlerapp/gitbutler/pull/6136#issuecomment-2628887841


Byron:
I tested this with `tauri dev` and it worked as expected.

mtsgrd (approved):

PR #6169, opened 2025-02-04
basic committing

Byron:
This PR introduces commit-mutation machinery that over time will evolve to handle all changes to commits that the UI could trigger.

For now this only involves creating full files, but the API is also sketched out to deal with hunks in a follow-up PR.

### Tasks

* [x] API to create a commit on top of the stack for full files
* [x] optional commit-signing based on GitButler settings
* [x] the first simple commit
* [x] Show whole-file application with difficult hunks with merge-commit
* [x] retry-support in case of conflicts
* [x] sketch how this would be called by the UI and show a JSON sample.
    - Don't expose the low-level engine, but provide individual commands for each capability.
* [x] make sure there are no hard assertions, make these errors so frontend has better errors
* [x] easily obtain changed tree in preparation for preparing an index for hooks
* [x] select patches to commit
* [x] tests for executable and symlinks, and changing file types
* [x] try submodule changes and apply them (also type changes)
* [x] tests for renames (hunks and whole file)
* [x] assure conflict-fallback test also shows it didn't remove files that weren't conflicting
     - [x] make sure this isn't an off-by-one - it's not - probably something about the example that is confusing.

### For follow-up PRs

**Remember: Frontend has 3 context-lines, and if there are any issues with this, we should try zero context lines**

In any order (probably)

* sparse repository must be excluded.
* set index to match new commit
* commit-signing tests
* commit *into* another commit, i.e. amend
* rebasing integration

### Further Functionality (can be done with old API)

* apply and unapply virtual branches

### Out of scope

* **hunk-dependencies**
* **atomic object writes**


Byron:
### Rebase Engine

* `reorder_stack()`
* reordering as special case of rebase
* empty branches point to the base
* intrinsic order of refs not good enough, as they might point tot he same commit

##### On conflict when recreating the top-level merge

* TBC

krlvi (approved):

PR #6195, opened 2025-02-06
Upgrade vite from v5 -> v6

mtsgrd:
- invalid "es2015" value in resolve section of config
- target should also be "modules"
- sync ui package target with rest at es2022


mtsgrd:
I had just missed committing a section of the lock file!

mtsgrd:
Fixes our problem mentioned in https://github.com/getsentry/sentry-javascript/issues/14516.

PR #6199, opened 2025-02-06
fix: header padding for platforms other than Mac OS

ndom91:
## 🧢 Changes

- Additional padding was added globally, however, this is only required when running on Mac OS and we place the traffic-lights in that corner

### DIFF
--- a/apps/desktop/src/components/ChromeHeader.svelte
+++ b/apps/desktop/src/components/ChromeHeader.svelte
@@ -34,7 +34,7 @@
 	let isNotificationsUnread = $state(false);
 </script>
 
-<div class="header">
+<div class="header" class:isMac={platformName === 'macos'}>
 	<div class="left">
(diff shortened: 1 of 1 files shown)


apps/desktop/src/components/ChromeHeader.svelte
```diff
@@ -116,7 +116,11 @@
 	.header {
 		display: flex;
 		justify-content: space-between;
-		padding: 14px 14px 14px 84px;
+		padding: 14px;
+
+		&.isMac {
+			padding: 14px 14px 14px 84px;
+		}
```
mtsgrd, line 123: Incorrect casing of class name, let's call it `mac`?

ndom91: Fixed in https://github.com/gitbutlerapp/gitbutler/pull/6234

PR #6203, opened 2025-02-06
feat: css gap utils

ndom91:
## 🧢 Changes

- Add `gap` utility classes
- @PavelLaptev do you know a better way to do this? To avoid using `!important` on these util classes?
	- The svelte component `Link` has a predefined `gap: 2px` in it's scoped styles for the class `link`. Which oddly also comes _before_ our prop  class,  so you'd think our prop classes would override it (i.e. something like `class="link ${classes}"`), but I guess it wins on specificity due to the additional svelte scoping auto-generated class name :shrug:

### DIFF
--- a/apps/desktop/src/components/StackContentPlaceholder.svelte
+++ b/apps/desktop/src/components/StackContentPlaceholder.svelte
@@ -112,7 +112,7 @@
 				target="_blank"
 				underline={false}
 				externalIcon={false}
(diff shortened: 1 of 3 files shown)


mtsgrd:
Following from our conversation earlier, can we try and accomplish this without `!importrant`?

Caleb-T-Owens:
This has been implemented separately :+1:

PR #6231, opened 2025-02-07
fix: simplify `StackTab.svelte`

ndom91:
## 🧢 Changes

- Simplify `StackTab.svelte`
- Remove unnecessary css and manual contextMenu state tracking


ndom91:
The scroller shadows still look pretty weird with this particular `linear-gradient` and `left: 0` / `right: 0` :thinking: I'll take a look at this in a follow-up 

image

apps/desktop/src/components/v3/StackTab.svelte
```diff
@@ -14,69 +14,65 @@
...
 
-	const { projectId, tab, first, last, selected }: Props = $props();
+	let nameEl = $state<HTMLDivElement>();
+	let nameWidth = $state<number>();
+
+	$effect(() => {
+		if (nameEl) {
+			nameWidth = nameEl.offsetWidth - 1;
+		}
+	});
```
mtsgrd, line 29: Hmm, I redid the tabs just yesterday and made it css only. What is the issue you are solving here? *edit: for reference: https://github.com/gitbutlerapp/gitbutler/commit/2a6a143cdbf5b830bfe92332010263163989ad05

apps/desktop/src/components/v3/StackTab.svelte
```diff
@@ -104,73 +101,64 @@
...
-		&::after {
-			transform: translateY(0);
-			background: var(--clr-theme-pop-element);
-			z-index: var(--z-ground);
+		&.selected {
+			&::after {
+				transform: translateY(0);
+				background: var(--clr-theme-pop-element);
+				z-index: var(--z-ground);
+			}
```
mtsgrd, line 161: I removed the css nesting for all these classes just yesterday since it doesn't really serve a useful purpose. How come you're adding it back in?

PR #6234, opened 2025-02-07
fix: absolutely center project select without need for pointer-events overrides

ndom91:
## 🧢 Changes

- Absolutely center the project selector in the chrome header in such a way that the element is only as large as its content and leverages auto margins for centering. 
	- This avoids the element being the full width of the viewport and therefore needing to hackaround it covering other buttons with `pointer-events`


mtsgrd:
@ndom91 could you check the commit I pushed and let me know what you think? The advantage is that there is no absolute positioning, shrinking as needed, and no overflow.

PR #7279, opened 2025-02-10
Chat message: Quote code

estib-vega:
This introduces the ability to select some lines from the patch diff hunks, and quote it in a message.

Introduces:
- HunkDiff selection handling (click toggles and drag selection)
- Chati input quote preview (display what part of the diff is being quoted)
- Propagate the diff selection to the parametes for sending a message
- Parse out the diff quote attributes from the message and display them

viz:

https://github.com/user-attachments/assets/68e3e5a6-cc9f-4d6c-ac22-5a69dc5b20e6

---
- <kbd>&nbsp;2&nbsp;</kbd> #7279 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #7274

### DIFF
--- a/apps/web/src/lib/chat/diffPatch.ts
+++ b/apps/web/src/lib/chat/diffPatch.ts
@@ -1,0 +1,63 @@
+import { SectionType, type ContentSection, type Line } from '@gitbutler/ui/utils/diffParsing';
+import type { DiffPatch } from '@gitbutler/shared/chat/types';
+
+function getSectionType(line: DiffPatch): SectionType {
+	switch (line.type) {
+		case 'added':
+			return SectionType.AddedLines;
+		case 'removed':
+			return SectionType.RemovedLines;
+		case 'context':
+			return SectionType.Context;
+	}
+}
+
+function cleanDiffLine(line: string, sectionType: SectionType): string {
+	if (sectionType === SectionType.AddedLines && line.startsWith('+')) {
+		return line.slice(1);
+	}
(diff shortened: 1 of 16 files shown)


estib-vega:
@PavelLaptev Sorry for the late tagging

The code changes are done from my side, there are only two main things that need to be style-vetted:

### 1: The selection state in the hunk diff

Screenshot 2025-02-10 at 20 07 00

### 2: The 'preview' of the quote.
This is shown in the input field to signal that there is code that will be quoted in the message that's about to be sent

Screenshot 2025-02-10 at 20 07 33

PavelLaptev:
Thank you for the PR, @estib-vega!  

I've added the code quotation design [here in Figma](https://www.figma.com/design/hrEqnFaxS6mDPBqKfIOOZg/Web-Client?node-id=1418-36223&t=6dWB839UXlFoTxdZ-1).  

#### Key edits:  

1. Rows are selected only when clicking on the count columns, not the entire row.  
2. The code quote box appears in the dialog only when clicking the **"quote"** button. The second button is just for **copying** the code.  

   image  

3. Selection should be indicated with an overlay box.  
4. Selection should support using the **Shift** key.  
5. It would be nice to support selection using **keyboard arrows**.  
6. Code quotes should be removable from the dialog box using a **"cross"** button.  

   image  

---

Let me know if you have any further questions! 🚀

estib-vega:
@PavelLaptev 
Thanks for the review! I'll do the changes 👍

PR #7293, opened 2025-02-15
Fix for renamed crate in rand 0.9.0 update

mtsgrd:


mtsgrd:
Fixes broken nightly: https://github.com/gitbutlerapp/gitbutler/actions/runs/13341176108/job/37265903857

mtsgrd:
@krlvi any idea why our tests didn't catch this?

krlvi:
> @krlvi any idea why our tests didn't catch this?

Yeah. Neither tests nor linting is running for this crate because of how it's setup. We can easily fix that by renaming the crate to the `gitbutler-` but I would like to understand first on Monday the story behind it. 

Dont have the full context here but it seems like we should have forked `clearlysid/tauri-plugin-decorum/` instead of vendoring it. It's not immediately obvious what modifications we made, and we also lose the ability to rebase from it's upstream in the future

PR #7295, opened 2025-02-16
commit and amend for in `tauri`

Byron:
Try to bring both commit and amend all the way into the frontend with an API and implementation that could actually work start to end.

Follow-up on #7287.

### Tasks

* [x] refactor
* [x] support for rewrites with change-ids
* [x] fix/port handling of obtaining commits for stacks so it deals with merged target branches
* [x] Assure that commit and author dates are updated for merge-commits and cherry-picks, and committer as well!
* [x] update tauri integration to be able to create a complete commit at any position (reachable) in the graph.
* [x] CLI verification of amending
* [x] tauri integration for amending commits

### For next PR

* [ ] index-removal via flag (!). Maybe a test for that as well.
* [ ] put all other logic (Vbranch-IO) into `but-workspace`
* [ ] settings for context-lines
* [ ] port cherry_rebase_group to `gix` fully (key is the auto-conflict handling) - needed to do allow in-memory config adjustments
* [ ] insert-empty-commit as feature of `but-rebase`
    - [ ] move ref-rewriting there to support direct callers of `but-rebase`, re-order, etc.

* empty commits aren't specifically highlighted, or handled or prevented. The UI could detect that and show them.
* Workspace commits with a single branch will get signed. This shouldn't be a real problem, and ideally it will go away soon enough.

### Notes

* Rebasing keeps existing headers like change-ids, but actually testing it is cumbersome but probably should be done nonetheless (needs to avoid 'stabilizing' change-ids).

### For follow-up PRs

**Remember: Frontend has 3 context-lines, and if there are any issues with this, we should try zero context lines**

In any order (probably)

* re-establish the worktree-lock in `tauri` for worktree-mutating operations.
    - read-only operations will probably be fine though, even though a read-lock would be required for correctness.
* commit-signing tests
* rebasing integration

### Further Functionality (can be done with old API)

* apply and unapply virtual branches

### Out of scope

* **hunk-dependencies**
* **atomic object writes**


krlvi:
can we put UNIDIFF_CONTEXT_LINES to gitbutler-settings

Byron:
> can we put UNIDIFF_CONTEXT_LINES to gitbutler-settings

A good point! I will do that in a follow-up PR.

PR #7299, opened 2025-02-16
Fix commit bug

mtsgrd:
Regression in pr #7286 caused all files to be committed regardless of selection.


krlvi:
@mtsgrd thanks for the fix - this was a total brainfart on my part

PR #7317, opened 2025-02-17
Fix traffic lights customizer import

mtsgrd:


krlvi:
solved without re-pinning all dependencies here https://github.com/gitbutlerapp/gitbutler/pull/7319

PR #7323, opened 2025-02-17
Change "Open in VS Code" to "Open in Editor" in drop-down menu

### LINKED ISSUE #7318: Hardcoded "Open in VS Code" while other editors are already available
Line:

https://github.com/gitbutlerapp/gitbutler/blob/80b369c4cef51d1c703a85734d9057078b5598f8/crates/gitbutler-tauri/src/menu.rs#L174

Potentially could be "Open in <whatever is configured", but can also be changed to "Open in editor" for now.

marinegor:
## 🧢 Changes
Change "Open in VS Code" to "Open in Editor" in drop-down menu

## ☕️ Reasoning
VS Code isn't the only editor GitButler can work with

## 🎫 Affected issues

Fixes: #7318


Byron (approved):
Thanks also lot for your help. Great catch!

PR #7325, opened 2025-02-18
feat: add scroll-snap and svelte-keepfocus to StackTabs

ndom91:
## 🧢 Changes

- Added scroll-snap to tabs
- Added `data-svelte-keepfocus` attribute to tab 
  - If folks are navigating via keyboard and press "Enter" on a tab, to navigate to that route, normally it would put focus back tot he top of the dom tree, but this keeps it on the selected tab, figured this is closer to user expectations and would be helpful later on with our focus/select management.
- Cleanup some of the existing code

image

image

### DIFF
--- a/apps/desktop/src/components/v3/StackTab.svelte
+++ b/apps/desktop/src/components/v3/StackTab.svelte
@@ -14,14 +14,15 @@
 		selected: boolean;
 	};
 
-	let kebabMenuTrigger = $state<HTMLElement>();
-	let contextMenuEl = $state<ContextMenu>();
-	let isContextMenuOpen = $state(false);
-
 	const { projectId, tab, first, last, selected }: Props = $props();
+
+	let kebabMenuTrigger = $state<HTMLElement>();
+	let contextMenu = $state<ContextMenu>();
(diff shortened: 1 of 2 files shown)


apps/desktop/src/components/v3/StackTab.svelte
```diff
@@ -93,6 +90,8 @@
 		border-right: 1px solid var(--clr-border-2);
 		overflow: hidden;
 		min-width: 100px;
+		flex: 0 0 auto;
```
mtsgrd, line 93: Here you are saying that tabs should not resize, which is not what we want. Can you drop this line?

mtsgrd (approved):
lgtm

mtsgrd:
image

Also, any chance you could figure out what the issue is with this shadow on linux?

apps/desktop/src/components/v3/StackTab.svelte
ndom91: Yeah so this was actually necessary, but looking at it again, only the second part - the shrink: 0 shorthand, was the important part :+1: 

If you look at the screenshots in the desc above it still grew/grows.

PR #7329, opened 2025-02-18
Refactor StackBranch

krlvi:
Make StackBranch setting of head and name go through a method - this will allow us to set/update real references in a controlled way.


Byron (approved):
LGTM

PR #7330, opened 2025-02-18
workspace update

Byron:
Make workspace updates possible, i.e. rebasing multiple branches in a workspace onto a new base.

Follow-up on #7315.

### Tasks

* [ ] most complex test-case
* [ ] multi-base support in rebase engine so it won't always builds a linear sequence
* [ ] capture worktree changes
* [ ] applied worktree changes onto new merge-commit, resolve conflicts to 'ours', reset worktree to match
* [ ] more testing probably
* [ ] see if merge-commits are detectable so we can prevent rebasing these
    - [ ] abort if a branch is meant to be merged
* [ ] what happens if the workspace commit cannot be remerged?

### For follow-up PRs

In any order (probably)

* Apply Branch/Unapply Branch
    - Don't rebase newly applied branch, just merge it in. 
    - Rebasing only when explicit update is required - avoid rebasing

### Out of scope

* **hunk-dependencies**
* **atomic object writes**

### DIFF
--- a/crates/but-rebase/tests/fixtures/workspace-update-rebase.sh
+++ b/crates/but-rebase/tests/fixtures/workspace-update-rebase.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# Three stacks of different length with a shared merge-base and an upstream with two added commits
+# and conflicting changes that can be swallowed.
+set -eu -o pipefail
+
+git init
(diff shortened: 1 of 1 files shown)


Byron:
Closing for now as it's going to be the last thing that will be implemented.

PR #7332, opened 2025-02-18
research apply-branch and implement head-info, stack-listing and stack-details

Byron:
Make workspace updates possible, i.e. rebasing multiple branches in a workspace onto a new base.

Follow-up on #7315.
Sibling of #7330

### Tasks

* [x] research
* [x] metadata API sketch
* [x] finish tent-pole ASCII drawings
* [x] strategy for buckets sitting on top of each other
* [x] virtual branches toml adapter for new branch-ref trait
    - [x] retrieve workspace
    - [x] delete workspace
    - [x] update workspace
    - [x] retrieve branch
    - [x] delete branch
    - [x] update branch
    - [x] Make handle carry the ref-name as well (or else it can be surprising)

### Next PRs

* sketch for archived stacks

### For follow-up PRs

In any order (probably)

### Out of scope

* **hunk-dependencies**
* **atomic object writes**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -779,6 +779,7 @@ dependencies = [
  "git2",
  "gitbutler-command-context",
  "gitbutler-commit",
+ "gitbutler-fs",
  "gitbutler-id",
  "gitbutler-oxidize",
  "gitbutler-project",
@@ -789,7 +790,10 @@ dependencies = [
  "insta",
  "itertools 0.14.0",
  "md5",
+ "regex",
  "serde",
+ "toml 0.8.19",
+ "tracing",
  "url",
 ]
 
--- a/crates/but-core/src/lib.rs
+++ b/crates/but-core/src/lib.rs
@@ -43,6 +43,7 @@
 
 use bstr::{BStr, BString};
 use gix::object::tree::EntryKind;
+use gix::refs::FullNameRef;
 use serde::Serialize;
 use std::any::Any;
 use std::ops::{Deref, DerefMut};
@@ -77,7 +78,7 @@ pub mod ref_metadata;
 pub trait RefMetadata {
     /// An implementation-defined wrapper for all data to keep additional information that it might need
     /// to more easily store the data.
-    type Handle<T>: Deref<Target = T> + DerefMut + ref_metadata::ValueInfo;
+    type Handle<T>: Deref<Target = T> + DerefMut + ref_metadata::ValueInfo + AsRef<FullNameRef>;
 
     /// Traverse all available metadata entries and see if their names still exist in the Git ref database.
     ///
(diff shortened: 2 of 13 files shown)


Byron:
A full journey from newly initialized repo to merging a feature branch

```text

┌───┐     ███████████ Create Commit ██████████████Create Stack █████████████Create Branch █████████████████████ Commit feat/1 ████████████████████ Insert Commit Feat █████████████████████████ Commit main ████████████████████████████████ Push main █████████████████████████████████ Push S:feat/1 and B:feat ████████████████ Merge feat/1 PR and fetch ████████████████████████████ Update Workspace █████████████████████████████ Prune Integrated ██████████████████████████
│   │
│ F │                                                                                                                    ┌──┐                                ┌──┐                                    ┌──┐                                      ┌──┐                                         ┌──┐                                             H:ws/1
│ r │                                                                                                          H:ws/1 ──▶│WS│────┐                 H:ws/1 ──▶│WS│───┐   ┌─┐                H:ws/1 ──▶│WS│───┐   ┌─┐                  H:ws/1 ──▶│WS│───┐   ┌─┐                     H:ws/1 ──▶│WS│───┐   ┌─┐ ┌── RTB/feat/1                 ┌─┐   │                                                ┌──┐
│ e │                                                                                                                    └──┘    │                           └──┘   └───└─┘◀─── S:feat/1             └──┘   └───└─┘◀─── S:feat/1               └──┘   └───└─┘◀─── S:feat/1                  └──┘   └───└─┘◀┴── S:feat/1    T:RTB/main ───▶└─┘◀──┼─────────┐                                 ┌────│WS│──────┐
│ s │                                                                                                         WS aids      │    ┌─┐                            │         │                             │         │                               │         │                                  │         │                                  │    ▼         │                                 │    └──┘      │
│ h │                                                                                                         traversal by │    └─┘◀─── S:feat/1               │        ┌┴┐                       ┌────┘        ┌┴┐                         ┌────┘        ┌┴┐                            ┌────┘        ┌┴┐ ┌──  RTB/feat                   │  ┌──┐        │                 T:RTB/main ─┐  ┌─┐     ▲       │                T:RTB/main ─┐  ┌─┐
│   │                                                           H:ws/1                     H:ws/1             parent       │     │                             ├────────┴─┘◀─── B:feat            │             └┬┘◀─── B:feat              │             └┬┘◀─── B:feat                 │             └┬┘◀┴── B:feat                      │  │WS│───┐   ┌─┐ ┌── RTB/feat/1     S:main ─┴─▶TMB─────┼───────┤                   H:S:main─┴─▶└─┘─────────────┐
│ I │                                                              │                          │                            ├─────┘                             │                                 ┌─┐             │                         ┌─┐             │                            ┌─┐             │                                  │  └──┘   └───└─┘◀┴── S:feat/1                   │      │       │                               WMB             │
│ n │                                                              ▼                          ▼                            │                                   │                       S:main ──▶└─┘──┐          │              S:main ─┬─▶└─┘──┐          │                 S:main ─┬─▶└─┘──┐          │                                  │    │         │                                 │   H:ws/1    ┌─┐ ┌── RTB/feat/1               TMB            ┌─┐
│ i │                                        ┌─┐                  ┌─┐                        ┌─┐ ┌─ S:feat/1              ┌─┐                                 ┌─┐                                     └─┬─┬──────┘          T:RTB/main ─┘       └─┬─┬──────┘             T:RTB/main ─┘       └─┬─┬──────┘                                  ├────┘        ┌┴┐ ┌──  RTB/feat                  │             └─┘◀┴── S:feat/1                  │             └─┘
│ t │          H:S:main         H:S:main────▶└─┘      S:main ────▶└─┘◀── S:feat    S:main ──▶└─┘◀┴─ B:feat      S:main ──▶WMB◀───── B:feat          S:main ──▶WMB                                       WMB                                       └─┘                                          └─┘                                         │             └┬┘◀┴── B:feat                     │              │                                │              │
│   │                                                                                                                                                                                                                                             WMB                                          WMB                                        ┌─┐             │                                 │             ┌┴┐ ┌──  RTB/feat                 │             ┌┴┐
│   │                                                                                                                                                                                                                                             TMB                                          TMB                             S:main ───▶└─┘──┐          │                                 │             └┬┘◀┴── B:feat                    │             └┬┘
└───┘                                                 Need ws/1 for metadata:      - Need ws/1 for metadata:                                                                                                                                                                                                                                   └─┬─┬──────┘                                ┌─┐             │                               ┌─┐             │
                                                      stack order                    stack + branch order                                                                                                                                                                                                                                        └─┘                                       └─┘──┐          │                               └─┘──┐          │
                                                                                   - Stack name changed to                                                                                                                                                                                                                                       WMB                                            └─┬─┬──────┘                                    └─┬─┬──────┘
                                                                                     feat/1                                                                                                                                                                                                                                                      TMB                                              WMB                                             └─┘



                                                                                                                                                                                                                                                                                                                                                                               - FF S:main                                     - delete all of S:feat/1 as we created it as well.
                                                                                                                                                                                                                                                                                                                                                                               - Detect S:feat/1 and B:feat are merged and     - delete ws/1 as its metadata can be inferred
                                                                                                                                                                                                                                                                                                                                                                                 do nothing.
                                                                                                                                                                                                                                                                                                                                                                               - remerge WS as S:main tip changed
                                                                                                                                                                                                                                                                                                                                                                               
```

PR #7337, opened 2025-02-19
Limit reference updates if new commits are created to the top-most ref.

Byron:
This is how the application acts currently, and previously we'd update too many branches.


krlvi (approved):
thanks!

PR #7340, opened 2025-02-19
feat: implement Stack selected commit right half of workspace page

ndom91:
## 🧢 Changes

- Add right-half StackDetails view when selecting a commit, including:
	- `FileList`
	- Commit header metadata
	- Show commit title + description as two separate styled text entities out of the 1 `commit.message` field
- Add required redux queries and mutations for getting commit changes and updating commit message
- Fix `StackTabs` dynamic border-radius changes
- Modify `UnifiedDiffView` to take a whole `TreeChange` prop in order to more easily show both worktree changes and committed changes via that 1 component

image


apps/desktop/src/components/v3/CommitRow.svelte
```diff
@@ -7,24 +7,61 @@
...
+		selectedCommitId?: string;
 	}
 
-	const { first, commit, last, lastBranch }: Props = $props();
+	let { first, commit, last, lastBranch, selectedCommitId = $bindable() }: Props = $props();
+
+	const commitTitle = $derived(
+		commit.message.includes('\n\n')
+			? commit.message.substring(0, commit.message.indexOf('\n\n'))
+			: commit.message
```
mtsgrd: Could we do a regex here to ensure the message includes `\n\n` specifically after the first line?

apps/desktop/src/components/v3/UnifiedDiffView.svelte
```diff
@@ -99,22 +89,26 @@
...
-									selection.hunks.some((h) => h.newStart === hunk.newStart))
-								? true
-								: false
-							: undefined}
-						onchange={(selected) => {
-							onchange(hunk, selected, diff.subject.hunks);
-						}}
-					/>
-				{/each}
+				{#if diff.subject.hunks.length}
```
mtsgrd: I think you can drop this if statement? Iterating of a zero length array is a no-op.

apps/desktop/src/lib/stacks/stackService.svelte.ts
```diff
@@ -97,6 +116,29 @@ function injectEndpoints(api: ClientState['backendApi']) {
...
+			getCommitChanges: build.query<
+				TreeChange[],
+				{ projectId: string; oldCommitId: string; newCommitId: string }
+			>({
+				query: ({ projectId, oldCommitId, newCommitId }) => ({
+					command: 'commit_changes',
+					params: { projectId, oldCommitId, newCommitId }
+				}),
+				providesTags: [ReduxTag.CommitChanges]
+			}),
```
mtsgrd: I just spoke with @Byron, he's going to change the api so we don't have to pass in the old commit id, and change the endpoint name in the same go. We'll have something more like `get_commit_changes`, taking only a `commit_id`.

mtsgrd: See #7346!

apps/desktop/src/components/v3/UnifiedDiffView.svelte
ndom91: Hmm yeah logically you're probably right. I had it here in order to be able ot render at least something (in this case the string "No Content"), when there are no hunks.

Ideally I'd like to just not allow the user to toggle them open when there are no hunks, however, due to the way we have it architected now, we don't know the contents of the file change, i.e. the diff, until when the user opens it. 

So its not straight forward to disable toggling open of a file which doesn't have a hunk (i.e. new empty file `Addition`) without checking them all beforehand, which is what we wanted to avoid, right. You have any good ideas? :thinking:

apps/desktop/src/components/v3/CommitRow.svelte
ndom91: Yeah, what're you thinking specifically? Because the first occurence of `\n\n` kind of defines what even is the "first line", you know what I mean? Of course we odn't want it to return the index of the second occurence of the substring `\n\n`, but `indexOf` will already trigger on the first occurence first.

apps/desktop/src/lib/stacks/stackService.svelte.ts
ndom91: Awesome, I see he's merged that. I've updated this as well :+1:

apps/desktop/src/components/v3/CommitRow.svelte
mtsgrd: I mean, if a commit message looks like this:
```
line1
line2

line3
```
Then git log, github, changelog generators etc will use `line1` as a summary. So I'm actually a bit incorrect in my suggestion above. What we want to do is split at the first newline, and trim the expected extra newline from the start of the description. Having two newlines after the title is standard, but it's technically optional.

apps/desktop/src/components/v3/UnifiedDiffView.svelte
mtsgrd: Good point.

@krlvi I think in v2 code we generate a synthetic hunk when there is no diff. It's easier for the front end if we keep that behaviour, but what do you think?

image

apps/desktop/src/components/v3/CommitRow.svelte
ndom91: Okay so we split it at the first `\n`, but then I'm a bit confused about what you want to trim exactly. In the example above

```
line1
---
line2

line3
```

So `line1` becomes title, `line2\n\nline3` becomes description. What needs to be trimmed / changed then?

ndom91: Ohhh I guess in the case of 

```
line1

line2
```

We'd have `line1` as title, and `\nline2` as desc, and you mean that additional preceding `\n` would have to be trimmed off of the desc, right? :sweat_smile: :joy:

mtsgrd: Yes, like `description.trim('\n')` probably makes sense.

mtsgrd: Haha, but also, since we're dealing with html that actually doesn't matter. It will render the same.

apps/desktop/src/components/v3/UnifiedDiffView.svelte
krlvi: Surely we could add a synthetic hunk, but on face value this feels like masking another problem. It's totally valid to have no hunks in the file (eg. file executable bit added/removed) - and the app could/should show that.
At the moment I feel like I'd rather favor correctness (of the API) over supporting the existing semantic.

Is there something that I might be missing here?

krlvi: And I guess the same applies for commits with no files in them - the app allows their creation, and it should be still possible to preview / edit the message

mtsgrd: No you're not missing anything, it's indeed more correct. If we had a synthetic hunk, it also take extra handling in dragging that to a commit. But, we do need think about how to render this. @PavelLaptev any chance you could help out here?

Currently it looks like this when you preview a new file with no content.
image

PavelLaptev: @mtsgrd thanks, I'll take a look today 👀

krlvi: This might be the time to create a view/design for how files look like when they have their flags changed, or if a file became a symlink etc

PavelLaptev: @krlvi got it. But it's better to discuss the design task for in a separate thread or issue

ndom91: @krlvi Yeah +1

I was just reading through the types again and it seems like we could disable the opening of these types of file with those flags too. Would it be a lot of work on your end to add a  `EmptyFile` `TreeChange.status` Flag?

apps/desktop/src/components/v3/CommitRow.svelte
ndom91: Yeah just to be on the safe side, ive added a `trim()` here and I've trimmed it in the [other place we render it in the commit metadata header](https://github.com/gitbutlerapp/gitbutler/pull/7340/files#diff-66565165eac8c72ddca3dd9dec2da64d2fe641ec3164ce9839df189f77c4fd08R26) already :+1:

apps/desktop/src/components/v3/UnifiedDiffView.svelte
mtsgrd: @ndom91 my intuition is that we should still allow the file to be opened, but show a message where we otherwise have a diff.

apps/desktop/src/components/SelectionView.svelte
```diff
@@ -1,27 +1,32 @@
 <script lang="ts">
 	import UnifiedDiffView from './v3/UnifiedDiffView.svelte';
+	import ReduxResult from '$components/ReduxResult.svelte';
```
ndom91: @mtsgrd you wanted to potentially get rid of this file btw, is that still the case? 

I had to fix one or two TS-related things here in this file in order to get the project to build and then just kept going a bit further haha

apps/desktop/src/components/v3/UnifiedDiffView.svelte
ndom91: Okay yeah I'm open to whatever. Maybe Pavel has a nice design idea haha

apps/desktop/src/components/SelectionView.svelte
mtsgrd: Yeah we can drop `SelectionView.svelte`!

apps/desktop/src/components/v3/UnifiedDiffView.svelte
mtsgrd: Not yet, but we will at some point. For now we can just drop some text in its place?

ndom91: Right. Currently there is some text there and it just says "No Content" i.e.:

image

apps/desktop/src/components/SelectionView.svelte
ndom91: Sounds good. Dropped!

apps/desktop/src/components/v3/StackCommitDetails.svelte
```diff
@@ -0,0 +1,63 @@
...
+		selectedCommitDetails?: Commit;
+		selectedBranchDetails?: WorkspaceBranch;
+	}
+
+	let {
+		selectedCommitId = $bindable(),
+		stackId,
+		selectedCommitDetails: commit,
+		selectedBranchDetails
+	}: Props = $props();
```
mtsgrd, line 21: Could we name these as follows perhaps?
```ts
	let {
		stackId,
		commit,
		branch
	}: Props = $props();
```

mtsgrd: Actually, I take that back.. Could we pass them as props and get them from the service? As in, stackid, commitId, and branchId?

mtsgrd (approved):

apps/desktop/src/components/v3/StackCommitDetails.svelte
ndom91: Yeah I'll refactor this in a follow-up :+1:

PR #7343, opened 2025-02-19
Write real git refs and keep them up to date

krlvi:


krlvi:
@Byron whenever you have a moment, would you mind giving this pr some 👀? The main interesting part is in `stack_branches.rs` where I added creating/updating/deleting of git reference via gitoxide, as long as a commit oid is available (as opposed to change ids which are being phased out)

Byron:
It will be done first thing in the morning tomorrow, today was one of those days that didn't quite go as planned.

crates/gitbutler-stack/src/stack_branch.rs
```diff
@@ -74,31 +79,136 @@ impl RepositoryExt for git2::Repository {
...
+                deref: false,
+            };
+            let create = RefEdit {
+                change: Change::Update {
+                    log: LogChange {
+                        mode: RefLog::AndReference,
+                        force_create_reflog: false,
+                        message: "GitButler reference".into(),
+                    },
+                    expected: PreviousValue::ExistingMustMatch(oid.into()),
```
Byron, line 173: This is probably a safe choice as it will overwrite existing references, at the destination, only if they are pointing to the same commit as the current reference (that is being renamed). That's *probably* OK to do.
However, if there is a chance that this is not fully proven to be desirable, one can change it to `PreviousValue::MustNotExist` to assure the destination is 'free'.

crates/gitbutler-stack/src/stack_branch.rs
```diff
@@ -74,31 +79,136 @@ impl RepositoryExt for git2::Repository {
...
+                name: reference.name().into(),
+                deref: false,
+            };
+            repo.edit_references([delete])?;
+        }
+        Ok(())
+    }
+
+    fn rename_real_reference(&self, name: &str, repo: &gix::Repository) -> Result<()> {
+        let current_name: BString = qualified_reference_name(self.name()).into();
```
Byron, line 150: If the current name was equal to the new name, the operation would probably fail as the transaction would try to obtain a lock for each ref-edit.
Thus the caller would have to assure the name is different or face an error.

crates/gitbutler-stack/src/stack_branch.rs
```diff
@@ -74,31 +79,136 @@ impl RepositoryExt for git2::Repository {
...
+        if let Some(reference) = repo.try_find_reference(&current_name)? {
+            let delete = RefEdit {
+                change: Change::Delete {
+                    expected: PreviousValue::MustExistAndMatch(oid.into()),
+                    log: RefLog::AndReference,
+                },
+                name: reference.name().into(),
+                deref: false,
+            };
+            repo.edit_references([delete])?;
```
Byron:
```suggestion
            repo.edit_reference(delete)?;
```

crates/but-workspace/src/commit_engine/refs.rs
```diff
@@ -32,6 +32,7 @@ pub fn rewrite(
         == 1;
     for (old, new) in changed_commits {
         let old_git2 = old.to_git2();
+        let mut already_updated_refs: Vec<bstr::BString> = Vec::new();
```
Byron: If one wants to name the type even though it could be inferred, this would be more common:
```suggestion
        let mut already_updated_refs = Vec::<BString>::new();
```

Byron (approved):
I left a few comments but don't see any blocker here when focussing on the `gitoxide` interactions.
The rest I skimmed and it looked good, too.

crates/but-workspace/src/commit_engine/refs.rs
krlvi: Thanks, this was actually an accidental leftover and didn't intend to keep the explicit type

crates/gitbutler-stack/src/stack_branch.rs
krlvi: Fixed

crates/gitbutler-stack/src/stack_branch.rs
krlvi: Oh i see! I am adding a noop return if the name is the same, just to avoid this codepath alltogether

krlvi:
> I left a few comments but don't see any blocker here when focussing on the `gitoxide` interactions. The rest I skimmed and it looked good, too.

Thanks for the review! Good points, addressing them all in a new commit

PR #7346, opened 2025-02-20
Use one commit only for obtaining changes for Commit

Byron:
## ☕️ Reasoning

That way it's easier to use from the frontend.
It's notable that this cements that we traverse the first parent only
when listing stack commits, which is now something we assume.
Should be alright though.

### DIFF
--- a/crates/gitbutler-tauri/src/diff.rs
+++ b/crates/gitbutler-tauri/src/diff.rs
@@ -24,19 +24,14 @@ pub fn tree_change_diffs(
 
 #[tauri::command(async)]
 #[instrument(skip(projects), err(Debug))]
-pub fn commit_changes(
+pub fn changes_in_commit(
     projects: tauri::State<'_, gitbutler_project::Controller>,
     project_id: ProjectId,
-    old_commit_id: Option<HexHash>,
-    new_commit_id: HexHash,
+    commit_id: HexHash,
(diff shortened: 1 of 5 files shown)


crates/gitbutler-tauri/src/diff.rs
```diff
@@ -24,19 +24,14 @@ pub fn tree_change_diffs(
 
 #[tauri::command(async)]
 #[instrument(skip(projects), err(Debug))]
-pub fn commit_changes(
+pub fn get_commit_changes(
```
Byron: Does this look better?

krlvi: why the get prefix?

ndom91 (approved):
Nice, thanks!

krlvi (changes requested):
since we are not prefixing getters with `get_`, let's not do this on the API as well https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

krlvi:
PS: sorry for being a pain in the ass here

Byron:
Thanks for your help!

We wanted to avoid inconsistency and also the `get_` prefix after all, so went with `changes_in_commit` instead.

krlvi (approved):

PR #7347, opened 2025-02-20
Refactor the stack_branches command - separating commit listing

krlvi:
The changes are as follows:

- The return of `stack_branches` no longer includes commits. Furthermore the archived state is available via a boolean field on `WorkspaceBranch`
- There is now a command `stack_branch_local_and_remote_commits` which provides the commits for a stacked branch which are in the workspace (ordered newest to oldest, just as in the previous API)
- There is also a command `stack_branch_upstream_only_commits` which provides the commits for a stacked branch which are at the upstream only (ordered newest to oldest, as before)

Semantically the two new APIs provide the exact same data that was previously returned as part of the `stack_branches` response.

Note: Both of the two new APIs are currenly only appropriate for accessing the commits of a stack branch in the workspace as they consider the "next head" in the stack as the end of the graph walk.


crates/but-cli/src/command/mod.rs
```diff
@@ -117,4 +119,15 @@ pub mod stacks {
         let ctx = CommandContext::open(&project, AppSettings::default())?;
         debug_print(stack_branches(id.to_string(), &ctx))
     }
+
+    pub fn branch_commits(id: &str, name: &str, current_dir: &Path) -> anyhow::Result<()> {
+        let project = project_from_path(current_dir)?;
+        let ctx = CommandContext::open(&project, AppSettings::default())?;
+        let local_and_remote =
+            stack_branch_local_and_remote_commits(id.to_string(), name.to_string(), &ctx);
+        debug_print(local_and_remote)?;
```
mtsgrd, line 128: 👀

krlvi: @mtsgrd this is intentional - `debug_print` is simply a function for printing in stdout in the testing CLI

Byron: And it's called `debug_*` because it invokes the `std::fmt::Debug` version of the Rust formatting system, basically the Rust object notation (RON), as the cheapest possible way to get an idea of what functions produce.

`but-cli` is used to bring new V3 implementations into the real world for the first time, and sometimes help with measuring performance in a non-interactive way.

PR #7355, opened 2025-02-21
Implement the review card

Caleb-T-Owens:
This commit introduces the but-review card such that it matches the current design. This commit and the following commits in this series are not interested in what the PR looks.

## The general strucutre.

We want to have the cards be part of a new "multi-card" which has one or two sections. I've achieved this by letting the over-arching "BranchReivew" component handle the border around the PR and BR card contents, rather than the two managing that themselves.

This means that the PullRequestCard is not really a card at all anymore, but is instead just the gubbins that live inside the card.

---
- <kbd>&nbsp;3&nbsp;</kbd> #7355 👈 
- <kbd>&nbsp;2&nbsp;</kbd> #7354 
- <kbd>&nbsp;1&nbsp;</kbd> #7353


schacon (commented):
Sorry, just a test.

PR #7386, opened 2025-02-24
assure there are no trees in commit-changes

Byron:
The reason for this shortcoming was quite literally that no test ever created a change in a directory.
This was done now to assure it won't regress.


mtsgrd:
Perfect, thanks!

PR #7389, opened 2025-02-24
Add "commit goes here" section

mtsgrd:
---
- <kbd>&nbsp;2&nbsp;</kbd> #7390 
- <kbd>&nbsp;1&nbsp;</kbd> #7389 👈

### DIFF
--- a/apps/desktop/src/components/v3/Branch.svelte
+++ b/apps/desktop/src/components/v3/Branch.svelte
@@ -3,9 +3,11 @@
 	import ReduxResult from '$components/ReduxResult.svelte';
 	import BranchCommitList from '$components/v3/BranchCommitList.svelte';
 	import BranchHeader from '$components/v3/BranchHeader.svelte';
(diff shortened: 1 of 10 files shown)


mtsgrd:
@PavelLaptev I have some work left, but the navigation stuff wrt to committing works here if you want to poke at it.

image

mtsgrd:
@Byron at the moment I can only commit changes to a specific location in the top branch, otherwise it fails as seen in this screenshot:
image

Byron:
Thanks for reeling me in. I did take a quick look and used the `but-cli` to amend to each of the commits that were referenced by named references in this setup:

Screenshot 2025-02-25 at 20 26 45

Then I inserted them below the tip of each branch, and besides ref-rewriting being unaware of the the 'bucket' nature of references, it seemed to have worked.

Screenshot 2025-02-25 at 20 33 20

So the next step of debugging would be to use the code in this branch, but I will do that tomorrow.

mtsgrd:
That's strange, and now I can't reproduce..

PR #7399, opened 2025-02-25
OS Notifications

estib-vega:
- **OS Notifications Module**: 
  - Introduces a new module for sending OS application notifications from the frontend.
  - Provides a utility for easier integration of notifications in the application.

- **Rust Toolchain Stabilization**:
  - Reverts to the stable version of the Rust toolchain.
  - Avoids the use of the experimental `checked_signed_diff` method for safe u32 subtraction.
  - Updates the hash snapshots to ensure compatibility and stability.

**Note**: macOS users must enable notifications for the Terminal application during development to receive notifications.


estib-vega:
Damnnnn, so the current blocker for this is that the application needs to be in the Cargo nightly toolchain, in order to access the experimental feature `windows_by_handle` https://github.com/rust-lang/rust/issues/63010, that exposes some needed metadata.

Tauri notifications have a transitive dependency that breaks on the Cargo nightly toolchain...

This kind of sucks

estib-vega:
Next thing to try: https://github.com/basharovV/musicat/issues/69#issuecomment-2567117090

Apparently we might be able to forgo the experimental Windows utils for that package.
Need to test that in a Windows machine, though

PR #7403, opened 2025-02-26
first steps for 'head-info'

Byron:
A first implementation of `head_info()` to provide information about where the user currently is.
It presents the world as a workspace with stacks, even though technically it might not exist.`

Follow-up on #7332

### Tasks

* [x] final polish of ref-metadta
* [x] more sketches for detached HEAD
* [x] basic `head-info` (untracked, detached)
* [ ] non-workspace cases
    - [x] simple branch
    - [ ] merge commit
    - [x] simple branch stacked
    - [ ] merge commit stacked
    - [ ] stacks and ambiguous stack references (i.e. lots of empty stack segments)
* [ ] workspace cases

### Next PRs

* ⚠️One probably wants to show all refs at a position (or indicate there are more) ⚠️

### For follow-up PRs

In any order (probably)

### Out of scope

* **hunk-dependencies**
* **atomic object writes**


Byron:
A couple of additional scenarios:

```text

┌───┐     ██████████████████ Unapply the last Stack (main) ████████████████████████ Apply main (or switch to it) ████████
│   │
│   │         H:S:main                           main                                        H:S:main
│ D │             │                                │                                             │
│ e │             ▼                                ▼                                             ▼
│ t │            ┌─┐                              ┌─┐                                           ┌─┐
│ a │            └─┘                      HEAD ──▶└─┘                                           └─┘
│ c │             │                                │                                             │
│ h │             │                                │                                             │
│   │             │                                │                                             │
│ H │            ┌─┐                              ┌─┐                                           ┌─┐
│ E │            └─┘                              └─┘                                           └─┘
│ A │
│ D │                              Unapplying the last stack from a
│   │                              workspace detaches the HEAD.
│   │
└───┘                              No stack is available now.



┌───┐     ██████████████████ Unapply the last Stack (main) ████████████████████████ Apply main (or switch to it) ████████
│   │
│   │         H:S:main                           main                                        H:S:main
│ D │             │                                │                                             │
│ e │             ▼                                ▼                                             ▼
│ t │            ┌─┐                              ┌─┐                                           ┌─┐
│ a │            └─┘                      HEAD ──▶└─┘                                           └─┘
│ c │             │                                │                                             │
│ h │             │                                │                                             │
│   │             │                                │                                             │
│ H │            ┌─┐                              ┌─┐                                           ┌─┐
│ E │            └─┘                              └─┘                                           └─┘
│ A │
│ D │                              Unapplying the last stack from a             Now the HEAD isn't detached anymore.
│   │                              workspace detaches the HEAD.
│   │
└───┘                              No stack is available now.


┌───┐     ██████████████████ List Stack Commits in WS ██████████
│   │                  ┌──┐
│ L │        H:ws/1 ──▶│WS│
│ a │                  └──┘
│ n │                    │                      ┌───┐┌───┐┌───┐
│ e │             ┌──────┼──────┐               │ A ││ B ││ C │
│   │             │      │      │               │┌─┐││┌─┐││┌─┐│
│ C │            ┌─┐    ┌─┐    ┌─┐              │└2┘││└4┘││└5┘│
│ o │   S:A   ──▶└2┘  ┌▶└4┘    └5┘◀─┐           │   ││┌─┐││   │
│ m │             │   │  │      │   │           │   ││└3┘││   │
│ m │             │   │  └─┬────┘   │           └───┘└───┘└───┘
│ i │             │ S:B    │      S:C
│ t │             │       ┌─┐          Each commit is only listed once, and
│ s │             │       └3┘          consistently based on an algorithm.
│   │             │        │
│   │            ┌─┐       │           This also means that one has to handle
└───┘            WMB───────┘           all commits at once.



┌───┐     ██████████████████ Lis Commits in ordinary Merge ██████████████
│   │                   ┌─┐
│ L │       H:S:main──▶ └3┘
│ a │                    │
│ n │             ┌──────┼──────┐               main┐
│ e │             │      │      │               │┌─┐│
│   │             │      │      │               │└3┘│
│ C │            ┌─┐    ┌─┐    ┌─┐              │┌─┐│
│ o │            └2┘    └4┘    └5┘              │└2┘│
│ m │             │      │      │               │┌─┐│
│ m │             │      └─┬────┘               │└1┘│
│ i │             │        │                    └───┘
│ t │             │       ┌─┐
│ s │             │       └6┘       First parent only traversal of merge
│   │             │        │        commits.
│   │            ┌─┐       │
└───┘            └1┘───────┘        Problem is that lanes wouldn't have
                                    names otherwise.

                                    Maybe one day we figure out something
                                    else, but this is safe.

                                    One could imagine allowing to 'switch
                                    junctions', decide which parent to
                                    walk along.
```

Byron:
Merging to avoid more conflicts, will pick it up after next intermediate topic is done.

PR #7438, opened 2025-02-27
Closing buts

Caleb-T-Owens:


packages/shared/src/lib/branches/branchService.ts
```diff
@@ -92,6 +92,32 @@ export class BranchService {
 			.createInterest();
 	}
 
+	async getBranch(uuid: string): Promise<Branch | undefined> {
```
Caleb-T-Owens, line 95: I don't think I ended up using this

estib-vega (approved):

PR #7441, opened 2025-02-27
insert commit at reference

Byron:
Assure that the commit-insertion respects the insertion position specified in the UI.

### Tasks

* [x] fix it in backend
* [x] frontend provides ref-info to drive backend


Byron:
@mtsgrd @ndom91 

This is a fix for the "cannot create commit in the right spot problem" - to really make it work, one will have to pass the short name of the stack segment (branch) in which the commit should be created. It's probably easy for you, but impossible for me, so @ndom91 and I thought it's best to hand it off.

Thank you.

apps/desktop/src/components/v3/NewCommit.svelte
```diff
@@ -69,6 +69,7 @@
 			stackId,
 			parentId,
 			message: message,
+			stackSegmentShortName: "top",
```
Byron, line 72: Here is where the actual stack-segment name needs to be added.

Screenshot 2025-02-27 at 15 33 55

Where ever the "Your commit goes here" marker is, that should inform the name to be passed. In the case above, it would be "top".

That way, the commit will always go into the right segment/bucket.

Byron:
Submitted using other PR

PR #7469, opened 2025-03-01
Updated navigation animation on scroll

PavelLaptev:
## 🧢 Changes

The animation before was jumpy and the layout didn't work correctly in between states.

Before:

https://github.com/user-attachments/assets/a03c42f3-6844-4752-809d-1af606fda24e

After:

https://github.com/user-attachments/assets/b53b732b-abdd-4583-b54d-4bb922c6c9c2


PavelLaptev:
@estib-vega fyi

PR #7470, opened 2025-03-02
Update minimized chat look and behaviour

PavelLaptev:
## 🧢 Changes

- chat should be fullscreen on mobile mode when you maximizer it because content on the background doesn;t make sence because it's too small behind the chat modal
- On table/mobile chat should be minimized by default
- Chat should have full width

Before:

https://github.com/user-attachments/assets/1825848d-c44b-4e58-baa9-16d7644df3c5

After:

https://github.com/user-attachments/assets/dbe70f0f-07fd-40b6-9631-68d53cf7b065


PavelLaptev:
@estib-vega fyi

PR #7474, opened 2025-03-03
ref-based-insertions

ndom91:
## 🧢 Changes

- Follow-up to @byron's #7441
	- Ref based insertions
- Extract empty commit state from living in branchCommitList
- Includes frontend changes to allow selecting a new (empty) dependent branch as target for your commit

> Assure that the commit-insertion respects the insertion position specified in the UI.
> 
> ### Tasks
> 
> * [x] fix it in backend
> * [x] frontend provides ref-info to drive backend


apps/desktop/src/components/v3/NewCommit.svelte
```diff
@@ -39,7 +41,30 @@
...
+			if (currentBranchIndex >= 0 && currentBranchIndex < branches.length - 1) {
+				const parentBranch = branches[currentBranchIndex + 1];
+				if (parentBranch?.name) {
+					const parentCommit = stackService.commitAt(projectId, stackId, parentBranch.name, 0)
+						.current.data;
+
+					if (parentCommit?.id) return parentCommit.id;
+				}
+			}
+		}
```
mtsgrd, line 63: Not sure if I'm reading correctly, but would this work if there are two empty branches, and you want to commit to the top one?

crates/but-cli/src/args.rs
```diff
@@ -31,6 +31,11 @@ pub enum Subcommands {
         /// The message of the new commit.
         #[clap(long, short = 'm')]
         message: Option<String>,
+        /// The name of the reference that the commit should be in.
+        ///
+        /// If there is ambiguity, this is what makes it ambiguous.
+        #[clap(long, short = 's')]
+        stack_segment_ref: Option<String>,
```
mtsgrd, line 38: @krlvi I'm wondering if `stack_segment` is a term we want to introduce? To me it would feel more natural to talk about branches.

apps/desktop/src/components/v3/NewCommit.svelte
ndom91: No you're right, it wouldn't. First thing that comes to mind, we could `while` loop down the branches until we find the first branch with a commit that will act as parent.

Alternatively, it might be easier to have an API that returns something like `StackBranch` with an array of commits with basic metadata on it already, what do you think @mtsgrd  @krlvi ?

mtsgrd: In theory we could have any number of empty branches in between the one you're committing to and a previous commit (or base), so I'd prefer to avoid the looping altogether by making the API capable of committing directly to a branch without specifying the parent commit. Let's chat in person during once Kiril is in the office.

crates/but-cli/src/args.rs
krlvi: For the UI we are gonna have to keep user-friendly language. 
On the backend, the term stack segment describes a range between two references (which is distinct from branches, which will technically include each other)

apps/desktop/src/components/v3/NewCommit.svelte
Byron: I think the API already takes a branch name that is currently used to insert into the correct 'bucket'. From there the parent commit could trivially be extracted.

A branch name will always be known, but it won't have to necessarily be pointing to a commit within the stack that the user sees. The V3 workspace information can represent such cases, but for now it should be sufficient if the UI passes the name it knows, or the parent commit, or both, the backend should be able to deal with it.

Byron:

It does, however, look like the PR was abandoned, as it doesn't merge cleanly and wasn't updated in quite some time.

PR #7527, opened 2025-03-06
Simplify all resizers

mtsgrd:


mtsgrd:
@PavelLaptev you might like this. 🥳

apps/desktop/src/components/BranchPreview.svelte
```diff
@@ -152,11 +140,8 @@
 			<Resizer
 				viewport={rsViewport}
 				direction="right"
-				minWidth={320}
-				onWidth={(value) => {
-					laneWidth = value / (16 * $userSettings.zoom);
-					lscache.set(laneWidthKey, laneWidth, 7 * 1440); // 7 day ttl
-				}}
+				minWidth={20}
```
estib-vega, line 143: Is it on purpose that the min width went from 320 to 20?

estib-vega (approved):

apps/desktop/src/components/BranchPreview.svelte
mtsgrd: Yes, it used to work with pixels, and now it works with `rem`.

estib-vega: Ah, nice 👍 

Ship itttttttt :)

PR #7536, opened 2025-03-07
Fix commit engine committing with multiple empty stacks

krlvi:
Because of incorrect reference updates, it appeared as if the commit was added to all of the empty stacks. This adds a check an additional check preventing that

### DIFF
--- a/crates/but-cli/src/command/commit.rs
+++ b/crates/but-cli/src/command/commit.rs
@@ -1,9 +1,11 @@
 use crate::command::debug_print;
 use anyhow::bail;
 use but_core::TreeChange;
-use but_workspace::commit_engine::{DiffSpec, ReferenceFrame, create_commit_and_update_refs};
+use but_workspace::commit_engine::{
+    DiffSpec, ReferenceFrame, StackSegmentId, create_commit_and_update_refs,
+};
 use gitbutler_project::Project;
-use gitbutler_stack::VirtualBranchesState;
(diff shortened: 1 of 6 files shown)


krlvi:
@Byron FYI

crates/but-cli/src/command/commit.rs
```diff
@@ -21,20 +23,34 @@ pub fn commit(
...
     if let Some(project) = project.as_ref() {
+        let destination = if amend {
+            if message.is_some() {
+                bail!("Messages aren't used when amending");
+            }
+            but_workspace::commit_engine::Destination::AmendCommit(parent_id)
+        } else {
+            let stack_segment = if let Some(stack_segment_ref) = stack_segment_ref {
+                let full_name = gix::refs::FullName::try_from(stack_segment_ref)?;
+                VirtualBranchesHandle::new(project.gb_dir())
```
Byron, line 36: Thanks for the ping @krlvi!

This will certainly work as a workaround, but also is something that I think must be avoided. Now there is a strong entanglement to V2 and its StackIds.

My intuition here is that disambiguation would need to introduce the stack reference, i.e. the top-most reference in a stack. Somehow I think this underlying problem is also improper validation of the workspace - in theory a single reference should only be assigned to a single stack and not multiple. I added a note about this to https://github.com/gitbutlerapp/gitbutler/pull/7403 and think this can be removed once V3 is more complete.

PR #7558, opened 2025-03-09
Use the latest models from OpenAI

vidluther:
## 🧢 Changes

* add o3-mini
* add o1-mini 
* remove GPT 3-5 Turbo
* remove GPT Turbo
* remove GPT 4o 

Update the service tests to make sure they pass.

### DIFF
--- a/apps/desktop/src/lib/ai/service.test.ts
+++ b/apps/desktop/src/lib/ai/service.test.ts
@@ -34,7 +34,7 @@ import type { SecretsService } from '$lib/secrets/secretsService';
 const defaultGitConfig = Object.freeze({
 	[GitAIConfigKey.ModelProvider]: ModelKind.OpenAI,
 	[GitAIConfigKey.OpenAIKeyOption]: KeyOption.ButlerAPI,
-	[GitAIConfigKey.OpenAIModelName]: OpenAIModelName.O3mini,
+	[GitAIConfigKey.OpenAIModelName]: OpenAIModelName.GPT4oMini,
 	[GitAIConfigKey.AnthropicKeyOption]: KeyOption.ButlerAPI,
 	[GitAIConfigKey.AnthropicModelName]: AnthropicModelName.Haiku
 });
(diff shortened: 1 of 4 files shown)


apps/desktop/src/lib/ai/service.test.ts
```diff
@@ -34,7 +34,7 @@ import type { SecretsService } from '$lib/secrets/secretsService';
 const defaultGitConfig = Object.freeze({
 	[GitAIConfigKey.ModelProvider]: ModelKind.OpenAI,
 	[GitAIConfigKey.OpenAIKeyOption]: KeyOption.ButlerAPI,
-	[GitAIConfigKey.OpenAIModelName]: OpenAIModelName.GPT35Turbo,
+	[GitAIConfigKey.OpenAIModelName]: OpenAIModelName.O3mini,
```
estib-vega: Let's set 4o mini as a default instead, please

apps/desktop/src/lib/ai/service.ts
```diff
@@ -123,7 +123,7 @@ export class AIService {
 	async getOpenAIModleName() {
 		return await this.gitConfig.getWithDefault<OpenAIModelName>(
 			GitAIConfigKey.OpenAIModelName,
-			OpenAIModelName.GPT35Turbo
+			OpenAIModelName.O3mini
```
estib-vega: Same here

apps/desktop/src/lib/ai/service.test.ts
vidluther: @estib-vega I just remembered the cost of o3-mini and o1-mini is the same.. they even recommend using o3-mini on the o1-mini page.. https://platform.openai.com/docs/models/o1-mini

"The o1 reasoning model is designed to solve hard problems across domains. o1-mini is a faster and more affordable reasoning model, but we recommend using the newer o3-mini model that features higher intelligence at the same latency and price as o1-mini." 

I can set it to o1-mini if you prefer.. but just wanted to point this out first.

estib-vega: @vidluther I was talking about **4o mini** :) That one is waaay cheaper

vidluther: oh yea.. duh

vidluther: The changes are now in my master branch, so I think this PR gets updated automatically now? I'm not 100% sure how this GH/PR magic works..

estib-vega (approved):

estib-vega:
@vidluther Nice!
Thanks a lot!

PR #7575, opened 2025-03-11
Run GPG in shell on Unix as well (#7467)

### LINKED ISSUE #7467: Broken git signing
### Version

0.14.8

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

I'm using macos gitbutler with nix-darwin based gpg. Whenever I try to enable git signing, I always get this error

Image

It used to work on a nightly version of gitbutler (I commented about it in #5022), but it's not working anymore. I don't know the exact version when it stopped working, but I can check it if you need it.

Also, git signing works normally on normal cli git.

_No response_

_No response_

```shell

```

Byron:
This way there are higher chances of picking up the required PATH configuration, making
signing operations more likely to work.

Fixes #7467 .

### Tasks

* [x] implementation
* [x] see if this improves anything on MacOS (where signing doesn't work for me either without running from a terminal)
    - It doesn't just magically work for me which is expected as `sh` isn't `zsh`, which would be more likely to work
* [x] try with `SHELL`.
     - It still doesn't work for me, despite `zsh`, but that's merely a problem of my setup which requires a login shell. Screenshot 2025-03-11 at 12 16 31
     - However, I think doing this will greatly improve the chances that signing works out of the box.
* [x] ❌ validate on windows (not easy, forgot there is a special incantation for ARM VMs), but nonetheless it doesn't startup, giving a null-pointer exception somewhere in tauri each time.

### DIFF
--- a/crates/but-core/src/cmd.rs
+++ b/crates/but-core/src/cmd.rs
@@ -8,7 +8,9 @@ pub fn prepare_with_shell(program: impl Into<OsString>) -> gix::command::Prepare
         // On Windows, this means a shell will always be used.
         .command_may_be_shell_script_disallow_manual_argument_splitting()
         // On Windows, this yields the Git-bundled `sh.exe`, which is what we want.
-        .with_shell_program(gix::path::env::shell())
+        .with_shell_program(
+            std::env::var_os("SHELL").unwrap_or_else(|| gix::path::env::shell().into()),
+        )
         // force using a shell, we want access to additional programs here
         .with_shell()
         // We know `program` is a path, so quote it.
--- a/crates/but-core/src/cmd.rs
+++ b/crates/but-core/src/cmd.rs
@@ -1,17 +1,16 @@
 use std::ffi::OsString;
 
-/// Prepare `program` for invocation with a shell on Windows, and directly everywhere else.
-pub fn prepare_with_shell_on_windows(program: impl Into<OsString>) -> gix::command::Prepare {
-    let prepare = gix::command::prepare(program);
-    if cfg!(windows) {
-        prepare
-            .command_may_be_shell_script_disallow_manual_argument_splitting()
-            // On Windows, this yields the Git-bundled `sh.exe`, which is what we want.
-            .with_shell_program(gix::path::env::shell())
-            // force using a shell, we want access to additional programs here
-            .with_shell()
-            .with_quoted_command()
-    } else {
-        prepare
-    }
+/// Prepare `program` for invocation with a Git-compatible shell to help it pick up more of the usual environment.
+///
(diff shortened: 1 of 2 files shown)


Byron:
@krlvi Maybe you could take a look and *try* the Windows version, something that I failed to do unfortunately due to strange technical issues on this machine. The reason it should be tried is that now the `SHELL` environment variable is read, and I'd hope this is nothing strange on Windows (like powershell) which might make something that worked before fail.

krlvi:
> @krlvi Maybe you could take a look and _try_ the Windows version, something that I failed to do unfortunately due to strange technical issues on this machine. The reason it should be tried is that now the `SHELL` environment variable is read, and I'd hope this is nothing strange on Windows (like powershell) which might make something that worked before fail. Thank you.

Thanks for this. I will try to validate this today

krlvi:
@Byron I was able to test signing working correctly on Windows with this branch. Both in the test UI as well as in commit creation.

Unfortunately setting gpg.program is not enough:
image

But if I set it in GitButler, signing works

image

krlvi (approved):

krlvi (approved):

Byron:
Thanks so much for validating!

It's interesting that the `--global` configuration doesn't work even though it should, and makes it seem like `gix` doesn't pickup `--global` options in this environment.
I took a look and `global` corresponds to `C:/Users/byron/.gitconfig` on a user's machine, which is a file that would be picked up (`gix config` shows it, too).

Something I find strange is the need to use msys-paths, but maybe `/`'ed paths would work as well , i.e. `c:/foo/gpg`. It's good to see that `/c/` paths also work, which means that using the Git shell is doing the trick.

Byron:
Now that I could build and run the application, I could finally check what's in the environment on Windows.

This is when launching the dev version from a terminal:

Screenshot 2025-03-12 at 08 59 07

The shell is the one set by the terminal. Unfortunately the debug-build itself doesn't bring up the UI so I could not see what actually happens in a 'normal' instance.
I'd assume that `SHELL` isn't set then, which means it gets the same value as before (the one determined by `gitoxide`).

krlvi:
@Byron FYI it was reported on Discord that this change possibly interferes with signing when the shell is Fish

> did something change with commit signing? I've been using fish for the the past 4 months, and all of a sudden in nightly nad dev, I'm seeing this error.. 
```
fish: $@ is not supported. In fish, please use $argv.
```
image

Byron:
🤦‍♂️Sorry for that, I really knew better, yet made this mistake.

Here is the fix: https://github.com/gitbutlerapp/gitbutler/pull/7623 .

PR #7578, opened 2025-03-11
Use libgit2 for logging butler requests

Caleb-T-Owens:
When using the gix logging, it appears that merge commits were also getting logged. Unfortunatly gix has no concept of "hide", so we've temporarily made the switch over to libgit2

### DIFF
--- a/crates/gitbutler-sync/src/stack_upload.rs
+++ b/crates/gitbutler-sync/src/stack_upload.rs
@@ -6,7 +6,7 @@ use anyhow::{bail, Result};
 use gitbutler_command_context::CommandContext;
 use gitbutler_commit::commit_headers::HasCommitHeaders as _;
 use gitbutler_oplog::reflog::{set_reference_to_oplog, ReflogCommits};
-use gitbutler_oxidize::{git2_signature_to_gix_signature, OidExt as _};
+use gitbutler_oxidize::{git2_signature_to_gix_signature, ObjectIdExt, OidExt as _};
 use gitbutler_repo::{commit_message::CommitMessage, signature};
 use gitbutler_stack::{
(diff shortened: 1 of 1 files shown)


crates/gitbutler-sync/src/stack_upload.rs
```diff
@@ -44,13 +44,12 @@ pub fn push_stack_to_review(
...
 
     let target_commit_id = vb_state.get_default_target()?.sha.to_gix();
-    let commits = repository
-        .rev_walk([top_branch.id])
-        .first_parent_only()
-        .with_boundary([target_commit_id])
-        .sorting(gix::revision::walk::Sorting::BreadthFirst)
-        .all()?
-        .filter_map(|e| Some(e.ok()?.id().detach()))
+    let git2_repository = ctx.repo();
```
Byron, line 47: Yes, for rev-walking there unfortunately is no equivalent replacement in `gix` yet, and we'd basically have to use `git2` whenever `hide` is desired.
Sorry for the hassle.

Caleb-T-Owens: @Byron I think we may have been seeing a gix bug at play, because in some cases we were seeing what looked like commits belonging to a parent of a merge commit that wasn't the first.

Pavel had a branch which was including commits from origin/master even though we specified `first_parent_only`, which should have been just the commits that he authored

Byron: Let's see if it still occurs once `gix` gets a real `hide` implementation.
A quick look at the `first_parent_only` implementation doesn't make it look like it could be incorrectly implemented either.
In any case, if `hide` is what's wanted here, then `with_boundary()` definitely isn't an adequate substitute, so this fix seems like the right way to go.

PR #7583, opened 2025-03-11
Bump tauri plugin single instance

krlvi:
The previous version had a null pointer error under windows


Byron:
Ah, right! So it should work now for me as well.
I shall remember to update all dependencies next time this happens.

PR #7597, opened 2025-03-12
discard changes

Byron:
Add V3 facilities for discarding changes in the worktree or index.

### Tasks

* [x] reset whole files (with safety-check as change must be matched)
    - [x] addition
    - [x] deletion
        - [x] symlinks
        - [x] submodules
        - [x] executable bit
    - [x] modification
        - [x] symlinks
        - [x] submodules
        - [x] executable bit
        - [x] type-change
    - [x] fix: re-generate submodule from local clone only
    - [x] restore type change from file to dir (dir is to be restored)
    - [x] replace file with directory (with tests for single and multi-discard)
    - [x] renamed file
        - [x] onto nothing
        - [x] onto directory
        - [x] onto other file
    - [x] renamed directory onto file
    - [x] conflicted (aren't listed, so don't need handling, double-check)
    - [x] walk-through-symlink-safety
* [x] index-update support (`.git/index` needs to represent the worktree) as needed
* [x] update `gix` to point to `main`
* [x] CLI integration
* [x] fix disk-space issue by avoiding an additional docker invocation, and fix the now failing tests.
* [x] `tauri`-integration to satisfy current UI capabilities
* [x] ~~restore rust-test parallelism~~ - no need, the cache works well and a full test-run is now below 5 minutes

### Next PR

* [ ] reset changed hunks in a file
* [ ] reset individual lines

### Notes for the Reviewer

* I flattened the Rust test setup, removing indirections and seemingly speeding it up, while reducing the chance that the runner runs out of disk space. For this PR, this now happened all the time, probably because running through docker consumes more disk space? Having a 'dumber' CI setup also seems like a win to me, despite less elegant handling of dependencies.
    - Installing dependencies takes 34s, and pulling the container takes 36. However, the 36s contains the Rust cache, which isn't included in the 34s count.
    - Overall Usage for Rust jobs is reduced from 36 minutes to ~~24~~ 19 minutes, and as less jobs are run more PRs can run in parallel.
    - Admittedly, removing docker from the mix was just an attempt to make CI work again, and it does indeed work now.
    - An uncached Rust test CI run can take 12 minutes, but with a cache this goes down to 4m 50s, which is less than the usual docker run (which seems a little slower under the same circumstances) Screenshot 2025-03-18 at 14 50 24
    - A CI run still takes just a little more than 5 minutes, so it remains at the sweet spot.

* It turns out that the frequent CI failures were caused by disguised out-of-disk-space errors. They appear to be an LLVM-LD crash, but apparently that is just a C-Programs way of saying that [the disk is full](https://github.com/llvm/llvm-project/issues/98988#issuecomment-2235270931).
    - Clearling the GH caches usually does the trick, but it's bound to re-occur.

### Shortcomings

* UX: submodules aren't automatically checked out if they are part of a directory. It's a relatively rare case as well.
* UX: Around renames, there can be a special case that makes the user discard a set of files twice before they go away. It's rare, and it's fixable if one wants to put in the time.

### Out of Scope

### Research

* **V2 implementation**


krlvi:
Awesome! One random thought - when it comes time to wire this as a tauri /CLI command, we could consider having the guard passed instead of it being created deeper in the call stack. Mainly because the snapshotting would need the same guard too - here's what i have in mind https://github.com/gitbutlerapp/gitbutler/commit/d3e1f42cc7160320c7cde35ed99c3f57fa528b64

Byron:
Right, thanks for the reminder! I remember the dead-lock problem that taught us that lesson 😅.
Maybe it's fair to say that these plumbing crates should never create this guard themselves, so either CLI or Tauri have to create it.

krlvi (approved):

krlvi:
Looking great, thanks Byron!

crates/gitbutler-tauri/src/workspace.rs
```diff
@@ -199,3 +199,33 @@ pub fn amend_commit_from_worktree_changes(
...
+#[tauri::command(async)]
+#[instrument(skip(projects), err(Debug))]
+pub fn discard_worktree_changes(
+    projects: State<'_, projects::Controller>,
+    project_id: ProjectId,
+    worktree_changes: Vec<but_workspace::discard::ui::DiscardSpec>,
+) -> Result<Vec<but_workspace::discard::ui::DiscardSpec>, Error> {
+    let project = projects.get(project_id)?;
+    let repo = but_core::open_repo(&project.worktree_path())?;
+    let _guard = project.exclusive_worktree_access();
```
Byron, line 217: This implementation is made around "discard only uses Git and probably there aren't issues with parallelism", so it doesn't take the write-permission itself (yet). The CLI doesn't obtain it, for instance, for that reason.

Something I thought about is to create a second, slightly higher-level function of `discard` that basically just takes a write-permission as argument, to nudge callers to use that instead. It felt, however, that it's enough to do that in `tauri` for now.

I could totally understand that there should rather be the habit of passing write permissions, even though I usually avoid any extras on the function under test, so a separate one would be easier to handle.

PR #7600, opened 2025-03-12
fix the hot reload override

estib-vega:
Updating the external packages (ui or shared) trigger a server restart

### DIFF
--- a/apps/desktop/vite.config.ts
+++ b/apps/desktop/vite.config.ts
@@ -72,8 +72,7 @@ export default defineConfig({
  */
 function debounceReload(): Plugin {
 	let timeout: NodeJS.Timeout | undefined;
-	let mustReload = false;
-	let longDelay = false;
+	let restartTimeout: NodeJS.Timeout | undefined;
 
 	return {
 		name: 'debounce-reload',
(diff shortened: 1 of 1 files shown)


apps/desktop/vite.config.ts
```diff
@@ -82,26 +81,75 @@ function debounceReload(): Plugin {
...
+					mustReload = true;
+					longDelay = true;
+					break defineFlags;
+				}
+
+				if (isSvelteKitFile(file)) {
+					mustReload = true;
+					break defineFlags;
+				}
 			}
```
mtsgrd: This labeled block feels a bit extra, would you object to a vanilla if/else thing?
```ts
if (isLocalPackageFile(file)) {
    mustReload = true;
    serverRestart = true;
    longDelay = true;
} else if (isNotInDesktopApp(file)) {
    mustReload = true;
    longDelay = true;
} else if (isSvelteKitFile(file)) {
    mustReload = true;
}
```

mtsgrd (approved):
Amazing.

apps/desktop/vite.config.ts
estib-vega: @mtsgrd haha, it is a bit extra build I'd argue it's more readable. Is it ok if I keep it?

apps/desktop/vite.config.ts
```diff
@@ -82,26 +81,66 @@ function debounceReload(): Plugin {
...
+				mustReload = true;
+				serverRestart = true;
+				longDelay = true;
+			} else if (isNotInDesktopApp(file)) {
 				mustReload = true;
 				longDelay = true;
-			} else if (file.includes('.svelte-kit')) {
+			} else if (isSvelteKitFile(file)) {
 				mustReload = true;
 			}
```
estib-vega, line 97: 🤮

PR #7607, opened 2025-03-12
Changes in branch API

krlvi:
Adds a new endpoint `changes_in_branch` which gives the tree changes for the entire branch (up to it's base).
The base of the branch is either the head of the branch it is stacked on top of or the merge base of the stack (when it is the bottom one)

---
⧓ Review in [Butler Review `#5ufpxPWxl`](https://gitbutler.com/krlvi/gitbutler/reviews/5ufpxPWxl)

**branch-changes-in-branch**

1 commit series (version 1)

| Series | Commit Title | Status | Reviewers | 
| --- | --- | --- | --- |
| 1/1 | [Changes in branch API](https://gitbutler.com/krlvi/gitbutler/reviews/5ufpxPWxl/commit/d951eae5-cd3f-4fbf-b40a-5160e9db022f) | 💬 | ,  |

_Please leave review feedback in the [Butler Review](https://gitbutler.com/krlvi/gitbutler/reviews/5ufpxPWxl)_

### DIFF
--- a/crates/gitbutler-tauri/src/diff.rs
+++ b/crates/gitbutler-tauri/src/diff.rs
@@ -1,7 +1,11 @@
 use crate::error::Error;
 use crate::from_json::HexHash;
 use but_core::ui::{TreeChange, WorktreeChanges};
+use but_workspace::StackId;
+use gitbutler_command_context::CommandContext;
+use gitbutler_oxidize::OidExt;
 use gitbutler_project::ProjectId;
+use gitbutler_stack::{stack_context::CommandContextExt, VirtualBranchesHandle};
 use tracing::instrument;
 
 /// Provide a unified diff for `change`, but fail if `change` is a [type-change](but_core::ModeFlags::TypeChange)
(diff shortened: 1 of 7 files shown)


crates/gitbutler-tauri/src/diff.rs
```diff
@@ -34,6 +38,63 @@ pub fn changes_in_commit(
...
+                (Some(branch), None)
+            } else {
+                (start, end)
+            }
+        });
+    let repo = ctx.gix_repository()?;
+
+    // Find the head that matches the branch name - the commit contained is our commit_id
+    let start_commit_id = repo
+        .find_reference(start.ok_or_else(|| anyhow::anyhow!("Branch {} not found", branch_name))?)?
```
Byron, line 81: Here it would be more intuitive to use `start.with_context(|| …)` I *think*.

crates/gitbutler-tauri/src/diff.rs
```diff
@@ -34,6 +38,63 @@ pub fn changes_in_commit(
...
+) -> anyhow::Result<Vec<TreeChange>, Error> {
+    let project = projects.get(project_id)?;
+    let ctx = CommandContext::open(&project, settings.get()?.clone())?;
+    changes_in_branch_inner(ctx, branch_name, stack_id).map_err(Into::into)
+}
+
+fn changes_in_branch_inner(
+    ctx: CommandContext,
+    branch_name: String,
+    stack_id: StackId,
```
Byron, line 58: I think right now it's hard to *not* use `StackId` as there is no real V3 counterpart for dealing with stacks or learning about them.

The `head_info()` function, for instance, would naturally return with a deterministic list of commits per stack-segment, which is exactly what's needed here - all it wants is the top-most and bottom-most commit belonging to a stack.

On the bright side, once that becomes available it should be quite straightforward to find all places with StackId and let them use the new APIs.

Byron (commented):
That's a cool new feature!

crates/gitbutler-tauri/src/diff.rs
krlvi: I forgot about that function. Will correct

PR #7615, opened 2025-03-13
use max_completion_tokens instead of max_tokens for OpenAI

vidluther:
## 🧢 Changes

max_tokens has been deprecated in favor of max_completion_tokens as described here: https://platform.openai.com/docs/api-reference/completions/create

### DIFF
--- a/apps/desktop/src/lib/ai/openAIClient.ts
+++ b/apps/desktop/src/lib/ai/openAIClient.ts
@@ -25,7 +25,7 @@ export class OpenAIClient implements AIClient {
 
 	async evaluate(prompt: Prompt, options?: AIEvalOptions): Promise<string> {
 		const response = await this.client.chat.completions.create({
-			max_tokens: options?.maxTokens ?? DEFAULT_MAX_TOKENS,
+			max_completion_tokens: options?.maxTokens ?? DEFAULT_MAX_TOKENS,
 			messages: prompt,
 			model: this.modelName,
 			stream: true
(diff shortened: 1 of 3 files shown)


Byron (approved):
Thanks a lot, looks good to me!
I couldn't try it though, so I will have to leave the merge to someone else.

apps/desktop/src/lib/ai/openAIClient.ts
```diff
@@ -24,8 +24,12 @@ export class OpenAIClient implements AIClient {
 	}
 
 	async evaluate(prompt: Prompt, options?: AIEvalOptions): Promise<string> {
+		// The 'max_tokens' parameter has been renamed to 'max_completion_tokens' in the OpenAI API.
+		// This change aligns with the updated API specification where 'max_completion_tokens'
+		// specifically controls the maximum number of tokens for the completion portion of the response.
+		// https://platform.openai.com/docs/api-reference/completions/create
```
estib-vega, line 30: Thanks for the docs here, could we just reduce this to this single line
```suggestion
		// https://platform.openai.com/docs/api-reference/chat/create?lang=node.js
```

I think we should only link to the updated docs, the rest of the comment is nice as ephemeral context in the PR for now

estib-vega (commented):
Tested and works perfectly, thanks!
Just one nit-picky comment

krlvi (approved):

krlvi:
Merging this so that we can kick off a build

PR #7618, opened 2025-03-13
allows a username other than git when using an ssh key

Hedroed:
## 🧢 Changes

Get username from ssh git URI and use it in the libgit2 API.

## ☕️ Reasoning

I have a use case where the ssh git URI use `gitea` as username and i need to use an ssh key.

### DIFF
--- a/crates/gitbutler-repo/src/credentials.rs
+++ b/crates/gitbutler-repo/src/credentials.rs
@@ -35,7 +35,7 @@ impl From<Credential> for git2::RemoteCallbacks<'_> {
                 key_path,
                 passphrase,
             }) => {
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot for contributing, it's much appreciated.

I also can't think of a reason to ignore the username provide in the URL when using SSH URLs and the `git2` backend.

@krlvi what do you think?

krlvi:
> Thanks a lot for contributing, it's much appreciated.
> 
> I also can't think of a reason to ignore the username provide in the URL when using SSH URLs and the `git2` backend.
> 
> @krlvi what do you think?

Yeah seems reasonable. I plan to make a release now _before_ this change, and then merge this, so that we can test this in the nightly for a day or so

PR #7623, opened 2025-03-13
Do not, ever, use login shells. (#7575)

Byron:
These could be anything, and don't have to be compatible to a posix shell at all or in the slightest.

There simply is no other way but to get the login environment from somewhere else if more terminal-like behaviour is desired.


krlvi (approved):

krlvi:
@Byron - here's a proposed middle ground approach https://github.com/gitbutlerapp/gitbutler/pull/7629

PR #7677, opened 2025-03-18
discard changes: hunks

Byron:

Follow-up of #7597.

### Tasks

* [x] figure out what's going on with the test and 'object not found' (and fix it)
* [x] fix content-conversions and hunk-references into said content for commit-engine 
* [x] fix tree/index/worktree bug
     - [x] assure tests for no-ops
     - [x] check with symlinks at least once
     - [x] ~~deduplicate match~~ - no need, will just be harder to read then

### For next PR

* [ ] reset changed hunks in a file
* [ ] merge https://github.com/GitoxideLabs/gitoxide/pull/1898 and refer to it from `main`.

### Notes for the reviewer

* I find it difficult to imagine how meaningful discarding parts of hunks in whole-file deletions and additions can be so I chose to disallow it. The UI can still offer discarding the single whole hunk that a deletion or addition of a file would come with, but internally it, as always, would discard the file instead. Discarding of lines or selections would have to be disallowed on the UI level though, as this is currently only available for modified (or renamed + modified) files.

### For the next PR

    - implement remaining ignored cases
* **gitoxide informs about diff-filter changes** and this information is passed to the frontend
* **reset individual lines** (but as free selections of sub-hunks)

### Shortcomings

### Out of Scope

### Research

* **V2 implementation**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1984,9 +1984,9 @@ dependencies = [
 
 [[package]]
 name = "flate2"
-version = "1.0.34"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a1b589b4dc103969ad3cf85c950899926ec64300a1a46d76c03a6072957036f0"
+checksum = "11faaf5a5236997af9848be0bef4db95824b1d534ebc64d0f0c6cf3e67bd38dc"
 dependencies = [
  "crc32fast",
  "libz-ng-sys",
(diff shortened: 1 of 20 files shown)


Byron:
@Caleb-T-Owens The issue mentioned in Discord is now fixed, and led to the realisation that an entire bag of problems with worktree changes was nearly entirely ignored previously. Now I see that what GitButler really is doing is to "pretend the index doesn't exist". For the typical user and GitButler-exclusive operation that is the case even, and the index is always matching the tree it was created from.
When users start to bypass GitButler, it's possible that `git add` is invoked which adds new complexity. Previously GitButler would just ignore these tree/index changes, but also erroneously show diffs only between the index and the worktree.
Now both views are merged in a match statement with 16 cases, with only 6 that I don't think can actually happen. 4 of these possible cases are a bit unclear to me still and I left them for another day.
It's notable that this merge essentially synthesises an index/worktree change.

Every handled case (even those that aren't implemented yet) has a specific test-case to trigger it, which makes deciding what should be done very easy.

Follow-up PRs have to deal with the `todo!()` cases, but additionally I am interested in seeing what the commit-engine does when these synthetic changes are passed in. The question is really around index handling, and I would assume that it won't work correctly. Yet my hope is that it 'somewhat' works and that it resolves itself without being (too) incorrect. After all, these cases can be considered rare for now, even though they are a possibility.

Thinking about it, now that this is implemented, the UI could be used for testing how certain edge-cases will work quite easily, the test-cases show how these states can be triggered.

And yes, I guess you can take this as invitation to break it :D.

@krlvi I am merging this without review as this is only refactors in preparation for chunk-handling, along with plenty of fixes.

Caleb-T-Owens:
Thanks @Byron.

I'm starting to use the v3 client where possible, so I can hopefully exercise most of these

Issue #7556, opened 2025-03-09
bug: use template toggle on / off should override content if blank


TGTGamer:
So, I have been using the AI Generation recently, and I like it - sometimes - sometimes it generates incorrect data. 

So I want to be able to delete the AI Content, and revert back to the template within the repository. At the moment I can't seem to do this via the UI. 

Probably don't want to override user content without a prompt or something, but if I select all and delete the content so it's blank, it should treat the form as empty / unmodified, and allow that template to be activated again.

Replication Steps: 

1. Create a branch
2. Create a commit (content irrelevant)
3. Click Create Pull Request
4. Click use PR template to toggle it on.
5. Click "AI Generation" to toggle it on.
6. Scroll to bottom of content box. Click Generate Description. 
7. Decide you don't like the description. 
8. Attempt to get back to your template..... 

https://github.com/user-attachments/assets/8217c9d8-8e05-42c8-a7be-99097be27ef4

Related: #4898 #5220

PR #7748, opened 2025-03-20
fix: ensure template toggle overrides blank content (#7556) only when…

TomasGonzalez:
Hello GitButler team! 👋
I'm Tomás, a software engineer who recently applied for the `Senior TypeScript Developer` position in Berlin. While familiarizing myself with your product, I noticed this issue (#7556 ) and wanted to contribute a small improvement. I'm genuinely excited about GitButler's mission and would love to join and contribute more in the future. 🚀

## 🧢 Changes
* Ensured that enabling the template toggle overrides blank content as expected.
* Adjusted logic to check if content is empty before applying the template.
* While investigating this, I realized that the Rust helper only checks .github/ for templates, even though GitHub supports templates in the root directory as well—this may be worth addressing separately.

## ☕️ Reasoning
The main fix ensures that when the template toggle is turned on, it properly overrides blank content as expected.

While working on this, I refactored the method for selecting a template for clarity while investigating why the templates I had in my root directory were not being read. I ensured the logic remains unchanged but easier to follow.

Additionally, I realized that the Rust helper only checks inside the directory specified for the current forge (in my case, .github) for templates, whereas GitHub itself recognizes template files located in the repository root as well. This might not be directly related to the issue at hand but could be worth investigating separately.

### DIFF
--- a/apps/desktop/src/components/PrTemplateSection.svelte
+++ b/apps/desktop/src/components/PrTemplateSection.svelte
@@ -38,17 +38,15 @@
 	}
 
 	$effect(() => {
-		if (templates) {
-			if ($lastTemplate && templates.includes($lastTemplate)) {
-				loadAndEmit($lastTemplate);
-			} else if (templates.length === 1) {
+		(() => {
+			if (!templates) return;
+			if ($lastTemplate && templates.includes($lastTemplate)) return loadAndEmit($lastTemplate);
+			if (templates.length === 1) {
 				const path = templates.at(0);
-				if (path) {
-					loadAndEmit(path);
-					lastTemplate.set(path);
-				}
+				if (!path) return;
+				return loadAndEmit(path), lastTemplate.set(path);
 			}
-		}
+		})();
 	});
 </script>
 
--- a/apps/desktop/src/components/PrTemplateSection.svelte
+++ b/apps/desktop/src/components/PrTemplateSection.svelte
@@ -38,15 +38,17 @@
 	}
 
 	$effect(() => {
-		(() => {
-			if (!templates) return;
-			if ($lastTemplate && templates.includes($lastTemplate)) return loadAndEmit($lastTemplate);
-			if (templates.length === 1) {
+		if (templates) {
+			if ($lastTemplate && templates.includes($lastTemplate)) {
+				loadAndEmit($lastTemplate);
+			} else if (templates.length === 1) {
 				const path = templates.at(0);
-				if (!path) return;
-				return loadAndEmit(path), lastTemplate.set(path);
+				if (path) {
+					loadAndEmit(path);
+					lastTemplate.set(path);
+				}
 			}
-		})();
+		}
 	});
 </script>
 
--- a/apps/desktop/src/lib/forge/prContents.svelte.ts
+++ b/apps/desktop/src/lib/forge/prContents.svelte.ts
@@ -1,4 +1,8 @@
-import { getEphemeralStorageItem, setEphemeralStorageItem } from '@gitbutler/shared/persisted';
+import {
+	getEphemeralStorageItem,
+	setEphemeralStorageItem,
+	clearEphemeralStorageItem
+} from '@gitbutler/shared/persisted';
 import type { DetailedCommit } from '$lib/commits/commit';
 
 const PERSITANCE_TIME_MIN = 5;
@@ -13,6 +17,7 @@ function getPersistedTitleKey(projectId: string, branchName: string) {
 
 export function setPersistedPRBody(projectId: string, branchName: string, body: string): void {
 	const key = getPersistedBodyKey(projectId, branchName);
+	if (!body) return clearEphemeralStorageItem(key);
 	setEphemeralStorageItem(key, body, PERSITANCE_TIME_MIN);
 }
 
--- a/apps/desktop/src/lib/forge/prContents.svelte.ts
+++ b/apps/desktop/src/lib/forge/prContents.svelte.ts
@@ -15,9 +15,13 @@ function getPersistedTitleKey(projectId: string, branchName: string) {
 	return 'seriesCurrentPRTitle_' + projectId + '_' + branchName;
 }
 
+export function clearPersistedPRBody(projectId: string, branchName: string): void {
+	const key = getPersistedBodyKey(projectId, branchName);
+	return clearEphemeralStorageItem(key);
+}
+
 export function setPersistedPRBody(projectId: string, branchName: string, body: string): void {
 	const key = getPersistedBodyKey(projectId, branchName);
-	if (!body) return clearEphemeralStorageItem(key);
 	setEphemeralStorageItem(key, body, PERSITANCE_TIME_MIN);
 }
 
@@ -124,6 +128,7 @@ export class ReactivePRBody {
 	}
 
 	reset() {
-		this.set('');
+		this._value = '';
+		clearPersistedPRBody(this.projectId, this.branchName);
 	}
 }
--- a/packages/shared/src/lib/persisted.ts
+++ b/packages/shared/src/lib/persisted.ts
@@ -54,6 +54,10 @@ export function persisted<T>(initial: T, key: string): Persisted<T> {
 	};
 }
 
+export function clearEphemeralStorageItem(key: string): void {
+	lscache.remove(key);
+}
+
 export function setEphemeralStorageItem(
 	key: string,
 	value: unknown,
--- a/apps/desktop/src/components/ReviewDetailsModal.svelte
+++ b/apps/desktop/src/components/ReviewDetailsModal.svelte
@@ -374,6 +374,27 @@
 		}, 2000);
 	}
 
+	function handlePRDescriptionFieldInput(
+		e: Event & { currentTarget: EventTarget & HTMLTextAreaElement }
+	) {
+		const target = e.currentTarget;
+		const isEmpty = !target.value;
+
+		if (isEmpty && $useTemplate && templateBody) {
+			// Force a reset to ensure an update even when the new value equals the current template
+			prBody.reset();
+			prBody.set(templateBody);
+			return;
+		}
+
+		if (isEmpty) {
+			prBody.reset();
+			return;
+		}
+
+		prBody.set(target.value);
+	}
+
 	export function show() {
 		modal?.show();
 	}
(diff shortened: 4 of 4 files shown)


apps/desktop/src/components/PrTemplateSection.svelte
```diff
@@ -38,17 +38,15 @@
...
 				const path = templates.at(0);
-				if (path) {
-					loadAndEmit(path);
-					lastTemplate.set(path);
-				}
+				if (!path) return;
+				return loadAndEmit(path), lastTemplate.set(path);
 			}
-		}
+		})();
```
estib-vega: Interesting.

If I'm reading correctly, the logic here isn't changed but the way it's written was.
You went for an iife in order to avoid the `if-else`s (which I also don't really like) by returning if the clauses are fulfilled.

I find it cool but in this case I'd say that it makes the code less readable.

We had a talk internally a few days ago about *labeled block statements* which I wanted to use in order to get rid of a chain of `if-else`s but in the end we want to keep thinks as clear as we can.

In this case is also a bit tricky since returning function calls makes it even more confusing in that it implies that some values are being returned.

Would like to hear your opinions on this, though :)

packages/shared/src/lib/persisted.ts
```diff
@@ -54,6 +54,10 @@ export function persisted<T>(initial: T, key: string): Persisted<T> {
 	};
 }
 
+export function clearEphemeralStorageItem(key: string): void {
+	lscache.remove(key);
+}
+
```
estib-vega, line 60: It's good to have that for sure 👍

apps/desktop/src/lib/forge/prContents.svelte.ts
```diff
@@ -13,6 +17,7 @@ function getPersistedTitleKey(projectId: string, branchName: string) {
 
 export function setPersistedPRBody(projectId: string, branchName: string, body: string): void {
 	const key = getPersistedBodyKey(projectId, branchName);
+	if (!body) return clearEphemeralStorageItem(key);
```
estib-vega: IMO this shouldn't be dealt with here.

The functions in this module are supposed to be straight forward and do only what their name imply.

Adding a 'hidden' side effect of clearing the stored PR body if the body is empty should should be replaced with some external handling.

Let me know what you think

estib-vega (commented):
Hola!

Thanks so much for your contribution. It's really appreciated :)
I left some feedback here for which I would be very interesting in hearing your thoughts about.

estib-vega:
The changes don't seem to solve the specified problem as described in the issue #7556. Deleting the content in the PR modal doesn't display the template again.

Look in the **prContents.svelte.ts** file. The fix should probably go in there :)
https://github.com/gitbutlerapp/gitbutler/blob/master/apps/desktop/src/lib/forge/prContents.svelte.ts

apps/desktop/src/components/PrTemplateSection.svelte
TomasGonzalez: Hey, thanks for the thoughtful feedback! 😊

You're absolutely right I didn't intentionally alter the logic. It was purely a stylistic refactor while diving into the code. My choice to use an IIFE was driven primarily by a personal preference for early returns, explicitly encapsulating side effects, and clearly highlighting the "happy path."

Personally, I find early returns beneficial because they explicitly state exit conditions upfront, helping me mentally parse the core logic faster without navigating nested conditional branches.

I hadn't considered using labeled block statements before, but now that you mention it, that's actually an interesting alternative, particularly as another clean approach to avoid deeply nested logic. I'm definitely open to exploring that more, especially if it aligns better with the team's readability standards.

I completely agree that clarity and maintainability should always come first, especially in collaborative environments. Happy to adjust the code back to the original nested structure or collaborate on an alternative approach. Let me know what you'd prefer!

apps/desktop/src/lib/forge/prContents.svelte.ts
TomasGonzalez: Good point, thanks for highlighting this!

My original thought here was that since the method already handles persisting the PR body, checking for empty states internally didn't feel entirely out of scope, it seemed like a natural part of maintaining consistency for how PR bodies are stored or cleared.

However, considering your perspective about keeping these methods strictly single-purpose and explicit, I actually agree with you. Moving this logic closer to the caller would indeed keep responsibilities clearer. I'm happy to adjust this! Let me know if you have additional suggestions.

TomasGonzalez:
> The changes don't seem to solve the specified problem as described in the issue #7556. Deleting the content in the PR modal doesn't display the template again.
> 
> Look in the **prContents.svelte.ts** file. The fix should probably go in there :) https://github.com/gitbutlerapp/gitbutler/blob/master/apps/desktop/src/lib/forge/prContents.svelte.ts

Hey, thanks for the feedback!

From my interpretation of the issue (#7556), the problem described was that after clearing the body content, the user couldn't re-activate the template again via the UI (i.e., toggling the template switch off and back on didn't restore the template). This happened because the previously saved (blank) body was persisted in local storage and prevented reloading the template.

My fix addresses this by ensuring that if the body content is cleared, the persistence state is also cleared—allowing the template to be toggled back on explicitly by the user afterward.

Automatically re-inserting the template immediately after clearing the body content might not be ideal, as clearing could be intentional. Additionally, it would cause unexpected behavior: if the user deliberately deletes the template text, it would immediately reappear.

My approach ensures the user explicitly decides when to restore the template, aligning more intuitively with typical user expectations.

That said, if the intended UX is to instantly re-display the template upon clearing the body, we could discuss adjustments in the prContents.svelte.ts component. Let me know your thoughts!

apps/desktop/src/lib/forge/prContents.svelte.ts
TomasGonzalez: I've made the adjustment—hope this version fits more cleanly with the structure you had in mind. Open to feedback if you spot anything else!

apps/desktop/src/components/PrTemplateSection.svelte
TomasGonzalez: I’ve reverted the refactor and kept the original structure, makes sense to stick with what’s most readable for the team.

estib-vega:
Hey Tomas, thanks for addressing the feedback!

> My fix addresses this by ensuring that if the body content is cleared, the persistence state is also cleared—allowing the template to be toggled back on explicitly by the user afterward.

I see. Yes, that works as specified 👍 

> Automatically re-inserting the template immediately after clearing the body content might not be ideal, as clearing could be intentional. Additionally, it would cause unexpected behavior: if the user deliberately deletes the template text, it would immediately reappear.
>
> My approach ensures the user explicitly decides when to restore the template, aligning more intuitively with typical user expectations.

This is a good point. I guess is about picking which one is the *least unexpected* behavior:
- Not being able to clear the content if you have the templates on
- Having to turn off and on the templates in order to reset the template

My gut feeling is that the former is the preferable one. 
If the user wants to clear the body content, they also should disable the template option.

What do you think? If you agree, could you adjust it so that it behaves like that instead?

TomasGonzalez:
> This is a good point. I guess is about picking which one is the _least unexpected_ behavior:
> 
> * Not being able to clear the content if you have the templates on
> * Having to turn off and on the templates in order to reset the template
> 
> My gut feeling is that the former is the preferable one. If the user wants to clear the body content, they also should disable the template option.
> 

Thanks for the follow-up! I totally get where you're coming from, this does feel like a case of choosing the least unexpected behavior for users.

Thinking about it a bit more though, I wonder if the real issue might be the UI model itself. Having a toggle control whether the template is applied feels a bit limited, especially since users can already select a specific template manually.

Maybe a cleaner approach would be to let users re-apply the selected template from the list at any time, rather than relying on the toggle to "force" it. That way the action is explicit, there’s no risk of overwriting on accident.

Totally open to going with your suggestion if that’s the direction you prefer, just wanted to throw this in as an alternative in case it resonates :)

estib-vega:
> Thinking about it a bit more though, I wonder if the real issue might be the UI model itself

It may be that. How I think of the current way the UI presents this option, is that the template is just the persisted starting point of whatever body content you want input/generate. 
In that 'flow' I think is okay to:
- Have the toggle and dropdown to select the whether and which template to use.
- Have the 'empty' mode always be the template selected.

That said, maybe a completely different UI layout might be less cumbersome.

If you're still willing to indulge this further I propose the following:
If you think of a better way of working with the PR templates I'll bring it up with @PavelLaptev and we can try to update that.
Alternatively let's go with having the 'empty state' be the template, if the toggle is enabled.

What do you think?

PavelLaptev:
Here’s an updated [design in Figma](https://www.figma.com/design/ShIiR6hI5dzH7sT5L03Jg1/App-Design-3.0?node-id=5876-104676&t=C874X8MoArTWHdmx-1) for the "Submit for Review" modal. It’s now a unified flow for both "Create a PR" and "Create a BR," rather than just a PR creation.

- Removed the toggle; if PR templates are detected, we now display the selector.
- Open question: Should we preselect the first template or keep "Please select a template"?

image

estib-vega:
@PavelLaptev Nice!

> Open question: Should we preselect the first template or keep "Please select a template"?

I think we shouldn't preselect a template. Instead, we can have an initial option `No template` and if the user decides to use one, then we persist that choice, as we do in v2.

I think it shouldn't say `Please select a template` since is not a requirement

TomasGonzalez:
Hey @PavelLaptev, this updated design looks awesome and super clean, I’m definitely in favor of dropping the toggle.

Totally agree that “No template” makes sense as the default, and no need for a “Please select…” prompt.

One small thing I’d love to confirm: would it be possible for the dropdown to allow re-selecting the same template to overwrite the body? That way, if I clear the text and want to bring the template back, I can just select it again. I prefer that to it auto-repopulating when the field is empty but happy to go with whatever fits best!

Let me know what you think and I’ll get started on the implementation :)

TomasGonzalez:
@PavelLaptev Also just a thought, maybe the AI toggle could eventually live in the same select as the templates? Since both are ways to generate starting content.

TomasGonzalez:
Hey @estib-vega  Just a quick update, I went ahead and implemented the behavior as originally discussed: clearing the body re-applies the template if the toggle is still on.

I’d love to explore the updated design further, but realistically I don’t have the bandwidth to take that on unless I were more deeply involved with the project.

That said, happy to offer thoughts or feedback if it moves forward later on :)

PavelLaptev:
@TomasGonzalez Thanks! We had a quick discussion about the "Submit for Review" modal design. Just small UI updates. I'll add a link here once I’ve updated it in Figma.

TomasGonzalez:
@PavelLaptev Great, feel free to ping me when it’s ready!

PavelLaptev:
Hey 😊 The current "Submit for Review" design is [here in Figma](https://www.figma.com/design/ShIiR6hI5dzH7sT5L03Jg1/App-Design-3.0?node-id=5876-104676&t=1hEY1ntPb9oH7XrV-1).  
Also, here's a [separate page](https://www.figma.com/design/ShIiR6hI5dzH7sT5L03Jg1/App-Design-3.0?node-id=6422-33818&t=1hEY1ntPb9oH7XrV-1) showing how "/" commands and popup menus look.

Byron:

It does, however, look like the PR was abandoned, as it doesn't merge cleanly and wasn't updated in quite some time.

PR #7749, opened 2025-03-20
Handle branch removal in v3

Caleb-T-Owens:


crates/gitbutler-branch-actions/src/branch_manager/branch_removal.rs
```diff
@@ -95,51 +96,100 @@ impl BranchManager<'_> {
...
+
+            let mut merge = gix_repo.merge_trees(
+                stack_head,
+                cwdt,
+                workspace_base,
+                gix_repo.default_merge_labels(),
+                merge_options,
+            )?;
+            let tree = merge.tree.write()?;
+            let tree = repo.find_tree(tree.object()?.id().to_git2())?;
```
Byron: `tree` is already a tree-id, so there should be no need for fetching its `object()` beforehand.

PR #7797, opened 2025-03-24
Fix a bug while rebasing

krlvi:


krlvi:
@Byron there was a bug where if during workspace update (rebasing the branch) there was a conflicted commit, the commit that follows had an incorrect tree, in some cases.
In this fix I re-implemented `find_cherry_pick_trees` in the spirit of the implementation that is in the `gitbutler-cherry-pick` crate

PR #7807, opened 2025-03-25
discard changes: hunk selections

Byron:

Follow-up of #7806.

### Tasks

* [x] hunk-selections for discard
     - [x] make hunk *section* specification clear
     - [x] old-new hunk-subtraction
        - [x] all relevant tests
        - [x] Show what happens if hunk selections aren't sorted (it should work, and keep working)
* [x] hunk-header support when discarding in the CLI

### Notes for the reviewer

* I find it difficult to imagine how meaningful discarding parts of hunks in whole-file deletions and additions can be so I chose to disallow it. The UI can still offer discarding the single whole hunk that a deletion or addition of a file would come with, but internally it, as always, would discard the file instead. Discarding of lines or selections would have to be disallowed on the UI level though, as this is currently only available for modified (or renamed + modified) files.

### For the next PRs

* ⚠️ **gitoxide informs about diff-filter changes** and this information is passed to the frontend
* **hunk-selections for committing**
    - Additive-only should trivially work by allowing sub-hunks.
* **reset individual lines** (but as free selections of sub-hunks)
    - How does this work with context lines? In theory, they will mess everything up so it's probably something to avoid. After all, they will combine multiple hunks into one and the UI really would have to provide hunks without them.
    - implement remaining ignored cases

### Shortcomings

### Out of Scope

### Research

* **V2 implementation**


Byron:
Hi @estib-vega, I noticed you have been dealing with hunk selections lately, so here is an example on how hunk-selections work for discard (but committing will be the same):

https://github.com/gitbutlerapp/gitbutler/blob/47b86e156741307df4c69de6404167998b2091d6/crates/but-workspace/tests/workspace/discard/hunk.rs#L368-L384

The idea is that discarding added lines anchors the selection on the old patch header, and vice-versa.
There is no need to concatenate selections, it's fine to pass single-line selections using the line numbers that the UI already knows.

If there are any questions, please let me know :).

Byron:
@krlvi I think this can be merged and hooked up the UI right away. However, please note the intentional limitation of refusing to discard hunks when the whole file was added or deleted (see "Notes for the Reviewer" for details). It should be no problem to allow it though if this is a requirement.

PR #7813, opened 2025-03-25
Fix an unwrap situation

krlvi:


crates/but-workspace/src/lib.rs
```diff
@@ -467,7 +467,10 @@ pub fn stack_branch_local_and_remote_commits(
     let stack = state.get_stack(Id::from_str(&stack_id)?)?;
 
     let branches = stack.branches();
-    let branch = branches.iter().find(|b| b.name() == &branch_name).unwrap(); //todo
+    let branch = branches
+        .iter()
+        .find(|b| b.name() == &branch_name)
+        .ok_or(anyhow::anyhow!("Could not find branch {:?}", branch_name))?;
```
Byron, line 473: This should be `with_context(|| format!(…))` as it will only run when there is an actual error, lazily.
An alternative to that is `.ok_or_else(|| anyhow::anyhow!(…))`, but that seems more noisy.

krlvi: Thanks! I keep forgetting to do that. I wish there was some kind of lint rule for these

krlvi: fixed here https://github.com/gitbutlerapp/gitbutler/pull/7851

PR #7637, opened 2025-03-14
fix: no hardcoded copyright year

et0and:
## 🧢 Changes

Uses a new Date() in place of a hardcoded copyright year.

## ☕️ Reasoning

Won't be outdated in the future! Currently set to 2023.

PR #7820, opened 2025-03-25
fix: update copyright year in tauri configuration

H1ghBre4k3r:
Updates the copyright year in the tauri configuration file  to reflect the current year, ensuring compliance with  intellectual property standards.


Byron:
You might find this PR interesting: #7637 - it automates setting the year. Maybe that can be adopted for this instance of copyright year as well?

H1ghBre4k3r:
@Byron I think, we could apply some kind of maintenance script or something like this to update this field in the configuration (which is kinda tedious). 

However, I am not sure whether we can set this information programmatically (e.g., via a `new Date()....` snippet.

PR #7823, opened 2025-03-25
return the review IDs in the integration outcome

estib-vega:
Integration outcome returns information about the review IDs that should be closed, if any


krlvi (approved):
lgtm

PR #7824, opened 2025-03-25
fix: auto detect system os to download correct target

### LINKED ISSUE #7772: Auto choose download package on website based on user OS.
Hi, I was just checking out GitButler for the first time. I downloaded it assuming it would be set for Windows, but saw it downloaded the dmg. I noticed that by default Apple is selected for the download button. 

Instead it would be great if it was changed such that if the default OS being chosen is based on the Client visiting the website. (By default people tend to assume, that the website would give the OS appropriate version for the default download)

Thanks!

(I think the tag would be @gitbutler/website?)

wleklinskimateusz:
## 🧢 Changes
- add function that triggers on CtaBlock component mount, that detects default download target for OS of the user
- extend necessary global interface according to this: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData - this allows to differentiate between Apple Silicon and Apple Intel 

## ☕️ Reasoning

- For Apple if available look for architecture: if something other than arm - change the target to Intel, if arm or architecture not available, stay with Apple Silicon
- For linux, I've set it to AppImage by default
- For windows it's set to windows - (BUT - I haven't had a chance to test it on windows, so it would appreciated)

If you don't want to use experimental browser features - I change it to: for all MacOS devices, default to Apple Silicon - we won't need to extend Navigator interface and it would simplify the logi

## 🎫 Affected issues

Fixes: #7772

### DIFF
--- a/apps/web/src/app.d.ts
+++ b/apps/web/src/app.d.ts
@@ -9,6 +9,21 @@ declare global {
 		// interface PageState {}
 		// interface Platform {}
 	}
+
+	declare interface Navigator {
+		userAgentData?: {
+			getHighEntropyValues(hints: string[]): Promise<{
+				readonly brands?: { readonly brand: string; readonly version: string }[];
+				readonly mobile?: boolean;
+				readonly platform?: string;
+				readonly architecture?: string;
+				readonly bitness?: string;
+				readonly formFactor?: string[];
+				readonly model?: string;
+				readonly platformVersion?: string;
+			}>;
+		};
+	}
 }
 
 export {};
--- a/apps/web/src/routes/(home)/components/CtaBlock.svelte
+++ b/apps/web/src/routes/(home)/components/CtaBlock.svelte
@@ -5,6 +5,7 @@
 	import { latestClientVersion } from '$lib/store';
 	import { quadIn } from 'svelte/easing';
 	import { fly } from 'svelte/transition';
+	import { onMount } from 'svelte';
 
 	let videoElement = $state<HTMLVideoElement>();
 	interface Props {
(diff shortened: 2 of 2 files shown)


apps/web/src/routes/(home)/components/CtaBlock.svelte
```diff
@@ -45,6 +46,33 @@
 	function handleClickOutside() {
 		showSelect = false;
 	}
+
+	async function detectDefaultDownload() {
```
Byron, line 50: Should that not be made typesafe and define the return value?

apps/web/src/routes/(home)/components/CtaBlock.svelte
```diff
@@ -45,6 +46,33 @@
...
+		const platform = navigator.userAgent.toLowerCase();
+
+		if (platform.includes('mac')) {
+			const userAgentData = await navigator.userAgentData?.getHighEntropyValues(['architecture']);
+			if (userAgentData?.architecture === 'arm' || userAgentData === undefined) {
```
Byron, line 55: If these terms were swapped I could imagine TS being smart enough to allow doing `userAgentData.architecture`.

Byron (changes requested):
Thanks so much for getting this started!

There is one potential issue with calling `targetDownload.set(undefined)`, but I'd hope that improving the type definition would catch the problem if it is one.

apps/web/src/app.d.ts
```diff
@@ -9,6 +9,21 @@ declare global {
 		// interface PageState {}
 		// interface Platform {}
 	}
+
+	declare interface Navigator {
+		userAgentData?: {
+			getHighEntropyValues(hints: string[]): Promise<{
```
Caleb-T-Owens, line 15: It's not impossible that a browser might implement `navigator.userAgentData` before `getHighEntropyValues`, so we should declare the `getHighEntropyValues` method as being optional

Caleb-T-Owens: We could also type the function like this so we can always access the returned properties:


```ts
type Hints = {
	readonly brands: { readonly brand: string; readonly version: string }[];
	readonly mobile: boolean;
	readonly platform: string;
	readonly architecture: string;
	readonly bitness: string;
	readonly formFactor: string[];
	readonly model: string;
	readonly platformVersion: string;
};

declare global {
	// ...
	declare interface Navigator {
		userAgentData?: {
			getHighEntropyValues?: <T extends [keyof Hints, ...(keyof Hints)[]]>(
				hints: T
			) => Promise<Pick<Hints, T[number]>>;
		};
	}
}
```

Usage:

Screenshot 2025-03-27 at 11 00 35

apps/web/src/routes/(home)/components/CtaBlock.svelte
```diff
@@ -45,6 +46,33 @@
...
+		if (platform.includes('linux')) {
+			return jsonLinks.downloads.linuxAppimage;
+		}
+	}
+
+	onMount(() => {
+		detectDefaultDownload().then(
+			(platformDownload) => platformDownload && targetDownload.set(platformDownload)
+		);
+	});
```
Caleb-T-Owens, line 75: We should put the logic to determine the default inside `store.ts` so it doesn't get lost if we ever replace this component.

Caleb-T-Owens (changes requested):
Thank you for the contribution! I've provided some feedback

Byron:

It does, however, look like the PR was abandoned, or it doesn't merge cleanly and wasn't updated in quite some time.

PR #7847, opened 2025-03-26
Fix crediting gitbutler in commits

Caleb-T-Owens:


crates/but-core/src/repo_ext.rs
```diff
@@ -77,7 +77,7 @@ impl RepositoryExt for gix::Repository {
             .context("No author is configured in Git")
             .context(Code::AuthorMissing)?;
 
-        let commit_as_gitbutler = !self
+        let commit_as_gitbutler = self
```
Byron, line 80: Oh my 🤦‍♂️.

Byron (approved):

PR #7849, opened 2025-03-27
fix: rename newBranchName to newName

H1ghBre4k3r:
## 🧢 Changes

Change the name of the argument passed to `update_branch_name` in `branchController.ts` to `newName` to reflect the naming on the Rust/Tauri side. 

## ☕️ Reasoning

Tauri uses named arguments for Tauri commands. Thus, you need to name each property in the passed object like the "counter part" on the rust side.


krlvi (approved):
Thank you for this fix! <3

PR #7852, opened 2025-03-27
fix series icon

### LINKED ISSUE #7850: The icon / tooltip for "single branch" vs "multiple branches" are swapped
### Version

0.14.13

### Operating System

Linux

### Distribution Method

AppImage (Linux)

### Describe the issue

These image and tooltips are presumably intended to show what virtual branch / lanes are single branches vs. stacked branches. But they're swapped from what they're supposed to be! This also shows up on the left hand sidebar incorrectly.

Image

Image

_No response_

_No response_

```shell

```

estib-vega:
fixes #7850

### DIFF
--- a/packages/ui/src/lib/SeriesIcon.svelte
+++ b/packages/ui/src/lib/SeriesIcon.svelte
@@ -12,7 +12,7 @@
 
 <Tooltip text={single ? 'Single branch' : 'Multiple branches'}>
 	<div class="stack-icon" class:outlined>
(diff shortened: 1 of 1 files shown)


Psykar:
@estib-vega I suspect this isn't solving the whole issue. The text matched the icon just fine, but the tooltip and the icon both implied multiple branches, but on a single branch - rather that the setting of the `single` variable is reversed?

PR #7856, opened 2025-03-27
Move focus manager event listening to last in line

mtsgrd:
Before this change the focus manager listened to the document in the capture phase, meaning it fired before anything else. This, however, prevents us from stopping propagation e.g. when resizing the commit drawer.

After this change it listens to the document still, but in the bubble phase, giving all other events a possibility to cancel the change in focus.


mtsgrd:
@Caleb-T-Owens fyi!

PR #7893, opened 2025-03-30
commit hunk selections

Byron:

Follow-up of #7807.

### Tasks

* [x]  **gitoxide informs about diff-filter changes** and this information is passed to the frontend
* [x] use latest `gix` from `main`
* [x] hunk-selections for committing
    - [x] trivial selections
    - [x] assure multiple selections per hunk actually work like they should (which won't work out of the box for sure)
    - [x] A UI-style selections with lots of one-liners - result should be the same no matter what (high-level)
    - [x] test for discarding only hunks that aren't matching, but commit the hunks that matched.
* [x] `but-cli` supports committing hunk headers

### Next PR

* https://github.com/gitbutlerapp/gitbutler/pull/7952#issue-2967230532
    - Why does it do the wrong thing once [these lines](https://github.com/gitbutlerapp/gitbutler/blob/92e722ddade85a195c7e7041317fbb3417467d56/crates/gitbutler-tauri/src/workspace.rs#L164-L167) are removed.
    - problem is that the passed workspace commit is already only having one parent only, even though it should have two.
          - Once that is clear, it should fail as it needs a base-substitute.
* **Discard hunk logic** can probably use the context-window = 0 upgrade, right now splits are computed even though I doubt this is always correct.

### DIFF
--- a/crates/but-cli/src/args.rs
+++ b/crates/but-cli/src/args.rs
@@ -28,6 +28,19 @@ pub struct Args {
 pub enum Subcommands {
     /// Commit or amend all worktree changes to a new commit.
     Commit {
+        /// The repo-relative path to the changed file to commit.
+        #[clap(requires_if(clap::builder::ArgPredicate::IsPresent, "hunk_headers"))]
+        current_path: Option<PathBuf>,
+        /// If the change is a rename, identify the repo-relative path of the source.
+        previous_path: Option<PathBuf>,
+        /// The 1-based pairs of 4 numbers equivalent to '(old_start,old_lines,new_start,new_lines)'
+        #[clap(
+            long,
+            requires_if(clap::builder::ArgPredicate::IsPresent, "current_path"),
+            num_args = 4,
+            value_names = ["old-start", "old-lines", "new-start", "new-lines"])
+        ]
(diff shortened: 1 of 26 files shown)


crates/but-core/src/lib.rs
```diff
@@ -140,6 +140,9 @@ pub enum UnifiedDiff {
     Patch {
         /// All non-overlapping hunks, including their context lines.
         hunks: Vec<unified_diff::DiffHunk>,
+        /// If `true`, a binary to text filter (`textconv` in Git config) was used to obtain the `hunks` in the diff.
+        /// This means hunk-based operations must be disabled.
+        is_result_of_binary_to_text_conversion: bool,
```
Byron: Hi @estib-vega , here is the promised flag that the UI would use to determine if hunk-selections of any kind should be allowed. That's the case only if this is `false`.

estib-vega: Could you add a serde rename to that it's camelCased, please? :)

Byron: Hehe, right, we didn't notice before because it was naturally camelCase.
Should be fixed now.

PR #7895, opened 2025-03-31
Unified diff: File deletions can't be partially selected

estib-vega:
Don’t let the user partially select the hunk if they are file deletions

TODO: don't allow partial selections on diff-filtered hunks
See #7893


Byron:
This is something that could also be relaxed if there was value in it.

When committing for instance, there is no such restriction yet, and I guess we will just try it to see how it feels before conforming the behaviour across operations.

Byron:
Exactly that. It's kind of… strange, but maybe useful?

estib-vega:
> Exactly that. It's kind of… strange, but maybe useful?

If that's the case, I'll undo it. That sounds cool

Byron:
Just to be clear: For discarding it won't work as the backend doesn't currently allow it, but the commit-engine doesn't have this limitation built-in. There is where that capability could be tried.

PR #7902, opened 2025-03-31
Fix commting with one stack

Caleb-T-Owens:


Caleb-T-Owens:
FYI @Byron

Byron:
The test-failure seems to be flake, and I have seen that too in other PR.
And since the removed lines don't trigger a legit test failure, I presume there is no reason to keep it. Further, I think with #7596 merged, the handling of 'framings' would be centralised and well tested.

Now I think it's time to fix the likely source of the issue, and finish https://github.com/GitoxideLabs/gitoxide/pull/1851 which will remove parallel iteration support.

Byron (approved):
While https://github.com/GitoxideLabs/gitoxide/pull/1851 is waiting for CI to be merged, I am rerunning CI here and merging it assuming it will go through this time.
The likely `gitoxide` fix will then be coming in with #7902.

Caleb-T-Owens:
Hey @Byron,

Here is a bit of a fuller explanation as to why I made this change. (I wrote this explanation two times already and v3 deleted them so I gave up and just pushed).

We were getting what seemed to be this out-of-sync error. As it turns out, there are some cases where we're not calling `update_workspace_commit` which results on the parent's of the workspace commit not always being a direct analogue for the heads of the stacks.

While we obviously need to tighten up where we call `update_workspace_commit`, it also seems reasonable to always try and get the stack head from a more reliable location, which is the stack head.

It also removed a code branch which makes everyone happy.

PR #7931, opened 2025-04-01
Toggling between v3 and v2 UI shouldn't screw up the workdir

krlvi:
Ideally we should do this for all projects. Let's first validate that this works


apps/desktop/src/lib/config/appSettingsV2.ts
```diff
@@ -1,4 +1,7 @@
 import { listen, invoke } from '$lib/backend/ipc';
+const projectsService = getContext(ProjectsService);
```
mtsgrd, line 2: Hmm, linter should have caught this..

PR #7952, opened 2025-04-02
Fix an issue with v3 comitting

krlvi:
It appears that something is not correctly updated by the commit engine in v3. This change seems to make things work.
Before this change we have been observing strange behavior in the commit engine (illustrated in the video)
@Byron there is probably a more elegant way of solving this - I have not yet understood exactly where the problem is, and I am not sure if this mitigates 100% of the issues, but trying it out until we understand this better.

https://github.com/user-attachments/assets/c4814d68-845f-45bf-a5f0-f1a99b29b8d3

### DIFF
--- a/crates/gitbutler-tauri/src/workspace.rs
+++ b/crates/gitbutler-tauri/src/workspace.rs
@@ -1,5 +1,6 @@
 use crate::error::Error;
 use crate::from_json::HexHash;
+use anyhow::Context;
 use but_hunk_dependency::ui::{
     hunk_dependencies_for_workspace_changes_by_worktree_dir, HunkDependencies,
 };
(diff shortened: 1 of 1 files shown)


krlvi:
also FYI @Caleb-T-Owens

Byron:
Thanks for reeling me in and for explaining the issue!
I made a note in [my current PR](https://github.com/gitbutlerapp/gitbutler/blob/92e722ddade85a195c7e7041317fbb3417467d56/crates/gitbutler-tauri/src/workspace.rs#L164-L167) and want to get that into a mergeable state today. Then I will look at this to solve the *missing parent of the workspace commit* :).

PR #7960, opened 2025-04-03
yeeehhh

Caleb-T-Owens:


Byron:
The only thing worse than `bash`-scripts is `bash`-scripts that can't be run locally 🔨💥.

PR #8006, opened 2025-04-05
Add lines added/removed to the tree_change_diffs API

krlvi:


crates/but-core/src/unified_diff.rs
```diff
@@ -215,6 +221,21 @@ impl UnifiedDiff {
     }
 }
 
+fn compute_line_changes(hunks: &Vec<DiffHunk>) -> (u32, u32) {
```
krlvi, line 224: @Byron is there a better way / place to do this?

Byron: I think this is a fine way - straightforward and it actually re-uses work that was already done (i.e. the patch itself).
I don't think I would have done it differently.

Byron (approved):
Let me merge this so it can be seen by #8008.

PR #8008, opened 2025-04-06
various improvements

Byron:
### Tasks

* [x] gix upgrade for security fix
    - [x] update `gix-features` when available
* [x] see if stats-performance can be improved by re-using a tree-diff.


Byron:
@krlvi Now that the existing changes as computed by diffing the trees are re-used it became clear the previously the line-stats were inversed. However, it's probably worth double-checking in the UI just to be sure.

PR #8012, opened 2025-04-06
But-CLI JSON

estib-vega:


krlvi:
i was thinking about the MCP yesterday - I wonder if it would be easier if we had our tauri commands served on a second endpoint - a regular http server locally. That way it could be less work to maintain as everything will be the same as the tauri commands, just served in one additional way.

estib-vega:
> i was thinking about the MCP yesterday - I wonder if it would be easier if we had our tauri commands served on a second endpoint - a regular http server locally. That way it could be less work to maintain as everything will be the same as the tauri commands, just served in one additional way.

@krlvi That would probably be the better way, yeah

Byron:
> i was thinking about the MCP yesterday - I wonder if it would be easier if we had our tauri commands served on a second endpoint - a regular http server locally. That way it could be less work to maintain as everything will be the same as the tauri commands, just served in one additional way.

As a developer tool, to me it's valuable to be able to do something outside of anything `tauri`. Also, a future production/user-level CLI *should* work without building `tauri` or having to use anything else than plumbing crates. 
Maybe with MCP you are thinking more along the lines of *certain* endpoints being served by a shared implementation.

Anyway, just thought I'd share my thoughts real quick.

estib-vega:
> As a developer tool, to me it's valuable to be able to do something outside of anything tauri

That's a good point too.
The price to pay for that would be that we'd need to extend the maintenance surface area

estib-vega:
I'll continue the avenue of accessing the CLI for now, since that seems like the faster path for now

Byron:
And I am probably also missing a piece of the conversation, so instead of appearing to be against something, I am definitely for making `but-cli` a great development tool. For me it works as is, but I am sure there are other angles too that are worth considering.

krlvi:
I think there are two different use cases hare and we should be able to hit both. One being CLI access to Gitbutler features, which is valuable in its own right. 
And a brand new use case is supporting MCP - for which, we can pick whichever implementation is more appropriate :) 

Maybe it will be easy to get started with but-cli, but a lot of the functionality has not been added to it.

I am happy to help out with getting a MCP thing out whichever approach we take.

estib-vega:
> And I am probably also missing a piece of the conversation, so instead of appearing to be against something, I am definitely for making `but-cli` a great development tool. For me it works as is, but I am sure there are other angles too that are worth considering.

The gist of the topic is that we want to add a [MCP](https://modelcontextprotocol.io/introduction) Server for GitButler, so that e.g. IDEs can be aware of the GitButler workspace state and even manipulate it if necessary.

In order to do so, though, there server needs to access GitButler APIs somehow. I chose to initially connect it to the CLI and Kiril brought up creating a secondary set of localhost API endpoints that the application would access, so that we get the APIs for free and don't have to keep adding things to the CLI.

---

I think both have pros and cons, but as a tangential takeaway, I also think that just having a great CLI tool would be nice :)

estib-vega:
@krlvi @Byron I'll merge this and see how we feel.

This should not disrupt anything, so even if we later decide to not use this for MCP, I think to have some JSON output options are a nice-to-have addition.

Byron:
I understand now, thanks for elaborating. The way that seems intuitive to me is to have a crate with MCP-enabled functionality in a format that is native to that protocol, and then expose it 'in a way that makes sense'. Can be a port on local-host, or a CLI that spits out the MCP results as JSON, for example.
It's based on my very partial MCP knowledge though, I hope that's disclaimer enough 😄.

PR #8027, opened 2025-04-07
Message editor: don't hide the buttons

estib-vega:
The buttons and other content is not hidden if the text editor content is too big to fit in the drawer.
Instead, make the content scrollable


estib-vega:
@PavelLaptev FYI, this are the changes that we discussed :)

- Disable the drawer-level scroll for the drawers that have a message editor
- Make the Message editor scrollable
- Increase the minimal resizable height for the drawers that have a message editor

PR #8094, opened 2025-04-09
fix faulty ordering in vendored notify debounce (#8026)

### LINKED ISSUE #8026: Crashes during large builds inside a repository (notify-rs)
### Version

0.14.14

### Operating System

Linux

### Distribution Method

Built from source

### Describe the issue

GitButler monitors the contents of .gitignored directories. When I build electron or the kernel. I have to switch GitButler to another repository or it will run ~~out of memory~~ tracking all the "useless" build artefacts. **EDIT**: It just randomly crashes in notify-rs.

As I already hinted at, ignoring .gitignored directories would be a nice way to solve the problem.

EDIT: I have to explain why I say ignored directories. I build in the context of a distribution so the whole electron/kernel build is in a directory that is .gitignored. If one would checkout the kernel and build - GitButler also has to ignore .gitignored patterns.

Do a large build

~~Not use up all the memory.~~ Not crash.

```shell

```

ghost:
ported from upstream fix notify-rs/notify#638

## 🧢 Changes

Fix sorting of rename events

## ☕️ Reasoning

The problem was fixed upstream: https://github.com/notify-rs/notify/pull/638

Fixes: #8026

### DIFF
--- a/crates/gitbutler-watcher/vendor/debouncer/src/lib.rs
+++ b/crates/gitbutler-watcher/vendor/debouncer/src/lib.rs
@@ -37,7 +37,8 @@ mod tests;
 #[cfg(not(test))]
 use std::time::Instant;
 use std::{
-    collections::{HashMap, VecDeque},
+    cmp::Reverse,
+    collections::{BinaryHeap, HashMap, VecDeque},
     path::PathBuf,
     sync::{
         atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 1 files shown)


krlvi (approved):
Thank you @rhizoome for taking the time to investigate this and backport the fix!

Byron:
Yes, I can only second that, thanks so much!

This makes me want to unvendor our code actually - we did it for Windows, but in V3 there might actually not be a need for that anymore.
Probably something to keep in mind as we make progress.

PR #8120, opened 2025-04-11
Refactor `useMutation` and prefer `mutate`

mtsgrd:
Both mutate and the trigger of useMutation will now throw if an error occurs.

Also, we should prefer mutate over let [trigger] = something.useMutation() since they do the same work, but one is free of side effects.

### DIFF
--- a/apps/desktop/src/components/CommitCard.svelte
+++ b/apps/desktop/src/components/CommitCard.svelte
@@ -146,17 +146,17 @@
 	let isUpdating = $state(false);
 
 	async function submitCommitMessageModal() {
+		if (!stack) {
+			return;
+		}
 		isUpdating = true;
 		commit.description = description;
-
-		if (stack) {
-			await updateCommitMessage({
-				projectId: project.id,
-				stackId: stack.id,
-				commitId: commit.id,
-				message: description
-			});
-		}
+		await updateCommitMessage({
+			projectId: project.id,
+			stackId: stack.id,
+			commitId: commit.id,
+			message: description
+		});
 
 		commitMessageModal?.close();
 		isUpdating = false;
--- a/apps/desktop/src/components/v3/CommitView.svelte
+++ b/apps/desktop/src/components/v3/CommitView.svelte
@@ -85,22 +85,12 @@
 			return;
 		}
 
-		const result = await updateCommitMessage({
+		const newCommitId = await updateCommitMessage({
 			projectId,
 			stackId,
 			commitId: commitKey.commitId,
 			message: commitMessage
 		});
-
-		if (!result.data) {
-			showToast({
-				message: `Update commit error`,
-				style: 'error'
-			});
-			return;
-		}
-
-		const newCommitId = result.data;
 
 		uiState.stack(stackId).selection.set({ branchName, commitId: newCommitId });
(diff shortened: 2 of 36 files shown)


apps/desktop/src/lib/state/customHooks.svelte.ts
```diff
@@ -224,22 +224,16 @@ export function buildMutationHooks<Definitions extends EndpointDefinitions>({
...
-				sideEffect?.(result.data, queryArg);
-			}
-
-			if (result.error && onError) {
-				onError(result.error, queryArg);
-			}
-
+			const result = await promise.unwrap().catch((error) => {
+				onError?.(error, queryArg);
+			});
```
mtsgrd, line 235: Does this result still throw or does `catch` swallow it?

apps/desktop/src/components/v3/NewCommitView.svelte
```diff
@@ -50,10 +50,14 @@
...
 
-	const defaultBranchName = uiState.global.draftBranchName.current;
-	const canCommit = $derived((branchName || defaultBranchName) && selection.current.length > 0);
+	const draftBranchName = $derived(uiState.global.draftBranchName.current);
+	const canCommit = $derived(
+		(selectedBranchName || draftBranchName) && selection.current.length > 0
+	);
 	const projectState = $derived(uiState.project(projectId));
 
+	$inspect(draftBranchName);
```
estib-vega, line 59: Inspect 🕵🏻‍♂️

apps/desktop/src/lib/state/customHooks.svelte.ts
estib-vega: That's swallowed by the catch

apps/desktop/src/components/CommitCard.svelte
```diff
@@ -146,17 +146,17 @@
...
-		}
+		await updateCommitMessage({
+			projectId: project.id,
+			stackId: stack.id,
+			commitId: commit.id,
+			message: description
+		});
 
 		commitMessageModal?.close();
 		isUpdating = false;
```
estib-vega, line 162: This would need to be wrapped in a try/finally

apps/desktop/src/components/v3/CommitView.svelte
```diff
@@ -85,23 +85,13 @@
...
-			showToast({
-				message: `Update commit error`,
-				style: 'error'
-			});
-			return;
-		}
-
-		const newCommitId = result.data;
-
 		uiState.stack(stackId).selection.set({ branchName, commitId: newCommitId });
```
estib-vega, line 95: This also probably needs a try/finally

apps/desktop/src/components/v3/stackTabs/StackTabNew.svelte
```diff
@@ -45,21 +44,11 @@
...
-					message: 'Failed to create new stack',
-					style: 'error'
-				});
-				return;
-			}
-
-			const id = result.data.id;
-			goto(stackPath(projectId, id));
+			goto(stackPath(projectId, stack.id));
 			createRefModal?.close();
```
estib-vega, line 52: This too

PR #8122, opened 2025-04-11
mutations are typed correctly

estib-vega:
The types for the mutation hooks where defined loosely and outside their value declarations. This lead to the hooks being mistyped by mistake.

- Move the hooks types to where the hook functions are.
- Strengthen the types so that we don't mistype by accident.

### DIFF
--- a/apps/desktop/src/lib/state/customHooks.svelte.ts
+++ b/apps/desktop/src/lib/state/customHooks.svelte.ts
@@ -1,4 +1,5 @@
-import { reactive } from '@gitbutler/shared/storeUtils';
+import { isTauriCommandError, type TauriCommandError } from '$lib/state/backendQuery';
+import { reactive, type Reactive } from '@gitbutler/shared/storeUtils';
 import {
 	type Api,
 	type ApiEndpointMutation,
@@ -7,15 +8,16 @@ import {
 	type EndpointDefinitions,
 	type MutationActionCreatorResult,
 	type MutationDefinition,
+	type MutationResultSelectorResult,
 	type QueryActionCreatorResult,
 	type QueryArgFrom,
 	type ResultTypeFrom,
 	type RootState,
 	type StartQueryActionCreatorOptions
 } from '@reduxjs/toolkit/query';
-import type { TauriCommandError } from '$lib/state/backendQuery';
 import type { CustomQuery } from '$lib/state/butlerModule';
 import type { HookContext } from '$lib/state/context';
+import type { Prettify } from '@gitbutler/shared/utils/typeUtils';
 
 type TranformerFn = (data: any, args: any) => any;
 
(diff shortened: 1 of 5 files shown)


apps/desktop/src/lib/state/customHooks.svelte.ts
```diff
@@ -217,24 +270,25 @@ export function buildMutationHooks<Definitions extends EndpointDefinitions>({
...
+		let promise = $state<MutationActionCreatorResult<D>>();
 
-		async function triggerMutation(queryArg: unknown) {
+		async function triggerMutation(queryArg: QueryArgFrom<D>) {
 			preEffect?.(queryArg);
-			const dispatchResult = dispatch(initiate(queryArg, { fixedCacheKey }));
-			promise = dispatchResult;
-			const result = await promise.unwrap().catch((error) => {
-				onError?.(error, queryArg);
-			});
```
estib-vega, line 235: don't swallow the error

apps/desktop/src/lib/state/customHooks.svelte.ts
```diff
@@ -217,24 +270,25 @@ export function buildMutationHooks<Definitions extends EndpointDefinitions>({
...
+			promise = dispatch(initiate(queryArg, { fixedCacheKey }));
+			try {
+				const result = await promise.unwrap();
+				sideEffect?.(result, queryArg);
+				return result;
+			} catch (error: unknown) {
+				if (onError && isTauriCommandError(error)) {
+					onError(error, queryArg);
+				}
+				throw error;
```
estib-vega, line 290: On error is called, but the error is thrown still

apps/desktop/src/lib/state/customHooks.svelte.ts
```diff
@@ -164,48 +166,99 @@ export function buildQueryHooks<Definitions extends EndpointDefinitions>({
...
+			sideEffect?.(result, queryArg);
+			return result;
+		} catch (error: unknown) {
+			if (onError && isTauriCommandError(error)) {
+				onError(error, queryArg);
+			}
+			throw error;
 		}
-
-		return result;
```
estib-vega, line 208: don't return the dispatch result

apps/desktop/src/lib/state/customHooks.svelte.ts
```diff
@@ -164,48 +166,99 @@ export function buildQueryHooks<Definitions extends EndpointDefinitions>({
...
+		const dispatchResult = dispatch(initiate(queryArg, { fixedCacheKey }));
+		try {
+			const result = await dispatchResult.unwrap();
+			sideEffect?.(result, queryArg);
+			return result;
+		} catch (error: unknown) {
+			if (onError && isTauriCommandError(error)) {
+				onError(error, queryArg);
+			}
+			throw error;
```
estib-vega, line 260: return the mutation result or throw

PR #8217, opened 2025-04-21
Distribute the CLI

estib-vega:
Even though the CLI alone is not fully designed for in-prod purposes, let’s start distributing it alonside the other binaries, to enable the usage of the MCP server.

We can keep building the CLI iteratively


estib-vega:
@krlvi @Byron little heads-up

Byron:
Thanks! I am looking forward to seeing this in action soon.

This also makes me want to have the `but .` functionality, along with an easy way to make the binary available in `PATH` so one can quickly open a GitButler window on any repository. One fine day :)…

estib-vega:
> Thanks! I am looking forward to seeing this in action soon.
> 
> This also makes me want to have the `but .` functionality, along with an easy way to make the binary available in `PATH` so one can quickly open a GitButler window on any repository. One fine day :)…

Uhhh that would be very cool

PR #8279, opened 2025-04-27
Fix issue with context menu behaviour when extending beyond the boundaries of the window

### LINKED ISSUE #7912: Context menu can go outside of window
This happens on both the web and the client. We should inspect the bounding boxes and adjust the positions.

imadx:
## 🧢 Changes

- Updates the x,y positions for the context menu on setAlignByMouse to handle top/left alignments for the context menu
- Checks the top position of the context menu when toggling the top or bottom alignment to prevent consistent toggling.

## ☕️ Reasoning

- The alignment is ignored when right-clicking to open context menus.

## 🎫 Affected issues

Fixes: #7912


Byron:

I tried it the way I thought it should reproduce, but the issue remained unchanged on MacOS at least. Clicking on the `...` button of a commit will show a context menu that grows up, and the context menu of the whole branch will always grow down, and with that easily become invisible towards the bottom of the window frame.

imadx:
@Byron Do you have a screenshot that I can refer to regarding the recreated issue? Here’s what I’m seeing on my local build with the fix on my Mac.

image

Byron:
I think I get it now - the repositioning is working within the frame of the GitButler window.

What I tried is to put it against the Screen, assuming it would avoid going off-screen as well.
Screenshot 2025-04-29 at 18 47 39

Something that is unrelated, but just so it's documented and with CC to @estib-vega, I saw that the algorithm currently flip-flops if the vertical height isn't sufficient.

https://github.com/user-attachments/assets/1c5ff317-6199-4286-b0f0-56807f217f25

imadx:
I've updated the same PR to avoid constant toggling on the context menu as well, in the same PR, since they both relate to the alignment behaviour on context menu.

estib-vega (approved):
Very nice!

PR #8286, opened 2025-04-28
On workspace update list any conflicting files

krlvi:
This affects v3 only and it is about showing any uncommitted changes that will conflict with the workspace update. 
It currently looks like this: 
image

FYI @PavelLaptev


PavelLaptev:
@krlvi thanks! I'll make it prettier 💅

PR #8289, opened 2025-04-28
Allow merged branches to be deleted locally

krlvi:
This is relevant for both v2 and v3

It currently looks like this:
image

@PavelLaptev I will need a bit of help with the design.

Currently this is not added to the v3 version of this modal. @estib-vega I noticed that v3 of `upstreamIntegrationService.svelte.ts`'s `integrateUpstream` method does not seem to take any resolutions from the UI.

This is a bug, I added it to the v3 tasks list. And when you get to fixing that, could you add this checkbox as well?


PavelLaptev:
@krlvi gotcha

krlvi:
> @krlvi gotcha

Just as an FYI - the toggle shows up only when the entire stack is integrated (simpler this way)

PavelLaptev:
@estib-vega let me update the design first.

estib-vega:
> @estib-vega let me update the design first.

Sounds good 👍

PR #8307, opened 2025-04-28
feat(CommitHeader): add tooltip to display full commit message

matthewevans:
## 🧢 Changes

Added a Tooltip instance to the CommitHeader component. This allows the commit title to be displayed when the user hovers over the commit in the list.

## ☕️ Reasoning

- Provides full commit title when title is truncated

Screenshot 2025-04-28 at 12 03 44 PM


Byron (approved):
Thanks so much for contributing!

I love it, but I might be biased ;).

What do you think, @krlvi?

krlvi (approved):
Thanks for the contribution, Matt!

PR #8309, opened 2025-04-28
Debug: propagete missing askpass error instead of panic

krlvi:
The `debug_assert!` panics, which means the requests the frontend makes will never get a response. Which messes up reactivity in the app in a confusing way.

This change makes the error get propagated to the frontend

Will show this error message instead of panic
image


Byron:
You are a hero!

PR #8315, opened 2025-04-29
Tame ref-updates

Byron:
When using the commit-engine, references pointed to any old commit are updated to the new commit.
However, this also is done for references 'outside' of the workspace, and ideally we reign that in to only happen when a reference is known to be 'inside' of the workspace.

### Tasks

* [x] 'real-world' emulating test
    - Update checked-out branch
    - branches in virtual-branches.toml
* [x] fix it


Byron:
Checked with @krlvi, and we determined that the ref-behaviour for rewrites should be that refs that pointed to a rewritten commit should universally point to the new commit afterwards, to maintain the appearance of the references pointing to the commit-graph.

It's notable that there is no concept for tags right now - lightweight tags can easily be rewritten, and it's unclear if they were pushed or not, so it's more like a user preference whether they should be rewritten or not. One could also make it so that tags trigger a warning because such history probably shouldn't be rewritten.

PR #8318, opened 2025-04-29
When getting the diffs of a branch in v3, compare against the default target sha

krlvi:


krlvi:
Okay this is actually not needed

PR #8319, opened 2025-04-29
feat(StackTabNew): Simplify branch creation button text

krlvi:


krlvi:
@PavelLaptev FYI - I simplified the text here - no need to introduce the "stack" concept until there is more than one branch in a "stack"

PR #8320, opened 2025-04-29
Remove reference creation for namespaced branches

krlvi:
This is from before the time where we were using real branches in git. Now that we do that, these serve no purpose.


Caleb-T-Owens (changes requested):
You can put the entire for loop in the if statement

krlvi:
> You can put the entire for loop in the if statement

Done

Caleb-T-Owens (approved):

Issue #8331, opened 2025-04-29
Better handling of merge-conflicts in workspace commit creation upon commit


Byron:
When a commit fails due to conflicts when re-merging the workspace commit, right now we see this (as originally [reported on Discord](https://discord.com/channels/1060193121130000425/1204524097241878629/1366739355007189125).

<img width="1999" alt="Image" src="https://github.com/user-attachments/assets/c645cfff-c463-4fe2-87d6-fc5ee8f1de79" />

```
command: create_commit_from_worktree_changes
params: {"projectId":"50fb464e-00e2-4744-9870-c92df0cfb84d","stackId":"5a0502ec-b843-492a-98a9-8aef91134503","message":"HM scorm token overrides","stackBranchName":"uk/hm-scorm-token-overrides","worktreeChanges":[{"pathBytes":[103,114,97,100,108,101,45,112,114,111,106,101,99,116,45,114,111,111,116,47,99,111,109,109,111,110,47,115,114,99,47,109,97,105,110,47,107,111,116,108,105,110,47,99,111,109,47,115,97,110,97,108,97,98,115,47,99,111,109,109,111,110,47,100,111,109,97,105,110,47,79,114,103,97,110,105,122,97,116,105,111,110,73,100,46,107,116],"hunkHeaders":[]},{"pathBytes":[103,114,97,100,108,101,45,112,114,111,106,101,99,116,45,114,111,111,116,47,99,111,114,101,47,115,114,99,47,109,97,105,110,47,107,111,116,108,105,110,47,99,111,109,47,115,97,110,97,108,97,98,115,47,99,111,114,101,47,115,99,111,114,109,47,83,99,111,114,109,80,114,111,118,105,100,101,114,67,111,110,116,114,111,108,108,101,114,46,107,116],"hunkHeaders":[]}]})

Encountered conflict when merging commits 0534e2c184f5a1fa907ea1120aba73fb9a98dc9d, d3ef09f3d61a3b182b711a7c08fe46e4f66c3026, ab4713b112fc9946b86edf8d50e8fa185435d750
```

Byron:
CC @krlvi .

Byron:
Next steps as discussed, written down for posterity: show modal with paths grouped by reason to inform the user about what went wrong, and advise how to fix it.

Probably it's a good idea to differentiate the fatal case, everything failed, from the case where one can retry without the failed paths.

krlvi:
[...]

PR #8333, opened 2025-04-29
backend exposes workspace commit merge-failures to frontend

Byron:
The idea is that top-level merge-conflicts are observable, and we match the diffspecs against them. This allows the frontend to respond to it.

Related to #8331.

### Tasks

* [x] reproduce
* [x] backend exposes a new rejection reason with diffspecs for merge failure
* [x] frontend type is aware of rejection reasons

### DIFF
--- a/apps/desktop/src/lib/stacks/stackService.svelte.ts
+++ b/apps/desktop/src/lib/stacks/stackService.svelte.ts
@@ -94,6 +94,22 @@ export interface BranchPushResult {
 	remote: string;
 }
 
+type RejectionReason =
+	| 'NoEffectiveChanges'
+	| 'CherryPickMergeConflict'
+	| 'WorkspaceMergeConflict'
+	| 'WorktreeFileMissingForObjectConversion'
+	| 'FileToLargeOrBinary'
+	| 'PathNotFoundInBaseTree'
+	| 'UnsupportedDirectoryEntry'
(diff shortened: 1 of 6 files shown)


apps/desktop/src/lib/stacks/stackService.svelte.ts
```diff
@@ -94,6 +94,22 @@ export interface BranchPushResult {
...
+	| 'WorkspaceMergeConflict'
+	| 'WorktreeFileMissingForObjectConversion'
+	| 'FileToLargeOrBinary'
+	| 'PathNotFoundInBaseTree'
+	| 'UnsupportedDirectoryEntry'
+	| 'UnsupportedTreeEntry'
+	| 'MissingDiffSpecAssociation';
+
+export type CreateCommitOutcome = {
+	newCommit: string;
```
estib-vega, line 109: Is the newCommit optional still? If so this probably should be `newCommit: string | null`

estib-vega:
The rest looks good, so I'll just merge it and make the adaptions on my side

estib-vega:
Thanks again for the changes :)

apps/desktop/src/lib/stacks/stackService.svelte.ts
Byron: That's a great catch! Indeed it is optional, but I forgot to adjust the type.

Issue #5661, opened 2024-11-25
💚 Feature Request - Elevate Signed commits to the global/profile level and allow override at a project level


GordonBeeming:
Currently, the **Sign commits** flag is at a project level

![GB-Screenshot 2024-11-25 at 17 46 46@2x](https://github.com/user-attachments/assets/9952eff8-e5dd-498c-9809-bcf10bdfd9c5)

and it would be great to set this once at the profile level and allow override at a project level when required

![GB-Screenshot 2024-11-25 at 17 45 33@2x](https://github.com/user-attachments/assets/65492916-50bb-483a-88a4-570c59772b01)

Byron:
Thanks a lot for sharing!

Just a moment ago I was thinking the same - typically those who want to sign their commits want to sign by default, but *might* want to turn off signing for some repositories.

But in fact, what I really think is that I'd expect it to just respect the Git configuration - using that it should be able to know what to do.

Somehow I have a feeling that this feature is a remnant of how application-made SSH keys were previously used, even though these are long gone.

CC @PavelLaptev and @krlvi .

schacon:
That's strange, I thought we did look at / set the config settings through git.

Byron:
When `gitbutler.signCommits` is unset, the UI shows it as disabled, and this is what the backend implements as well.

https://github.com/gitbutlerapp/gitbutler/blob/329322b056596624aa46284c38ee19f83c2335fb/crates/but-rebase/src/commit.rs#L44

The UI should be able to detect if that configuration is set, and if not indicate that Git settings will be followed - the backend could then implement it accordingly. This would make `gitbutler.signCommits` some sort of override.

If #8238 makes commit-signing work for most out of the box, the setting can certainly also be made less prominent.

PR #8340, opened 2025-04-30
Re-use terminal environment

### LINKED ISSUE #4750: Failed to sign commits using gpg
### Version

0.12.20

### Operating System

Mac OS X

### Distribution Method

dmg (Apple Silicon)

### Describe the issue

Test signing with gpg failed with:

> Signing is not working correctly
> No such file or directory (os error 2)

The command in the log runs fine in the shell:

```
❯ "gpg" "--status-fd=2" "-bsau" "04278459" "-"                                                              at 14:58:33
[GNUPG:] KEY_CONSIDERED 91E148533A8CA28D4FB5054313D6365B04278459 2
[GNUPG:] BEGIN_SIGNING H10
hi there
[GNUPG:] PINENTRY_LAUNCHED 50639 mac 1.1.1 /dev/ttys000 xterm-256color - 20620/501/4 501/20 0
[GNUPG:] SIG_CREATED D 22 10 00 1724482720 91E148533A8CA28D4FB5054313D6365B04278459
-----BEGIN PGP SIGNATURE-----

iHUEABYKAB0WIQSR4UhTOoyijU+1BUMT1jZbBCeEWQUCZsmEoAAKCRAT1jZbBCeE
WSA5AP9VG1LWpZpgxa8XfD6KUC1jf5ssKH3bvR6zX4o74+u+IwEAjWLzqgP6Fe44
CmJyQk0HUol2i+vAtP8pLc3+/N0w7QM=
=aZUC
-----END PGP SIGNATURE-----
```

I'm using fish and homebrew-installed gpg if that helps. Thank you.

### How to reproduce

_No response_

### Expected behavior

_No response_

### Relevant log output

```shell
2024-08-24T06:42:08.568557Z ERROR check_signing_settings: crates/gitbutler-tauri/src/repo.rs:37: error=Error(failed to spawn "gpg" "--status-fd=2" "-bsau" "04278459" "-"

Caused by:
    No such file or directory (os error 2)) [3mid[0m[2m=[0m66ca6d82-42cd-4de4-9e12-6151f68415b0
```

### LINKED ISSUE #8238: Commit-signing takes too long due to shell-startup time
### Version

0.14.18

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

When performing amends or anything that performs a rebase of many commits, the time it takes to sign the rewritten commits starts to dominate the operation.

This turned out to be due to invoking GPG with a shell, just like Git does. On MacOS, we prefer the login shell as long as it's posix, but that may also lead to large startup times as it may pull in a lot of configuration.

Fix this by using a shell (`sh`) just like Git for compatibility, but make it possible to inject the cached environment of the user's login shell so GPG can be found out all.

_No response_

_No response_

```shell

```

Byron:
Try to acquire the login shell (`$SHELL`) environment and apply it to the parent process so we can use the Git shell (usually `sh`) to run helper programs like GPG.
That should significantly reduce the time it takes to invoke, as using `$SHELL` can add 100ms and more, something that accumulates when doing rebases or rewrites of a stack.
Additionally, this will make GPG work on more Unix systems (like mine).

Fixes #8238.
Fixes #4750.

### Tasks

* [x] global environment acquisition
* [x] test with posix shell
* [x] test with non-POSIX shell: `fish`
    - it works implicitly, as we know `fish` supports `-c` which seems quite universal

### Follow-Up issues

* #8345 
* #5661 

### Notes for the reviewer

* On Unix systems, a shell isn't used anymore, which is similar to how Git does it. On Windows, a Git shell is used to provide GPG from the programs that Git ships with.
* On Unix, GPG will now basically work out of the box, so it should be possible to change the behaviour of `gitbutler.signCommits`.

### DIFF
--- a/crates/but-core/src/cmd.rs
+++ b/crates/but-core/src/cmd.rs
@@ -1,26 +1,98 @@
+use bstr::BStr;
 use std::ffi::OsString;
+use std::path::PathBuf;
+use std::process::Stdio;
 
-/// Prepare `program` for invocation with a Git-compatible shell to help it pick up more of the usual environment.
+/// Prepare `program` for invocation with a Git-compatible shell to help it pick up more of the usual environment on Windows.
 ///
 /// On Windows, this specifically uses the Git-bundled shell, further increasing compatibility.
-pub fn prepare_with_shell(program: impl Into<OsString>) -> gix::command::Prepare {
-    let login_shell = std::env::var_os("SHELL");
-    // if login shell is set and it is a POSIX shell, use it, otherwise use the default shell
-    let shell_program = login_shell
-        .filter(|s| {
-            let s = s.to_string_lossy();
-            s.ends_with("/sh") || s.ends_with("/bash") || s.ends_with("/zsh")
-        })
-        .unwrap_or_else(|| gix::path::env::shell().into());
+pub fn prepare_with_shell_on_windows(program: impl Into<OsString>) -> gix::command::Prepare {
+    if cfg!(windows) {
+        gix::command::prepare(program)
+            // On Windows, this means a shell will always be used.
+            .command_may_be_shell_script_disallow_manual_argument_splitting()
+            // force using a shell, we want access to additional programs here
+            .with_shell()
+            // We know `program` is a path, so quote it.
+            .with_quoted_command()
+    } else {
+        gix::command::prepare(program)
+    }
+}
 
-    gix::command::prepare(program)
-        // On Windows, this means a shell will always be used.
-        .command_may_be_shell_script_disallow_manual_argument_splitting()
-        // On Windows, this yields the Git-bundled `sh.exe`, which is what we want.
-        // Only use a login shell here if it is a POSIX one.
-        .with_shell_program(shell_program)
-        // force using a shell, we want access to additional programs here
-        .with_shell()
-        // We know `program` is a path, so quote it.
-        .with_quoted_command()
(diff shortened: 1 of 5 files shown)


krlvi:
@Byron I tested this branch, and unfortunately, I still have this signing/fork exec overhead....

https://github.com/user-attachments/assets/b709fb3d-45fd-44a5-94e3-1e274d356f71

Byron:
For you this is indeed much slower. However, I think it's worth retrying, as it will only run `gpg` (in theory), which takes as long as it takes.
Worst-case this is the thing that's so slow for you. Generally, for me there is also a 30x difference between signing and not signing, sadly, despite only GPG being involved with a standard key and GPG agent for credentials.

Byron:
Just for posterity, here is a screen recording on the performance impact of signing on MacOS. GPG is executed directly here, without intermediate Git-shell as is the case on Windows.
The video has about 20 commits that are rewritten.

https://github.com/user-attachments/assets/4c498444-d350-40d7-b625-3f07d2c4e4ea

Performance numbers are 882ms with signing and 30,6ms without, a 30x difference that is solely attributed to GPG invocations. This is about in-line with my experience with rebasing signed commits in `stg` (StackedGit), which often feels slower than that.

krlvi (approved):
Thanks! Let's land this and take it from there

PR #8358, opened 2025-04-30
Allow creation of empty commits

krlvi:


krlvi:
@Byron what do you think of this? Scott has expressed the wish to create empty commits that he can later amend into. What do you think of this approach, is it reasonable?

Byron:
At first glance it looks like a minimal change to make it work, which is great.
I will add a couple of tests and push them here before merging it.

krlvi:
Thanks @Byron. Should I leave this PR "as is" for you to finalize in this case?

Byron:
Yes, I will take it from here now (already on it), but will leave the GUI code as is.

crates/but-cli/src/command/commit.rs
```diff
@@ -87,11 +87,11 @@ pub fn commit(
...
     } else {
         let destination = if amend {
-            if message.is_some() {
-                bail!("Messages aren't used when amending");
-            }
             let parent_id = parent_id.unwrap_or(repo.head_id()?.detach());
-            but_workspace::commit_engine::Destination::AmendCommit(parent_id)
+            but_workspace::commit_engine::Destination::AmendCommit {
+                commit_id: parent_id,
+                new_message: message.map(ToOwned::to_owned),
```
estib-vega, line 93: Nice! I wanted to add this at some point for the MCP as well. Cool!

Byron:
@krlvi @estib-vega `insert_blank_commit` could now be implemented in V3 directly, as Amending supports empty changes and a new message.

estib-vega (approved):
Looks good!
Should the frontend use the amend endpoint when updating the commit message now?

PR #8359, opened 2025-04-30
fix(UI): Remove duplicate "Delete local branch" option in Integrate Upstream modal

matthewevans:
## 🧢 Changes

- Removed redundant rendering logic for the "Delete local branch" option within the `IntegrateUpstreamModal.svelte` component.

## ☕️ Reasoning

When the Integrate Upstream modal displayed a single branch that was already integrated, the UI would incorrectly show the "Delete local branch" text and checkbox twice.

This happened because:
1. The `IntegrationSeriesRow.svelte` component correctly renders this UI element internally when `series.length === 1` and the branch status is `'integrated'`.
2. The `IntegrateUpstreamModal.svelte` component was *also* rendering the same UI element within the child content (`{#snippet stackStatus...}`) it passed to `IntegrationSeriesRow` under the same conditions (`stackFullyIntegrated(stackStatus)` being true and `series.length === 1`).

This commit removes the redundant rendering from the child content in `IntegrateUpstreamModal.svelte`, ensuring the option is only displayed once by the `IntegrationSeriesRow` component itself.

## 🖼️ 
Before
Screenshot 2025-04-30 at 11 13 59 AM

After
Screenshot 2025-04-30 at 11 23 16 AM

### DIFF
--- a/apps/desktop/src/components/IntegrateUpstreamModal.svelte
+++ b/apps/desktop/src/components/IntegrateUpstreamModal.svelte
@@ -22,7 +22,6 @@
 	import { getContext } from '@gitbutler/shared/context';
 	import Badge from '@gitbutler/ui/Badge.svelte';
 	import Button from '@gitbutler/ui/Button.svelte';
-	import Checkbox from '@gitbutler/ui/Checkbox.svelte';
 	import IntegrationSeriesRow, {
 		type BranchShouldBeDeletedMap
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot for contributing this fix!

Everything looks good to me, but @estib-vega should definitely take another look to be sure.

estib-vega (approved):
Thanks for this!

PR #8360, opened 2025-04-30
feat(ui): apply text styles from settings to MessageEditor

matthewevans:
## 🧢 Changes

- Passed user diff settings (font family, tab size, ligatures) from `MessageEditor.svelte` down to the CSS layer using CSS custom properties (`--code-block-font`, `--code-block-tab-size`, `--code-block-ligatures`).
- Updated `standard-theme.css` to consume these new custom properties for code blocks (`.StandardTheme__code`), inline code (`.StandardTheme__textCode`), and the plain text container (`.lexical-container.plain-text`). Sensible defaults are provided in case the properties are not set.

## ☕️ Reasoning

To apply a consistent visual representation of text styling across components.

### DIFF
--- a/apps/desktop/src/components/v3/editor/MessageEditor.svelte
+++ b/apps/desktop/src/components/v3/editor/MessageEditor.svelte
@@ -3,8 +3,9 @@
 	import MessageEditorRuler from '$components/v3/editor/MessageEditorRuler.svelte';
 	import CommitSuggestions from '$components/v3/editor/commitSuggestions.svelte';
 	import { showError } from '$lib/notifications/toasts';
+	import { SETTINGS, type Settings } from '$lib/settings/userSettings';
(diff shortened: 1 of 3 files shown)


PavelLaptev:
Thanks @matthewevans!

I was checking the PR and noticed that when I change the font in the input, only the font in the hunk header reflects the change — there’s no effect on anything else:

Hunk Header

Also, there are no changes in the editor itself:

Editor Screenshot

Does it work as expected on your end?

matthewevans:
@PavelLaptev Oh nice catch. The settings dialog preview has apparently not worked... ever? I've added the relevant prop so it's applied now
Screenshot 2025-04-30 at 3 27 34 PM
Screenshot 2025-04-30 at 3 27 20 PM

PavelLaptev:
Thanks! Works now :-)

PR #8374, opened 2025-05-02
Some CSS experiments with multi lane

krlvi:
I touched some CSS for this experimental mode thingie

Screenshot 2025-05-02 at 17 36 42

### DIFF
--- a/apps/desktop/src/components/v3/StackCard.svelte
+++ b/apps/desktop/src/components/v3/StackCard.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-	import BranchLabel from '$components/BranchLabel.svelte';
 	import { UiState } from '$lib/state/uiState.svelte';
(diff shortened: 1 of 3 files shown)


apps/desktop/src/components/v3/StackCard.svelte
```diff
@@ -20,12 +19,6 @@
 </script>
 
 <div class="stack-card text-15 text-bold text-" class:selected>
-	<div class="header">
-		<BranchLabel name={stackName} readonly />
-		<div>
-			{@render contextMenu()}
-		</div>
-	</div>
```
mtsgrd, line 28: Let's not drop the context menu? It's the only way to unapply a stack.

krlvi: Oh yeah, forgot about that. Maybe we could invent a context menu that looks nice in a "headless" mode (without stack name)

PR #8383, opened 2025-05-03
feat(mcp): add CreateCommitOutcomeSchema and rejection summary handling for commit outcome

estib-vega:
### Description

- Added `CreateCommitOutcomeSchema` to the MCP module.
- Implemented rejection summary handling for commit outcomes.

---
- <kbd>&nbsp;2&nbsp;</kbd> #8383 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #8382

### DIFF
--- a/packages/mcp/src/tools/client/commit.ts
+++ b/packages/mcp/src/tools/client/commit.ts
@@ -1,6 +1,7 @@
 import { BaseParamsSchema, DiffSpec } from './shared.js';
 import { listStackBranches, listStacks } from './status.js';
 import { executeGitButlerCommand, hasGitButlerExecutable } from '../../shared/command.js';
+import { CreateCommitOutcomeSchema } from '../../shared/entities/stacks.js';
 import { CallToolResult, GetPromptResult } from '@modelcontextprotocol/sdk/types.js';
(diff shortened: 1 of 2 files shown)


packages/mcp/src/tools/client/commit.ts
```diff
@@ -38,6 +39,55 @@ function populateDiffSpec(
...
+	const groupedRejections = new Map<string, Set<string>>();
+	for (const path of outcome.pathsToRejectedChanges) {
+		const [reason, filePath] = path;
+		const paths = groupedRejections.get(reason) || new Set<string>();
+		paths.add(filePath);
+		groupedRejections.set(reason, paths);
+	}
+
+	const rejectionMessages = Array.from(groupedRejections.entries())
+		.map(([reason, paths]) => `${reason}: ${Array.from(paths).join(', ')}`)
```
Byron, line 61: Since it's a `Set`, is the order in `paths` defined? `Array.from` probably won't do the sorting, and if that is true one might want to do it here.
The same might be true for the reasons, i.e. `groupedRejections.entries()`, which in practice would be less of an issue as it's unlikely there will be more than one, but possible.

PR #8400, opened 2025-05-05
Implement v3 move commit to another stack

krlvi:


apps/desktop/src/components/v3/BranchHeader.svelte
```diff
@@ -79,6 +82,14 @@
 	let renameBranchModal = $state<BranchRenameModal>();
 	let deleteBranchModal = $state<DeleteBranchModal>();
 
+	const dzHandler = $derived(
+		new MoveCommitDzHandler(
+			stackService,
+			args.type === 'stack-branch' ? args.stackId : '', // Only used for stack-branch
```
krlvi: It sucks that I have to do this here... This component feels kinda bloated, having to drag unrelated functionality for preview branches (referred to as 'normal-branch' in here)

Issue #7671, opened 2025-03-17
`GLIBC_2.38' not found


FrancescoLucia:
### Version

2.38

### Operating System

Linux

### Distribution Method

AppImage (Linux)

### Describe the issue

Hi, I got this error when launching the AppImage (distro PopOS)
```
(AppImageLauncher:2438309): GLib-CRITICAL **: 17:57:06.312: g_file_test: assertion 'filename != NULL' failed
gitbutler-tauri: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by gitbutler-tauri)
gitbutler-tauri: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by gitbutler-tauri)
gitbutler-tauri: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by gitbutler-tauri)
gitbutler-tauri: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by /tmp/.mount_GitButtT8C8k/usr/lib/libwebkit2gtk-4.1.so.0)
gitbutler-tauri: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by /tmp/.mount_GitButtT8C8k/usr/lib/libwebkit2gtk-4.1.so.0)
gitbutler-tauri: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.31' not found (required by /tmp/.mount_GitButtT8C8k/usr/lib/libwebkit2gtk-4.1.so.0)
gitbutler-tauri: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.32' not found (required by /tmp/.mount_GitButtT8C8k/usr/lib/libwebkit2gtk-4.1.so.0)
gitbutler-tauri: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by /tmp/.mount_GitButtT8C8k/usr/lib/libgtk-3.so.0)
gitbutler-tauri: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by /tmp/.mount_GitButtT8C8k/usr/lib/libgtk-3.so.0)
[...]

PR #8404, opened 2025-05-05
linux build improvements

Byron:
This is a research PR which could even be merged if the experiment is successful.

Try to see how low we can get when building a Tauri App.

### Prerequisites
* [x] learn about minimal support version of Ubuntu/Debian for Tauri 2
    - Seems to be 22.04

### Tasks

* [x] try CI build with Ubuntu 22
* [x] try to release on Ubuntu 22 locally (by running relevant CI steps and try build)
    - works
* [x] make local build available for testing: https://github.com/gitbutlerapp/gitbutler/releases/tag/release%2F0.12.9
* [x] offer *untested* (or only locally tested) CI changes that might need to be reverted.

### ~~Optional Tasks~~

No need to try to bypass available runner images as the lowest bound is determined by Gtk4.1.

* [ ] See how Docker works on GitHub to build and extract artefacts
* [ ] Try to build on Ubuntu 20 (locally)
* [ ] possibly setup a Dockerfile to build a release binary and package it as
    * [ ] .deb
    * [ ] AppImage

Related to https://github.com/gitbutlerapp/gitbutler/issues/7671#issuecomment-2850618406.

### Notes for the Reviewer

> [!WARNING]
> This may have to be reverted as it could fail the publishing. With that in mind, I think it has good chances to work given my local testing.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -92,7 +92,7 @@ jobs:
         platform:
           - macos-13 # [macOs, x64]
           - macos-latest # [macOs, ARM64]
-          - ubuntu-24.04 # [linux, x64]
+          - ubuntu-22.04 # [linux, x64]
           - windows-latest # [windows, x64]
 
     runs-on: ${{ matrix.platform }}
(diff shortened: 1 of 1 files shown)


Byron:
Let's not do this until the new Ubuntu 22 build was verified. It's quite likely that it will fail at runtime, judging from past experience.

PR #8422, opened 2025-05-07
basic head-info

Byron:
A first implementation of `head_info()` to provide information about where the user currently is.
It presents the world as a workspace with stacks, even though technically it might not exist.`

Follow-up on #7596.

* [x] prevent non-workspace cases (or assure that the new system doesn't get to show them)
    - [x] figure out how it knows that one moved away from GitButler workspace
        - `operating_mode()` is called when the workspace branch changes, which tells the UI it's not on the right branch. So no need to alter the current implementation.
* [ ] `heads_info` with workspace cases - reflect what's currently possible
* [x] research on how to Integrate new `heads_info()` with current API interface
    - [x] move UI types into `ui` modules
    - [x] check if current data is sufficient to provide all that's needed to current API
        -  ~~`get_base_branch_data()`~~ (skipped), `stacks()`, `stack_details()`
* [ ] implement intermediate V3 versions of these functions 
    - [x] `stacks_v3()`
        - [ ] basic test (needs heads-info test setup)
    - [ ] `branch_details_v3()`
    - [ ] `stack_details_v3()`
* [ ] make use of `stacks_v3` and `stack_details_v3()` in tauri

## Notes for the Reviewer

* This is a transitory PR that replaces VirtualBranchesToml with the RefMetadata trait to prepare the code to transition to other data stores.
* The code is clearly transitory while `StackId`  is still a thing - in the new world stacks or stack-segments are referred to by their reference name or by their index in the parent-list of the top-level merge commit (if there is one).
* Ideally, one the UI will be able to make just one `heads_info` call, and can consume the data directly (even though it may have been processed to further facilitate consumption).
* `branch_details()` already works on references
* `BranchDetails` are probably the data structure of choice for the UI, and it's just the question if there is stack information or not.
* There is no notion of using stacks in branch listings outside of what we are currently looking at, i.e. where `HEAD` is. Thus, for this we will probably keep using branch details of sort.

## Next PR

- intermediate V3 version of `get_base_branch_data()`.

## Conversion check

Does the new types have enough information to provide what's needed to the UI?

### Stack
- `stack_id` (deprecated, maybe emulated with workaround to map an index to an id maybe)
- `heads`
    - `name` ✓
    - `tip` ✓

### Stack Details
- derived_name ✓ (stack name, top-most branch in stack)
- push_status (enum) (derived)
- is_conflicted (has a conflicting commit somewhere) (derived)
- branch details
    - name ✓
    - remote_tracking_branch ✓ (from `Stack::upstream`, but it's deduced seemingly)
    - description ✓
    - pr_number ✓
    - review_id ✓
    - tip ✓
    - base_commit (derived)
    - push_status: (derived with remote tracking branch)
    - last_updated_at (from RefInfo::updated_at ✓)
    - authors: Vec<Author> ✓
    - is_conflicted ✓
    - commits: Vec<Commit> ✓
    - upstream_commits: Vec<UpstreamCommit> ✓
    - is_remote_head ✓

### Commit 
- id ✓
- parent_ids ✓
- message ✓
- has_conflicts ✓
- state: CommitState (derived)
- created_at ✓
- author: Author ✓

### CommitState

All derived once the commit-analysis is done

- LocalOnly (derived)
- LocalAndRemote(id) (derived)
- Integrated (derived)

### UpstreamCommit

- id ✓
- message ✓
- created_at ✓
- author ✓

### Author

- name ✓
- email ✓
- gravatar_url ✓

### Follow Up Tasks

* [ ] first non-workspace cases
    - [ ] merge commit
    - [ ] merge commit stacked
    - [ ] stacks and ambiguous stack references (i.e. lots of empty stack segments)

### Next PRs

* ⚠️One probably wants to show all refs at a position (or indicate there are more) ⚠️

### For follow-up PRs

In any order (probably)

### Out of scope

* **hunk-dependencies**
* **atomic object writes**

### DIFF
--- a/crates/but-workspace/src/branch.rs
+++ b/crates/but-workspace/src/branch.rs
@@ -515,6 +515,19 @@ pub struct Stack {
     pub stash_status: Option<StashStatus>,
 }
 
+impl Stack {
+    /// Return the name of the top-most [`StackSegment`].
+    ///
+    /// It is `None` if this branch is the top-most stack segment and the `ref_name` wasn't pointing to
+    /// a commit anymore that was reached by our rev-walk.
+    /// This can happen if the ref is deleted, or if it was advanced by other means.
+    pub fn name(&self) -> Option<&gix::refs::FullNameRef> {
+        self.segments
+            .first()
+            .and_then(|name| name.ref_name.as_ref().map(|name| name.as_ref()))
+    }
+}
+
 /// A list of all commits
 #[derive(Debug, Clone)]
 pub struct BranchCommit {
--- a/crates/but-workspace/src/branch_details.rs
+++ b/crates/but-workspace/src/branch_details.rs
@@ -1,0 +1,220 @@
+use crate::head_info::function::workspace_data_of_default_workspace_branch;
+use crate::ui::{CommitState, PushStatus};
+use crate::{state_handle, ui};
+use anyhow::{Context, bail};
+use but_core::RefMetadata;
+use gitbutler_command_context::CommandContext;
+use gitbutler_error::error::Code;
+use gitbutler_oxidize::OidExt;
+use gix::remote::Direction;
+use std::collections::HashSet;
+use std::path::Path;
+
+/// Returns information about the current state of a branch.
+pub fn branch_details(
+    gb_dir: &Path,
+    branch_name: &str,
+    remote: Option<&str>,
+    ctx: &CommandContext,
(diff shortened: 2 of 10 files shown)


Byron:
Intermediate merge to capture refactoring - none of the original functions were altered/are functionally different, only new functions were added (that aren't even functional yet).

PR #8463, opened 2025-05-08
Remove commit status label for now

mtsgrd:
Reading "pushed commit" for a commit on e.g. master is not helpful.
Removing these descriptions until we can provide something more
accurate.

image


PavelLaptev:
@mtsgrd To me, this is helpful for a few reasons:

- The status label doesn’t take up much space.  
- What exactly is the helpful part here? Is it the commit SHA? I’d argue it’s more than that.  
- We added this label to make the commit status **clear and obvious**, without requiring users to hover over icons and guess based on color.  
- If we remove the status label, we reduce the **discoverability** of the commit state. Then we’d rely solely on icon hover tooltips, and we don’t explain those colors anywhere else in the UI.

@krlvi thoughts?

krlvi:
Maybe we can come up with better copy for this. "Pushed" perhaps doesn't read very nicely, language wise. But I don't have any suggestion for another phrase at the moment...

PR #8497, opened 2025-05-12
Unified diff: Locks are shown independent of the stack

estib-vega:
Since we no longer have stacks tied to pages, we no longer can (or shold) hide the locks dependeing on which stack you’re viewing.

- Display the locks independently of the stack
- Re-enable the tests

---
- <kbd>&nbsp;2&nbsp;</kbd> #8499 
- <kbd>&nbsp;1&nbsp;</kbd> #8497 👈


krlvi:
How about using the current "commit goes here" selection @estib-vega?

PR #8517, opened 2025-05-12
feat(settings): add autoFetchIntervalMinutes to mock settings

estib-vega:
Include autoFetchIntervalMinutes with a default value of 15 in the
mock app settings to simulate fetch behavior during tests. This change
improves test coverage for features relying on fetch intervals.


ig-raiderler:
The performance optimizations here are quite clever. Nice engineering work!

PR #8554, opened 2025-05-13
Fix deduplication error in creating new stack

mtsgrd:
If say branch `test` is taken we want to deduplicate to `test-1`
rather than `test 1`, since the latter is not a valid ref.


krlvi (commented):
Thanks!!!!

PR #8588, opened 2025-05-15
Better FE Error on nonexclusive project access

Byron:
This way it's possible for the UI to correctly explain what's going on.

### Tasks

* [x] backend makes lock-error distinguishable
* [x] frontend uses this new Code variant


estib-vega:
@Byron added the FE part of this

PR #8608, opened 2025-05-16
feat(baseBranch): improve `shortName` calculation

### LINKED ISSUE #8596: GitHub API Error while creating pull request
### Version

0.14.23

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

When creating a pull request, from the gitbutler app, I get:

Error while creating pull request
```
Validation Failed: {"resource":"PullRequest","field":"head","code":"invalid"} - https://docs.github.com/rest/pulls/pulls#create-a-pull-request
```

Image

1. Create Pull Request
2. Error while creating pull request

_No response_

```shell

```

matthewevans:
## 🧢 Changes
- Renamed original `shortName` getter to `lastPathComponent` to better reflect its actual functionality
- Added an enhanced implementation of `shortName` that properly handles Git reference formats
- Added support for stripping common Git prefixes (`refs/remotes/remoteName/`, `remoteName/`, `refs/heads/`)

## ☕️ Reasoning
The previous `shortName` implementation only extracted the last path component of a branch name without handling Git's reference formats properly. The new implementation normalizes branch names by intelligently stripping Git-specific prefixes based on context (remote vs. local), resulting in more consistent and user-friendly branch name representation throughout the application.

Further context:

- This affected me when I had my target set to `origin/release/2025.18`. When I tried to create a PR it was attempting to use `2025.18` as the base branch. GitHub refused and GitButler showed an error. This fix allowed me to successfully create the PR. I also invested the other locations that leveraged `shortName` and they all appeared to use it in a context where they would expect the full branch name.

## 🎫 Affected issues

Fixes: #8596 (I believe – I was receiving the same error in my repro case)


Byron:
That's great, thanks so much! It would be lovely if this could fix the linked PR-creation issue.
Do you think it's possible to add a simple task list to the PR so it's possible to see what's missing? Maybe even turn it into a draft.
The reason I am asking is that the new `shortName()` function doesn't seem to be used yet.
Thanks again

CC @estib-vega .

matthewevans:
> That's great, thanks so much! It would be lovely if this could fix the linked PR-creation issue. Do you think it's possible to add a simple task list to the PR so it's possible to see what's missing? Maybe even turn it into a draft. The reason I am asking is that the new `shortName()` function doesn't seem to be used yet. Thanks again
> 
> CC @estib-vega .

Forgive me if I'm misunderstanding, but the _new_ `shortName` function (that's been renamed) is just a copy of the old `shortName`.

1. Renamed `shortName` to `lastPathComponent`
2. Implemented a fixed version of `shortName` that works with branches that contain `/`. This is the actual fix of this PR.
3. `lastPathComponent` is left purely for posterity (if there's a use-case that desires just the last part of the branch path).

estib-vega:
@matthewevans another great contribution!
Much appreciated :)

Caleb-T-Owens:
We should probably use the rust function `given_name` for this since remove names can include forward slashes

apps/desktop/src/lib/baseBranch/baseBranch.test.ts
```diff
@@ -0,0 +1,178 @@
...
+
+describe('BaseBranch.shortName', () => {
+	const testCases = [
+		// Cases with remoteName
+		{
+			branchName: 'refs/remotes/origin/feature/foo',
+			remoteName: 'origin',
+			expected: 'feature/foo',
+			description: 'full ref with remote'
+		},
```
Caleb-T-Owens, line 12: you could have some horror of a scenario where you have a remote called `origin/feature` and the branch name is only `foo`

PR #8613, opened 2025-05-18
git style diffs

Byron:
This PR should improve the diff quality as it introduces a `Sink` that calculates Git diffs with hunks similar to what `xdiff` does.
At least in theory, as I didn't vet the algorithm, just the implementation.

This makes using the new `Sink` a bit of a leap of faith, as it uses enough slice-indices to make panics quite possible. There is also not a lot of tests, nor is there a fuzzer setup, so anything could happen.
Ideally, it will just work though, and if it gives trouble it should be easy to revert the commit.

It will probably be used in `bat` soon: https://github.com/sharkdp/bat/pull/3235#issuecomment-2889210009.

And lastly, [here is two diffs](https://github.com/GitoxideLabs/gitoxide/blob/55056467e8b1d2ee327e4f29df058224fb64db5e/gix-diff/tests/diff/blob/git_diff.rs#L88-L139
) that show where the new algorithm differs - the slider is at the bottom of the diff, and the upper diff is more desirable thanks to the new algorithm.

### Tasks

* [x] update `gix` for `GitDiff` `Sink` 
* [x] Bonus: Additional updates to auto-updateable dependencies
* [x] use new `Sink` when creating unified diffs

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,21 +34,21 @@ version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
 dependencies = [
- "getrandom 0.2.15",
+ "getrandom 0.2.16",
  "once_cell",
  "version_check",
 ]
 
 [[package]]
 name = "ahash"
-version = "0.8.11"
+version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


Byron:
The E2E tests seem to be genuinely failing, even though I have trouble seeing how this is related.
The frontend shows diffs like normal.

Byron:
Checked with the team and we decided it's OK to merge it, noting that:

* there is a risk this panics and if that happens we can undo this change easily in `unified_diff.rs`
* it has the potential to fix a current problem where V2 and V3 diffs don't match up

CC @krlvi .

PR #8685, opened 2025-05-21
fix: Unset the last project used

estib-vega:


apps/desktop/src/components/v3/profileSettings/GeneralSettings.svelte
```diff
@@ -97,6 +99,8 @@
 		isDeleting = true;
 		try {
 			await settingsService.deleteAllData();
+			await projectsService.reload();
+			projectsService.unsetLastOpenedProject();
```
Byron, line 103: I wonder what happens if someone delete the last opened project in another Window or on disk so it's genuinely not present anymore next time the UI comes up.
And I tried it, `rm  ~/Library/Application\ Support/com.gitbutler.app.dev/projects.json` and bringing up the app correctly puts it into the uninitialised state 🎉.

PR #8709, opened 2025-05-22
various improvements

Byron:
Various changes between #8590 and its follow-up.

### Tasks

* [x] update to latest version of `git2-hooks` 
* [x] create empty commits if none of the changes could be applied. That way nothing is lost.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2455,13 +2455,13 @@ dependencies = [
 [[package]]
 name = "git2-hooks"
 version = "0.4.0"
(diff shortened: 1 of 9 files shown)


Byron:
This is only a partial fix as it will allow the creation of no-op commits in case the cherry-pick fails, a common reason for this.
There it automatically retries.

However, with the final workspace merge we can't currently retry, even though it's technically possible.
So I guess for now we get good at communicating the failure, and then we either allow the merge to retry (if there is something to retry), or we figure out a way to use hunk-dependencies to guide the user.

PR #8717, opened 2025-05-22
feat: Pretty display of commit rejection reasons

estib-vega:
Display a modal if there are any rejection reasons for commiting

### DIFF
--- a/apps/desktop/cypress/e2e/errorHandling.cy.ts
+++ b/apps/desktop/cypress/e2e/errorHandling.cy.ts
@@ -1,16 +1,16 @@
 import { clearCommandMocks, mockCommand } from './support';
-import MockBackend from './support/mock/backend';
 import { PROJECT_ID } from './support/mock/projects';
+import BranchesWithChanges from './support/scenarios/branchesWithChanges';
 
 describe('Error handling - commit actions', () => {
-	let mockBackend: MockBackend;
+	let mockBackend: BranchesWithChanges;
 
 	const UPDATE_COMMIT_ERROR_MESSAGE = 'Error updating commit message';
 	const COMMIT_ERROR_MESSAGE = 'Error creating commit';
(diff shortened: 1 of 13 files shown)


apps/desktop/cypress/e2e/errorHandling.cy.ts
```diff
@@ -138,6 +139,138 @@ describe('Error handling - commit actions', () => {
 		expect(mockBackend.getDiff).to.have.callCount(0);
 	});
 
+	it('Fully failing to commit with rejection reasons shuold be handled graceful', () => {
```
estib-vega, line 144: @PavelLaptev I've added these two tests for the new commit failure modal.

There are two types:
- Full failure --> No commit was created
- Partial failure --> Commit was created, but some file changes were not committed as intended

estib-vega: If you change this line to:
```ts
it.only('Fully ....')
```

Only this test will run, so it will be easier to change the design of the modal

apps/desktop/cypress/e2e/errorHandling.cy.ts
```diff
@@ -138,6 +139,138 @@ describe('Error handling - commit actions', () => {
...
+		cy.getByTestId('global-modal-action-button').should('be.visible').click();
+		cy.getByTestId('global-modal-commit-failed').should('not.exist');
+
+		// The commit drawer should be open still
+		cy.getByTestId('new-commit-drawer').should('be.visible');
+		cy.getByTestId('commit-drawer-title-input').should('have.value', newCommitMessage);
+		cy.getByTestId('commit-drawer-description-input').should('contain', newCommitMessageBody);
+	});
+
+	it('Fully failing to commit with rejection reasons shuold be handled graceful', () => {
```
estib-vega: Same here, if you only want to try this test to see the other version of the modal, just add 

```ts
it.only('Fully failing to commit...')
```

PR #8756, opened 2025-05-24
Handle condition where notify event has no associated paths

### LINKED ISSUE #8754: Debounces events without path
### Version

0.14.26

### Operating System

Linux

### Distribution Method

source (Linux)

### Describe the issue

Related to #8094 

It seems there can be events without paths on linux. I didn't dig deep yet, because I wanted to get feedback first. Let me know if you need more information.

```shell
thread 'notify-rs inotify loop' panicked at /home/ganwell/testing/lasting/aports-but/testing/gitbutler/src/gitbutler-release-0.14.26/crates/gitbutler-filemonitor/vendor/debouncer/src/lib.rs:242:32:
index out of bounds: the len is 0 but the index is 0
```

I fixed the problem with this patch.

```patch
diff --git a/crates/gitbutler-filemonitor/vendor/debouncer/src/lib.rs b/crates/gitbutler-filemonitor/vendor/debouncer/src/lib.rs
index 1ec5edbf..15ec147b 100644
--- a/crates/gitbutler-filemonitor/vendor/debouncer/src/lib.rs
+++ b/crates/gitbutler-filemonitor/vendor/debouncer/src/lib.rs
@@ -239,6 +239,9 @@ impl<T: FileIdCache> DebounceDataInner<T> {
             return;
         }

+        if event.paths.is_empty() {
+            return;
+        }
         let path = &event.paths[0];

         match &event.kind {
```

This solves the issue

ghost:
Such an event cannot be cached, so we skip it

## 🧢 Changes

Skip debounce cache on events without path

## ☕️ Reasoning

I assume not all fs-notify backends produce the exact same events and on linux it seems possible to have an event without paths (I haven't checked what kind of event this is)

## 🎫 Affected issues

Fixes: #8754

### DIFF
--- a/crates/gitbutler-filemonitor/vendor/debouncer/src/lib.rs
+++ b/crates/gitbutler-filemonitor/vendor/debouncer/src/lib.rs
@@ -239,6 +239,9 @@ impl<T: FileIdCache> DebounceDataInner<T> {
             return;
         }
 
+        if event.paths.is_empty() {
+            return;
+        }
         let path = &event.paths[0];
 
         match &event.kind {


krlvi (approved):
Thank you!

PR #8772, opened 2025-05-26
Posthog for the latency of rust api calls

krlvi:
Let's figure out what is slow and make it fast

### DIFF
--- a/apps/desktop/src/lib/state/backendQuery.ts
+++ b/apps/desktop/src/lib/state/backendQuery.ts
@@ -47,9 +47,15 @@ export async function tauriBaseQuery(
 	};
 
 	try {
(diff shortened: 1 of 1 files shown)


apps/desktop/src/lib/state/backendQuery.ts
```diff
@@ -47,9 +47,12 @@ export async function tauriBaseQuery(
 	};
 
 	try {
+		const startTime = performance.now();
```
mtsgrd: `performance` is a node interface, we need to use `Date.now()` instead.

krlvi: 🙈 - thanks for the catch, the AI completion hallucinated this

PR #8781, opened 2025-05-26
Create a lightweight SQLite setup and migrate hunk_assingments to use it for persistence

krlvi:
Outstanding issues:
- [x] When a repository is opened for the first time there is a `Failed to run migrations: database is locked` error - likely due to a race condition
    - Byron says: It should be much less likely now, but it's not defeated. It's strange as well.
- [x] Fix CI

### DIFF
--- a/crates/but-db/diesel.toml
+++ b/crates/but-db/diesel.toml
@@ -1,0 +1,9 @@
+# For documentation on how to configure this file,
+# see https://diesel.rs/guides/configuring-diesel-cli
+
+[print_schema]
+file = "src/schema.rs"
+custom_type_derives = ["diesel::query_builder::QueryId", "Clone"]
+
+[migrations_directory]
+dir = "/Users/kiril/src/gitbutler/crates/but-db/migrations"
--- a/crates/but-db/diesel.toml
+++ b/crates/but-db/diesel.toml
@@ -6,4 +6,4 @@ file = "src/schema.rs"
 custom_type_derives = ["diesel::query_builder::QueryId", "Clone"]
 
 [migrations_directory]
-dir = "/Users/kiril/src/gitbutler/crates/but-db/migrations"
+dir = "migrations/"
--- a/crates/gitbutler-command-context/src/lib.rs
+++ b/crates/gitbutler-command-context/src/lib.rs
@@ -10,16 +10,19 @@ pub struct CommandContext {
     project: Project,
     /// A snapshot of the app settings obtained at the beginnig of each command.
     app_settings: AppSettings,
+    db_handle: but_db::DbHandle,
 }
 
 impl CommandContext {
     /// Open the repository identified by `project` and perform some checks.
     pub fn open(project: &Project, app_settings: AppSettings) -> Result<Self> {
         let repo = git2::Repository::open(&project.path)?;
+        let db_handle = but_db::DbHandle::new(&project.gb_dir())?;
         Ok(Self {
             git_repo: repo,
             project: project.clone(),
             app_settings,
+            db_handle,
         })
     }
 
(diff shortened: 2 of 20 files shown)


crates/but-db/diesel.toml
```diff
@@ -0,0 +1,9 @@
+# For documentation on how to configure this file,
+# see https://diesel.rs/guides/configuring-diesel-cli
+
+[print_schema]
+file = "src/schema.rs"
+custom_type_derives = ["diesel::query_builder::QueryId", "Clone"]
+
+[migrations_directory]
+dir = "/Users/kiril/src/gitbutler/crates/but-db/migrations"
```
krlvi: We should probably edit this to be a relative path

Byron: That will be it! Let me try that.
The other error, related to a relative path being passed, also reproduced this generic error so I thought it would work. But maybe with this change it will - these errors reproduce locally.

Byron: This is only relevant for running the CLI I think, the path for embedding migrations seems alright.

krlvi: Yeah this is only read when running diesel_cli

Byron: It turns out that the normalization that I added didn't is very lazy and only really does something for `../` paths, very much by design.
Then a `./.git/…` path was passed, which is known to not work.
Now there are tests for that, and concurrent intra-process access is also being validated and generally works, for reading. When writes are intermixed, one has [some tuneables](https://github.com/the-lean-crate/criner/blob/c36b108c763270a3e5b3adc9c562e0a12646a00d/criner/src/persistence/mod.rs#L25-L28) that can be added to the connection initialization.

Byron:
I think you can take over now, @krlvi . It will be interesting to see if the added concurrency settings help with that other "Database is locked" issue.

There is still some bits that can still be done if this is still encountered, basically making the DB wait until locks are released instead of failing.

> There are two measures to make this work. Firstly, there should be a busy handler which, in the simplest case, sleeps a little. It should be installed when opening a connection to the database.

https://github.com/crates-io/criner/blob/c36b108c763270a3e5b3adc9c562e0a12646a00d/criner/src/persistence/mod.rs#L59-L60

Secondly each transaction that plans to read before writing should begin in IMMEDIATE mode. Without it, it will first get a read lock for the read portion, and then try to upgrade to a write lock when needed. In my case, this didn't trigger the busy handler and would fail permanently.

> https://github.com/crates-io/criner/blob/c36b108c763270a3e5b3adc9c562e0a12646a00d/criner/src/persistence/table.rs#L135-L160
>
> connection.transaction_with_behavior(rusqlite::TransactionBehavior::Immediate)?
The above is the key to making this work. In conjunction with a busy handler that sleeps…
>
> https://github.com/crates-io/criner/blob/c36b108c763270a3e5b3adc9c562e0a12646a00d/criner/src/persistence/mod.rs#L96-L100
>
> …this call will block until a write lock was obtained before any read happened.

Byron (approved):
@krlvi The concurrency issue seems fixed - I wasn't able to reproduce it anymore.
This works thanks to the setup of a busy-handler which waits up to 30s before failing permanently with "database locked".

I would not expect this to be a problem as we only read and write in short bursts, with data (soon) being distributed over multiple tables.

crates/gitbutler-command-context/src/lib.rs
```diff
@@ -32,6 +35,10 @@ impl CommandContext {
         &self.git_repo
     }
 
+    pub fn db(&mut self) -> &mut but_db::DbHandle {
```
Byron: I'd definitely prefer to do this lazily - with `&mut self` that's easy, and won't spawn a database each time `git_head()` is called (this is where it failed for me with "database locked".

Byron: For context, I think this happens when running the migration, something that strangely enough isn't reproducible in tests at all.
Ideally we don't risk blocking a call if it's not going to use the database.

krlvi: Yeah it does make sense to do this in a lazy way

Byron:
I also tried this (change order of Pragma): https://github.com/diesel-rs/diesel/issues/2365#issuecomment-2899347817 .
Let's see.

PR #8801, opened 2025-05-28
Setup a but-auto crate for automations

krlvi:
This should be possible to invoke both from the UI as well as from the MCP endpoint.
It will be good to separate concerns as well as dependencies


krlvi:
@estib-vega this setup now allows us to have this "handle_changes" be used from anywhere:
- as a tauri command
- as a regular cli command
- as an MCP endpoint

PR #8805, opened 2025-05-28
feat(ui): add hunk dependencies computation for specific changes

estib-vega:
Introduce a new function `hunk_dependencies_for_changes` that computes hunk
dependencies from an explicit set of tree changes, decoupling it from the
worktree directory. Refactor the existing `hunk_dependencies_for_workspace_changes_by_worktree_dir`
to call this new function after obtaining changes from the worktree.

Add a Tauri command `hunk_dependencies_for_specific_changes` to expose
hunk dependency computation for arbitrary changes via the UI, enabling
more flexible and targeted dependency queries.

These changes improve modularity and allow the UI to request dependencies
for any given set of changes, not just those from the current worktree.

### DIFF
--- a/crates/but-hunk-dependency/src/ui.rs
+++ b/crates/but-hunk-dependency/src/ui.rs
@@ -6,6 +6,24 @@ use gitbutler_stack::StackId;
 use serde::{Deserialize, Serialize};
 use std::path::Path;
 
+/// Compute the hunk dependencies of a set of tree changes.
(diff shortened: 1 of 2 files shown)


estib-vega:
> @estib-vega I noticed this while creating release notes. I think it may be worth sync-ing with Mattias - now the tree_changes API provides the full set of locks, so I suspect this may not be needed anymore

Not by the UI (right now), but for mcp things it will

PR #8819, opened 2025-05-29
Use `gitoxide` to do author-check on project-add.

Byron:
Previously `git2` was used, which left the possibility for it to not pick up authors if there were complex/unsupported Git configuration.

With `gitoxide`, all of this is supported so the chances of picking something up are higher to the point where it should 'just work'.

Triggered by https://discord.com/channels/1060193121130000425/1206670506271707156/1377404899750318200.


Byron:
It doesn't seem to obviously break anything, merging.

PR #8823, opened 2025-05-29
avoid panics when calculating hunk dependencies.

Byron:
These are now part of workspace changes (which might be a problem by itself), but until that can be changed, let's be sure it fails more gracefully by making these failures optional.

Thus, no panics, and failures to get hunks are non-fatal, while being logged.

This fixes a panic for me that prevented worktree changes to show up.

### DIFF
--- a/crates/but-hunk-assignment/src/lib.rs
+++ b/crates/but-hunk-assignment/src/lib.rs
@@ -195,8 +195,8 @@ impl HunkAssignment {
     }
 }
 
-#[instrument(skip(ctx), err(Debug))]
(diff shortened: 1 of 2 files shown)


Byron:
@krlvi Even though this fixes a panic, I don't know if there are any constraints that this violates, i.e. one assignment per change.

Further, it seems the performance is a problem at least for my test repository (`deno`) where there is a single applied branch with 1077 commits in it. The repository has 10141 tracked files, but no worktree changes.

krlvi (approved):
Thanks!!

PR #8825, opened 2025-05-29
hunk-assignments performance optimisations

Byron:
### Tasks

* [x] don't calculate if there are no worktree changes
* [x] calculate worktree changes only once.


Byron:
@krlvi This PR seems to improve performance by avoiding duplicate worktree_changes computations, and by setting up gix::Repository instances with a large object cache to speed up tree-diffs.

It does, however, not address the issue that unified diffs can fail to compute if the involved change is a submodule.
I couldn't easily fix it as each change should have a unified diff, and I didn't want to open a rabbit hole. But in theory, if it could be `Option<UnifiedDiff>`, then the compiler would tell which places need adjustments. Since you know the logic best, maybe that's something you want to tackle.

krlvi (approved):
Thanks!!

PR #8828, opened 2025-05-29
Fix exiting edit mode

mtsgrd:
When we exit edit mode we need to checkout the tree of the new workspace
commit before we apply the uncommitted changes.

### DIFF
--- a/crates/gitbutler-edit-mode/src/lib.rs
+++ b/crates/gitbutler-edit-mode/src/lib.rs
@@ -325,6 +325,13 @@ pub(crate) fn save_and_return_to_workspace(
     let new_workspace = WorkspaceState::create(ctx, perm.read_permission())?;
     let uncommtied_changes = get_uncommited_changes(ctx)?;
 
+    let workspace_oid = update_workspace_commit(&vb_state, ctx)?;
+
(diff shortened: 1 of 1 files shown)


crates/gitbutler-edit-mode/src/lib.rs
```diff
@@ -325,6 +325,13 @@ pub(crate) fn save_and_return_to_workspace(
     let new_workspace = WorkspaceState::create(ctx, perm.read_permission())?;
     let uncommtied_changes = get_uncommited_changes(ctx)?;
 
+    let workspace_oid = update_workspace_commit(&vb_state, ctx)?;
+
+    repository.checkout_tree(
+        repository.find_commit(workspace_oid)?.tree()?.as_object(),
+        Some(CheckoutBuilder::new().force().remove_untracked(true)),
```
Byron, line 332: I think the edit mode went through a couple of iterations in regard of worktree checkouts and I wonder if @krlvi has all the context. Without a doubt it's not currently working as intended, but I wonder if such forceful checkouts have previously already been removed for other reasons.

Something I'd be interested in is if untracked changes in the workspace can now survive entering edit mode, and exiting it.

PR #8837, opened 2025-05-30
Implement an install_cli command

krlvi:
For now, making this internal only. 
Also... windows is not done yet

image

### DIFF
--- a/crates/gitbutler-tauri/src/cli.rs
+++ b/crates/gitbutler-tauri/src/cli.rs
@@ -1,0 +1,89 @@
+use std::path::Path;
+
+use crate::error::Error;
+use anyhow::Context;
+use tracing::instrument;
+
+#[tauri::command(async)]
+#[instrument(err(Debug))]
+pub fn install_cli() -> anyhow::Result<(), Error> {
+    #[cfg(windows)]
+    {
+        // TODO
+        return Err(anyhow::anyhow!(
+            "CLI installation is not supported on Windows. Please install manually."
+        ))
+        .map_err(Error::from);
+    }
+
+    let cli_path = get_cli_path()?;
+    let link_path = Path::new("/usr/local/bin/but");
+    let bin_dir = link_path.parent().context("Cant find bin dir")?;
+
+    if link_path.exists() {
+        let current_link = std::fs::read_link(link_path)
+            .context(format!(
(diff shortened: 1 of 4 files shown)


crates/gitbutler-tauri/src/cli.rs
```diff
@@ -0,0 +1,89 @@
...
+            "CLI installation is not supported on Windows. Please install manually."
+        ))
+        .map_err(Error::from);
+    }
+
+    let cli_path = get_cli_path()?;
+    let link_path = Path::new("/usr/local/bin/but");
+    let bin_dir = link_path.parent().context("Cant find bin dir")?;
+
+    if link_path.exists() {
```
Byron, line 23: I would get `symlink_metadata(link_path)`. From there one can more easily handle the case that there is an actual binary in which case we probably want to say so and abort the attempt.

Right now we would also abort, but with an error that says something about failing to read a link.

crates/gitbutler-tauri/src/cli.rs
```diff
@@ -0,0 +1,89 @@
...
+use tracing::instrument;
+
+#[tauri::command(async)]
+#[instrument(err(Debug))]
+pub fn install_cli() -> anyhow::Result<(), Error> {
+    #[cfg(windows)]
+    {
+        // TODO
+        return Err(anyhow::anyhow!(
+            "CLI installation is not supported on Windows. Please install manually."
```
Byron, line 14: Is the binary shipped on Windows? If so, then the CLI path could be mentioned here and users could help themselves. This would buy time to one day make it nicer, particularly if there is demand.

Byron (commented):
Even though ordinarily I'd be a bit anxious about the `-f` flag in `ln`, I think there would be no issue as long as we validate that we can't (easily) overwrite an existing non-symlink.

*And I am too late, the PR merged while I was typing this :D*

krlvi:
@PavelLaptev - I added this as an internal/admin only setting in the global settings. Would you mind taking a look at the thing, and perhaps making it look consistent with design? I think the command could benefit from a copy-paste thing.

crates/gitbutler-tauri/src/cli.rs
krlvi: yeah, should be there, with an exe suffix

PR #8843, opened 2025-05-30
install cli follow up

Byron:
- **CLI-install: Make error handling around the spot being occupied clearer**
- **Don't try to create `/usr/local/bin` as it's unlikely to be in PATH then**
- **refactor for readability**

### DIFF
--- a/crates/gitbutler-tauri/src/cli.rs
+++ b/crates/gitbutler-tauri/src/cli.rs
@@ -7,36 +7,51 @@ use tracing::instrument;
 #[tauri::command(async)]
 #[instrument(err(Debug))]
 pub fn install_cli() -> anyhow::Result<(), Error> {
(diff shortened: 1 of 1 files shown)


krlvi (approved):
thanks!

crates/gitbutler-tauri/src/cli.rs
```diff
@@ -1,64 +1,86 @@
...
-            &format!(
-                "do shell script \" \
-                    mkdir -p \'{}\' && \
+    if cfg!(target_os = "macos") {
+        let status = std::process::Command::new("/usr/bin/osascript")
+            .args([
+                "-e",
+                &format!(
+                    "do shell script \" \
                     ln -sf \'{}\' \'{}\' \
```
krlvi, line 61: do we leave the `-f`?

PR #8845, opened 2025-05-30
Revert "Fix exiting edit mode"

krlvi:
Reverts gitbutlerapp/gitbutler#8828

Unfortunately, while this change fixed a real problem, this also causes uncommitted changes to be wiped when entering/exiting edit mode.

Since the second regression is more severe than the (also problematic) issue the PR addresses, let's revert this for now, as we implement a proper fix for edit mode.

cc @mtsgrd @Caleb-T-Owens

### DIFF
--- a/crates/gitbutler-edit-mode/src/lib.rs
+++ b/crates/gitbutler-edit-mode/src/lib.rs
@@ -325,13 +325,6 @@ pub(crate) fn save_and_return_to_workspace(
     let new_workspace = WorkspaceState::create(ctx, perm.read_permission())?;
     let uncommtied_changes = get_uncommited_changes(ctx)?;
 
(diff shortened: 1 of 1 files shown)


Byron:
I just want to add that my testing (with this change) might not have shown the whole truth. After all, it's not enough to know if this change really introduced these issues. They could already have been present when entering edit mode, for instance, and it would need additional testing to know for sure.

Byron:
This video was recorded on `master` with this branch applied, and it looks like the wiping also happened before this change.

https://github.com/user-attachments/assets/4d1ddad1-4047-4d89-a507-0ac037e03c3b

Please note that this should be validated in another repository as well as my test-repo might be very weird or otherwise won't represent a healthy standard repo.

PR #8880, opened 2025-06-02
Implement a reload signal that can be consumed in the frontend and reload the workspace

krlvi:
In the frontend we can do something like
```ts

	import { listen } from '$lib/backend/ipc';
	import { ClientState } from '$lib/state/clientState.svelte';
	import { getContext } from '@gitbutler/shared/context';
	const clientState = getContext(ClientState);
	$effect(() => {
		const unsubscribe = listen(`project://${projectId}/reload`, (event) => {
			clientState.backendApi.util.resetApiState();
		});
		return () => unsubscribe();
	});
```

However, it seems like `clientState.backendApi.util.resetApiState();` does not seem to work, so skipping this for now

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -699,6 +699,7 @@ dependencies = [
  "gitbutler-branch",
  "gitbutler-branch-actions",
  "gitbutler-command-context",
+ "gitbutler-filemonitor",
  "gitbutler-operating-modes",
  "gitbutler-oplog",
  "gitbutler-oxidize",
--- a/crates/but-action/Cargo.toml
+++ b/crates/but-action/Cargo.toml
@@ -27,6 +27,7 @@ gitbutler-oplog.workspace = true
 gitbutler-oxidize.workspace = true
 gitbutler-project.workspace = true
 gitbutler-reference.workspace = true
+gitbutler-filemonitor.workspace = true
 but-core.workspace = true
 but-db.workspace = true
 but-workspace.workspace = true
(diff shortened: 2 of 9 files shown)


Byron:
I didn't look into if this is working, but assume it would given the filesystem monitor is relatively mature at this point. My fear here would be that removing the file triggers another event so there is some self-triggering going on.

If the purpose is to refresh the UI if something changed, then I'd think ideally we can update the current filesystem monitor implementation to respond to Git changes as well, a feature needed for Single-Branch mode.

This could be a tool to help debug fielsystem events, it also shows the events the filesystem watcher actually gets, along with the filtered-down events that the application gets to see.

```
RUST_LOG=trace cargo run -p but-cli -- --trace watch
```

If there is truly a new mechanism needed to let both communicate I think they could do with sockets or pipes, there [is a crate for that](https://lib.rs/crates/interprocess) (even though I never used it), while Cargo uses the [`jobserver`](https://lib.rs/crates/jobserver) crate to let multiple instances talk to each other (including `make`) to prevent over-allocation of threads.

mtsgrd:
Was this still relevant? I mentioned this over chat but it might have gotten lost, `resetApiState` is an action and needs to be dispatched using the dispatcher from `ClientState`. I need to fix the same reset for github/gitlab api's when switching projects!

krlvi:
> Was this still relevant? I mentioned this over chat but it might have gotten lost, `resetApiState` is an action and needs to be dispatched using the dispatcher from `ClientState`. I need to fix the same reset for github/gitlab api's when switching projects!

I think that we maaaay do a different implementation here using sqlite hooks

PR #8892, opened 2025-06-03
Fix integrate upstream modal again

Caleb-T-Owens:
When I tested this before without a forge, I tested when there were no branches applied. As it turns out, there was a second place where the code assumed implicitly assumed that a forge was present.

This refactor removes the undefined state of the filteredReviews. They are now always an array.

This also now “syncronously” sets the filteredReviews in the open function. This means that we don’t need to try and look at the filteredReviews themselves to see if the thing has finished loading.

This refactor also gets rid of the “v2” variant because they are the same fucking thing and we don’t fucking need the other one.

---
⧓ Review in [Butler Review `#1aWKJWty4`](https://gitbutler.com/cto/gitbutler-client-d443/reviews/1aWKJWty4)

**Fix integrate upstream modal again**

When I tested this before without a forge, I tested when there were no branches applied. As it turns out, there was a second place where the code assumed implicitly assumed that a forge was present.

This refactor removes the undefined state of the filteredReviews. They are now always an array.

This also now “syncronously” sets the filteredReviews in the open function. This means that we don’t need to try and look at the filteredReviews themselves to see if the thing has finished loading.

This refactor also gets rid of the “v2” variant because they are the same fucking thing and we don’t fucking need the other one.

1 commit series (version 1)

| Series | Commit Title | Status | Reviewers | 
| --- | --- | --- | --- |
| 1/1 | [Fix integrate upstream modal again](https://gitbutler.com/cto/gitbutler-client-d443/reviews/1aWKJWty4/commit/e842a63b-fad1-40cb-a116-736846210376) | ⏳ |  |

_Please leave review feedback in the [Butler Review](https://gitbutler.com/cto/gitbutler-client-d443/reviews/1aWKJWty4)_


apps/desktop/src/components/v3/IntegrateUpstreamModal.svelte
```diff
@@ -64,15 +64,16 @@
...
+	let filteredReviews = $state<PullRequest[]>([]);
 	const reviewMap = $derived(new Map(filteredReviews?.map((r) => [r.sourceBranch, r])));
 
 	const isDivergedResolved = $derived(base?.diverged && !baseResolutionApproach);
 	const [integrateUpstream] = $derived(upstreamIntegrationService.integrateUpstream(projectId));
 
 	$effect(() => {
 		if (!modal?.imports.open) return;
-		if (branchStatuses?.type !== 'updatesRequired' || filteredReviews === undefined) {
+		if (branchStatuses?.type !== 'updatesRequired') {
```
Caleb-T-Owens, line 76: This was a very flawed condition that was trying to express "If the filtered reviews are loading, don't bother with anything else". This broke down when there were no forge integrations.

apps/desktop/src/components/v3/IntegrateUpstreamModal.svelte
```diff
@@ -122,27 +123,19 @@
 				.then((result) => {
 					targetCommitOid = result;
 				});
+		} else {
+			// If there is no divergence we should set this to undefined.
+			targetCommitOid = undefined;
```
Caleb-T-Owens, line 128: This is not relevant to the main fix, but it is in general a good practice to always have the else condition to unset any variables that were set in the true condition. This is important because of svelte's atomic reactivity.

Essentially, if this first ran and set `targetCommitOid` because `base.diverted` was true, if `base.diverged` then changed to false, we would then want to set it back to `undefined`.

apps/desktop/src/components/v3/IntegrateUpstreamModal.svelte
```diff
@@ -170,23 +163,20 @@
...
 
 	async function fetchAppliedBranches() {
-		return await stackService
-			.fetchStacks(projectId)
-			.then(
-				(stacksResponse) =>
-					stacksResponse.data?.flatMap((stack) => stack.heads.map((head) => head.name)) ?? []
-			);
+		const stacksResponse = await stackService.fetchStacks(projectId);
+		return stacksResponse.data?.flatMap((stack) => stack.heads.map((head) => head.name)) ?? [];
```
Caleb-T-Owens, line 167: This change here is a no-op, it's the same thing just expressed using async/await syntax

apps/desktop/src/components/v3/IntegrateUpstreamModal.svelte
```diff
@@ -170,23 +163,20 @@
...
 	}
 
 	export async function show() {
 		integratingUpstream = 'inert';
 		branchStatuses = undefined;
-		filteredReviews = undefined;
+		filteredReviews = [];
```
Caleb-T-Owens, line 173: When we start to show the modal, clear out the old value of `filteredReviews` with an empty array. We are going to set it "synchronously" before `branchStatuses` (the thing the loading state depends on) gets set.

apps/desktop/src/components/v3/IntegrateUpstreamModal.svelte
```diff
@@ -410,9 +400,7 @@
 				wide
 				style="pop"
 				disabled={isDivergedResolved || !branchStatuses}
-				loading={integratingUpstream === 'loading' ||
-					!branchStatuses ||
-					(forgeListingService && filteredReviews === undefined)}
+				loading={integratingUpstream === 'loading' || !branchStatuses}
```
Caleb-T-Owens, line 403: Now `branchStatuses` will never get set before we have finished setting `filteredBranches`. (If there is no forge integration, the function to set `filteredBranches` just returns early)

apps/desktop/src/components/v3/IntegrateUpstreamModal.svelte
```diff
@@ -122,27 +123,19 @@
...
-			});
-		}
-	});
+	async function setFilteredBranches(appliedBranches: string[]) {
+		if (!forgeListingService) return;
+
+		const reviews = await forgeListingService.fetchByBranch(projectId, appliedBranches);
+		// Find the reviews that have a "mergedAt" timestamp
+		filteredReviews = reviews.filter((r) => !!r.mergedAt);
+	}
```
Caleb-T-Owens, line 138: This is a rewrite of the old effect as a function. We are safe to write this as a function like this because the dependencies of this function were only set during the `open` function call anyway.

One change we did make is in how we filter the reviews. Previously a PR with no `sourceBranch` was considered "merged", but that only works if the GH repository is configured to automatically delete branches. This change makes use of the `mergedAt` property, which is at timestamp of when the PR was merged, and will only ever be set if the PR has in fact been merged.

PR #8911, opened 2025-06-05
Fallback when user.signingkey isn't configured

### LINKED ISSUE #8907: Git gpg signing fails if user.signingkey is not set
### Version

0.14.19

### Operating System

Linux

### Distribution Method

nix (nix builds from [source](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/gi/gitbutler/package.nix))

### Describe the issue

git gpg signing doesn't require user.signingkey to be set, instead it uses gpg key associated with the committer identity when unset.

In ~/.config/git/config set:
```
[commit]
    gpgSign = true
[gitbutler]
    signCommits = true
```
Try commit in gitbutler, but the unhandled exception occurs "No signing key found" 

gitbutler is expected to use the gpg key associated with the committer identity by default instead of throwing an exception

```shell
2025-06-04T17:34:52.970312Z  WARN save_edit_and_return_to_workspace: crates/gitbutler-repo/src/repository_ext.rs:347: Commit signing failed but remains enabled as gitbutler.signCommits is explicitly enabled globally project_id=77037f81-e371-4a93-88ec-8862e1185c20
2025-06-04T17:34:52.970926Z ERROR save_edit_and_return_to_workspace: crates/gitbutler-tauri/src/modes.rs:73: error=Error(Failed to commit new commit

Caused by:
    No signing key found) project_id=77037f81-e371-4a93-88ec-8862e1185c20
```

Byron:
Fixes #8907.

Now one can still turn signing on, but no key has to be set. The UI will highlight this in red, but it works nonetheless.

Screenshot 2025-06-05 at 05 44 37

The format for the user is the same as Git uses when signing.

```
05:29:31.359915 run-command.c:655       trace: run_command: gpg --status-fd=2 -bsau 'Sebastian Thiel <sebastian.thiel@icloud.com>'
error: gpg failed to sign the data
fatal: failed to write commit object
```

### DIFF
--- a/crates/but-rebase/src/commit.rs
+++ b/crates/but-rebase/src/commit.rs
@@ -1,5 +1,5 @@
 use anyhow::{Context, anyhow, bail};
-use bstr::{BString, ByteSlice};
+use bstr::{BStr, BString, ByteSlice};
 use but_core::cmd::prepare_with_shell_on_windows;
 use but_core::{GitConfigSettings, RepositoryExt};
 use gitbutler_error::error::Code;
@@ -96,10 +96,7 @@ pub(crate) fn update_committer(
 pub fn sign_buffer(repo: &gix::Repository, buffer: &[u8]) -> anyhow::Result<BString> {
     // TODO: support gpg.ssh.defaultKeyCommand to get the signing key if this value doesn't exist
     let config = repo.config_snapshot();
-    let Some(signing_key) = config.string("user.signingkey") else {
-        bail!("No signing key found");
-    };
-    let signing_key = signing_key.to_str().context("non-utf8 signing key")?;
+    let signing_key = signing_key(repo)?;
     let sign_format = config.string("gpg.format");
     let is_ssh = if let Some(sign_format) = sign_format {
         sign_format.as_ref() == "ssh"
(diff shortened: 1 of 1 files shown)


crates/but-rebase/src/commit.rs
```diff
@@ -184,12 +181,12 @@ pub fn sign_buffer(repo: &gix::Repository, buffer: &[u8]) -> anyhow::Result<BStr
...
             );
 
-        let mut cmd = into_command(prepare_with_shell_on_windows(gpg_program.as_ref()).args([
-            "--status-fd=2",
-            "-bsau",
-            signing_key,
-            "-",
-        ]));
+        let mut cmd = into_command(
+            prepare_with_shell_on_windows(gpg_program.as_ref())
```
Byron, line 185: While at it

crates/but-rebase/src/commit.rs
```diff
@@ -228,12 +225,28 @@ fn into_command(prepare: gix::command::Prepare) -> std::process::Command {
     cmd
 }
 
-fn as_literal_key(maybe_key: &str) -> Option<&str> {
-    if let Some(key) = maybe_key.strip_prefix("key::") {
-        return Some(key);
+fn as_literal_key(maybe_key: &BStr) -> Option<&BStr> {
```
Byron, line 228: While at it, I removed the `String` conversion in favor of supporting any encoding. This is particularly relevant when using the author identity, as it truly just is a bundle of bytes and we wouldn't want these to be causing issues down the road.

PR #8919, opened 2025-06-05
graph segmentation spike

Byron:
Produce a datastructure that can represent any commit-graph as segments, which is the level of detail that we are interested in.
These segments can be traversed forward and backwards, and they know the commits that are uniquely assigned to it.
Segments can have inter-segment relations, to help with dealing with related remote tracking branches.

### Problems this should solve

This PR should allow us to see if this is the right way to go, as it should solve a couple of problems, all at once:

* [x] make complex commit-graphs graspable
    - Even though there are limits inherent to the tree-view, but it seems to do well enough for our idea of stacks.
* [x] make any commit-graph representable
  * [ ] map any such graph to stacks
* [ ] make complex workspace updates possible
    - [x] imagine updates to the rebase-engine to be able to transplant complex graphs as well
        - This would work by having well-known operations, supporting only one at a time for now, and then to return
           a pick-list for the rebase engine that can be reviewed before actually running it (i.e. dry-run built-in).

### Tasks

* [x] research datastructure and visualization
    - use `petgraph` and visualize as dot (with edge/node getters) and using `termtree` for tree-like visualization
* [x] sketch data structure as graph
* [x] visualise hand-built data structures for testing and validation
* [x] basic initialisation from graph-walk
    - ~~Find a way to do permutation testing for when there is multiple tips (and there is most of the time)~~ - probably too hard for now with dubious value, there are many additional tests to write.
    - [x] Split across merges
    - [ ] Split across potential stack-segments
* [ ] map graphs to stacks.
* [ ] Show how typical segment-walks can now be facilitated with standard algorithms

#### Petgraph as basis for graphs

It has efficient data structures for graphs and would certainly be a good basis.
This also makes it easy to output graphs in `dot` format, but besides using the `dot` verbatim, it's of no use for visualization. Maybe it can be used from time to time to visualize it by hand via `pbpaste | dot -Tsvg > out.svg`.

#### `dot` (Graphviz) for visualisation

As ugly as it might have been, it has a `dot -Tascii` plugin, but only from v13. And that is unavailable for installation, so I couldn't even see if it is suitable.
It's impossible to build it on a Mac just to get the latest version, and clearly it's not feasible to use `-Tascii` because of that.

### Related

* #8905


Byron:
Merging this first version to avoid conflicts and set the stage for more to come.
This code doesn't affect anything else meaningfully yet, but some refactoring were necessary.

PR #8922, opened 2025-06-05
Fixes a bug where sqlite connection fails on windows

krlvi:


crates/but-db/src/lib.rs
```diff
@@ -44,7 +44,7 @@ impl DbHandle {
         let db_file_path = db_file_path
             .to_str()
             .ok_or_else(|| anyhow::anyhow!("Invalid UTF-8 in database file path"))?;
-        Self::new_at_url(format!("file://{}", db_file_path))
+        Self::new_at_url(db_file_path)
```
Byron, line 47: I can't believe I fell for that… again :/.

PR #8935, opened 2025-06-07
graph segmentation

Byron:
Produce a datastructure that can represent any commit-graph as segments, which is the level of detail that we are interested in.
These segments can be traversed forward and backwards, and they know the commits that are uniquely assigned to it.
Segments can have inter-segment relations, to help with dealing with related remote tracking branches.

### Problems this should solve

This PR should allow us to see if this is the right way to go, as it should solve a couple of problems, all at once:

* [x] make complex commit-graphs graspable
    - Even though there are limits inherent to the tree-view, but it seems to do well enough for our idea of stacks.
* [x] make any commit-graph representable
  * [x] map any such graph to stacks
* [x] make complex workspace updates possible
    - [x] imagine updates to the rebase-engine to be able to transplant complex graphs as well
        - This would work by having well-known operations, supporting only one at a time for now, and then to return
           a pick-list for the rebase engine that can be reviewed before actually running it (i.e. dry-run built-in).

### Tasks

- [x] better debugging with on-the-fly dot-graphs.
- [x] Split across merges
- [x] integrate workspace branches automatically
    - [x] propagate flags
- [x] figure out if edge-weights could be removed in favor of a graph that can be traversed
    - Keep edge weight as debugging tool - it allows to validate constraints when visualising the graph (i.e. last commit must start the edge, first must receive it), just in case it gets mutated.
    - [x] ~~Use only validated connection methods, which means we make connections 'at the right time' when post-processing.~~ - do not fail, instead rely on visual debugging (without highlights for violations).
- [x] Split across potential stack-segments
    - [x] handle empty segments (as listed in the workspace)
    - [x] reconnect segments linked to commits that are now in different segments
    - [x] create segments without name even when there are ambiguous ref-names
        - ~~No, as this would create anonymous non-empty segments which later have to be dealt with when mapping the graph into stacks. Such segments would have to be logically connected/merged into the segment above them, which either means they can't have graph-based segment ids, or the UI would have to do it. Segment IDs seem nice for the UI to have as well, and it seems easier to be able to say that one segment stays one segment when mapping them (and each segment is either empty or has a name).~~
        - Actually, since we have to deal with segments that are not empty and have no name, because the type-system supports it, we need to handle this when mapping. So the solution is to make the segment-id optional.
- [x] integrate remotes
    - [x] also have tests without workspace
    - [x] test eager remote traversal and assigning of previously ambiguous segment name in post
- [x] integrate the target branch
    - [ ] make sure that integrated commits can be hidden (for instance, new (virtual) stacks shoudn't have a commit.
    - [ ] early abort should work as well, check the queue to stop if there is only 'hidden' tips left.
- [ ] split across stacks (those that are resting at the target ref (or maybe the merge base?))
    - We must be able to split out stacks of empty commits resting at the base, that is segments that will be mapped into stacks later. For that we connect them to the governing workspace.
- [ ] ⚠️ fix related TODOs and known issues
- [ ] map graphs to stacks.
    - [ ] make sure all existing tests work and/or are adapted
    - [ ] Put `has_conflicts` into `Commit` and remote `RemoteCommit` type
- [ ] Show how typical segment-walks can now be facilitated with standard algorithms

### Not to forget

### Related

* #8905

### Previous

* #8919


Byron:
Intermediate merge.

PR #8985, opened 2025-06-12
Add 3.5 UI feature toggle

mtsgrd:
Key combination "3 5" toggles the new workspace layout.


PavelLaptev (approved):
Thanks! But shouldn't we do it

PR #9023, opened 2025-06-16
Improve hunk retain function

Caleb-T-Owens:
This introduces a new updateLines function to handle the retaining of individually selected lines. It has the following behaviour:

- If every line is selected, they should remain all selected
- If some of the lines are selected, then filter out any old selected lines that no longer exist.
- If the selection ends up empty, return undefined which should signal that the entire selection should be dropped.

This could be made more intelligent in the future (IE updating the selection based on the similarity of old lines to new lines) but this seems to work well enough for now.

This also goes to the effort to introduce an everyLineSelected function which can be used in a follow up to update the checkLine logic to make it go from “Array that holds all the lines” to “Empty array"

---

Something worth touching on is that this makes use of Sets to try and avoid any O(n^2) logic. This is technically more complex, but I don’t think it impacts readability much.

### DIFF
--- a/apps/desktop/src/lib/selection/uncommitted.ts
+++ b/apps/desktop/src/lib/selection/uncommitted.ts
@@ -481,6 +481,7 @@ function updateAssignments(
 		treeChangeAdapter.getInitialState(),
 		action.payload.changes
 	);
+	const oldAssignments = state.hunkAssignments;
 	state.hunkAssignments = hunkAssignmentAdapter.addMany(
 		hunkAssignmentAdapter.getInitialState(),
 		action.payload.assignments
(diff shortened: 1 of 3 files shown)


apps/desktop/src/lib/selection/uncommitted.ts
```diff
@@ -496,17 +497,100 @@ function updateAssignments(
...
 	return state;
 }
+
+/**
+ * Updates the lines in the selection based on the new assignment.
+ *
+ * If the old assignment was full, we will keep the lines that are still
+ * selected.
+ * If the old assignment was partial, we will keep the lines that are present.
+ *   If there are no lines left, we will return undefined.
```
mtsgrd, line 530: Formatting

Caleb-T-Owens: That was an intentional indentation as it is only relevant after the filtering step

PR #9064, opened 2025-06-17
Result<Option>> for unified_diff()

Byron:
This way, no code can accidentally run into errors anymore for trying to diff undiffable things.

See https://discord.com/channels/1060193121130000425/1384601119338270872 for motivation.

### Tasks

* [x] ❌couldn't reproduce the issue when just obtaining worktree changes, but it's clearly there in the UI which doesn't treat submodules specifically yet.
* [x] fix

### DIFF
--- a/crates/but-core/src/diff/worktree.rs
+++ b/crates/but-core/src/diff/worktree.rs
@@ -819,7 +819,7 @@ impl TreeChange {
         &self,
         repo: &gix::Repository,
         context_lines: u32,
-    ) -> anyhow::Result<UnifiedDiff> {
+    ) -> anyhow::Result<Option<UnifiedDiff>> {
         let mut diff_filter = crate::unified_diff::filter_from_state(
             repo,
             self.status.state(),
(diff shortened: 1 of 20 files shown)


krlvi:
@Byron I added option handling for the scenario that caused the most pain https://github.com/gitbutlerapp/gitbutler/pull/9066

There are of course the other places but they probably don't get triggered or if they do, its less problematic. It is still necessary though to have this return Option

krlvi:
It appears that this didnt take care of the issue so i did some more places where we assume there will be a diff https://github.com/gitbutlerapp/gitbutler/pull/9068

Byron:
I have modified everything and basically moved the handling of 'no-diff' to the call-site. All changes where previously legitimate errors would be ignored have been replaced with "don't ignore the error, but ignore a missing diff".

That way I think that the meaning of the code that dealt with diffs didn't change, and those parts that could skip over missing diffs still will, now including skips of submodules.

If this doesn't fix the issues they describe we should now see exactly where it came from.

PR #9071, opened 2025-06-18
fix-amend-hunk-assignments

estib-vega:
- Added logic to retrieve hunk assignments for current selections, limited to worktree files.
- Extended uncommittedService to support assignment retrieval by single and multiple paths.
- Updated amend functionality to read and apply hunk assignments when amending files into commits.
- Modified drop handler and utils to process hunk assignments, ensuring only selected hunks are amended.
- Integrated hunk assignment logic into draggable components for improved drag and drop operations.

### DIFF
--- a/apps/desktop/src/lib/commits/utils.ts
+++ b/apps/desktop/src/lib/commits/utils.ts
@@ -1,6 +1,7 @@
+import { isDefined } from '@gitbutler/ui/utils/typeguards';
 import type { DetailedCommit } from '$lib/commits/commit';
 import type { TreeChange } from '$lib/hunks/change';
-import type { DiffSpec } from '$lib/hunks/hunk';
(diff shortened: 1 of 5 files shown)


apps/desktop/src/lib/commits/utils.ts
```diff
@@ -35,14 +36,20 @@ export function findLastDivergentCommit(
...
 		const previousPathBytes =
 			change.status.type === 'Rename' ? change.status.subject.previousPathBytes : null;
+		const assignment = assignments?.[change.path];
+		const hunkHeaders = assignment?.map((a) => a.hunkHeader).filter(isDefined) ?? [];
+
 		return {
 			previousPathBytes,
 			pathBytes: change.pathBytes,
-			hunkHeaders: []
+			hunkHeaders
```
estib-vega, line 52: Check whether there is some assignment information related to this file. if so, use it

PR #9081, opened 2025-06-18
Enable reordering of stacks

mtsgrd:
This isn't styled, but it should be working.


PavelLaptev (approved):

PavelLaptev:
I'll check it out today a bit later and then merge it :-)

PavelLaptev:
@mtsgrd Thank you! I updated only a few things. It would be cool if we can see reordering like in V2 where we have the flip animation

PR #9084, opened 2025-06-18
Add copy commit hash

hosmelq:
## 🧢 Changes

This PR adds an option to `CommitContextMenu` that allows copying the hash to the clipboard.

## ☕️ Reasoning

At times, I need to copy the git hash. For instance, in V3, I have to click on the commit and copy it from the opened panel.

### DIFF
--- a/apps/desktop/src/components/CommitContextMenu.svelte
+++ b/apps/desktop/src/components/CommitContextMenu.svelte
@@ -102,6 +102,13 @@
 			/>
 		{/if}
 		<ContextMenuItem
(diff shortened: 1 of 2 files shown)


Byron:

To me it makes sense to have this feature in the right-click menu as well, and a member of the frontend team will take a look at it shortly!

It would also be interesting to learn why frequent copying of the hash is still necessary in the first place - maybe there is a workflow that GitButler could support better?
Thanks again.

PavelLaptev (commented):
Looks good, thanks! :-)

PR #9115, opened 2025-06-19
Create persiste Workflow entries

krlvi:
These are currently being triggered by an MCP action. Since it's ran async, we need a separate entry with it's status etc


krlvi:
@estib-vega i added an entity for the async tasks that get are being performed as a result of a MCP call to update_branches. This way we can build the designs we discussed today. 

In the future this should also be able to support more advanced workflows as well

PR #9133, opened 2025-06-23
feat(windows): add custom Windows title bar to replace the default

SilkePilon:
- Introduce a new WindowsTitleBar component for a native-like title
  bar experience on Windows platform.
- Conditionally render WindowsTitleBar in the main layout when running
  on Windows.
- Modify window creation logic to disable default window decorations
  (title bar) on Windows, enabling the custom title bar to be used.
- Keep default decorations on other platforms to preserve native look.
- Improve app branding by showing version and build type badges in the
  custom title bar.
- Enhance user experience with zoom controls and window management
  integrated into the custom title bar.

This change improves UI consistency and usability for Windows users by
providing a tailored window chrome that matches platform conventions.

- [ ] Linux tests
- [x] MaxOS tests
- [x] Windows tests


Byron:

Here is a screenshot for how it looks on MacOS - the window in the foreground is the current nightly.

Screenshot 2025-06-23 at 16 43 54


We probably have to check it on Linux as well, but I will skip this for now.

I also pushed some minor fixes, leaving only some CSS related lints unaddressed.

SilkePilon:
The current issue is i cant build or preview my changes on my pc atm, as i need to restart my pc every time i make an change as the port becomes locked

SilkePilon:
Here is an demo of how it will look on windows (old ui):

{D95DAC5B-834F-4D49-A2AC-49F7168E4A3F}

SilkePilon:
Needs to be tested on mac and linux to see if this does not break anything on those platforms

SilkePilon:
Fully working on windows 1:1

SilkePilon:
Here is an demo video of the new ui (current):

https://github.com/user-attachments/assets/468fd8e0-989d-4d9a-bbc0-24411803918a

SilkePilon:
Added an preview badge to display in what IDE the project gets opened

image

SilkePilon:
The custom titlebar is now optional (will be moved down in the settings menu in the future)

https://github.com/user-attachments/assets/a702f0fd-048c-4b98-adb1-2e54a2321fb7

SilkePilon:
Is now at the bottom:

image

Byron:
Wow, I am so impressed! And the new custom menu bar integrated into the title on Windows is something I thought I wanted to have on MacOS as well.

And even though there seem to have been side-effects, I think the other platforms are something we will be able to take care of.

Screenshot 2025-06-24 at 04 51 51

SilkePilon:
> And even though there seem to have been side-effects, I think the other platforms are something we will be able to take care of.
> 
> Screenshot 2025-06-24 at 04 51 51

Will have an look at this issue, seems to be related to the css i changed

SilkePilon:
@Byron the issue should be fixed now

Byron:
That's great 🎉, I can confirm that 🙏.

Screenshot 2025-06-24 at 10 05 28

Should we assume that MacOS is the same as Linux?

And also, is it possible to have the menubar integrated into the title bar on other platforms as well? I think this will be useful on MacOS, and on Linux it will also look better than the built-in menu bar.

SilkePilon:
> And also, is it possible to have the menubar integrated into the title bar on other platforms as well? I think this will be useful on MacOS, and on Linux it will also look better than the built-in menu bar.

It's fully possible at the moment - there's an if-check that determines if the platform is Windows before showing the custom title bar, which can be removed to make it active for every platform. However, the issue is that on Mac, the close, minimize, and maximize buttons have a different style. If we used this approach on Mac, it would look out of place in my opinion, though this can be adjusted if needed.

SilkePilon:
However, based on a suggestion from Pavel Laptev, I'm probably going to change the menu item selectors to look close to this:
image
But I'll keep the close, minimize, and maximize button section the same as it is (at least for Windows).

PavelLaptev:
> However, based on a suggestion from Pavel Laptev, I'm probably going to change the menu item selectors to look close to this

@SilkePilon thank you for the PR and your work! 🙏  This is exactly what I was hoping for — it would be great if we can make it look like the screenshot. The goal is to have it resemble a native menu, but with a more polished, refined look.

Here are a few good examples — custom, but subtle.

image
image
image

PavelLaptev:
@SilkePilon here are colors that I used:

- Header background: `--clr-bg-3`
- text and icons: `--clr-text-1` but menu items have 0.5 opacity


image

---

Font sizes are default, I think it's 12px

SilkePilon:
Ready for review!

SilkePilon:
Needs to be retested on mac and linux

krlvi:
Thank you a ton for this PR! I am doing some testing - as a minor issue, in the light theme the hover state of these buttons is very faint
Capture1

Additionally, at least on windows 10 it looks like a very thin border is missing at the top of the frame. I'm including a screenshot from the nightly as a comparison. At the same time, VSCode has no such border at all...

Capture2
Capture3
Capture4

apps/desktop/src/lib/backend/tauri.ts
```diff
@@ -1,10 +1,37 @@
...
+		const isMaximized = await window.isMaximized();
+		if (isMaximized) {
+			await window.unmaximize();
+		} else {
+			await window.maximize();
+		}
+	}
+
+	async close() {
+		const window = getCurrentWindow();
```
mtsgrd: Since we work with an instance of this class, could we make `window` a private class field? `private window = getCurrentWindow()`

apps/desktop/src/lib/backend/tauri.ts
```diff
@@ -1,10 +1,37 @@
...
 import { check } from '@tauri-apps/plugin-updater';
+import { getCurrentWindow } from '@tauri-apps/api/window';
 
 export class Tauri {
 	invoke = invokeIpc;
 	listen = listenIpc;
 	checkUpdate = check;
 	currentVersion = getVersion;
+
+	// Window control methods
```
mtsgrd: I'm a bit allergic to comments that don't really convey new information. Could we maybe drop this one?

packages/ui/src/lib/Badge.svelte
```diff
@@ -24,15 +25,21 @@
...
 	}: Props = $props();
 </script>
 
 <Tooltip text={tooltip}>
 	<!-- A badge is not a clickable UI element, but with exceptions. No button styling desired. -->
 	<!-- svelte-ignore a11y_no_static_element_interactions -->
 	<!-- svelte-ignore a11y_click_events_have_key_events -->
-	<div class="badge {style} {kind} {size}-size" class:reversedDirection {onclick}>
+	<div
+		class="badge {style} {kind} {size}-size"
```
mtsgrd: To be consistent with the rest of the code base, could you declare as follows?
```ts
<div
    class="badge"
    class:style
    class:kind
    ...
>
...
</div>
```

apps/desktop/src/components/WindowsTitleBar.svelte
```diff
@@ -0,0 +1,555 @@
...
+	import { shortcuts } from '$lib/utils/hotkeys';
+	import { openExternalUrl, getEditorUri } from '$lib/utils/url';
+	import { getContext } from '@gitbutler/shared/context';
+	import { getContextStoreBySymbol } from '@gitbutler/shared/context';
+	import Badge from '@gitbutler/ui/Badge.svelte';
+	import ContextMenuItem from '@gitbutler/ui/ContextMenuItem.svelte';
+	import ContextMenuSection from '@gitbutler/ui/ContextMenuSection.svelte';
+	import DropDownButton from '@gitbutler/ui/DropDownButton.svelte';
+	import KeyboardShortcutsModal from '$components/KeyboardShortcutsModal.svelte';
+	import type { Writable } from 'svelte/store';
```
mtsgrd, line 20: It might be that you need to build `packages/no-relative-imports` for vscode to automatically sort your imports?

apps/desktop/src/components/WindowsTitleBar.svelte
```diff
@@ -0,0 +1,555 @@
...
+	let buildType = $state<'stable' | 'nightly' | 'dev'>('stable');
+	let appIcon = $state<string>('');
+
+	// Update project when page route changes
+	$effect(() => {
+		$page.url.pathname; // Reactive dependency
+		projectsService
+			.getActiveProject()
+			.then((activeProject) => (project = activeProject))
+			.catch(() => (project = undefined));
```
mtsgrd: I'm a bit confused by this, could you help me understand what you are solving for?

apps/desktop/src/components/WindowsTitleBar.svelte
```diff
@@ -0,0 +1,555 @@
...
+	const getBadgeText = () => {
+		switch (buildType) {
+			case 'nightly':
+				return `Nightly v${appVersion}`;
+			case 'dev':
+				return 'Development';
+			default:
+				return appVersion || 'Stable';
+		}
+	};
```
mtsgrd: For consistency with the rest of the code, could you refactor this into a named function pls?

apps/desktop/src/components/WindowsTitleBar.svelte
```diff
@@ -0,0 +1,555 @@
...
+	const getBadgeStyle = () => {
+		switch (buildType) {
+			case 'nightly':
+				return 'warning';
+			case 'dev':
+				return 'error';
+			default:
+				return 'neutral';
+		}
+	};
```
mtsgrd: Same here.

apps/desktop/src/components/WindowsTitleBar.svelte
```diff
@@ -0,0 +1,555 @@
...
+
+	// Zoom functionality
+	const ZOOM_CONFIG = { MIN: 0.375, MAX: 3, DEFAULT: 1, STEP: 0.0625 };
+
+	function updateZoom(newZoom: number) {
+		const zoom = Math.min(Math.max(newZoom, ZOOM_CONFIG.MIN), ZOOM_CONFIG.MAX);
+		document.documentElement.style.fontSize = zoom + 'rem';
+		userSettings.update((s) => ({ ...s, zoom }));
+	}
```
mtsgrd: What would it take to reuse existing zoom code here?

apps/desktop/src/components/WindowsTitleBar.svelte
```diff
@@ -0,0 +1,555 @@
...
+	// Menu action handlers
+	const menuActions = {
+		addLocalRepository: () => projectsService.addProject(),
+		cloneRepository: () => goto(clonePath()),
+		switchTheme: () =>
+			userSettings.update((s) => ({ ...s, theme: s.theme === 'light' ? 'dark' : 'light' })),
+		zoomIn: () => updateZoom($userSettings.zoom + ZOOM_CONFIG.STEP),
+		zoomOut: () => updateZoom($userSettings.zoom - ZOOM_CONFIG.STEP),
+		resetZoom: () => updateZoom(ZOOM_CONFIG.DEFAULT),
+		openDevTools: () => import.meta.env.DEV && console.log('Opening developer tools...'),
```
mtsgrd: 👀

apps/desktop/src/components/WindowsTitleBar.svelte
```diff
@@ -0,0 +1,555 @@
...
+		'share-debug': menuActions.shareDebugInfo
+	}).forEach(([key, handler]) => shortcutService.on(key, handler));
+
+	// Always show custom title bar on Windows
+	const showTitleBar = $derived(platformName === 'windows');
+</script>
+
+{#snippet editorBadgeSnippet()}
+	{#if editorDisplayName}
+		<Badge style="neutral" size="icon" borderRadius="var(--radius-s)">
```
mtsgrd: We should avoid passing css values through props like this, but I'm also wondering if we really want to make the badge radius configurable. @PavelLaptev could you help out here?

apps/desktop/src/components/WindowsTitleBar.svelte
```diff
@@ -0,0 +1,555 @@
...
+			>
+				<svg width="10" height="10" viewBox="0 0 10 10">
+					<path d="M1 1l8 8M9 1L1 9" stroke="currentColor" stroke-width="1" />
+				</svg>
+			</button>
+		</div>
+	</div>
+{/if}
+
+<KeyboardShortcutsModal bind:this={keyboardShortcutsModal} />
```
mtsgrd: Is there any way we can avoid having a second instance of this modal? It's currently present in `ChromeSidebar.svelte`.

apps/desktop/src/components/WindowsTitleBar.svelte
```diff
@@ -0,0 +1,555 @@
...
+		margin: 0;
+	}
+
+	/* Hide dropdown icons, separators, and vertical lines */
+	.title-bar__menu :global(.dropdown-wrapper .btn .icon),
+	.title-bar__menu :global(.dropdown-wrapper .btn .separator),
+	.title-bar__menu :global(.dropdown-wrapper .btn svg),
+	.title-bar__menu :global(.dropdown-wrapper .btn::after),
+	.title-bar__menu :global(.dropdown-wrapper::after),
+	.title-bar__menu :global(.separator) {
```
mtsgrd, line 584: Could you look at customizing the dropdown through props rather than overriding css using `:global`?

mtsgrd (commented):
Awesome stuff, sorry for leaving so many comments(!)

Could you also address the check failures from github actions? Unfortunately eslint will fail silently until you build `packages/no-relative-imports`.

apps/desktop/src/lib/backend/tauri.ts
SilkePilon: Made `window` a private class field (`private window = getCurrentWindow()`) and updated all methods to use `this.window` instead of calling `getCurrentWindow()` repeatedly. This improves performance and follows better OOP practices.

apps/desktop/src/lib/backend/tauri.ts
SilkePilon: Removed the "Window control methods" comment as it didn't provide meaningful information beyond what the method names already convey.

packages/ui/src/lib/Badge.svelte
SilkePilon: Reverted the `borderRadius` prop addition from Badge.svelte entirely and formatted the component to match the codebase conventions with proper line spacing and indentation.

apps/desktop/src/components/WindowsTitleBar.svelte
SilkePilon: Built the `packages/no-relative-imports` package which resolved the ESLint/import sorting issues. Imports are now properly merged and sorted in WindowsTitleBar.svelte.

apps/desktop/src/components/WindowsTitleBar.svelte
SilkePilon: Improved the comment to "Ensures project updates when route changes" which better explains that this creates a reactive dependency on the pathname to trigger project reloading when navigation occurs.

apps/desktop/src/components/WindowsTitleBar.svelte
SilkePilon: Refactored the project update $effect to call a named function `updateProjectFromRoute()` for consistency with the rest of the codebase.

apps/desktop/src/components/WindowsTitleBar.svelte
SilkePilon: Removed the duplicated zoom logic and now reuse the same constants and update function pattern used in `ZoomInOutMenuAction.svelte`. The zoom functionality now uses shared `ZOOM_CONFIG` constants and a `updateZoom` function that matches the existing pattern exactly.

apps/desktop/src/components/WindowsTitleBar.svelte
SilkePilon: Removed the `borderRadius` prop from Badge.svelte entirely and removed its usage in WindowsTitleBar.svelte. The Badge component now uses its default styling without prop-based customization.

apps/desktop/src/components/WindowsTitleBar.svelte
SilkePilon: Removed the `KeyboardShortcutsModal` import, state, and component usage from WindowsTitleBar.svelte. The keyboard shortcuts action now relies on the global shortcut registration system instead of maintaining a local modal instance, avoiding duplication with the modal in ChromeSidebar.svelte.

apps/desktop/src/components/WindowsTitleBar.svelte
SilkePilon: Checked the DropDownButton component and confirmed it doesn't support prop-based visual customization for title bar styling. The global CSS overrides are necessary but have been made more targeted and removed the `!important` declarations. The styling is now scoped specifically to `.title-bar__menu` context and follows a cleaner approach.

SilkePilon:
@mtsgrd Thanks for the review, resolved most of the issues.

SilkePilon:
@krlvi — this seems to be a Windows 10-specific issue; it works fine on Windows 11. I’ll look into it, though given that Windows 10 is nearing end-of-life, we might not need to address it.

SilkePilon:
Lint seems to be failing atm

krlvi:
> @krlvi — this seems to be a Windows 10-specific issue; it works fine on Windows 11. I’ll look into it, though given that Windows 10 is nearing end-of-life, we might not need to address it.

Understood! No need to worry about windows 10 then - I just didnt know :)

crates/gitbutler-tauri/src/window.rs
```diff
@@ -255,6 +255,22 @@ pub fn create(
...
+    )
+    .resizable(true)
+    .title(handle.package_info().name.clone())
+    .disable_drag_drop_handler()
+    .min_inner_size(1000.0, 600.0)
+    .inner_size(1160.0, 720.0)
+    .decorations(true) // Start with decorations enabled, frontend will disable if user has custom title bar enabled
+    .build()?;
+
+    #[cfg(not(any(target_os = "windows", target_os = "macos")))]
```
Caleb-T-Owens, line 273: Is this not identical to below?

apps/desktop/src/components/WindowsTitleBar.svelte
```diff
@@ -556,7 +555,7 @@
 
 	.native-control-button:hover {
 		background-color: rgba(255, 255, 255, 0.1);
-	}
+	}
```
SilkePilon: ?

PavelLaptev: might be auto-formatting. I'll clean this up

packages/ui/src/lib/Badge.svelte
```diff
@@ -32,7 +32,21 @@
...
+		class:neutral={style === 'neutral'}
+		class:pop={style === 'pop'}
+		class:success={style === 'success'}
+		class:error={style === 'error'}
+		class:warning={style === 'warning'}
+		class:purple={style === 'purple'}
+		class:solid={kind === 'solid'}
+		class:soft={kind === 'soft'}
+		class:icon-size={size === 'icon'}
+		class:tag-size={size === 'tag'}
```
PavelLaptev: @SilkePilon let's keep the code simple. That's why it was `{style} {kind}`

PavelLaptev (changes requested):

PavelLaptev:
So what I would like to avoid is code duplication. Here we repeat all menu items that we have in the rust file. And I already see some inconsistencies like this:

Screenshot 2025-06-25 090456

There should be "..." at the end of the "Share debug info" because it opens a modal it's not an instant "send" action.

@krlvi is there a way to handle this?

PavelLaptev:
@SilkePilon When you click on the menu item it opens the menu and when click again it should close the menu, not trigger it. Should work like a toggle:

https://github.com/user-attachments/assets/e9320949-00a2-4b87-a322-4df5038640ef

PavelLaptev:
Another thing is that it looks like the custom title bar can maximize the window, but not minimize it:

https://github.com/user-attachments/assets/720d06b9-a02b-46cc-904f-e1b8f97e1de4

packages/ui/src/lib/Badge.svelte
SilkePilon: reverted

SilkePilon:
> Another thing is that it looks like the custom title bar can maximize the window, but not minimize it:

Weird, this seems to work for me just fine.

SilkePilon:
> @SilkePilon When you click on the menu item it opens the menu and when click again it should close the menu, not trigger it. Should work like a toggle:
> 
>  Screen.Recording.2025-06-25.072019.mp4

Change it so its an toggle

PavelLaptev:
> Change it so its an toggle

Thanks! but it should be smarter I think because currently I have to click two times to trigger the dropdown (check the video below)

https://github.com/user-attachments/assets/8fb6aec9-ff56-471f-a57b-71e434a11926

PavelLaptev:
This issue is caused because you modified the dropdown component using global styles:

> Thanks! But it should be smarter, I think, because currently I have to click twice to trigger the dropdown (check the video below).

However, this is not actually a dropdown button component. A proper dropdown button component triggers the menu when clicking on the arrow and hide the menu just fine. But in your implementation not the arrow triggers the menu to show, it's the whole button. 

https://github.com/user-attachments/assets/ea4cdef0-6ed8-4afd-8974-efa6e999040d

Let's remove the dropdown component from here. It doesn’t make sense to use a component whose logic and styles have been overridden. Instead, it should be a custom element specifically designed for this custom title bar.

@mtsgrd you also asked if we can modify the dropdown button. In my view, it's not a dropdown button component anymore — it looks and work differently.

So it could be just a simple `button` but styled

SilkePilon:
>Let's remove the dropdown component from here. It doesn’t make sense to use a component whose logic and styles have been overridden. Instead, it should be a custom element specifically designed for this custom title bar.

👍 I'm on it

Issue #9128, opened 2025-06-23
Uncommitted Changes don't respond initially, but then even show other project's changes across restarts


Byron:
### Version

Version 0.14.35 (20250620.070100)

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

It's very strange, but the file `Releases.md` isn't even in the [small.zip](https://github.com/user-attachments/files/20856354/small.zip) project.

<img width="730" alt="Image" src="https://github.com/user-attachments/assets/db47b46d-37a9-4bb9-b345-c056353625b6" />

### How to reproduce

* Open the `small` project
* note that worktree changes aren't tracked, it's like filesystem monitoring isn't working even though it triggering an action
* go to another project and observe a worktree change there
* go back to `small` and notice that the last seen worktree changes persist
* restart the application and note that the worktree changes persist

### Relevant log output (Optional)

```shell

```

Byron:
Probably this is only a problem with strange branch configurations like in my test repository, but in case it's not, CC @estib-vega .

Byron:
Removing `.git/gitbutler/but.sqlite` does the trick.

The problem is a broken database migration which prevents the app from seeing any workspace update.

PR #9137, opened 2025-06-23
more robust app startup in the light of database failures

Byron:
Helps to make #9128 more obvious by allowing `set_project_active` to check for possible errors.

### Tasks

* [x] implementation
* [x] test UI and see if this can be a popup.
* [ ] @estib-vega the is_exclusive check doesn't seem to kick in in my testing and I don't know why. Maybe it's the application caching this value somehow?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3277,6 +3277,7 @@ dependencies = [
  "backtrace",
  "but-action",
  "but-core",
+ "but-db",
  "but-graph",
  "but-hunk-assignment",
  "but-hunk-dependency",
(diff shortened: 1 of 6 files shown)


Byron:
@krlvi I think this is a good way of dealing with the kind of error we discovered in #9128, and I'd hope @estib-vega could move it over the finishing line.

And even though I have seen the new error on database corruption…
Screenshot 2025-06-23 at 15 05 18


…I have not seen this popup anymore for some reason that is probably me being very stupid somewhere.

Screenshot 2025-06-23 at 15 05 52

Thanks for your help with this @estib-vega (the branch is in the gitbutler repo so you can take it over).

krlvi:
So it will automatically move the file in case of an error like this? I like it!

Byron:
Yes, it's the same logic as used for `projects.json`, and adapted so that bubbles up to the UI.
It's also in the logs for good measure :).

krlvi (approved):

PR #9152, opened 2025-06-24
always-show-locking-information

Caleb-T-Owens:
## Goal

-   Show a lock icon next to a file or diff to indicate that it is locked to a
    specific branch.

## Investigation

-   We now have lock information from the HunkAssignments.
    -   We can also get file dependencies through it's own endpoint.
    -   The issue is that the HunkAssignment information is fallible. This means
        that we probably shouldn't rely on it at the minute?
-   We're going to compute the locking information separately and then pass it
    into the hunk assignments seperatly.
    -   We could either pass in just the `diffs` property of
        the `HunkDependencies` struct, or we could just pass the entire thing.
    -   I think passing the entire think makes more sense. The other property is the
        `errors` struct which could be helpful in the future.
-   We already have the `DependencyService` which is currently responsible for
    returning locking information. It feels like there are a handful of options that
    we could take to adapt the existing logic:
    1.  Inject the `WorktreeService` into the `DependencyService` and then use
        the `worktreeData` endpoint to get the dependencies.
    2.  Remove the `DependencyService` in favor of having it's existing methods
        sit in `WorktreeService`
    -   For now I am inclined to consume the `WorktreeService` from inside the
        `DependencyService`. This is because the `DependencyService` is using an
        entity adapter, and it feels like there is _enough_ logic going on there
        that it warrents it's own home.
-   `getLineLocks` is currently used to extract individual lines that are locked
    from the the information.
    -   Currently, this takes a `excludeStackId` as a filter, which hides a
        particular stack from the output, and if filter is `undefined`, it will
        filter out everything.
        -   In `Unassigned changes` we presumibly will always want to show the
            changes.
        -   In `Assigned changes` we may still want to show them, or might want
            to show them differently.
            -   Talking to pavel, it seems like we are going to want to show
                them differently. So I'm going to remove the input `stackId`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3400,6 +3400,7 @@ dependencies = [
  "anyhow",
  "but-core",
  "but-hunk-assignment",
+ "but-hunk-dependency",
  "but-settings",
  "gitbutler-branch-actions",
  "gitbutler-command-context",
--- a/apps/desktop/cypress/e2e/errorHandling.cy.ts
+++ b/apps/desktop/cypress/e2e/errorHandling.cy.ts
@@ -24,9 +24,6 @@ describe('Error handling - commit actions', () => {
 		mockCommand('undo_commit', () => {
 			throw new Error(COMMIT_UNDO_ERROR_MESSAGE);
 		});
-		mockCommand('hunk_dependencies_for_workspace_changes', (params) =>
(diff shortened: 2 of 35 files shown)


Caleb-T-Owens:
bugbot run

Caleb-T-Owens:
bugbot run

PR #9162, opened 2025-06-25
Update sveltey deps

Caleb-T-Owens:


estib-vega:
It seems that this update broke the deployments to `gitbutler-web` on Vercel.

https://community.vercel.com/t/svelte-builds-fail-file-etc-ssl-certs-does-not-exist/13231

PR #9191, opened 2025-06-27
Fix git prune killing oplog

Caleb-T-Owens:


krlvi:
OMG such a great find! How did you find it?

PR #9193, opened 2025-06-27
DB-hook experiment.

Byron:
Here is an example on how to do update hooks.
However, it turns out these only work for same-process changes.

Possible avenues would be to implement polling, but that makes it hard to detect updates, while additions and deletions would be quite straightforward.
It's also quite inefficient to do polling, but maybe with our table sizes it would be alright.

Brute force we could just compare the entire table stable.

### Tasks

* [x] try `update_hook()` - doesn't work, not inter-process
* [x] sample brute-force implementation per table
* [x] remove `rusqlite` as `libsqlite3-sys` is enough.

### Notes for the reviewer

* Only tested with `cargo run -p but-testing -- --app-suffix nightly -d -C /Users/byron/dev-gb/gitbutler watch-db` - it detected assignment changes correctly.
* If this really becomes a cornerstone of the application, it *should* be changed to watching timestamps on the databsae instead, or at least be gated by that.


crates/but-testing/src/command/mod.rs
```diff
@@ -469,6 +470,22 @@ pub async fn watch(args: &super::Args) -> anyhow::Result<()> {
     }
     Ok(())
 }
+pub fn watch_db(args: &super::Args) -> anyhow::Result<()> {
+    let (_repo, project) = repo_and_maybe_project(args, RepositoryOpenMode::General)?;
+    let project = project.context("Couldn't find GitButler project in directory, needed here")?;
+    let mut ctx = CommandContext::open(&project, AppSettings::default())?;
+    let db = ctx.db()?;
+    let rx = db.poll_changes(
+        ItemKind::Actions | ItemKind::Assignments | ItemKind::Workflows,
```
krlvi, line 479: I wonder - do we gain anything performance wise by not checking certain tables? For example assignments are set from the same process (the gui app) so in theory we only care about actions and workflows. Or is it just about the same amount of work?

Byron: This command wants to check everything for testing purposes. But other subscribers can register for a subset and that will impose less work.
To me this implementation is just something that seems to work for now, and that should probably be revised once it acts up.

estib-vega (approved):

PR #9194, opened 2025-06-27
ai client: Prefer the GB client credentials

estib-vega:
When using the actions, prefer the GB client credentials


krlvi:
I wonder why though - if a user has openAI configured, nothing wrong with it, and may even be less running cost

estib-vega:
> I wonder why though - if a user has openAI configured, nothing wrong with it, and may even be less running cost

This is more of a band-aid fix.
Ideally, this reads what you have configured as the source and not just what api keys are available.
Both can be available at the same time

estib-vega:
I can fix it properly in a bit

krlvi:
no rush - i was just curious

PR #9210, opened 2025-06-30
Commit flow fixes and tests: Your commit goes here position and commit selection after creation

estib-vega:
### Description

- ~~Select the new commit automatically after successful creation~~.
- Pass the selected commit ID as the parent when creating a new commit.
- Improve commit flow tests to verify commit parent correctness and UI behavior.
- Set commit row test IDs to their parent commits for better test accuracy.
- Ensure "Your commit goes here" position is correctly asserted in tests.
- Select stack ID when clicking on one of its commits.
- Add an alias for the GitHub API umbrella interceptor for cleaner code.

### DIFF
--- a/apps/desktop/cypress/e2e/commitActions.cy.ts
+++ b/apps/desktop/cypress/e2e/commitActions.cy.ts
@@ -1,6 +1,7 @@
 import { clearCommandMocks, mockCommand } from './support';
 import MockBackend from './support/mock/backend';
 import { PROJECT_ID } from './support/mock/projects';
+import BranchesWithChanges from './support/scenarios/branchesWithChanges';
 import LotsOfFileChanges from './support/scenarios/lotsOfFileChanges';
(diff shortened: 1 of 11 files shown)


krlvi:
> Select the new commit automatically after successful creation.

I wonder - is this desirable though? It introduces content shift... and arguably not needed?

estib-vega:
> > Select the new commit automatically after successful creation.
> 
> I wonder - is this desirable though? It introduces content shift... and arguably not needed?

Good point.
The intention is that when you're committing multiple times in the same branch, not having the auto-select would make it so that the user would have to manually click on the parent.

Content shift is also not great, though.

I'll get rid of that for now, and then we can take a closer look at the workflow for creating multiple commits in a row

krlvi:
I think we used to have this and we took it out at some point because it was feeling un-ergonomic

PR #9226, opened 2025-06-30
Fix the performance issue of the DB subscription

estib-vega:
Make the DB watcher fast by creating tokio async threads and only subscribing (and unsubscribing) to the DB when setting the project active
  
- Removed unnecessary Rustend and frontend database-related code and services.
- Updated DB watcher to instantiate when setting the default project, terminating any active watcher before starting a new one.
- Introduced async polling for item changes using Tokio, replacing synchronous polling with an efficient async task.
- Refactored DB watcher to use async polling, improving performance and preventing UI blocking.


estib-vega:
During manual testing this, I noticed that even when creating the watcher thread when the project was set didn't improve things.

So what I did was to create tokio-async version of the polling thread (similar to how the file watcher does) to see if that made it faster and **it did**!

This works for me in dev, but any other improvements to this are very much welcome

estib-vega:
We should release in any case before merging this

crates/but-db/src/poll.rs
```diff
@@ -87,4 +89,141 @@ impl DbHandle {
             })?;
         Ok(rx)
     }
+
+    /// Register async polling at `interval` for any `kind` of data and return a channel to be informed about the changes
+    /// for the respective kind. Uses async tokio task and channel for efficiency.
+    pub fn poll_changes_async(
```
Byron, line 95: The sync/threaded version can now probably be removed. The CLI uses it, but either it's changed to `async` which is already done there, or it's removed entirely. No problem, it was just for showing that it works, but that can now be seen in other ways as well.

apps/desktop/src/routes/[projectId]/workspace/+page.svelte
```diff
@@ -33,16 +27,6 @@
 	const firstStack = $derived(firstStackResult.current.data);
 	const stackResult = $derived(stackId ? stackService.stackById(projectId, stackId) : undefined);
 
-	// Start watching the database for changes.
```
Comment (Byron) [inline apps/desktop/src/routes/[projectId]/workspace/+page.svelte:36]: Maybe there is a plan for this commented out section?

Byron (approved):
Thanks for hailing me :)!

And sorry for the os-threaded initial implementation, I had no clue it would end up being integrated into the user interface. Also it's good to know that anything with threads that isn't working so well with `tauri` and its `tokio` based concurrency.

With that said, I think this is a good copy of all the mechanism that the filesystem watcher employs, and it seems to work the same.

Personally I think that, at least compared to the os-threaded version, there is much more needed, seemingly, to interrupt a task and that's a little strange, but now is not the time to question this.

apps/desktop/src/routes/[projectId]/workspace/+page.svelte
Comment (estib-vega) [reply apps/desktop/src/routes/[projectId]/workspace/+page.svelte:36]: No need, this change removes these previously commented out lines

crates/but-db/src/poll.rs
estib-vega: I'll leave it for now, just to have this in the CLI

PR #9231, opened 2025-07-01
fix auto-commit and improve it

estib-vega:
Fix the auto-commit workflow and add some improvements on top.

- fix: Use the v2 stacks call to get the data about the applied stacks, this used to be broken because it would be able to find empty branches
- Improvement: Commit tool creates a branch is none is found.
    This speeds up the workflow by reducing a two-tool call operation to one
- Improvement: Tool failures won't stop a workflow, we just encode the tool error into the response

### DIFF
--- a/crates/but-tools/src/workspace.rs
+++ b/crates/but-tools/src/workspace.rs
@@ -1,9 +1,11 @@
 use std::sync::Arc;
 
-use but_graph::VirtualBranchesTomlMetadata;
+use but_workspace::ui::StackEntry;
 use gitbutler_command_context::CommandContext;
 use gitbutler_oplog::{OplogExt, SnapshotExt};
+use gitbutler_stack::{PatchReferenceUpdate, VirtualBranchesHandle};
 use schemars::{JsonSchema, schema_for};
+use serde_json::json;
 
 use crate::emit::EmitStackUpdate;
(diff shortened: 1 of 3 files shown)


crates/but-tools/src/workspace.rs
```diff
@@ -122,43 +239,69 @@ pub fn create_commit(
...
 
 fn stacks(
     ctx: &CommandContext,
     repo: &gix::Repository,
 ) -> anyhow::Result<Vec<but_workspace::ui::StackEntry>> {
-    let meta = VirtualBranchesTomlMetadata::from_path(
-        ctx.project().gb_dir().join("virtual_branches.toml"),
-    )?;
-    but_workspace::stacks_v3(repo, &meta, but_workspace::StacksFilter::InWorkspace)
+    but_workspace::stacks(
```
krlvi, line 249: You can do 
```rust
    if ctx.app_settings().feature_flags.ws3 {
        let meta = ref_metadata_toml(ctx.project())?;
        but_workspace::stacks_v3(&repo, &meta, filter.unwrap_or_default())
    } else {
        but_workspace::stacks(&ctx, &project.gb_dir(), &repo, filter.unwrap_or_default())
    }
    .map_err(Into::into)
    ```
    and this way the code will be transitioned smoothly when ws3 si ready

PR #9236, opened 2025-07-01
Rewrote stack reordering to use simpler (and not broken) logic

Caleb-T-Owens:
This moves the “ondragover” handler to the lanes themselves rather than the whole workspace view.

This means we can avoid writing any logic with reguards to calculating widths and can just focus on simply when to or when not to swap the elements


apps/desktop/src/components/v3/MultiStackView.svelte
```diff
@@ -117,7 +121,7 @@
 	});
 
 	// Throttle calls to the reordering code in order to save some cpu cycles.
-	const throttledDragOver = throttle(onReorderDragOver, 250);
```
Caleb-T-Owens, line 120: Making it 25 instead of 250 makes it much more responsive and the performance is still  good

PR #9266, opened 2025-07-03
fix symlink committing

Byron:
It seems symlinks are added as symlinks, but their content being the content of the file they
link to, instead of their original value as read by `readlink()`.

This is a huge oversight and I can't believe that I didn't test this precisely enough.
It also means that all symlinks currently created by the app are broken.

### Task

* [x] reproduce in Nightly
* [x] reproduce in test or check existing tests accuracy
    - it's not obviously broken on a low level
    - workspace changes are shown correctly
    - it's probably related to the UI passing in hunk information, so the new link is a single-line change `path: "ho", hunk_headers: [HunkHeader("-1,0", "+1,1")]`
* [x] fix - allow it, but do it correctly
     - This works as ultimately the operation on a single-line buffer is valid, and we work with Git directly without having to deal with the worktree.
* [x] what about discarding? Added protection, barking loudly, so this can be handled better in future if needed.
    - The UI will allow passing hunks when discarding individual lines in a symlnk, but it doesn't do anything.

### DIFF
--- a/crates/but-db/Cargo.toml
+++ b/crates/but-db/Cargo.toml
@@ -23,7 +23,7 @@ anyhow = "1.0.98"
 diesel_migrations = { version = "2.0.0", features = ["sqlite"] }
 chrono = { version = "0.4.41", features = ["serde"] }
 # other things
-tokio = { workspace = true, features = ["rt-multi-thread", "parking_lot"] }
+tokio = { workspace = true, features = ["rt-multi-thread", "parking_lot", "time", "sync"] }
 tracing.workspace = true
 
 [dev-dependencies]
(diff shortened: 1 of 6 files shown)


Byron:
@PavelLaptev Would it be possible to get a different icon for symlink changes?
Below, `ho` is a symlink, the other change is a file.

Screenshot 2025-07-03 at 13 41 17

I noticed that I was continuously confused when looking at the changes and kept resorting to the CLI just to be sure I was actually looking at a symlink change. The frontend already knows the type of the change, it could use a different icon.
Thanks for your help 🙏

Byron:
Another interesting case @PavelLaptev from a UX perspective.
When files have been renamed, there is a label that says so, but there is no indication what the file original was. Technically this is part of the change which, just like hunks, should be visualised.
In the following case a file was modified and renamed, but the rename source is invisible.

Screenshot 2025-07-03 at 13 51 13

Do we have a design for that?

Byron:
@estib-vega Ideally, when the UI sees that all hunks are selected when committing a file, it should pass no hunks at all to the backend. The backend can then be in whole-file mode, which is always faster and better.

PR #9286, opened 2025-07-04
Refactor claude hooks

krlvi:


krlvi:
fyi @schacon the hooks are now 
`but claude pre-tool`
`but claude post-tool`
`but claude stop`

PR #9314, opened 2025-07-08
Refactor command name handling for back end queries

mtsgrd:
In order to better capture metrics we need the backend command name to 
be defined on the rtkq queries directly, rather than as part of the 
arguments to the base query.

What would be ideal is if we could modify the `MutationDefintion` to 
have the extra properties directly, rather than constantly nesting them 
inside `extraOptions`.

For now I am leaving `build.query()` definitions alone, and reliant on 
passing `command` through the query args. Before we migrate them over we
need to implement the posthog stuff on the query hooks (the ones touched
here are all mutations).

### DIFF
--- a/apps/desktop/src/lib/actions/actionService.svelte.ts
+++ b/apps/desktop/src/lib/actions/actionService.svelte.ts
@@ -35,10 +35,12 @@ function injectEndpoints(api: ClientState['backendApi']) {
 	return api.injectEndpoints({
 		endpoints: (build) => ({
 			autoCommit: build.mutation<void, { projectId: string; changes: TreeChange[] }>({
-				query: ({ projectId, changes }) => ({
+				extraOptions: {
 					command: 'auto_commit',
-					params: { projectId, changes },
 					actionName: 'Figure out where to commit the given changes'
+				},
+				query: ({ projectId, changes }) => ({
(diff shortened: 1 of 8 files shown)


apps/desktop/src/lib/actions/actionService.svelte.ts
```diff
@@ -47,10 +49,12 @@ function injectEndpoints(api: ClientState['backendApi']) {
...
 			autoBranchChanges: build.mutation<void, { projectId: string; changes: TreeChange[] }>({
-				query: ({ projectId, changes }) => ({
+				extraOptions: {
 					command: 'auto_branch_changes',
-					params: { projectId, changes },
 					actionName: 'Create a branch for the given changes'
+				},
+				query: ({ projectId, changes }) => ({
+					params: { projectId, changes }
 				}),
```
mtsgrd, line 58: All these kinds of arguments can be reduced to `(x)=>x` once all the `build.query()` endpoints have been updated to use extraOptions.

README.md
```diff
@@ -1,3 +1,5 @@
+Hello
```
krlvi: Hi :)

mtsgrd: 😂

PR #9365, opened 2025-07-12
Add Cypress component test for MessageEditor 

mtsgrd:
- nabled component testing by replacing reactive variables with Svelte 
  ;stores and removing $effect usage in .svelte.ts constructors.
- Sacrificed lifecycle management and automatic unsubscription for simpler testing setup.
- Improved browser compatibility to support Cypress component testing.
- Added a Cypress component test for the CommitMessageEditor.


krlvi:

PR #9429, opened 2025-07-16
Update copy of drop to assign placeholder

krlvi:
image

Trying a different copy here, FYI @PavelLaptev so you can update figma


PavelLaptev:
Got it, thanks!

PR #9465, opened 2025-07-18
Reduxify Mode Service and start working on RemoveFile removal

Caleb-T-Owens:


apps/desktop/src/components/EditMode.svelte
```diff
@@ -181,23 +161,34 @@
...
+			const { conflictEntryPresence } = initialFileMap.get(fileEntry.path) || {};
+			if (!conflictEntryPresence) return 'unknown';
+			return getConflictState(conflictEntryPresence, file.data.content);
+		});
+
+		const manuallyResolved = toStore(() => manuallyResolvedFiles.has(fileEntry.path));
+
+		return derived([conflictState, manuallyResolved], ([conflictState, manuallyResolved]) => {
+			return fileEntry.conflicted && conflictState === 'conflicted' && !manuallyResolved;
+		});
```
estib-vega, line 184: Is it possible to only read the file if the fileEntry.conflicted is true? Just to avoid to make that call if not needed

Caleb-T-Owens: `conflictedEntryPresence` is only set if there is a conflict. `fileEntry.conflicted` is actually set based off of the presence of `conflictedEntryPresence`

apps/desktop/src/routes/+layout.svelte
```diff
@@ -151,6 +152,10 @@
 	);
 	const feedFactory = new FeedFactory(data.tauri, stackService);
 	const rulesService = new RulesService(clientState['backendApi']);
+	const modeService = $derived(new ModeService(clientState['backendApi']));
+	$effect.pre(() => {
+		setContext(ModeService, modeService);
+	});
```
mtsgrd, line 158: Why do you have a `$derived()` on line 155? It doesn't look necessary.

PR #9494, opened 2025-07-21
Remove unnecessary derived

estib-vega:
Fetching the name of the branch doesn’t need to be double-wrapped in a derived


estib-vega:
@mtsgrd  I think we don't need the other derived

PR #9503, opened 2025-07-21
POC: Dynamically generated Backend service

estib-vega:
Create the hooks dynamically out of the query definitions and expose them from a singleton class. The rules service consumes this

### DIFF
--- a/apps/desktop/src/lib/backend/backendService.svelte.ts
+++ b/apps/desktop/src/lib/backend/backendService.svelte.ts
@@ -1,1 +1,155 @@
-export default class BackendService {}
+import {
+	createMutationEndpoint,
+	createQueryEndpointWithTransform,
+	type CustomBuilder,
+	type EndpointMap
+} from '$lib/state/butlerModule';
+import { invalidatesItem, invalidatesList, providesItems, ReduxTag } from '$lib/state/tags';
+import { createEntityAdapter, type EntityState } from '@reduxjs/toolkit';
+import type {
+	CreateRuleRequest,
+	UpdateRuleRequest,
+	WorkspaceRule,
+	WorkspaceRuleId
+} from '$lib/rules/rule';
+import type { BackendApi } from '$lib/state/clientState.svelte';
+
+function typedEntries<T extends Record<string, unknown>>(obj: T): [keyof T, T[keyof T]][] {
+	return Object.entries(obj) as [keyof T, T[keyof T]][];
+}
+
+function typedFromEntries<T extends Record<string, unknown>>(entries: [keyof T, T[keyof T]][]): T {
+	return Object.fromEntries(entries) as T;
+}
+
+export default class BackendService {
+	private static instance: BackendService;
+	private mutationApi: ReturnType<typeof injectMutationEndpoints>;
+	private queryApi: ReturnType<typeof injectQueryEndpoints>;
+
+	private constructor(backendApi: BackendApi) {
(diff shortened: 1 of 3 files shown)


apps/desktop/src/lib/backend/backendService.svelte.ts
```diff
@@ -1 +1,155 @@
...
+		>(
+			builder,
+			'list_workspace_rules',
+			(response: WorkspaceRule[]) => {
+				return workspaceRulesAdapter.addMany(workspaceRulesAdapter.getInitialState(), response);
+			},
+			(result) => providesItems(ReduxTag.WorkspaceRules, result?.ids ?? [])
+		)
+	} satisfies EndpointMap;
+}
```
estib-vega, line 149: So first we defined the endpoints.
The injection happens for queries and mutations separately in order to preserve the type safety when iterating over the entries

apps/desktop/src/lib/backend/backendService.svelte.ts
```diff
@@ -1 +1,155 @@
...
+
+export default class BackendService {
+	private static instance: BackendService;
+	private mutationApi: ReturnType<typeof injectMutationEndpoints>;
+	private queryApi: ReturnType<typeof injectQueryEndpoints>;
+
+	private constructor(backendApi: BackendApi) {
+		this.mutationApi = injectMutationEndpoints(backendApi);
+		this.queryApi = injectQueryEndpoints(backendApi);
+	}
```
estib-vega, line 33: We then store them in private properties

apps/desktop/src/lib/backend/backendService.svelte.ts
```diff
@@ -1 +1,155 @@
...
+			)
+		) as FetchMap;
+
+		return {
+			...mutate,
+			...useMutation,
+			...useQuery,
+			...fetchMap
+		};
+	}
```
estib-vega, line 95: And then, out of the endpoint entries, we generate the map of functions

apps/desktop/src/lib/rules/rulesService.svelte.ts
```diff
@@ -1,87 +1,31 @@
...
 
 export default class RulesService {
-	private api: ReturnType<typeof injectEndpoints>;
+	private backendService: BackendService;
+	private apis: ReturnType<typeof this.backendService.get>;
 
 	constructor(backendApi: BackendApi) {
-		this.api = injectEndpoints(backendApi);
+		this.backendService = BackendService.getInstance(backendApi);
+		this.apis = this.backendService.get();
```
estib-vega, line 10: The backend service here is consumed as a singleton

apps/desktop/src/lib/rules/rulesService.svelte.ts
```diff
@@ -1,87 +1,31 @@
...
-		return this.api.endpoints.updateWorkspaceRule.useMutation();
+		return this.apis.updateWorkspaceRuleUseMutation();
 	}
 
 	listWorkspaceRules(projectId: string) {
-		return this.api.endpoints.listWorkspaceRules.useQuery(
+		return this.apis.listWorkspaceRulesUseQuery(
 			{ projectId },
 			{ transform: (result) => workspaceRulesSelectors.selectAll(result) }
 		);
```
estib-vega, line 29: If the result is transformed into an entity state instance, we need to select it when consuming it.

Issue #8457, opened 2025-05-08
GitButler thinks branch is not incorperated upstream.


Icantjuddle:
### Version

Version 0.14.19 (20250428.160452)

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

After merging a PR made with gitbutler using the squash and merge strategy on github, my local gitbutler client will not remove my local branch and rather thinks it is conflicted. 

![Image](https://github.com/user-attachments/assets/349411f7-617c-41cd-9b2e-3edc2192ffe6)

### How to reproduce (Optional)

1. create a PR with with gb
2. Merge with squash and merge and delete the branch on github
3. Update workspace on local gitbutler

### Expected behavior (Optional)

It recognizes that the branch is incorporated upstream and offers to delete the local branch.

### Relevant log output (Optional)

```shell

```

Byron:
Thanks a lot for reporting!
It seems that the failure to detect that the branch is already integrated allowed to perform a rebase which then had to cause a conflict.
I hope we can tackle this in the next major release.

Icantjuddle:
Is there a better way to work around this thank to manually fixup the `.git/gitbutler/virtual_branches.toml` to make it forget about it? 

UI operations seem to end up in some loop of conflicted states that cannot be resolved?

Byron:
> Altering the configuration file directly might be possible - there are flags for whether something is integrated or not. Maybe [@krlvi](https://github.com/krlvi) can advise?

krlvi:
I have an idea for approaching this.

The root cause here is that whenever there is a squash commit, it is difficult to recognize that the branch that is still in the workspace is indeed integrated - the commits will be different, and the commit message of the squash commit is not reliable either. 

The app also looks at the content - i.e. "does the incoming version contain the exact code that is in the branch". But if there were follow up changes, e.g. the branch was squash merged but then the code was modified by another merged PR it can no longer detect it. 

--

So I am thinking about implementing a "treat as integrated" override option for the workspace update operation. This override can be triggered by the UI, based on the Pull Request status.

This solution is a little bit cheating because it would not work universally, especially if a person does not have a forge integration enabled. But maybe helping some of the cases is better than helping none

Byron:
[...]

PR #9504, opened 2025-07-21
workspace robustness

Byron:
‼️ if branches in the workspace are advanced outside of the workspace, it currently doesn't see these commits and the branch looses its name in the workspace.
The traversal really should try to add the known tips explicitly for traversal and expose that information.

### Tasks

- [x] What about scenarios where there is no stack ID?
    - we already have a bunch of cases where there is no stack-id, and for now we don't set it as the UI as to be able to deal with it being optional. Depending on the usecase, there can probably be adjustments once we know what we need.
- [x] fixes around amending and integration checks
- [x] take a look at misclassification below
- [ ] expose segments that were moved out of the workspace

### Next related tasks

* 🏎️ Use per-commit metadata to avoid recomputing changeset IDs for each commit, enabling processing more and more commits with the 1s compute budget.
* 🏎️ graph-based merge-base computation would be much faster if a bitmap was used. Could be intrinsic or separate, with intrinsic certainly being better.
* ⚠️ fix related TODOs and known issues

### Not to forget

* There might be an issue with the way it uses searches - a tip with a search might be blocked at an existing commit, discovered by a tip with a different search, and even though the thing it searches is reachable through that, it stops looking.
* the amount of commits of remotes ahead of their local branch doesn't seem to always match git (particularly when it's a lot of them)

### Related issues

* #8457
* #9408

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1385,7 +1385,7 @@ version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fde0e0ec90c9dfb3b4b1a0891a7dcd0e2bffde2f7efed5fe7c9bb00e5bfb915e"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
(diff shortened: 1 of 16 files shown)


Byron:
CleanShot_2025-07-22_at_10 19 25

[Thread on Discord](https://discord.com/channels/1060193121130000425/1397132709691916350)

CleanShot_2025-07-22_at_12 19 49

Byron:
Early merge to get bugfixes in.

PR #9515, opened 2025-07-22
Provide a special hint on how to deal with a missing keychain

### LINKED ISSUE #9507: [0.15.0] secrets error spam upon startup
### Version

0.15.3

### Operating System

Linux

### Distribution Method

deb (Linux)

### Describe the issue

After upgrading to 0.15.3, I get the following error thrown a bunch (see also screenshot below):

```
The name org.freedesktop.secrets was not provided by any .service files
```

Instead of this, I used to get an error about my keyring being deleted, or something along those lines.

The impact is that I have to dismiss all the errors, and re-login to both gitbutler and github whenever I restart my computer.

I am running GitButler in an Ubuntu distrobox on Bluefin.

Image

_No response_

_No response_

```shell

```

Byron:
On linux it's possible to not have any keychain to allow persistence of secrets beyond reboot.
Then even attempts to read passwords will result in an error like:

```
The name org.freedesktop.secrets was not provided by any .service files
```

The solution is to install a keychain, which we will now propose.

Related #9507

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3270,6 +3270,7 @@ name = "gitbutler-secret"
 version = "0.0.0"
 dependencies = [
  "anyhow",
+ "gitbutler-error",
  "gix",
  "keyring",
  "serde",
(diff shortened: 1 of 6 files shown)


Byron:
@Caleb-T-Owens Here is what I whipped up to provide a helpful and not too bad looking message in case there is no keychain on linux.

Screenshot 2025-07-22 at 16 14 21

However, since secrets retrieval is happening a lot particularly after opening a project, there is a lot of error boxes.

Do you think it's possible to deduplicate it somehow?

This PR I think is ready to be merged, but maybe it's something that sparks your interest after improving the hint system today.

PR #9520, opened 2025-07-23
optional stack-id for single-branch support

Byron:
That way it's possible to start consuming the underlying workspace data, instead of relying on `stack_details`.

Doing so would also enable GB to follow the user to any checked-out branch.

Related to #9516, which would be needed to recover from commits that put a stack-segment's ref outside of the workspace.

### Tasks

* [ ] ~~backend endpoint and RefInfo serialisation~~
    - not for now, prefer a small step.
* [x] stacks returns stackentry with optional stack-id
* [x] stack_details with optional stack-id
* [x] test for `stacks` and `stack_details()` outside workspace
* [x] adapt frontend to optional stack-id
* [ ] ~~(optional) let `ws3` not respond with a 'need workspace' screen, but instead let it refresh the view~~
    - Even though I could change the backend, I would have no idea how to make this do anything useful in the frontend.
       Thus the frontend still wouldn't encounter a non-workspace branch in the wild.
* [x] validate UI still works

### Notes for the reviewer

* I validated that the UI still works normally with limited UI testing.
* the old `stacks` and `stack_details` APIs are retrofitted to use optional `stack_id`, indicating that there is no workspace and just a single stack.
* detached heads can't be represented with the old API as branches must have names, it needs the new `head_info()` data.
* The frontend is adjusted to stop missing stack-ids at runtime, to allow adjusting it in a controlled fashion.
* The backend code that consumes the old APIs, `stacks` and `stack_details`, was adjusted to basically filter-out and ignore such stacks, and I suppose it would go to `head_info` directly.

### DIFF
--- a/apps/desktop/src/lib/stacks/stackService.svelte.ts
+++ b/apps/desktop/src/lib/stacks/stackService.svelte.ts
@@ -34,7 +34,7 @@ import type { LocalFile } from '$lib/files/file';
 import type { DefaultForgeFactory } from '$lib/forge/forgeFactory.svelte';
 import type { TreeChange, TreeChanges } from '$lib/hunks/change';
 import type { DiffSpec, Hunk } from '$lib/hunks/hunk';
-import type { BranchDetails, Stack, StackDetails } from '$lib/stacks/stack';
+import type { BranchDetails, Stack, StackOpt, StackDetails } from '$lib/stacks/stack';
 import type { PropertiesFn } from '$lib/state/customHooks.svelte';
 
 type BranchParams = {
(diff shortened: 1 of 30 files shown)


apps/desktop/src/lib/stacks/stackService.svelte.ts
```diff
@@ -846,24 +846,34 @@ export class StackService {
...
 }
 
+function transformStacksResponse(response: Stack[]) {
+	response.forEach((stack) => {
+		// To keep it simple, what's cast as `Stack` is actually `StackOpt`
+		// (as returned by the backend).
+		// So here we cast it back and stop any optional stack-id in its tracks
+		// until the code can actually cope with it.
+		const stackOpt = stack as StackOpt;
+		if (!stackOpt.id) {
```
Byron, line 858: It's notable that this can only happens once the backend sends a new operating mode, but that's not implemented at all.
That could be a viable next step once it's clear how that should look like.
I was thinking of introducing a new operating mode, for now, which is feature-toggled with `ws3` as well.

In future, however, I'd think there is no such thing as an operating mode anymore, and instead information could be derived from the reference name that `HEAD` points to, like "We see you have are looking at a commit with `HEAD` set to `gitbutler/edit`, so you must be in edit mode right now".

PR #9547, opened 2025-07-24
Fix: commit title stale value on click after generate

estib-vega:
Fixes an issue where clicking on the commit title immediately after generating it would apply a stale value due to improper state updates. Updates CommitMessageEditor to correctly synchronize title/description with parent state on AI generation. Refactors NewCommitView to robustly update new commit message state using partial updates, preventing loss of information when updating title or description separately.


apps/desktop/src/components/NewCommitView.svelte
```diff
@@ -219,11 +219,23 @@
...
+				...newCommitMessageUpdate,
+				description
+			};
+		}
+
+		if (newCommitMessageUpdate) {
+			stackState.newCommitMessage.set({
+				...stackState.newCommitMessage.current,
+				...newCommitMessageUpdate
+			});
```
estib-vega, line 236: UI state doesn't seem to behave in a synchronous way

PR #9552, opened 2025-07-24
Open in finder

PavelLaptev:
## 🧢 Changes

There are two `FileContextMenu.svelte` files. I only updated the one for V3. We should probably refactor and remove the old one later, but I’ll do that in a separate PR.

Added "Open In Finder" in two places.

image
image

### DIFF
--- a/apps/desktop/src/components/ProjectSettingsMenuAction.svelte
+++ b/apps/desktop/src/components/ProjectSettingsMenuAction.svelte
@@ -37,6 +37,16 @@
 		);
 	});
 
+	shortcutService.on('show-in-finder', async () => {
+		const project = await projectsService.fetchProject(projectId);
+		if (!project) {
+			throw new Error(`Project not found: ${projectId}`);
+		}
+		// Show the project directory in the default file manager (cross-platform)
+		const { invoke } = await import('@tauri-apps/api/core');
+		await invoke('show_in_finder', { path: project.path });
+	});
+
 	shortcutService.on('history', () => {
 		$showHistoryView = !$showHistoryView;
 	});
--- a/apps/desktop/src/components/ProjectSettingsMenuAction.svelte
+++ b/apps/desktop/src/components/ProjectSettingsMenuAction.svelte
@@ -8,7 +8,7 @@
 	import { SHORTCUT_SERVICE } from '$lib/shortcuts/shortcutService.svelte';
 	import * as events from '$lib/utils/events';
 	import { unsubscribe } from '$lib/utils/unsubscribe';
-	import { getEditorUri, openExternalUrl } from '$lib/utils/url';
+	import { getEditorUri, openExternalUrl, showFileInFolder } from '$lib/utils/url';
 	import { inject } from '@gitbutler/shared/context';
 	import { onMount } from 'svelte';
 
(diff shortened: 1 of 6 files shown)


crates/gitbutler-tauri/src/main.rs
```diff
@@ -267,6 +267,8 @@ fn main() {
                     modes::edit_initial_index_state,
                     modes::edit_changes_from_initial,
                     open::open_url,
+                    #[cfg(target_os = "macos")]
```
estib-vega: I think you need to remote this config flag so that the command is also available in other platforms :)

PavelLaptev: updated :-)

apps/desktop/src/components/ProjectSettingsMenuAction.svelte
```diff
@@ -37,6 +37,16 @@
...
 	});
 
+	shortcutService.on('show-in-finder', async () => {
+		const project = await projectsService.fetchProject(projectId);
+		if (!project) {
+			throw new Error(`Project not found: ${projectId}`);
+		}
+		// Show the project directory in the default file manager (cross-platform)
+		const { invoke } = await import('@tauri-apps/api/core');
+		await invoke('show_in_finder', { path: project.path });
```
Caleb-T-Owens: Could you use the `showFileInFolder` function you defined further down?

Caleb-T-Owens: As a note, we should really try to avoid dynamic imports (the `await import` syntax)  because it makes static analysis of the code significantly harder. Currently we don't do anything like that, but there may come a point where we try to do tree shaking or something.

As another note, we should always use the invoke function defined in `$lib/backend/ipc`  because it does some extra error handling for us.

Your function defined down below is already using the correct `invoke`

apps/desktop/src/lib/utils/url.ts
```diff
@@ -21,6 +21,15 @@ export async function openExternalUrl(href: string) {
...
 
+export async function showFileInFolder(filePath: string) {
+	try {
+		await invoke<void>('show_in_finder', { path: filePath });
+	} catch (error) {
+		console.error('Failed to show in file manager:', error);
+		throw error;
+	}
+}
+
```
Caleb-T-Owens: We can entirely remove this try/catch here since we have global error handling set up

apps/desktop/src/components/v3/FileContextMenu.svelte
```diff
@@ -373,6 +386,25 @@
...
+									const projectPath = project?.path;
+									if (projectPath) {
+										const absPath = await join(projectPath, changes[0]!.path);
+										await showFileInFolder(absPath);
+									}
+									contextMenu.close();
+								} catch (error) {
+									console.error('Failed to show in file manager:', error);
+									toasts.error('Failed to show in file manager');
+								}
```
Caleb-T-Owens: If we removed this try/catch it would mean the error is handled by the global error handler, which means it would get fed into posthog so we could get some details about how many users this fails for.

Using a toast might be a design choice  though... though I wonder if we could do something that lets it still go through the global error handler so we still get the posthog reports

PR #9596, opened 2025-07-28
fix: use correct symbol for referencing branches on forges

H1ghBre4k3r:
## 🧢 Changes

Use the correct symbol for referencing branches on different forges. 

## ☕️ Reasoning

The issue arose when using stacked branches on GitLab in a merge request. We used `#` instead of `!` which worked fine for GitHub but did not work  for GitLab. 

Now, we utilize the already present information in the `ForgePrService`  such that we always use the correct symbol when referencing branches.

### DIFF
--- a/apps/desktop/src/lib/forge/shared/prFooter.ts
+++ b/apps/desktop/src/lib/forge/shared/prFooter.ts
@@ -20,7 +20,7 @@ export async function updatePrDescriptionTables(prService: ForgePrService, prNum
 		const prs = await Promise.all(prNumbers.map(async (id) => await prService.fetch(id)));
 		const updates = prs.filter(isDefined).map((pr) => ({
 			prNumber: pr.number,
-			description: updateBody(pr.body, pr.number, prNumbers)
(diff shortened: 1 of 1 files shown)


apps/desktop/src/lib/forge/shared/prFooter.ts
```diff
@@ -33,18 +33,23 @@ export async function updatePrDescriptionTables(prService: ForgePrService, prNum
 /**
  * Replaces or inserts a new footer into an existing body of text.
  */
-function updateBody(body: string | undefined, prNumber: number, allPrNumbers: number[]) {
+function updateBody(
```
Caleb-T-Owens, line 36: Could we pass the symbol in here too, so that only `updatePrDescriptionTables` needs a ForgeService in order to be tested?

H1ghBre4k3r: Sure, that would reduce the needed mocks!

Caleb-T-Owens (approved):
LGTM. Fire at will

PR #9607, opened 2025-07-28
Run pre-commit hooks on amend

mtsgrd:
How It Works:

1. When a user drags changes to amend a commit:
- Frontend calls runPreCommitHooks() with the changes
- If hooks fail, operation stops with error message
- If hooks succeed, amend request is sent to backend

2. Backend receives amend request:
- Creates a temporary tree with the changes
- Runs hooks::pre_commit_with_tree()
- If hooks fail, returns error
- If hooks succeed, proceeds with amend

Both frontend and backend now compile successfully, and pre-commit
hooks will run before amend operations just like they do before regular 
commits!

### DIFF
--- a/apps/desktop/src/components/BranchCommitList.svelte
+++ b/apps/desktop/src/components/BranchCommitList.svelte
@@ -18,6 +18,7 @@
 		type DzCommitData,
 		SquashCommitDzHandler
 	} from '$lib/commits/dropHandler';
+	import { projectRunCommitHooks } from '$lib/config/config';
 	import { DRAG_STATE_SERVICE } from '$lib/dragging/dragStateService.svelte';
 	import { draggableCommitV3 } from '$lib/dragging/draggable';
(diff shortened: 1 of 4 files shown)


crates/gitbutler-branch-actions/src/actions.rs
```diff
@@ -240,6 +241,38 @@ pub fn amend(
 ) -> Result<git2::Oid> {
     ctx.verify(ctx.project().exclusive_worktree_access().write_permission())?;
     ensure_open_workspace_mode(ctx).context("Amending a commit requires open workspace mode")?;
+
+    // Run pre-commit hooks if changes are provided
```
krlvi: Why do we need to run the hook here (as well)?

krlvi (changes requested):
I'd like to discuss the approach here. In the commit case, the hook execution is a discrete operation, performed separately. Why do we choose to conflate the hook execution with the amend command?

krlvi (approved):

PR #9624, opened 2025-07-29
feat(settings): set ws3 feature flag default to true

krlvi:


Byron:
I have no objections 👍 - let's do it!

PR #9625, opened 2025-07-30
Reduce app shared state removing project controller

krlvi:


crates/gitbutler-testsupport/src/lib.rs
```diff
@@ -289,3 +286,9 @@ pub(crate) static DRIVER: Lazy<PathBuf> = Lazy::new(|| {
 /// Note that this can't be used if secrets themselves are under test as it' doesn't act
 /// like a real store, i.e. stored secrets can't be retrieved anymore.
 pub mod secrets;
+
+pub fn set_test_data_dir<P: AsRef<std::path::Path>>(path: P) {
+    unsafe {
+        std::env::set_var("TEST_APP_DATA_DIR", path.as_ref());
```
Byron, line 292: Not sure about this - it's set to a temp path, which is usually done per test, so tests would interfere with each other now.

Byron (commented):
I think using an env-var for tests (with code setting it to a tempdir) is definitely asking for trouble.

This is where I think code that needs application directories should have a handle to some object to allow injections and isolation.

crates/gitbutler-testsupport/src/lib.rs
krlvi: Discussed on discord... In general, we want to rely on "project" less and less in the app. And these are mostly legacy tests

PR #9626, opened 2025-07-30
New look — new me!

estib-vega:
### Description

- Added reusable GitHub and Google OAuth button components for login/signup pages.
- Introduced redirect guard components to manage user navigation based on authentication and account finalization status.
- Refactored Navigation and Header to use new login/signup flow with webRoutes and removed old redirect logic.
- Updated app layout to provide LoginService, integrate redirect guards, and conditionally show Navigation.
- Display logged out view with prompt on main app page when user is not logged in.
- Fixed user login checks to use USER_SERVICE instead of auth token for project reviews and rules pages.
- Added new authentication and user-related pages including login, signup, finalize account, resend confirmation, and password reset flows.
- Refactored root layout to improve authentication and token restoration logic using webRoutes.


aspleenic (approved):
THis looks good to me.

PR #9634, opened 2025-07-30
Leave settings and feedback enabled at all times

### LINKED ISSUE #9633: Project settings inaccessible if branch isn't `gitbutler/workspace`
This also means that users have to be on `gitbutler/workspace` to remote a project.

There shouldn't be a technical reason for this, so ideally project settings should be enabled at all times.

Byron:
Still needs testing, does it work at all?

Fixes #9633.

Everything below the line is AI generated.

### Tasks

* [x] see that 'feedback' works - it does ✅
* [x] see that settings work - they don't ❌
* [x] don't try to enable project settings - seems to be harder

----

Here's a well-formatted, concise PR description:

---

# Enable project settings access outside workspace branch

## Problem
Project settings are disabled when users switch away from `gitbutler/workspace`, preventing access to repository configuration.

## Solution
Always enable the project settings and feedback buttons in the sidebar, regardless of current branch.

### Changes
- Project settings button: `{disabled}` → `disabled={false}`
- Feedback button: `{disabled}` → `disabled={false}`

### Rationale
Project settings manage repository-level configuration (git config, commit signing, credentials) that doesn't require workspace branch access. Workspace-dependent features (Workspace, Branches, History) remain appropriately disabled.

### DIFF
--- a/apps/desktop/src/components/ChromeSidebar.svelte
+++ b/apps/desktop/src/components/ChromeSidebar.svelte
@@ -283,7 +283,7 @@
 				onclick={() => {
 					shareIssueModal?.show();
 				}}
-				{disabled}
+				disabled={false}
 			/>
 		</div>
 	</div>


Byron:
It looks like project settings really don't work outside of the gitbutler workspace branch. Maybe this is fixable?

https://github.com/user-attachments/assets/e50b71d7-8a2e-4ae0-9e62-03cf1f7dea58

Byron:
@estib-vega The PR can probably be merged as is as the feedback button definitely works. 
However, the project-settings don't work that way - if it's easy, can they be enabled? If not, I'd hope the PR can be merged as is. 
Thank you.

estib-vega:
Hmm I see.

apps/desktop/src/components/ChromeSidebar.svelte
```diff
@@ -283,7 +283,7 @@
 				onclick={() => {
 					shareIssueModal?.show();
 				}}
-				{disabled}
+				disabled={false}
```
estib-vega, line 286: This works as intended, but just for future reference (a small nitpick) this can also be achieved by completely removing the prop, i.e. just having removed line 286.

This basically forces the component to use the default value for the prop in question, which in this case is undefined or false :)

estib-vega:
As for the settings page, the issue is that the button needs to be enabled and the actual settings page content needs to be displayed.

At the root of the project routes **apps/desktop/src/routes/[projectId]/+layout.svelte** we actively hide other pages if the user is not on the workspace branch.

Let me try to fix that for you

apps/desktop/src/components/ChromeSidebar.svelte
Byron: I completely missed your feedback, which tells me something about how I catchup on the GitHub notifications these days 😅.

Thanks for the heads-up, I will keep that in mind.

estib-vega: No no all good!
I basically commented this as I was merging it. Was just a quick drive by

PR #9638, opened 2025-07-31
Enable project settings access outside workspace

### LINKED ISSUE #9633: Project settings inaccessible if branch isn't `gitbutler/workspace`

estib-vega:
- Allow access to project settings from state outside the gitbutler workspace branch by rendering the settings via NotOnGitButlerBranch
- Add the ability to go back to the workspace from project settings (using goto and page events)
- Update ChromeSidebar and layout handling to support navigation differences depending on workspace/settings context
- Update settings page to close back to workspace on exit

### DIFF
--- a/apps/desktop/src/components/ChromeSidebar.svelte
+++ b/apps/desktop/src/components/ChromeSidebar.svelte
@@ -198,13 +198,18 @@
 				{/if}
 				<Button
 					kind="outline"
-					onclick={() => goto(newProjectSettingsPath(projectId))}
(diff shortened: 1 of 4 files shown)


Byron:
That's great, thank you!

This will become so much more valuable once we have single-branch mode as well, settings would definitely be accessible at all times.

PR #9651, opened 2025-08-01
Introduce but-api crate

Caleb-T-Owens:
---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #9652 
- <kbd>&nbsp;1&nbsp;</kbd> #9651 👈

PR #9652, opened 2025-08-01
Introduce but-server

Caleb-T-Owens:
---
- <kbd>&nbsp;2&nbsp;</kbd> #9652 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #9651

### DIFF
--- a/crates/but-server/Cargo.toml
+++ b/crates/but-server/Cargo.toml
@@ -1,0 +1,47 @@
+[package]
+name = "but-server"
+version = "0.0.0"
+edition = "2024"
+repository = "https://github.com/gitbutlerapp/gitbutler"
+license-file = "../../LICENSE.md"
+description = "A GitButler CLI tool"
+authors = ["GitButler <gitbutler@gitbutler.com>"]
+readme = "../../README.md"
+publish = false
+
+[[bin]]
+name = "but-server"
+path = "src/main.rs"
+doctest = false
+
+[features]
+
+[dependencies]
+serde.workspace = true
+axum = { version = "0.8.4", features = ["ws"] }
+futures-util = { version = "0.3", default-features = false, features = [
+    "sink",
+    "std",
+] }
+tower = "0.5.2"
+tower-http = { version = "0.6.5", features = ["cors"] }
+tokio = { workspace = true, features = ["full"] }
+tokio-util = "0.7.15"
+anyhow.workspace = true
+serde_json = "1.0.140"
+tracing.workspace = true
+tracing-subscriber = { version = "0.3", features = [
+    "env-filter",
+    "std",
+    "fmt",
+] }
+reqwest = { version = "0.12", features = ["json"] }
+uuid.workspace = true
+
+but-api.workspace = true
+but-path.workspace = true
+but-settings.workspace = true
+gitbutler-feedback.workspace = true
+gitbutler-user.workspace = true
+gitbutler-project.workspace = true
+gitbutler-watcher.workspace = true
--- a/crates/but-server/src/lib.rs
+++ b/crates/but-server/src/lib.rs
@@ -1,0 +1,469 @@
+use std::sync::Arc;
+
+use axum::{
+    Json, Router,
+    extract::{
+        WebSocketUpgrade,
+        ws::{Message, WebSocket},
+    },
+    response::IntoResponse,
+    routing::{any, get},
+};
+use but_api::{
+    IpcContext, NoParams,
+    broadcaster::Broadcaster,
+    commands::{
+        askpass, cli, config, diff, forge, git, github, modes, open, projects as iprojects,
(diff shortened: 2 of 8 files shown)


crates/but-server/src/lib.rs
```diff
@@ -0,0 +1,470 @@
...
+            any({
+                let broadcaster = broadcaster.clone();
+                async move |req| handle_ws_request(req, broadcaster).await
+            }),
+        )
+        .layer(ServiceBuilder::new().layer(cors));
+
+    // run our app with hyper, listening globally on port 6978
+    let port = std::env::var("BUTLER_PORT").unwrap_or("6978".into());
+    let host = std::env::var("BUTLER_HOST").unwrap_or("0.0.0.0".into());
```
Byron: Should the default listening address be local at least, something like `127.0.0.1`?

I am thinking that our backend server should at least be as locked-down as the tauri backend is.

```
❯ curl -vvv  http://localhost:6667/1cffa8ff-bb56-465e-847d-d57131381786/workspace
* Host localhost:6667 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:6667...
* connect to ::1 port 6667 from ::1 port 52304 failed: Connection refused
*   Trying 127.0.0.1:6667...
* Connected to localhost (127.0.0.1) port 6667
> GET /1cffa8ff-bb56-465e-847d-d57131381786/workspace HTTP/1.1
> Host: localhost:6667
> User-Agent: curl/8.7.1
> Accept: */*
>
* Request completely sent off
* Received HTTP/0.9 when not allowed
* Closing connection
curl: (1) Received HTTP/0.9 when not allowed
```

I have no idea how it works though, and if it's possible to speak HTTP with it at all. That's not to say we have to emulate `tauri`, just to be reasonably sure the server can't be abused.

Maybe running as Electron sidecar automatically sorts out that kind of sandboxing though.

crates/but-server/src/lib.rs
```diff
@@ -0,0 +1,470 @@
...
+fn run_cmd<
+    D: DeserializeOwned,
+    S: Serialize,
+    Fun: Fn(&IpcContext, D) -> Result<S, but_api::error::Error>,
+>(
+    ipc_ctx: &IpcContext,
+    params: serde_json::Value,
+    fun: Fun,
+) -> Result<serde_json::Value, but_api::error::Error> {
+    let result = fun(ipc_ctx, serde_json::from_value(params).to_error()?)?;
```
Byron, line 151: Here it might be interesting to know what happens if `fun` panics. Does this bring down the server? If so, maybe that's what we want initially.

crates/but-server/src/lib.rs
```diff
@@ -0,0 +1,470 @@
...
+    active_projects: Arc<Mutex<ActiveProjects>>,
+}
+
+pub async fn run() {
+    let cors = CorsLayer::new()
+        .allow_methods(Any)
+        .allow_origin(Any)
+        .allow_headers(Any);
+
+    let config_dir = but_path::app_config_dir().unwrap();
```
Byron, line 58: Unwraps are easy to use for debugging and are probably OK for first pre-release versions.

crates/but-server/Cargo.toml
```diff
@@ -0,0 +1,81 @@
...
+gitbutler-secret.workspace = true
+but-workspace.workspace = true
+but-core.workspace = true
+but-hunk-assignment.workspace = true
+gitbutler-branch.workspace = true
+gitbutler-branch-actions.workspace = true
+gitbutler-command-context.workspace = true
+gitbutler-oxidize.workspace = true
+gitbutler-reference.workspace = true
+gitbutler-stack.workspace = true
```
Byron: Aren't most of these in `but-api`?

Byron (commented):
This looks good to me, it's definitely a start and having a platform for e2e tests is of great value.

There is just a question about the fallback listening address which seems too broad.

crates/but-server/src/lib.rs
Caleb-T-Owens: I'll change it to `127.0.0.1` since it's a "more secure" default.


Security in an electron environment is something that we will need to study in more detail if/when we start looking into packaging GitButler in electron

crates/but-server/src/lib.rs
Caleb-T-Owens: Even in release modes, having the app panic if it can't find "essential" directories like this doesn't feel like the _worst_ thing ever.

Before we would ship it in electron we do need to explore how to handle the sidecar dying for whatever reasons, including at startup

crates/but-server/src/lib.rs
Caleb-T-Owens: The thread panics, but the server remains running - which I think is the desired behaviour

crates/but-server/Cargo.toml
Caleb-T-Owens: Yes

PR #9653, opened 2025-08-01
Remove exit calls from main.rs error handling

estib-vega:
This commit removes explicit exit calls (std::process::exit(2)) from the main.rs file in the but crate. The removed exit calls were present in the error handling sections of the ClaudePreTool, ClaudePostTool, and ClaudeStop subcommands. The change avoids abrupt termination of the process on error, which triggers a loop on the CC agent run


estib-vega:
cc @krlvi

PR #9655, opened 2025-08-01
Introduce development electron app

Caleb-T-Owens:
# Introduce development electron app

This does not explore packaging electron or anything like that. This simply uses electron like a browser, similar to the e2e tests.

This PR looks huge, but the majority of that is the lockfile. The real diff is more like **`+350 -10`**


Caleb-T-Owens:
Shit. I hit the wrong button. Reveerting

PR #9668, opened 2025-08-03
Improve upstream commit filtering

estib-vega:
Modified the logic in stacks.rs to filter upstream commits more accurately by comparing tree changes between the last local commit and the upstream commit being looked at. 

Added import for tree_changes function from but_core::diff module. Adjusted upstream commit processing to skip upstream commits that do not introduce new changes relative to the last local commit, preventing redundant upstream commits from being shown after operations such as squash or rebase.

### DIFF
--- a/crates/but-workspace/src/stacks.rs
+++ b/crates/but-workspace/src/stacks.rs
@@ -6,6 +6,7 @@ use crate::{RefInfo, StacksFilter, branch, head_info, ref_info, state_handle, ui
 use anyhow::{Context, bail};
 use bstr::BString;
 use but_core::RefMetadata;
+use but_core::diff::tree_changes;
 use but_graph::VirtualBranchesTomlMetadata;
 use gitbutler_command_context::CommandContext;
 use gitbutler_commit::commit_ext::CommitExt;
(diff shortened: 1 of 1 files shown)


krlvi:
Nice one! I wonder - should we look into porting this into the graph implementation?

Byron:
I think it would need a test so it can be ported [to the RefInfo test-suite](https://github.com/gitbutlerapp/gitbutler/blob/cbfb19d69c1b9d895a947457b97ad429c3511d74/crates/but-workspace/tests/workspace/ref_info/with_workspace_commit/journey/mod.rs#L3).
Otherwise it's ephemeral at best.

Chances are that this is already working there, but also, that it works differently.

estib-vega:
> I think it would need a test so it can be ported [to the RefInfo test-suite](https://github.com/gitbutlerapp/gitbutler/blob/cbfb19d69c1b9d895a947457b97ad429c3511d74/crates/but-workspace/tests/workspace/ref_info/with_workspace_commit/journey/mod.rs#L3). Otherwise it's ephemeral at best.
> 
> Chances are that this is already working there, but also, that it works differently.

Maybe I can add a test in a bit. The WS3 version would look like this: https://github.com/gitbutlerapp/gitbutler/pull/9676

Before merging this, maybe we should discuss how we want to display this information in the brave new world of WS3.

The motivation for this change is to not show commits that are irrelevant, but if we're in a single branch world, maybe we still want to see all commits, event the ones considered 'irrelevant'.

That's why I had only added it to the non-WS3 version of the details.

Issue #8031, opened 2025-04-07
Gitbutler makes pointless rebase/merge for PRs


p1k0chu:
### Version

0.14.14

### Operating System

Linux

### Distribution Method

AppImage (Linux)

### Describe the issue

Every time i apply a pr branch gitbutler rebases it for no reason (requires force push now) or if i disable force pushes it merges the master into pr branch, even if pr is up to date. (merge commit IS empty)

### Relevant log output (Optional)

```shell

```

Byron:
Thanks a lot for reporting, and sorry for the hassle.

Git commit-graphs can be very complex and right now I don't think the analysis of what to do is always correct.
It's definitely on our list of things to address, but it can only happen incrementally, unfortunately.

hey2022:
What is the status on this? (still happens on v0.15.2)
From what I can see, this issue always happens even on simple use cases, such as applying a branch that can be fastforwarded onto an empty workspace.

Byron:
The branch-application logic is scheduled for a complete rewrite, and we will move away from the rebase-by-default policy entirely.

It's still a little while out, but I'd expect these features to trickle in faster once our new technological foundation is in place.

peaklabs-dev:
As the following PR https://github.com/gitbutlerapp/gitbutler/pull/9710 will close the issue where I left more details about the pointless rebases and then in turn needed force pushes. I will link it here for reference and so it does not get lost: https://github.com/gitbutlerapp/gitbutler/issues/2847#issuecomment-3148533731.

**Copy paste out of that issue:**

## `Integrate Upstream` (yellow button)

- It appears on each branch individually as soon as there are changes - checked every minute (with a setting `check interval`) or after a manual check (like it is now).
- **It should also appear on PR branches. Currently, if a PR is applied from a remote (not on the origin) and there is a push/ new change in the PR, the change is only reflected if I hit `Unapply Stack` and then delete the local branch and then reapply the PR branch. Related to: https://github.com/gitbutlerapp/gitbutler/issues/9664**
- When I click `Integrate Upstream` it should directly integrate the commits into the branch and that’s it -> no merge commit (not always possible), no push and certainly no force push needed.
[...]

Issue #2847, opened 2024-02-20
Remote commits are not pulled, causing the danger of overwriting them with local commits


gtupak:
There is currently a way to overwrite the remote commits without knowing. 

To repro:
0. Ensure "Allow force pushing" is not selected in the GitButler's project settings
1. Make a new virtual branch
2. Make a commit `test 1`, push the branch and make a PR
3. In Github, open the PR and edit a file on the branch, and commit it with `test 2`
4. Make a commit locally and call it `test 3`. 
5. Force push the changes - even though "allow force pushing" is not enabled.
6. Notice that the `test 2` commit was overwritten.

This is quite dangerous in a situation where you contribute with other developers on the same branch as you can easily overwrite their work without knowing.

This issue is the result of [this thread on Discord](https://discord.com/channels/1060193121130000425/1073202153163857920/1209610637680971816).

gtupak:
I think this is an even broader problem which is a deal breaker when using the tool: the inability of merging remote changes to the matching virtual branch. This happens all the time when there are at least 2 developers working on the same feature branch :/

esolitos:
☝🏼  I have been wondering about this myself. :) 
From reading the documentation the expected workflow in this case would be to create a new "trunk" branch which is the "base branch" shared with other developers and then use PRs against this base branch.
This should work fine, but it is somewhat an hassle not being able to use pull the changes in any way, but at the same time I understand that this would be problematic to integrate, because then you are updating only the "virtual branch" instead of the "base".

I hope my understanding&explanation is correct and makes sense. 😅

mtsgrd:
We're making some [progress e.g. here](https://github.com/gitbutlerapp/gitbutler/pull/2856), but it's hidden since we need to refactor the [Rust code](https://github.com/gitbutlerapp/gitbutler/blob/0084b1b0ec026248a4dbf08deaea6873b202f161/gitbutler-app/src/virtual_branches/virtual.rs#L1307) that is meant to merge in upstream changes.

Once we have this working again we can build a `--force-with-lease` like flag to prevent accidentally overwriting commits pushed by others!

tgecho:
I'm not sure how it would interact with the rewriting you're doing internally, but I'm also a fan of the `--force-if-includes` flag to help avoid some scenarios where `--force-with-lease` might not cover things.

peaklabs-dev:
[...]

PR #9679, opened 2025-08-04
feat: add `--force-if-includes` flag to force push

peaklabs-dev:
## 🧢 Changes
- feat: add `--force-if-includes` flag to force push

## ☕️ Reasoning

- I am aware that the hole branch code and integrate upstream (yellow and cyan button) is being refactored at the moment, which is great and highly needed. In my opinion, we should almost never need a force push for a lot of actions - see this issue https://github.com/gitbutlerapp/gitbutler/issues/8031 & https://github.com/gitbutlerapp/gitbutler/issues/2847#issuecomment-3148533731 for more details but if we do need to force push for example after a rebase from main, it should be done as securely as possible so that no commits can be lost.
- The `--force-if-includes` protects commits even more than just `--force-with-lease` as it ensures that any remote changes have been integrated locally (not just fetched). For more details see https://stackoverflow.com/a/78298145

## Note

- I am unsure if this is the only change needed to use both flags all the time when force pushing.

### DIFF
--- a/crates/gitbutler-git/src/repository.rs
+++ b/crates/gitbutler-git/src/repository.rs
@@ -362,7 +362,7 @@ where
     args.push(&refspec);
 
     if force {
-        args.push("--force-with-lease");
+        args.push("--force-with-lease --force-if-includes");
     }
 
     let (status, stdout, stderr) =


Byron (approved):

To me this makes sense and I agree this should make force-pushing a little safer. I'd hope that GitButler will be able to help in the case the push then fails as well. Right now, it probably doesn't until it fetches the next time, or the user hits the fetch button.

In any case, I think a second set of eyes would be good @krlvi .

krlvi:
Thanks @peaklabs-dev for researching and sharing this. We have a small discussion over on discord https://discord.com/channels/1060193121130000425/1204524097241878629/1401918272596873307

I think there is an outstanding question if introducing this would restrict the app in an undesirable ways. For instance, there may be legitimate cases where one wishes to overwrite the content on the remote. 

The GitButler push already does a fetch https://github.com/gitbutlerapp/gitbutler/blob/7c103cc65648e61b49a581a4bcfb4b49ade5b410/crates/gitbutler-branch-actions/src/stack.rs#L176 before pushing and @schacon had a suggestion here https://discord.com/channels/1060193121130000425/1204524097241878629/1401922071117955082 about implementing a high level version of this in our push flow, where we pop a modal in the UI and warn the user and ask them what they wanna do.

krlvi:
Okay, since there is a bit of a chance of some people getting blocked by this, I introduced this in a way that can be configured https://github.com/gitbutlerapp/gitbutler/pull/9691

@peaklabs-dev For now there is no UI for this, but you can edit the config file directly (on mac this is `~/Library/Application\ Support/gitbutler/settings.json` )

I think ultimately we could make this true by default and have the ability to disable it as an escape hatch

peaklabs-dev:
Thank you so much. I will try and and close this PR for now.

PR #9682, opened 2025-08-04
V3 of remove_branch

Byron:
Change the existing `create_branch()` and `create_virtual_branch()` functions to
to use the new `but_workspace::branch::create_reference()` function if `ws3` is enabled.

### Tasks

* [x] research
* [x] undo/redo - document progression plan in code
* [x] `remove_reference` (idempotent)
    - [x] also: test with anonymous segment which is now OK to remain
    - [x] assure it won't delete special references
* [x] CLI integration for `create_reference()` to show what it can really do
* [x] integrate V3 into API behind `ws3` toggle
    - [x] use snapshots
* [x] CLI integration for `remove_branch()` to see how it *could* be.
* [x] fix zero-commit-id issue

### Research

* can delete anything as long as one named branch remains to be the stack. No limitations otherwise.
* `metadata` is not deleted, just the reference!

### Performance Tasks

* 🏎️ Expensive computations only for a single stack (the one that updated)
* 🏎️ Use per-commit metadata to avoid recomputing changeset IDs for each commit, enabling processing more and more commits with the 1s compute budget.
* 🏎️ graph-based merge-base computation would be much faster if a bitmap was used. Could be intrinsic or separate, with intrinsic certainly being better.

### Not to forget

* ⚠️Right now there is no occasion where branch-metadata would be deleted, so it's likely to go stale. Ideally we will be able to delete it as soon as it leaves our 'sphere of influence', but at the latest once the local branch doesn't exist anymore. Probably best to have a GC/cleanup step of sorts.
* ⚠️single-branch mode currently doesn't limit itself to only show what's not reachable by extra-target
* There might be an issue with the way it uses searches - a tip with a search might be blocked at an existing commit, discovered by a tip with a different search, and even though the thing it searches is reachable through that, it stops looking.
* the amount of commits of remotes ahead of their local branch doesn't seem to always match git (particularly when it's a lot of them)


Byron:
### Undo Queue Research

* Undo/Redo isn't assumed to be complete and can fail as the user can always perform ops outside of GB that disrupt undo/redo. That's OK, and on error the queue is simply wiped.
* There are non-undoable operations like `push` and `fetch`, and that's totally fine.
* Thanks to the immutability of the object database, any undo operation boils down to setting a reference and setting reference metadata.
    - For that reason, undoing can be fully generalised into a sequence of these operations which can undo/redo themselves. This means one won't have to test undo/redo deeply, a simple test will do as the primitives are well-tested.
    - existing operations can be retro-fitted, like adding it to `create_reference()` and `remove_reference()` is straightforward.
* The undo-queue would be part of the `App` and the API layer is adding commands to the queue. The menu bar deals with interacting with the queue via a simple API.
* Overall, this whole system can be put together in a good day, and it *should* be easy to maintain despite being somewhat invasive (i.e. support is implemented inside of the actual mutating functionality).
* In theory, the sequence of undoable operations that mutating functions now return could be made so it's not yet executed, which would allow tests to be done in read-only mode just by observing that sequence.

#### Outcome

Skip for now, but plan to use snapshot system as vehicle for undo/redo queue.

PR #9700, opened 2025-08-05
Fix ReduxResult rendering with filteredChange

estib-vega:
Don’t render anything if there are no filtered changes


mtsgrd:
@estib-vega should we perhaps add back some kind of empty result logic instead of using an ad-hoc solution this way?

PR #9703, opened 2025-08-05
Fix handling of stop and post tool call for no changes

estib-vega:
Basically, don’t create empty stacks

Modified crates/but-claude/src/hooks/mod.rs to add logic in handle_stop() and handle_post_tool_call() functions to properly check for changes before proceeding. Added early return with appropriate ClaudeHookOutput when no changes are detected, preventing unnecessary further processing and stack creation.


estib-vega:
cc @krlvi

PR #9710, opened 2025-08-05
feat: add force push protections

### LINKED ISSUE #2847: Remote commits are not pulled, causing the danger of overwriting them with local commits

peaklabs-dev:
## ToDo & Limitations

- **Limitation:** Currently when a force push is prevented through the git flags, the `Integrate Upstream` (yellow) button will not work anymore. As @Byron is refactoring that part, this should not be an issue for too long I hope 😀 ~~(or maybe it is an easy fix??)~~.
- ~~**ToDo: https://github.com/gitbutlerapp/gitbutler/pull/9710#issuecomment-3160360029**~~

## 🧢 Changes

- feat(core): add the ability to use more secure flags when force pushing. Every force push when the `Force push protection` setting is enabled will now happen with `--force-with-lease` and `--force-if-includes` instead of with `--force` to help make sure new remote commits are not lost.
- feat(ui): if there is new (not locally integrated) remote commits and `Force push protection` is enabled, a modal will be shown with all the commits that would be overwritten by the force push and allow the user to choose `Force push anyway` which will overwrite the remote commits (sometimes we want that) or to integrate the commits manually.
- feat(ui): add a `Force push protection` setting to the `Git stuff` Project settings page to enable the more secure way of force pushing.
- fix: at the moment when force pushing, the `--force` flag is always used, so the effect of any other `--force...` flag with protection never takes effect
- chore: removed `force_if_includes` settings flag as it is no longer needed

## ☕️ Reasoning

- At the moment there is a big issue with GitButler that a lot of things require force pushing https://github.com/gitbutlerapp/gitbutler/issues/8031, which hopefully will be eliminated in a future version. To make force pushes more secure when they are absolutely needed like after rebasing, there are flags that can be used instead of `--force` to make it less likely to lose any remote commits.
- How the 2 flags work can be read here as the git documentation is quite sparse: https://stackoverflow.com/a/78298145
- At the moment force pushing seems to be done via `--force-with-lease` and not `--force` when looking at the implementation here https://github.com/gitbutlerapp/gitbutler/blob/a86174ccb67d5eeeb59d8ead5876959d1e12e9e4/crates/gitbutler-git/src/repository.rs#L363-L365 Unfortunately this is not working and this flag is never used for 2 reasons:
  - 1: Every time GitButler pushes it first does a fetch, which basically removes all effect of `--force-with-lease` as `force-with-lease` compares the local and remote ref at push time but if we fetch the local branch before pushing in the background, we will be on the same head as the remote which will mean any newly added commits on the remote will be overwritten as this flag now allows us to force push, as it does not see any difference (which means we could loose remote commits).
  - 2: This flag even though it could take effect if we removed the automatic fetch before (and do not manually fetch and integrate the changes before ofc) would still not take effect as we always force push with `--force`. Yes `--force` is always added to the git push command which removes any effect of the flag `--force-with-lease` or even `--force-with-lease` and `--force-if-includes` at the same time. This was one of the hardest things to find as we do not add `--force` explicitly but rather it is added here via a `+` https://github.com/gitbutlerapp/gitbutler/blob/a86174ccb67d5eeeb59d8ead5876959d1e12e9e4/crates/gitbutler-repo-actions/src/repository.rs#L166-L167 A `+` in git is basically the same as adding `--force` to the git push command and as soon as `--force` is added to the git push command none of the other flags that replace `--force` will take effect.
  - As you can see the more secure way of force pushing (that is currently implemented) is never used and at the moment we basically always do a straight `--force` push. 
  - Some time ago I proposed adding a second flag `--force-if-includes` in addition to `--force-with-lease` to protect against the first issue of fetching before pushing but as the `--force-if-includes` flag was never used at all, I decided to introduced a new setting on the `Git stuff` Project settings page to enable `Force pushing protection` which uses both new flags at the same time.

## Issues

- fix https://github.com/gitbutlerapp/gitbutler/issues/2847. Merging this PR should fix this issue completely as overwriting remote commits will now be nearly impossible, so this issue could be closed as soon as this PR is merged.

## Preview

**Git stuff Project settings page:**

image

**Warning modal:**

image

### DIFF
--- a/apps/desktop/src/components/GitForm.svelte
+++ b/apps/desktop/src/components/GitForm.svelte
@@ -16,6 +16,10 @@
 	async function onForcePushClick(project: Project, value: boolean) {
 		await projectsService.updateProject({ ...project, ok_with_force_push: value });
 	}
+
+	async function onForcePushProtectionClick(project: Project, value: boolean) {
+		await projectsService.updateProject({ ...project, force_push_protection: value });
+	}
 </script>
 
 <Section>
@@ -28,22 +32,40 @@
 	<Spacer />
 	<ReduxResult {projectId} result={projectResult.current}>
 		{#snippet children(project)}
-			<SectionCard orientation="row" labelFor="allowForcePush">
-				{#snippet title()}
-					Allow force pushing
-				{/snippet}
-				{#snippet caption()}
-					Force pushing allows GitButler to override branches even if they were pushed to remote.
-					GitButler will never force push to the target branch.
-				{/snippet}
-				{#snippet actions()}
-					<Toggle
-						id="allowForcePush"
-						checked={project.ok_with_force_push}
-						onchange={(checked) => onForcePushClick(project, checked)}
-					/>
-				{/snippet}
-			</SectionCard>
+			<div class="stack-v">
+				<SectionCard orientation="row" labelFor="allowForcePush" roundedBottom={false}>
+					{#snippet title()}
+						Allow force pushing
+					{/snippet}
+					{#snippet caption()}
+						Force pushing allows GitButler to override branches even if they were pushed to remote.
+						GitButler will never force push to the target branch.
+					{/snippet}
+					{#snippet actions()}
+						<Toggle
+							id="allowForcePush"
+							checked={project.ok_with_force_push}
+							onchange={(checked) => onForcePushClick(project, checked)}
+						/>
+					{/snippet}
+				</SectionCard>
+				<SectionCard orientation="row" labelFor="forcePushProtection" roundedTop={false}>
+					{#snippet title()}
+						Force push protection
+					{/snippet}
+					{#snippet caption()}
+						Protect remote commits during force pushes. This will use Git's safer force push flags
+						to avoid overwriting remote commit history.
+					{/snippet}
+					{#snippet actions()}
+						<Toggle
+							id="forcePushProtection"
+							checked={project.force_push_protection}
+							onchange={(checked) => onForcePushProtectionClick(project, checked)}
+						/>
+					{/snippet}
+				</SectionCard>
+			</div>
 		{/snippet}
 	</ReduxResult>
 </Section>
--- a/apps/desktop/src/components/PushButton.svelte
+++ b/apps/desktop/src/components/PushButton.svelte
@@ -9,6 +9,7 @@
 		branchRequiresForcePush
 	} from '$lib/stacks/stack';
 	import { STACK_SERVICE } from '$lib/stacks/stackService.svelte';
+	import { combineResults } from '$lib/state/helpers';
 	import { UI_STATE } from '$lib/state/uiState.svelte';
 	import { getBranchNameFromRef } from '$lib/utils/branch';
 	import { splitMessage } from '$lib/utils/commitMessage';
@@ -50,30 +51,28 @@
 	const branchDetails = $derived(stackService.branchDetails(projectId, stackId, branchName));
 	const projectResult = $derived(projectsService.getProject(projectId));
 	const [pushStack, pushResult] = stackService.pushStack;
-	const potentialOverwrittenCommits = $derived(
-		stackService.upstreamCommits(projectId, stackId, branchName).current.data || []
+	const upstreamCommitsResult = $derived(
+		stackService.upstreamCommits(projectId, stackId, branchName)
 	);
+	const upstreamCommits = $derived(upstreamCommitsResult.current.data);
 
-	let _currentProject = $state<any>(undefined);
-
-	function handleClick(requiresForce: boolean, project?: any) {
-		_currentProject = project;
-
+	function handleClick(args: { withForce: boolean; skipForcePushProtection: boolean }) {
 		if (multipleBranches && !isLastBranchInStack && !$doNotShowPushBelowWarning) {
 			confirmationModal?.show();
 			return;
 		}
 
-		push(requiresForce, project?.force_push_protection);
+		push(args);
 	}
 
-	async function push(requiresForce: boolean, forcePushProtection: boolean) {
+	async function push(args: { withForce: boolean; skipForcePushProtection: boolean }) {
+		const { withForce, skipForcePushProtection } = args;
 		try {
 			const pushResult = await pushStack({
 				projectId,
 				stackId,
-				withForce: requiresForce,
-				forcePushProtection: forcePushProtection,
+				withForce,
+				skipForcePushProtection,
 				branch: branchName
 			});
 
@@ -112,29 +111,25 @@
 	let forcePushProtectionModal = $state<ReturnType<typeof Modal>>();
 </script>
 
-<ReduxResult {projectId} result={branchDetails.current}>
-	{#snippet children(branchDetails)}
-		{@const requiresForce = branchRequiresForcePush(branchDetails)}
+<ReduxResult {projectId} result={combineResults(branchDetails.current, projectResult.current)}>
+	{#snippet children([branchDetails])}
+		{@const withForce = branchRequiresForcePush(branchDetails)}
 		{@const hasThingsToPush = branchHasUnpushedCommits(branchDetails)}
 		{@const hasConflicts = branchHasConflicts(branchDetails)}
 
-		<ReduxResult {projectId} result={projectResult.current}>
-			{#snippet children(project)}
-				<Button
-					testId={TestId.StackPushButton}
-					kind={isFirstBranchInStack ? 'solid' : 'outline'}
-					size="tag"
-					style="neutral"
-					{loading}
-					disabled={!hasThingsToPush || hasConflicts}
-					tooltip={getButtonTooltip(hasThingsToPush, hasConflicts)}
-					onclick={() => handleClick(requiresForce, project)}
-					icon={multipleBranches && !isLastBranchInStack ? 'push-below' : 'push'}
-				>
-					{requiresForce ? 'Force push' : 'Push'}
-				</Button>
-			{/snippet}
-		</ReduxResult>
+		<Button
+			testId={TestId.StackPushButton}
+			kind={isFirstBranchInStack ? 'solid' : 'outline'}
+			size="tag"
+			style="neutral"
+			{loading}
+			disabled={!hasThingsToPush || hasConflicts}
+			tooltip={getButtonTooltip(hasThingsToPush, hasConflicts)}
+			onclick={() => handleClick({ withForce, skipForcePushProtection: false })}
+			icon={multipleBranches && !isLastBranchInStack ? 'push-below' : 'push'}
+		>
+			{withForce ? 'Force push' : 'Push'}
+		</Button>
 
 		<Modal
 			title="Push with dependencies"
@@ -142,7 +137,7 @@
 			bind:this={confirmationModal}
 			onSubmit={async (close) => {
 				close();
-				push(requiresForce, _currentProject?.force_push_protection);
+				push({ withForce, skipForcePushProtection: false });
 			}}
 		>
 			<p>
@@ -163,11 +158,13 @@
 						onclick={() => {
 							$doNotShowPushBelowWarning = false;
 							close();
-						}}>Cancel</Button
+						}}
 					>
-					<Button testId={TestId.StackConfirmPushModalButton} style="pop" type="submit" width={90}
-						>Push</Button
-					>
+						Cancel
+					</Button>
+					<Button testId={TestId.StackConfirmPushModalButton} style="pop" type="submit" width={90}>
+						Push
+					</Button>
 				</div>
 			{/snippet}
 		</Modal>
@@ -179,35 +176,35 @@
 			bind:this={forcePushProtectionModal}
 			onSubmit={async (close) => {
 				close();
-				push(requiresForce, false);
+				push({ withForce, skipForcePushProtection: true });
 			}}
 		>
 			<p class="description">
 				Your force push was blocked because the remote branch contains <span
 					class="text-bold text-nowrap"
-					>{potentialOverwrittenCommits.length === 1
-						? '1 commit'
-						: `${potentialOverwrittenCommits.length} commits`}</span
+					>{upstreamCommits?.length === 1 ? '1 commit' : `${upstreamCommits?.length} commits`}</span
 				>
 				your local branch doesn’t include. To prevent overwriting history,
 				<span class="text-bold">cancel and pull & integrate</span> the changes.
 			</p>
-			<div class="scroll-wrap">
-				<ScrollableContainer maxHeight="16.5rem">
-					{#each potentialOverwrittenCommits as commit}
-						{@const commitUrl = forge.current.commitUrl(commit.id)}
-						<SimpleCommitRow
-							title={splitMessage(commit.message).title ?? ''}
-							sha={commit.id}
-							date={new Date(commit.createdAt)}
-							author={commit.author.name}
-							url={commitUrl}
-							onOpen={(url) => openExternalUrl(url)}
-							onCopy={() => writeClipboard(commit.id)}
-						/>
-					{/each}
-				</ScrollableContainer>
-			</div>
+			{#if upstreamCommits}
+				<div class="scroll-wrap">
+					<ScrollableContainer maxHeight="16.5rem">
+						{#each upstreamCommits as commit}
+							{@const commitUrl = forge.current.commitUrl(commit.id)}
+							<SimpleCommitRow
+								title={splitMessage(commit.message).title ?? ''}
+								sha={commit.id}
+								date={new Date(commit.createdAt)}
+								author={commit.author.name}
+								url={commitUrl}
+								onOpen={(url) => openExternalUrl(url)}
+								onCopy={() => writeClipboard(commit.id)}
+							/>
+						{/each}
+					</ScrollableContainer>
+				</div>
+			{/if}
 
 			{#snippet controls(close)}
 				<div class="controls">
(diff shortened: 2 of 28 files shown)


crates/gitbutler-git/src/repository.rs
```diff
@@ -368,6 +368,7 @@ where
     }
 
     if force && force_if_includes {
+        args.push("--force-with-lease");
```
krlvi: Hmm, this should be added above, on line 367

peaklabs-dev: Ah yes, then I have no clue why it's not working. It still always overwrites my remote commits when I force push. Maybe the --force-if-includes flag does not work the way I think it would 🤔.

krlvi: What happens if you perform the test with manually pushing the git reference from the CLI?

peaklabs-dev: I just did a lot of testing, and it works correctly in the terminal but not in GitButler.

peaklabs-dev: @krlvi I found the issue and I will push a fix soon.

crates/gitbutler-repo-actions/src/repository.rs
```diff
@@ -160,16 +159,11 @@ impl RepoActionsExt for CommandContext {
...
         refspec: Option<String>,
         askpass_broker: Option<Option<StackId>>,
     ) -> Result<()> {
         let refspec = refspec.unwrap_or_else(|| {
-            if with_force {
-                format!("+{}:refs/heads/{}", head, branch.branch())
-            } else {
-                format!("{}:refs/heads/{}", head, branch.branch())
-            }
+            format!("{}:refs/heads/{}", head, branch.branch()) // for force pushing we previously had "+{}:refs/heads/{}" which was removed because it bypasses the force push protection flags as it is equivalent to --force
```
Byron, line 168: I see, this is the key here.
It also seems like a bug in Git to not error *or* respect the flags, but that could be brought up with them separately.

crates/gitbutler-git/src/repository.rs
```diff
@@ -363,11 +363,12 @@ where
     args.push(remote);
     args.push(&refspec);
 
-    if force {
-        args.push("--force-with-lease");
+    if force && !force_push_protection {
```
Byron: I certainly wouldn't mind to see this if condition cleaned up to not be so redundant and error prone anymore.

```rust
if force {
    if force_push_protection {}
}

**Comment by Byron** [comment]

> * **Limitation:** Currently when a force push is prevented through the git flags, the `Integrate Upstream` (yellow) button will not work anymore. As @Byron is refactoring that part, this should not be an issue for too long I hope 😀 (or maybe it is an easy fix??).

Actually I would appreciate if you could look into this to see *if* there is an easy fix, but that's after this PR.

Even though I didn't see the usage of `--force-if-includes` in the diff, I also didn't expand it. Maybe that will change if the argument logic is cleaned up a bit.

It's notable that I didn't run this yet, and I think this PR needs another set of eyes, @krlvi.

PS: When looking at the CI failure, I recommend running `cargo check --all-targets` more often, and probably `cargo clippy --all-targets` as well (or instead of `cargo check`).

File: crates/gitbutler-git/src/repository.rs (hunk 3 above)
Comment (peaklabs-dev) [reply crates/gitbutler-git/src/repository.rs]: Thanks. Done.

File: crates/gitbutler-repo-actions/src/repository.rs (hunk 2 above)
Comment (peaklabs-dev) [reply crates/gitbutler-repo-actions/src/repository.rs:168]: Yes, I also think it should be classified as a bug in Git because, in my opinion, adding the flags `--force --force-with-lease --force-with-includes` should apply them to the force push. However, Git seems to think differently because it does not use them as protection flags, it uses them as a replacement to `--force`. Therefore, you can use either one to perform a safer force push or even use both for maximum security when doing the force push, but not in combination with `--force` (or a `+` for that matter).

**Comment by Byron** [comment]
Thanks a lot!

Maybe you could add a todo-list so it's more clear what's missing for this draft to become eligible for review and merging. Thank you.

**Comment by peaklabs-dev** [comment]
The following things need to be done for this to be ready:

- [x] Show a modal with the commits that would be overwritten.
- [x]  Add a way to catch the error and display the commit difference in the modal.
- ~~Search for a fix for integrate upstream.~~

**Comment by Byron** [comment]
Thank you. I thought that with this fix to how `--force-*` flags are used they will be effective. Is this not by itself valuable and worth merging?

**Comment by PavelLaptev** [comment]
@peaklabs-dev @krlvi @Byron I have a few thoughts:
- Different projects need different approaches, so this should be under the **project settings**.
- It doesn’t sound like an experimental feature, it’s more of a protection level. So this probably should be here `Project settings / Project`

**Comment by peaklabs-dev** [comment]
@PavelLaptev Yes, I agree. I thought that earlier today as well. What about putting it under `Git Stuff` below `Allow force pushing` or grouping it with `Allow force pushing`?

**Comment by PavelLaptev** [comment]
> @PavelLaptev Yes, I agree. I thought that earlier today as well. What about putting it under `Git Stuff` below `Allow force pushing` or grouping it with `Allow force pushing`?

Sounds good! Yes, let's actually connect it into a single block like this:

image

**Comment by peaklabs-dev** [comment]
@PavelLaptev Done moved the setting as requested.

@Byron I tried to fix the yellow button, but I don't think it's that easy. The rebasing and the entire logic needs improvement, and I'm not yet completely familiar with all parts of the codebase so I am sure if I will be able to fix it.

**Comment by PavelLaptev** [comment]
> @PavelLaptev Done moved the setting as requested.

Thanks! I can implement the modal. When exactly should it appear? If the flag is on?

**Comment by Byron** [comment]
> @Byron I tried to fix the yellow button, but I don't think it's that easy. The rebasing and the entire logic needs improvement, and I'm not yet completely familiar with all parts of the codebase so I am sure if I will be able to fix it.

No worries, I think it makes sense to minimise the scope of this PR so that we can merge what works.

**Comment by peaklabs-dev** [comment]
@PavelLaptev When the flag is active and it throws an error (code 1 I believe). We will need to add some error handling for the code and figure out how to get the commits somewhere around here probably https://github.com/peaklabs-dev/gitbutler/blob/1af31e5f22f4e7fc9342d4e15bcbfcb407308f8c/crates/gitbutler-git/src/repository.rs#L378

**Comment by PavelLaptev** [comment]
> @PavelLaptev When the flag is active and it throws an error (code 1 I believe). 

@peaklabs-dev I just realized I’m not sure how to get the difference between commits. So I’ll use a dummy array for now that you can replace later. Is that okay?  
Or do you want to create the modal, and I’ll just style it after?

**Comment by peaklabs-dev** [comment]
@PavelLaptev dummy array is fine. I need to figure how to get the commits as well 😅.

**Comment by PavelLaptev** [comment]
@peaklabs-dev I added the modal.  
@Byron @peaklabs-dev feel free to tweak any of my code related to data consumption or similar.  
I used the same design for the commit list as we did for `IntegrateUpstreamModal`.

image


The "Don't show again" checkbox isn’t needed here, since the user enabled the "force push protection" flag. The modal should keep appearing until they turn it off.

**Comment by peaklabs-dev** [comment]
@PavelLaptev Thank you so much for the UI help! The only issue I'm not sure how to address is that sometimes I need to reload the entire view to get the commits/ updated commits into the modal.

@Byron and @krlvi Thanks for all the help. This PR is now ready for review and then hopefully also merge.

**Comment by PavelLaptev** [comment]
>  sometimes I need to reload the entire view to get the commits/ updated commits into the modal.

Yeah, probably something wrong with reactivity. Not sure. When I toggle the flag, the app doesn’t react to it (see video), and I have to reload the app manually to see the update. @Byron do you know who might know which direction to dig here?


https://github.com/user-attachments/assets/2ca78149-3914-4c07-b954-3adfb35e0818


@estib-vega unrelated, but I ran into this API error with rules:
image

**Comment by peaklabs-dev** [comment]
Sorry to tag you again @Byron but I think I may have uncovered a deeper issue with the `integrate upstream` (yellow button) feature. Now, when I fetch and try to integrate without doing a force push with protection on --> the protection is on but I did not try to use the force push button first, I pulled and then hit `integrate upstream` it breaks with the following error:
```
command: integrate_upstream_commits
params: {"projectId":"XX","stackId":"XX","seriesName":"next"})

Picked commit already exists in a previous step
```

**Comment by Byron** [comment]
I just rebased the PR on top of `master` in the hopes to get rid of the rule-related API errors. But this is minor.

Related to the other issues:

* I don't know a thing about reactivity or how to make it work, unfortunately.
* `Picked commit already exists in a previous step` is an error when trying to rebase something, but the list of commits that was generated didn't pass basic validation. To be honest, I am surprised it works at all, sometimes. Maybe this can be fixed in the current implementation.
    - My real question here is if this issue if related to the changes here, or was there before. **My guess is that it's unrelated.**

I hope this helps. (*Please note that I didn't try this PR myself, just looked at the diff*)

**Comment by PavelLaptev** [comment]
@estib-vega might be you know why it could work like this https://github.com/gitbutlerapp/gitbutler/pull/9710#issuecomment-3165684168?

File: apps/desktop/src/components/GitForm.svelte [hunk 4]
```diff
@@ -28,22 +32,40 @@
...
-				{/snippet}
-				{#snippet actions()}
-					<Toggle
-						id="allowForcePush"
-						checked={project.ok_with_force_push}
-						onchange={(checked) => onForcePushClick(project, checked)}
-					/>
-				{/snippet}
-			</SectionCard>
+			<div class="stack-v">
```
mtsgrd, line 35: let's remove the class since it's not used for anything

apps/desktop/src/components/PushButton.svelte
```diff
@@ -78,34 +109,40 @@
...
-			tooltip={getButtonTooltip(hasThingsToPush, hasConflicts)}
-			onclick={() => handleClick(requiresForce)}
-			icon={multipleBranches && !isLastBranchInStack ? 'push-below' : 'push'}
-		>
-			{requiresForce ? 'Force push' : 'Push'}
-		</Button>
+
+		<ReduxResult {projectId} result={projectResult.current}>
+			{#snippet children(project)}
+				<Button
```
mtsgrd: Instead of two redux result components, could you combine the two results using `combineResults`? There should be plenty of examples in the code base.

crates/gitbutler-git/src/repository.rs
```diff
@@ -364,40 +364,50 @@ where
...
+                }))?
+            } else {
+                Err(Error::<E>::Failed {
+                    status,
+                    args: args.into_iter().map(Into::into).collect(),
+                    stdout,
+                    stderr,
+                })?
+            }
         }
```
mtsgrd: This section feels more difficult to read now with a combination of match, and two if blocks. Could you try and simplify it a bit?

crates/gitbutler-repo-actions/src/repository.rs
```diff
@@ -66,13 +66,27 @@ impl RepoActionsExt for CommandContext {
         let refname =
             RemoteRefname::from_str(&format!("refs/remotes/{remote_name}/{branch_name}",))?;
 
-        match self.push(commit_id, &refname, false, false, None, askpass) {
+        match self.push(
+            commit_id,
+            &refname,
+            false,
+            self.project().force_push_protection,
```
mtsgrd: Can you declare a variable for this since it's used more than once?

apps/desktop/src/components/PushButton.svelte
```diff
@@ -32,32 +44,51 @@
...
 	const uiState = inject(UI_STATE);
+	const forge = inject(DEFAULT_FORGE_FACTORY);
 	const branchDetails = $derived(stackService.branchDetails(projectId, stackId, branchName));
+	const projectResult = $derived(projectsService.getProject(projectId));
 	const [pushStack, pushResult] = stackService.pushStack;
+	const potentialOverwrittenCommits = $derived(
+		stackService.upstreamCommits(projectId, stackId, branchName).current.data || []
+	);
+
+	let _currentProject = $state<any>(undefined);
```
mtsgrd: Can you help me understand the purpose of this variable?

crates/gitbutler-branch-actions/src/stack.rs
```diff
@@ -158,6 +158,7 @@ pub fn push_stack(
     ctx: &CommandContext,
     stack_id: StackId,
     with_force: bool,
+    force_push_protection: bool,
```
mtsgrd: Is this value ever different from `ctx.project().force_push_protection`? I might be missing something, but if it isn't we don't need to have it sent from the front end.

mtsgrd (commented):
Thanks for all the hard work. 💪 I left a few comments if you wouldn't mind having a look?

crates/gitbutler-branch-actions/src/stack.rs
peaklabs-dev: Yes, it can be different when users wants to force push even with protection (selectable in the modal).

apps/desktop/src/components/PushButton.svelte
peaklabs-dev: Not sure, this was added by @PavelLaptev here [`66e2171` (#9710)](https://github.com/gitbutlerapp/gitbutler/pull/9710/commits/66e2171880ec9825698be48954c4374618077f53)

apps/desktop/src/components/PushButton.svelte
peaklabs-dev: This is needed on the `onSubmit` of the `Push with dependencies` modal and in the `handleClick` function

crates/gitbutler-branch-actions/src/stack.rs
mtsgrd: Right, ofc!

mtsgrd: I think we need to change this to something like `ignore_force_push_protection`, and get the value for `force_push_protection` directly from the application context in the back end. That way it's clear that it is asking to ignore such a setting, reducing the likelihood of future errors. Do you see where I'm coming from?

apps/desktop/src/components/PushButton.svelte
mtsgrd: I understand. Would you still be able to make this change?

crates/gitbutler-branch-actions/src/stack.rs
peaklabs-dev: I see where you're coming from but I feel like that makes it more complex. We can skip force push protection by disabling this flag which seems pretty straightforward to me.

apps/desktop/src/components/PushButton.svelte
mtsgrd: Right, I see. If you remove the second `ReduxResult` and use `combineResults` for the first one, then you'll have `project` in scope for the modal click. We can then pass just the variable `forcePushProtection` to the `push` function (and remove `_currentProject`).

crates/gitbutler-branch-actions/src/stack.rs
mtsgrd: Right, but I need to insist we separate the override from the actual setting. It invites confusion otherwise, as evidenced by my experience reviewing this pr. Does that work for you? *edit: if you want I can accept this as is, but add these changes myself.

apps/desktop/src/components/PushButton.svelte
peaklabs-dev: Done

apps/desktop/src/components/PushButton.svelte
peaklabs-dev: Done

crates/gitbutler-branch-actions/src/stack.rs
peaklabs-dev: Thanks. It would be great if you could add the changes.

mtsgrd:
@peaklabs-dev could you have a look at what I force pushed to this branch and let me know if that looks good to you?

crates/gitbutler-branch-actions/src/stack.rs
```diff
@@ -186,6 +187,9 @@ pub fn push_stack(
         remote: default_target.push_remote_name(),
         branch_to_remote: vec![],
     };
+
+    let force_push_protection = !skip_force_push_protection && ctx.project().force_push_protection;
```
mtsgrd, line 191: @peaklabs-dev here it either skips, or uses the value from project settings

crates/gitbutler-repo-actions/src/repository.rs
mtsgrd: I don't think we actually need to set this here, since it's a test pushing to a test branch.

mtsgrd (approved):
@Byron I've made some changes. Are you ok with landing this?

crates/gitbutler-branch-actions/src/stack.rs
peaklabs-dev: Sounds good.

peaklabs-dev:
@mtsgrd Everything looks good to me

mtsgrd:
Also, I address the project settings reactivity issue here: https://github.com/gitbutlerapp/gitbutler/pull/9784

PavelLaptev (approved):

PR #9716, opened 2025-08-05
fix: project settings menu

peaklabs-dev:
## 🧢 Changes

- fix: Project Settings menu (in the menu bar) currently points to a different Settings page (with horizontal tabs) instead of the v3 one.

### DIFF
--- a/apps/desktop/src/components/ProjectSettingsMenuAction.svelte
+++ b/apps/desktop/src/components/ProjectSettingsMenuAction.svelte
@@ -2,11 +2,7 @@
 	import { goto } from '$app/navigation';
 	import { vscodePath } from '$lib/project/project';
 	import { PROJECTS_SERVICE } from '$lib/project/projectsService';
-	import {
-		isNewProjectSettingsPath,
-		newProjectSettingsPath,
-		workspacePath
-	} from '$lib/routes/routes.svelte';
+	import { newProjectSettingsPath } from '$lib/routes/routes.svelte';
 	import { historyPath } from '$lib/routes/routes.svelte';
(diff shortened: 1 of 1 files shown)


apps/desktop/src/components/ProjectSettingsMenuAction.svelte
```diff
@@ -20,7 +24,11 @@
 	$effect(() =>
 		mergeUnlisten(
 			shortcutService.on('project-settings', () => {
-				goto(projectSettingsPath(projectId));
+				if (isNewProjectSettingsPath()) {
+					goto(workspacePath(projectId));
+				} else {
+					goto(newProjectSettingsPath(projectId));
+				}
```
estib-vega: Let's just keep the navigation to the new project settings path, please.

It's fine for the gear button to have that, but the menu shouldn't do the toggle-ing 👍 

```ts
() => {
    goto(newProjectSettingsPath(projectId));
}
```

estib-vega (commented):
This is an great catch, much appreciated!
Just that tiny comment and then we can merge it

apps/desktop/src/components/ProjectSettingsMenuAction.svelte
peaklabs-dev: Thanks. Done. Why does the gear icon need both and the menu not?

estib-vega: The menu should always navigate since it's ephemeral (when you click the menu option, the menu closes)

The button has that extra functionality because the button is still visible, even if you're in the project settings page.
In this case, the ability to use the same button to toggle is more of an ergonomic hidden feature, rather than an actual requirement

PR #9719, opened 2025-08-05
feat: add Default terminal setting & Open in terminal menu

peaklabs-dev:
## ToDo & Limitations

- **Limitation: Currently, this change only works for macOS because the app names are specific to macOS.** I am not 100% sure how we would make this change cross platform as app names could be different on windows.
- ToDo: Instead of using links to open all the code editors (like now), we could refactor that part of the code to open via app names as well using one clean function. This would be helpful because once every default terminal and code editor is opened via the app name, we could allow custom apps we do not have in the list as well. To add a custom app the users could select an app from the app folder, which would be very useful because we would not have to add every editor and the user can just select one. Good idea??

## 🧢 Changes

- feat(ui): add default terminal settings to the `General` settings page
- feat(core): add `Open in Terminal` to project menu which opens the current project in the default terminal

## Preview

**Settings**

image

**Menu**

image

### DIFF
--- a/crates/but-api/src/commands/open.rs
+++ b/crates/but-api/src/commands/open.rs
@@ -150,3 +150,24 @@ pub fn show_in_finder(_app: &App, params: ShowInFinderParams) -> Result<(), Erro
 
     Ok(())
 }
+
+#[derive(Deserialize)]
+pub struct OpenInTerminalParams {
+    pub app_name: String,
+    pub path: String,
+}
+
+pub fn open_in_terminal(_app: &App, params: OpenInTerminalParams) -> Result<(), Error> {
+    #[cfg(target_os = "macos")]
+    {
+        use std::process::Command;
+        Command::new("open")
+            .arg("-a")
+            .arg(&params.app_name)
+            .arg(&params.path)
+            .status()
+            .with_context(|| format!("Failed to show '{}' in Finder", params.path))?;
+    }
+
+    Ok(())
+}
--- a/apps/desktop/src/components/ProjectSettingsMenuAction.svelte
+++ b/apps/desktop/src/components/ProjectSettingsMenuAction.svelte
@@ -6,7 +6,7 @@
 	import { historyPath } from '$lib/routes/routes.svelte';
 	import { SETTINGS } from '$lib/settings/userSettings';
 	import { SHORTCUT_SERVICE } from '$lib/shortcuts/shortcutService';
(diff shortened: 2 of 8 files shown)


PavelLaptev:
@peaklabs-dev thank you for the PR :-)  
I made a small commit. I moved "default terminal" and "default editor" into a single group because they both relate to default tools.

image

PavelLaptev:
@peaklabs-dev Is this still something we want to merge right?

peaklabs-dev:
@PavelLaptev Yes, but I'm not sure how to implement this for Linux and Windows. Also, should we switch to opening the app for code editors as well instead of links, to allow custom app selection?

crates/but-api/src/commands/open.rs
```diff
@@ -150,3 +150,24 @@ pub fn show_in_finder(_app: &App, params: ShowInFinderParams) -> Result<(), Erro
...
+pub struct OpenInTerminalParams {
+    pub app_name: String,
+    pub path: String,
+}
+
+pub fn open_in_terminal(_app: &App, params: OpenInTerminalParams) -> Result<(), Error> {
+    #[cfg(target_os = "macos")]
+    {
+        use std::process::Command;
+        Command::new("open")
```
Byron, line 164: If `open::with_command(…)` would be used instead, I think this would be portable provided viable app-names. The `.app` suffix on MacOS also wouldn't be needed then, I think

Generally I think it would be OK to start supporting MacOS and add Windows/Linux later. This probably would be gated where the available app-names are configured.

PavelLaptev:
> @PavelLaptev Yes, but I'm not sure how to implement this for Linux and Windows.

I see. Yes, it would be nice to handle it for all platforms.

> should we switch to opening the app for code editors as well instead of links, to allow custom app selection?

What do you mean?

sebastianhuus:
@peaklabs-dev @PavelLaptev What's the status here? Looks like it went stale. 

Looks like the entire app has had major file structure changes so it will be difficult to refactor this branch into something that will pass checks. I think it's easier to rewrite the implementation in a new PR. 

I'm interested in trying this and can also try implementing cross platform support. Thoughts?

Byron:
Thanks for bringing this to our attention @sebastianhuus!
I completely agree that a great first step is to close this issue due to staleness (and probably many more).

peaklabs-dev:
@sebastianhuus Yes it would be great if you could finish this. I am currently busy with other work, so closing this out is ok.

PR #9735, opened 2025-08-07
Improve handling of stacked remote commits

### LINKED ISSUE #9733: squash-merged commits aren't detected when remote commits are involved
That makes sense, this check isn't performed there, but it should be.
Here is a simple example, but these can probably be more complex, too.

> So I've tested the most basic case:
> 
> Before: squash-merged first branch in github but not updated the workspace
> Image
> 
> After: tried to update the workspace
> Image
> 
> what's good: first branch is gone, 
> bad: don't know what to do with these upstream commits that are actually incorporated
> 
>  

 _Originally posted by @MFattakhov in [#8457](https://github.com/gitbutlerapp/gitbutler/issues/8457#issuecomment-3160711037)_

Byron:
Fixes #9733.

### Tasks

* [x] manual reproduction and exploration of fix
* [x] test
* [x] fix
    - find a way to pick create segments for remote branches only when seen from another remote branch
    - maybe this can be done as a post-process, so all we need is remote branches to be seen on the commits in question.

### DIFF
--- a/crates/but-graph/src/init/mod.rs
+++ b/crates/but-graph/src/init/mod.rs
@@ -265,7 +265,17 @@ impl Graph {
         let configured_remote_tracking_branches =
             remotes::configured_remote_tracking_branches(repo)?;
         let refs_by_id = repo.collect_ref_mapping_by_prefix(
-            std::iter::once("refs/heads/").chain(if collect_tags {
+            [
+                "refs/heads/",
+                // Remote refs are special as we collect them into commits to know about them,
+                // just to later remove them unless they are on an actual remote commit.
+                // In that case, we also split the segment there if the previous segment then wouldn't be empty.
+                // Naturally we only pick them up and segment them if they are added by the local tracking branch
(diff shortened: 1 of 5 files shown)


Byron:
### Investigation

After squashing, we see this if the branch above wasn't pushed as well.

Screenshot 2025-08-07 at 10 31 48

Now the UI wants to integrate, even though what we really want is to force-push `b2`.

Screenshot 2025-08-07 at 10 32 25

After that, everything is back in order.

Screenshot 2025-08-07 at 10 35 17

Screenshot 2025-08-07 at 10 35 41


**All the above was only to the point where we match the actual issue**, so plenty of traps to deal with when squashing.

After the workspace update, we have this:

Screenshot 2025-08-07 at 10 44 39

Screenshot 2025-08-07 at 10 42 51

Which looks like this in the UI as we currently collect all the remote segments, instead of just one. So probably we just want to stop after one to avoid this issue, which is possible as long as the remote branches are properly pushed.

PR #9746, opened 2025-08-07
fix: singular and plural wording

peaklabs-dev:
## 🧢 Changes

- fix: singular and plural wording. When there is 1 commit or 1 change, it should use the singular word: `commit` and `change` and not `commits` or `changes`.


Byron (approved):
Thanks a lot, this looks good to me!

PR #9749, opened 2025-08-07
Undo the filtering logic for usptream commits v2

estib-vega:
- Remove unused import `but_core::diff::tree_changes` from stacks.rs
- Adjust commit filtering to push upstream commits directly when not matched locallyr


estib-vega:
cc @krlvi 
I have no reason to believe it's this but it's the first thing that I find that changed in the stack_details in between release 0.15.7 and 0.15.8

PR #9754, opened 2025-08-08
Add `create_reference()` to API to allow putting refs anywhere.

Byron:
### Tasks

* [x] endpoint in backend
* [x] simplify/flatten types
* [x] ~~types in the frontend at least (or some lifesign that it can work?)~~ I don't think I can do this 😅, definitely worth to stop here and align on the types first.

### Notes for the Reviewer

* Are the nested enums easy enough to handle? The benefit is that only valid operations are representable, but a flat simple structure could also do the trick and the backend can just error if something isn't representable.
    - I decided to flatten the structure to keep typing in the FE simple.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -880,6 +880,7 @@ dependencies = [
  "gitbutler-sync",
  "gitbutler-user",
  "gix",
+ "insta",
  "open",
  "reqwest 0.12.22",
  "serde",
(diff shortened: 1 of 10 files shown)


Byron:
Thanks @estib-vega for the feedback! I am totally looking forward to seeing this used in the UI :).

PR #9755, opened 2025-08-08
MainViewport: remove AsyncRender wrapper

estib-vega:
Remove  the AsyncRender wrapper around the middle slot rendering. 

This *seem* to fix a condition in which the stacks would sometimes not render correctly


krlvi:
fixes https://github.com/gitbutlerapp/gitbutler/issues/9740

PR #9762, opened 2025-08-10
fix: delete button on empty branches & deletion error

### LINKED ISSUE #9662: [BUG] Empty (virtual) branches cannot be deleted
### Version

0.15.8

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

Create a new independent branch and try to delete it immediately after creating it. There is no `delete` option.

_No response_

I think there should be a delete option, like on any other local/ remote branch.

Image

```shell

```

peaklabs-dev:
## 🧢 Changes

- fix(ui): show `Delete` button on empty branches (no commits) as they can be deleted
- fix: deletion should invalidate stacks and stack details. This fixes the following error that always appeared on branch deletion:
```
command: stack_details 
params: {"'projectId": "XX", "stackId": "XX" }) 
Couldn't find XX even when looking at virtual_branches.toml directly
```

## 🎫 Affected issues

Fixes: https://github.com/gitbutlerapp/gitbutler/issues/9662

## 📌 Todos

- [x] Can we improve the clarity of the following error message? 
A branch (stacked or independent) cannot be deleted when there is 1 or more unpushed commits on the branch. The following error message will appear:
```
command: remove_branch
params: {"projectId":"XX","stackId":"XX","branchName":"p-branch-1"})

Refusing to delete last named segment 'p-branch-1' as it would leave an anonymous segment
```


Byron:
Thanks so much for improving branch deletion!

> Refusing to delete last named segment…

This means it can't delete stacks that have commits in them. It can, however, delete stacks that don't. The UI would have to check for this to avoid such errors.

In future, I believe it will be possible to delete an applied stack with commits which would be similar to `unapply + delete`.

peaklabs-dev:
@Byron Just adjust the error message slightly to make it more understandable. I would love to see a `unapply + delete` feature.

Byron:
> Just adjust the error message slightly to make it more understandable.

I note that it's not what a user would see as the UI would avoid such calls. The message itself is on point, but understanding it needs some understanding of more internal and recently added terminology.

peaklabs-dev:
The error occurs for me, which is why I wanted to adjust the message.

image

crates/but-workspace/src/branch/remove_reference.rs
```diff
@@ -56,7 +56,7 @@ pub(crate) mod function {
                     < 2)
         {
             bail!(
-                "Refusing to delete last named segment '{}' as it would leave an anonymous segment",
+                "Cannot delete the last named segment '{}' because the stack has unpushed commits. Deleting it would leave an anonymous segment.",
```
Byron:
```suggestion
                "Cannot delete the last named segment '{}'. Deleting it would leave an anonymous segment.",
```

It's not correct to say this is related to unpushed commits.

Byron (changes requested):
@peaklabs-dev I think what's missing here is that the UI won't try to delete the last stack segment. It's not possible right now to do that, hence the error message, but it's internal and users shouldn't have to see it.

Ideally, the "Delete" menu item is shown *only* if there are more than one segments (`stackLength` here) OR if there is a single stack-segment `stackLength==1` without any commits.

Thanks again for going the extra mile with me.

Byron (approved):
Works great, thanks so much!

PR #9773, opened 2025-08-11
Add auto opt-in for WS3 feature flag

estib-vega:
Auto-opt-in the users into WS3, allowing them to go back after, of course


estib-vega:
cc @krlvi

PR #9776, opened 2025-08-11
Update chat model to gpt-5-mini

estib-vega:
Changed the AI model used for the generation of commit messages and branch names requests in crates/but-action/src/generate.rs from "gpt-4o" to "gpt-5-mini".


estib-vega:
cc @krlvi 
Use smaller-faster-newer models for the hook generations

PR #9777, opened 2025-08-11
New CC types and persistence

krlvi:


crates/but-db/migrations/2025-08-11-130145_claude_sessions_and_messages/down.sql
```diff
@@ -0,0 +1,7 @@
+-- This file should undo anything in `up.sql`
+DROP TABLE IF EXISTS `claude_sessions`;
+DROP INDEX index_claude_sessions_on_current_id;
```
Caleb-T-Owens: FWIW, dropping the table will automatically drop the indexes

krlvi: fixed, thanks

PR #9779, opened 2025-08-11
Create ref in stack

estib-vega:
Ability to create a anywhere in the stack.

### DIFF
--- a/apps/desktop/src/components/BranchCommitList.svelte
+++ b/apps/desktop/src/components/BranchCommitList.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import CardOverlay from '$components/CardOverlay.svelte';
 	import CommitAction from '$components/CommitAction.svelte';
-	import CommitContextMenu, { type CommitMenuContext } from '$components/CommitContextMenu.svelte';
+	import CommitContextMenu from '$components/CommitContextMenu.svelte';
 	import CommitGoesHere from '$components/CommitGoesHere.svelte';
 	import CommitRow from '$components/CommitRow.svelte';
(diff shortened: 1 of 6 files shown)


estib-vega:
cc @Byron :)

Byron:
Hi @estib-vega! Loving this feature.

Two quick notes:

* The backend uses a different terminology because it has to (it does new stacks as well as dependent branches in one function), which in the end, is just a ref. However, I don't think we should introduce this term in the frontend necessarily and rather keep calling it dependent branch (or similar). Maybe this was already aligned though, and I'd retract my comment :D.

https://github.com/user-attachments/assets/e769bfa7-870d-49e6-b5e5-4fc8e5dde95a

**The UI would have to avoid showing the 'below' option when there are commits already.**

https://github.com/user-attachments/assets/cefc8405-265c-4aec-821d-6de22953a167

Empty stacks will work though - they must be fully empty though.

estib-vega:
Hey @Byron, thanks for the feedback!

> Hi @estib-vega! Loving this feature.
> 
> Two quick notes:
> 
> * The backend uses a different terminology because it has to (it does new stacks as well as dependent branches in one function), which in the end, is just a ref. However, I don't think we should introduce this term in the frontend necessarily and rather keep calling it dependent branch (or similar). Maybe this was already aligned though, and I'd retract my comment :D.
> 

Makes total sense 👍 

>  Screen.Recording.2025-08-11.at.16.43.41.mov 
> **The UI would have to avoid showing the 'below' option when there are commits already.**
> 
>  Screen.Recording.2025-08-11.at.16.46.38.mov 
> Empty stacks will work though - they must be fully empty though.

Right, forgot about that

PR #9781, opened 2025-08-11
Make Claude Code hooks use the rules structure for session<>stack associations

krlvi:


apps/desktop/src/components/Rule.svelte
```diff
@@ -85,6 +85,10 @@
 		{:else if filter.type === 'semanticType'}
 			<Icon name="tag" opacity={0.6} />
 			<span class="text-12 truncate">{semanticTypeToString(filter.subject.type)}</span>
+		{:else if filter.type === 'claudeCodeSessionId'}
+			<Icon name="tag" opacity={0.6} />
+			<!-- TODO: Make an API call to get a rich name for the session, instead of rendering a UUID -->
```
krlvi, line 90: FYI @estib-vega

krlvi: API doesnt exist yet. But it will exist soon

PR #9791, opened 2025-08-12
rules-realtime-notify-filters

estib-vega:
- Backend DB watcher and Tauri integration now emit rule-updates events for Rules item kind
- Frontend listens to rule-updates events and invalidates cache on rules DB changes
- More robust handling of rule filter types, including special pill/action behavior for session UUID (claudeCodeSessionId)
- Improved UI/UX for rule filter entry, display, and operations


estib-vega:
cc @krlvi 
This updates the FE on claude rules and makes the claude rules session selector actually a static label.

PR #9800, opened 2025-08-12
Set rule trigger to null in RulesList

estib-vega:
Leave the trigger alone when updating


estib-vega:
cc @krlvi
- Don't update the trigger unnecessarily

PR #9801, opened 2025-08-12
Emit hunk assignment updates as distinct event

estib-vega:
Make the hunk assignment database updates emit their own event. Update the backend to emit 'hunk-assignment-update' instead of bundling with other DB updates, and update the frontend to listen for and handle these events so stacks can update correctly.


estib-vega:
cc @krlvi 
- Update the stacks if they are created by the rules (invalidate the stacks on hunk assignment changes)

PR #9823, opened 2025-08-13
A quick-fix to let the frontend observe if a stack-segmetn is checked out.

Byron:
That way it can decide what to do for 'solo' mode.

The idea here is to add flags to achieve certain goals, without providing everything the information needs.

### DIFF
--- a/crates/but-action/src/lib.rs
+++ b/crates/but-action/src/lib.rs
@@ -228,7 +228,7 @@ fn stacks(ctx: &CommandContext, repo: &gix::Repository) -> anyhow::Result<Vec<St
     let project = ctx.project();
     if ctx.app_settings().feature_flags.ws3 {
         let meta = ref_metadata_toml(ctx.project())?;
-        but_workspace::stacks_v3(repo, &meta, but_workspace::StacksFilter::InWorkspace)
+        but_workspace::stacks_v3(repo, &meta, but_workspace::StacksFilter::InWorkspace, None)
     } else {
         but_workspace::stacks(
             ctx,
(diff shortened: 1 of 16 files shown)


Byron:
@mtsgrd With this PR merged, you can apply this patch locally for 'isolation' mode. Then if a stack or stack-segment is checked out, the UI will only see this, not the surrounding ones:

```patch
diff --git a/crates/but-workspace/src/stacks.rs b/crates/but-workspace/src/stacks.rs
index 2ea49cec99..9913d44960 100644
--- a/crates/but-workspace/src/stacks.rs
+++ b/crates/but-workspace/src/stacks.rs
@@ -229,7 +229,7 @@ pub fn stacks_v3(
     }
 
     let unapplied_stacks = unapplied_stacks(repo, meta, &info.stacks)?;
-    Ok(match filter {
+    let out = match filter {
         StacksFilter::InWorkspace => into_ui_stacks(repo, info.stacks, meta),
         StacksFilter::All => {
             let mut all_stacks = unapplied_stacks;
@@ -237,6 +237,11 @@ pub fn stacks_v3(
             all_stacks
         }
         StacksFilter::Unapplied => unapplied_stacks,
+    };
+    Ok(if out.iter().any(|s| s.is_checked_out) {
+        out.into_iter().filter(|s| s.is_checked_out).collect()
+    } else {
+        out
     })
 }
 

```

With the new flag this kind of filtering could also happen in the frontend, which might be more convenient even.

mtsgrd:
Awesome, thank you!

PR #9824, opened 2025-08-13
Namespace rule analytics props and add helper

estib-vega:
- Renamed returned metrics object to ruleMetrics instead of returning directly.
- Added namespaceProps helper function that takes EventProperties and a namespace string and returns a new object with keys prefixed by `${namespace}:`.
- Updated function to return namespaceProps(ruleMetrics, 'rule') so all rule-related analytics properties are namespaced (e.g., 'rule:totalWorkspaceRules').


estib-vega:
cc @krlvi
Rule metrics are namespaced:  `rule:`

PR #9825, opened 2025-08-13
fix: change rule pill text from "All files matched" to "All changes"

krlvi:


krlvi:
@PavelLaptev did minor copy changes (I changed the suggestion that I made earlier)

PR #9844, opened 2025-08-14
Workspace rules for Claude sessions will use the original session ID if available

krlvi:


crates/but-claude/src/hooks/mod.rs
```diff
@@ -404,6 +407,16 @@ pub fn handle_post_tool_call() -> anyhow::Result<ClaudeHookOutput> {
...
 
+fn original_session_id(ctx: &mut CommandContext, current_id: String) -> Result<String> {
+    let original_session_id =
+        crate::db::get_session_by_current_id(ctx, Uuid::parse_str(&current_id)?)?;
+    if let Some(session) = original_session_id {
+        Ok(session.id.to_string())
+    } else {
+        Ok(current_id)
+    }
+}
```
Caleb-T-Owens, line 418: should this live in the but-claude/db.rs and use the functions inside db.rs so we have the types mapped for us?

krlvi: Great idea, lemme do that real quick

krlvi: Actually no - this is specific handling for the Claude Hooks Input types

PR #9850, opened 2025-08-14
Add nudge for rules feature flag

estib-vega:
Adds a new autoOptInRules method and a nudge routine to the SettingsService. This ensures the rules feature flag is toggled on for users who have not already engaged with it, storing completion state. Mirrors existing logic for ws3AutoToggle, so users are prompted to try the rules feature automatically. Updates constructor logic accordingly.


estib-vega:
cc @krlvi 
Nudge people into ruling their workspace with an iron fist

PR #9857, opened 2025-08-14
ui-fixes-123

PavelLaptev:
- Add missing "unassigned" mode to UnassignedView
- Show Select All checkbox only when with changes (work)
- Add prop to AsyncButton apply spacing to CommitView buttons
- Simpl AsyncButton props and loading flow for clarity
- Remove dependent branch options from commit context menu
- Remove unused AddDependentBranchModal imports and state
- Show generated slug helper when branch name differs
- Remove double bottom border from SelectionView

-- some screenshots (not all) --

image
image
image
image

### DIFF
--- a/apps/desktop/src/components/CommitContextMenu.svelte
+++ b/apps/desktop/src/components/CommitContextMenu.svelte
@@ -55,7 +55,6 @@
 		KebabButton,
 		TestId
 	} from '@gitbutler/ui';
-	import type { AnchorPosition } from '$lib/stacks/stack';
 
 	type Props = {
 		flat?: boolean;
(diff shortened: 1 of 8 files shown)


apps/desktop/src/components/CommitContextMenu.svelte
```diff
@@ -214,24 +194,6 @@
 						}}
 					/>
 				</ContextMenuSection>
-				<ContextMenuSection>
-					<ContextMenuItem
-						label="Create dependent branch above"
-						disabled={refCreation.current.isLoading}
-						onclick={async () => {
-							await handleCreateNewRef(ensureValue(stackId), commitId, 'Above');
```
Byron, line 222: @PavelLaptev This hurts 😅 as it removes functionality that can't be achieved in another way right now.

Maybe from a UX perspective, we can rethink this operation so it's more of a 'split'? Could this be achievable differently? I keep thinking about the insertion-marker when creating new commits - inserting a reference at a selected point feels similar.

PR #9866, opened 2025-08-15
fix: Push status - If there are remote commits, force push is required

estib-vega:
If there are upstream commits in a branch, despite it having local or local-and-remote commits, pushing would remove the remote only commits and, in turn, require a force push


Byron:
If you have a specific case in mind, a test would be needed for that as well to pin it down. Even if existing tests break it's often best to add another one with the specific scenario you are fixing. It might overlap, but it's likely subtly different so it's definitely not a waste.

crates/but-workspace/tests/workspace/ref_info/with_workspace_commit/journey/integrate_with_merges.rs
```diff
@@ -492,7 +492,7 @@ fn remote_ahead() -> anyhow::Result<()> {
                             ],
                             commits_outside: None,
                             metadata: "None",
-                            push_status: NothingToPush,
+                            push_status: UnpushedCommitsRequiringForce,
```
estib-vega, line 495: @Byron I think this one is good enough as a test. It's basically checking whether remote ahead, not integrated, should have the push state

Byron: Indeed! I was looking at it an still thought there should be nothing to push, but it's true that pushing would require force here so the test was clearly wrong.
I have a feeling that in future, there will be more inter-op between "integrate remote" and these push buttons - after all one would want to make a fast-forward here really easy to do, while still allowing a force-push behind a safety check.

PR #9868, opened 2025-08-16
fix: resolve branch name parsing causing git reference errors

matthewevans:
## 🧢 Changes

  - **Add new `parseBranchName` utility function** to properly extract remote prefixes
  from branch names
  - **Fix BranchesView.svelte** to use `parseBranchName` when creating selection IDs,
  preventing unparsed branch names from reaching the backend
  - **Optimize StackView.svelte** to parse branch name and remote in a single reactive
  derivation instead of two separate calls
  - **Add comprehensive test suite** covering all parsing scenarios including edge cases
  and error conditions
  - **Clean up code quality** by removing redundant comments and eliminating duplicate
  parsing logic

  ## ☕️ Reasoning

  Users were encountering git reference errors like `"The reference
  'refs/heads/origin/feat/APL-2641-target-path-redirect-after-auth' did not exist"` when
  interacting with remote branches in the UI.

  **Root Cause**: Branch names containing remote prefixes (e.g.,
  `"origin/feat/branch-name"`) were being passed directly to the backend without parsing.
  The backend would then construct invalid git references by prepending `refs/heads/` to
  the unparsed name, resulting in malformed refs like `refs/heads/origin/feat/branch-name`
   instead of the correct `refs/remotes/origin/feat/branch-name`.

  **Solution**: The new `parseBranchName` function intelligently separates remote prefixes
   from branch names, ensuring the backend receives clean branch names and proper remote
  information. This allows the backend to construct valid git references and eliminates
  the reference errors.

  **Impact**: Fixes branch selection and viewing for all remote branches, improving the
  user experience when working with branches from multiple remotes.

Screenshot 2025-08-16 at 1 22 32 PM

### DIFF
--- a/apps/desktop/src/components/BranchesView.svelte
+++ b/apps/desktop/src/components/BranchesView.svelte
@@ -30,6 +30,7 @@
 	} from '$lib/selection/key';
 	import { STACK_SERVICE } from '$lib/stacks/stackService.svelte';
 	import { UI_STATE } from '$lib/state/uiState.svelte';
+	import { parseBranchName } from '$lib/utils/branch';
 	import { inject } from '@gitbutler/shared/context';
 	import { persisted } from '@gitbutler/shared/persisted';
 	import { AsyncButton, Button, Modal, TestId } from '@gitbutler/ui';
@@ -70,7 +71,8 @@
 			const branchName = current.remote
 				? current.remote + '/' + current.branchName
 				: current.branchName;
-			return createBranchSelection({ stackId: current.stackId, branchName });
+			const { branchName: parsedBranchName } = parseBranchName(branchName);
+			return createBranchSelection({ stackId: current.stackId, branchName: parsedBranchName });
 		}
 	});
 
--- a/apps/desktop/src/components/StackView.svelte
+++ b/apps/desktop/src/components/StackView.svelte
@@ -30,6 +30,7 @@
 	import { STACK_SERVICE } from '$lib/stacks/stackService.svelte';
 	import { UI_STATE } from '$lib/state/uiState.svelte';
 
+	import { parseBranchName } from '$lib/utils/branch';
 	import { computeChangeStatus } from '$lib/utils/fileStatus';
 	import { inject } from '@gitbutler/shared/context';
 	import { persistWithExpiration } from '@gitbutler/shared/persisted';
@@ -104,7 +105,16 @@
 	);
 
 	const commitId = $derived(selection.current?.commitId);
-	const branchName = $derived(selection.current?.branchName);
+	const { branchName, branchRemote } = $derived.by(() => {
+		const rawBranchName = selection.current?.branchName;
+		if (!rawBranchName) {
+			return { branchName: undefined, branchRemote: undefined };
+		}
+
+		const { branchName: parsedBranchName, remote } = parseBranchName(rawBranchName);
+		return { branchName: parsedBranchName, branchRemote: remote };
+	});
+
 	const upstream = $derived(selection.current?.upstream);
 
 	const activeSelectionId: SelectionId | undefined = $derived.by(() => {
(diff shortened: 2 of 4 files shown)


Byron:
Thanks so much for tackling this, I am sure this PR is desirable as it fixes a problem, and one I have seen before as well.

I will leave the review to the frontend team but just wanted to mention that ultimately, the frontend has to use full ref names internally in future to avoid any ambiguity when interacting with the backend. 
If the reference is a remote tracking branch, the backend has to provide the remote name as well as remotes can contain slashes, i.e. `origin/other` is a perfectly valid remote name, so `origin/other/branch` is either using the `origin` remote, or `origin/other`.

This is a bigger transformation, but it will have to happen at some point or (possibly subtle) bugs will keep creeping up.

krlvi:
Thanks for this fix @matthewevans! This seems like something we wanna to have but just cc-ing @estib-vega @mtsgrd here since they work mostly in that part of the code

apps/desktop/src/components/StackView.svelte
```diff
@@ -341,11 +351,14 @@
...
-		branchName
-	})}
+	{@const changesResult = (() => {
+		return stackService.branchChanges({
+			projectId,
+			stackId: stackId,
+			branchName,
+			remote: branchRemote
+		});
+	})()}
```
estib-vega, line 361: I don't think the IIFE is necessary here

apps/desktop/src/lib/utils/branch.ts
```diff
@@ -71,3 +72,44 @@ export function getBestRemote(remotes: string[]): string | undefined {
...
+
+	const parts = branchName.split(BRANCH_SEPARATOR);
+
+	// No separator means it's a simple branch name
+	if (parts.length <= 1) {
+		return { remote: providedRemote, branchName };
+	}
+
+	const [firstPart, ...remainingParts] = parts;
+	const branchNameWithoutPrefix = remainingParts.join(BRANCH_SEPARATOR);
```
estib-vega, line 100: Sadly I think this part might need some extra logic.
Remotes in git can have slashes in them. So the only think we can do here is a best effort (without the help from the backend).

Could you change this function so that if there's a `providedRemote` defined, we test whether the branchName starts with that string as is?

That way, if the `providedRemote` is e.g. `feature/remote` and the branch name is `feature/remote/my-cool-branch` it will correctly determine what the branch name is and what the remote name is.

Also, a test for that (admittedly) edge case would be covered

estib-vega (commented):
Thanks so much @matthewevans.
This already would improve something that is in a visibly broken state. I just left a couple of notes.

Ideally, in a short term, the backend handles more of this logic, since the FE doesn't have the full picture (it doesn't know what remotes are available).

apps/desktop/src/components/BranchesView.svelte
```diff
@@ -70,7 +71,8 @@
 			const branchName = current.remote
 				? current.remote + '/' + current.branchName
 				: current.branchName;
-			return createBranchSelection({ stackId: current.stackId, branchName });
+			const { branchName: parsedBranchName } = parseBranchName(branchName);
```
estib-vega, line 74: What happens here if the local branch is named `fork/my-branch`?

If it doesn't have a remote, the name `fork/my-branch` is passed to the function and it'll assume that the remote is called `fork`.

Of course, *who would namespace their branch `fork`?* but the point still stands.
That would yield unexpected behavior

estib-vega:
@matthewevans I went and did the changes on the backend to enable doing the correct thing and deferring handling the remote to the backend.
https://github.com/gitbutlerapp/gitbutler/pull/9885

Sorry to eat your lunch but this required a bit more context about the Backend & the selection system in order to get it done.

Thanks again for putting the effort on this PR.
It makes us really happy to get your contributions on this side of the PR :)

matthewevans:
@estib-vega nice! Always down for the _correct_ fix. Does that BE fix nullify this PR then? If so I can close. Thanks!

estib-vega:
> @estib-vega nice! Always down for the _correct_ fix. Does that BE fix nullify this PR then? If so I can close. Thanks!

Yes, we can close this one 👍

PR #9869, opened 2025-08-17
fix: Detect force push requirements for GitButler virtual branches

matthewevans:
🧢 Changes

TBD

  ☕️ Reasoning

  GitButler uses a unique "virtual branch" approach that creates local branches without setting up standard Git remote tracking relationships. This caused the force push detection logic to
  fail, resulting in users seeing only a "Push" button (which would fail) instead of the expected "Force push" button when:

  - Dragging files into existing commits to amend them
  - Local and remote branches have diverged requiring force push
  - Standard Git tracking lookup returns None for virtual branches

### DIFF
--- a/crates/but-workspace/src/branch_details.rs
+++ b/crates/but-workspace/src/branch_details.rs
@@ -2,6 +2,7 @@ use crate::ref_info::function::workspace_data_of_default_workspace_branch;
 use crate::ui::{CommitState, PushStatus, UpstreamCommit};
 use crate::{state_handle, ui};
 use anyhow::{Context, bail};
+use bstr::ByteSlice;
 use but_core::RefMetadata;
 use gitbutler_command_context::CommandContext;
 use gitbutler_error::error::Code;
@@ -114,6 +115,13 @@ pub fn branch_details(
 /// Note that for stacks, we shouldn't call `stack_details_v3`, but instead [`head_info()`](crate::head_info()) to get all stacks
 /// reachable from the current HEAD.
 ///
+/// ### Remote Reference Detection
+///
+/// This function uses Git's standard tracking branch lookup first. If that fails (as is common with GitButler
+/// virtual branches that don't follow standard Git tracking conventions), it falls back to constructing
+/// remote references based on the integration branch and branch name. This fallback is essential for
+/// accurate force push detection in GitButler workflows.
+///
 /// ### Implementation
 ///
 /// Note that the following fields aren't computed or are only partially computed.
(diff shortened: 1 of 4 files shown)


Byron:
Thanks for tackling this!

`branch_details_v3()` is used in a different context, in the branch listing pane, which doesn't allow to push anything. There was a issue fixed recently which might be related (#9866), which might well be what this PR attempts to fix.

Please feel free to add videos showing the before and after this commit is applied and how it affects the push-related workflows. Right now I'd have trouble seeing how this is effective in practice.

Independently of that, It's notable that the push-status label seems wrong in more cases than one…

Screenshot 2025-08-17 at 06 32 40

…but that's expected at this point, all of this is scheduled for an overhaul and what's here is merely 'made to somewhat work', meanwhile.

Last but not least, could you clarify if this is (co-)written by an LLM?
Thank you 🙏.

matthewevans:
@Byron - Great call out. This PR unfortunately caught your attention but (at this point) was part of the testing flow for me attempting to reproduce the issue to see if the above fixed it.

I converted the PR to a draft because, well, it didn't fix it :)

---

I am still investigating two things:

1. Why branches don't have tracking set when they are initially pushed (as virtual branches). I think solving this is the real fix. When we push a branch, tracking should be set up automatically (and it used to do so), but that doesn't seem to be the case anymore.

2. Attempting to locate the remote to automatically set up tracking is I think a complicated way of approaching this, so I'm backing off that for now.

Regarding the PR you linked, if that's part of the latest nightly build, then the issue I'm experiencing it doesn't seem to be fixed, as that's what I was using today when I noticed the issue.

To address your last question, yes, I use Claude Code, though I never one-shot. This specific _attempt_ went through multiple iterations, and with all contributions, I make every effort to ensure the possible solution is idiomatic, follows existing repository patterns, and includes test cases when applicable.

My mistake in this instance was pulling the trigger on the PR too early, hence the conversion to Draft prior to your comment.

The updated PR will include before/after evidence.

Byron:
Thanks for the clarification!

> The updated PR will include before/after evidence.

I think seeing the issue in a video would already help a lot. `branch_details_v3()` is most definitely not the place for a fix though, and even though the status computed there probably isn't correct, it's acceptable for now as no actions are taken based on that.

> Regarding the PR you linked, if that's part of the latest nightly build, then the issue I'm experiencing it doesn't seem to be fixed, as that's what I was using today when I noticed the issue.

Yes, that fix is included. But the way this fix works should be what you need to fix the other issue you are experiencing.

PR #9870, opened 2025-08-17
create-branch-for-commits

PavelLaptev:
- Simplify context menu labels for creating branches (above/below) in the UI
- Restore "branch from/after commit" actions to the commit menu

image


Byron (commented):
This looks great! 

I particularly like the rewording so the menu items aren't longer than the longest were before, restoring some balance.
It's nice to see the same treatment in the branch context menu, this works well.

Let me merge this in joy :).

PR #9894, opened 2025-08-19
Use the deduced push-remote if there is no deduced fetch remote.

Byron:
That way, the typical fork workflow would work naturally as one will push branches to the push remote and they will be shown.

Setting up the push remote like this…

Screenshot 2025-08-19 at 12 00 34

…and pushing to it…

Screenshot 2025-08-19 at 12 02 08

…still shows up as if there is no remote as it's not picked up automatically. It must be deduced.

### Tasks

* [x] improve virtual branches adapter for access to push remotes
* [x] reproduce in test (with fetch remote also available, and without)
* [x] adjust deduction to prefer the push remote for display
* [ ] see if UI tooltip can show the remote name


Byron:
CC @PavelLaptev

Byron:
@estib-vega is taking care of the tooltips to just consistently show up and say something about what's going to happen when the button is pushed. Thanks again, this excites me more than it has any right to :). @PavelLaptev

PR #9904, opened 2025-08-20
Enable Dependabot updates for Rust toolchain

JamieMagee:
## 🧢 Changes

Enables Dependabot updates for Rust toolchain

## ☕️ Reasoning

The project currently targets `nightly-2025-02-14`, which is quite out-of-date. Dependabot now supports updating `rust-toolchain.toml`: https://github.blog/changelog/2025-08-19-dependabot-now-supports-rust-toolchain-updates/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -31,3 +31,8 @@ updates:
     groups:
       rust-updates:
         update-types: [minor, patch]
+
+  - package-ecosystem: 'rust-toolchain'
+    directory: '/'
+    schedule:
+      interval: 'monthly'


Byron:
Thanks a lot!

This seems useful, even though I wonder if this is the kind of churn we want right now. @krlvi would probably know if we want to keep Rust toolchain updates manual (and deal with Clippy then), or have deal with clippy once a month with a chance that we observe new errors.

My preference would be to get to stable, but I think it's Windows builds that prevent that ATM.

kmk142789 (approved):

Mcrich23 (approved):

Byron:
I think this PR can now be closed as we managed to switch over to `stable` 🎉.

Issue #9900, opened 2025-08-19
API error: (stacks) - Update workspace button stuck in loading state.


tobeyboe:
### Version

0.5.1476 (0.5.1476) (Nightly) and Version 0.15.12 (0.15.12) (Standard)

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

I am unable to update workspace, its just spinning when clicked. 
In version 0.15.12 I get no error, only the spinning button. In Nightly build I get a api error:

> `command: stacks
> params: {"projectId":"5d79bd5c-85b8-471c-a3df-xxxxxxxx"})
> 
> Segment NodeIndex(8) has only 11 commit(s), cannot split at 13`

<img width="1011" height="815" alt="Image" src="https://github.com/user-attachments/assets/0578c94c-9f7a-4f17-a654-0bb8233d6ec5" />

### How to reproduce (Optional)

1. click Upstream commit
2. click Update Workspace

### Expected behavior (Optional)

1. should sync down commits, or give error. 
2. Should not get stuck in looping loading state.

### Relevant log output (Optional)

```shell

```

Byron:
Thanks a lot for reporting!

Did the error messages in the background always show for you? Or did that happen after the workspace update was started?

Finally, could you launch the [Nightly version](https://gitbutler.com/downloads) of the application from the terminal like `LOG_LEVEL=debug /Applications/GitButler\ Nightly.app/Contents/MacOS/gitbutler-tauri` and search for `panic` after pressing the update button? The common reason for a never-ending request is a fatal backend failure, something which really shouldn't happen.

Thanks for your help!

tobeyboe:
Its shows on startup

<img width="1436" height="622" alt="Image" src="https://github.com/user-attachments/assets/7e6b9520-58e7-4d81-9f64-95a6d19faf4c" />
And it is also triggered a extra when clicking "upstream commits"

<img width="1434" height="616" alt="Image" src="https://github.com/user-attachments/assets/39d2f5e9-df7e-4f0c-a3a4-2ddb1bedcf44" />

I never get to click the "Update workspace" button, it is already spinning when the menu opens from clicking "upstream commits"

There is no Panic in the log when clicking "upstream commits"
```
 Segment NodeIndex(8) has only 11 commit(s), cannot split at 13
[...]

PR #9905, opened 2025-08-20
Allow submitting anonymised dot-files with a separate button.

Byron:
It seems we have unforseen issues like #9900 with the post-processing stage of the graph,
so we could have a way to:

* anonymise graph dot files so no branch names contain PII anymore, just prefixes and consistent names.
* submit a dot-file of the graph with post-processing when possible or
* submit a dot-file of the graph without post-processing if that fails

### Tasks

* [x] graph without post-processing
* [x] PII cleanup of graph
* [x] flag to allow submitting a dot file in archival tool, with auto-fallback if post-processing fails
* [x] UI checkbox for this


Byron:
@estib-vega  I am seeing strange, possibly racy, behaviour and I don't know if it's something I did. Maybe reproducible, but probably not important enough to look into. Just in case you see something obvious.

https://github.com/user-attachments/assets/b8841335-1acc-44e8-957d-8e2dac896ecd

PR #9921, opened 2025-08-20
avoid auto-user setup

### LINKED ISSUE #8832: First Git run: Don't persist dummy-user to local configuration, instead fill it in on the fly while communicating the issue in the UI
### Version

0.14.26 and newer

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

The way this is currently implemented the user will see a popup once (without the option to fill in their name it seems), then a dummy name is written to the local repository configuration file.

This persistence then is a problem because users may have no idea why their actual username isn't picked up even after they set it globally.

> I did try to replicate the issue just now by renaming my .gitconfig file then trying to clone a project with GitButler, but GitButler seems to be using my information from GitHub for the author information (I might be wrong on this) instead of defaulting to firstname lastname. There was no popup when this happened and the git author name and email was silently written to the repository config file

So there seems to be more, maybe the frontend writing this information, which also means it will not respond to changes the user does themselves.

*Actually*, it wasn't well remembered if there even was a popup related to the lack of configured author.

This flow should be handled better. It seems easy enough to fill in the author name dynamically and let the UI know when switching to a project if no user exists. The UI can facilitate a fix then, maybe even by offering to set it globally.

Generally, I don't think GitButler should write the user identification unless the user triggered it and is aware of the scope of such action.

Byron:
We only have to get up and running, it's fine to fail commits for now, assuming that we iterate on this to help set a user via prompt.

I thin kit's a remnant of the past as well.

Fixes #8832.

### Task 

* [x] avoid making up a user
* [x] mark errors related to failed commit creation if there is no author configured
* [x] endpoint for storing the user name and email globally
* [ ] ~~`set_project_active` checks if the Git user is set~~

### Handed off to @estib-vega 

* [ ] Frontend-form that captures the user information and uses the new endpoint to store it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -860,6 +860,7 @@ dependencies = [
  "but-graph",
  "but-hunk-assignment",
  "but-hunk-dependency",
+ "but-rebase",
  "but-rules",
  "but-settings",
  "but-workspace",
--- a/crates/but-api/Cargo.toml
+++ b/crates/but-api/Cargo.toml
@@ -26,6 +26,7 @@ but-workspace.workspace = true
 but-core.workspace = true
 but-hunk-assignment.workspace = true
 but-action.workspace = true
+but-rebase.workspace = true
 gitbutler-branch.workspace = true
 gitbutler-branch-actions.workspace = true
 gitbutler-command-context.workspace = true
--- a/crates/but-api/src/commands/config.rs
+++ b/crates/but-api/src/commands/config.rs
@@ -29,3 +29,29 @@ pub fn set_gb_config(_app: &App, params: SetGbConfigParams) -> Result<(), Error>
         .set_git_settings(&params.config.into())
         .map_err(Into::into)
 }
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct StoreAuthorGloballyParams {
+    pub project_id: ProjectId,
+    pub name: String,
+    pub email: String,
+}
+
+pub fn store_author_globally_if_unset(
+    _app: &App,
+    StoreAuthorGloballyParams {
+        project_id,
+        name,
+        email,
+    }: StoreAuthorGloballyParams,
+) -> Result<(), Error> {
+    let repo = but_core::open_repo(gitbutler_project::get(project_id)?.path)?;
+    but_rebase::commit::save_author_if_unset_in_repo(
+        &repo,
+        gix::config::Source::User,
+        name.as_str(),
+        email.as_str(),
+    )?;
+    Ok(())
+}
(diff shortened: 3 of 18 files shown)


Byron:
Without a user setup, it's perfectly fine to add a new local repository.

When creating a new reference it fails like this:

Screenshot 2025-08-20 at 20 38 32

When creating a new commit it fails like this:

Screenshot 2025-08-20 at 20 39 29


Cloning a repository I didn't test, but `gitoxide` isn't going to fail even without a user, just like Git it makes it work.

This means I think there isn't anything holding us back to merge this, but to be sure lets have another set of eyes on it @krlvi . 

PS: For fun copilot was added as well, let's see what it has to say.

krlvi:
There is an inherent tradeoff here. It's not that GB wants to override any existing setting but rather, cover the case where none have been setup (which may be the case for some users who are new to coding and version control in general).

Can we solve this differently? I think producing these error messages on reference or commit creation is not desirable, and so I think changing this needs to be accompanied by some UX for handling the case of no name/email

Byron:
Let's do that!

I will update this PR to also make the kind of error detectable. Then the frontend can provide a popup that configures the user. This configuration could then even be done globally.

krlvi:
How about this though - this PR stays as is, and then in parallel we create a "check user" and "set user" APIs that are invoked immediately after a repo has been added, which hook up to a modal. For bonus points, the set user may offer to set the value globally.

Byron:
I see, so it's a proactive thing!

Then here is my version of it - same effect, but easier:

* finish the "set user globally endpoint" - started this before AFK
* add a check to `set_project_active` the sets a boolean to show if the user is setup.
* the frontend reacts to that accordingly and uses the new endpoint to set the user

Working on it just to warp this up.

Byron:
@estib-vega I think this PR has everything that you'd need to whip-up an error handler that responds to `Code::AuthorMissing`, picks up a name and email from the user, and uses the new endpoint to store that information.

~~A caveat: I didn't actually test that endpoint.~~ 
Maybe you can superseding this PR with your own, then wire everything up, as you probably can't push into this PR as it's in my fork.

Leaving this PR unmerged meanwhile.
Thanks again

PS: It looks like Playwright fails because there is no auto-fallback for the user anymore, and it will be keen to fill in the user and email into that new prompt :D. Amazing. Note that this also means that playwright needs to be isolated when running it locally, by controlling the [home of the user](https://git-scm.com/docs/git-config#Documentation/git-config.txt-gitconfig) or by setting [GIT_CONFIG_GLOBAL](https://git-scm.com/docs/git-config#Documentation/git-config.txt-GITCONFIGGLOBAL).

Byron:
It's in `user-setup-popup` now.

PR #9936, opened 2025-08-22
feat(ui): add Vue file type icon mapping

nshcr:
## 🧢 Changes

The file list view was missing the icon for `.vue` files.
I noticed the corresponding SVG icon already existed, but the extension-to-icon mapping hadn't been added.
This PR adds it.

Before:

After:


Byron (approved):
Thanks for this, great catch!

PR #9943, opened 2025-08-22
Author missing modal

### LINKED ISSUE #8832: First Git run: Don't persist dummy-user to local configuration, instead fill it in on the fly while communicating the issue in the UI

estib-vega:
Pop a modal asking for information if that’s missing from the git config

Summary
- Add authorInfo query and integrate it into GitService via injected endpoints.
- Add new "author-missing" GlobalModal variant and UI to prompt when commit author info is absent.
- Ensure UI checks for author info at project load and displays modal when missing.
- Add E2E test verifying commit flow without global git author config.

What changed
- injectEndpoints helper defining authorInfo query (extraOptions.command 'get_author_info', passthrough args, providesTags ReduxTag.AuthorInfo).
- GitService now accepts BackendApi in constructor, stores this.api, and exposes getAuthorInfo(projectId).
- Export AuthorInfo type and import cache tag helpers.
- Extend GlobalModalType with 'author-missing' and add AuthorMissingModalState (projectId, authorName, authorEmail).
- Minimal Author Missing modal UI branch in GlobalModal.svelte with sizing, test id, padding, and close button.
- Add GlobalModal_AuthorMissing test id.
- Project load now ensures git author is configured and triggers modal when absent.
- E2E test added for committing without global author config.

Testing
- New E2E test confirming graceful handling of missing global git author during first commit.

fixes https://github.com/gitbutlerapp/gitbutler/issues/8832


e2e/playwright/tests/startTheApp.spec.ts
```diff
@@ -84,3 +84,82 @@ test('should start the application and be able to commit', async ({ page, contex
 	await expect(commitRows).toHaveCount(1);
 	await expect(commitRows.first()).toHaveText(newCommitMessage);
 });
+
+test('no author setup - should start the application and be able to commit', async ({
```
Byron, line 89: Loving this test - and no mocking required!

Byron (commented):
I love that we are getting a way to help the user fill in what's needed, and also because it's an opportunity.

An opportunity to be better, not only with this modal in particular, but with 'this kind of situation' in general.

Here is my pitch: instead of bugging users with stuff they are not interested in *at that moment*, we bug them only when truly required. We could be extra good by adding a mechanism to show "attention required". The visual language used could be consistent with other places where attention is needed, like when conflicts need to be resolved, etc.

For instance, instead of showing a popup right after the user just wanted to add their first project, we could instead show some sort of info-bubble that reveals the problem "Hey, just know that no user is setup which is required for committing. Click to fix…". It's non-blocking, of course. Then, when the user tries to commit, without reading it, the UI could integrate the user information *directly* into the commit view and *prevent* the user from hitting the button until it is set. I think this would be endgame level of perfection.

Until that is achievable, I'd go with showing this popup reactively when something fails due to the author missing. That's not super great as the user might lose a commit message, but maybe the UI is already *not* discarding it if the commit fails.

What's your thoughts, @PavelLaptev, @krlvi and @estib-vega ?

estib-vega:
> Here is my pitch: instead of bugging users with stuff they are not interested in _at that moment_, we bug them only when truly required. We could be extra good by adding a mechanism to show "attention required". The visual language used could be consistent with other places where attention is needed, like when conflicts need to be resolved, etc.
> 
> For instance, instead of showing a popup right after the user just wanted to add their first project, we could instead show some sort of info-bubble that reveals the problem "Hey, just know that no user is setup which is required for committing. Click to fix…". It's non-blocking, of course. Then, when the user tries to commit, without reading it, the UI could integrate the user information _directly_ into the commit view and _prevent_ the user from hitting the button until it is set. I think this would be endgame level of perfection.


Actually I'd prefer this for a couple of reasons. Both technical constraint and UX.

### The technical
The way we catch errors and then surface them as toasts works at the window level. We attach a global listener that basically is the last line of defense when dealing with Exceptions.
The FE purposefully throws things in order to have that single point in which we raise the toast.

We can't (easily) control whether a modal is displayed from that point. Toasts are more like free agents that can be instanced  from anywhere.

We could change this a bit, but that would entail a bigger effort

### The UX
I think we *should* ask this in the beginning.
My understanding is that the scenario in which you start gitbutler and still haven't setup your git credentials is rare enough that it would be better to just check out of the go.
Asking for this information right before a commit seems more intrusive almost.
If we preemptively ask for this, we don't throw an error and interrupt an action.

What do you guys think?

Byron:
> Asking for this information right before a commit seems more intrusive almost.
> If we preemptively ask for this, we don't throw an error and interrupt an action.

In the endgame scenario, I was seeing the author information as part of the commit-view, and it's the truth that it's required to make a commit. It's now needed, and there is no way to ignore it, but at least we would show it proactively.

The MVP version I suggested, responding to an error (independently of technical feasibility) is really just that, and it's not great, but might be less of a problem if the UI shows the same message again as it knows something went wrong (so *not much* lost). I clearly optimise for 'unblocking' the path of users and not 'nag' them with stuff they aren't interested in (in that moment).

With that said, if it's technically easier to prompt right away then *this* should be the MVP, and my comment earlier was more about the end-game of the UI when it comes to this. And there is more in that direction: target-ref selection, Git setup, etc. Ideally, there is none of that until truly required (and the new backend can handle it, too, for that reason).

PavelLaptev:
> Here is my pitch: instead of bugging users with stuff they are not interested in at that moment, we bug them only when truly required.

I really like this approach. The less a user has to configure at the beginning, the better. They won’t get bored, and we break the larger setup into smaller, contextual steps.

I think it could ask only when the user is trying to do some git related stuff — commit, push, etc.  
But if it’s easier to just do it upfront, let’s go with what’s easier since it’s a rare case.

PavelLaptev:
@estib-vega I made a few changes:
- added cancel button
- updated copy
- added email validation
- refactored some components (new ModalFooter and EmailTextbox with built-in logic)

Also, I also sync all [modals in Figma](https://www.figma.com/design/ShIiR6hI5dzH7sT5L03Jg1/App-Design-3.5?node-id=14413-51145&t=gojfkr9Iy1Dic4KM-1). Thanks for the ping.

https://github.com/user-attachments/assets/cde331de-cfbb-4242-8615-2bff215c2e43

Byron:
That's great to hear!

> I think it could ask only when the user is trying to do some git related stuff — commit, push, etc.
> But if it’s easier to just do it upfront, let’s go with what’s easier since it’s a rare case.

I think, while going with the technically easiest version now, we can also be driven by metrics and see what percentage of our users bumps into this popup. Can we have that?

I am thinking that in future, once GB becomes the default Git client, there will be many who didn't setup Git, and probably who don't even have it installed. We should know either way.

PR #9938, opened 2025-08-22
user-setup-popup

estib-vega:
Summary
- Replace automatic user setup with a popup flow that defers user creation and allows later failure handling.
- Expose missing author condition via an attached error code for discoverability.
- Add a new endpoint to store a global user when none is set.
- Ensure deterministic git config in e2e tests by providing a .gitconfig fixture and propagating GIT_GLOBAL_CONFIG to test and server processes.

What changed
- Removed automatic user setup; operations that require an author may now fail and surface a prompt-based setup later.
- Introduced an error code to make "no author" situations detectable programmatically.
- Implemented an API endpoint to persist a global user configuration.
- Standardized git identity/config for Playwright e2e tests and updated env/test/server scripts to use a shared GIT_GLOBAL_CONFIG.

PR #9944, opened 2025-08-22
cleanup but-rebase

Byron:
Follow up on #9938 and the PR that creates the new endpoint in the first place.

It's probably fine to use `but-api` and move it 'down' only when other plumbing crates also need the new functionality.


Byron:
I'd love to use `but-api` more for these kinds of functions, i.e. plumbing that other plumbing crates *do not* call, but due to `Params` it doesn't feel like a place where testing is going to look too good.

PR #9947, opened 2025-08-23
Add file icon mapping for .cts and .mts extension

4O4:
## 🧢 Changes

Before:
ts-icons-before

After:
ts-icons-after

## ☕️ Reasoning
These extensions were introduced in TypeScript 4.7

See: 
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#new-file-extensions


Byron:
Wonderful, thanks!
Many thanks for the before/after as well!

Byron (approved):

PR #9948, opened 2025-08-23
Add file icon mapping for .graphqls extension

4O4:
## 🧢 Changes
Before:
graphqls-icon-before

After:
graphqls-icon-after

## ☕️ Reasoning
It's a less common extension for GraphQL schema files, but some projects just choose to use this one 🤷🏻‍♂️  It's not a complete niche because it's recognized across most of the GraphQL tooling. VSCode built-in icon theme also has this association.

References:
- https://github.com/graphql/graphiql/tree/b3c42a65160ba4878b5ea87d9e72385889856464/packages/vscode-graphql-syntax#graphql-syntax-support
- https://github.com/ardatan/graphql-tools/blob/0a7cc399c89ca6e94cd20502e961200395958173/packages/load-files/src/index.ts#L95
- https://github.com/microsoft/vscode/blob/a86edb56835c46cd6d233455a375ffb92b93a011/extensions/theme-seti/icons/vs-seti-icon-theme.json#L1621

### DIFF
--- a/packages/ui/src/lib/components/file/typeMap.ts
+++ b/packages/ui/src/lib/components/file/typeMap.ts
@@ -94,6 +94,7 @@ export const symbolFileExtensionsToIcons: { [key: string]: string } = {
 	gql: 'graphql',
 	gradle: 'gradle',
 	graphql: 'graphql',
+	graphqls: 'graphql',
 	h: 'h',
 	haml: 'haml',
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks so much!

It looks a bit like we could just mirror the 2500 lines of VSCode extension mappings ;).

PR #9949, opened 2025-08-23
feat(icons): add icons for Godot Engine related files

nshcr:
## 🧢 Changes

Add five icons for [Godot Engine](https://github.com/godotengine/godot) related files:

1. godot -> `.godot`
2. gdscript -> `.gd`
3. gdshader -> `.gdshader`
4. tres -> `.tres`
5. tscn -> `.tscn`

SVG from this PR: https://github.com/vscode-icons/vscode-icons/pull/3719

A few points to note:

1. I compressed the original SVG and removed unnecessary tags to reduce size.
2. I slightly darkened the fill color of some icons to make them easier to recognize.
3. The gdscript icon was too similar to the gear icon, which is often used for environment files, and could cause confusion.

Before:

After:

And here's how it looks inside Godot Engine:


Byron:
Thanks so much!

This looks good to me, but I want to reel in @PavelLaptev just to double-check the Icon/SVG work here. I think it might be important to know where these SVGs are coming from and if the license is acceptable for GitButler.

nshcr:
@Byron I hadn't noticed where the SVG files in that PR came from, I thought they were from the official repo of Godot. But after compared them, I found the code wasn't the same (maybe compressed).
I've now replaced all the SVGs with the versions from the official repo, all under the MIT license, so it should be fine now.




- https://github.com/godotengine/godot/blob/master/modules/gdscript/icons/GDScript.svg
- https://github.com/godotengine/godot/blob/master/editor/icons/Godot.svg
- https://github.com/godotengine/godot/blob/master/editor/icons/PackedScene.svg
- https://github.com/godotengine/godot/blob/master/editor/icons/Shader.svg
- https://github.com/godotengine/godot/blob/master/editor/icons/ResourcePreloader.svg

PavelLaptev:
@Byron thanks for pinging me! Yes, I’d like to align it with our icons before merging.

PavelLaptev:
@nshcr thanks for the PR!  

Most of the icons you added are generic, so they should share the same style with the rest generic icons.  
We have a [library file in Figma](https://www.figma.com/design/g5vfx1IlY9xuK99NhHW5d0/%F0%9F%91%BE-File-Icons?node-id=20521-84115&t=QT7XochLF3e2PtvS-1) with all the icons.  

File formats here are specific to Godot, but at a higher level they are generic:  
- gdshader → shader  
- tres → package  
- tscn → scene (clipboard)  
- gd → script file (gear)  

I also took this PR as an opportunity to update the generic icon colors.  

image

@nshcr I’ll plan to merge this later today unless you’d like to make any changes.

nshcr:
@PavelLaptev I'm fine with it! The generic icon actually looks better than the official one and keeps the overall style consistent. I don't have anything else to change.

---

By the way, I'd like to ask an unrelated question (I know Discord might be a better place, but I'm not used to it, so I'll just ask here).

Besides Godot, I'd also like to add icon support for some other file formats I often use. Some seem clearly suitable for a PR, while others I'm less sure about.

I'm a developer from China, and in the Chinese developer ecosystem there are a number of file extensions mostly used locally, often as wrappers around standard formats.

For example:
- `.wxml`: non-standard extension of HTML
- `.wxss`: non-standard extension of CSS
- `.uvue`: non-standard extension of Vue
- `.uts`, `.ets`: non-standard extensions of TypeScript

They're not universal standards, but are widely used within their respective domains, each with a large developer base.
Supporting them usually just means mapping to existing generic icons rather than creating new ones.
Would PRs for adding such mappings be acceptable? Thanks for your time!

PavelLaptev:
@nshcr thanks! Sure we can add icons for all these formats.

Just wondering, I've tried to find `.uvue`. Is it [this repo](https://github.com/universal-vue/uvue?tab=readme-ov-file)? It appears they haven't updated it in years, and many links are dead.

nshcr:
@PavelLaptev No, not that repository.

The `uvue` I'm referring to is the file extension used by [uni-app](https://github.com/dcloudio/uni-app). It adds cross-platform native compilation support to Vue (compiling to Web, iOS, and Android, somewhat like Flutter, but using Vue as the UI framework) [1].
It's usually used together with `uts` [2], in the some use cases, they're compiled into `wxml` + `wxss` + `js`.
`ets` is a relatively new file extension, used for the [ArkTS language](https://en.wikipedia.org/wiki/ArkTS).

Chinese only docs:
1: https://doc.dcloud.net.cn/uni-app-x/vue/
2: https://doc.dcloud.net.cn/uni-app-x/uts/

I'll add support for those new icons in a few days. Thanks a lot!

PavelLaptev:
@nshcr thanks for the explanation. 
Added support for these icons:

`.wxml`: green html icon
`.wxss`: green CSS icon
`.uvue`: green Vue icon
`.uts`: green TS icon
`.ets`: orange TS icon

image

nshcr:
@PavelLaptev Thanks for adding support for those icons!

Actually, besides `.wxml` and `.wxss`, there's also `.wxs`, which is an extension of JavaScript [1]. Sorry I forgot to mention it earlier, since this file extension is less commonly used directly.
From what I see in your additional commit, `js-green` could be a suitable icon for `.wxs`?

For reference, here's how the [official JetBrains plugin](https://plugins.jetbrains.com/plugin/24687-wechat-mini-program) handles these three formats (HTML, CSS, and JavaScript icons with an extra green logo).



By the way, since you've added `.ets`, maybe `.ejs` ([Embedded JavaScript templates](https://github.com/mde/ejs)) could also be included — though they are completely different technologies.
There's also `.nvue` (meaning Native Vue) [2], which belongs to the same ecosystem as `.uvue`, so maybe it could use the same icon.

Sorry if this causes you extra work!
I was originally planning to prepare a complete version in a PR in a few days, so the examples I gave here weren't fully comprehensive.
It should be almost complete now though — thanks for taking care of it ahead of time! I really appreciate it.

The three additional icons:
- `.wxs`
- `.nvue`
- `.ejs`

1: https://developers.weixin.qq.com/miniprogram/en/dev/framework/view/wxs/
2: https://uniapp.dcloud.net.cn/tutorial/nvue-outline.html

PavelLaptev:
@nshcr thanks and no worries, it's good that you mentioned these formats. I didn’t include the WeChat logo since our icons are usually smaller, and the logo would just look like a messy small dot.  

image

PavelLaptev:
Going to merge this one. Feel free to request more icons if needed 🙂

PR #9971, opened 2025-08-26
sc-but-exp

schacon:
Some more but commands as an experiment.


krlvi:
@schacon run `cargo fmt`

PR #9996, opened 2025-08-27
Open in Editor hotkey Short cut 

DevAbdoTolba:
## 🧢 Changes

Add a hotkey shortcut "Open in Editor" item with accelerator to the project menu CmdOrCtrl+Shift+A, just like GitHub Desktop

## ☕️ Reasoning

Make the app more relevant and familiar to those who came from GitHub Desktop background

### DIFF
--- a/crates/gitbutler-tauri/src/menu.rs
+++ b/crates/gitbutler-tauri/src/menu.rs
@@ -158,7 +158,11 @@ pub fn build<R: Runtime>(
                 .build(handle)?,
         )
         .separator()
-        .text("project/open-in-vscode", "Open in Editor");
+        .item(
+            &MenuItemBuilder::with_id("project/open-in-vscode", "Open in Editor")
+                .accelerator("CmdOrCtrl+Shift+A")
+                .build(handle)?,
+        );
 
     #[cfg(target_os = "macos")]
     {


Byron:

This makes a lot of sense, and I remember @krlvi having a lot of details about hotkeys already sketched out, he'd probably be best for the review.

krlvi:
@DevAbdoTolba thank you very much for this PR. There are a few things to be considered here - one related specifically to the "open in editor" action and another for the broader subject of shortkeys in the app.

Regarding "open in editor" or other actions that are performed on a specific file or a hunk - this is connected to the notion of "current focus". Currently the app has some shortcomings in that regard that are being eliminated. And this should of course extend to opening a specific hunk. So whenever the shortkey for "open in editor" is pressed, the expected thing gets opened. 

Broadly speaking regarding shortkeys - this is something we will be able to tackle more systematically as soon as we have some of the focus and navigation primitives nailed.  Here is an _out of date_ example of the type of structure we are considering, just to give a taste. And when it comes to the key combinations, I agree that it is reasonable to do keys that are familiar.
https://docs.google.com/spreadsheets/d/1Ed88h2WYg2OGiJjKqnDChrB0Wr1auLwg8RAm5jm_1pw/edit?usp=sharing

With all of that said, I am afraid we can't take the PR at this time because of the focus issue mentioned earlier. However this is something we will be able to address in the very near future.

PR #9997, opened 2025-08-27
fix: remove semicolons from text editor line wrapping

martypenner:
## 🧢 Changes

When writing commit or PR messages in the floating text editor, I noticed that backspacing in first wrapped line of a multiline… line… within a bullet (e.g. `- `) inserts a semicolon at the beginning of that wrapped line. It also removes the first character of the whole line. I made a one-line change to strip the semicolon from the `parseBullet` function, which seems like a typo to me. If I'm missing some context here or there's a good reason for this, feel free to close without comment. 😊


Byron:
Thanks a lot for contributing a fix, it's much appreciated!

If tests pass, I'd think @mtsgrd would know if this was intentional or not. Please note that from the description, I wasn't able to run into this issue, so it's probably not triggered trivially.

mtsgrd:
This is great, thank you for finding this!

Issue #10008, opened 2025-08-28
Failed to determine storage location for Git user configuration


LeiYangGH:
### Version

0.15.15

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

i am new user and after installing and open a local repo, it prompts for user name and mail.
this is strange to me already because i have been using git for long time, and sure to have them already configured globally.

after input some name and mail, error dialog shows

> command: store_author_globally_if_unset
> params: {"projectId":"4819f530-4274-49fc-9709-fb20ea2fb033","name":"leiyang","email":"lei@yang.com"})
> 
> Failed to determine storage location for Git user configuration

<img width="373" height="158" alt="Image" src="https://github.com/user-attachments/assets/47db3679-eb29-4dbe-adcb-a85c2f79bc9d" />

### Relevant log output (Optional)

```shell

```

Byron:
Thanks a lot for reporting!

Could you run this command from a Git bash to show which configuration is loaded on your system?

```
git config -l --show-origin | awk '{ print $1 }' | sort -u
```

Additionally, this one would be good to have to see exactly where your existing user related configuration is stored.

```
git config --show-origin user.name
```

Once it's clear where your user is stored, I think we can fix it by letting it fallback to the respective configuration source, like `System`.

Thanks a lot, and sorry for the hassle.

PS: [This code](https://github.com/GitoxideLabs/gitoxide/blob/17835bccb066bbc47cc137e8ec5d9fe7d5665af0/gix-config/src/source.rs#L100-L108) is probably the culprit as it doesn't find any of the relevant environment variables, and thus fails to find the user configuration.

Byron:
Once #10010 is merged, you should be able to try it out tomorrow in the latest [Nightly](https://gitbutler.com/downloads) - your feedback would be appreciated, I'd hope my blind fix can work around this.

LeiYangGH:
```bash
MINGW64 /c/GL/internal-cid (main)
$ git config -l --show-origin | awk '{ print $1 }' | sort -u
file:.git/config
file:C:/Git/etc/gitconfig
file:C:/Users/yang.lei/.gitconfig

/c/GL/internal-cid (main)
$ git config --show-origin user.name

 /c/GL/internal-cid (main)

```

`git config --show-origin user.name` gives emtpy, but i have no problem making commits, i think git automatically detects my system user?

[...]

PR #10010, opened 2025-08-28
Automatically fallback to at most one other configuration source if the first fails (#10008)

Byron:
Realted to #10008


crates/but-rebase/src/commit.rs
```diff
@@ -38,8 +39,16 @@ pub fn save_author_if_unset_in_repo<'a, 'b>(
...
+    let config_path = Some(destination)
+        .into_iter()
+        .chain(match destination {
+            Source::System => Some(Source::Git),
+            Source::Git => Some(Source::User),
+            Source::User => Some(Source::Git),
+            _ => None,
+        })
+        .filter_map(|source| source.storage_location(&mut |name| std::env::var_os(name)))
+        .next()
```
Byron, line 53: @copilot This is what you should have done. Look at what you did in #10009! Bad, bad bot!

PR #10017, opened 2025-08-29
Windows environment fix

### LINKED ISSUE #10008: Failed to determine storage location for Git user configuration

Byron:
It looks like #10008 is having trouble to find the correct environment variables
from the terminal, which affects our ability to write Git configuration files for the user
on Windows.

~~See if we can use the Git bash specifically to achieve that, and fixes #10008 in the process,
and related issues.~~

We cannot and should not use Git bash as shell on Windows due to the special unix paths in its environment, something that works very differently in Windows where these paths are relative to the drive of the current working directory.

Now we only make that clear and use a fixed version of `gix` which can detect the homedir on Windows even if no `HOME` variable is set.

### Tasks

* [x] sort out environment hassle
* [x] revert attempt to use `git bash` - doesn't work as paths are Unix style
* [x] validate `gix` fix works
* [x] update `gix` to latest version from main

### The Trap

One path, two ways of interpreting it, which was very confusing in the context of this issue.

Screenshot 2025-08-29 at 18 31 04
Screenshot 2025-08-29 at 18 30 19

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3767,7 +3767,7 @@ dependencies = [
 [[package]]
 name = "gix"
 version = "0.73.0"
-source = "git+https://github.com/GitoxideLabs/gitoxide?branch=main#f7f087cb7b2d239688203f58d166fde42d4bde8c"
+source = "git+https://github.com/GitoxideLabs/gitoxide?branch=main#10a20d6b3027179791c04a95731e1697001fb5f8"
 dependencies = [
  "gix-actor 0.35.4",
  "gix-attributes 0.27.0",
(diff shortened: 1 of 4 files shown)


Byron:
I am having trouble running GitButler without a terminal on Windows - the nightly debug build automatically brings up a terminal, which makes it impossible to test the code that runs only when there is no terminal.

Let's build a new nightly ~~after this is merged~~ from this branch and try it out on Windows (the only platform affected by this change).

PR #10025, opened 2025-08-31
fix: correct redirect path for customize prompts button

nshcr:
## 🧢 Changes

The button (Customize prompts) under Project Settings -> AI options was using `/settings/ai` as its redirect path, while other pages use `/new-settings` through a shared helper, it seems that `/settings` has already been deprecated.

This PR fixes the inconsistency by using the same method as other pages.

Reference code:

https://github.com/gitbutlerapp/gitbutler/blob/fea03750de890153acb7c23ba42e857908678a16/apps/desktop/src/lib/routes/routes.svelte.ts#L92-L101

https://github.com/gitbutlerapp/gitbutler/blob/fea03750de890153acb7c23ba42e857908678a16/apps/desktop/src/components/Feed.svelte#L134-L136

Before & After screenshot video:
https://github.com/user-attachments/assets/000c598c-ea4d-40d2-8a38-17728f0975e7

### DIFF
--- a/apps/desktop/src/components/CloudForm.svelte
+++ b/apps/desktop/src/components/CloudForm.svelte
@@ -5,6 +5,7 @@
 	import WelcomeSigninAction from '$components/WelcomeSigninAction.svelte';
 	import { projectAiExperimentalFeaturesEnabled, projectAiGenEnabled } from '$lib/config/config';
 	import { focusable } from '$lib/focus/focusable';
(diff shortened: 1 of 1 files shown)


Byron (approved):

This looks great to me, but for lack of some context it's best to wait for a second review.

mtsgrd:
Looks great, thanks for fixing this! I'll go ahead and rename `new-sttings` -> `settings` in a separate pr.

PR #10095, opened 2025-09-04
Simplify but-api functions with a procmacro

krlvi:
This is a continuation from https://github.com/gitbutlerapp/gitbutler/pull/10093 where it removed an unnecessary `app` dependency that was being passed in all commands for no reason. 

The aim here is to further simplify the functions that we create in `but-api` such that they can be conveniently used from various high level sources:
- tauri commands
- but-server endpoints (electron)
- but (CLI)
- AI/Agent tools
- Tests 

The `but-api` functions already give us a high level functions/ surface area, so the aim here is to make that conveniently usable without having to create "params" structs for all functions.

This uses the `workspace` module as the starting point, with the intention of converting all funcs in a follow up PR

### DIFF
--- a/crates/but-api-macros/src/lib.rs
+++ b/crates/but-api-macros/src/lib.rs
@@ -1,0 +1,72 @@
+use convert_case::{Case, Casing};
+use proc_macro::TokenStream;
+use quote::{format_ident, quote};
+use syn::{FnArg, ItemFn, Pat, parse_macro_input};
+
+/// To be used on functions, so a function `func` will be turned into:
+/// * `func` - the original item, unchanged
+/// * `func_params(FuncParams)` taking a struct with all parameters
+/// * `func_cmd` for calls from the frontend, taking `serde_json::Value` and returning `Result<serde_json::Value, Error>`
+#[proc_macro_attribute]
+pub fn api_cmd(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let input_fn = parse_macro_input!(item as ItemFn);
+
+    let vis = &input_fn.vis;
+    let sig = &input_fn.sig;
+    let fn_name = &sig.ident;
+    let output = &sig.output;
+
+    // Collect parameter names and types
+    let mut fields = Vec::new();
+    let mut param_names = Vec::new();
+    for arg in &sig.inputs {
+        if let FnArg::Typed(pat_type) = arg {
+            let ty = &pat_type.ty;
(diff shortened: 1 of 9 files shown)


crates/but-api-macros/src/lib.rs
```diff
@@ -0,0 +1,68 @@
+use convert_case::{Case, Casing};
+use proc_macro::TokenStream;
+use quote::{format_ident, quote};
+use syn::{FnArg, ItemFn, Pat, parse_macro_input};
+
+#[proc_macro_attribute]
```
Byron, line 10: I think it would be useful to document in rough terms what functions it generates and what their purpose is.

```suggestion
/// To be used on functions, so a function `func` will be turned into:
/// * `func` - the original item, unchanged
/// * `func_params(FuncParams)` for calls from the frontend
/// * `func_cmd` for calls from CLI tools (?)
#[proc_macro_attribute]
```

Byron (approved):
This is fantastic and a perfect use of proc-macros!

I love the macro itself as well, it's surprisingly readable. Is this based on Claude by any chance? Maybe this is how you do write proc-macros in 2025 - as someone who doesn't have much experience with them it looks very good.

krlvi:
> Is this based on Claude by any chance?

I have not gotten used to Claude yet, but I did use my regular AI in Zed where I ask it what is idiomatic and telling it "pls make this snippet more idiomatic" :D

crates/but-api-macros/src/lib.rs
krlvi: Good point. Done

Byron:
I see! I usually take these single-line comments as evidence, humans do that much less. That's not to say that you wouldn't be able to come up with this on your own, proc-macros never really opened up to me.

krlvi:
The way i did this particular one was asking it to:
1. explain how the tauri macro worked (i gave it the source as context)
2. with the explanation, i concocted a bad and ugly version that did what I want
3. I told it "make this code idiomatic" 

The step 3 is where the magic happens because it understood what I was trying to do and gave me a nice readable version with comments that did exactly what was intended.

krlvi:
This whole thing makes me think... how cool would it be if the VCS could capture these state transitions of the code... and the AI interaction...

Byron:
Thanks for sharing the workflow, I will keep that in mind while inevitably trying to use these tools more myself. 
Particularly when I am uncertain or am afraid, I should use them, clearly.

PR #10111, opened 2025-09-05
Reduce file watcher debounce timeout from 250ms to 50ms (non-windows)

mtsgrd:
The 250ms delay has been annoying me for a while, but might be necessary
to prevent issues on Windows.

### DIFF
--- a/crates/gitbutler-filemonitor/src/file_monitor.rs
+++ b/crates/gitbutler-filemonitor/src/file_monitor.rs
@@ -15,7 +15,13 @@ use crate::events::InternalEvent;
 const DEBOUNCE_TIMEOUT: Duration = Duration::from_secs(60);
 
 // The internal rate at which the debouncer will update its state.
-const TICK_RATE: Duration = Duration::from_millis(250);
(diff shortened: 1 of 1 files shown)


crates/gitbutler-filemonitor/src/file_monitor.rs
```diff
@@ -15,7 +15,13 @@ use crate::events::InternalEvent;
 const DEBOUNCE_TIMEOUT: Duration = Duration::from_secs(60);
 
 // The internal rate at which the debouncer will update its state.
-const TICK_RATE: Duration = Duration::from_millis(250);
+// Keeping a higher timeout on Windows because of file-system issues related
+// to `virtual_branches.toml`.
+const TICK_RATE: Duration = if cfg!(target_os = "windows") {
```
Byron:
```suggestion
const TICK_RATE: Duration = if cfg!(windows) {
```

Byron (approved):
LGTM!

Also interesting that in the end, Claude still didn't reach the most idiomatic version of this, but got better incrementally.

PR #10123, opened 2025-09-05
Add desktop README with run instructions for the server

estib-vega:


Byron:
THIS IS AWESOME 🙏.

And what's best is that it won't force rebuilding the world all the time, so this will be powering my dev-experience from now on!

PR #10134, opened 2025-09-07
Use the but_api workspace functions directly as tauri commands

krlvi:


Byron (commented):
It's probably a good idea to use `tauri` annotations directly if possible, and while we obviously depend on it.
My intuitive response was "but this breaks separation", but on second thought it's the path of least resistance which will save time.

I have just created #10135 and it's moving instrumentation to `but-api`, and maybe a follow-up could continue the work started here even. To be seen.

krlvi:
One thing that is different now from before is we no longer have shared state that gets injected by Tauri, so the function signatures are reusable:
( https://github.com/gitbutlerapp/gitbutler/pull/10057 followed by https://github.com/gitbutlerapp/gitbutler/pull/10093 )

I agree - it is unfortunate that we add a dependency to tauri in `but-api`. I wish I could somehow scope it to only the tauri macro `#[tauri::command(async)]`.
But at the same time, the indirection and the extra function wrapping was just not nice at all.

PR #10144, opened 2025-09-08
Improve no-relative-imports

Caleb-T-Owens:
I've been doing some experiments off-hours to look into handling tsconfig 'references' which are commonly used in the vite templates for SSR stuff


Caleb-T-Owens:
Keeping this draft PR around because it has some relevant test cases that we need to address.

krlvi:

Caleb-T-Owens:
This is still relevant. Especially for but lite.

PR #10192, opened 2025-09-10
Fix e2e tests by avoiding error popup

Byron:
Even though this should work, it's just a quick-fix before the proper fix.

### Tasks

* [x] try backend hack - works
* [x] try actual fix in image


Byron:
Seems to work.

PR #10194, opened 2025-09-10
better rename handling

Byron:
Fixes #10193.

### Tasks

* [x] impl


Byron:
Works, only tested manually for `a` -> `a/b` and `a/b` -> `a`.

Tests would be needed for the V3 implementation, let's not forget.

Issue #10198, opened 2025-09-10
Strange `virutal_branches.toml` file prevents worktree changes from being shown


Byron:
These changes, however, are retrieved by the frontend, and are non-empty. There are no stacks in the workspace.

### How to reproduce

Add the `dua-cli` repo as a project, and see the changes with `git-status`. They don't show in the UI though.

[dua-cli.zip](https://github.com/user-attachments/files/22253794/dua-cli.zip)

There is also no error anywhere.

PR #10202, opened 2025-09-10
Worktree changes: Dependecy errors toast and fallback

estib-vega:
- If the calculation of dependencies and assignments fails, fallback to assuming there are no dependencies
- Surface any errors regarding the calculation of dependencies or hunk assignments when fetching the worktree changes

Has to do (but doesn't fully take care of): https://github.com/gitbutlerapp/gitbutler/issues/10198


estib-vega:
cc @Byron

PR #10230, opened 2025-09-11
Migrate GitConfigService API calls to RTK Query

estib-vega:
This migrates the git config service to use RTK Query for all get/set/remove global config methods, leveraging Redux and query cache/invalidations for improved data consistency and less manual fetching. Includes new tags, endpoint injection, and integration into the app's layout for cache invalidation on navigation.


estib-vega:
cc @Byron 

This makes the data from the git config be cached (on a per item basis) for 30s, and it gets invalidated when moving away from pages

PR #10250, opened 2025-09-12
feat(ui): make codegen input scrollable with and actions overlay

SilkePilon:
Move to a borderless textarea wrapper and add maxRows so the input can grow up to 8 lines before scrolling. Constrain the input container with overflow hidden and a fixed bottom padding to make space for action buttons. Add a semi-opaque fade overlay at the bottom to visually mask the scroll boundary, and place the action toolbar above the fade using z-index so buttons remain accessible. Also add a subtle border, radius, and background to match other dialogs.

### DIFF
--- a/apps/desktop/src/components/codegen/CodegenInput.svelte
+++ b/apps/desktop/src/components/codegen/CodegenInput.svelte
@@ -58,9 +58,10 @@
 		autofocus
 		placeholder="What would you like to make..."
 		borderless
-		maxRows={8}
+		maxRows={10}
+		minRows={2}
 		padding={{
-			bottom: 96
+			bottom: 64
 		}}
 	/>
 
@@ -142,7 +143,7 @@
 		padding: 0;
 		overflow: hidden;
 		border: 1px solid var(--clr-border-2);
-		border-radius: var(--radius-ml);
+		border-radius: var(--radius-m);
 		background-color: var(--clr-bg-1);
 	}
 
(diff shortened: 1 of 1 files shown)


SilkePilon:
Preview:

https://github.com/user-attachments/assets/d10f8c12-4bb1-4512-8eb1-abe950ad215e

I also changed the border radius to be the same as the other ui elements, tho this can be reverted

krlvi:
Thank you very much for the fix.
Just a thought about the fade - Wouldn't it make more sense to fade the top of text when it's out of view instead of the bottom?
@PavelLaptev if you remember, we had this fade for the commit message box, but at some point we realized that it was incontinent when typing to not see the characters that you just typed

PavelLaptev:
> @PavelLaptev if you remember, we had this fade for the commit message box, but at some point we realized that it was incontinent when typing to not see the characters that you just typed

That was a different issue. It was always on the right and blocked the text.

PavelLaptev:
@SilkePilon I'll make some changes

PavelLaptev:
Changes:
- The inner radius has been reduced.
image

- The bottom padding has been reduced. The gradient scale appears less appealing when the scale is longer, but the bottom padding was too large.
image

PavelLaptev:
@SilkePilon I can merge it if you okay with these changes

SilkePilon:
👍 Looks good to me!

PavelLaptev:
@krlvi @SilkePilon there is an issue with the scrollbar

image
image

This could mean that we need to treat the input differently, avoiding the use of inner textarea padding. I’ll address this in the next PR. I see if I can save this nice gradient without adding extra JS logic

PR #10255, opened 2025-09-12
fix UTF-8 errors when diffing content

Byron:
Turns out this is easy to reproduce, and that the source of the issue is in `gitoxide`!

### Tasks

* [x] upgrade to the fixed version of `gitoxide`. https://github.com/gitbutlerapp/gitbutler/pull/10255 for content-diff fix
* ~~see if non-utf8 branch names work (on Linux, won't really work no MacOS)~~ - ignore, as it's unlikely to have ever worked, and will work after the rewrite.


Byron:
https://github.com/user-attachments/assets/eee6f717-9aeb-4fd9-ad62-b1262e9715a1

Byron:
@krlvi Maybe we get to bash more bugs, but this one here is definitely a reason for a patch. The issue was introduced in `gitoxide` quite recently (but now won't regress again).

PR #10258, opened 2025-09-12
Don't run parallel e2e tests

estib-vega:


Byron:
@estib-vega Is this because of the flakiness check being unreliable?

estib-vega:
> @estib-vega Is this because of the flakiness check being unreliable?

Yes. I have a strong suspicion that this was leaking. 

Also, even without the parallelization the tests themselves take 53s, which is more than fine

PR #10261, opened 2025-09-12
Make it clear that we unapply stacks if they conflict with the new one on apply

Byron:
Related to #10239

### DIFF
--- a/crates/gitbutler-branch-actions/src/branch_manager/branch_creation.rs
+++ b/crates/gitbutler-branch-actions/src/branch_manager/branch_creation.rs
@@ -271,7 +271,12 @@ impl BranchManager<'_> {
         self.ctx.add_branch_reference(&branch)?;
 
         match self.apply_branch(branch.id, perm, old_workspace, old_cwd) {
-            Ok((_, unapplied_stacks)) => Ok((branch.id, unapplied_stacks)),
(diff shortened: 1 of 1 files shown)


Byron:
@krlvi @PavelLaptev Is this something you'd find acceptable?
It turns the silent surprise about unapplied stacks into a loud and obvious one.

It's a temporary hack as this behaviour won't be available at first with the new implementation.

https://github.com/user-attachments/assets/1070c7d4-afc8-4567-8556-a84e4c566e2d

PavelLaptev:
@Byron If this is a rare and temporary case, we can handle it with a toast. However, if the error is likely to persist in the future, a modal would be a better choice. Modals feel more user-friendly and expected, while toast errors can come across as overly technical and unexpected.

Byron:
Actually… thanks for the hint!

I missed that the UI is already trying to do that, it's just that it is not working (anymore). There even is a test for it which probably does see it working, so it must be something else, a regression at least.

And… it turns out that I am 😣🤪 as apparently I made my change first, then tested it, instead of validating that the assumption is true that there is no popup.

Turns out there is one:

Screenshot 2025-09-12 at 19 08 11

And everything is fine - I took the issue at face value and assumed there is no note about that special behaviour at all.

PR #10284, opened 2025-09-15
Fix terminals opening on windows

Caleb-T-Owens:


QDaed:
thank god someone already do this for me , it really annoying that terminal keep opening when using codegen

Caleb-T-Owens:
@QDaed We're planning a release for this afternoon/evening which will include this fix :+1: Sorry for the trouble

PR #10302, opened 2025-09-15
Changes in branch show only the changes in the branch

estib-vega:
In the application, showing the changes for a stacked branch would display the changes inside that and all other branches below it in the stack.
This is kind of unexpected as we want to see the changes in that branch and not all changes up to the tip of the commit.

This makes it so that only the changes that will end up in that branch are shown


estib-vega:
Heads up: @Byron @krlvi

Byron:
Thank you! This is a genuine bugfix, and I can't believe that, erm, the original author, intended to ignore the branch 😅.

PR #10343, opened 2025-09-17
fix(ui): provide GitLab state in project settings modal

yuncai836:
Provide GitLab state and related services to the Project Settings modal so ForgeForm can persist and access platform-specific values.

- inject BASE_BRANCH_SERVICE, SECRET_SERVICE and GITLAB_CLIENT for modal scope
- derive repoInfo from base branch service to obtain repository metadata
- create GitLabState with secrets and repo info and provide it as GITLAB_STATE
- set gitLabClient to the newly created GitLabState in the modal context

This fixes scoped state handling by ensuring GitLab-related state and client are available inside the modal, enabling proper persistence and form behavior.


Byron:
Thanks a lot for the fix, this works!

https://github.com/user-attachments/assets/44b6aba2-3494-48c8-bcee-4a3ff04d56e1

Running `pnpm begood` should do the trick here.

estib-vega:
Thanks for this fix @yuncai836 !

The providing of the GitLab State seems to work, but sadly the correct fix needs a bit more restructuring.
The issue seems to be that the newly implemented modals broke injection consistency of the GitLab state, as they are not part of the project tree anymore (it seems). 

We need to provide the GitLab State from the deps.ts file, and then update the repo information form the project layout (just as we do with the GitHub client).

Let me take a look

estib-vega:
This is the preferred fix https://github.com/gitbutlerapp/gitbutler/pull/10362

Thanks again @yuncai836 for your effort though, and sorry for taking so long before fixing this issue

yuncai836:
> Thanks for this fix @yuncai836 !
> 
> The providing of the GitLab State seems to work, but sadly the correct fix needs a bit more restructuring. The issue seems to be that the newly implemented modals broke injection consistency of the GitLab state, as they are not part of the project tree anymore (it seems).
> 
> We need to provide the GitLab State from the deps.ts file, and then update the repo information form the project layout (just as we do with the GitHub client).
> 
> Let me take a look

Haha, today I took a closer look at the project, and you're absolutely right—the state management is more complex than I initially thought. I apologize, as I'm not particularly well-versed in frontend development, but I'm glad it didn’t cause any confusion in the project. GitButler is an excellent product that I use daily, and I hope to contribute to it in the future as well. So, what would you suggest we do with this PR ? @estib-vega

yuncai836:
> This is the preferred fix #10362
> 
> Thanks again @yuncai836 for your effort though, and sorry for taking so long before fixing this issue

Oh, I see the correct fix, I should close this pull request.

estib-vega:
> Haha, today I took a closer look at the project, and you're absolutely right—the state management is more complex than I initially thought. I apologize, as I'm not particularly well-versed in frontend development, but I'm glad it didn’t cause any confusion in the project. GitButler is an excellent product that I use daily, and I hope to contribute to it in the future as well. 

No need to apologize! Your contribution made it so much easier to pinpoint the root cause of the issue. 
You were just missing some context of our internal conventions which is normal.
This is truly appreciated!

> So, what would you suggest we do with this PR ? @estib-vega

We can close it. But keep opening others in the future :)

PR #10344, opened 2025-09-17
Fixes tooltip copy for pull-requests

schpet:
## 🧢 Changes

tooltip copy for pull requests

## ☕️ Reasoning

previously would incorrectly use ‘commit’ copy

### DIFF
--- a/apps/desktop/src/components/editor/MessageEditor.svelte
+++ b/apps/desktop/src/components/editor/MessageEditor.svelte
@@ -58,6 +58,7 @@
 		testId?: string;
 		forceSansFont?: boolean;
 		useRuler?: boolean;
+		messageType: 'commit' | 'pr';
 	}
 
 	let {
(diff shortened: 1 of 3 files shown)


apps/desktop/src/components/editor/MessageEditor.svelte
```diff
@@ -224,12 +226,17 @@
 		medium: 320
 	};
 
+	const GENERATE_MESSAGES: Record<typeof messageType, string> = {
```
Byron, line 229: That's pretty neat as it probably forces you to add a message if a new `messageType` is added.

Byron (approved):

This looks good to me and I dare to merge it myself even as I could reproduce the issue on nightly, but not anymore with this PR applied.

Screenshot 2025-09-17 at 13 56 02

Byron:
CC @estib-vega just in case.

apps/desktop/src/components/editor/MessageEditor.svelte
schpet: yeah exactly. similarly switch can be exhaustive (e.g. https://stackoverflow.com/a/75217377) 

not as nice as rust enums but it's something

PR #10357, opened 2025-09-18
Improve defaults after settings.json was deleted + recreated

### LINKED ISSUE #10356: Settings are incorrect on 'fresh' installations/after settings.json was deleted
When first opening the application *without* `settings.json` file, it doesn't come up as expected, but probably I am missing something regarding how it works as well. Maybe it's due to not going through the onboarding if the App otherwise was started already.

Image

Note the lack of "new workspace backend", which should be enabled.

Here is the file after opening the application:

```json
{
  "telemetry": {
    "appDistinctId": "user_<redacted>"
  }
}
```

This goes along with a user reporting that persistence doesn't seem to work, which I *cannot* reproduce.

However, the app-settings really are returned like this by the backend:

```ron
AppSettings {
    context_lines: 3,
    onboarding_complete: false,
    telemetry: TelemetrySettings {
        app_metrics_enabled: true,
        app_error_reporting_enabled: true,
        app_non_anon_metrics_enabled: false,
        app_distinct_id: Some(
            "<redacted>",
        ),
    },
    github_oauth_app: GitHubOAuthAppSettings {
        oauth_client_id: "<redacted>",
    },
    feature_flags: FeatureFlags {
        ws3: false,
        cv3: false,
        undo: false,
        actions: false,
        butbot: false,
        rules: false,
        single_branch: false,
    },
    extra_csp: ExtraCsp {
        hosts: [],
    },
    fetch: Fetch {
        auto_fetch_interval_minutes: 15,
    },
    claude: Claude {
        executable: "claude",
        notify_on_completion: false,
        notify_on_permission_request: false,
        dangerously_allow_all_permissions: false,
        auto_commit_after_completion: true,
        use_configured_model: false,
    },
}
```

So I think the problem is that the default settings are used for initialisation and merging, but certain features aren't enabled there even though they should in this case.

Byron:
Fixes #10356.

When `settings.json` gets deleted, the user doesn't get the onboarding step
again to initialise the most important values.

This leads to defaults not being respected, while enabling telemetry by default.

### DIFF
--- a/crates/but-settings/assets/defaults.jsonc
+++ b/crates/but-settings/assets/defaults.jsonc
@@ -19,11 +19,11 @@
 	},
 	"featureFlags": {
 		// Deprecated, was used for the new version of the UI.
-		"v3": false,
+		"v3": true,
 		// Enables the v3 safe checkout.
 		"cv3": false,
 		/// Enable the usage of V3 workspace APIs.
-		"ws3": false,
+		"ws3": true,
 		/// Enable undo/redo support.
 		"undo": false,
 		/// Enable the usage of GitButler Acitions.
--- a/crates/gitbutler-branch-actions/tests/reorder.rs
+++ b/crates/gitbutler-branch-actions/tests/reorder.rs
@@ -469,7 +469,9 @@ fn file(ctx: &CommandContext, commit_id: gix::ObjectId) -> String {
 }
 
 fn command_ctx(name: &str) -> Result<(CommandContext, TempDir)> {
-    gitbutler_testsupport::writable::fixture("reorder.sh", name)
+    gitbutler_testsupport::writable::fixture_with_settings("reorder.sh", name, |settings| {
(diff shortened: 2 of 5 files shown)


Byron:
CC @krlvi - I think the recreation of `settings.json` without onboarding needs special consideration, as the config comes up incorrectly unless the defaults are adjusted, or the app onboarding is restarted then. The latter would probably be preferred as it deals with telemetry as well, but I will bring up another issue for that.

crates/but-settings/assets/defaults.jsonc
```diff
@@ -19,11 +19,11 @@
 	},
 	"featureFlags": {
 		// Deprecated, was used for the new version of the UI.
-		"v3": false,
```
krlvi, line 22: Oh I see - when we made these on by default, I updated the parsing default but never changed this!

krlvi:
Good catch!

PR #10374, opened 2025-09-18
Move branches round and around and around and around

estib-vega:
Ability to move branch on top of each other in order to stack them


apps/desktop/src/lib/dragging/draggable.ts
```diff
@@ -241,6 +243,26 @@ function setupDragHandlers(
...
 
+/////////////////////////////
+//// BRANCH DRAGGABLE ///////
+/////////////////////////////
+
+function createBranchElement(label: string | undefined): HTMLDivElement {
+	const cardEl = createElement('div', ['draggable-branch-card', 'text-15', 'text-bold'], label);
+
+	return cardEl;
+}
```
estib-vega, line 254: @PavelLaptev  This is creating the basic branch header drag chip

apps/desktop/src/components/LineOverlay.svelte
```diff
@@ -51,6 +54,12 @@
 				background-color: var(--clr-theme-pop-element);
 			}
 		}
+
+		&:not(.hovered).advertize {
+			& .indicator {
+				background-color: var(--clr-theme-pop-soft-hover);
+			}
+		}
```
estib-vega, line 62: @PavelLaptev I also created this optional class that just lights up, so that it's easier to discover where you can drop a branch. But probably this could be prettier

krlvi:
oh exciting new func

PR #10377, opened 2025-09-18
fix(config): use correct config path

TheGB0077:
## ☕️ Reasoning

It breaks reading settings on Linux

## 🎫 Affected issues

Fixes: https://discord.com/channels/1060193121130000425/1418270491046838425


Byron (approved):
Thanks so much for figuring this out, this is fantastic!

I think the underlying problem was that tauri had its own view on what the path should look like, so changes were stored in one place, but reads are done from another, strangely enough (I wasn't aware of this).

In any case, it's probably better to leave the path in `but-path` as is to match what's documented. But then again, as writes went into the `config_dir()`, it would certainly be better to now also start reading from it so people don't lose their config. But then again, due to this bug they could never use their configuration anyway, so ultimately it's best to keep everything where it was supposed to be.

Maybe you could double-check that this works for you as well.

CC @krlvi in case I am getting something wrong here.

krlvi:
Hmm actually i am a bit confused why this works.
`app_config_dir` gives us
```rust

pub fn app_config_dir() -> anyhow::Result<PathBuf> {
    if let Ok(test_dir) = std::env::var("E2E_TEST_APP_DATA_DIR") {
        return Ok(PathBuf::from(test_dir).join("gitbutler"));
    }
    dirs::data_dir()
        .ok_or(anyhow::anyhow!("Could not get app data dir"))
        .map(|dir| dir.join("gitbutler"))
}
```
So it will be gitbutler nested twice

krlvi:
Actually i think this will break something.

crates/gitbutler-tauri/src/main.rs
```diff
@@ -37,7 +37,7 @@ fn main() {
     let mut tauri_context = generate_context!();
     gitbutler_secret::secret::set_application_namespace(&tauri_context.config().identifier);
 
-    let config_dir = dirs::config_dir()
+    let config_dir = but_path::app_config_dir()
         .expect("missing config dir")
         .join("gitbutler");
```
krlvi, line 42: I think this join should be skipped since the but_path::app_config_dir() adds the gitbutler folder.
@Byron

Byron: I broke it :/, and how could I miss this :/.

Thanks for the quick fix, see #10379 for the follow-up.

PR #10379, opened 2025-09-18
fix(settings): Enhance settings file observability

TheGB0077:
It was pointed out that the changes could possibly be breaking, but such is not the case. 

The persistence flow had been writing to ./config/gitbutler through dirs::config_dir() all along, so in every Linux distribution the settings are currently stored on that folder. Standardizing to dirs::data_dir() is breaking and also an antipattern for Linux expectations.

Other platforms remain unaffected, as dirs::config_dir() and dirs::data_dir() resolve to the same file system address. See [config_dir](https://docs.rs/dirs/6.0.0/dirs/fn.config_dir.html) and [data_dir](https://docs.rs/dirs/6.0.0/dirs/fn.data_dir.html)

## Further Findings:

ModifyKind (in the context of Linux at the very least) doesn't trigger changes when the settings file is "updated", that's because a temp file is created and then proceed to replace the original settings.json file.

When settings.json is removed, the RemoveKind Event happens and then the watcher must be updated to watch the newly created settings.json


crates/but-path/src/lib.rs
```diff
@@ -13,7 +13,7 @@ pub fn app_config_dir() -> anyhow::Result<PathBuf> {
     if let Ok(test_dir) = std::env::var("E2E_TEST_APP_DATA_DIR") {
         return Ok(PathBuf::from(test_dir).join("gitbutler"));
     }
-    dirs::data_dir()
+    dirs::config_dir()
```
Byron, line 16: I see now how my logic was wrong, thanks for the fix, again.

crates/but-settings/src/watch.rs
```diff
@@ -130,7 +132,20 @@ impl AppSettingsWithDiskSync {
                             send_event(update)?;
                         }
                     }
-
+                    Ok(Ok(Event {
+                        kind: notify::event::EventKind::Remove(RemoveKind::File),
+                        ..
+                    })) if cfg!(target_os = "linux") => {
```
Byron, line 141: I think it's a safe choice to limit this to run on Linux for now, while adding that I don't think we get our event-processing right here.

In theory, we should always be handling all related events, if they occur or not, with a reasonable amount of debouncing sprinkled in, so this wouldn't have to be linux only, but I am slightly anxious about double-loads on MacOS or other platforms, maybe even Linux on different filesystems or implementations.

Byron (approved):
Thanks a lot for the second take!
I am happy we went through the additional iteration as it revealed the watch-related problem.

Please note that I didn't validate this on Linux, but assume that it's good because it works for you.

Maybe it happened all the time, and it was expected that we don't respond to removals?
Usually the removal/recreation happens when the application itself changes the settings, so probably it's intentional not to respond to them.

So the underlying problem always was that the watcher lost track on Linux, not necessarily that it didn't respond to removals with an event.

@krlvi I think on linux, we might only want to re-watch the file, instead of sending an update event. What do you think?

TheGB0077:
> So the underlying problem always was that the watcher lost track on Linux, not necessarily that it didn't respond to removals with an event.

It was a dual issue! After the path mismatch was fixed, I noticed that the toggles remained unsuccessful at tracking the file past an update and considering the UI update logic was tied to the file event watching, I started debugging from there. 

I suppose that other OSes might handle the update more gracefully with an Update event, but what I ultimately found was the removal of the file which, unaccounted for, made settings changes only refresh upon restarting the app.

crates/but-settings/src/watch.rs
TheGB0077: While debugging this code path I did not notice some surge of events triggered, I would tend to believe the underlying code hooks into inotify syscalls to manage this, so it should be a negligible change on Linux.

TheGB0077: I set the OS target there under the belief that the same errors cannot be observed on Windows or Mac. At least on Mac people that discussed with me on Discord did not seem to have those issues while changing settings, so I guess it is working as expected.

krlvi (approved):
Looks reasonable! Thank you fox fixing

krlvi:
Regarding re-watching I think it makes sense indeed

PR #10382, opened 2025-09-19
Add the `head_info` endpoint

Byron:
With that the frontend can consume all relevant workspace data just once, instead of once for `stacks()` and once per stack (i.e. 1 + N vs 1).

### Tasks

* [x] special UI copy of the existing type just for good measure, also because some data needs to be duplicated and processed, like branch names.
* [x] split branch names into bytes and short-names
* [x] split remote names into remote name and short name.
* [x] filter entrypoint - remove everything not reachable by entrypoint


Byron:
The new `head_info(project_id)` endpoint is now available, in a very first version on which I think we will iterate 🙌.

PR #10387, opened 2025-09-19
Reword the "git hooks" settings description

mtsgrd:
It was saying only commit hooks run, while we have commit-msg and 
pre-push as well.


apps/desktop/src/components/DetailsForm.svelte
```diff
@@ -52,10 +52,11 @@
 
 <SectionCard labelFor="runHooks" orientation="row">
 	{#snippet title()}
-		Run commit hooks
+		Run git hooks
```
Byron:
```suggestion
		Run Git hooks
```

PR #10399, opened 2025-09-22
Do not persist credentials during checkout to prevent exfiltration or accidental logging

Byron:
Based on this commit in Gitoxide:

https://github.com/GitoxideLabs/gitoxide/pull/2187/commits/a235ac803598df05cefb2f3eff46598e42a16b96

### Tasks

* [x] add directive to all non-publish workflows
* [x] prevent future checkouts with persisted credentials in everything but publish.yml

Note that the exception of `publish.yml` is merely done out of fear of testing this, as it's possible that
this workflow relies on persisted credentials.

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -13,6 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
         with:
+          persist-credentials: false
           repository: 'gitbutlerapp/gitbutler'
       - uses: actions/labeler@v6
         with:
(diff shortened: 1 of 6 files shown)


Byron:
CC @krlvi This PR is a clone of [a similar one in `gitoxide`](https://github.com/GitoxideLabs/gitoxide/pull/2187), and this suggestion comes directly from the GitHub security team so I thought I'd port it here.

`publish.yaml` is left out, but if you think it's not relying on persisted credentials we could disallow doing so there as well. I just didn't want to risk breaking it, especially knowing that this workflow is only run by trusted parties.

krlvi:
Nice one, thank you

PR #10404, opened 2025-09-22
Don't strip trailing newlines but keep them in the diff.

Byron:
This *should* work as we transform to a normalised form before diffing. However, I think the reason for the change ultimately was Windows, where crlf/lf differences might make diffs light up unnecessarily.

Maybe there will be more polish in that regard. Unfortunately it's also not clear anymore why stripping was deemed so very necessary in the first place.

### Tasks

* [ ] "Design" around it for now
* [ ] frontend needs to differentiate between files and no diff being shown - can infer
    - [ ] it seems that "generate" for the commit message will think the file was deleted. I thought it should think that 'nothing changed'.

### Future Tasks

* [ ] send newlines to the frontend at least so it can determine what changed
* [ ] somehow strip newlines in the frontend, but visualise changes somehow?

### DIFF
--- a/crates/but-claude/src/lib.rs
+++ b/crates/but-claude/src/lib.rs
@@ -109,6 +109,7 @@ pub struct UserInput {
     /// The user message
     pub message: String,
     /// Optional attached file references
+    #[serde(default)]
     pub attachments: Option<Vec<PromptAttachment>>,
 }
 
@@ -235,4 +236,46 @@ pub async fn send_claude_message(
         payload: json!(message),
     });
     Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_user_input_backwards_compatibility_without_attachments() {
+        let json_without_attachments = r#"{
+            "message": "foo"
+        }"#;
+
+        let user_input: UserInput = serde_json::from_str(json_without_attachments)
+            .expect("Failed to deserialize UserInput without attachments field");
+
+        assert_eq!(user_input.message, "foo");
+        assert!(
+            user_input.attachments.is_none(),
+            "Attachments should default to None for backwards compatibility"
+        );
+    }
+
+    #[test]
+    fn test_user_input_with_attachments() {
+        let json_with_attachments = r#"{
+            "message": "bar",
+            "attachments": [
+                {
+                    "type": "file",
+                    "path": "src/main.rs"
+                }
+            ]
+        }"#;
+
+        let user_input: UserInput = serde_json::from_str(json_with_attachments)
+            .expect("Failed to deserialize UserInput with attachments field");
+
+        assert_eq!(user_input.message, "bar");
+        assert!(user_input.attachments.is_some());
+        let attachments = user_input.attachments.unwrap();
+        assert_eq!(attachments.len(), 1);
+    }
 }


Byron:
With this fix, newlines are relevant in diffs.

https://github.com/user-attachments/assets/34f195e8-cad2-4c90-b0ab-483fdd22bb34

They could benefit from special rendering in the frontend as well.

It's unclear if there are corner-cases on Windows that were bypassed previously, maybe stripping newlines was never necessary though or it stopped being necessary over time due to improvements in other places.

CC @krlvi as this might need some testing on Windows. We don't seem to have a breaking test though so but I am not sure they cover everything.

Byron:
Here is my comment from Discord for a second life:

----

Can we fix the 'line change diff shows up empty' issue in post :sweat_smile:? 
Right now, changing just the `\n` at the end of the file by adding or removing it creates empty diffs. Fixing this is more involved and probably won't be possible to do nicely without more frontend awareness, and it's a rabbit hole I'd like to defer going down to focus on `apply/unapply`.
So here is my MVP improvement for this: The UI detects if it gets empty hunks, and instead shows something along the lines of 'line separators changed', or 'newline was added or removed from the file'. The latter is a very common reason for empty diffs.
CC @mtsgrd 

This picture would have to be changed to something else, to buy time until we can tackle the actual problem properly.

Screenshot_2025-09-23_at_04 43 54

mtsgrd:
Ok, if it's a deep rabbit hole then let's try and manage this in the front end. I won't be able to get to this in the next few days, but I can try and make it so if there are no flags changes, and the file is modified, then we switch the message to something about newlines.

Byron:
Thank you, appreciated!
I just want to prioritise 'apply/unapply' now and be more aggressive in my pursuit - it's hanging around for too long and I want something to be usable (even if not complete) ASAP.

mtsgrd:
@Byron could you confirm what effect this change has and if it should be merged?

estib-vega:
> @Byron could you confirm what effect this change has and if it should be merged?

@Byron and I talked in person about this change. There are some little changes needed in the FE before this can be merged

Byron:
Commit original and place changed in the worktree to see the effect of newlines.

[original.zip](https://github.com/user-attachments/files/23363140/original.zip)
[changed.zip](https://github.com/user-attachments/files/23363141/changed.zip)


# task

* [x] why does the backend not pickup the whole line?

Byron:
Here is what's happening:

* the version in Git doesn't have a trailing newline at the end of the file
* the worktree version has a trailing newline (but we can't show it in the UI yet as our diffs don't include newlines)

Screenshot 2025-11-10 at 10 10 34

On disk, we clearly have a trailing newline.

Screenshot 2025-11-10 at 10 15 14

After committing in GitButler, the worktree version is unchanged.

Screenshot 2025-11-10 at 10 20 34

But the file in Git has all the changes, but the trailing newline change at the end.

Screenshot 2025-11-10 at 10 22 05

### Conclusion

The problem is that the frontend doesn't see the changes, and thus doesn't pass them to the backend.

The solution is to *not send hunks* if *all hunks are selected*.

[small.zip](https://github.com/user-attachments/files/23450256/small.zip)

This can be implemented as hack as well, that keeps a flag to say: user didn't edit hunks, and turns that off in the moment any hunk is edited. Then that's a fix already for most cases.

CC @estib-vega -  I think this is a high-value problem as it's a small thing that's really puzzling to everyone involved (even me 😅).

PR #10430, opened 2025-09-24
Use focus manager for navigation context menus

mtsgrd:
Removes custom keyboard navigation for context menus, and replaces it 
with the new focus manager functionality. Note that ContextMenuItem 
contains two focusables, one for navigation and one to support f-keys.

### DIFF
--- a/packages/ui/src/lib/components/ContextMenu.svelte
+++ b/packages/ui/src/lib/components/ContextMenu.svelte
@@ -329,13 +329,6 @@
 		return isVisible;
 	}
 
-	function handleKeyNavigation(e: KeyboardEvent) {
-		if (e.key === 'Escape') {
-			e.preventDefault();
-			close();
-		}
-	}
-
 	$effect(() => {
 		if (!menuContainer) return;
 		const config = { attributes: false, childList: true, subtree: true };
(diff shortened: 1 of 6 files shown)


mtsgrd:
https://github.com/user-attachments/assets/d0eba3e0-186e-498a-a332-c76912813641

@PavelLaptev could you have a look at this and let me know what you think? *edit: still need to fix tab for moving to next/prev item.

PavelLaptev:
@mtsgrd thanks for tagging. Looks nice, but let’s keep only one outline instead of two.

image

I think the inner one should stay since we use the arrow to navigate.

mtsgrd:
To me it looks like I included a change with `outline: none` to exactly prevent the double border. Could you be needing to recompile?

PavelLaptev:
> To me it looks like I included a change with `outline: none` to exactly prevent the double border. Could you be needing to recompile?

@mtsgrd sorry, I was looking at the video you attached.

PavelLaptev:
@mtsgrd how can I trigger the context menu using the keyboard? I tried the "Enter" key but it didn’t open it.

https://github.com/user-attachments/assets/9a338ccf-e4b7-4229-a10c-c05b7ec03aaa

mtsgrd:
I'll investigate!

mtsgrd:
Ok, should be fixed. I added enter key handling to KebabButton.svelte.

PavelLaptev:
Thanks! Looks like it works for branches but not for commits

https://github.com/user-attachments/assets/f9af3e6f-7227-4850-a990-99845bbadf98

mtsgrd:
@PavelLaptev just fyi, I still intend to land something like this but I'm currently trying to work through an inconsistency between Chrome and Safari wrt to `:focus-visible`. It's proving to be a bit tricky.

PavelLaptev:
> @PavelLaptev just fyi, I still intend to land something like this but I'm currently trying to work through an inconsistency between Chrome and Safari wrt to `:focus-visible`. It's proving to be a bit tricky.

Gotcha! Let me know if you need me to test anything or if you need any help :-)

kmk142789 (approved):

PR #10449, opened 2025-09-26
Ensure that the target branch isn't showing up as workspace segment.

Byron:
Unless they are specifically added to the workspace metadata.

Related to #10415.

### Tasks

* [x] reproduce
* [x] fix


Byron:
This fix isn't nice and I think there should be a way to fix it at its core, which also auto-resolves the issue we are seeing here, left with a TODO.

PR #10467, opened 2025-09-27
More defensively build the format string in case the workspace commit isn't managed (#10466)

### LINKED ISSUE #10466: Bug: At Least One User Commit On Top
### Version

0.16.7

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

I get this occasionally in my project and there is no way to clear it or address it.  I'm not sure what to do with this.  If I can't get this to go away, occasionally I'll have to remove my project and re-add my project, but then I have to re-connect my project to github which is annoying.

Image

1. I'm not exactly sure how this happened
2. I was just using only gitbutler for claude based development, I didn't make any changes anywhere else
3. I occasionally get one of these errors that I can't dismiss about a random commit on top
4. Sometimes I get this after pulling from github
5. Sometimes I get this if I make a change from cursor outside of gitbutler 
6. But this time I didn't pull from github or use cursor

Also, I can't install the CLI from the settings, it straight up doesn't work, like nothing happens when I click the button.  I attempted to create a symlink, so maybe that's working.

1. I'd love to be able to dismiss the error. 
2. I'm not sure what the significance of the error is, does this prohibit me from anything?
3. What are the steps to remediate the situation?  
4. Is remediation necessary?

```shell

```

Byron:
Note that it's still unclear how the graph workspace can report that it's not pointed to by the managed workspace ref, yet the first commit it finds seems to be the managed workspace commit.

This really shouldn't be possible.

Fixes #10466 .

### DIFF
--- a/crates/but-workspace/src/lib.rs
+++ b/crates/but-workspace/src/lib.rs
@@ -221,8 +221,8 @@ pub struct RefInfo {
 #[derive(Debug, Clone)]
 pub struct AncestorWorkspaceCommit {
     /// The commits along the first parent that are between the managed workspace reference and the managed workspace commit.
-    /// The vec is never empty.
-    pub commits_outside: Vec<crate::ref_info::Commit>,
+    /// The vec *should* not be empty, but it can be empty in practice for reasons yet to be discovered.
+    pub commits_outside: Vec<ref_info::Commit>,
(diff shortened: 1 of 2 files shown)


notorioustig:
Thanks for the communication, I'll let you know if it happens to me again!

PR #10477, opened 2025-09-28
Implement but branch new stacked branch

krlvi:
When doing `but branch new <branchname>`
you can also pass `--anchor <branch|commit>` or `-a <branch|commit>` to 
create a stacked branch

### DIFF
--- a/crates/but/src/branch/mod.rs
+++ b/crates/but/src/branch/mod.rs
@@ -1,5 +1,7 @@
 use std::path::Path;
 
+use but_settings::AppSettings;
+use gitbutler_command_context::CommandContext;
 use gitbutler_project::Project;
 
 #[derive(Debug, clap::Parser)]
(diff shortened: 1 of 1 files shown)


crates/but/src/branch/mod.rs
```diff
@@ -14,21 +16,88 @@ pub enum Subcommands {
...
+                let request = but_api::stack::create_reference::Request {
+                    new_name: branch_name.clone(),
+                    anchor,
+                };
+                but_api::stack::create_reference(project.id, request)?;
+                println!("Created branch {branch_name}");
+                Ok(())
+            } else {
+                // Create an independent branch
+                let req = gitbutler_branch::BranchCreateRequest {
```
Byron, line 92: That's interesting, I thought `but_api::create_branch()` does the same, but using the new API.

https://github.com/gitbutlerapp/gitbutler/blob/29a0d937602e6f2523011a861a1f5fafbf3f3a69/crates/but-api/src/commands/stack.rs#L96

That's what the UI calls.

krlvi: ~Nice catch. This commit only had the `but_api::virtual_branches::create_virtual_branch` so here I only added the create reference. But this should be updated indeed~

krlvi: Hmm the frontend still does the `create_virtual_branch` command so this just does the same
Maybe it should be updated in both places....
image

Byron: Ah, right, I was confused. Somehow I thought there would only be one function for that, but in fact there are two, this one and the one I mentioned, and *both* are used by the UI for some reason :). Both support `ws3` and will call the respective new function, so there is no issue with it - probably this one has better 

This is probably where one would apply another fix then.

I also think what I was responding to here is to generally avoid calling anything with `virtual` in it 😅.

Byron: I opened https://github.com/gitbutlerapp/gitbutler/pull/10492 for this, with the kind request to validate it as I didn't do that

PR #10493, opened 2025-09-29
Handle non-UTF8 encodings better when preparing unified diffs for the frontend (#10475)

### LINKED ISSUE #10475: Latin1 charset rendered incorrectly in diff.
### Version

0.16.7

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

GitButler seems to brutally assume an UTF-8 encoding of all text files, with no option to do otherwise.

Image

Either a per file or per repo setting would help a lot to remedy this. Or to have actual auto detection.

1. Create a non-utf8 encoded file.
2. Watch gitButler fail to parse it correctly.

1. GitButler should display characters correctly. It should have an option to set character encoding when GitButler fails to detect it correctly.

```shell

```

Byron:
Fixes #10475.

### Tasks

* [x] test for reproduction
* [x] fix based on #10481

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -985,6 +985,7 @@ dependencies = [
  "anyhow",
  "bstr",
  "but-testsupport",
+ "chardetng",
  "gitbutler-error",
  "gitbutler-serde",
  "gix",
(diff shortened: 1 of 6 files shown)


Byron:
Before:

Screenshot 2025-09-29 at 09 23 40

After:

Screenshot 2025-09-29 at 09 23 18


The hardest part was to get a sample with the expected encoding 😅. Turns out my LLM generated sample that was supposed to use accents from a certain encoding did use UTF8 accents that didn't encode cleanly in the desired codepage. Thus, `?` was added by the editor, for a file that genuinely contained these question marks, but I kept thinking that the decoding didn't work as it added replacement characters (even though it claimed it didn't).

PR #10517, opened 2025-10-01
feat(ui): format tokens and cost using locale-aware formatter

nshcr:
Using formatted numbers can make the token consumption and dollar cost much easier to read.

Before and after screenshots:

Screenshot 2025-10-01 at 10 12 08

### DIFF
--- a/apps/desktop/src/components/codegen/CodegenSidebarEntry.svelte
+++ b/apps/desktop/src/components/codegen/CodegenSidebarEntry.svelte
@@ -36,6 +36,16 @@
 	}: Props = $props();
 
 	let isOpen = $state(false);
+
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks so much!

I take it that the left version is 'after', and if so it's a massive difference indeed.

nshcr:
> I take it that the left version is 'after', and if so it's a massive difference indeed.

Actually, the part on the right side of the screenshot is the "after".

To demonstrate the thousand separators, I added a really large random number while debugging in the dev version…
Surely no one would burn through thousands of dollars in a Claude Code session, right?

Caleb-T-Owens:
Thanks for the contribution. This looks great!

PR #10519, opened 2025-10-01
Improve the 'worksapce branch is advanced by the user' Git command for a quick fix.

Byron:


Byron:
I think it's better, because it's simpler, while assuring that the user has no problems with stashes and what-not.

https://github.com/user-attachments/assets/e6b18995-c012-4faa-8342-215cd95604f7

PR #10528, opened 2025-10-02
use the stable toolchain to safe space on CI

Byron:
It seems we don't use any nightly features anymore, and dealing with an extra toolchain costs space and time.
Also it adds friction to contributors who need to download a new toolchain when trying to do anything with the project.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2664,6 +2664,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "file-id"
+version = "0.2.3"
+source = "git+https://github.com/Byron/notify#1ef23915e81b9bd780c6cf76c9a649d11bf60d57"
+dependencies = [
+ "windows-sys 0.60.2",
+]
+
+[[package]]
 name = "filetime"
 version = "0.2.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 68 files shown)


Byron:
@krlvi if this PR goes through, I will trigger nightly build right away to assure this doesn't break anything.
If it does, I will be on it.
It's notable that CI can be faster now, also because caches are smaller due to the new settings.

We can even consider to disable debug information on CI for even more space savings.

Byron:
It seems after #10522 CI went down [to ~32m](https://github.com/gitbutlerapp/gitbutler/actions/runs/18184891491/usage) without cache [from ~37m](https://github.com/gitbutlerapp/gitbutler/actions/runs/18186398256/usage).

With cache *and* with stable it's now [at 20m](https://github.com/gitbutlerapp/gitbutler/actions/runs/18187496762/usage), and *before stable and with more debug info* it's [at ~23.5m](https://github.com/gitbutlerapp/gitbutler/actions/runs/18170655686/usage).

Byron:
I tested the latest nightly, built with the changes regarding `file-id`, and it seems to work fine.

I was able to fetch a repo as usual, and [it looks like](https://github.com/gitbutlerapp/gitbutler/blob/eb053ceddc56e6c834bd5af4e38a39b2198c4737/crates/gitbutler-git/src/repository.rs#L112-L115) the stat code always runs and definitely doesn't fail.

So I think it's fine to keep it.

PR #10536, opened 2025-10-02
Pin the stable version to whatever the latest CI image contains!

Byron:
Otherwise builds will once again take longer (but we also can keep using our caches longer).

Note that the image documentation says it's Rust 1.89, but the Rust cache says it's
1.90. Let's trust that then.


Byron:
@krlvi I'd so love to be able to say "stable" for CI and something specific locally, but … actually you can: `RUSTUP_TOOLCHAIN`

PR #10537, opened 2025-10-02
Setting polling frequency

### LINKED ISSUE #3257: Allow user to adjust fetching frequency
Currently, we're fetching every 15 minutes. But It would be nice to allow users to adjust it and fetch earlier or later.
I'm thinking about frequency presents like `1 min`, `5 min`, `10 min` or `15 min`

Caleb-T-Owens:
Closes https://github.com/gitbutlerapp/gitbutler/issues/3257


krlvi:
This feels like something that would ideally be expressed on a per-project basis. And this value is not consumed anywhere on the rust side, so I am curious, why did you choose put it here rather than a frontend value?

PR #10546, opened 2025-10-04
Fix the link provided in the GitButler workspace commit.

Byron:
Maybe we could also have a redirect from

https://docs.gitbutler.com/features/virtual-branches/integration-branch

to

https://docs.gitbutler.com/features/branch-management/integration-branch


Byron:
@estib-vega @PavelLaptev Would it be good to have a redirect from the old to the new link?
If it's a hassle, it's probably not worth it though as all these workspace commits get rewritten all the time, so the fix in this PR should do the trick.
Maybe a redirect would be a good safety measure in case the old link is used in more places.

PavelLaptev:
@Byron good catch! Let's merge this one. And I'll add a redirect thingy to the docs

PR #10562, opened 2025-10-07
feat: allow COLOR_OVERRIDE env to force colored output rendering

krlvi:


crates/but/src/main.rs
```diff
@@ -25,6 +25,14 @@ mod status;
 
 #[tokio::main]
 async fn main() -> Result<()> {
+    if let Ok(color_override) = std::env::var("COLOR_OVERRIDE") {
+        match color_override.to_lowercase().as_str() {
+            "true" | "1" | "on" | "yes" => colored::control::set_override(true),
+            "false" | "0" | "off" | "no" => colored::control::set_override(false),
+            _ => {}
+        }
+    }
```
Byron, line 34: The last time I delved into this a bit I found that there are some common conventions for this.
Fortunately, there already seems to be an implementation for it, right there in `colored`: https://docs.rs/colored/latest/colored/control/struct.ShouldColorize.html#method.from_env . Maybe that's better, particularly in `but` which should be used by everyone?

krlvi: I was not aware of this. Changing now, thank you for flagging

PR #10567, opened 2025-10-07
Prevent description reset when editing commit message

mtsgrd:
The value would reset when toggling between inline and modal view.

### DIFF
--- a/apps/desktop/src/components/CommitMessageEditor.svelte
+++ b/apps/desktop/src/components/CommitMessageEditor.svelte
@@ -199,8 +199,11 @@
 		{aiIsLoading}
 		{suggestionsHandler}
 		useRuler={uiState.global.useRuler.current}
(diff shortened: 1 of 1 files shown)


apps/desktop/src/components/CommitMessageEditor.svelte
```diff
@@ -199,8 +199,11 @@
 		{aiIsLoading}
 		{suggestionsHandler}
 		useRuler={uiState.global.useRuler.current}
-		onChange={(text: string) => {
-			onChange?.({ description: text });
+		onChange={(value) => {
+			// Updating the description prevents resetting the editor value
+			// when toggling between inline and modal view.
+			description = value;
```
estib-vega, line 205: Nicee. I though this would have broken the ability to cancel and reset the values of the message editor.

Works as expected!

PR #10572, opened 2025-10-07
chore: ignore additional build artifact "but-*"

nshcr:
Screenshot 2025-10-08 at 05 43 32

I believe these two production build files should be ignored.

They appear when building a production version locally, and other similar binaries in the same directory are already ignored.


Byron:
Thanks a lot, much appreciated!

PR #10574, opened 2025-10-07
feat: add UI settings to support native title bar style

nshcr:
Screenshot 2025-10-08 at 05 42 48

This PR adds a toggle for enabling the native title bar style in the UI settings.
I added it as a hidden option in the internal JSON config instead of exposing it in the UI, since enabling it requires restarting the app — and I think using the native title bar somewhat breaks the app's overall design.
It's better not to expose it until the UI fully supports it.

Enabling the native title bar has a few benefits (mostly due to Tauri's limited native feature support), such as better drag behavior (three finger drag support) and more consistent double-click zoom behavior.

Tauri currently binds double-click on `data-tauri-drag-region` to "maximize", but on macOS this doesn't always match system behavior depending on user settings, which makes GitButler behave inconsistently compared to other apps.

https://github.com/tauri-apps/tauri/blob/19fb6f7cb0d702cb2f25f6f2d1e11014d9dada5d/crates/tauri/src/window/scripts/drag.js#L46-L48

Part of this PR depends on my previous one (https://github.com/gitbutlerapp/gitbutler/pull/10573). If that one can't be merged, I can modify this PR to make it standalone.
If you think adding this setting makes sense but I've missed something, please let me know how I can improve it. If you don't think this setting should exist, feel free to close it.

### DIFF
--- a/apps/desktop/src/components/ChromeHeader.svelte
+++ b/apps/desktop/src/components/ChromeHeader.svelte
@@ -45,6 +45,7 @@
 	const isWorkspace = $derived(isWorkspacePath());
 	const canUseActions = $derived($settingsStore?.featureFlags.actions ?? false);
 	const singleBranchMode = $derived($settingsStore?.featureFlags.singleBranch ?? false);
+	const useCustomTitleBar = $derived(!($settingsStore?.ui.useNativeTitleBar ?? false));
 	const backend = inject(BACKEND);
 
 	const mode = $derived(modeService.mode({ projectId }));
@@ -112,12 +113,12 @@
 <div
 	class="chrome-header"
 	class:mac={backend.platformName === 'macos'}
-	data-tauri-drag-region
+	data-tauri-drag-region={useCustomTitleBar}
 	class:single-branch={singleBranchMode}
 	use:focusable
 >
-	<div class="chrome-left" data-tauri-drag-region>
-		<div class="chrome-left-buttons" class:macos={backend.platformName === 'macos'}>
+	<div class="chrome-left" data-tauri-drag-region={useCustomTitleBar}>
+		<div class="chrome-left-buttons" class:has-traffic-lights={useCustomTitleBar}>
 			<SyncButton {projectId} disabled={actionsDisabled} />
 
 			{#if isHasUpstreamCommits}
@@ -138,7 +139,7 @@
 		</div>
 	</div>
 
-	<div class="chrome-center" data-tauri-drag-region>
+	<div class="chrome-center" data-tauri-drag-region={useCustomTitleBar}>
 		<div class="chrome-selector-wrapper">
 			<Select
 				searchable
@@ -244,7 +245,7 @@
 		{/if}
 	</div>
 
-	<div class="chrome-right" data-tauri-drag-region>
+	<div class="chrome-right" data-tauri-drag-region={useCustomTitleBar}>
 		{#if $ircEnabled}
 			<NotificationButton
 				hasUnread={isNotificationsUnread}
@@ -429,8 +430,8 @@
 		gap: 8px;
 	}
 
-	/** Mac padding added here to not affect header flex-box sizing. */
-	.mac .chrome-left-buttons {
+	/** Mac padding added here to not affect header flex-box sizing, only applied when using custom title bar. */
+	.mac .chrome-left-buttons.has-traffic-lights {
 		padding-left: 70px;
 	}
 
--- a/apps/desktop/cypress/e2e/support/mock/settings.ts
+++ b/apps/desktop/cypress/e2e/support/mock/settings.ts
@@ -33,5 +33,8 @@ export const MOCK_APP_SETTINGS: AppSettings = {
 	},
 	reviews: {
 		autoFillPrDescriptionFromCommit: true
+	},
+	ui: {
+		useNativeTitleBar: false
 	}
 };
--- a/apps/desktop/src/lib/config/appSettingsV2.ts
+++ b/apps/desktop/src/lib/config/appSettingsV2.ts
@@ -67,6 +67,10 @@ export class SettingsService {
 		await this.backend.invoke('update_fetch', { update });
 	}
 
+	async updateUi(update: Partial<UiSettings>) {
+		await this.backend.invoke('update_ui', { update });
+	}
+
 	private async nudge(): Promise<void> {
 		await this.autoOptInWs3();
 		await this.autoOptInRules();
@@ -154,6 +158,8 @@ export type AppSettings = {
 	claude: Claude;
 	/** Settings related to code reviews and pull requests */
 	reviews: Reviews;
+	/** UI settings */
+	ui: UiSettings;
 };
 
 export type TelemetrySettings = {
@@ -205,4 +211,9 @@ export type Claude = {
 export type Reviews = {
 	/** Whether to auto-fill PR title and description from the first commit when a branch has only one commit. */
 	autoFillPrDescriptionFromCommit: boolean;
+};
+
+export type UiSettings = {
+	/** Whether to use the native system title bar. */
+	useNativeTitleBar: boolean;
 };
--- a/crates/but-api/src/commands/settings.rs
+++ b/crates/but-api/src/commands/settings.rs
@@ -1,7 +1,7 @@
 //! In place of commands.rs
 use but_api_macros::api_cmd;
 use but_settings::api::{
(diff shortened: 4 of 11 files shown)


nshcr:
Creating a stack of PRs from GitButler didn't work, so I'm submitting it directly like this.

The first commit in this PR is from the previous one — hope that won't cause any trouble when merging.

Byron:
Thanks a lot for giving this a go!

I didn't see the new settings in "Appearance" (as mentioned in the PR text as well), but when adding 

```json
  "ui": {
    "useNativeTitleBar": true
  },
```

to the settings file at `~/Library/Application\ Support/gitbutler/settings.json` *and* creating a new application window (or restarting the App) it will show.
This picture shows how the app came up, and a new application window after the settings were changed.

Screenshot 2025-10-08 at 05 17 47

With all that said, as opposed to the traffic-lights PR (#10573), this PR is firmly in the hands of @PavelLaptev :).

nshcr:
@Byron Sorry, I forgot to mention where the new setting is stored in the PR description, thanks for pointing that out!

Let me also add a bit more context about how Tauri custom title bar affects certain macOS system behaviors:

Screenshot 2025-10-08 at 12 43 52

1. macOS allows users to configure double-clicking a window title to Fill / Zoom / Minimize. However, custom title bar (after setting the `data-tauri-drag-region` attribute) only behaves like Zoom, regardless of the system setting. Double-click to minimize is actually quite useful when Stage Manager is enabled.

Screenshot 2025-10-08 at 12 44 40

2. Three-finger drag is a niche feature, I don't personally use it, but it's clear that custom title bar doesn't currently support it.

By the way, I've resolved the conflicts caused by the last PR merge.
This PR looks pretty good now, thanks again!

Byron:
Now that the website is out, maybe @PavelLaptev has a moment to look at this? To me it seems valuable to have, but I don't know if it's worth the trade-off of usage vs. ongoing maintenance.

PavelLaptev:
Hey! Thanks for the PR @nshcr !
As far as I know, other applications don’t provide this option. To me, it feels like we’re trying to mask a Tauri limitation and make it look better — which I don’t like from a design perspective. I’m okay with the current variant and don’t want to make any design changes around it, since that would require a complex redesign, not just margin adjustments.

image

If we do want to expose this setting, let’s keep it under experimental settings. It’s very niche and confusing — like when you offer a native title bar, it makes me wonder if something’s wrong with the current one is it broken?

PavelLaptev:
If you add a native title bar, there shouldn’t be a margin on the left, since it was reserved for the traffic lights.

image

nshcr:
@PavelLaptev Thanks for your reply!

> As far as I know, other applications don’t provide this option.

Actually, VSCode has always offered an option to switch between a native title bar and a custom one.

Screenshot 2025-10-13 at 19 22 42

Obsidian also provides a similar toggle between two custom styles and the native style.

Screenshot 2025-10-13 at 19 23 05

> To me, it feels like we’re trying to mask a Tauri limitation and make it look better — which I don’t like from a design perspective. 

That's absolutely true — I don't like that approach either.

But I see this more as a workaround for Tauri's current limitations, since we may not be able to fix the upstream compatibility issues ourselves.

If Tauri supported custom title bars as well as Electron — reacting properly to system settings — we wouldn't need this at all.

As you can see from the code I shared earlier, what Tauri does for its custom title bar is simply hide the native one and inject a JavaScript file to handle drag and double-click-to-zoom behavior for custom elements.

> I’m okay with the current variant and don’t want to make any design changes around it, since that would require a complex redesign, not just margin adjustments.

That's totally fine with me. My goal with this PR wasn't to suggest a design change, but rather to make GitButler's window behavior functionally consistent with other macOS apps.

The toggle can remain an internal setting for users who specifically want it — I don't think it needs to be exposed in the UI, even under experimental settings (a short note in the docs would be enough).

Anyone who chooses to enable it would already understand that they're trading some visual consistency for functionality.

> It’s very niche and confusing — like when you offer a native title bar, it makes me wonder if something’s wrong with the current one is it broken?

As mentioned before, in some situations, Tauri's custom title bar can indeed be considered "broken", since it doesn't behave according to macOS system settings.

---

Finally, please feel free to close this PR if you decide not to include this option.

If anyone brings up issues about GitButler's window behavior in the future, this PR could still serve as a useful reference for discussion.

Byron:
Do I understand @PavelLaptev correctly that in order to get this merged, the negative space would have to be removed?
If so, is it something we can maintain and keep working? Or would it break next time we touch the titlebar?

Since other major application have this feature, to me it seems that we'd have to pick this feature up as we grow anyway, and this is already a first step in that direction.

A feature toggle in the settings could slowly graduate until it ends up in the UI, and the Window still looks half-decent with it.

But obviously I really have no answer, with about 80 lines added it's also not incredibly lightweight to support, but it would be my hope that @nshcr would be using it, and will keep submitting PRs should it break :).

PavelLaptev:
@Byron, you’re right. @nshcr we have a block in the file [this file](https://github.com/gitbutlerapp/gitbutler/blob/2ada6e22abfb3c4cb07b22f85937cf99f1e72536/apps/desktop/src/components/ChromeHeader.svelte#L432-L435) that adds the left padding for macOS. However, there should be another check to determine if the native title bar is enabled or not.

Here’s the current code:

```css
/** Mac padding added here to avoid affecting header flex-box sizing. */
.mac .chrome-left-buttons {
    padding-left: 70px;
}
```

It would be more appropriate to change to something like this:

```css
/** Additional padding for traffic lights */
.has-traffic-lights .chrome-left-buttons {
    padding-left: 70px;
}
```

nshcr:
@Byron Thanks for your support!

> but it would be my hope that @nshcr would be using it, and will keep submitting PRs should it break :).

Of course I will. The reason I submitted this PR to add this option is that I really need GitButler's window behavior to stay consistent with other applications. The only way I could think of to achieve that was by enabling the native title bar style.

So even if this PR doesn't get merged, I'll still maintain this feature in my own fork build.

> there should be another check to determine if the native title bar is enabled or not.

@PavelLaptev Thanks for pointing out this style change!

While I think keeping the left padding helps maintain visual consistency, removing the blank space to align with the Windows and Linux builds is also a good idea (I haven't tried GUI builds on those two platforms yet, so I'm not entirely sure what the actual layout looks like).

I've updated the code to include this condition.

Byron:
Thanks everyone! I am glad the disposition for this PR is to merge!

Just to help out, I ran the latest version of the PR locally to take some screenshots.

(enabled)

Screenshot 2025-10-14 at 05 32 33

(disabled, the default)

Screenshot 2025-10-14 at 05 34 07

Note that I didn't drag the window to adjust the traffic lights.

PavelLaptev (approved):

apps/desktop/src/components/ChromeHeader.svelte
```diff
@@ -112,12 +113,12 @@
 <div
 	class="chrome-header"
 	class:mac={backend.platformName === 'macos'}
-	data-tauri-drag-region
+	data-tauri-drag-region={useCustomTitleBar}
 	class:single-branch={singleBranchMode}
 	use:focusable
 >
-	<div class="chrome-left" data-tauri-drag-region>
-		<div class="chrome-left-buttons" class:macos={backend.platformName === 'macos'}>
```
PavelLaptev, line 120: @nshcr it looks like you removed the condition for whether it's macOS or not and only added the custom title bar check.

PavelLaptev (changes requested):

apps/desktop/src/components/ChromeHeader.svelte
nshcr: This is a duplicate class that isn't used anywhere. I think having the `.mac` class on the parent element is sufficient.

https://github.com/gitbutlerapp/gitbutler/blob/953a5795a05270a69bcb230768ec4dd629ba337e/apps/desktop/src/components/ChromeHeader.svelte#L112-L118

Of course, it's possible that it's being used somewhere I'm not aware of — if that's the case and it does need to be added back, I'll make that change.

PavelLaptev: I see. Yes, that makes sense, I didn’t notice it and thought we were going to combine the two checks into a single one.

PavelLaptev (approved):

Byron:
Fantastic, thanks so much, both @nshcr for making it happens and @PavelLaptev for the review!
Let's see if and when we will be able to make this more official :).

PR #10582, opened 2025-10-09
fix(docs): update broken logo icon link

nshcr:
This PR (https://github.com/gitbutlerapp/gitbutler/pull/10547) updated the application icons and changed the path for the release version icons, but it missed updating README.md.

I'm not sure if this reflects the intended look for the new icons, but fixing a broken link is never a bad idea.


Byron (approved):
Thanks so much!!

Yes, let's fix the Logo right away as anything is better than this :D

Image

With the PR, it looks like this:

Image

Let's be sure @PavelLaptev knows in case the fixed Icon isn't the desired one.

PR #10584, opened 2025-10-09
Revert "Adjust traffic light size to match the new native size"

nshcr:
This reverts commit 5357fa8d7e28977bbb8707d997b139fb4e56b3e7.

Introduced by https://github.com/gitbutlerapp/gitbutler/pull/10573

Screenshot 2025-10-09 at 14 34 24

In the latest nightly build, the traffic light icons have shifted significantly downward.

Considering that only builds made with the macOS 26 image seem to produce larger traffic lights, and that the original positioning still looks ok even with the larger buttons.

Screenshot 2025-10-09 at 18 08 28

I suggest keeping the original code.

Reference: https://github.com/actions/runner-images


Byron:
Apologies, I didn't receive notifications for three days and thus didn't see you already did the work.

PR #10586, opened 2025-10-09
Assure the traffic-light offset works for MacOS Pre-Tahoe...

Byron:
...because that's the system we build on.

This means the traffic lights will be smaller, which affects the offset. This is quite terrible and it would be better if we wouldn't mess with it at all, especially given the buggy location right after startup due to https://github.com/tauri-apps/tauri/issues/13044.

### DIFF
--- a/crates/gitbutler-tauri/src/window.rs
+++ b/crates/gitbutler-tauri/src/window.rs
@@ -327,7 +327,14 @@ pub fn create(
     if let Some(window) = window.get_window(label) {
         // Note that these lights get reset when the Window label is changed!
         // See https://github.com/tauri-apps/tauri/issues/13044 .
-        window.setup_traffic_lights_inset(LogicalPosition::new(16.0, 28.0))?;
+        // TODO: stop doing this entirely and work with the defaults, my preference,
+        //       create a build script that makes it clear which MacOS version we are using
+        //       to conditionally compile the right value.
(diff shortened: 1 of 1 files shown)


Byron:
CC @estib-vega and @PavelLaptev - traffic lights fix is in so release should be fine. But local builds on Tahoe will look off until we can detect the exact version with a `build.rs` script and pass that info on to the compile-time..

nshcr:
This PR looks great, I've closed mine (https://github.com/gitbutlerapp/gitbutler/pull/10584) since they conflict.

It's really surprising that macOS 26 affects the built app's appearance (even the window corner radius looks slightly different.)

Maybe try spinning up a temporary GitHub Actions job using a macOS 26 image to see if the CI-built version changes, that could help decide whether system-specific handling is necessary.

https://github.com/gitbutlerapp/gitbutler/blob/973751aac74453d9d1891847b1d1b228ccfcb6a1/.github/workflows/publish.yaml#L94-L98

But since macos-latest probably won't migrate to macOS 26 anytime soon, it's not something that needs to be addressed urgently.

https://github.com/actions/runner-images

Byron:
I think system-specific handling would be needed if we kept adjusting the traffic lights. However, it's at least my hope that we can stop doing that, as it also leads to popping behaviour after setting the window label, with no fix in sight.

Also, I have created #10610 to pin the macOS version for now, thanks to your hint :) 🙏.

PR #10605, opened 2025-10-11
Add debug log to learn why some branches are skipped when pushing.

Byron:
Turns out that the old workspace engine is used here and it thinks everything is integrated.


Byron:
Here is a reproducer, try to push anything in the workspace.

[small.zip](https://github.com/user-attachments/files/22862557/small.zip)

I leave it for now as this functionality will need an overhaul anyway.

PR #10601, opened 2025-10-10
make curesors smaller

PavelLaptev:
Before:

https://github.com/user-attachments/assets/908e550a-47c0-434d-8e12-b0cd051c5ddf

After:

https://github.com/user-attachments/assets/9800fd22-9230-45cb-a846-5c91dac71f5c

PR #10607, opened 2025-10-11
fix(cursor): conditionally use 2x custom cursors on WebKit with supports rule

PavelLaptev:
Looks like Firefox doesn't support `image-set` in CSS urls

Before:

https://github.com/user-attachments/assets/7516953f-e78c-440e-b412-e18f3c75328f

After:

https://github.com/user-attachments/assets/d245a158-4102-4e3c-b54e-dbab04f12c4e

### DIFF
--- a/apps/web/src/routes/+layout.svelte
+++ b/apps/web/src/routes/+layout.svelte
@@ -178,7 +178,12 @@
 		flex-direction: column;
 		width: 100%;
 		cursor: var(--cursor-custom);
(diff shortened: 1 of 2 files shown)


PavelLaptev:
Address this issue here: https://github.com/gitbutlerapp/gitbutler/pull/10601#issuecomment-3392961421

PavelLaptev:
@krlvi could you please confirm if this works for you

krlvi:
it is indeed half as large now. thanks

PR #10608, opened 2025-10-11
feat: add Trae IDE protocol handler support

TheNomadInOrbit:
This pull request adds **Trae IDE** as an editor option in GitButler.

Now, users can open files directly in **Trae IDE** using the same pattern followed for **VSCode**, **Cursor**, **Zed**, and **Windsurf**.

## Changes Made
- Added `trae` scheme to the allowed URL schemes list in `crates/but-api/src/commands/open.rs`
- Added Trae IDE option to the editor selection dropdown in `apps/desktop/src/components/profileSettings/GeneralSettings.svelte`

## Testing
- Verified that the code compiles successfully
- Confirmed that Trae appears as an option in the editor selection dropdown
- Confirmed that selecting Trae IDE generates the correct `trae://` protocol URLs


Byron (approved):
Thanks a lot, looks good to me!

PR #10616, opened 2025-10-13
Enable more fine-grained caching for Rust enabled CI pipelines

Byron:
This should speed it up, particularly for the new `but` build.
The general idea is to just reuse caches from `master`, but to not bother saving them.

Also, keep specialised caches for related jobs.

### Tasks

* [x] rust cache optimisation
* [x] skip custom image (45s) for dependency installation.
* [x] ~~avoid apt-update~~ - not possible

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -94,8 +94,6 @@ jobs:
     needs: changes
     if: ${{ needs.changes.outputs.rust == 'true' }}
     runs-on: ubuntu-latest
-    container:
-      image: ghcr.io/gitbutlerapp/ci-base-image:latest
     env:
       CARGO_TERM_COLOR: always
     steps:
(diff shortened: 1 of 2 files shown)


Byron:
Note that the usage is up for now as this PR can't use caches yet. However, once it ran on `master` caches will be updated, which will lead to a speedup of all PRs once again, and overall *should* mean savings in CI time.

For reference, this is the run **without cache**, but **with containers**.

Screenshot 2025-10-13 at 10 30 00

This the run **without cache**, and **without containers** - expected to be a bit slower as dependency installation takes longer. 

Screenshot 2025-10-13 at 10 36 57

Maybe `apt update` should be skipped as it takes 12 seconds? it's the question if the apt-database is populated. Let's try it.

No, that doesn't work, nor does it work to speed up `apt`.

Byron:
Merging now, and will follow-up with statistics once the cache is present and available for use.

Byron:
**With container** as Apt cache it's now this:

The longest runtime is 13m38s.

Screenshot 2025-10-13 at 13 54 47


**With container** and **with rust cache hit**

The longest runtime is 4m50s .

Screenshot 2025-10-13 at 14 41 31

And this is the old usage, with cache hit: 26m is now down to ~19m, 7m savings. Unfortunately, the slowest job didn't get much faster, with 5m for the Rust test compared to 4m50s.

Screenshot 2025-10-13 at 14 45 08

PR #10619, opened 2025-10-13
`but worktree new`

Caleb-T-Owens:
Currently takes in an unqualified branch name, the same as branch tips.

e2e tests will be added for the `but` cli after `but worktree list` is added


crates/but-worktrees/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+/// A worktree entry representing a Git worktree.
+/// This gets exposed in public JSON outputs, be careful when removing
+/// properties.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Worktree {
+    /// The filesystem path to the worktree.
+    pub path: PathBuf,
+    /// The git reference this worktree was created from.
+    pub reference: String,
```
Caleb-T-Owens, line 33: This would ideally be a BString, but currently our DB mappings only use real strings. Currently the `reference` is a name that we generate, and will be utf-8.

We will be able to migrate forward to BStrings in a compatible manner

Byron: ~~Despite the typing, when reading this I'd wonder if this is a full name or a short-name. 
My guess is that it's a full name, but then it would have to use `gix::refs::FullName`, so I would end up thinking it's a short name.~~
I think I am confused as to why this isn't `gix::refs::FullName`.

crates/but-worktrees/tests/fixtures/worktree.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+set -eu -o pipefail
+CLI=${1:?The first argument is the GitButler CLI}
```
Byron, line 3: ~~I hope we can find a way to not proliferate the `gitbutler-cli` - it shouldn't be needed anymore.~~

It turns out this wants to say `but`-CLI, as that's actually used.

I still believe it's too heavy to use and in practice, it's less noisy to create commit graphs the way it's done in `but-workspace` (and much faster to execute as well).

crates/but-worktrees/tests/worktree/main.rs
```diff
@@ -0,0 +1,189 @@
...
+        let guard = test_ctx.ctx.project().exclusive_worktree_access();
+        let repo = test_ctx.ctx.gix_repo_for_merging()?;
+        let meta = VirtualBranchesTomlMetadata::from_path(
+            test_ctx
+                .ctx
+                .project()
+                .gb_dir()
+                .join("virtual_branches.toml"),
+        )?;
+        let stacks = stacks_v3(&repo, &meta, StacksFilter::InWorkspace, None)?;
```
Byron, line 47: For what to use when testing, I recommend taking a look at `but_workspace::create_reference()`, or `but_workspace::branch::apply()`. `stacks_v3` is an adapter to link the old and the new, and basically deprecated.

crates/gitbutler-testsupport/src/lib.rs
```diff
@@ -114,6 +114,49 @@ pub mod writable {
...
+
+    /// Use the `but` CLI instead of `gitbutler-cli` for fixtures that need stacking support.
+    pub fn but_fixture_project(
+        script_name: &str,
+        project_directory: &str,
+    ) -> anyhow::Result<(Project, TempDir)> {
+        let root = gix_testtools::scripted_fixture_writable_with_args(
+            script_name,
+            Some(BUT_DRIVER.display().to_string()),
+            gix_testtools::Creation::ExecuteScript,
```
Byron, line 148: Simple copying isn't possible anymore because of all the absolute paths that `but` will inject into configuration files and such, which makes such scripts, if writable, unnecessarily slow.

crates/but-worktrees/src/git.rs
```diff
@@ -0,0 +1,33 @@
+use std::path::Path;
+
+use anyhow::{Result, bail};
+use tracing::{Level, event};
+
+pub(crate) fn git_worktree_add(
+    project_path: &Path,
+    path: &Path,
+    branch_name: &str,
```
Byron, line 9: This should definitely be strongly typed.

crates/but-worktrees/src/git.rs
```diff
@@ -0,0 +1,33 @@
...
+            .current_dir(project_path)
+            .arg("worktree")
+            .arg("add")
+            .args(["-B", branch_name])
+            .arg(path.as_os_str())
+            .arg(commit.to_string())
+            .output()?;
+
+    event!(Level::INFO, "{}", str::from_utf8(&output.stdout)?);
+    event!(Level::ERROR, "{}", str::from_utf8(&output.stderr)?);
```
Byron, line 23: Why not `tracing::error|info!()`?

crates/but-worktrees/src/db.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+impl TryFrom<Worktree> for but_db::Worktree {
+    type Error = anyhow::Error;
+
+    fn try_from(value: Worktree) -> Result<Self, Self::Error> {
+        let source = serde_json::to_string(&value.source)?;
+        let base = value.base.to_hex().to_string();
+        let path = value.path.to_string_lossy().to_string();
```
Byron, line 64: Let's convert to bytes instead to not be lossy here.

crates/but-db/src/worktrees.rs
```diff
@@ -0,0 +1,58 @@
...
+    Debug, Clone, PartialEq, Serialize, Deserialize, Queryable, Selectable, Insertable, Identifiable,
+)]
+#[diesel(table_name = crate::schema::worktrees)]
+#[diesel(check_for_backend(diesel::sqlite::Sqlite))]
+#[diesel(primary_key(path))]
+pub struct Worktree {
+    pub path: String,
+    pub reference: String,
+    pub base: String,
+    pub source: String,
```
Byron, line 19: These definitely must be documented.

crates/but-db/src/worktrees.rs
```diff
@@ -0,0 +1,58 @@
...
+use serde::{Deserialize, Serialize};
+
+#[derive(
+    Debug, Clone, PartialEq, Serialize, Deserialize, Queryable, Selectable, Insertable, Identifiable,
+)]
+#[diesel(table_name = crate::schema::worktrees)]
+#[diesel(check_for_backend(diesel::sqlite::Sqlite))]
+#[diesel(primary_key(path))]
+pub struct Worktree {
+    pub path: String,
```
Byron, line 16: Let's avoid putting filesystem paths when we can. Here we can, as the worktree has a name, the directory where its `.git` directory is located in `.git/worktrees/<name>`. This is enough to discover everything, when the project itself is known.

crates/but-worktrees/src/db.rs
```diff
@@ -0,0 +1,73 @@
+//! Database operations for worktrees.
+
+use anyhow::Result;
+use gitbutler_command_context::CommandContext;
```
Byron, line 4: We shouldn't pull in `CommandContext` or anything from `gitbutler_*` for that matter.
It's insta-legacy that way.

Byron (commented):
I have given this a good look, yet my comments aren't exhaustive by any means.

Overall, I think this is a mix of old and new and work is needed to bring it into the new world. I fully understand that this isn't a well-trodden path yet, but I think it's time that needs to be spent.

Unfortunately I won't make it into the office in Thursday, but I will be there on Friday to maybe give a tour on testing in `but-workspace`.

crates/but-worktrees/src/db.rs
Caleb-T-Owens: This is addressed here: https://github.com/gitbutlerapp/gitbutler/pull/10648/files

crates/but-db/src/worktrees.rs
Caleb-T-Owens: This is resolved here https://github.com/gitbutlerapp/gitbutler/pull/10648/files

crates/but-worktrees/src/git.rs
Caleb-T-Owens: Because I came across this in the tracing documentation first 😅; resolved here https://github.com/gitbutlerapp/gitbutler/pull/10648/files

crates/but-worktrees/src/git.rs
Caleb-T-Owens: Resolved here https://github.com/gitbutlerapp/gitbutler/pull/10648/files

crates/but-worktrees/tests/fixtures/worktree.sh
Caleb-T-Owens: The misleading comment is corrected here: https://github.com/gitbutlerapp/gitbutler/pull/10648/files

crates/but-worktrees/src/lib.rs
Caleb-T-Owens: This is resolved here: https://github.com/gitbutlerapp/gitbutler/pull/10648/files

crates/but-worktrees/src/db.rs
Caleb-T-Owens: Currently the CommandContext acts as the interface for working with the database, so unless this work encapsulated adding a different way of accessing it, this is at least one requirement.

This work will rely on `Stack::as_rebase_steps()` for the rebase-engine, in the absence of newer primitives. I'd rather not include making a graph-based rebase engine into this set of work

crates/gitbutler-testsupport/src/lib.rs
Caleb-T-Owens: Could you clarify which are the behaviours that the old CLI didn't exhibit, but the new one will?

crates/but-worktrees/tests/fixtures/worktree.sh
Caleb-T-Owens: In what regards is the `but` CLI too heavy? If that is the case, should we not try and improve that?

crates/but-worktrees/src/db.rs
Byron: > Currently the CommandContext acts as the interface for working with the database, so unless this work encapsulated adding a different way of accessing it, this is at least one requirement.

Thanks for the clarification, I also see that `but-hunk-assignments` does the same, probably for similar reasons.
I was not aware, and am a bit shocked that the new database hides behind something in a `gitbutler_*` crate.

How hard would it be to consume the database directly?

I also think we have a 'problem' as even if the database would be passed directly, now we have `meta` and `db`, with `meta` being metadata attached to ref-names, which eventually will be stored in… the `db` :D.

Passing the whole `db` also feels like a danger as it gives access to all data, not just the data the function actually consumes.

In this case, and on the bright side, `meta` probably isn't required as the `workspace` should be passed directly.

Byron: > This work will rely on `Stack::as_rebase_steps()` for the rebase-engine, in the absence of newer primitives. I'd rather not include making a graph-based rebase engine into this set of work

I see - I'd definitely refactor this even if it's just a 'hack' at first to allow using the rebase utility with the new data structure, but I see how this would be postponed since you are in a hurry.

crates/but-worktrees/tests/fixtures/worktree.sh
Byron: It's not what would be minimally required to make this work. For that, it needs only a Git repository. Workspaces aren't special anymore either and can quite easily be created with commits only - the workspace will be discovered based on that alone.

Anything else that's needed can be created on the fly.
The idea is to let the plumbing be as specific as possible in its needs, which is also reflected by the parameters passed the functions. It's the least amount of data that it needs to function.

The minimisation of dependencies is probably what the above boils down to, which helps with focus and prevents things to get messy (one can only mess around with what you have).

This is only my reasoning, not an attempt to convince you.

crates/gitbutler-testsupport/src/lib.rs
Byron: Sorry, this comment seems only half-ready.

The idea is that `ExecuteScript` should be avoided in preference of `CopyFromReadOnly`. If that's not possible, and I don't think it is here as `git worktree` will put absolute paths everywhere, then the script being executed should be as light as possible.

With `but` in the mix that's not the case anymore. Using bare Git repositories is ideally all that's ever needed for any test we write.

crates/but-db/src/worktrees.rs
Byron: @Caleb-T-Owens Just to bring this to your attention, one last time.

PR #10626, opened 2025-10-13
try to make E2E playwright faster on CI

Byron:
### Tasks

* [x] try to accelerate tauri deps
* [x] try to accelerate Rust build
* [x] see if `sudo apt install -y libnss3` can make chrome work with the cache, as faster alternative to not having a cache.

### DIFF
--- a/.github/workflows/test-e2e-playwright.yml
+++ b/.github/workflows/test-e2e-playwright.yml
@@ -13,6 +13,8 @@ jobs:
   test:
     name: e2e-playwright
     runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/gitbutlerapp/ci-base-image:latest
     timeout-minutes: 60
     env:
       DESKTOP_PORT: 3000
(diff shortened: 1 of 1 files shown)


Byron:
CC @estib-vega just to be sure I am not missing anything, or broke something. Particularly the relative paths for artefact uploads *should* work, but I didn't test it.

Byron:
### Before: 11m 51s

Screenshot 2025-10-13 at 19 14 09

### After: 10m 36s (without Rust cache)

Screenshot 2025-10-13 at 19 13 16

This is still not much faster than before, but once the Rust-cache hits, the speedup *should* be very meaningful.

Byron:
More performance data is here: https://github.com/gitbutlerapp/gitbutler/pull/10630#issuecomment-3398496233.

In short, we are hitting 5m if the Rust cache is hit. 6m30s for Playwright due to Sveltekit.

PR #10630, opened 2025-10-13
Make E2E blackbox run faster with improved caching

Byron:
### Tasks

* [x] works in container?
* [x] works with Rust cache?


Byron:
Turns out that with a Rust cache hit, the E2E time goes down to 5m!!!

Let's just hope the cache is actually available after running it on master.

Screenshot 2025-10-13 at 19 42 32

Byron:
Looks like Playwright is slower due to sveltekit, but could also hit 5m if that wasn't built.

Screenshot 2025-10-13 at 19 43 30

PR #10640, opened 2025-10-14
publishing performance

Byron:
The potential for savings is huge:

Screenshot 2025-10-14 at 05 54 40

And I wouldn't be surprised if the slowest runners finish in 10m or so, with publishes going through in less than 15m.

### Tasks

* [x] add rust caching
* [x] create a release to build the cache
* [x] another release to see how fast it is now
* [x] see the timings with multi-threaded and how different `lto` options affect the publish
    - On CI this doesn't gain much, but locally it's a world of a difference, which I think helps development as the default release build is (relatively) fast now.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -187,6 +187,7 @@ jobs:
             gir1.2-webkit2-4.1=2.44.0-2;
 
       - uses: actions/download-artifact@v5
+        name: Download SvelteKit build output
         with:
           name: sveltekit-build
           path: ./apps/desktop/build/
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -134,6 +134,10 @@ jobs:
       - uses: actions/checkout@v5
         with:
           token: ${{ secrets.PAT_JUNON }} # custom token here so that we can push tags later
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v2.8.1
+        with:
+          shared-key: app-release-build
       - name: Init Node Environment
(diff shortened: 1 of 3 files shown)


Byron:
The slowest job is now 15m faster, with just a Rust cache.

Screenshot 2025-10-14 at 08 05 37

The problem here is the single-threadedness, which I don't think is worth it until we measure it and decide it's worth it.
And of course, the actual problem is `tauri` which invalidates a lot of the build artefacts.

Byron:
It turns out that builds without threading limitation (and **without cache**) and with LTO = "thin" isn't faster at all, like there are no cores to spare anyway. Particularly for the final binary I would have thought threading makes a difference.

Screenshot 2025-10-14 at 09 24 45

I am tempted to say that the time it took to handle the cache now is a net-loss, so the first build will have  penalty.

Compare this to the original build times, without cache, and single-threaded "fat" LTO builds:

Screenshot 2025-10-14 at 09 30 54

I don't worry about this at all, as particularly the slowest macOS 13 build can vary greatly, it seems, and it alone can be responsible for the entire difference in total runtime. Generally, it's responsible for the worst-case performance.

Byron:
It turns out that some runners profit more than others from the improved concurrency, but ultimately it's the slowest runner that decides on the overall latency to a new release.

And actually, MacOS 13 also got a big speedboost, from 35m to 27m.

Screenshot 2025-10-14 at 09 56 25

This means we can now publish in 31m!

Screenshot 2025-10-14 at 09 56 40

Down from 55m, a 24m improvement.

Screenshot 2025-10-14 at 09 57 05

PR #10645, opened 2025-10-14
But worktree list

Caleb-T-Owens:
Currently, how to e2e test the worktree list is undecided, so I'm going to leave the actual CLI untested for now. The internal command has full coverage though

### DIFF
--- a/crates/but-worktrees/src/list.rs
+++ b/crates/but-worktrees/src/list.rs
@@ -1,0 +1,48 @@
+use anyhow::Result;
+use but_graph::VirtualBranchesTomlMetadata;
+use but_workspace::{StacksFilter, stacks_v3};
+use gitbutler_command_context::CommandContext;
+use gitbutler_project::access::WorktreeReadPermission;
+use serde::{Deserialize, Serialize};
+
+use crate::{Worktree, WorktreeHealthStatus, db::list_worktrees, gc::get_health};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+/// This gets used as a public API in the CLI so be careful when modifying.
+pub struct ListWorktreeOutcome {
+    pub entries: Vec<WorktreeListEntry>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+/// This gets used as a public API in the CLI so be careful when modifying.
+pub struct WorktreeListEntry {
+    pub status: WorktreeHealthStatus,
+    pub worktree: Worktree,
+}
+
+/// Creates a new worktree off of a branches given name.
+pub fn worktree_list(
+    ctx: &mut CommandContext,
+    _perm: &WorktreeReadPermission,
+) -> Result<ListWorktreeOutcome> {
+    let repo = ctx.gix_repo_for_merging()?;
+    let meta = VirtualBranchesTomlMetadata::from_path(
+        ctx.project().gb_dir().join("virtual_branches.toml"),
+    )?;
+    let stacks = stacks_v3(&repo, &meta, StacksFilter::InWorkspace, None)?;
+    let heads = stacks.into_iter().flat_map(|s| s.heads).collect::<Vec<_>>();
+
+    let entries = list_worktrees(ctx)?
+        .into_iter()
+        .map(|w| {
+            Ok(WorktreeListEntry {
+                status: get_health(&repo, &w, &heads)?,
+                worktree: w,
+            })
+        })
+        .collect::<Result<Vec<_>>>()?;
+
+    Ok(ListWorktreeOutcome { entries })
+}
--- a/crates/but-worktrees/tests/worktree/main.rs
+++ b/crates/but-worktrees/tests/worktree/main.rs
@@ -23,7 +23,7 @@ mod util {
     }
 }
 
-mod stacked_and_parallel {
+mod worktree_new {
     use super::*;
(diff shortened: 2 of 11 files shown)


crates/but-worktrees/src/list.rs
```diff
@@ -0,0 +1,48 @@
...
+/// This gets used as a public API in the CLI so be careful when modifying.
+pub struct WorktreeListEntry {
+    pub status: WorktreeHealthStatus,
+    pub worktree: Worktree,
+}
+
+/// Creates a new worktree off of a branches given name.
+pub fn worktree_list(
+    ctx: &mut CommandContext,
+    _perm: &WorktreeReadPermission,
```
Byron, line 28: This permission system shouldn't be needed in `plumbing` crates - these are to do only what they need to do, with the minimal amount of data and information needed.
Hence we won't pass `ctx`, but whatever it actually needs to perform that job.

The concurrency-protection (and snapshots, and what not) will then be added by the `but-api` crate.

crates/but-worktrees/src/list.rs
```diff
@@ -0,0 +1,48 @@
...
+) -> Result<ListWorktreeOutcome> {
+    let repo = ctx.gix_repo_for_merging()?;
+    let meta = VirtualBranchesTomlMetadata::from_path(
+        ctx.project().gb_dir().join("virtual_branches.toml"),
+    )?;
+    let stacks = stacks_v3(&repo, &meta, StacksFilter::InWorkspace, None)?;
```
Byron, line 34: Similar to how it's done in `but-workspace`, if something needs to know about the workspace, it would consume `&but_graph::projection::Workspace`.

crates/but-worktrees/tests/worktree/main.rs
```diff
@@ -187,3 +187,104 @@ mod stacked_and_parallel {
...
+        std::process::Command::from(gix::command::prepare(gix::path::env::exe_invocation()))
+            .current_dir(&ctx.project().path)
+            .arg("worktree")
+            .arg("remove")
+            .arg("-f")
+            .arg(d.created.path.as_os_str())
+            .output()?;
+
+        // remove `feature-c` branch from workspace
+        // It would be nice to invoke the `but` cli here...
```
Byron, line 252: It would be nice to have `but_workspace::branch::unapply()` here :).

crates/but-worktrees/tests/worktree/main.rs
```diff
@@ -187,3 +187,104 @@ mod stacked_and_parallel {
...
+            "New reference :D",
+        )?;
+        std::process::Command::from(gix::command::prepare(gix::path::env::exe_invocation()))
+            .current_dir(&c.created.path)
+            .arg("switch")
+            .arg("new-ref")
+            .output()?;
+
+        // delete d's worktree
+        std::process::Command::from(gix::command::prepare(gix::path::env::exe_invocation()))
```
Byron, line 243: Actually, there is but_testsupport::git()` for this, it's much less boilerplate.

crates/but-worktrees/src/list.rs
```diff
@@ -0,0 +1,48 @@
...
+
+/// Creates a new worktree off of a branches given name.
+pub fn worktree_list(
+    ctx: &mut CommandContext,
+    _perm: &WorktreeReadPermission,
+) -> Result<ListWorktreeOutcome> {
+    let repo = ctx.gix_repo_for_merging()?;
+    let meta = VirtualBranchesTomlMetadata::from_path(
+        ctx.project().gb_dir().join("virtual_branches.toml"),
+    )?;
```
Byron, line 33: If it needs `meta`, it's passed as parameter.

Byron (commented):
And here a few more comments which I hope help shape future work.

crates/but-worktrees/src/list.rs
Caleb-T-Owens: I thought the whole point of the permissions thing was to have a better representation of what reading and or writing to repo/working repository.

The functions in `but-api` get called directly, so are we not back to where we were before the permissions thing was introduced in the first place?

crates/but-worktrees/src/list.rs
Caleb-T-Owens: Where? how do you mean?

crates/but-worktrees/src/list.rs
Byron: These primitives solve the application-level concurrency that we get and help mitigate problems around racy file-IO.

It's already clear that any mutation of anything we have is inherently racy as we can't atomically update a DB *and* what's on disk. That can only be accepted, with the expectation that the function itself is consistent, but with no expectation that it can handle races perfectly.

The synchronisation we have on application level is only made for a single application, multiple ones (i.e. windows) will still race, nothing is truly perfect in that regard outside of maybe a database.

I hope that makes sense.

> The functions in `but-api` get called directly, so are we not back to where we were before the permissions thing was introduced in the first place?

This means the `but-api` is responsible for obtaining the required lock, even if it's not passed down.
Some of the `gitbutler-*` crate functions had this as parameter, and it might even have been me who added it, but merely as a crutch to know that certain portions won't be callable without synchronisation, and having a compiler who tells me all the call sites.

The `but-api` is the new entrypoint, and it's expected to do everything correctly, so most will just call that.

crates/but-worktrees/src/list.rs
Byron: I'd replace `_perm` with `meta: &impl RefMetadata`.

PR #10651, opened 2025-10-14
Use new APIs

Caleb-T-Owens:


crates/but-worktrees/src/gc.rs
```diff
@@ -1,16 +1,13 @@
 use crate::{Worktree, WorktreeHealthStatus, WorktreeSource};
 use anyhow::Result;
-use but_workspace::ui::StackHeadInfo;
 
 pub fn get_health(
     repo: &gix::Repository,
     worktree: &Worktree,
-    heads: &[StackHeadInfo],
+    ws_segment_names: &[gix::refs::PartialName],
```
Byron, line 7: Let's avoid using `PartialName`, and prefer full names. Partial names can be ambiguous.

Byron:
@Caleb-T-Owens I am trying hard to find a way to not use `gitbutler-*` crates. The problem seems to arise from the database. And as all data we collect is tied to the workspace *and* read-only, I think the data we need can be put into a file into `.git/worktrees/<name>/` without problems.

That would bring down the truly required information to (hopefully) `repo` and `workspace`, while differentiating GitButler workspaces from ones the user may have.

What do you think about this?

PR #10659, opened 2025-10-15
but-worktree-integrate

Caleb-T-Owens:
This PR implements the `integrate` API, as well as makes changes to make integrating any worktree possible by making the metadata optional


crates/but-db/migrations/2025-10-15-105125_update_worktrees_schema/up.sql
```diff
@@ -0,0 +1,15 @@
+-- Update worktrees table to new schema
+-- SQLite doesn't support ALTER COLUMN, so we need to recreate the table
+
+-- Create new table with updated schema
+CREATE TABLE `worktrees_new`(
+	`path` BLOB NOT NULL PRIMARY KEY,
+	`base` TEXT NOT NULL,
+	`created_from_ref` BLOB
```
Byron, line 8: These two items can be stored directly in the worktree, so the database isn't needed for these write-once items.

These items would just be standard Git refs that can be stored per worktree, for instance in `.git/worktrees/<name>/BASE|CREATED_FROM`.

crates/but-worktrees/src/integrate.rs
```diff
@@ -0,0 +1,278 @@
...
+
+    status
+        .stack
+        .set_heads_from_rebase_output(ctx, status.rebase_output.references)?;
+    let after = WorkspaceState::create(ctx, perm.read_permission())?;
+    update_uncommited_changes(ctx, before, after, perm)?;
+    let vb_state = VirtualBranchesHandle::new(ctx.project().gb_dir());
+    update_workspace_commit(&vb_state, ctx)?;
+
+    git_worktree_remove(&ctx.project().path, path, true)?;
```
Byron, line 85: Worktrees *really* must not be identified by `path`, but by their [id](https://docs.rs/gix/latest/gix/struct.Worktree.html#method.id). This also exists on the worktree proxies that are iterated: [`id`](https://docs.rs/gix/latest/gix/worktree/struct.Proxy.html#method.id).

Byron (commented):
Thanks for the ping. I took a quick look and made comments, but also feel I am repeating myself so I will stop bringing those up in front of you in future. I am definitely not happy with the way data is stored, but also think can be changed without too much trouble.

Further, when running the tests I noticed that it tried to sign a commit and failed: `Error: errors.commit.signing_failed` in `test_create_unrelated_change_and_reintroduce`, but I didn't investigate. Maybe something about opening `gix` repositories within these plumbing functions instead of passing them (causing them to be opened without isolation).

Another good reason to not rely on the binary for setting up tests is latency, as now a code change will also have to rebuild the binary and double the time it takes till the test can start. Plumbing doesn't (or shouldn't) rely on metadata that can't be created in memory, to keep the setup costs minimal to what's needed: a git repository and maybe some metadata that is injected on the fly.

I think relying on a lot of the old world is fine to get the actual integration done more quickly, and it can be reworked later.

Besides that, I enthusiastically think you are pioneering what at some point in the future will be standard capabilities of GitButler, with worktrees as part of the workspace, and re-integrating them like any other ref that advanced outside of the workspace, with the speciality that one can offer to remove the worktree when done.

PR #10661, opened 2025-10-15
GitHub Rust pt. I: Initial migration to the Rust end & Multiple GitHub users

estib-vega:
- Create a GitHub client in the RustEnd (very basic at the moment)
- Completely ignore the access token stored in the GitButler API server
- Allow for the user to have multiple GitHub accounts active
- Allow the user to select an account to use per project


estib-butler:
This effectively will force the user to re-authenticate. 
I don't think that's too bad, specially since now we have the prompt in the workspace.
I'll look into adding some migration code, but don't want to make it to complicated either

PR #10662, opened 2025-10-15
fix(ui): replace generic user check with forge-specific validation

vitallium:
## 🧢 Changes

Hi! This PR replaces generic user check with forge-specific validation:

If the user is configured and the forge type is GitHub, check the user. Display an error toast to inform the user that
the GitHub integration is not configured.

CleanShot 2025-10-15 at 21 25 16@2x

If the user is not configured and the forge type is GitLab, skip the user check and instead verify the GitLab configuration state. Display an error toast to provide feedback that the GitLab integration is not configured.

CleanShot 2025-10-15 at 21 25 33@2x

## ☕️ Reasoning

When the user clicks on the `Create MR` button then nothing happens. The old check was too generic and didn't handle GitHub and GitLab properly. This PR makes sure each integration is checked correctly and users get proper feedback. I put some generic error messages there but happy to adjust them. Thanks!

## 🎫 Affected issues

Fixes: https://github.com/gitbutlerapp/gitbutler/issues/9964

### DIFF
--- a/apps/desktop/src/components/ReviewCreation.svelte
+++ b/apps/desktop/src/components/ReviewCreation.svelte
@@ -21,6 +21,7 @@
 	import { DEFAULT_FORGE_FACTORY } from '$lib/forge/forgeFactory.svelte';
 	import { mapErrorToToast } from '$lib/forge/github/errorMap';
 	import { GitHubPrService } from '$lib/forge/github/githubPrService.svelte';
+	import { GITLAB_STATE } from '$lib/forge/gitlab/gitlabState.svelte';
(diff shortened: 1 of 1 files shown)


Byron (approved):

To me this change feels like a net win over what's there and should be merged for that reason alone.

A future improvement could be to let the forge decide if it's ready to be used, but I don't know if now should be the time to change it.

Reeling in @estib-vega for that :).

estib-vega (approved):
Thanks for this improvement!
It's highly appreciated 🎉

PR #10669, opened 2025-10-16
Mark `tokens` as sensitive and follow the existing pattern to prevent log-leakage

Byron:
The idea is that `Sensitive` parts are clearly marked and the typesystem enforces special handling.

That must not be circumvented or avoided.

Follow-up of #10661.


Byron:
@krlvi I think it's something for the engineering meeting to point out that all secrets must be marked with `Sensitive`, and that the pattern must not be avoided or circumvented.

The typesystem was continuously conveying a message, but it was bypassed instead of following the existing pattern.
Thank you

PR #10679, opened 2025-10-16
Store the GH users in one secret

estib-vega:
- Don’t use the settings to track the authenticated users
- Use only one secret to store the users authenticated to the application

---
- <kbd>&nbsp;1&nbsp;</kbd> #10679 👈


crates/but-github/src/token.rs
```diff
@@ -0,0 +1,75 @@
...
+    Ok(map.into_iter().map(|entry| entry.username).collect())
+}
+
+const GITHUB_OAUTH_ACCESS_TOKEN_MAP_KEY: &str = "github_oauth_access_token_map";
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+struct UserToken {
+    pub username: String,
+    pub access_token: String,
+}
```
Byron, line 71: Can we differentiate the token type by looking at the `access_token`? I am asking because of the semantic difference between device tokens and personal access tokens.

crates/but-github/src/token.rs
```diff
@@ -0,0 +1,75 @@
...
+        &serialized_map,
+        secret::Namespace::Global,
+    )
+}
+
+fn retrieve_github_oauth_access_token_map() -> Result<Vec<UserToken>> {
+    static FAIR_QUEUE: Mutex<()> = Mutex::new(());
+    let _one_at_a_time_to_prevent_races = FAIR_QUEUE.lock().unwrap();
+    if let Some(serialized_map) =
+        secret::retrieve(GITHUB_OAUTH_ACCESS_TOKEN_MAP_KEY, secret::Namespace::Global)?
```
Byron, line 90: Since the access is now `Global`, wouldn't we run into this macOS26 bug of nightly not being able to access stable and vice versa?
Or is this now working, somehow?

Byron (commented):
Before I go, I also wanted to mention that there is this hack which completely changes what secrets we use (and how we store them) in dev mode.

https://github.com/gitbutlerapp/gitbutler/blob/9c5a1d540dafde37da018ae1504abe5ca01103fa/crates/gitbutler-tauri/src/main.rs#L103-L105

So to test this, you'd always have to do a release build, something that now at least is much faster than before.

crates/but-github/src/token.rs
estib-vega: Not sure, but might as well bake in some nice differentiation now 👍

crates/but-github/src/token.rs
estib-vega: Thought of leaving that for now as it's not going to change it's effect (at least on macOS) but if we can fix it then it should just improve automatically

Byron: What if `Global` would always use `com.gitbutler.app`? Then there would be more re-use and it's all gravitating towards the stable release.
Also, if we change our minds, stable users wouldn't notice the change at all.

~~I definitely feel like if we are going forget users current GitHub login when switching to the mapping, we might as well make this change, too, maybe with a migration.~~ Actually I am unsure what this macOS26 bug really is, let me try to experience it for real before thinking about this further 😅.

PR #10680, opened 2025-10-16
When setting the base branch, also perform a checkout (#10674)

### LINKED ISSUE #10674: Trunk branch change Clarification

My situation is very similar to this comment https://github.com/gitbutlerapp/gitbutler/issues/2507#issuecomment-2181078977:

- In addition to the main branch we have one branch per supported release
- Sometimes I need to checkout one of these release branches to add or cherrypick changes

This is what #2507 was about, and indeed I can update the target branch in the Project Settings > Remote Configuration.

Expected behaviour:

- I start with a clean workspace with the main branch as target
- I change the target branch
- I end with a clean workspace with the release branch as target

Actual behaviour:

- I start with a clean workspace with the main branch as target
- I change the target branch
- I end with the release branch as target but the checked out files of the main branch

I want the equivalent of git checkout, but gitbutler seems to do the equivalent of git update-ref. Changing the branch but not the filesystem content.

I'm not sure if I'm missing the intended way to use the feature, or if this should be a separate feature request.

Notably all files which existed in the old target branch and don't exist in the new target branch are still on the filesystem but now count as untracked.
Cleaning them up without affecting 'really' untracked files seems impossible. If I manually check out the new target branch to prepare the filesystem into the correct state (bypassing the gitbutler workspace branch) I cannot change the target branch in gitbutler anymore.

What is the intended workflow, starting from a clean workspace, to change the target branch and end with a clean workspace?
I think this feature is too new to be documented yet?

_No response_

### Does the docs page already exist? If so, please link to it.

_No response_

Byron:
As we change the HEAD, we should try to set the worktree to match.

I still don't know why this doesn't work anymore, or if it ever worked.

Fixes #10674 .


Byron:
@krlvi Please take a look at the video for the before-after behaviour.

https://github.com/user-attachments/assets/bfcbe7c6-9e4c-426d-b77d-f5d291278ecd

**The question is if I am missing something about the intended behaviour.**

In general, `safe_checkout` can truly be used as such, and if there is a conflict with working tree changes it will fail with an error naming the conflicting file. If that happens, the result is still the same, and the target will have been switched.

I will merge as it works according to my tests, but we can follow it up if I am missing something.

PR #10695, opened 2025-10-17
Parse Gerrit status tags like [WIP] and [PRIVATE]

schacon:
Fix parsing of Gerrit remote push lines so status tags (e.g. [WIP], [NEW], [PRIVATE]) at the end of the line are recognized and stripped from the commit title.

The old logic only detected exactly "[NEW]" and failed to remove or recognize other tags, causing URLs and titles not to be parsed correctly when using options like %wip. This change generalizes detection of a trailing status tag, trims it safely, and adds unit tests for WIP, PRIVATE, and edge cases (titles containing brackets).

### DIFF
--- a/crates/but-gerrit/src/parse.rs
+++ b/crates/but-gerrit/src/parse.rs
@@ -106,7 +106,8 @@ enum ChangeSection {
 fn parse_change_info(line: &str) -> Option<ChangeInfo> {
     // Parse lines like:
     // "http://15a45d4cba1a/c/gerrit-test/+/42 aaaaaaa [NEW]"
-    // or "http://15a45d4cba1a/c/gerrit-test/+/41 sup5"
+    // "http://15a45d4cba1a/c/gerrit-test/+/41 sup5"
+    // "http://192.168.178.83:8080/c/testing/+/36 tt [WIP]"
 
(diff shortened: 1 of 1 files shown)


schacon:
This doesn't seem to actually fix the WIP problem (`but push` with a wip flag does not associate a url with the pushed commits), but it must be somewhere around this code.

krlvi:
@schacon i got a working fix here https://github.com/gitbutlerapp/gitbutler/pull/10708

Btw - completely meta, the prompts are not version controlled, but i was able to one-shot instruct claude to fix this with the following prompt:

```
The parse module of the but-gerrit crate handles the text output when pushing to a gerrit server. It parses the output into a structure. However it seems like this does not correctly in the case when the push includes the %wip option.
Heres and example output that needs to be parsed:
"remote: \rremote: Processing changes: refs: 1, new: 1 \rremote: Processing changes: refs: 1, new: 1 \rremote: Processing changes: refs: 1, new: 1 \rremote: Processing changes: refs: 1, new: 1, done \nremote: \nremote: SUCCESS \nremote: \nremote:
http://15a45d4cba1a/c/gerrit-test/+/47
 hello [WIP] [NEW] \nremote:"

Please add handling for this as well
```

PR #10705, opened 2025-10-17
fix: Correct the params type to account for the serialization

estib-vega:
Align the expected params that need to be serialized


Byron:
@estib-vega Did I break something here? I ask because I did do some 'cleanup' for serde related derives recently and I was under the assumption that it would blow up at compile time if I get it wrong. But I am asking just to make sure 😅. Thanks.

estib-vega:
@Byron no no, this was totally a self-inflicted wound. 
You're good!

PR #10706, opened 2025-10-17
fix(preferences): default snapshot_threshold to 20 when unset

PavelLaptev:
image

### DIFF
--- a/apps/desktop/src/components/PreferencesForm.svelte
+++ b/apps/desktop/src/components/PreferencesForm.svelte
@@ -47,7 +47,7 @@
 						type="number"
 						width={100}
 						textAlign="center"
-						value={project.snapshot_lines_threshold?.toString()}
+						value={project.snapshot_lines_threshold?.toString() ?? '20'}
 						minVal={5}
 						maxVal={1000}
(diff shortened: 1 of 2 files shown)


PavelLaptev:
@krlvi I just hardcoded the default threshold amount in the frontend, but there’s the same hardcoded default value in Rust.  
Is there a good way to bind these defaults and use a single source?

```rs
pub fn snapshot_lines_threshold(&self) -> usize {
    self.snapshot_lines_threshold.unwrap_or(20)
}
```

krlvi:
the app doesnt do this anymore so this can be removed from the UI entirely

PavelLaptev:
The branch was corrupted and is in a weird state now. I'll leave it unchanged so we can investigate the issue later. I'll create another PR for this.

Byron:
@PavelLaptev Please keep a copy of the folder around if still possible 😅, it's probably something to investigate.

PavelLaptev:
@Byron yep. I wouldn't remove the branch :-)

Byron:
Great, thanks, let's see what we can sort out in office then.

PR #10716, opened 2025-10-18
Improve login form in the app

PavelLaptev:
## Add Delayed Cancel Button to Sign-in Flow
Improves sign-in UX by showing the cancel button only after a timeout, preventing accidental cancellations while still allowing users to abort long-running operations.

Flow:
1. User clicks "Log in / Sign up" → button shows loading spinner and prevents multiple clicks
2. After 6 seconds (if still loading) → cancel button appears
3. User can cancel operation if needed (cancel button also prevents double-clicks)

Same behavior for "Copy login link" with 3-second timeout.

Benefits: Prevents accidental cancellations, provides escape route for slow operations, prevents multiple clicks/double submissions, cleaner loading UX.

https://github.com/user-attachments/assets/35b586b7-0783-4a53-8d23-03b23128681f


Byron:
Loving your profile picture :D! I never knew cuddling with a light string has such great effect.

Oh, and the improvement is great as well, the feature did feel half baked before and now has more gravity to it.

PR #10729, opened 2025-10-20
fix: Migrate the project git dir before validation

estib-vega:
Move up the migration step for the project’s git dir so that it happens before the validation.

This fixes the issue in which existing projects would not be ‘found’ even if they were still there

### DIFF
--- a/crates/gitbutler-project/src/controller.rs
+++ b/crates/gitbutler-project/src/controller.rs
@@ -207,6 +207,7 @@ impl Controller {
         #[cfg_attr(not(windows), allow(unused_mut))]
         let mut project = self.projects_storage.get(id)?;
         // BACKWARD-COMPATIBLE MIGRATION
+        project.migrate()?;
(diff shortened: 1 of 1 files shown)


estib-vega:
Hey @Byron does this look ok?

I could not open any of my existing projects on dev because it attempted to `open_isolated` before the `git_dir` would be set

Byron (commented):
😱Thanks so much, and I am very sorry for the hassle!
It looks like I really messed up there after the big refactor, while also now learning that none of our tests catches it. That… does make sense as they'd have to be using a `project.json` from an older version, and try to open it with the new one.

This is probably also why I didn't notice it - all the projects I used were already migrated.

In any case, I am glad this was caught before anyone else will notice…releasing a new nightly now (if not done).

Edit: looks like a recent enough one was already published.

estib-vega:
No hassle :)

PR #10739, opened 2025-10-20
Codegen attach files

PavelLaptev:
https://github.com/user-attachments/assets/9144aa44-d8db-4b89-9e0d-154766d246a8

https://github.com/user-attachments/assets/23935d91-1bb9-409a-991c-0f6894b4bd3e

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -966,6 +966,7 @@ name = "but-claude"
 version = "0.0.0"
 dependencies = [
  "anyhow",
+ "base64 0.22.1",
  "but-action",
  "but-broadcaster",
  "but-core",
--- a/apps/desktop/src/components/CardOverlay.svelte
+++ b/apps/desktop/src/components/CardOverlay.svelte
@@ -64,9 +64,6 @@
 
 <style lang="postcss">
 	.dropzone-wrapper {
-		--dropzone-fill: oklch(from var(--clr-scale-pop-50) l c h / 0.1);
-		--dropzone-stroke: oklch(from var(--clr-scale-pop-50) l c h / 0.8);
-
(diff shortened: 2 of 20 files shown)


krlvi:
Is it attaching the file path or the file contents?

PavelLaptev:
@krlvi yeah, I didn't get how to make Claude work only with file paths, so it's attaching the content as well. Not gonna merge this.
I've asked @Caleb-T-Owens to take a look at this to see if we can analyze files using only paths.

PavelLaptev:
We can completely change how it works, but the frontend here is ready to use. 
cc @mtsgrd

krlvi:
In my prompts I make references to files in the project very frequently. 

Actually almost all of my prompt mention something like "use the api functions found in `crates/but-api/src/commands/stack.rs`

It's good at reading the file by itself. 

At the same time, improving the ergonomics and visuals would definitely be nice. 

When it comes to files I can think of 3 different types of references:

1. A file that is in the project (like my example, I use this very often)
2. A file that is not in the project (like examples.json or something). In this case bringing the file content is probably needed if Claude can't access outside the repo. I can see the value of this but haven't had the need yet. 
3. A file diff that is either committed or not committed. This would refer to the changes made to a file in a specific commit - that would be something that can be dragged and dropped in the app. The way I imagine that would be used is "Look at how the database stuff was refactored in <project.rs> - do things like it"


For this use case 3. I have previously given it a commit hash, telling it something like "look I how I did stuff here, please do it that way", but I can see it's also valuable to be specific to the file level too. 


---

Of course pictures are entirely different - I can totally see how nice those will be when dealing with frontend related prompts.

Issue #2880, opened 2024-02-22
Way to delete your account


schacon:
Whoopsie. We don't actually have a way to delete your own account once you create one other than asking us. We all hate those fucking companies, so let's get some sort of "delete my account" self service thing somewhere.

This may require us to have a "go to gitbutler cloud" type button under your user settings to get to where you can do it easily.

To be GDPR compliant, we'll have to delete everything associated with the user we can think of. Mostly this is the user record which has PII data and the associated Auth0 record, plus any backups or project data the user has sent to us.

Byron:
Definitely still needed.

MikaelFangel:
Any updates on this topic?

Byron:
There is now a button to delete the user account. However, it's not yet complete as it will only delete associated data in the database and in Sentry, but not yet in Posthog or backups. The latter might be acceptable, but if so I think we should be clear about the retention.

Still work to be done but a step in the right direction.

PR #10749, opened 2025-10-21
Web Profile Page: Delete the account

estib-vega:
Add a button in the profile page that allows the user to delete their account.

Of course, add a confirmation modal as well

Screenshot 2025-10-21 at 09 54 01

Screenshot 2025-10-21 at 09 54 09

### DIFF
--- a/apps/web/src/lib/user/userService.ts
+++ b/apps/web/src/lib/user/userService.ts
@@ -90,6 +90,15 @@ export class UserService {
 		this.user.set(undefined);
 	}
 
+	async deleteAccount(): Promise<void> {
+		try {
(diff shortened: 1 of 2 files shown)


Byron:
This is awesome! Would that fix #2880 ?

Also, is it true that no data remains, not even in backups or posthog?

If it's true, perfect, but how can we assure this remains true? This is more of a risk mitigation question as we wouldn't want to accidentally hold on to PII data, and then just as accidentally see it exposed.

estib-vega:
This is only a partial fix. It deletes it from our DB and from Sentry. Need to add a deletion mechanism for Posthog

PR #10751, opened 2025-10-21
remove frontend snapshot lines

PavelLaptev:


PavelLaptev:
@krlvi I removed the snapshot threshold from the frontend only. Does PR look good to you?

PR #10756, opened 2025-10-21
but: Status review

estib-vega:
Display the associated revies for the branches in the status


estib-vega:
Follow-ups:
1. Improve performance
We shouldn't fetch all open PRs, but only PRs relevant to the applied branches
2. Show more information about the associated PRs

PR #10775, opened 2025-10-22
Fix codegen page

mtsgrd:
The distinction between returning undefined or null for queries used in 
`combineResults` is important. If undefined the `children` template is 
never rendered.


mtsgrd:
@krlvi fyi - looks like the sessions are created on demand now.

PR #10781, opened 2025-10-23
do not log secrets

Byron:
Do not log secrets, and let's be mindful whenever something has the `Sensitive` or indicates that secrets are handled.

Related to #10777 .


Byron:
@krlvi I'd consider this an incident as GitLab tokens are definitely put into logs now for a couple of releases.

Additionally, I think we should not expose secrets in the UI as if they were normal strings, but I will create a separate issue for that.

Screenshot 2025-10-23 at 04 36 12

Screenshot 2025-10-23 at 04 37 27

In theory, the next release should wipe all application logs.

PR #10794, opened 2025-10-24
spike: avoid keychain popups on macOS when 'but' accesses GUI keychain entries

Byron:
### Tasks

* [x] identifier cleanup
* [x] prep `but` for re-use in tauri
* [x] make `gitbutler-tauri` a valid `but` executable
    -  [x] symlink the tauri executable instead (what about Windows though? It's not used there at all)
* [x] validate on new nightly build 
* [x] remove extra binaries from distribution, but on non-windows only.
* [x] validate distribution
* [x] mini-migration - play it safe, no elevated permissions or popups.


Byron:
@krlvi With the changes in this PR, I have made it so that neither `but-testing` nor `but` are included in the distribution, on all platforms. This also works on Windows which will need some love if people want to use the CLI anyway.

This will lead to a 75MB reduction in the installation size on MacOS, down to 100MB, and I will see how big the distributable is.

Please note that locally, we still build the `but` binary and symlink it as it's more convenient in dev mode, and the whole thing can be feature-toggled with Cargo.  I hope to have more data later once the builds are finished.

krlvi:
This is really cool. I wonder if this would make it easier to implement a `but .` functionality

krlvi:
Btw - one thing to consider is we might need to update our brew setup (but can be installed that way too)

Byron:
> This is really cool. I wonder if this would make it easier to implement a `but .` functionality

I think these are independent. The reason I didn't add this just for fun is that it needs some frontend support, and multi-window support, there is plenty of unknowns for me at least.

> Btw - one thing to consider is we might need to update our brew setup (but can be installed that way too)

Right, this one as well. It looks like the formula lives elsewhere though.
On that note, all users of the CLI would lose access to it and have to click the install button again. We could have a migration for that though.

Asset sizes are way down now as well, pretty neat.

Screenshot 2025-10-24 at 20 43 54

krlvi:
Do users who have previously installed the CLI (created the symlink) need to do something or the mini migration makes it just work?

Byron:
It should just do the right thing. 

The migration checks if the symlink is there *and* points to a non-existing file. Only then it will replace it with the new target. So it shouldn't ever auto-install the CLI or overwrite what people chose.

krlvi:
> It should just do the right thing.
> 
> The migration checks if the symlink is there _and_ points to a non-existing file. Only then it will replace it with the new target. So it shouldn't ever auto-install the CLI or overwrite what people chose.

Hmm, maybe something didnt migrate, after updating, I did not have a `but` in `/usr/local/bin/`. Well i reinstalled it and it's fine now

Byron:
That's strange, and maybe something to investigate. I tried it with 'broken 'but' link' and 'valid but link', and both cases worked as expected.
If there is no but link at all, it will also do nothing. Maybe we can figure this out.

PR #10803, opened 2025-10-24
Forge Settings

estib-vega:
Create a build-type-specific storage for forge-related settings.
Currently, the only information stored there is the GitHub account information.

Now:
- Known accounts and their metadata are stored in the file-based storage
- Secrets (access tokens) are stored plainly in the secret storage

This includes a migration step that runs only once and cleans up the previous secret that kept metadata stored.

### DIFF
--- a/crates/but-forge-storage/src/storage.rs
+++ b/crates/but-forge-storage/src/storage.rs
@@ -1,7 +1,8 @@
-use std::path::PathBuf;
+use std::{path::PathBuf, sync::Mutex};
 
-use crate::settings::ForgeSettings;
+use crate::settings::{ForgeSettings, GitHubAccount};
 use anyhow::Result;
+use serde::{Deserialize, Serialize};
 
 const FORGE_SETTINGS_FILE: &str = "forge_settings.json";
 
@@ -18,13 +19,15 @@ impl Storage {
     }
 
     pub fn read(&self) -> Result<ForgeSettings> {
-        match self.inner.read(FORGE_SETTINGS_FILE)? {
+        let settings = match self.inner.read(FORGE_SETTINGS_FILE)? {
             Some(settings) => {
                 let settings: ForgeSettings = serde_json::from_str(&settings)?;
-                Ok(settings)
+                settings
             }
-            None => Ok(Default::default()),
-        }
+            None => Default::default(),
+        };
+
+        self.migrate_accounts(settings)
     }
 
     pub fn save(&self, settings: &ForgeSettings) -> Result<()> {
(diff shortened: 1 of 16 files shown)


estib-vega:
cc @Byron

crates/but-forge-storage/src/storage.rs
```diff
@@ -0,0 +1,145 @@
...
+                    new_settings.github.known_accounts.push(migrated_account);
+                }
+            }
+            self.save(&new_settings)?;
+            return Ok(new_settings);
+        }
+
+        Ok(settings)
+    }
+}
```
estib-vega, line 62: This is the only part that feels a bit weird to me, but it seems to work.
It's also needed only for the people that have been using the nightly build since two days ago, so we can remove it after a couple of days

estib-vega:
Related to nice conversations in https://github.com/gitbutlerapp/gitbutler/issues/10784

crates/but-forge-storage/src/storage.rs
Byron: Would this migrate accounts as they are currently in `0.16.10`? I thought there we still only have a single GitHub secret.

crates/but-forge-storage/src/storage.rs
```diff
@@ -0,0 +1,145 @@
+use std::{path::PathBuf, sync::Mutex};
+
+use crate::settings::{ForgeSettings, GitHubAccount};
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+
+const FORGE_SETTINGS_FILE: &str = "forge_settings.json";
```
Byron, line 7: Is this needed because it's our first 'per build kind' settings file?
If so, I wonder how hard it would be to add a build-kind `settings.json` and merge it into the one that is shared among all builds. We have the functions for that, and adding it there seems more sustainable (but also more complex with a chance for a rabbit hole).

Byron (commented):
I just took a quick look for now, so I can share only some vague ideas and coments for now.

Generally, and even though there is no issue with creating a new crate for everything, I think here it would make sense to have one crate that deals with authentication, or forges, and put things there. Maybe that way more can be shared, and it's a bit clearer then how we do things.

My approach here is like this:

* create a new crate if the dependencies of it really need to be 'contained'
* prefer to stay 'on topic' of existing crates and extend them, also to re-use their test utilities.
* crate new crates if one expects it to get big or be its very own and unique topic.

Quite fuzzy, still, and in `gitoxide` it's clearly very 'sub-system' based, but here I really like to let crates grow more, and split if there is a good reason for it, rather than having many small crates.

estib-vega:
> I just took a quick look for now, so I can share only some vague ideas and coments for now.
> 
> Generally, and even though there is no issue with creating a new crate for everything, I think here it would make sense to have one crate that deals with authentication, or forges, and put things there. Maybe that way more can be shared, and it's a bit clearer then how we do things.
> 
> My approach here is like this:
> 
> * create a new crate if the dependencies of it really need to be 'contained'
> * prefer to stay 'on topic' of existing crates and extend them, also to re-use their test utilities.
> * crate new crates if one expects it to get big or be its very own and unique topic.
> 
> Quite fuzzy, still, and in `gitoxide` it's clearly very 'sub-system' based, but here I really like to let crates grow more, and split if there is a good reason for it, rather than having many small crates.

Hey thanks for the feedback. Here's the justification for a separate crate:

There are three main categories on the forge architecture:
- (**A**) Forge-Provider-specific crate: (currently only **but-github**, but soon enough **but-gitlab** and others)
- (**B**) Forge-agnostic crate: **gitbutler-forge**, this basically exposes the API functions that care only about the project Id, and internally figure out which forge provider to use
- (**C**) Forge-storage: **but-forge-storage** (new kid on the block)

**B** depends on **A**, of course, in order to direct the requests towards the correct forge-specifc logic.
Both of them depend on **C**, though.

It sadly can't be integrated into **B** since **A** also depends on it, and that would create a circular dependency

crates/but-forge-storage/src/storage.rs
estib-vega: No, this just migrates the users from the last mechanism. Basically this step is just for us the developers running nightly every day.

The migration of users from `0.16.10` is handled in another place already

crates/but-forge-storage/src/storage.rs
estib-vega: This is not the first build-kind specific file. Projects are also build-kind specific.

I don't like the idea of merging this information with the application settings for the following reasons:
- We've already established it as a manually editable file since it allows people to add e.g. CSP allow-list items there, [in the docs](https://docs.gitbutler.com/troubleshooting/custom-csp#adding-a-custom-csp) and we don't want them looking at the accounts and messing stuff up
- Not merging it into the settings file makes allows us to not have to migrate people into the build-kind specific settings yet. I also think we *should* move towards that, but at the moment it's not a pressing issue and would like to do that deliberately and after aligning with the rest of the team

Byron: Thanks for elaborating, and… right, `projects.json`! How could I not see that.

Something I am a bit worried about is the precedent this might create. You want to store something? Make your own file because… we are lacking a concept for this?

Is there something we could do now to make this easier in future and not create something we are likely going to change later once 'the concept' has arrived?

That might be something to bring up sooner than later with the rest of the team.

Byron:
Thanks! Another argument for separate forge-specific crates is that they might need different dependencies to access their APIs.
Something I am worried about there is that we pull in dependencies quite indiscriminately, in the sense that in the end, they make HTTPS requests. Yet, they could use vastly different crate versions or crates for that, and we end up with a lot of duplicated functionality within the dependency tree.
Ultimately, we definitely don't want more in the `gitbutler-` namespace as it's supposed to go away, while also avoiding creating `but-` crates like it's the `gitbutler-' namespace. This might just be my internal guideline though, but to me it seems `gitbutler-forge` shouldn't exist or be migrated.

In my ideal world, there is a `but-forge` crate that is able to efficiently make HTTP API calls against various forges and probably deal with their AUTH* requirements, and it's all very, very similar.
This is not me requesting a change, just to share a vision (independently of how unrealistic it might be 😅).

PR #10804, opened 2025-10-24
Add file auto-complete to codegen input

mtsgrd:


krlvi (approved):
Just wanted to follow up since we discussed this on a call but at the time I had not seen the implementation. 

This [ignore](https://docs.rs/ignore/latest/ignore/index.html) library is an excellent choice for the task! I didn't realize this was by `BurntSushi` - his work is really great.

Also just relized that `igrnore` is a component for `ripgrep` which is very popular :)

mtsgrd:
@krlvi that was Claude's choice, but I'm glad you approve. 😏 Let me know if you have any feedback after giving it a spin?

krlvi:
It's scary how good Claude is...

krlvi:
I thought, damn, this code is pretty clean

Byron:
Just for completeness, let me add that `gix` also has a dirwalker with 'ignore' capability, and it's what we'd ultimately want to use as it's the most correct implementation of `.gitignore`. It also is a dependency we already have. Mattias and I already talked about this and left with me eventually replacing `ignore`.

I may add a concern that we are loading up on dependencies quite a bit these days, probably facilitated by codegen as well, and it's going to cost us not only in compile-time.

Issue #10188, opened 2025-09-09
Gerrit forge integration


torarnv:
I'd like to see Gerrit being one of the supported forges, including pushing branches for review. I might be able to work on this, but first wanted to open up a discussion if this has been looked into, and what considerations I should take if looking at this. Thanks!

Byron:
Thanks a lot for sharing, it's much appreciated!

This is the first issue about this topic, and I don't recall anything about it either, except that supporting Gerrit might be more involved as it requires its own form of a change-id?
Pushing branches there might also be quite different, even though maybe in the end its just (force)pushing to one particular ref?

With all that said, my biggest concern would be the change-id, as the existing forge abstraction should be strong enough to adapt to Gerrit. Of course, there will be a lot that I am missing 😅.

torarnv:
A few notes after brainstorming with @krlvi 

- Gerrit has the concept of pushing to normal refs (branches), just like any other git server, so we should not necessarily equate  "pushing to Gerrit" as "create reviews". Just like GitHub/GitLab, those are in principle two steps: push, and create review
- The most natural mapping from GB's virtual branches are a set of individual changes in a Gerrit [relation chain](https://gerrit-review.googlesource.com/Documentation/concept-changes.html#related-changes). We don't want to treat each gerrit change (commit) as a separate virtual branch in GB
- Gerrit changes can be tagged with a "topic", which would map well to the named virtual branch
- GB currently associates a review with a virtual branch, based on the GitHub/GitLab PR model, but in Gerrit reviews are per commit. Ideally we'd show a link/button in the UI for each of the commits, pointing to the associcated gerrit change, but as a stepping stone we can have a "review link" for the virtual branch that points to the gerrit topic, or the last change in the relation chain
- For pushing and pulling changes we probably want to base things off the Git protocol, like we do for other forges, but it may be too unreliable to try to create reviews using the `refs/for/foo` magic ref, as the protocol doesn't (?) give us the resulting `refs/changes/123` ref. As an alternative we should consider the [Gerrit REST API](https://gerrit-review.googlesource.com/Documentation/rest-api-changes.html), which hopefully allows creating changes from existing refs (?)

DanieleSassoli:
As someone that is quite involved with the Gerrit world, I thought I'd chime in here.

> Gerrit changes can be tagged with a "topic", which would map well to the named virtual branch

[...]

PR #10806, opened 2025-10-24
Automatically enable gerrit_mode if remote is a gerrit one

krlvi:
Related to https://github.com/gitbutlerapp/gitbutler/issues/10188

### DIFF
--- a/crates/but-core/src/settings.rs
+++ b/crates/but-core/src/settings.rs
@@ -142,6 +142,12 @@ pub mod git {
                     if sign_commits { "true" } else { "false" },
                 )?;
             };
+            if let Some(gerrit_mode) = self.gitbutler_gerrit_mode {
+                config.set_raw_value(
+                    &GITBUTLER_GERRIT_MODE,
+                    if gerrit_mode { "true" } else { "false" },
+                )?;
+            };
             if let Some(signing_key) = &self.signing_key {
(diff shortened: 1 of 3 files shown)


krlvi:
@Byron I was getting compile errors in CI (the second commit) but not while building locally.... Granted I was just running the app. So I added the `"blocking-network-client"` feature flag to make this work. 
I find it curious... how come it worked when just building the client? I am guessing the feature flag gets inherited transitively from somewhere... 

Anyways, I thought it's probably good to ping you just to make sure I am not doing something insane. I am actually quite pleased that I was able to do it all with `gix` :)

Byron:
I looked at the code and wouldn't change a thing - it really does look pretty good :).
Regarding the "blocking-network-client" feature, I didn't find it in use here so it shouldn't just have activated for you locally. It's unclear to me how that could ever compile for you without it.

**Important note**: this code now has access to only `file://`, `git://` and `ssh://`. For HTTP(S) you need "blocking-http-transport-curl" (as replacement for the current one). This means that currently, if the remote is HTTPS, the code would not be able to figure out it's a gerrit remote, even if it is.

krlvi:
Thanks!

> This means that currently, if the remote is HTTPS, the code would not be able to figure out it's a gerrit remote, even if it is.

How can i work around this?

PR #10810, opened 2025-10-24
ui-fixes

PavelLaptev:
Preview lanes now take the full height. This was needed for:
- Making preview lanes always look the same, which helps from a UX perspective
- Enabling future features like a "slide-in" preview, where having 100% height is necessary
- Simplifying conditions, since the codegen should be a full-height component

image

image

image

image

---

Summary
- fixes and refinements across multiple components.

What changed
- Compute and display rounded context usage; hide kebab menu when value is zero.
- Use monospace font only when needed (restore regular font for typical messages).
- Unify CSS variable and remove extra wrapper for codegen messages.
- Update drawer to take full height when in codegen mode or when no file preview is present.

Notes
- These are visual/UX refinements; no functional behavior changes beyond display adjustments.

### DIFF
--- a/apps/desktop/src/components/ChangedFiles.svelte
+++ b/apps/desktop/src/components/ChangedFiles.svelte
@@ -75,6 +75,8 @@
 	{noshrink}
 	bottomBorder={changes.length > 0}
 	persistId={`changed-files-drawer-${projectId}-${stackId || 'default'}`}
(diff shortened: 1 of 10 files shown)


PavelLaptev:
cc @krlvi

krlvi:
nice!

PR #10818, opened 2025-10-26
various Cargo updates and upgrades

Byron:
Maybe that helps with build times, or at least prevents bugs that are already fixed.

### Tasks

* [x] updates
* [x] upgrades

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 *.sh eol=lf
+/Cargo.lock merge=binary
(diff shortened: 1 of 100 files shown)


Byron:
I am CC'ing you @krlvi and @Caleb-T-Owens  because you may have been very involved in the MCPs in this codebase, and would be able to validate it's still working after I upgraded `rmcp` to 0.8 (from 0.1).  I hope it works, but if it's broken beyond repair, it could be reverted with fe81b5be719d4c3db9582349fe184d97df2f732f .
I also double-checked with #10819 and AI seems to do most of the same.

PR #10820, opened 2025-10-26
Simplify codegen attachment struct

mtsgrd:
Use flat `type` and properties rather than `type` and `subject` pair.

### DIFF
--- a/crates/but-claude/src/lib.rs
+++ b/crates/but-claude/src/lib.rs
@@ -69,36 +69,21 @@ pub enum ClaudeMessageContent {
     GitButlerMessage(GitButlerMessage),
 }
 
-/// File attachment with full content.
-#[derive(Serialize, Deserialize, Debug, Clone)]
-#[serde(rename_all = "camelCase")]
-pub struct AttachedHunk {
-    pub path: String,
-    pub start: usize,
-    pub end: usize,
-}
-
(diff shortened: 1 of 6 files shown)


mtsgrd:
@Caleb-T-Owens since you and I talked about this irl, are you ok with this change?

crates/but-claude/src/lib.rs
```diff
@@ -69,36 +69,21 @@ pub enum ClaudeMessageContent {
...
+        path: String,
+        start: usize,
+        end: usize,
+    },
+    File {
+        path: String,
+    },
+    Commit {
+        commit_id: String,
+    },
```
Caleb-T-Owens, line 86: If we have the named structs, it should still be flattened out if the `content = "subject"` property is omitted.

It _can_ be nice to have named structs so you can have `impl`s on them individually

krlvi:
This may be something nice to do across the board for other Structs - the flattening of type + the fields.
It's not necessary to inline the struct fields though, as Caleb mentioned.
Actually it would be really good to have a consistent way of serializing enums

Caleb-T-Owens:
@krlvi I started a discord thread on this - there are some downsides which might make it impossible to apply across the board https://discord.com/channels/1060193121130000425/1432012443080589435/1432012448306696243

PR #10831, opened 2025-10-26
Use adjacent tagging for serialized claude messages

mtsgrd:
Instead of `{type: 'claudeOutput', subject: {..}}` this code now serializes messages into `{type: 'claudeOutput', ...}`.


crates/but-claude/src/db.rs
```diff
@@ -138,10 +139,23 @@ pub fn list_messages_by_session(
...
+                Err(e) => {
+                    tracing::warn!(
+                        message_id = %message_id,
+                        error = %e,
+                        "Failed to deserialize claude message, skipping"
+                    );
+                    None
+                }
+            }
+        })
```
mtsgrd: @krlvi fyi - about ignoring read errors for `list_messages_by_session`

PR #10833, opened 2025-10-27
feat(codegen): make drop handler async and add multiple files support

PavelLaptev:
https://github.com/user-attachments/assets/60fb7329-57a0-4c19-b40f-c3daf8259636


PavelLaptev:
@mtsgrd added multiple files support

PR #10834, opened 2025-10-27
feat(but): Add shell completion generation support

nshcr:
https://github.com/user-attachments/assets/806d70d4-b421-48c2-97b2-a1b99ce3b3d5

This PR adds command-line completion support for the `but` command based on [`clap_complete`](https://docs.rs/clap_complete/latest/clap_complete/).

The code was primarily generated by Claude Sonnet 4.5, I reviewed and tested it, and made several adjustments after checking the `clap_complete` documentation and examples from similar projects (such as [Typst](https://github.com/typst/typst/blob/main/crates/typst-cli/src/completions.rs)) to ensure the generated code works correctly.

Supported shells for completion generation:

- bash
- zsh
- fish
- elvish
- powershell

Next step: integrate the completion setup into shell environments via installation scripts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -776,6 +776,7 @@ dependencies = [
  "but-worktrees",
  "chrono",
  "clap",
+ "clap_complete",
  "colored",
  "command-group",
  "gitbutler-branch",
@@ -1673,6 +1674,15 @@ dependencies = [
  "clap_lex",
  "strsim",
  "terminal_size",
+]
+
+[[package]]
+name = "clap_complete"
+version = "4.5.59"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2348487adcd4631696ced64ccdb40d38ac4d31cae7f2eec8817fcea1b9d1c43c"
+dependencies = [
+ "clap",
 ]
 
 [[package]]
--- a/crates/but/Cargo.toml
+++ b/crates/but/Cargo.toml
@@ -23,6 +23,7 @@ serde.workspace = true
 tokio = { workspace = true, features = ["rt-multi-thread", "io-std"] }
 strum = { version = "0.27", features = ["derive"] }
 clap = { version = "4.5.48", features = ["derive", "env", "wrap_help"] }
+clap_complete = "4.5.59"
 chrono.workspace = true
 bstr.workspace = true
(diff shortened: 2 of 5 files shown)


Byron (approved):
That's fantastic, thanks so much!


I also finally took the liberty to embrace UTF8 in terminals and change this…

Image

to this…

Image

For a sleeker look.

nshcr:
Thanks for helping refactor and merge it!

When I was checking the Gitoxide completions command, I noticed how they handled the default shell, but for some reason, I totally forgot to add that part myself. 🥲

Also, honestly, I couldn't tell the difference between the two screenshots you posted without looking at the code. 🤣

Lastly, regarding adding command completion support in the installation script, I think it would be quite easy to implement through Homebrew.

However, considering how `but` integrates with `gitbutler-tauri`, and that GitButler is distributed as a cask rather than a formula, this might be beyond my abilities.

Gitoxide:
https://github.com/Homebrew/homebrew-core/blob/0afaef59e16e639f65cb4b3c8807de2297ee9e16/Formula/g/gitoxide.rb#L28-L33

Typst:
https://github.com/Homebrew/homebrew-core/blob/0afaef59e16e639f65cb4b3c8807de2297ee9e16/Formula/t/typst.rb#L31-L38

Byron:
> Also, honestly, I couldn't tell the difference between the two screenshots you posted without looking at the code. 🤣

Since I feel so strongly about that ellipsis apparently, here is a re-focused screenshot 😁.

Screenshot 2025-10-27 at 07 56 41

Regarding homebrew, I realise the that with the `but` symlink speciality, we'd have to update the [Cask](https://github.com/Homebrew/homebrew-cask/blob/HEAD/Casks/g/gitbutler.rb). @krlvi mentioned this as well, and it would have to be done once the new stable release of GitButler is available. This will be hard to time, but I think it's OK to update the cask earlier as it would work, at least, despite having a symlink to a `gitbutler-tauri` binary that doesn't act like `but` just yet.

I am also thinking to keep an authoritative copy of the cask along with instructions on how to validate it in this repo, as I am also not familiar with it.

krlvi:
Thank you @nshcr for this PR, this is really awesome!

And re: brew, indeed lets time an update approx as we release

PR #10837, opened 2025-10-27
build: optimize frontend build size

nshcr:
- Previous build: [build.zip](https://github.com/user-attachments/files/23162677/build.zip)
- This PR build: [build-min.zip](https://github.com/user-attachments/files/23162679/build-min.zip)

Build command: `pnpm build:desktop -- --mode nightly`

Uncompressed build: 46.5 MB -> 25.5 MB (~ 45% reduction)
ZIP compressed build: 24.8 MB -> 7.4 MB (~ 70% reduction)

For more details: https://discord.com/channels/1060193121130000425/1432062926625177600/1432292511551979592

References: https://svelte.dev/docs/kit/adapter-node#Options-precompress

### DIFF
--- a/apps/desktop/svelte.config.js
+++ b/apps/desktop/svelte.config.js
@@ -12,7 +12,7 @@ const config = {
 			pages: 'build',
 			assets: 'build',
 			fallback: 'index.html',
-			precompress: true,
+			precompress: false,
 			strict: false
 		})
 	},
--- a/apps/desktop/vite.config.ts
+++ b/apps/desktop/vite.config.ts
@@ -60,7 +60,7 @@ export default defineConfig({
 		// minify production builds
 		minify: !process.env.TAURI_ENV_DEBUG ? 'esbuild' : false,
 		// ship sourcemaps for better sentry error reports
-		sourcemap: 'inline'
+		sourcemap: true
 	},
 	test: {
 		includeSource: ['src/**/*.test.{js,ts}'],
--- a/apps/web/vite.config.ts
+++ b/apps/web/vite.config.ts
@@ -34,7 +34,7 @@ export default defineConfig({
 		setupFiles: ['./vitest-setup.js']
 	},
 	build: {
-		sourcemap: 'inline'
+		sourcemap: true
 	},
 	resolve: {
 		alias: {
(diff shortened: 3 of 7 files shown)


Byron:
Thanks a million!

Is this something that kicks in automatically when building a release? If so, I could push this commit into a branch here and build a nightly from it. The artefact size of svelte should reveal the change, and the overall build product should be smaller.

I also wonder how this affects local development, if at all. So in any case, @mtsgrd and @estib-vega would probably need to take a look.

Ideally, the only thing that magically gets smaller is the bundles we distribute - if that's not the case because it's compressed anyway, then applying compression here might make no meaningful difference.

nshcr:
@Byron Yes, you can cherry-pick this commit into a branch and trigger a nightly build to see if the final build size improves.

The `precompress` setting essentially follows web development practices rather than being intended for local app distribution.

It generates both a gzip and a Brotli-compressed version for each code file, meant for web servers to serve smaller bundles over HTTP.

Using inline `sourcemap` also increases file size, since the `.js.map` content is base64-encoded and appended to the end of the original file — this typically increases the total size by about one-third.

It doesn't affect local development.

Byron:
Thanks for the update!

Here is a release build, which shows a reduction of the Svelte bundle to ~7MB, from 24MB.

https://github.com/gitbutlerapp/gitbutler/actions/runs/18863546998

A local build distributable for macOS clocks in at 78MB, with the bundle going down to 33.1 MB, a notable improvement.

Screenshot 2025-10-28 at 05 02 23

nshcr:
I realized that this PR also modified some web-related configurations.

For web distribution, setting `precompress` to `true` and `sourcemap` to `inline` is essentially a trade-off — it increases bundle size in exchange for fewer network requests and faster individual response times.

I'm not sure what the current build setup for the official website is, so I'll leave it to you to decide whether to keep this change.

However, I believe that for desktop app distribution, optimizing for frontend network requests is completely unnecessary.

Byron:
This is a huge improvement!

* bundle size: 53,7MB - 36,5MB  = 17,2 MB (33%) saving
* application size: 104,2MB - 87MB = 17,2 MB (~15%) saving

This is exactly the reduction of the svelte bundle size, which seems to be attached to the binary verbatim.

Screenshot 2025-10-28 at 05 25 16

Since the publishing step seems to be fine, and since the application works, I think this is good to merge.

@estib-vega @mtsgrd Please double-check that source-maps don't have problems now that there went from `inline` to `true`. If so, please feel free to revert the commit, even though I'd be glad if its effects could stay.

PR #10839, opened 2025-10-27
Fix the linking issue introduced by a dependency update in 78a5cb236add69e11f96d2e9d5c4d32b2bb5d68e

Byron:
See if it's still an issue with `cargo run -p but-server` and this commit reverted.


Byron:
CC @mtsgrd this seems to fix the issue. More context is here: https://discord.com/channels/1060193121130000425/1432052722294587472/1432344447034589224

Byron:
I am creating a new nightly [here](https://github.com/gitbutlerapp/gitbutler/actions/runs/18841391074) just to be sure that this change doesn't mess with macOS on non macOS 26.

PR #10840, opened 2025-10-27
faster Rust testing with Tauri split

Byron:
- **Remove `rust-doc` job as we don't actually care about rust-docs (yet).**
- **Split tauri tests into their own job to prevent cache trashing.**


Byron:
Note that this will only come to fruition once caches are recreated.

PR #10852, opened 2025-10-28
build: Add cross-compilation support for macOS x86_64 target

nshcr:


Byron:
Thanks for getting this started!

Let's try it and see what happens. Here is the [publish job](https://github.com/gitbutlerapp/gitbutler/actions/runs/18873619754).

My feeling is that the artefact suffixes *might* not match, but then again, I also don't know were these are actually set.

Screenshot 2025-10-28 at 12 41 39

But in 30m we know more.

Byron:
Closing in favor of #10856.

PR #10853, opened 2025-10-28
feat(status): prefer first line of message in non-verbose mode

krlvi:


Byron:
It's strange that you had to update that screenshot, and that nothing seems changed. Maybe metadata? 

Screenshot 2025-10-28 at 14 15 26

Something to watch out for I think.

PR #10854, opened 2025-10-28
fix: Correctly handle the key events on file suggestions

estib-vega:
Correctly handle the file suggestions events so that they don't swallow the rest of the keyboard inputs

### DIFF
--- a/apps/desktop/src/components/codegen/CodegenInput.svelte
+++ b/apps/desktop/src/components/codegen/CodegenInput.svelte
@@ -11,11 +11,13 @@
 		CodegenHunkDropHandler
 	} from '$lib/codegen/dropzone';
 	import { newlineOnEnter } from '$lib/config/uiFeatureFlags';
+	import { FILE_SERVICE } from '$lib/files/fileService';
 	import { showError } from '$lib/notifications/toasts';
 	import { inject } from '@gitbutler/core/context';
 	import { Tooltip, AsyncButton, RichTextEditor, FilePlugin } from '@gitbutler/ui';
 	import { fade } from 'svelte/transition';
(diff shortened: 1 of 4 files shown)


estib-vega:
cc @mtsgrd 
I changed a couple of things so that the FileSearch component is more of a pure style component now.

Happy to iterate on this, but for now I'd just merge this fix and then change it as needed

apps/desktop/src/components/codegen/CodegenInput.svelte
```diff
@@ -52,6 +54,70 @@
...
+	function exitFileSuggestions() {
+		fileSuggestionsPlugin?.exitFileSuggestions();
+	}
+
+	function onFileSuggestionUpdate(update: FileSuggestionUpdate) {
+		if (update.loading) {
+			loadingFileSuggestions = true;
+			indexOfSelectedFile = undefined;
+			return;
+		}
```
estib-vega, line 76: All these new added variables and functions could be abstracted to a dedicated class, similar to how we handled mentions on the chat

packages/ui/src/lib/richText/plugins/FilePlugin.svelte
```diff
@@ -1,87 +1,101 @@
...
+
+		exit();
+	}
+
+	export function exitFileSuggestions() {
+		exit();
 	}
 </script>
+
+<TypeAhead onExit={exit} testMatch={getFileMatch} {onMatch} />
```
estib-vega, line 101: Use the typeahead utility component to reduce the need to customize the matching of patterns

mtsgrd:
Amazing, thanks!

PR #10856, opened 2025-10-28
macOS cross compilation testing

Byron:
The motivation is twofold:

* macOS 15 uses ARM runners which are much faster than the ones for macOS 13.
* macOS 13 is deprecated and these runners may one day go away
* there seems to be no disadvantage to build tauri apps on more recent macOS versions, as they are supposedly working on older macOS' as well.

Inspired by https://github.com/tauri-apps/tauri-action and https://github.com/tauri-apps/tauri-action/pull/749.
More context [on Discord](https://discord.com/channels/1060193121130000425/1432062926625177600/1432228586328424579).

Supersedes #10852.

### Tasks

* [x] rust-cache split
* [x] make Windows build work - it shouldn't have to care at all.
* [x] validate build and suffixes
* [x] make rust-cache work
* [x] final publish to validate rust caches

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -141,10 +141,11 @@ jobs:
       - name: Install Rust targets for macOS
         if: runner.os == 'macOS' && matrix.target != ''
         run: rustup target add ${{ matrix.target }}
-      - name: Rust Cache
+      - name: Rust Cache ${{ matrix.target }}
         uses: Swatinem/rust-cache@v2.8.1
+        if: runner.os == 'macOS' && matrix.target != ''
         with:
-          shared-key: app-release-build-${{ github.event.inputs.channel || 'nightly' }}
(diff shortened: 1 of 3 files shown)


Byron:
The build now works, and the artifact is definitely an Intel build.

Screenshot 2025-10-28 at 18 29 38

The ARM build is also still arm, so it's definitely working.

Byron:
With macOS 15 on a cold cache, it takes 15m (ARM) to 17m (Intel) to build. The other runners, Ubuntu with 12m and Windows with 15m had a hot cache, so it's safe to say that now macOS won't be dragging the publish process down and releases in the sub-20m should be the norm for Nightlies now.

Screenshot 2025-10-28 at 19 25 31

Screenshot 2025-10-28 at 19 25 21

It's worth noting that the rust-cache intentionally won't cache anything produced in the current workspace, so there is always a certain amount of work to be done.

PR #10864, opened 2025-10-29
Prevent windows from building the frontend again despite having downloaded it

Byron:
This gives another 2minutes speedup by avoiding duplicate work.

### Tasks

* [x] works locally - before the script doesn't fail when building it locally
* [x] validate nightly build for Windows (actually try it)


Byron:
Turns out that the RPM code would be harder to debug, and with Windows still being 2m slower than ubuntu which could gain 3m or so if RPM wasn't so slow, it's clear that improving this isn't useful until Ubuntu actually becomes the bottleneck.

Byron:
Screenshot 2025-10-29 at 07 09 05

All seems to work

PR #10867, opened 2025-10-29
feat(profile): move logout section and refine profile-pic hover transitions to improve layout and UX

PavelLaptev:
image


PavelLaptev:
сс @Byron

PR #10884, opened 2025-10-30
Fix safe-checkout so it actually works even in more difficult cases

Byron:
It's probably just a small thing that it is missing, but with grave consequences.

### Tasks

* [x] fix metadata for superseded stacks
* [ ] reproduce the issue in testing
* [ ] fix it


Byron:
Merging this because I didn't realise it contained a fix that really should have been merged a long time ago.

PR #10885, opened 2025-10-30
try to accelerate Windows publishes by not installing perl

Byron:
Maybe it works, the need for strawberry perl was from 2022, maybe the runner image is suitable now.

### Tasks

* [x] try nightly build: https://github.com/gitbutlerapp/gitbutler/actions/runs/18929936428


Byron:
It very much looks like this works, and we now have an incentive to make ubuntu at least 1 minute faster 😁.

Screenshot 2025-10-30 at 06 17 51
Screenshot 2025-10-30 at 06 18 55

PR #10886, opened 2025-10-30
try speeding up publishing by not initialising node

Byron:
Node isn't actually used anymore, not even on Windows where it was accidentally used before.

### Tasks

* [ ] validate publish: https://github.com/gitbutlerapp/gitbutler/actions/runs/18931308535

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -108,21 +108,15 @@ jobs:
       channel: ${{ env.channel }}
 
     steps:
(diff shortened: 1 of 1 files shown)


Byron:
```
Run ./scripts/release.sh \
building:
	channel: nightly
	version: 0.5.1628
	os: linux
	arch: x86_64
	dist: ./release
	sign: true
	target: default
./scripts/release.sh: line 95: pnpm: command not found
```

Right, pnpm is needed. But at this stage, what we really need is tauri, which is faster to install given that we don't need frontend dependencies anymore.

PR #10890, opened 2025-10-30
Move VirtualList into ui package and add to storybook

mtsgrd:


mtsgrd:
@PavelLaptev just fyi

PR #10902, opened 2025-10-30
fix(Gitlab): remove unnecessary characters from draft prefix

ValeraS:
## 🧢 Changes
Remove `:` from added prefix

## ☕️ Reasoning

GitLab uses title prefix to mark drafts: "Draft:", "[Draft]", or "(Draft)". If add `[Draft]: ` to the title and then switch MR to ready to merge, GitLab will only remove the `[Draft]` prefix, leaving the `: ` intact.


krlvi (approved):
Thank you for this

Byron:
Thanks again for the fix!

I still remember how I researched this and it seemed both variants are fine. Probably hallucinations were involved as well 😅.

PR #10912, opened 2025-10-30
Virtual List: Avoid potential type errors

estib-vega:
Handle potential undefined states correctly.

The `rows`  variable can be undefined, so we need to guard against trying to read properties from it correctly.
This fixes the type errors on the branches page

### DIFF
--- a/packages/ui/src/lib/components/VirtualList.svelte
+++ b/packages/ui/src/lib/components/VirtualList.svelte
@@ -164,7 +164,7 @@
 		let i = start;
 
 		while (i < chunks.length) {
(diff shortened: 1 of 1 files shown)


estib-vega:
cc @mtsgrd

krlvi:
I think this change is causing the following issue:
https://github.com/user-attachments/assets/2eb979d7-30ea-430d-82f3-57fcba7b637f

Reverting this commit makes it works correctly again:
https://github.com/user-attachments/assets/0d784705-4b87-4822-a2af-2ad036e45806

PR #10913, opened 2025-10-30
PR listing: Clean the source branch

estib-vega:
Correctly parse the source branch name out from the data.

### DIFF
--- a/apps/desktop/src/lib/forge/interface/types.ts
+++ b/apps/desktop/src/lib/forge/interface/types.ts
@@ -64,6 +64,14 @@ export interface PullRequest {
 	reviewers: ForgeUser[];
 }
 
+function getCleanBranchName(sourceBranch: string): string {
+	if (sourceBranch.includes(':')) {
+		const parts = sourceBranch.split(':');
+		return parts[parts.length - 1] ?? '';
+	}
+	return sourceBranch;
+}
+
 export function mapForgeReviewToPullRequest(pr: ForgeReview): PullRequest {
 	return {
 		htmlUrl: pr.htmlUrl,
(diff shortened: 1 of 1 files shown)


apps/desktop/src/lib/forge/interface/types.ts
```diff
@@ -64,6 +64,14 @@ export interface PullRequest {
 	reviewers: ForgeUser[];
 }
 
+function getCleanBranchName(sourceBranch: string): string {
+	if (sourceBranch.includes(':')) {
```
Byron, line 68: Is this new that it contains a `:`? I always thought that's just a full ref name.

estib-vega: GitHub always returns them in the format <owner>:<branch-name>.

In my recent change to list PRs from the rust-end, I made it so to return the branch as it was. But forgot to clean it up

Byron: Is the format documented in the documentation somewhere?
I am quite certain I would fall for it, still (while also not knowing what GitHub needs `:` for).

It could also be that I confused `sourceBranch` with this, which is essentially the remote tracking ref name if a local tracking branch was created from it to apply it to the workspace:

https://github.com/gitbutlerapp/gitbutler/blob/9c5a1d540dafde37da018ae1504abe5ca01103fa/crates/but-graph/src/virtual_branches_legacy_types.rs#L47

estib-vega: > Is the format documented in the documentation somewhere?
Not on our code but here: https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests

That said, I noticed it's actually better to just use the correct value in the rust-end.
This will have to be eventually be aligned to whatever GitLab gives us, so is better to align it in the backend than having to flip it in the frontend.

This is the more correct fix: https://github.com/gitbutlerapp/gitbutler/pull/10921

PR #10921, opened 2025-10-31
GitHub PR: Use the ref for a PR source branch

estib-vega:
The expected value for the source and target branches should have been the branch name only, e.g. the ref


crates/gitbutler-forge/src/review.rs
```diff
@@ -191,12 +191,22 @@ impl From<but_github::GitHubPrLabel> for ForgeReviewLabel {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
+/// Represents a user from a forge platform (e.g., GitHub, GitLab).
+///
+/// This structure contains information about a user account on a forge platform,
+/// including their identification details and profile information.
```
Byron, line 197: Yes, that's the way :)!

crates/but-github/src/client.rs
```diff
@@ -206,8 +206,8 @@ impl From<octorust::types::PullRequestSimple> for PullRequest {
             author,
             labels,
             draft: pr.draft,
-            source_branch: pr.head.label,
-            target_branch: pr.base.label,
+            source_branch: pr.head.ref_,
+            target_branch: pr.base.ref_,
```
Byron, line 210: Loving this backend fix!

PR #10985, opened 2025-11-04
but review: Add a branch selector

estib-vega:
If no branch is specified, present an interactove selector to pick out which branches you want to create reviews for.

Screenshot 2025-11-04 at 16 18 46

### DIFF
--- a/crates/but-workspace/src/branch_details.rs
+++ b/crates/but-workspace/src/branch_details.rs
@@ -397,3 +397,12 @@ fn local_commits_gix(
     }
     Ok(out)
 }
+
+/// Returns all local commits for the branch identified by `branch_id` that are not in `integration_branch_id`.
+pub fn local_commits_for_branch(
+    branch_id: gix::Id<'_>,
+    integration_branch_id: gix::ObjectId,
+) -> anyhow::Result<Vec<ui::Commit>> {
+    let mut authors = HashSet::new();
+    local_commits_gix(branch_id, integration_branch_id, &mut authors)
+}
--- a/crates/but/src/ui/mod.rs
+++ b/crates/but/src/ui/mod.rs
@@ -1,0 +1,1088 @@
+use std::{io, ops::ControlFlow};
+
+pub mod test_utils;
+use anyhow::Result;
+use bstr::ByteSlice;
+use ratatui::{
+    crossterm::event::{self},
+    style::Stylize,
+    widgets::Widget,
+};
+
+/// Run a terminal UI to select branches from the given stacks.
+pub fn run_branch_selector_ui(stacks: Vec<SimpleStack>) -> Result<Vec<String>> {
+    let mut terminal = ratatui::init();
+    let mut state = BranchSelectorState {
+        stacks,
+        ..Default::default()
+    };
+    state.run(&mut terminal)?;
+    ratatui::restore();
+
+    Ok(state.selected_branches)
+}
+
+#[derive(Debug, Default)]
+pub struct SimpleCommit {
+    /// The commit title.
+    pub title: String,
+    /// The short SHA of the commit.
+    pub sha: String,
+}
+
+impl From<but_workspace::ui::Commit> for SimpleCommit {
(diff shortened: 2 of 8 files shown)


crates/but-workspace/src/branch_details.rs
Byron, line 373: I am so sorry, but… `with_hidden()` actually doesn't really work and it has issues that make it unreliable.
Thus, we still need to use `git2` if we really need a rev-walk.

On the bright side, we shouldn't need one anymore as we must use `but-graph::Graph` for discovering commits relevant to us. Probably the easiest fix is to use `git2`, but with me a little support I think it's no problem to use `but-graph::Graph` either.

crates/but/src/ui/mod.rs
```diff
@@ -0,0 +1,1088 @@
...
+        ");
+
+        // Move focus right to second stack's head (stack-2-branch-1)
+        selector.handle_key_event(event::KeyCode::Right.into());
+        assert_eq!(selector.focused_branch, Some("stack-2-branch-1".into()));
+        buf = Buffer::empty(Rect::new(0, 0, 80, 10));
+        selector.render(buf.area, &mut buf);
+        let buffer_str = buffer_to_string(&buf);
+        insta::assert_snapshot!(buffer_str, @r"
+        ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━Workspace━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
```
Byron, line 927: These tests are pretty awesome!
I would have hoped there is a built-in way to get a string from the buffer, like `Buffer::from_lines()` can create one, but it doesn't look they use the buffer that way.

The `Debug` implementation might be useful though, but I see how it's more than one would want.

```
[crates/but/src/ui/mod.rs:534:9] &buf = Buffer {
    area: Rect { x: 0, y: 0, width: 80, height: 10 },
    content: [
        "┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━Workspace━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓",
        "┃┌branch-1─────────────────────────────┐┌branch-2─────────────────────────────┐┃",
        "┃│0 commits                            ││0 commits                            │┃",
        "┃└─────────────────────────────────────┘└─────────────────────────────────────┘┃",
        "┃                                                                              ┃",
        "┃                                                                              ┃",
        "┃                                                                              ┃",
        "┃                                                                              ┃",
        "┃                                                                              ┃",
        "┗ Done <Enter> Cancel <Q> Select branch <S> Select all <A> Move with arrow keys┛",
    ],
    styles: [
        x: 0, y: 0, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
        x: 35, y: 0, fg: Reset, bg: Reset, underline: Reset, modifier: BOLD,
        x: 44, y: 0, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
        x: 7, y: 9, fg: Blue, bg: Reset, underline: Reset, modifier: BOLD,
        x: 15, y: 9, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
        x: 22, y: 9, fg: Red, bg: Reset, underline: Reset, modifier: BOLD,
        x: 26, y: 9, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
        x: 40, y: 9, fg: Green, bg: Reset, underline: Reset, modifier: NONE,
        x: 44, y: 9, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
        x: 55, y: 9, fg: Green, bg: Reset, underline: Reset, modifier: NONE,
        x: 59, y: 9, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
    ]
}
```

Byron (commented):
Really cool to see `ratatui` show up here :). And the tests are awesome as well!

I just found one correctness issue related to a `gix` bug unfortunately, and it's something that will probably be a bug sooner or later (even though it often kind of works).

PR #10991, opened 2025-11-05
Also provide the names of the stacks that conflicted, so they can be shown.

Byron:
The motivation here is that this is a quick change that will live disproportionally better
for users, and gives us some time to rethink the workflow more thoroughly.

### Tasks

* [x] backend also sends names
* [x] frontend uses them

### DIFF
--- a/apps/desktop/cypress/e2e/support/mock/backend.ts
+++ b/apps/desktop/cypress/e2e/support/mock/backend.ts
@@ -623,6 +623,7 @@ export default class MockBackend {
 
 		// Do nothing for now
 		return {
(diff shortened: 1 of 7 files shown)


Byron:
Thanks for your help with this, @estib-vega, you will be able to take over the PR and push right into it.

Byron:
I thought maybe I could just wrap this up myself, and this is the outcome.

Screenshot 2025-11-10 at 09 45 43

Also, it's meant to be very defensive and deal with the old code-path, but either way it will now more correctly talk about a single stack if just one stack was unapplied.

PR #11019, opened 2025-11-06
fix(ui): Incorrect file path display in file names

PavelLaptev:
Test path `.github/actions/run-frontend-tests/file.svelte` 👇

Before:
image

After:
image


PavelLaptev:
Addresses https://github.com/gitbutlerapp/gitbutler/issues/10604

PR #11047, opened 2025-11-06
but: Actionable information about the account states

### LINKED ISSUE #11028: GH publishing does not seem to work
```
❯ but review publish -b sc-branch-8
Pushing branch 'sc-branch-8'
Push completed successfully
Pushed to remote: origin

Publishing review for branch 'sc-branch-8' targetting 'main
Error: Failed to publish reviews for branches.

Caused by:
    No GitHub access token found for account 'OAuth: schacon'
```

however:

```
❯ but forge  list-users
Known GitHub usernames:
- OAuth: schacon
```

Why does this not work?

Also, if the review publish isn't successful because of an auth problem, we should try to hint to the user how to fix it (ie: "try running `but forge auth`")

estib-vega:
- Improve the error message when failing to get account information 
- Add more information about the account status when listing the known accounts

closes: https://github.com/gitbutlerapp/gitbutler/issues/11028

Screenshot 2025-11-06 at 12 16 21

### DIFF
--- a/crates/but-api/src/commands/github.rs
+++ b/crates/but-api/src/commands/github.rs
@@ -163,3 +163,12 @@ pub async fn list_known_github_accounts() -> Result<Vec<but_github::GithubAccoun
         .await
         .map_err(Into::into)
 }
+
+pub async fn check_github_credentials(
+    account: &but_github::GithubAccountIdentifier,
+) -> Result<but_github::CredentialCheckResult, Error> {
+    let storage = but_forge_storage::controller::Controller::from_path(but_path::app_data_dir()?);
+    but_github::check_credentials(account, &storage)
+        .await
+        .map_err(Into::into)
+}
--- a/crates/but-github/src/lib.rs
+++ b/crates/but-github/src/lib.rs
@@ -242,6 +242,31 @@ pub async fn get_gh_user(
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum CredentialCheckResult {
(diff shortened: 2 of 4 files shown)


Byron:
What does it mean for credentials to be `invalid`?
I understand missing credentials, but feel that `invalid` would mean they don't work for authentication, but then we could also say that. Yet I doubt that the listing will try to use the credentials, at least not without an additional flag.

estib-vega:
> What does it mean for credentials to be `invalid`? I understand missing credentials, but feel that `invalid` would mean they don't work for authentication, but then we could also say that. 

Invalid means that you have a stale/false token, meaning exactly that. 
It uses the credentials

> Yet I doubt that the listing will try to use the credentials, at least not without an additional flag.

I think adding an extra flag is unnecessary. Would be more annoying than useful.
If there is no network connection, this will show `(unknown)` so it won't explode either

Byron:
I see! Then I think `invalid` is too generic, and there should be something that indicates that they were rejected or can't be used for successful authentication.

`invalid` for me is strongly associated with the format of the string or something, i.e. refers to the makeup of the token itself, instead of its it being rejected when used with a remote.

estib-vega:
Any rename suggestions welcome!

Byron:
Hmm, instead of `Invalid Credentials` I'd find "rejected by remote" or "rejected by server" to be closer to what it does.

And from there, the sentence that calls them "invalid" is clear to me, I know they are invalid because the server said so.

PR #11082, opened 2025-11-07
Remove 'log' command and related code

schacon:
The 'log' subcommand and its CommandName variant were removed along with the module and handling in lib.rs because the project no longer supports or needs a separate 'but log' command. This cleans up the CLI enum, command dispatch, module imports, and help groupings to reflect that 'status' is now the sole inspection command.


Byron:
To see more easily what's affected, you can run `cargo check --all-targets`.

PR #11110, opened 2025-11-07
sc-fix-but-help

### LINKED ISSUE #11099: show completion instead
Currently out `but` output looks like this at the end:

```
...

Other Commands:
  init         Initializes a GitButler project from a git repository in the current directory
  mcp          Starts up the MCP server
  forge        Commands for interacting with forges like GitHub, GitLab (coming soon), etc
  review       Command for creating and publishing code reviews to a forge
  completions  Generate shell completion scripts for the specified or inferred shell
  absorb

Options:
  -C, --current-dir <PATH>  Run as if but was started in PATH instead of the current working directory [default: .]
  -j, --json  Whether to use JSON output format
  -h, --help  Print help
```

I would prefer to either hide the `completion` subcommand (because running it is useless by itself, it's super confusing), or just show how to use it:

```
...

Other Commands:
  init         Initializes a GitButler project from a git repository in the current directory
  mcp          Starts up the MCP server

To add command completion, add this to your shell rc (for example ~/.zshrc):
   eval "$(but completions zsh)"

Options:
  -C, --current-dir <PATH>  Run as if but was started in PATH instead of the current working directory [default: .]
  -j, --json  Whether to use JSON output format
  -h, --help  Print help
```

schacon:
CleanShot 2025-11-07 at 17 38 03@2x

fixes #11099

### DIFF
--- a/crates/but/src/args.rs
+++ b/crates/but/src/args.rs
@@ -187,6 +187,7 @@ For examples see `but rub --help`."
         #[clap(value_enum)]
         shell: Option<clap_complete::Shell>,
     },
+    /// Automatically absorb uncomitted changes to an existing commit
     Absorb {
         /// If the CliID is an uncommitted change - the change will be absorbed
         /// If the CliID is a stack - anything assigned to the stack will be absorbed accordingly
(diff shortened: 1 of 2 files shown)


schacon:
close in favor of #11147

PR #11112, opened 2025-11-07
Prevent axum abort on client disconnect

mtsgrd:
This ensures the lock is removed after the claude process terminates.


mtsgrd:
@Caleb-T-Owens @krlvi  with this applied I don't get those claude is busy messages. How do we proceed?

krlvi:
LGTM - lets land this. 
PS: It's so cool that i can apply this thing now with the `apply3` feature flag without rebasing it and push a commit

PR #11113, opened 2025-11-07
Reconcile stacks for the `gitbutler/workspace`, not for the HEAD workspace.

Byron:


Byron:
CC @mtsgrd for the fix.
CC @krlvi in case this should be a patch. The worst case that can happen is that edit mode doesn't work, and possibly that people can't get out of edit mode if they closed the app while in it (which generally isn't a good idea as they are in neverland then). So, I seem to be talking myself into thinking this PR should be a patch, which fortunately is able to fix issues the previous version introduced as well.

PR #11121, opened 2025-11-08
Fixi but-server claude websocket events

mtsgrd:
Move reading of the child process stdout into a dedicated blocking
thread and forward lines through an unbounded tokio mpsc channel to the
async task. Replace the previous direct async BufReader lines() loop
with a blocking std::thread::spawn that calls BufReader::lines() and
sends each Ok line over tx; the async side now awaits rx.recv().

This prevents blocking the tokio runtime on stdio reads, avoids
calling blocking I/O in async context, and preserves line-by-line
delivery for downstream JSON parsing and handling.


krlvi:
@mtsgrd note that this affects not just `but-server` but also the tauri API - so worth verifying all is still good there too

PR #11125, opened 2025-11-08
Add db watcher to but-server

mtsgrd:
Fixes among other things permission requests.

### DIFF
--- a/crates/but-server/src/projects.rs
+++ b/crates/but-server/src/projects.rs
@@ -3,7 +3,9 @@ use std::collections::HashMap;
 use anyhow::{Context as _, Result};
 use but_api::{App, error::ToError};
 use but_broadcaster::FrontendEvent;
+use but_db::poll::DBWatcherHandle;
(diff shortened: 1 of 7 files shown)


crates/but-server/src/projects.rs
```diff
@@ -17,8 +19,15 @@ struct SetProjectActiveParams {
     id: ProjectId,
 }
 
+struct ProjectHandles {
+    #[allow(dead_code)] // Watchers are kept alive; Drop handles cleanup
+    file_watcher: WatcherHandle,
+    #[allow(dead_code)] // Watchers are kept alive; Drop handles cleanup
+    db_watcher: DBWatcherHandle,
```
Caleb-T-Owens: I _think_ the idiomatic rust thing is to name these with `_` in cases like this

Caleb-T-Owens:
LGTM :+1:

crates/but-server/src/projects.rs
mtsgrd: Thanks

PR #11126, opened 2025-11-08
Emit new commit system message in codegen chat after stop hook

mtsgrd:
This is an experiment with adding a commit confirmation to codegen. It could be as simple as a message saying something has been committed, or as this poc shows, using the existing `CommitDetails` component inline in the chat.

image


PavelLaptev:
The UI doesn't fit. It's not our styles. But I can update it after, so fill free to merge

PavelLaptev (approved):

mtsgrd:
It would be great if we can click through to the commit or something.

PR #11133, opened 2025-11-09
Show draft codegen row when starting new session

mtsgrd:
image


PavelLaptev:
What’s the purpose of displaying the codegen row if there are no messages?

PavelLaptev:
If there are no messages (0 context used), it makes more sense not to display this at all might be. What if someone clicks here just out of curiosity? How can I remove this codegen attachment now?

If we’re saying that clicking the AI button creates a new session, and even if it’s empty, we should still show the code generation row. A user should have a mechanism to remove it.

mtsgrd:
Good points, let's not merge this!

PR #11148, opened 2025-11-10
Single-branch support for `but status`

Byron:
This is to help AI to use newer APIs, maybe, and to show that it can already be done.

### Tasks

* [x] refactors
* [x] API proposal

### Next PR 

* [ ] put `gitbutler-` dependencies behind `legacy` feature toggle.
* [ ] status uses `but_workspace::ref_info()`

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -112,6 +112,10 @@ jobs:
           save-if: ${{ github.ref == 'refs/heads/master' }}
       - run: cargo fmt --check --all
       - run: cargo check --workspace --all-targets
+      - run: |
+          # should work without any features
+          cargo check -p but-project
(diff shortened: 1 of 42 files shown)


Byron:
@krlvi I started out wanting to make `but status` single-branch compatible by using the new APIs, and noticed that it's strangely impossible to do so. So I thought I'd start changing this and came up with `but-project`, that has optional legacy support (behind a feature toggle). This makes it necessary to 'stay clean', and seems to show the way forward to minimal abstractions while still supporting older or not-fully-ported code.

Tomorrow I think it's a good idea to clean up more and hopefully get the single-branch mode done in `but-status`, which really is no more than traversing `but_workspace::RefInfo` instead.

PR #11153, opened 2025-11-10
but-actions is no more

estib-vega:
Remove the remnants of the but action. Puns intended

---
- <kbd>&nbsp;2&nbsp;</kbd> #11159 
- <kbd>&nbsp;1&nbsp;</kbd> #11153 👈


Byron:
Does that mean that the `but-actions` crate is going to be deleted? And if so, what would be used in its place? (This is me trying to catch up, I missed its entire story actually).

estib-vega:
> Does that mean that the `but-actions` crate is going to be deleted? And if so, what would be used in its place? (This is me trying to catch up, I missed its entire story actually).

Not yet, at least. We might use this in other iterations of features in the future.

This is more like removing the unused components from the FE and also removing the reliance of watching the DB updates in order to update the workspace info

Byron:
Thanks, I see, that sounds like it's going to remain a pillar, akin to `gitbutler-branch-actions`.

PR #11158, opened 2025-11-10
Add UpDownPlugin and wire up user input recall

mtsgrd:
Introduce a new UpDownPlugin for rich text editors to allow navigating
history entries and inserting prior user messages. Export the plugin
from the UI library and include it in the RichTextEditor story so it
appears in Storybook.

Wire backend support and command routing for retrieving user messages:
add a "claude_get_user_message" handler in the server command map and
call the new backend method from the desktop CodegenInput component.
When a historical user message contains attachments, add them to the
attachment service so they are available in the editor.

Fix a typo in the Claude DB accessor: request user messages by the
MessagePayloadDbType::User tag (previously used UserInput) so lookups
correctly return stored user messages.


mtsgrd:
https://github.com/user-attachments/assets/3dd87753-6a00-4a60-a7cc-12008c1122b4

PR #11167, opened 2025-11-11
Deep link support

estib-vega:
Add support for deeplinking the application.
This will enable opening the application from the web or from the terminal

Related to: https://github.com/gitbutlerapp/gitbutler/issues/11025

### DIFF
--- a/crates/but/src/gui.rs
+++ b/crates/but/src/gui.rs
@@ -1,43 +1,104 @@
-use std::process::Stdio;
+use std::env;
 
 use anyhow::{Context, Result, bail};
 
 /// Open the GitButler GUI application for `possibly_project_dir`.
 ///
-/// This function attempts to launch the GitButler desktop application, and only works
-/// if this binary is either co-located with `gitbutler-tauri` or *is* `gitbutler-tauri`
-/// itself.
-///
-/// Then we open the GUI application and provide the path to figure out how to open it.
+/// This expects that the GUI application is present and has correctly registered URL
+/// schemes for the different channels.
 pub fn open(possibly_project_dir: &std::path::Path) -> Result<()> {
-    let exe_path = std::env::current_exe()?.canonicalize()?;
-    let stem = exe_path
-        .file_stem()
-        .context("Need file-stem of current executable")?;
-    let gui_exe_path = if stem != "gitbutler-tauri" {
-        exe_path.with_file_name(if cfg!(windows) {
-            "gitbutler-tauri.exe"
-        } else {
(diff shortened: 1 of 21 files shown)


crates/but/src/gui.rs
```diff
@@ -1,43 +1,74 @@
...
+            AppChannel::Nightly => "gb-nightly",
+            AppChannel::Release => "gb",
+            AppChannel::Dev => "gb-dev",
+        };
+
+        let url = format!("{}://open?path={}", scheme, possibly_project_dir.display());
+
+        #[cfg(target_os = "macos")]
+        {
+            std::process::Command::new("open")
```
Byron: If system URL associations are used, then I think the `open` crate is a more portable choice here.
Disclaimer: It's a crate I maintain, and it's already available via `open.workspace = true`.

crates/gitbutler-tauri/tauri.conf.release.json
```diff
@@ -21,6 +21,11 @@
 				"https://app.gitbutler.com/releases/release/{{target}}-{{arch}}/{{current_version}}"
 			],
 			"pubkey": "dW50cnVzdGVkIGNvbW1lbnQ6IG1pbmlzaWduIHB1YmxpYyBrZXk6IDYwNTc2RDhBM0U0MjM4RUIKUldUck9FSStpbTFYWUE5UkJ3eXhuekZOL2V2RnpKaFUxbGJRNzBMVmF5V0gzV1JvN3hRblJMRDIK"
-		}
+		},
+		"deep-link": {
+      "desktop": {
+        "schemes": ["gb"]
```
Byron: Could this not be `but`? Probably no other application wants to use it either ;).
I had to ask :D.

crates/but/src/gui.rs
estib-vega: Uhhh that's cool

crates/gitbutler-tauri/tauri.conf.release.json
estib-vega: Haha, you're right :)

estib-vega:
There are some rough edges still:

In order to know which application to open, we probe the CHANNEL and IDENTIFIER which don't seem to be set when calling but.
That needs to be fixed in order to be able to open the application from the but CLI

Byron:
Thanks so much!

I have updated [the task list](https://github.com/gitbutlerapp/gitbutler/issues/11025#issuecomment-3497503042) accordingly and would hope we can sort out the URL issue today 🎉.

PR #11176, opened 2025-11-11
Fix codegen file mention bug that deletes rest of line

mtsgrd:
image

- Select inserted text
- Do not truncate the rest of the line
- Handle multiple occurrences of file trigger
- Improve cursor placement after inserting file path
- Default to no-op when the expected file match is absent


PavelLaptev (approved):
nice. thanks!

mtsgrd:
@PavelLaptev this solution doesn't yet turn mentioned files into the right style yet, I'll try and fix that.

PR #11177, opened 2025-11-11
Add inline code node and plugin for rich text

mtsgrd:
Using a new `InlineCodeNode` extension to the lexical editor.


mtsgrd:
@PavelLaptev fyi - if you could test this out?

PR #11197, opened 2025-11-12
Handle network errors when fetching GitHub user

mtsgrd:
Detect and surface network connectivity problems when calling the GitHub
API from get_gh_user.

Expose the GitHub error object through the forge factory and related 
hooks so the frontend can react to network or API errors.

Prevents showing the integration setup prompt on network errors.

### DIFF
--- a/crates/but-github/src/lib.rs
+++ b/crates/but-github/src/lib.rs
@@ -226,10 +226,23 @@ pub async fn get_gh_user(
         let gh = account
             .client(&access_token)
             .context("Failed to create GitHub client")?;
-        let user = gh
-            .get_authenticated()
-            .await
-            .context("Failed to get authenticated user")?;
+        let user = match gh.get_authenticated().await {
+            Ok(user) => user,
+            Err(client_err) => {
+                // Check if this is a network error before converting to anyhow
+                if is_network_error(&client_err) {
+                    return Err(anyhow::Error::from(client_err).context(
+                        but_error::Context::new_static(
+                            but_error::Code::NetworkError,
+                            "Unable to connect to GitHub.",
+                        ),
+                    ));
+                }
+                return Err(
+                    anyhow::Error::from(client_err).context("Failed to get authenticated user")
+                );
+            }
+        };
         Ok(Some(AuthenticatedUser {
             access_token,
             login: user.login,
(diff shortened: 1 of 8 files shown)


crates/but-github/src/lib.rs
```diff
@@ -290,3 +318,49 @@ pub async fn list_known_github_accounts(
...
+    }
+
+    #[test]
+    fn test_is_not_network_error_auth_failure() {
+        let err = anyhow::anyhow!("401 Unauthorized");
+        assert!(
+            !is_network_error(&err),
+            "Should not detect auth failures as network errors"
+        );
+    }
```
mtsgrd, line 374: 👀

crates/but-github/src/lib.rs
```diff
@@ -242,6 +251,25 @@ pub async fn get_gh_user(
     }
 }
 
+/// Check if an error is a network connectivity error.
+///
+/// This includes DNS resolution failures, connection timeouts, connection refused, etc.
+fn is_network_error(err: &anyhow::Error) -> bool {
```
Byron: It would be worth trying to downcast this error to its original one, [shown here](https://github.com/gitbutlerapp/gitbutler/blob/d87b098c3fe1af27fc38ba57e3dfcb23c3a39f51/crates/but-github/src/client.rs#L34), and then see if it contains the information in a more type-safe format.

mtsgrd: Thanks. Does this look better now?

crates/but-github/src/lib.rs
```diff
@@ -242,6 +251,25 @@ pub async fn get_gh_user(
...
+    {
+        return reqwest_err.is_timeout() || reqwest_err.is_connect() || reqwest_err.is_request();
+    }
+
+    // Fallback: check for reqwest::Error directly in the error chain
+    err.chain().any(|cause| {
+        cause
+            .downcast_ref::<reqwest::Error>()
+            .is_some_and(|e| e.is_timeout() || e.is_connect() || e.is_request())
+    })
```
Byron: Downcast actually takes care of walking the error chain, so this can be removed entirely.

crates/but-github/src/lib.rs
```diff
@@ -242,6 +251,25 @@ pub async fn get_gh_user(
...
+/// Check if an error is a network connectivity error.
+///
+/// This includes DNS resolution failures, connection timeouts, connection refused, etc.
+fn is_network_error(err: &anyhow::Error) -> bool {
+    // Check if it's a ReqwestError wrapped in ClientError
+    if let Some(octorust::ClientError::ReqwestError(reqwest_err)) =
+        err.downcast_ref::<octorust::ClientError>()
+    {
+        return reqwest_err.is_timeout() || reqwest_err.is_connect() || reqwest_err.is_request();
+    }
```
Byron: Yes, this looks very much like it!

Something I'd always worry about, is if the downcast is actually to the right type. So I'd try to let `gh.get_authenticated()` return its original error type so a downcast isn't necessary in the first place.
If that's not easily possible though, I think keeping this is fine.

crates/but-github/src/lib.rs
Byron: Definitely, thanks for trying it out!

crates/but-github/src/lib.rs
mtsgrd: Got it, makes more sense!

PR #11222, opened 2025-11-13
Upstream integration: Remove the `force_integrated_branches` option

estib-vega:
We don’t use it anymore, and we probably will rethink the way we handle this states in a nicer way


krlvi:
Nice cleanup. The cool thing is that now that we use but-graph, we are much better at detecting when branches are integrated, including the squash merged situations, so hopefully we never need this style of approach again

PR #11226, opened 2025-11-13
Add draft mechanism for codegen

mtsgrd:
Extend the StackDraft component to also handle draft codegen sessions, 
using the message queue in order to facilitate sending the first 
message.


mtsgrd:
@estib-vega no matter what I do, I can't get this pr past the Client Frontend Integration Tests. Commit actions seem to lead to "aww snap", both in CI and local. Could you lend a hand?

estib-vega:
> @estib-vega no matter what I do, I can't get this pr past the Client Frontend Integration Tests. Commit actions seem to lead to "aww snap", both in CI and local. Could you lend a hand?

Yeap looking into that now.
I've seen them fail a couple of times due to memory issues. I think I'll just port them to playwright

PavelLaptev (approved):

PR #11234, opened 2025-11-14
Deeplink: Clear the URLS after processing

estib-vega:
Clear the deeplink URLs after emission, that way the application doesn’t navigate back after reload

### DIFF
--- a/apps/desktop/src/lib/backend/tauri.ts
+++ b/apps/desktop/src/lib/backend/tauri.ts
@@ -93,28 +93,52 @@ export default class Tauri implements IBackend {
 	}
 }
 
+const LAST_PROCESSED_DEEP_LINK_URL_KEY = 'lastProcessedDeepLinkUrl';
+
+function alreadyProcessedDeepLinkUrl(url: string): boolean {
+	const lastProcessedUrl = localStorage.getItem(LAST_PROCESSED_DEEP_LINK_URL_KEY);
+	return lastProcessedUrl === url;
+}
+
+function markDeepLinkUrlAsProcessed(url: string): void {
+	localStorage.setItem(LAST_PROCESSED_DEEP_LINK_URL_KEY, url);
+}
+
 function handleDeepLinkUrls(urls: string[], handlers: DeepLinkHandlers) {
-	// For now, only handle a single URL at a time
-	if (urls.length !== 1) return;
-	for (const url of urls) {
-		if (!isValidDeepLinkUrl(url)) {
-			console.warn('Received invalid deep link URL:', url);
-			continue;
-		}
+	// We don't care about the previous URLs, only the last one.
+	const url = urls[urls.length - 1];
+	if (!url) return;
+	if (!isValidDeepLinkUrl(url)) {
+		console.warn('Received invalid deep link URL:', url);
+		return;
+	}
 
-		const result = parseDeepLinkUrl(url);
-		if (!result) {
-			console.warn('Failed to parse deep link URL:', url);
-			continue;
-		}
+	const result = parseDeepLinkUrl(url);
+	if (!result) {
+		console.warn('Failed to parse deep link URL:', url);
+		return;
+	}
 
-		const [topLevel, params] = result;
-		switch (topLevel) {
-			case 'open': {
-				const path = params.get('path');
-				if (path) {
-					handlers.open(path);
-				}
+	const [topLevel, params] = result;
+	const timestamp = params.get('t');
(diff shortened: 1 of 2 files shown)


estib-vega:
hmmm, this doesn't seem to work.
We need to know whether the application already processed the deeplink, so that on app reload we don't redirect them again.
This doesn't seem to be cleared (or clearable) through the Tauri APIs...?

Byron:
That's strange and has the makeup of a bug.
I never read the tauri docs on deep linking though, so maybe there is something special to be done.
There is a chance, of course, that it's half-baked as well.

Would a front-end based fix be tenable, so it remembers that it followed the link already? Or is it too baked in and the URL jump is triggered by Tauri?

estib-vega:
> That's strange and has the makeup of a bug. I never read the tauri docs on deep linking though, so maybe there is something special to be done. There is a chance, of course, that it's half-baked as well.
> 
https://v2.tauri.app/plugin/deep-linking/

Nothing I can immediately find. It's weird and might indeed be a bug.

> Would a front-end based fix be tenable, so it remembers that it followed the link already? Or is it too baked in and the URL jump is triggered by Tauri?

https://github.com/tauri-apps/plugins-workspace/issues/1879#issuecomment-2458108929

I see some mention on workarounds from the FE that I could still try out. Man, it sucks that this behavior seems to be known but not documented

Byron:
Thanks for the links! I couldn't make much of the issue, but at least could take a look at the docs for the first time.

They don't seem to mention how to remove the deeplink URLs or clear its state, which I understand is the problem here.
And… I don't see know how `getCurrent` can be tuned to not return the same URLs each time.
My hope would be that it pushes onto the list, so each deep link is added to the end of it. Then it would be possible to keep track of the last seen list length in the Window state, non-persisting.
But I am also sure that's not how it works, isn't it?

What is most annoying about this certainly is the debugging workflow, as it needs a real nightly build each time. Doing this locally *should* be OK by now as it builds relatively quickly. But 2m is still not fast, but faster than waiting on CI.

Byron:
And I did try to build my own local copy with this patch to learn about URLs…

```patch
diff --git a/apps/desktop/src/lib/backend/tauri.ts b/apps/desktop/src/lib/backend/tauri.ts
index 151c4d23a0..1790a4536c 100644
--- a/apps/desktop/src/lib/backend/tauri.ts
+++ b/apps/desktop/src/lib/backend/tauri.ts
@@ -41,12 +41,14 @@ export default class Tauri implements IBackend {
 	async initDeepLinking(handlers: DeepLinkHandlers): Promise<void> {
 		// Check if app was launched via deep link
 		const urls = (await getCurrent()) ?? [];
+		console.log(urls);
 		if (urls.length > 0) {
 			handleDeepLinkUrls(urls, handlers);
 		}
 
 		// Listen for new deep links while app is running
 		await onOpenUrl((urls) => {
+			console.log(urls);
 			handleDeepLinkUrls(urls, handlers);
 		});
 	}
diff --git a/crates/gitbutler-tauri/src/main.rs b/crates/gitbutler-tauri/src/main.rs
index b2da0947b2..469131e865 100644
--- a/crates/gitbutler-tauri/src/main.rs
+++ b/crates/gitbutler-tauri/src/main.rs
@@ -187,7 +187,8 @@ fn main() -> anyhow::Result<()> {
                 });
 
                 let app_handle_for_deep_link = app_handle.clone();
-                app_handle.deep_link().on_open_url(move |_| {
+                app_handle.deep_link().on_open_url(move |e| {
+                    dbg!(e.urls());
                     // Get main window
                     if let Some(window) = app_handle_for_deep_link.get_window("main") {
                         let _ = window.unminimize();
```

…but it didn't work at all. It seems it needs a real build with signature.

And indeed, the only problem seems to be that app reloads jump to the most recent deeplink.
But I really wonder if this is only one URL or more, because if it's more, I think this can be used to avoid responding to a known URL (by index into the URLs array).
But seems too good to be true… .

estib-vega:
> But I really wonder if this is only one URL or more, because if it's more, I think this can be used to avoid responding to a known URL (by index into the URLs array).
> But seems too good to be true… .
>

It seems that the array always includes one URL, no matter what 🤦‍♂️ 

What I though might actually circumvent this issue is to add a timestamp to the deeplink and persist the last URL we processed. That way we'll only process a new one.

Byron:
That's a great idea! And one that definitely works 🎉!

krlvi:
this timestamp approach is a really cool solution! 🎉

PR #11238, opened 2025-11-14
Make Claude sessions optionally project-wide

mtsgrd:


krlvi:
without a mechanism for distributing changes produced by claude to branches, i am not quite sure this fits the bill. I'd suggest leaving this for a future project looking at this more holistically (workspace sessions)

krlvi:
Having slept on this I have an idea for how to make workspace-scoped chats work. 
We could make the prompt have a "target" - similar to how a model can be chosen, we could have some option for selecting between stack, branch or a commit. 

If a stack or a branch is selected, a new commit is created there, if a commit is selected, then the new changes will be amended into that commit. 

This "target" can be changed between prompts and the context can be cleared on demand. 

---

The would require changing things up in how committing is done but it could be a nice UX.

In any case, I would like to take this as a new project after we have shipped the current thing

PR #11246, opened 2025-11-14
simplify branch creation

PavelLaptev:
I tried to simplify branch creation, but the context menu approach didn't work well after testing it. Branch creation is now 3 clicks away instead of the previous 2 clicks. Also, this context mixing creates confusion between the meanings of those two actions.

Before:

https://github.com/user-attachments/assets/3c8a81e8-8489-4f2b-a7e7-cfaf34cd4342

After:

https://github.com/user-attachments/assets/097dd6df-0089-46ff-b32f-7825edd55881


PavelLaptev:
cc @krlvi

PR #11249, opened 2025-11-15
Show tool input while tool call processing

mtsgrd:
image

### DIFF
--- a/apps/desktop/src/components/codegen/CodegenToolCall.svelte
+++ b/apps/desktop/src/components/codegen/CodegenToolCall.svelte
@@ -44,12 +44,12 @@
 		{/snippet}
 
 		{#snippet content()}
-			{#if toolCall.result}
-				<div class="stack-v gap-6 m-b-8">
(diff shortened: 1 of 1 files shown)


mtsgrd:
@PavelLaptev can you merge or amend this if you agree with it?

PavelLaptev:
Thanks @mtsgrd!
In my opinion, let's not do this now and either close it or ask the rest of the team on Monday.
On one hand, I understand the motivation for disabling it, the tool call hasn't finished its work yet. On the other hand, I'd like to be able to expand the tool call if it's taking too much time. Another thing to consider: our tool calls aren't expanded by default, and the choice to expand or not is completely up to the user. So if they want to expand it and it won't cause any errors or issues, let them do it.

mtsgrd:
Sorry, I could have explained myself better. Without this change you can expand the tool call, but the container grows only with a few pixels while loading.. but therein lies the real problem, let's render the input while waiting for output?

PR #11271, opened 2025-11-17
Jt/repo absence test

jonathantanmy2:
I'm not sure if this is an issue with my configuration, but I ran into this error.

If you don't have the same problem on Mac, the error message might be OS-specific, in which case I think we can use a `[..]` wildcard on the error message to be compatible across OSes. If this is the case, let me know and I can update this PR.

### DIFF
--- a/crates/but/tests/but/journey.rs
+++ b/crates/but/tests/but/journey.rs
@@ -37,7 +37,7 @@ fn from_scratch_needs_work() -> anyhow::Result<()> {
     let env = Sandbox::empty()?;
 
     env.but("status").assert().failure().stderr_eq(str![[r#"
-Error: Could not find a git repository in '.' or in any of its parents
+Error: Could not find a git repository in '.' or in any of its parents[..]
 
 "#]]);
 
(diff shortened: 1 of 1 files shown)


jonathantanmy2:
The error message was indeed configuration-specific, so I've updated the PR to support both cases. PTAL.

Caleb-T-Owens:
Hi @jonathantanmy2,

Thanks for the contribution.

I am a little curious as to what these configuration specific differences are in the output?

Could you share some examples so I can get a better idea as to wether this output _should_ be config-sensitive.

jonathantanmy2:
The difference is in the commit message - copying and pasting for convenience:
```
---- expected: tests/but/journey.rs:194:20
++++ actual:   stderr
   1      - Error: Could not find a git repository in '.' or in any of its parents
        1 + Error: Could not find a git repository in '.' or in any of its parents within device limits below '/'
```

Byron:
Thanks a lot! Let's make tests work for you!
It would definitely be interesting to get to the bottom of this, and I will look out for it when I (finally) get to update `but status` to use the new workspace APIs.

PR #11279, opened 2025-11-18
Hunk Selection: Commit always uses 0-anchored hunk headers

estib-vega:
The frontend alway uses the 0-based anchoring scheme when passing changes to the backedn for committing.
- If the file is completely selected, empty hunk header lists are passed
- Even when selecting complete hunks, we turn them into multiple hunk headers that are anchored to the 0th line


estib-vega:
cc @Byron 
This should address the topic regarding the committing issues (hopefully) :)

Byron:
Thanks a lot 🙏, it's much appreciated.
Let's hope it works until people want a different selection experience, which is probably when all of that needs to be touched again. And I hope that doesn't happen anytime soon 😅.

PR #11286, opened 2025-11-18
Fix key up and key down in file lists

Caleb-T-Owens:
Fixes regression in https://github.com/gitbutlerapp/gitbutler/pull/11270


mtsgrd:
Hmm, is the title of this pr correct?

Issue #11263, opened 2025-11-17
Overhaul CLI Ids


krlvi:
## Current state
The `but` CLI defines and makes use of a type `CliId` which is used when:
- Displaying relevant objects in the terminal
- Referring to relevant objects in command arguments

The canonical examples here are `but status` and `but rub <source> <destination>`

Some of the types of objects that currently receive a CLI id are:
- Commits
- Branches
- Uncommitted files
- Files that are part of a commit

### Ergonomics
The main purpose of the CLI ids is to offer an ergonomic way of referring to relevant items. For that reason, currently the CLI ids are made up of just 2 lowercase alpha characters (ease of typing).
In the case of commit shas, the first 2 chars are taken directly, whereas for files and branch names, a 2 character id is generated on the fly.

### Interop
It is also intended that the CLI IDs can be replaced with the full identifier - e.g you can use either `7d` or `7d72456`, and also either `ve` (generated ID) or `Readme.md` and either `tw` or `my-branch`.

### Collision handling
Currently it is possible for the system to generate more than one "objects" with the same ID. When a CLI Id is being passed as an argument, if there is an ambiguity, the application will prompt the user to select which object they were referring to. The implementation is unfortunately not applied consistently across all sub-commands.

## Issues with the current implementation
- Identifiers change / dont feel predictable: If an uncommitted file is amended into a commit (or vice versa), it gets a new ID. This also happens if the file is assigned to a different stack.
- There are bugs in handling the "full" identifiers (eg. branch names) where a generated id like `br` collides with a branch named `my-branch` (due to the br in branch)
- The implementation for disambiguating CLI ids is not used uniformly for all relevant commands.
- Inefficiency - the current implementation loads the full workspace state to "discover" all CLI IDs in order to match a CLI id to an internal object - perhaps this can be implemented in a better way. (Note by @Byron: This is fast if expensive bits of information are skipped, which should be the default here)

## New CLI ID goals
It would be good to take a step back and re-consider how the CLI ids function and how they are implemented.
The key goals would be to:
- Address known existing bugs / limitations (addressing the sub-issues)
- Improve code quality & maintainability
- Improve ergonomics for users (eg. can we do better than we do now?)
- Make this scale for doing operations on Hunks as well
- @Byron's notes: can we have user journeys that are significantly simplified through CliIDs? How 'stable' would these have to be then?
- We can consider using the commit Change IDs when deriving CLI ids for commits, thus making them stable

[...]

PR #11291, opened 2025-11-19
id: canonical IDs for most branch names

jonathantanmy2:
## 🧢 Changes

(copied from commit message)

Currently, all CLI IDs of branches are hashes (deterministically pseudorandom). Generating them based on the branch names would result in a more comfortable user experience, but this was not possible without the risk of collisions, since each branch ID was computed independently of every other branch (e.g. taking the first 2 characters as ID would cause "aaa" and "aab" to collide). However, an ID database was introduced recently, which makes it possible to compute branch IDs in view of all branch names.

Therefore, teach it to supply branch names.

The ID of each branch is the first pair of characters in the branch name that are alphanumeric and cannot be confused with another branch name or an OID. If no such ID is possible, fall back to the existing way of computing the ID (i.e. hashing). Ideally, we would check this ID for collisions against all other IDs and alter it slightly if there is a collision, but this cannot be done until we ensure that all CLI commands interpret and/or generate IDs in exactly one place, and does not add/rename/remove any branches within that place. This is left for future work.

(end copy)

Related to #11263

From a code health perspective, this also enables future changes (e.g. when generating file `CliId`s, we can ensure that those don't conflict with any branch `CliId`s).

### DIFF
--- a/crates/but/src/absorb.rs
+++ b/crates/but/src/absorb.rs
@@ -53,7 +53,7 @@ pub(crate) fn handle(
                 // Absorb this particular file
                 absorb_file(project, &path, assignment, &assignments, &dependencies, out)?;
             }
-            CliId::Branch { name } => {
+            CliId::Branch { name, .. } => {
                 // Absorb everything that is assigned to this lane
                 absorb_branch(project, &name, &assignments, &dependencies, out)?;
             }
--- a/crates/but/src/branch/mod.rs
+++ b/crates/but/src/branch/mod.rs
@@ -189,7 +189,7 @@ pub async fn handle(
                             position: but_workspace::branch::create_reference::Position::Above,
                         })
                     }
-                    crate::id::CliId::Branch { name } => Some(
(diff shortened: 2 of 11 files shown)


Byron (approved):
Thanks a lot, let's merge this!

Please note that I fixed merge conflicts (and *think* did so correctly, but better double-check). Additionally I put a refactor commit on top. The to me most important thing it adds is a TODO that hopes for the arrival of unit-level tests for `IdMap`. Cargo would just pick these up and I think it will allow to show the behaviour we want with more specific and more numerous tests than what would be desirable with these `but` program runs we otherwise do.

I wouldn't mind if `tests/but/id` goes away completely, after all there is no `id` subcommand.

jonathantanmy2:
Thanks, your resolution of the merge conflicts looks good. Removing `tests/but/id` in favor of unit tests for `IdDb` is a good idea too, but probably needs to wait until we change it to not use legacy code (so we don't have to figure out how to set up the unit test for legacy code, and then replace that with non-legacy code). I'll work on that now.

PR #11294, opened 2025-11-19
sc-refactor-cli-documentation

schacon:
This PR refactors the subcommands definitions to a single new file called `subcommands.rs` and adds a new `but-clap` bin crate that will take the Clap documentation (mostly from that file) and automatically generate the mdx files from it.

It also removes the `but log` command, rather than just hiding it.


crates/but/src/lib.rs
```diff
@@ -501,3 +497,9 @@ mod trace {
         Ok(())
     }
 }
+
+/// Get the clap Command for documentation generation
+pub fn get_command() -> clap::Command {
```
Byron: I recommend to omit this function entirely and use `but::Args` directly. The `clap::Command` conversion can be done by the caller.

Byron (commented):
I appreciate the refactoring and thought it's a good opportunity to bring everything related to `clap` into the `args` module. This might mean `subcommands.rs` also moves there.

crates/but/src/lib.rs
schacon: sounds good, will do.

schacon:
> I appreciate the refactoring and thought it's a good opportunity to bring everything related to `clap` into the `args` module. This might mean `subcommands.rs` also moves there.

Yeah, it's a good point, I'll merge this back into args.rs.

schacon:
ok @Byron I moved it all back to args.rs  and removed that command thing.

PR #11297, opened 2025-11-19
`gitbutler-command-ctx` to `but-ctx`

Byron:
That way, `but-ctx::Context` can replace all usages of `CommandContext`, paving the way for further de-legacy-isation down the road.

A major side-effect of this will be that everything related to context and project ids is then within one crate, with legacy code clearly marked.

### Tasks

* [x] make `repo()` lazy and on the fly.
* [x] make `git2_repo()` lazy and on the fly (kind-of - needs adjustments later)
* [x] make `git2_repo()` non mutable for safety (won't panic)
* [x] switch `repo()` over to refcell version
* [x] ~~try to 'take' existing DBs into a thread-safe context~~
* [x] fold `gitbutler-command-context`
* [x] make everything `pub` in the Context and just state the invariant for ease of use.
* [x] look for `new_from_legacy_project_and_settings` to have a version that doesn't use default settings
* [x] replace the `project_data_dir()`

Follow-up of #11292
Related to #11236


krlvi (approved):
discussed over call

PR #11305, opened 2025-11-20
some-tabbing-fixes

PavelLaptev:
- Previously, some modals didn’t have a visible focus-outline.
In this PR, I added more focus styles for UI components like buttons, toggles, checkboxes, and radio buttons to ensure that the outline is visible for all color modes and whether they are checked or not, as we used the same color for checked states and the focus outline.
- I removed the `.focus-state` style to use only a single source of outline style for focus states, which is now `.focus-visible`.
- The `.focus-visible` class is now a part of the design-core package.
- I made some fixes for the `tabbable` and `focusManager` components. In some modals, some buttons previously didn’t have a proper focus outline (see the video below).

Here are some buttons without a focus outline and inconsistent focus styles:

https://github.com/user-attachments/assets/ae031154-182b-4ae6-a474-98190f284c7f

After the changes:

https://github.com/user-attachments/assets/d52f493f-e4ee-4e3e-be97-5acd5d5d8a4b

https://github.com/user-attachments/assets/5921d09d-075a-4d17-9198-1bdd7997a9ed

https://github.com/user-attachments/assets/e09a34e0-f752-4b17-b7e1-4d7bb680ad13

—

The previous outline had an outer offset and was cropped.

https://github.com/user-attachments/assets/974f9c37-b602-4e7e-9383-a33f0106c503

The updated version:

https://github.com/user-attachments/assets/9573b82b-7737-43f4-9938-568b98adf366

### DIFF
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -25,7 +25,7 @@
 	"devDependencies": {
 		"@anthropic-ai/sdk": "^0.59.0",
 		"@gitbutler/core": "workspace:*",
-		"@gitbutler/design-core": "^1.3.2",
+		"@gitbutler/design-core": "1.3.4",
 		"@gitbutler/shared": "workspace:*",
 		"@gitbutler/svelte-comment-injector": "workspace:*",
 		"@gitbutler/ui": "workspace:*",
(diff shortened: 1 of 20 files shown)


PavelLaptev:
cc @mtsgrd

PR #11311, opened 2025-11-20
Generate the TS defintions from Rust

estib-vega:
Add the ts type definitions from the Rust types


.github/workflows/push.yaml
```diff
@@ -60,8 +60,23 @@ jobs:
...
+    container:
+      image: ghcr.io/gitbutlerapp/ci-base-image:latest
+    env:
+      CARGO_TERM_COLOR: always
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/init-env-node
+      - run: pnpm generate-ts-definitions
```
Byron, line 76: Can we have another step that prints a warning or even errors so I can see the files that changed?
I find the (incredibly hard to read for me) node based linting quite indirect in comparison.

krlvi:
oh this is really cool!

PR #11313, opened 2025-11-20
"Lanes & branches" new global settings domain

PavelLaptev:
Related to https://github.com/gitbutlerapp/gitbutler/pull/11309

Since we had different settings for lanes and branches here and there from the UX POv it's better to consolidate them in a single place. 

---

1. Implement an auto-select feature for branch names when creating a new branch.

https://github.com/user-attachments/assets/0cb9a4c3-ff60-40ff-9927-3f166ddc4017

2. Move the “new lane placement” option to the settings menu. Additionally, add a button to the “create new branch” modal to access this option.

https://github.com/user-attachments/assets/0137069f-9424-4e25-b4a8-8d0209a05a0c


PavelLaptev:
cc @krlvi

jpsirois:
OMG This is perfect!! Thank you! 😁

Byron:
Thanks so much @PavelLaptev !

krlvi:
@PavelLaptev looks great. I am just thinking, regarding the first option (auto select input) - is there a case where this shouldnt be the case?
Perhaps that could be the default behavior and then we dont need even a setting for it

PR #11317, opened 2025-11-21
Use the latest version of linux dependencies when publishing

Byron:
That way, it seems to not freeze anymore.

This is an experiment that needs validation, but if it builds, it's probably good.

### Tasks

* [x] create new Nightly: [publish](https://github.com/gitbutlerapp/gitbutler/actions/runs/19562568756)
* [ ] validate it: https://github.com/gitbutlerapp/gitbutler/issues/9955#issuecomment-3561760600

Related to  #9955

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -184,12 +184,12 @@ jobs:
             xdg-utils;
 
           sudo apt install -y \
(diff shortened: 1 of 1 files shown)


Byron:
The experiment failed, version pinning was done to prevent another issue :D.

PR #11327, opened 2025-11-21
RS 2 TS: Stacks and StackDetails

estib-vega:
Use the correct rust-generated TS types for the Stacks, StackDetails and their type dependencies

Adapt a bunch of code around this


estib-vega:
Bypassing the reviewers again as these are only TS type changes

PR #11332, opened 2025-11-23
maintenance and CLI refactor

Byron:
### Tasks

* [x] upgrade `gix`
* [x] let `but` crate support its own utilities, by clearly separating command code.
* [x] nightly fmt


crates/but/src/id.rs
```diff
@@ -217,15 +217,15 @@ impl CliId {
         if s.len() > 2 {
             // For longer strings, try prefix matching on CliIds
             let mut cli_matches = Vec::new();
-            crate::status::all_files(ctx)?
+            crate::command::status::all_files(ctx)?
```
Byron, line 220: Just wanted to bring this PR to the attention of @jonathantanmy2, as it should greatly facilitate cleaning up the Id related code. This can probably be led by rewriting the integration tests to not use the `but` CLI, but to use the type directly.
It's find to export it like `but` is a library crate, but it's just as fine to keep everything private and use unit-level tests instead, merely because it helps to weed out parts that aren't actually used.

PR #11333, opened 2025-11-23
legacy-commands for but-cli

Byron:
That way we can keep working on it, but also have a fighting chance of porting commands over explicitly, which also tests the (possibly new) APIs.

Note that this should be implemented so that the code stays as clean as possible without extensive feature-toggle attributes.
It is at this point still possible that this PR is abandoned if it turns out to be too ugly, but ideally it's the same as in
the `but-api`.

### Tasks

* [x] PoC of a minimally invasive feature toggle.
* [x] first version of a building and running minimal non-legacy `but` CLI
* [x] move all legacy commands into `commands::legacy`
* [ ] ~~see if we can have a non-legacy test just for '-h'~~ not really, no way to remove default features.


crates/but/src/legacy/id.rs
```diff
@@ -158,7 +158,7 @@ impl CliId {
 
         // Only try SHA matching if the input looks like a hex string
         if sha_prefix.chars().all(|c| c.is_ascii_hexdigit()) && sha_prefix.len() >= 4 {
-            let all_commits = crate::utils::commits::all_commits(ctx)?;
+            let all_commits = crate::legacy::commits::all_commits(ctx)?;
```
Byron, line 161: @jonathantanmy2 I had to move the `Id` related impl to `legacy` because it's now quite clear that it uses legacy code.
From tomorrow I'd hope to be able to provide more of a blueprint on how to 'de-legacy' it.

PR #11334, opened 2025-11-23
feat(ui): add scrollToTop and auto-scroll on tab change in settings modal

nshcr:
## 🧢 Changes

The Global settings and Project settings windows in GitButler don't reset the scroll position when switching tabs, which causes each newly switched tab to start in the middle.

This PR automatically resets the scroll position when switching tabs, for a better user experience.

Before and after comparisons:

https://github.com/user-attachments/assets/057ef2c6-9ee9-443c-8ef6-8aebba6c401f

### DIFF
--- a/apps/desktop/src/components/ConfigurableScrollableContainer.svelte
+++ b/apps/desktop/src/components/ConfigurableScrollableContainer.svelte
@@ -18,6 +18,13 @@
 			scrollableContainer.scrollToBottom();
 		}
 	}
+
+	// Export method to scroll to top
+	export function scrollToTop() {
+		if (scrollableContainer?.scrollToTop) {
+			scrollableContainer.scrollToTop();
+		}
+	}
 </script>
 
 <ScrollableContainer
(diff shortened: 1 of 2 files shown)


Byron (commented):
Thanks so much for the fix, it's much appreciated!

I can definitely reproduce it as well, and can validate that this PR fixes it.

https://github.com/user-attachments/assets/30a22dc4-01f0-41ee-bd0f-100c86aa20f5

The video shows that it now scrolls up visibly, and I wonder if this can be invisible as well as it's done ahead of time?

In any case, @estib-vega or @Caleb-T-Owens would want to take a look as well.

nshcr:
@Byron Thank you for verifying this PR!

I agree that making the scrolling process invisible would be better, and I also tried making the scroll position jump back to the top instantly with no upward scrolling animation.

However, after testing, I found that the effect wasn't ideal. Although the page content does jump back to the top immediately, the scrollbar indicator on the right stays in its previous position instead of resetting in sync.

This creates a slight visual inconsistency. Keeping the brief scrolling motion actually makes the whole interaction feel smoother and more natural.

Of course, if you have a better implementation in mind, feel free to directly modify my PR or just close it.

Or, perhaps @PavelLaptev might have a better suggestion from a UX perspective?

PavelLaptev:
@nshcr thanks for the PR! I'll take a look.

PavelLaptev:
@Byron @nshcr, let’s merge the two until we find a better solution.
I’ve tried to make the scroll instantaneous, but then the custom scrollbar position didn’t update properly.

PR #11335, opened 2025-11-23
feat: add support for JSON variants (jsonc/json5/jsonl)

nshcr:
## 🧢 Changes

Add support for several JSON variant file formats.

At the moment, JSON5 doesn't have good syntax highlighting yet, but JSONC and JSONL work well.


Caleb-T-Owens:
@nshcr Could you provide some screenshots of what these look like.

I'm wondering if we could just use a JS parse for JSON5?

nshcr:
@Caleb-T-Owens Thanks for pointing that out! That makes sense.

Using the JS parser to highlight JSON5 definitely works. I've pushed the branch again with the corresponding changes, and here's a before/after comparison.




By the way, here are the highlighting results for the other two formats, JSONC can also be highlighted using the JS parser to handle comments as well.

Caleb-T-Owens:
Looks great! Thank you so much for the contribution. These little QOL changes make a big difference.

PR #11348, opened 2025-11-24
Avoid infinite loop by tracking previous head

krlvi:
Prevent infinite reactivity loops by storing the previous head value and
only invalidating stacks and details when the head actually changes. The
change introduces a previousHead state variable and uses untrack to 
update it and call stackService.invalidateStacksAndDetails() only when 
head differs from previousHead.


krlvi:
did not fix the issue

krlvi:
Or maybe it did fix it https://discord.com/channels/1060193121130000425/1073202153163857920/1442609982238625893

PR #11359, opened 2025-11-25
workspace reconciliation for the refmetadata adapter

Byron:
This makes it possible to reconcile whenever new code may have created the need for it, so we can be sure that the old code will see the workspace similarly to new code.


Byron:
@Caleb-T-Owens It's happening! This is the thing we talked about, and reconciliation now also happens when new code alters metadata in such a way that it will be written back. This might help some, but it's also not perfect. In any case, I like that the code now lives with `but-meta`, where it belongs (most).

PR #11366, opened 2025-11-25
Clarifying `assignments_with_fallback`

jonathantanmy2:
I was investigating reducing legacy code in "status" and came across `changes_in_worktree`  in `crates/but-api/src/legacy/diff.rs`, which uses this function. There were some error cases that I didn't fully understand, so I made some refactorings that show how my understanding of the situation (and my understanding may be wrong or incomplete).

In particular, I don't think that the fallback in `assignments_with_fallback` is necessary. But if it is necessary, we shouldn't have `anyhow::Error` twice in the return type, but have an enum as the inner type (so, `Result<(Vec<HunkAssignment>, Option<AssignmentError>)>`) so that callers can know exactly the cases in which we fallback (I didn't do that in this PR).

PTAL - the first 2 commits can be merged as-is, but the 3rd commit (but-hunk-assignment: clarify return value of assignments_with_fallback) has an open question.

### DIFF
--- a/crates/but-hunk-assignment/src/lib.rs
+++ b/crates/but-hunk-assignment/src/lib.rs
@@ -275,7 +275,7 @@ pub fn assign(
         &applied_stacks,
         MultipleOverlapping::SetMostLines,
         true,
-    )?;
+    );
 
     // Reconcile with the requested changes
     let with_requests = reconcile::assignments(
@@ -284,7 +284,7 @@ pub fn assign(
         &applied_stacks,
         MultipleOverlapping::SetMostLines,
         true,
-    )?;
+    );
 
     // Reconcile with hunk locks
     let lock_assignments = hunk_dependency_assignments(deps)?;
@@ -294,7 +294,7 @@ pub fn assign(
         &applied_stacks,
         MultipleOverlapping::SetNone,
         false,
-    )?;
+    );
 
     state::set_assignments(db, with_locks.clone())?;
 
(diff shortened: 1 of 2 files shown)


Byron (approved):
That's a great catch, thanks a lot!

It took me a good squint to understand that apparently, all that `reconcile_worktree_changes_with_worktree_and_locks` does was to wrap its error into an `Option`. The refactoring made that clear to the point where I learned that `?` can apparently already do that automatically.

In any case, `assignments_with_fallback` doesn't seem necessary then.

I will leave it to @krlvi to take another look though as he should be more familiar.

krlvi (approved):
This looks like an appropriate change. 

Some context to this - the `but-hunk-assignment` code is the bit that facilitates multiple "staging" areas - which is simply a mechanism that allows users to "assign" an uncommitted change to a particular lane. The persistence of that state happens outside of git itself (sqlite, table `hunk_assignments`).

It was our intent to make this as an "optional / auxiliary" subsystem such that if it fails, it doesn't disrupt the regular functionality in the app. Over time, the the frontend (which consumes the same API) grew more dependent on having the assignments information available in all cases, thus the fallback mechanism were added. The fallback being - returning an empty array.

Thank you for the cleanup! This double error definitely was unnecessary.

jonathantanmy2:
> It took me a good squint to understand that apparently, all that `reconcile_worktree_changes_with_worktree_and_locks` does was to wrap its error into an `Option`. The refactoring made that clear to the point where I learned that `?` can apparently already do that automatically.

Just for clarification, it also converts `core::TreeChange` into `HunkAssignment`, and can automatically generate `core::TreeChange` and `HunkDependencies` if they are not provided by the caller.

jonathantanmy2:
> Thank you for the cleanup! This double error definitely was unnecessary.

Thanks! I'll go ahead and merge this. We probably still need a better name than `assignments_with_fallback`, but that can be done later.

PR #11371, opened 2025-11-26
Single-branch support for `but branch apply`

Byron:
Port `but branch apply` over to non-legacy code, make sure it builds in non-legacy mode as well.

### Tasks

* [x] add CI validation that it now exists in non-legacy builds.
* [x] add all the tests for the status quo
* [x] de-'legacy' oplog to be able to use 
* [x] new 'apply' function in but-api
* [x] make existing tests work
* [x] fix CI 💢

### Next PR
* [ ] #11400 

Follow-up on #11236.


krlvi:
Looks good to me. One thing that i found interesting in the PR title "Single-branch support for but branch apply" - I think I know what you meant but also, per definition the apply function is about adding an additional branch to the workspace, so perhaps 'single branch' may be a misnomer in this case

krlvi (approved):
thanks, for this, LGTM

Byron:
Thanks you!

> [..] One thing that i found interesting in the PR title "Single-branch support for but branch apply" - I think I know what you meant but also, per definition the apply function is about adding an additional branch to the workspace, so perhaps 'single branch' may be a misnomer in this case

What if I told you that apply *can* be used in single branch mode, e.g. while on `main`, apply a feature branch and it will automatically create a workspace for you. Without target, you will now see `main` and `feature`. But if `main` is configured as target, you will see only `feature`.

Single-branch mode for me means that all functions work everywhere, or at least have meaningful results or graceful failure if it's just not possible (like unapplying a branch, that really is only for real workspaces, from what I can tell).

PR #11373, opened 2025-11-26
Fix text editor wrapping and improve test reliability

mtsgrd:
- Fix bug where typing in middle of paragraph lost characters during 
rewrap Replace setTimeout with deterministic waits in HardWrapPlugin 
tests Remove legacy MarkdownTransitionPlugin and simplify initialization
- Add tests for automatic wrapping on init (enabled/disabled)
- Fix typos and improve null checking in RichTextEditor


mtsgrd:
@PavelLaptev fyi - if you could keep your eyes out for any regression related to this pr. It's a pretty substantial simplification to the rich text editor in that it removes the `plainText` prop.

packages/ui/src/lib/richText/RichTextEditor.svelte
```diff
@@ -355,19 +315,6 @@
...
-		z-index: 0;
-		position: relative;
-		flex-direction: column;
-		height: 100%;
-		overflow: auto;
-		border: 0;
-		outline: 0;
-		/* It's unclear why the resizer is on by default on this element. */
-		resize: none;
-	}
```
mtsgrd, line 369: @PavelLaptev this looks a bit sus, but also it wasn't being used.

PR #11375, opened 2025-11-26
Generalize testsupport mode normalization by filetype

slarse:
## 🧢 Changes
Continuation of #11363 with a suggestion to generalize the testsupport file mode normalization.

### DIFF
--- a/crates/but-testsupport/src/lib.rs
+++ b/crates/but-testsupport/src/lib.rs
@@ -289,6 +289,8 @@ pub fn visualize_tree(tree_id: gix::Id<'_>) -> termtree::Tree<String> {
     visualize_tree(tree_id.object().unwrap().peel_to_tree().unwrap().id(), None).unwrap()
 }
 
+const FILETYPE_MASK: u32 = 0o170000;
+
 /// Visualize a tree on disk with mode information.
 /// For convenience, skip `.git` and don't display the root.
 ///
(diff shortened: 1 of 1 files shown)


crates/but-testsupport/src/lib.rs
```diff
@@ -304,20 +306,22 @@ pub fn visualize_tree(tree_id: gix::Id<'_>) -> termtree::Tree<String> {
...
-            0o100777 => 0o100755,
-            0o120775 => 0o120755,
-            0o120777 => 0o120755,
-            other => other,
+        let normalize_permission_bits = |mode: u32| if mode & 0o100 == 0 { 0o644 } else { 0o755 };
+
+        match mode & FILETYPE_MASK {
+            // File types Git cares about
+            directory @ 0o40000 => directory | normalize_permission_bits(mode),
+            symlink @ 0o120000 => symlink | normalize_permission_bits(mode),
```
slarse, line 314: From a Git perspective it's a bit weird to propagate permission bits on directories and symlinks, as Git does not store those. But the current snapshots do so.

Do we want to "properly" normalize this to a more Git-aligned view (i.e. zero out the permission bits for these types in existing snapshots), or is this fine?

slarse: I realize that if we don't care about "proper" Git normalization for the snapshots, this entire thing could be reduced to just:

```rust
    fn normalize_mode(mode: u32) -> u32 {
        let normalize_permission_bits = |mode: u32| if mode & 0o100 == 0 { 0o644 } else { 0o755 };
        (mode & 0o170000) | normalize_permission_bits(mode)
    }
```

Because every match arm is just doing the same thing right now, except for the panic, which I don't feel strongly about keeping.

Byron: Yes, why not go with the minimal version that also can't panic. Let's merge this as well and you can follow it up.

Byron (approved):
Thanks so much for picking this up!

Despite merging the PR now, I'd also be looking forward to the two-line version that won't panic :).

crates/but-testsupport/src/lib.rs
slarse: @Byron Your word is my law, here's the followup: #11389 

I had too much fun playing with bits not to do it...

PR #11377, opened 2025-11-27
Revert "Add custom PR message prompt support for concistency" - didn't work properly

Byron:
This reverts commit 52c34ac86347f8fda613b7db4a76aa58662bd5eb.

See https://github.com/gitbutlerapp/gitbutler/pull/11370#issuecomment-3583299801 for more information.

Also when testing it in the nightly, it didn't do anything anymore. Very strange.

It's also notable how I made assumptions about: "I can edit the first bubble, and it picks up, so it works",
even though the second bubble can't be edited.
Also, now it doesn't seem to pickup anymore.

### DIFF
--- a/apps/desktop/src/components/AIPromptEdit.svelte
+++ b/apps/desktop/src/components/AIPromptEdit.svelte
@@ -7,7 +7,7 @@
 	import type { Prompts, UserPrompt } from '$lib/ai/types';
 
 	interface Props {
-		promptUse: 'commits' | 'branches' | 'pullRequests';
+		promptUse: 'commits' | 'branches';
 	}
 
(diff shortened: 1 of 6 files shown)


krlvi:
@Byron I think Pavel forward-fixed this here https://github.com/gitbutlerapp/gitbutler/pull/11376 so maybe revert was not necessary

Byron:
Right, I didn't really notice this, but maybe insta-revert was an overreaction. 

However, I did try it again in Nightly before it was reverted and it didn't work anymore, as in: had no effect. Even though the first test I made with a local build worked at least with system message changes. It's puzzling, and might show that there is more to this that requires more attention than I was able to give.

PR #11378, opened 2025-11-27
Add AppStream MetaInfo file

NN708:
## 🧢 Changes

This PR introduces [AppStream MetaInfo file](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html), which is essential for Linux packaging like Flatpak.

## ☕️ Reasoning

Currently, the MetaInfo file for GitButler is maintained in the [flathub/com.gitbutler.gitbutler](https://github.com/flathub/com.gitbutler.gitbutler) repository. However, [Flathub's requirements](https://docs.flathub.org/docs/for-app-authors/requirements#required-metadata) recommend including it in the upstream repository, as it is not specific to Flatpak and contains release information that could be better maintained upstream.

### DIFF
--- a/crates/gitbutler-tauri/com.gitbutler.gitbutler.metainfo.xml
+++ b/crates/gitbutler-tauri/com.gitbutler.gitbutler.metainfo.xml
@@ -1,0 +1,157 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.gitbutler.gitbutler</id>
+
+  <metadata_license>FSFAP</metadata_license>
+  <project_license>LicenseRef-FSL-1.1-MIT=https://github.com/gitbutlerapp/gitbutler/blob/master/LICENSE.md</project_license>
+
+  <name>GitButler</name>
+  <summary>Git, finally designed for humans</summary>
+
+  <developer id="com.gitbutler">
+    <name>GitButler Inc.</name>
+  </developer>
+
+  <description>
+    <p>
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot for helping with this, it's much appreciated.

Can you help me understand how this file needs to be maintained moving forward?
How does it interact with the flathub workflow (if there is any)?

Thanks in advance.

Also CC @krlvi who is mostly doing our releases, which is quite some steps to follow for the stable ones.

NN708:
@Byron This file contains all the metadata to be displayed on the [Flathub website](https://flathub.org/en/apps/com.gitbutler.gitbutler) and in GNOME Software. Most of this information won't need constant updating, only the `releases` section should be prepended to with each new release.

Since this is just metadata, it doesn't interact directly with the Flathub workflow. If it isn't updated for a release, the app can still be built, but the website won't display the correct version information.

Byron (approved):
Thanks a lot for elaborating, I think I have a good understanding now.

From what I see, it's something we'd want to support the Linux releases better, and it seems to be quite easy to maintain.

If @krlvi agrees to this in principle so this can be merged, maybe he can have a look at the actual metadata and see if it fits still.

krlvi:
yeah seems good to me. Interesting with the version numbers in this mata info file... this will be easy to forget to update

krlvi (approved):

Byron:
Yeah, it's definitely another bullet point on your release-list. And let's see how it goes - personally I still feel very detached from that file, but that is true for the Linux distribute process in general.

PR #11389, opened 2025-11-27
Simplify testsupport mode normalization

slarse:
## 🧢 Changes
Continuation on #11375 to further simplify the mode normalization. Tried it with `umask=0077/0022/0002/0000`, so from very strict to wildly lax, all good. There really shouldn't be any issues unless the umask affects the user/owner bits, but at that point there's really little that can be done.


Byron (approved):
Loving it 🙏

PR #11390, opened 2025-11-27
Redesign branhces reordering

PavelLaptev:
https://github.com/user-attachments/assets/cdc6140a-0b1b-43d8-baea-b60f8a773b53


krlvi:
This design is awesome!

PavelLaptev:
> This design is awesome!

Thanks! I’m currently battling a failing test and will merge it as soon as I resolve it 🤪

We’ll also make some nice updates to the documentation for this feature.

krlvi:
I wonder if we can give similar treatment to commit reordering as well

PavelLaptev:
> I wonder if we can give similar treatment to commit reordering as well

Displaying a “dropping line” between each commit wouldn’t be visually good, unlike branches, they are too close to each other. Additionally, there could be many of them, which would appear more like strange separators rather than indicators that you can drop a commit between two items. Another reason against doing so is that we also have drop zones for commits there.

image

I believe our current implementation is quite appropriate and familiar. Here’s a linear example of reordering tasks.

https://github.com/user-attachments/assets/9ef89ba3-32dc-4185-a396-23fbc892b08a

---

We can improve the commits dragging functionality by displaying the label operation label on the draggable element when it hovers over the drop area. Currently, if I drag a commit into another one, I don’t see the label operation label because the draggable element covers it.

image

PR #11402, opened 2025-11-28
When running `pnpm tauri dev`, make sure askpass is checked in the right directory

mathijs81:
When running `pnpm tauri dev`, make sure askpass is checked in the right directory

## ☕️ Reasoning

I ran into issues with errors that askpass can't be found when running the tauri app in dev mode. This fixes it for me.


Byron:
Thanks so much for getting this started!

Also for the motivation to try another route which, in hindsight, might not be much better except that it's focussing on helping the user to do the right thing, instead of keeping to tell them the wrong thing and than deal with that more smartly.

PR #11405, opened 2025-11-28
Allow inline message for but describe

mtsgrd:
Add an optional -m/--message flag to the describe command so users can
supply a new commit message or branch name on the command line instead
of opening an editor. This change threads the optional message through
the describe handlers and uses it to validate and apply the supplied
value, when absent it falls back to the existing editor-based flow.

Add tests for 'but describe' commit message handling.

### DIFF
--- a/crates/but/tests/but/command/describe.rs
+++ b/crates/but/tests/but/command/describe.rs
@@ -1,0 +1,154 @@
+use snapbox::str;
+
+use crate::utils::{Sandbox, setup_metadata};
+
+#[test]
+fn describe_commit_with_message_flag() -> anyhow::Result<()> {
+    let env = Sandbox::init_scenario_with_target_and_default_settings("one-stack")?;
+    insta::assert_snapshot!(env.git_log()?, @r"
+    * edd3eb7 (HEAD -> gitbutler/workspace) GitButler Workspace Commit
+    * 9477ae7 (A) add A
+    * 0dc3733 (origin/main, origin/HEAD, main) add M
+    ");
+
+    setup_metadata(&env, &["A"])?;
+
+    // Use describe with -m flag to change commit message (using commit ID)
+    env.but("describe 9477ae7 -m 'Updated commit message'")
+        .assert()
+        .success()
+        .stdout_eq(str![[r#"
+Updated commit message for [..] (now [..])
+
+"#]]);
+
+    // Verify the commit message was updated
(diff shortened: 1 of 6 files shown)


mtsgrd:
@Byron this is largely claude generated, but it makes it much easier to edit commit messages with `but` through our codegen. Could you have a quick look and let me know what I need to do in order to merge this?

crates/but/tests/but/command/describe.rs
```diff
@@ -0,0 +1,154 @@
...
+    let log = env.git_log()?;
+    assert!(log.contains("Updated commit message"));
+    assert!(!log.contains("add A"));
+
+    Ok(())
+}
+
+// Note: Branch rename test is omitted because the test scenario uses single-character
+// branch names ("A") which don't meet the 2-character minimum requirement for CLI IDs.
+// The branch rename functionality with -m flag is tested manually and works correctly.
```
Byron: @jonathantanmy2 

I thought this would be interesting to you - is it CLI IDs that have to adjust, or is it the commands that should be able to fallback to "could this be a very short name for an entity"? To me it feels that CLI Ids could handle this, as they usually have to know the entire set of possible entities and thus can deal with short names, which it seems like could just be an ID on their own.

Byron (commented):
Thanks for the invite - I will push some minor refactors here and merge the result.

Also, I realise that the commit-details function used here is only legacy because it uses legacy types, and it would take a final push to be able to turn it into non-legacy with the api-cmd macro so that `ProjectId` style functions can be become 'modern'. So I think I'd want to do that this week as well.

Byron:
The main behavioural change I made was that changing something to the same message is not an error. If the user says they want it to be message "X", and it is message "X", it's what they want which is good - that's the motivating factor behind that.
One could also say that no-change is a good change. Similar behaviour is already implemented in "branch apply" - applying the same branch again does nothing, and is happy about that.

Besides that, generally we'd want to limit these slow (but fortunately not flaky) tests to the bare minimum, typically happy paths *and* UX or messages that are really important to us. What happens when it can't find something might be something we want to test, but I have a feeling that with some framework around selecting stacks, commits, and hunks we can test this centrally and don't have to test it for each command.

Lastly, I think AI does a fine job but should do better when it comes to deduplication, and of course not spamming tests.

mtsgrd:
Awesome, thanks @Byron!

Issue #5282, opened 2024-10-23
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...


esamson:
### Version

0.13.6

### Operating System

Linux

### Distribution Method

AppImage (Linux)

### Describe the issue

Upon launch, I just get an empty window.

I'm using Fedora 40 (KDE Plasma) with Wayland.

### Relevant log output

```shell
** (git-butler:3207): WARNING **: 21:29:40.424: webkit_settings_set_enable_offline_web_application_cache is deprecated and does nothing.
2024-10-23T13:29:40.445664Z  INFO crates/gitbutler-tauri/src/main.rs:64: system git executable for fetch/push: "git"
2024-10-23T13:29:40.445781Z  INFO crates/gitbutler-tauri/src/main.rs:101: starting app version=0.13.6 name=GitButler
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...

** (git-butler:3207): WARNING **: 21:30:05.447: atk-bridge: get_device_events_reply: unknown signature
```

Keloran:
if you need to download the old one to get your system working again
```
https://releases.gitbutler.com/releases/release/0.13.5-1375/linux/x86_64/git-butler_0.13.5_amd64.AppImage.tar.gz
```

Byron:
Thanks a lot for reporting!

The compatibility story for the Linux distribution, particularly related to AppImage, isn't great, and sometimes there is no good solution, from my experience.

But also, I am fortunately not the right one to talk to about this, but @ndom91 is and knows all about it.

ndom91:
Hey thanks for reporting this. So EGL is referring to OpenGL. I ran into this yesterday as well and spent a bunch of time going back and forth between different opengl / mesa versions trying to pin this down, I'm on NixOS. Downgrading to the version on the `24.04` nixpkgs (`mesa@24.0.7`) allowed me to run these newer releases again.

What I've found so far:

- Nightly `0.5.826` is the last working version for me ([AppImage download](https://releases.gitbutler.com/releases/nightly/0.5.826-1374/linux/x86_64/git-butler-nightly_0.5.826_amd64.AppImage.tar.gz))
[...]

PR #11412, opened 2025-11-30
Add Ubuntu 22.04 builds

slarse:
## 🧢 Changes
Replaces the default Ubuntu 24.04 build with a Ubuntu 22.04 build.

The Ubuntu 24.04 build is hidden away under a different S3 object key. Depending on how the access control is configured on the S3 bucket, configuration may be needed to access the Ubuntu 24.04 build.

<details>
<summary>The original reasoning of this post (pretty outdated now)</summary>
This PR does two (highly related) things.

1. Adds Ubuntu 22.04 builds for x86_64 and ARM64 (side note: are the Linux ARM64 builds actually used??)
2. ~Upgrades the core Linux dependencies, most notably libwebkit2gtk-4.1~
    - Due to large forward-compatibility issues, we'll take this separately

I've been fiddling around with this in my fork to try stuff out in the actual runners, and produced [the Ubuntu 22.04 and Ubuntu 24.04 builds found here](https://github.com/slarse/gitbutler/actions/runs/19802219644).

> Note: Don't mind the fact that the artifacts are 500 MB each, I was veeery lazy and just added the entire `target/release/bundle` directory as an artifact. In other words, the artifacts contain way more stuff than they need to.

The Ubuntu 24.04 AppImage appears to be working very well on my Arch Linux system under Wayland.

I was hoping the Ubuntu 22.04 AppImage would also work, but I'm having some very strange linking issues with it when running on my Arch machine (see below). So, I do believe we need to build for both older and newer systems, independently.

<details>
<summary>Possible linker error when running Ubuntu 22.04 AppImage on Arch Linux</summary>

I run into this error when trying to work with remotes.

```bash
command: fetch_from_remotes
params: {"projectId":"ff91a7a9-1b3d-4b71-abee-6254bc1b7731","action":"modal"})

backend error: git command exited with non-zero exit code 128: ["fetch", "--quiet", "--prune", "upstream", "+refs/heads/*:refs/remotes/upstream/*"]

STDOUT:


STDERR:
git: /tmp/.mount_GitButGDbenp/usr/lib/libpcre2-8.so.0: no version information available (required by git)
/usr/lib/git-core/git: /tmp/.mount_GitButGDbenp/usr/lib/libpcre2-8.so.0: no version information available (required by /usr/lib/git-core/git)
/usr/lib/git-core/git-remote-https: symbol lookup error: /usr/lib/libcurl.so.4: undefined symbol: nghttp2_option_set_no_rfc9113_leading_and_trailing_ws_validation
fatal: remote helper 'https' aborted session
```

The strangeness is that `nghttp2_option_set_no_rfc9113_leading_and_trailing_ws_validation` is absolutely defined in `/usr/lib/libcurl.so.4`, so there has to be some linker issue. This error does not occur when using the Ubuntu 24.04 AppImage.
</details>

## ☕️ Reasoning

[See this comment](https://github.com/gitbutlerapp/gitbutler/issues/7671#issuecomment-3591558807) for a pretty comprehensive justification for the addition of the Ubuntu 22.04 build.

As for the dependency upgrades, they are well overdue. The WebkitGTK build that's currently used is years old and pretty buggy. Upgrading the dependencies fixes the freeze issue described in #9955 on my Arch Linux machine, and probably touches many of the other issues reported on the Linux bundles, the AppImage in particular.

## 🎫 Affected issues

Fixes: #9955 (The dependency upgrades in the Ubuntu 24.04 image seems to eliminate this problem)

Also improves upon #7671 with the Ubuntu 22.04 build, which links to GLIBC 2.35 instead of 2.38 (as is the case in the Ubuntu 24.04 build). This will let the AppImage run on older distros.

Very likely touches upon more of the Linux-specific issues, but I have not had the time to evaluate more of them.

## TODO
On 1/3 systems I've tested on (another Arch Linux+Wayland machine), I'm getting `Could not create default EGL display: EGL_BAD_PARAMETER. Aborting ...`. Which is strange. I want to figure that out before proceeding here.

### Testing matrix (AppImage)

| Machine | AppImage | Status | Workaround |
| ------------- | --------------- | --------- | ------------------ |
| Arch Linux + Wayland (mesa) | Ubuntu 22.04 | :x: EGL_BAD_PARAMETER | https://github.com/gitbutlerapp/gitbutler/issues/5282#issuecomment-3598600651 (poor xwayland performance) |
| Arch Linux + Wayland (mesa) | Ubuntu 24.04 | :x: EGL_BAD_PARAMETER | https://github.com/gitbutlerapp/gitbutler/issues/5282#issuecomment-3598600651 (poor xwayland performance) |
| Arch Linux + Wayland (Nvidia) | Ubuntu 22.04 | :warning: Cannot work with remotes ||
| Arch Linux + Wayland (Nvidia) | Ubuntu 24.04 | :heavy_check_mark: No issues detected| |
| Ubuntu 22.04 + X11 (mesa) | Ubuntu 22.04 | :heavy_check_mark: No issues detected | |
| Ubuntu 22.04 + X11 (mesa) | Ubuntu 24.04 | :x: Cannot launch due to GLIBC mismatch | |

</detalis>

## Important caveat: The Ubuntu 22.04 build may not last
I don't know for how long the Ubuntu 22.04 build will keep on building, as the WebKit project dropped official support for it [on 2025-04-25](https://docs.webkit.org/Ports/WebKitGTK%20and%20WPE%20WebKit/DependenciesPolicy.html). Which is odd as Debian Bookworm (the Ubuntu 22.04 base distro) has support until June 2026. `libwebkit2gtk-4.1` is however [still an officially Ubuntu-maintained package](https://packages.ubuntu.com/jammy/libwebkit2gtk-4.1-dev), and the current upstream version 2.50.1 [was released in October](https://webkitgtk.org/2025/10/10/webkitgtk2.50.1-released.html). I think that's sufficient recent activity to make it worthwhile to start running the build.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -96,9 +96,9 @@ jobs:
             target: ''
           - platform: macos-15 # [macOs, x64] - cross-compile
             target: x86_64-apple-darwin
-          - platform: ubuntu-24.04 # [linux, x64]
+          - platform: ubuntu-22.04 # [linux, x64]
             target: ''
-          - platform: ubuntu-24.04-arm # [linux, ARM64]
+          - platform: ubuntu-22.04-arm # [linux, ARM64]
             target: ''
           - platform: windows-latest # [windows, x64]
             target: ''
@@ -141,7 +141,7 @@ jobs:
       - name: Rust Cache ${{ matrix.target }}
         uses: Swatinem/rust-cache@v2.8.2
         with:
-          shared-key: app-release-build-${{ github.event.inputs.channel || 'nightly' }}${{ matrix.target }}
+          shared-key: app-release-build-${{ github.event.inputs.channel || 'nightly' }}${{ matrix.target }}${{ matrix.platform }}
       - name: Init Node Environment
         uses: ./.github/actions/init-env-node
 
@@ -183,13 +183,14 @@ jobs:
             librsvg2-dev \
             xdg-utils;
 
+          webkit_version='2.50.*'
           sudo apt install -y \
-            libwebkit2gtk-4.1-0=2.44.0-2 \
-            libwebkit2gtk-4.1-dev=2.44.0-2 \
-            libjavascriptcoregtk-4.1-0=2.44.0-2 \
-            libjavascriptcoregtk-4.1-dev=2.44.0-2 \
-            gir1.2-javascriptcoregtk-4.1=2.44.0-2 \
-            gir1.2-webkit2-4.1=2.44.0-2;
+            libwebkit2gtk-4.1-0="$webkit_version" \
+            libwebkit2gtk-4.1-dev="$webkit_version" \
+            libjavascriptcoregtk-4.1-0="$webkit_version" \
+            libjavascriptcoregtk-4.1-dev="$webkit_version" \
+            gir1.2-javascriptcoregtk-4.1="$webkit_version" \
+            gir1.2-webkit2-4.1="$webkit_version";
 
       - uses: actions/download-artifact@v7
         name: Download SvelteKit build output
@@ -348,9 +349,9 @@ jobs:
             target: ''
           - platform: macos-15 # [macOs, x64] - cross-compile
             target: x86_64-apple-darwin
-          - platform: ubuntu-24.04 # [linux, x64]
+          - platform: ubuntu-22.04 # [linux, x64]
             target: ''
-          - platform: ubuntu-24.04-arm # [linux, ARM64]
+          - platform: ubuntu-22.04-arm # [linux, ARM64]
             target: ''
           - platform: windows-latest # [windows, x64]
             target: ''


.github/workflows/publish.yaml
```diff
@@ -184,12 +189,12 @@ jobs:
...
-            libjavascriptcoregtk-4.1-0=2.44.0-2 \
-            libjavascriptcoregtk-4.1-dev=2.44.0-2 \
-            gir1.2-javascriptcoregtk-4.1=2.44.0-2 \
-            gir1.2-webkit2-4.1=2.44.0-2;
+            'libwebkit2gtk-4.1-0=2.50.*' \
+            'libwebkit2gtk-4.1-dev=2.50.*' \
+            'libjavascriptcoregtk-4.1-0=2.50.*' \
+            'libjavascriptcoregtk-4.1-dev=2.50.*' \
+            'gir1.2-javascriptcoregtk-4.1=2.50.*' \
+            'gir1.2-webkit2-4.1=2.50.*';
```
slarse: I believe locking down on the _minor_ version is a good enough tradeoff. That way, we get patches as well as Debian/Ubuntu-specific changes automatically, while minimizing risk that we get a breaking update.

Minor version bumps in libwebkit does seem to introduce drastic changes, so I think it's a good idea to perform such updates under more controlled forms.

.github/workflows/publish.yaml
```diff
@@ -172,6 +176,7 @@ jobs:
         if: runner.os == 'Linux'
         run: |
           sudo apt update;
+          sudo apt upgrade;
```
slarse: The base image for the runner is [most often outdated](https://docs.github.com/en/actions/how-tos/manage-runners/github-hosted-runners/customize-runners), I think it makes sense to ensure that we get the latest and greatest of _everything_ that's installed for whichever distro we're building for.

.github/workflows/publish.yaml
```diff
@@ -100,6 +100,10 @@ jobs:
             target: ''
           - platform: ubuntu-24.04-arm # [linux, ARM64]
             target: ''
+          - platform: ubuntu-22.04 # [linux, x64]
+            target: ''
+          - platform: ubuntu-22.04-arm # [linux, ARM64]
+            target: ''
```
slarse, line 102: As noted in OP, I'm really unsure if this build is desirable.

Byron:
Thanks so much, this PR could be very impactful to the linux users out there!

Without much ado, I am creating a new [nightly](https://github.com/gitbutlerapp/gitbutler/actions/runs/19814434277) from it and invite you to test out the official builds, as downloaded from the main site.

Something I am wondering about is if the additional ubuntu 22 images will confuse the website. And even if not, I think we will need an upgrade so these builds can be differentiated.

> (side note: are the Linux ARM64 builds actually used??)

They are downloadable, but… I am sure I am missing something.

---- 

And it turns out that the build is failing: https://github.com/gitbutlerapp/gitbutler/actions/runs/19814434277/job/56762812224 .

My feeling is that it *might* re-use caches from other ubuntu versions as the shared key only contains the target, and maybe that causes trouble. The failed ubuntu 22 job used `v0-rust-app-release-build-nightly-Linux-x64-341760e0-379a590c`, and the working ubuntu 24 job used `v0-rust-app-release-build-nightly-Linux-x64-341760e0-379a590c` as well. If nothing else, it's probably racy when writing back. But is this really the cause of the failure? I don't know. But I do know that the weirdest errors happen if the Cargo cache is borked, something that regularly happens even when switching compiler versions. And maybe that's the real issue here.

```
 --> crates/but-claude/src/notifications.rs:3:5
  |
3 | use notify_rust::Notification;
  |     ^^^^^^^^^^^ can't find crate

   Compiling gitbutler-edit-mode v0.0.0 (/home/runner/work/gitbutler/gitbutler/crates/gitbutler-edit-mode)
For more information about this error, try `rustc --explain E0463`.
error: could not compile `but-claude` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
failed to build app: failed to build app
       Error [tauri_cli_node] failed to build app: failed to build app
 ELIFECYCLE  Command failed with exit code 1.
```

Let me try that.

Byron:
Ok, here is the new publish: https://github.com/gitbutlerapp/gitbutler/actions/runs/19814866150, and if it passes *and* the website doesn't choke, this PR can definitely be merged when you think it's acceptable.

The website could be updated in a separate step then.

Byron:
The build went through this time, but I am a bit anxious that the download links aren't correct. The filenames aren't qualified enough to differentiate the different linux platforms, and I am afraid that the Ubuntu 22/24 builds are randomly swapped or something. 

Could you validate the linux bundles available on [the website](https://gitbutler.com/nightly)?

Thanks again.

slarse:
@Byron So sorry, I think you caught this _just_ before I made it a draft as I encountered the EGL_BAD_PARAMETER problem on one machine. See the testing matrix I edited into the PR body. I think it's an upstream issue in Tauri, it's reported all over the place. From some very cursory investigation it appears to be a MESA issue, because I don't encounter it when using a Nvidia GPU with proprietary drivers.

Should be noted that I don't encounter the issue on Ubuntu 22.04 using the Ubuntu 22.04 AppImage, even though that machine is running Intel with MESA drivers, but that may be due to having older drivers. Unsure.

But I need to put in some more work here and do some more compatibility testing. I really want an AppImage that works well across all my Arch machines.

> The build went through this time, but I am a bit anxious that the download links aren't correct. The filenames aren't qualified enough to differentiate the different linux platforms, and I am afraid that the Ubuntu 22/24 builds are randomly swapped or something.

Right, I meant to look at this too but somehow it slipped my mind. Will look into that as well.

Byron:
Thanks! Let me create a new nightly build then. This would give you another 20 minutes to download the bundles, I hope it's enough.

slarse:
Darn, I think I win this year's miscommunication award. I'm working full time right now (not on GitButler :( ), and then I have family responsibilities, so work on GB is generally confined to late evenings and weekends. So I didn't have time to pull down those build artifacts :(. Just quickly popped my head in when my work computer had the EGL_BAD_PARAMETER issue.

Sorry I wasted your time today, I was just really eager to get this in after having spent so much time fiddling with builds on different systems. A bit too eager, should have done more testing!

On the bright side, I've diagnosed the EGL_BAD_PARAMETER issue on my system. It's, wait for it... Also a linker issue! Why that doesn't happen with the currently pinned versions is anyone's guess. I put down a pretty little writeup over here: https://github.com/gitbutlerapp/gitbutler/issues/5282#issuecomment-3598600651

The gist of it is that the system's mesa driver links incorrectly with the bundled `libwayland-client.so.0` lib. If I instead use the system's lib by preloading it, it seemingly works fine:

```bash
$ ./GitButler.AppImage
<snip>
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
$ LD_PRELOAD=/usr/lib/libwayland-client.so ./GitButler.AppImage
<snip>
TADAAAAA (it does not actually say that, but you get it)
```

This is a workaround at best. I will give it more of a go tomorrow, but even if it works for me there's no telling if it'll work on a Fedora box, for example.

Linker issues seem to be rather abundant in AppImage Tauri-land, and they are working on something to resolve it: https://github.com/tauri-apps/tauri/pull/12491. I think _that_ will be the be-all-end-all solution to the non-portability of the AppImage.

Not entirely sure what we do in the mean time. I think adding the Ubuntu 22.04 build would be worthwhile regardless, but I'm a bit unsure about bumping the core dependencies given the amount of trouble it causes. Then again, _not_ bumping the deps also causes a whole bunch of random issues. Perhaps the preload workaround is fine for now, and we could always just roll back if _that_ causes too many issues for people.

What do you think?

Byron:
> Darn, I think I win this year's miscommunication award. I'm working full time right now (not on GitButler :( ), and then I have family responsibilities, so work on GB is generally confined to late evenings and weekends. So I didn't have time to pull down those build artifacts :(. Just quickly popped my head in when my work computer had the EGL_BAD_PARAMETER issue.

No worries, at all! It's good that the binaries can still be downloaded, they have just be superseded by a more recent version.
So I don't think any time was wasted (and even if some was, that's life :D).

> Not entirely sure what we do in the mean time. I think adding the Ubuntu 22.04 build would be worthwhile regardless, but I'm a bit unsure about bumping the core dependencies given the amount of trouble it causes. Then again, _not_ bumping the deps also causes a whole bunch of random issues. Perhaps the preload workaround is fine for now, and we could always just roll back if _that_ causes too many issues for people.
> 
> What do you think?

Actually I can't really judge if the trade-off is worth it. To me right now it *looks* like the old dependencies allow the app to launch 'more often', but it may freeze later, and the new dependency version prevent a launch more often unless the `LD_PRELOAD` workaround is used, but it fixes a bunch of random issues as well.

To me neither seems great and I'd love to downgrade linux support to 'experimental' or something. When that's done, we could move to the latest greatest, have a documentation page with workarounds and make sure people know about it, while waiting for https://github.com/tauri-apps/tauri/pull/12491 to land in the hopes for the one good bundle we can provide.

slarse:
> Actually I can't really judge if the trade-off is worth it. To me right now it looks like the old dependencies allow the app to launch 'more often', but it may freeze later, and the new dependency version prevent a launch more often unless the LD_PRELOAD workaround is used, but it fixes a bunch of random issues as well.

It's hard. But to add to the argument for not staying on the old dependencies: they don't get security updates. And WebKitGTK is effectively a web browser. There's a lot to be said about not sticking around on yesterday's web browser version.

I did try the preload workaround a bit more today, and found that it works well. Except that, for some reason, the XWayland session shoots away in CPU usage (10x increase in CPU usage, from ~5% to ~50%). That could just be a performance regression in the system's Wayland client (it happens quite regularly), but it could also be that the system libs don't quite want to play with the AppImage libs. I'm going to try the deb distribution as well and see if that works.

> To me neither seems great and I'd love to downgrade linux support to 'experimental' or something.

This seems like the only reasonable approach. The Linux story is just a bit poor at the moment. It doesn't work on old distros because of the hard glibc dependency, and it doesn't work on new distros because of random WebKitGTK issues.

I'll bring this up in the Discord so we can have a slightly more interactive discussion around it.

scripts/release.sh
```diff
@@ -219,7 +219,15 @@ else
 	BUNDLE_DIR=$(readlink -f "$PWD/../target/release/bundle")
 fi
 
+# The release dir determines a (significant portion of) the final S3 object key.
 RELEASE_DIR="$DIST/$OS/$ARCH"
+if [ "$OS" = "linux" ]; then
+  # We build for multiple linux distros and need distinct keys for them.
+  RELEASE_DIR="$DIST/$OS/$(lsb_release -cs)/$ARCH"
+fi
```
slarse: @Byron As you thought, there was probably a last-write-wins situation going on with the S3 uploads for the Linux releases. This should address that (if I've finally actually understood how the S3 keys are determined), BUT it is not enough. There's a part of the `publish.yaml` workflow that notifies the GB API of the new version https://github.com/gitbutlerapp/gitbutler/blob/0e8b4cc233ce8fbe6826efaba9d6c76830489559/.github/workflows/publish.yaml#L398-L406

A quick peek into that API shows this:

```bash
$ curl -s https://app.gitbutler.com/releases/ | jq
{
  "version": "0.18.2",
  "notes": "- Improved UX of branch reordering, making creating / splitting stacks of branches more intuitive\n- Fixes a bug with commit message / PR description inputs preventing creating new lines (paragraphs)\n- Codegen - fixes a bug with full commands not being shown in permission requests",
  "pub_date": "2025-11-29T14:53:27+00:00",
  "platforms": {
    <snip>
    "linux-aarch64": {
      "signature": "dW50cnVzdGVkIGNvbW1lbnQ6IHNpZ25hdHVyZSBmcm9tIHRhdXJpIHNlY3JldCBrZXkKUlVUck9FSStpbTFYWUxuWGxGNEdyRDhKRHMzdGpQamhacEhOcVJRYTJNYzJHdVl5RDJ0OG9jbytlUXBUYVphWkNBbkw1RWxSYVFPQ1ZMUHZkMWVGV1J3cjRENmZXOTQ4S1FFPQp0cnVzdGVkIGNvbW1lbnQ6IHRpbWVzdGFtcDoxNzY0NDI0NTg3CWZpbGU6R2l0QnV0bGVyXzAuMTguMl9hYXJjaDY0LkFwcEltYWdlLnRhci5negpXSEhmOVhIUHcwVi9mSS9PWkJHOUxwVmxGbmhxY1ZlMjZPQ25LRWYzQXBEc2plVnM5Qms3N0hUV0wrVDNvZjh5bGYrVGIzZEdJWnRzaURYUkFmYUdCQT09Cg==",
      "url": "https://releases.gitbutler.com/releases/release/0.18.2-2660/linux/aarch64/GitButler_0.18.2_aarch64.AppImage.tar.gz"
    },
    "linux-x86_64": {
      "signature": "dW50cnVzdGVkIGNvbW1lbnQ6IHNpZ25hdHVyZSBmcm9tIHRhdXJpIHNlY3JldCBrZXkKUlVUck9FSStpbTFYWUQ0cnVYQ3VEYzhKUWhlUHMwdmovamtwTnlMWE1jY2Y2MGlIUms4c21BSWNUK3RUNzNYbDNDT0E2TU1EN2VnQmg0aUZkQnlYS2hqbGRjUUNRY2t1S1FVPQp0cnVzdGVkIGNvbW1lbnQ6IHRpbWVzdGFtcDoxNzY0NDI0NTIwCWZpbGU6R2l0QnV0bGVyXzAuMTguMl9hbWQ2NC5BcHBJbWFnZS50YXIuZ3oKbnc2SFNoc2dqcDRReDUzTDZxdEhuWEhxMVUwSy9EQkZYU2ZSNHJjVzRPL2o4TFA0aytRbXR4Z0hOWUhpd2tXUVVaeXE0UG1NVGZNU0xIU2tyd3ZSQ2c9PQo=",
      "url": "https://releases.gitbutler.com/releases/release/0.18.2-2660/linux/x86_64/GitButler_0.18.2_amd64.AppImage.tar.gz"
    }
  }
}
```

That API needs to be updated, along with the website page that shows downloads: https://gitbutler.com/downloads

And potentially also the Tauri updater plugin URLs, I don't really know how those work.

This super easy, barely an inconvenience little thing turned out to be all kinds of inconvenient :)

Byron: Thanks for validating my worst fears 😅.

With this PR not updating dependencies anymore, I think what's left to be done here is to update the backend to deal with this.
It's something I can't do, but @Caleb-T-Owens could. Next week @krlvi is back as well and I'd hope he can double-check the course of action, which is aptly summarised in [your comment on Discord](https://discord.com/channels/1060193121130000425/1445541648686780587/1445858454789947532).

scripts/release.sh
```diff
@@ -227,7 +227,19 @@ else
...
 RELEASE_DIR="$DIST/$OS/$ARCH"
+if [ "$OS" = "linux" ] && [ $(lsb_release -cs) = "noble" ]; then
+  # Our default Linux build is Ubuntu jammy (22.04). We still build for noble
+  # (24.04) but put the builds in a hidden-away part of the S3 bucket. The
+  # primary reason for this build existing is that the noble-built AppImage
+  # tends to work better with newer distros than the jammy-built AppImage.
+  RELEASE_DIR="$DIST/$OS-$(lsb_release -cs)/$ARCH"
+fi
+
+echo "Resolved RELEASE_DIR=$RELEASE_DIR"
```
slarse: @Byron I figured we could do something of a middleground. We make the Ubuntu 22.04 build the "default" Linux build, which will be linked from the website without any changes to any infra.

But we still build the Ubuntu 24.04 (noble) build, and just stash it away under a different S3 key. So we can direct users to "try that AppImage instead" if they report issues. At least it'll be accessible.

Depending on how the S3 access control is configured, there may be a need for some updates there, but the items should be uploaded at least.

Seems like a good middleground, don't you think? Could you try making a nightly and we'll see what happens?

slarse: Lemme know if you feel like "nah, let's just go with the Ubuntu 22.04 build alone", and we'll do that.

Byron:
Alright, let's do it!

Here is the test-publish: https://github.com/gitbutlerapp/gitbutler/actions/runs/20125554514 .

If it works, I think we can merge this PR, but let's please clean it up to avoid the 'master' merge, and have just one commit or so.

Thanks again.

Edit: it looks like the Ubnutu22 build is broken (now?). It's strange as I thought we had that working already.

slarse:
> Edit: it looks like the Ubnutu22 build is broken (now?). It's strange as I thought we had that working already.

Oh bother, libwebkit 2.44 doesn't exist on jammy... When we got it to build we used 2.50 across the board, which is available both on jammy and noble. Oversight on my part. I've updated the script to choose separate dep versions for the different builds. So, ironically, the jammy build will use a newer version of libwebkit even though it builds on an older platform.

The only other choice on jammy was 2.36, which is just too bloody old.

I've also got [this lead from a community member](https://github.com/gitbutlerapp/gitbutler/issues/5282#issuecomment-3639443229) about a potentially more portable AppImage, which I intend to investigate this weekend. But regardless of that we need the jammy build for the deb and rpm packages for the older glibc symbols.

Byron:
Thanks so much, and… please go easy on the weekend work 😅!

Here is the new release: https://github.com/gitbutlerapp/gitbutler/actions/runs/20163867079 - I am curious how the non-appimage bundles will work, given the glibc downgrade and the dependency update.

Caleb-T-Owens:
Hey!

I just wanted to hop in here to go over what went awry with the downloads APIs.

Our API provides a `platform` property (known as `platform_human` internally). When generating this, it first tried to get a "os_human" which it concatenated with another string. This `os_human` expected the `os` to be either "windows", "darwin", or "linux", so the "linux-..." values caused `os_human` to return `nil` and subsequently fail when we tried to concatenate that into a string.

It's our fault for not having any validation on what we expected the `os` value to be. I've made a PR that adds in appropriate validations & gracefully handles the cases when unexpected data is present. I'll be deploying this on Monday.

It does however make me wonder how we should go about labelling these different builds. For our different linux builds in the past they've been part of the file name.

We could make the required changes to the backend & downloads page to support new OS variants.

slarse:
> Hey!
> 
> I just wanted to hop in here to go over what went awry with the downloads APIs.
> 
> Our API provides a `platform` property (known as `platform_human` internally). When generating this, it first tried to get a "os_human" which it concatenated with another string. This `os_human` expected the `os` to be either "windows", "darwin", or "linux", so the "linux-..." values caused `os_human` to return `nil` and subsequently fail when we tried to concatenate that into a string.

Thanks for the lore! My third PR causing some kind of major problem is a new personal record :)

> It's our fault for not having any validation on what we expected the `os` value to be. I've made a PR that adds in appropriate validations & gracefully handles the cases when unexpected data is present. I'll be deploying this on Monday.

I prefer not to talk about fault, it's a precarious thing, even when self-assigned. The context of the "faulty" decision is almost always missing from the equation. We break things, we fix things, we move on. Sorry about the sermon :)

But good on the fix for Monday!

> It does however make me wonder how we should go about labelling these different builds. For our different linux builds in the past they've been part of the file name.

I don't think it's a particularly significant decision as the amount of data is so tiny and can easily be reordered later with a small data migration. But I would personally categorize different Linux distributions as different operating systems. Ubuntu 22.04 and Ubuntu 24.04 are different OSes in much the same sense that Windows 10 and Windows 11 are. Typically that which is built for the predecessor works on the successor (but not always), and what's built on the successor does not work on the predecessor (but sometimes it does).

> We could make the required changes to the backend & downloads page to support new OS variants.

Let's not put any effort into this right now, as the Linux situation is a bit fluid. We're charting out a path that entails solid support for Ubuntu 22.04 and 24.04, but it hasn't entirely materialized yet, so effort should be kept to a minimum. This PR has well and truly spiraled beyond all intentions.

If it's possible to keep pushing both Ubuntu builds to S3, but let the website keep displaying only the `os=linux` build, that'd be ideal. The push to S3 worked as expected, btw, [here is the noble AppImage](http://releases.gitbutler.com/releases/nightly/0.5.1741-2685/linux-noble/x86_64/GitButler_Nightly_0.5.1741_amd64.AppImage.tar.gz), btw.

If your Monday fix makes that possible - let's do that. If it would require additional work, then let's just drop the Ubuntu 24.04 build for now, I think.

Byron:
As this PR approaches a merge, could we rebuild it into a couple of commits on top of `master`, without the `master` merges? As I keep making test releases from here, I'd have to rebase it on top of `master` before doing so and in the past, I couldn't do that (but would instead have had to merge `master` again).

Regarding the Ubuntu-24 build, I think we can easily remove it for now, wait for the backend fixes, and setup another PR that re-adds them, as only then we have a chance to make them show up on the website anyway.

I'd love to merge this long-standing monster :D.

slarse:
> As this PR approaches a merge, could we rebuild it into a couple of commits on top of `master`, without the `master` merges? As I keep making test releases from here, I'd have to rebase it on top of `master` before doing so and in the past, I couldn't do that (but would instead have had to merge `master` again).
> 
> Regarding the Ubuntu-24 build, I think we can easily remove it for now, wait for the backend fixes, and setup another PR that re-adds them, as only then we have a chance to make them show up on the website anyway.
> 
> I'd love to merge this long-standing monster :D.

I'm kind of short on time, so I hope you don't mind my just throwing away the entire commit history here. There are some interesting things in it so I've pushed a backup with the full commit history at https://github.com/slarse/gitbutler/tree/improve-linux-distro-compatibility-backup (I have low trust for GitHub keeping track for me :)).

As I sign commits, rewriting this long a commit history is kind of a pain, and something that GitButler doesn't really help with at the moment.

Byron:
Thanks a lot - one commit is perfect for me to handle!

I just rebased the PR on top of `master` and [kicked off a new publish](https://github.com/gitbutlerapp/gitbutler/actions/runs/20240941001) for a nightly, which hopefully is ready within 30m or so.

Once you give it your thumbs up, we can certainly merge this PR alongside #11534.

slarse:
@Byron Don't tell Kiril, but this was effectively a squash with more steps (I do sincerely find squashing useful at times).

Anyhow, build seems to work fine for me on my main machine (Arch Linux + nvidia) as well as on my Ubuntu 22.04 machine. Tested the AppImage and deb package. I expect the https://github.com/gitbutlerapp/gitbutler/issues/5282 to show up for folks running Intel graphics on newer distros on some hardware configurations.

Byron:
And it's done!

Thanks so much for your tremendous help with Linux, it's indispensable.

slarse:
> And it's done!
> 
> Thanks so much for your tremendous help with Linux, it's indispensable.

Right back at you!

PR #11433, opened 2025-12-02
Make `but_api::legacy::diff::commit_details()` non-legacy with macro-support.

Byron:
This PR will add the machinery that's needed to be able to port legacy functions that take `ProjectId` into the modern world, so they take `Context` instead.
The main difficulty I see is to make `api_cmd` deal with the conversion of `tauri` and the server, down to either `&Context`, or `&mut Context`.

Let's see what kind of simplifications happen on the way, as `Context` is also doable, at the expense of automated caching,
i.e. re-use of the context for a particular project which would require the `Context` to be passed as reference.

### Tasks

* [x] manual port of `commit_details()`
    - [x] make sure line stats aren't computed unnecessarily
    - [x] move CommitDetails into `but-core`
* [x] `api_cmd` update to be able to generate the tauri and server versions.
    - This will get us the 'cmd' version, the one that takes the params struct for the but-server implementation
* [x] let UI use the new version of the function, remove the old one and remove the `_v2` suffix.
* [x] address PR review
    - [x] auto ident rename when mapping
    - [x] update but_api docs
    - [x] double-check `try_from` error handling


crates/but-api/src/diff.rs
```diff
@@ -0,0 +1,75 @@
...
+/// This function just exists to make the frontend work with different parameter names, and without
+/// the need for line-stats to be enabled explicitly.
+// TODO: the frontend shouldn't have to care or be able to remap these names internally.
+#[but_api_macros::but_api(json::CommitDetails)]
+#[instrument(err(Debug))]
+pub fn commit_details_with_line_stats(
+    project_id: &Context,
+    commit_id: gix::ObjectId,
+) -> anyhow::Result<CommitDetails> {
+    commit_details(project_id, commit_id, ComputeLineStats::Yes)
```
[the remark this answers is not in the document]
Byron: Intentional

crates/but-api-macros/src/lib.rs
```diff
@@ -19,202 +20,439 @@ pub fn api_cmd_tauri(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+                    // We control these conversions, and must just make them work to keep this simple.
+                    param_conversions.push(match from_mode {
+                        FromMode::From => {
+                            quote! {
+                                let mut #name = <#ty>::from(#name);
+                            }
+                        }
+                        FromMode::TryFrom => {
+                            quote! {
+                                let mut #name = <#ty>::try_from(#name)?;
```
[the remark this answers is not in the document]
Byron: It's a pre-emptive `mut`, intended. Without it, `&mut` wouldn't be supported.

crates/but/src/command/legacy/status/mod.rs
```diff
@@ -1,8 +1,11 @@
 use std::collections::BTreeMap;
 
+use crate::CLI_DATE;
 use assignment::FileAssignment;
 use bstr::{BString, ByteSlice};
-use but_core::ui::{TreeChange, TreeStatus};
+use but_api::diff::ComputeLineStats;
+use but_core::ui::TreeChange;
+use but_core::{TreeStatus, ui};
```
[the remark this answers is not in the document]
Byron: Great catch.

crates/but-api/src/diff.rs
```diff
@@ -0,0 +1,74 @@
...
+    CommitDetails::from_commit_id(commit_id.attach(&repo), line_stats.into())
+}
+
+/// This function just exists to make the frontend work with different parameter names, and without
+/// the need for line-stats to be enabled explicitly.
+// TODO: the frontend shouldn't have to care or be able to remap these names internally.
+#[but_api(json::CommitDetails)]
+#[instrument(err(Debug))]
+pub fn commit_details_with_line_stats(
+    project_id: &Context,
```
krlvi: Why the naming discrepancy?

krlvi: Oh I know why - in order to not change the frontend. And also to not leak `ctx` as a name into the frontend.

crates/but-api-macros/src/lib.rs
```diff
@@ -9,7 +10,7 @@ use syn::{Expr, FnArg, ItemFn, Pat, parse_macro_input};
 /// * `func_cmd` for calls from the frontend, taking `serde_json::Value` and returning `Result<serde_json::Value, Error>`
 /// * `func_tauri` for calls from the tauri, args and returning `Result<serde_json::Value, Error>`, with `tauri` support.
 #[proc_macro_attribute]
-pub fn api_cmd_tauri(attr: TokenStream, item: TokenStream) -> TokenStream {
+pub fn but_api(attr: TokenStream, item: TokenStream) -> TokenStream {
```
krlvi, line 16: Perhaps worth adding something to the documentation that as far as callers of the JSON API are concerned, `Context` is equivalent to `ProjectId`. It's really cool that we can do this in a macro but it is potentially a point of confusion when implementing commands

crates/but-ctx/src/legacy.rs
```diff
@@ -63,6 +63,15 @@ impl Context {
     }
 }
 
+impl TryFrom<LegacyProjectId> for Context {
+    type Error = anyhow::Error;
+
+    fn try_from(value: LegacyProjectId) -> Result<Self, Self::Error> {
+        let project = gitbutler_project::get(value)?;
+        Context::new_from_legacy_project(project)
```
krlvi, line 71: Since this gets executed from within a macro... in theory error could be extra confusing. I wonder, perhaps this is a place we could be helpful when debugging if we included some additional error info with the anyhow's `.context`? What do you think?

krlvi (approved):
I left some reflections/ comments. Curious about the E2E test failure but other than that looks good

crates/but-api/src/diff.rs
Byron: Actually, we will always have that problem and last night I was just in a hurry.
Fresh in the morning I think this renaming must be done in the macro, that's what it is there for :).

This function is still needed so the frontend can always get the line stats, but at least it's very explicit now.

crates/but-api-macros/src/lib.rs
Byron: Yes, and this is a good reminder that the docs should probably be updated.
Besides that, I think the macro will for some time do a considerable amount of magic that is quite 'unspecified' when looking at just the name, but maybe that generic name is good from that point of view: It can do anything we need it to do to serve the `but-api`.

crates/but-ctx/src/legacy.rs
Byron: I assumed that the calls themselves will provide useful error information, but I will double-check. My expectation is that this typically won't be the reason a call fails.

Byron:
Thanks a lot for the review @krlvi, and I am glad codeowners was there. You see I already made it a rule to use Copilot by default, and if there is no rush I think I will try and not auto-merge things, particularly the bigger ones to get more eyes on it.

Byron:
For the E2E failure, the crucial lesson learned here that one must not only search for `'commit_details'`, but also for `"commit_details"` as playwright uses the `but-server` as backend. Makes perfect sense as well, and I hope to remember it for a while now.

crates/but-ctx/src/legacy.rs
Byron: I checked and each of these two calls provide specific contexts already, so errors here should be clear and identify the underlying problem. Thus I think it's OK.

crates/but-api/src/diff.rs
krlvi: @Byron One challenge is that for a person implementing an endpoint in JavaScript, they can't look at the method signature anymore and understand what they need to pass...
I don't know how to solve this though.

Byron: Yes, that's true.

It's a crutch and it comes at a cost, but I can see the light… *<after fending off the heart attack by pounding the chest a couple of times, the light disappears and I continue writing>* and … actually there will always be this difference between the UI passing `projectId` or `projectHandle`, which has to get transformed into `ctx`.

To my mind the problem is mostly that the frontend has to rely on parameter names.
Maybe this is just something we document and live with? Maybe it's something to change to `projectHandle` or something shorter as well, or maybe `projectId` is fine even if the type eventually becomes the self-describing `ProjectHandle`.

In the end, I also don't see any other solution but to document that clearly, for those frontend implementations that rely on parameter names at least. Dioxus for example wouldn't have that problem at all and calls the Rust API directly.

krlvi: I think documenting and living with it sounds like the way to go here

Byron: Alright! I double-checked the docs and here is the result: https://github.com/gitbutlerapp/gitbutler/pull/11460 .

kritsada1998mt-create:
[]()

Issue #11437, opened 2025-12-03
but rub command has critical usability issues with path resolution and CLI ID handling


et84121:
### Version

0.18.2

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

## Environment

- **GitButler CLI version**: 0.18.2
- **Git version**: 2.48.1
- **OS**: macOS (Darwin 25.1.0)

---

## Summary

The `but rub` command has several critical issues that make it difficult or impossible to assign files to branches in common scenarios. These issues compound each other, leaving users with no viable workaround when CLI IDs collide or paths contain hyphens.

---

### Description
File paths containing hyphens are incorrectly parsed as range syntax (`start-end`), making it impossible to assign such files.

### Reproduction
```bash
but rub apps/web-app/src/types/my-component.d.ts feature/my-branch
```

### Error
```
Error: Range format should be 'start-end', got 'apps/web-app/src/types/my-component.d.ts'
```

### Impact
Extremely common in modern web projects where hyphenated filenames are standard (e.g., `vue-carousel.d.ts`, `user-profile.ts`, `api-client.js`).

---

### Description
When multiple files receive the same CLI ID, the error message suggests disambiguation methods, but **none of them work**.

### Reproduction

1. Observe CLI ID collision in status:
```
╭┄00 [Unassigned Changes]
┊   ab D apps/web-app/src/types/User.ts
┊   cd D apps/web-app/src/types/Auth.ts
┊   ab M apps/web-app/src/components/UserCard.vue
```

2. Attempt to use the conflicting CLI ID:
```bash
but rub ab feature/my-branch
```

3. Receive error with suggestions:
```
Error: Source 'ab' is ambiguous. Matches: ab (an uncommitted file), ab (an uncommitted file).
Try using more characters, a longer SHA, or the full branch name to disambiguate.
```

4. Try **all** suggested methods - all fail:

| Suggested Method | Command Attempted | Result |
|-----------------|-------------------|--------|
| "more characters" | `but rub types/User feature/my-branch` | `Source 'types/User' not found` |
| filename | `but rub User.ts feature/my-branch` | `Source 'User.ts' not found` |
| partial path | `but rub src/types/User feature/my-branch` | `Source 'src/types/User' not found` |
| quoted path | `but rub 'User.ts' feature/my-branch` | `Source 'User.ts' not found` |

### Root Causes
1. CLI ID generation algorithm creates duplicates
[...]

PR #11445, opened 2025-12-03
Unambiguous committed files CLI IDs

jonathantanmy2:
Sample output (copied from a commit message):
```
╭┄000 [Unassigned Changes]
┊
┊╭┄ux [b00] (no commits)
├╯
┊
┊╭┄wo [work]
┊●   89f03ba test commit
┊│     g0 A a
┊│     h0 A b
┊│     j0 A c
├╯
┊
┊╭┄i0 [i0] (no commits)
├╯
┊
┴ dba337d (common base) [origin/main] 2025-11-11 Merge pull request #2254 from ralphmodales/credent
```

Helps one of several issues in #11437

If any Rust expert out there can think of how to improve my code, feel free to leave suggestions. I struggled particularly with how the Vec binary search returns indexes instead of references to the found element (necessitating an unchecked `v[index]` when `index` is known to be safe because it was returned by binary search) and how I cannot declare a Vec to only contain one specific variant of an enum.

### DIFF
--- a/crates/but/src/command/legacy/rub/commits.rs
+++ b/crates/but/src/command/legacy/rub/commits.rs
@@ -1,7 +1,7 @@
 use std::collections::HashSet;
 
 use anyhow::{Context as _, Result};
-use bstr::ByteSlice;
+use bstr::BStr;
 use but_core::{DiffSpec, diff::tree_changes};
 use but_ctx::Context;
 use but_hunk_assignment::HunkAssignmentRequest;
@@ -15,7 +15,7 @@ use crate::{
 
 pub fn commited_file_to_another_commit(
     ctx: &mut Context,
-    path: &str,
+    path: &BStr,
     source_id: gix::ObjectId,
     target_id: gix::ObjectId,
     out: &mut OutputChannel,
@@ -30,7 +30,7 @@ pub fn commited_file_to_another_commit(
     let tree_changes = tree_changes(&repo, Some(source_commit_parent_id.detach()), source_id)?;
     let relevant_changes = tree_changes
         .into_iter()
-        .filter(|tc| tc.path.to_str_lossy() == path)
+        .filter(|tc| tc.path == path)
         .map(Into::into)
         .collect::<Vec<DiffSpec>>();
 
@@ -56,7 +56,7 @@ pub fn commited_file_to_another_commit(
 
 pub fn uncommit_file(
     ctx: &mut Context,
-    path: &str,
+    path: &BStr,
     source_id: gix::ObjectId,
     target_branch: Option<&str>,
     out: &mut OutputChannel,
@@ -72,7 +72,7 @@ pub fn uncommit_file(
         let tree_changes = tree_changes(&repo, Some(source_commit_parent_id.detach()), source_id)?;
         tree_changes
             .into_iter()
-            .filter(|tc| tc.path.to_str_lossy() == path)
+            .filter(|tc| tc.path == path)
             .map(Into::into)
             .collect::<Vec<DiffSpec>>()
     };
--- a/crates/but/src/command/legacy/rub/mod.rs
+++ b/crates/but/src/command/legacy/rub/mod.rs
@@ -14,7 +14,10 @@ use gitbutler_oplog::{
     entry::{OperationKind, SnapshotDetails},
 };
 
-use crate::{legacy::id::CliId, utils::OutputChannel};
+use crate::{
+    legacy::id::{CliId, IdDb},
+    utils::OutputChannel,
+};
 
 pub(crate) fn handle(
     project: &Project,
@@ -23,7 +26,8 @@ pub(crate) fn handle(
     target_str: &str,
 ) -> anyhow::Result<()> {
     let ctx = &mut Context::new_from_legacy_project(project.clone())?;
-    let (sources, target) = ids(ctx, source_str, target_str)?;
+    let id_db = IdDb::new(ctx)?;
+    let (sources, target) = ids(ctx, &id_db, source_str, target_str)?;
 
     for source in sources {
         match (&source, &target) {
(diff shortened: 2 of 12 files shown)


jonathantanmy2:
Going back to how I didn't like the Vec binary search API, I added a new commit "id: use BTreeSet instead of sorted Vec". The BTreeSet API has its own weaknesses, but I prefer it over the Vec API. I'm happy to submit the PR with or without this commit.

crates/but/src/command/legacy/rub/commits.rs
```diff
@@ -72,7 +72,7 @@ pub fn uncommit_file(
         let tree_changes = tree_changes(&repo, Some(source_commit_parent_id.detach()), source_id)?;
         tree_changes
             .into_iter()
-            .filter(|tc| tc.path.to_str_lossy() == path)
+            .filter(|tc| tc.path == path)
```
Byron, line 75: Wonderful!

crates/but/src/command/legacy/rub/mod.rs
```diff
@@ -131,9 +156,14 @@ fn makes_no_sense_error(source: &CliId, target: &CliId) -> String {
...
-    let sources = parse_sources(ctx, source)?;
-    let target_result = crate::legacy::id::CliId::from_str(ctx, target)?;
+fn ids(
+    ctx: &mut Context,
+    id_db: &IdDb,
+    source: &str,
+    target: &str,
+) -> anyhow::Result<(Vec<CliId>, CliId)> {
+    let sources = parse_sources(ctx, id_db, source)?;
+    let target_result = id_db.parse_str(ctx, target)?;
```
Byron, line 166: \<rambling>
I always found it strange (out of ignorance) why `IdDb` wouldn't initialise itself with all desired context upon instantiation. Why does it do any work when getting IDs from some entity? After all, how would it know it generates unique CLIIds if it doesn't know all the 'to-be-accessed' context in advance?

Byron (approved):
Thanks so much! Let's merge this, as it's a clear improvement.

I put some minor refactors on top with guides on future performance improvements, as this type inevitable marches towards non-legacy.
It's great that it now starts using the `RefInfo` structure.
Regarding Vec + binary search vs BTreeSet - to me it seems having a more convenient API is preferable here as well.

Please note that I didn't look at all at the algorithm, but it's something I'd certainly love consuming by reading its tests once they exist on unit level. Generally, I think this type must amp up its testing, and I'd love to help with that a well if there are impediments.

Beyond that, here are a couple of thoughts just for sharing them with you, nothing actionable.

Note: I seem to keep writing `IdMap` when I mean to say `IdDb`. I also think it's a better name.

#### Regarding bd806f6ff86ff1fe6d77ce19762e4e85a60ff428

The general invariant seems to be that
1. once the 'context' (i.e. everything contributing to CLI ID generation) is specified, only then we can query the CLI IDs, or get a mapping of any sort
1. if the CLI ID mapping is recreated on another invocation, ideally IDs remain as stable as possible even if new entities are added (i.e. a new branch was inserted).

Regarding 1), I think this can be enforced by something like `let map = IdMap::build_context().add_foo(…).add_bar(…).into_mapping(); map.query(…)`. That way, the query of IDs can't be done until all context is specified.
Maybe this is not how code is using the IdMap today, but we might as well have a new version of it that eventually replaces all other uses.

Also… now that I looked at the code I realise that right now it's instantiated from `ctx` anyway, so it's expected to initialise itself from *everything*. Maybe one day this can be limited to only parts of the workspace, to safe time.

Further, I think that `IdMap::new()` should take a `RefInfo` reference, so that we can reuse it. For instance, `but status` will need one to visualise, but it's also used for `IdMap`, no need to compute it twice. `IdMap` should be plumbing, hence it takes as input only what it needs.

This is inconvenient to do in this PR as `new()` is used in a lot of places, but as the `but status` refactor progresses, I think there should be another constructor at least to phase out `new`.
Alternatively, `new()` could stay but make the `RefInfo` accessible afterwards.

crates/but/src/command/legacy/rub/mod.rs
jonathantanmy2: > After all, how would it know it generates unique CLIIds if it doesn't know all the 'to-be-accessed' context in advance?

It doesn't generate unique `CliId`s - that's why I have been trying to centralize generation of IDs.

jonathantanmy2:
> Regarding 1), I think this can be enforced by something like `let map = IdMap::build_context().add_foo(…).add_bar(…).into_mapping(); map.query(…)`. That way, the query of IDs can't be done until all context is specified. Maybe this is not how code is using the IdMap today, but we might as well have a new version of it that eventually replaces all other uses.

I'll do this if I can't do the below...

> Also… now that I looked at the code I realise that right now it's instantiated from `ctx` anyway, so it's expected to initialise itself from _everything_. Maybe one day this can be limited to only parts of the workspace, to safe time.
> 
> Further, I think that `IdMap::new()` should take a `RefInfo` reference, so that we can reuse it. For instance, `but status` will need one to visualise, but it's also used for `IdMap`, no need to compute it twice. `IdMap` should be plumbing, hence it takes as input only what it needs.
> 
> This is inconvenient to do in this PR as `new()` is used in a lot of places, but as the `but status` refactor progresses, I think there should be another constructor at least to phase out `new`. Alternatively, `new()` could stay but make the `RefInfo` accessible afterwards.

I think that if we can make a stub `RefInfo`, we can pass it to the `IdMap` constructor and then we'll be able to unit test it. (Then we won't need the `build_context` you describe above.) Also, I'll change the name to `IdMap` when I change the constructor to take `RefInto` as callers would then only need to change once.

Byron:
Sounds like a plan!

> I think that if we can make a stub `RefInfo`, we can pass it to the `IdMap` constructor and then we'll be able to unit test it. (Then we won't need the `build_context` you describe above.) Also, I'll change the name to `IdMap` when I change the constructor to take `RefInto` as callers would then only need to change once.

I just wanted to add that we'd still have to create a real repository to avoid having to make up an entire `RefInfo` struct.

As a side-note, I also noticed that it contains `SegmentIndex` fields which refer to a `Graph` that it doesn't contain anymore, so it's not self-contained and it's something that could be addressed.

PR #11446, opened 2025-12-04
id: remove dependence on legacy code for commits

jonathantanmy2:
Follow up on #11445, and much smaller in scope.

(copy and paste from commit description)

Also, simplified commit ID parsing in `IdDb`. It used to be done before uncommitted files, committed files, and branches if the commit ID given was at least 4 characters long (with another redundant parse after), and after uncommitted files, committed files, and branches otherwise. But it does not matter whether it's done before or after, since commit IDs never collide with any of those (which all start with g-z), so do it before in all cases.

In the absence of a way to unit test `id.rs`, I've added integration tests once again. Once we migrate `id.rs` completely off legacy code, we may be able to narrow down what exactly is needed (probably not the whole `Context`) and thus be able to make unit tests.


Byron:
Thanks a lot @jonathantanmy2! 

> In the absence of a way to unit test id.rs, I've added integration tests once again. Once we migrate id.rs completely off legacy code, we may be able to narrow down what exactly is needed (probably not the whole Context) and thus be able to make unit tests.

I truly believe this is possible, and think I will setup an example to replace the integration test with a more unit'y test.
And now that this is there I think improving `IdDb` will be much easier as well - sorry for letting you wait so long.

Byron:
Let's merge this as another step towards a great `IdMap` implementation.

Issue #9946, opened 2025-08-22
Feature request: Image comparison tool


HermannBjorgvin:
Being able to compare an image diff would be useful for comparing before/after snapshots from testing frameworks like cypress or playwright.

Now it just says binary file but it would be nice to be able to do a side by side comparison.

Byron:
Thanks for sharing, I think so, too!

I also think we could start by showing images even in non-diff situations, along with other binary formats that are easily displayable in a browser engine.

HermannBjorgvin:
Closed in #11456 released in `0.18.3`

Thanks!!!

PR #11456, opened 2025-12-04
feat: add ImageDiff component to load and show before/after images

nshcr:
This PR adds before-and-after visual diff previews for image files in the diff view.

Most of the code was generated with the help of Claude Sonnet 4.5, and was reviewed by me before submitting the PR to ensure code quality and correct functionality.

It covers the following scenarios:

1. Adding

2. Modifying

3. Renaming

4. Deleting

5. Displayed side-by-side when the window width is sufficient


apps/desktop/src/components/UnifiedDiffView.svelte
```diff
@@ -86,6 +87,23 @@
...
+			'.jpg',
+			'.jpeg',
+			'.gif',
+			'.webp',
+			'.bmp',
+			'.ico',
+			'.heic',
+			'.heif',
+			'.avif'
+		];
```
nshcr: For some image formats such as `heic`, `heif`, and `avif`, they may not render correctly on systems with older OS versions.
Maybe should consider dropping support for them.

BTW, `svg` files could also be displayed as image diffs.
However, they are currently treated as text, and I didn't add a special-case handling for them.

crates/but-api/src/legacy/repo.rs
nshcr, line 1: I avoided using `project.open_git2()` as I saw it is marked deprecated, and used the gix API instead.
I'm not entirely sure if this is the correct approach here.

nshcr:
@Byron — could you please help review the Rust changes? I added a new function to `but-api` and modified an existing one.

@PavelLaptev — could you please help review the frontend changes and adjust the styling so that the UI better matches GitButler's design?

Thanks in advance for taking the time to review this and for any feedback or improvements you can suggest!

nshcr:
Oh, this PR could fix https://github.com/gitbutlerapp/gitbutler/issues/9946.

crates/but-api/src/legacy/repo.rs
Byron: I think it's a good call. The `repo` needs to be opened anyway, and then using `gix` is faster.
Since it's all in legacy-land, I think it's fine to do what's needed.

However, I don't think that anything but full hashes may be supported here. The frontend already has blob hashes for anything one would want to diff.

Also I think there is a way to read an arbitrary file from within the worktree, in case of files that want comparison between worktree/Git.

Byron (commented):
Thanks a lot for getting this started, that's such a useful feature!

I just went ahead and made the changes I'd need to see with some explanation in the commit message. Probably I don't understand where the need for arbitrary revspecs is coming from.
With that it looks good on the Rust side.

Something I could use your help with is to understand how your are diffing worktree/Git changes. ~~I also vaguely remember the presence of a function to obtain a worktree file, but I can't find it now.~~ Ah, right, it's `readFromWorkspace` and it's already used.

Byron:
Oh, and allow me unleash Copilot on this PR as I actually value what it has to say, most of the time 😅. Please feel free to ignore entirely.

crates/but-api/src/legacy/repo.rs
```diff
@@ -84,6 +92,24 @@ pub fn get_workspace_file(project_id: ProjectId, relative_path: PathBuf) -> Resu
     project.read_file_from_workspace(&relative_path)
 }
 
+#[but_api]
+#[instrument(err(Debug))]
+pub fn get_blob_file(
+    project_id: ProjectId,
+    relative_path: PathBuf,
+    blob_id: String,
+) -> Result<FileInfo> {
```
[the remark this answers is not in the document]
Byron: A valid point - this will help when we eventually port it to non-legacy, which would already be possible actually.

PavelLaptev:
@Byron @nshcr, I’ve updated the diff component.

1. Moved the component to the UI package.
2. Added “swipe” and “onion modes.” Should we merge this?
3. Added metadata like image dimensions and file size.
4. Shown the difference in how smaller or larger the new file is.
5. Made some style updates.

Here’s the updated component:

https://github.com/user-attachments/assets/6829cde5-13c4-4c8a-92bf-19452d6053ed

---

Two things we can do in a follow-up PRs:
- Treat SVGs as images but also allow switching to the code preview (sometimes it matters).
- Commit generation doesn’t see image files (probably all binaries, see image below).

image

---

Should we merge it?

nshcr:
@PavelLaptev Sorry I haven't had time to handle this PR over the past few days, and thank you so much for the updates you've made!

If Byron thinks the Rust changes look good and are ready to be merged, I'm completely fine with that as well — honestly, I don't really have a clear idea of what should be changed on the Rust side...

As for the frontend changes — they're fantastic! Far beyond what I originally expected for this feature.

The Swipe and Onion Skin modes are incredibly useful. I was thinking: for Swipe mode, would it be more intuitive to place the drag handle in the center of the image and use a semi-transparent center line to separate the before / after images? (similar to [Squoosh](https://squoosh.app/))

Also, I think replacing the "before / after" text with "deleted / added" would be more appropriate and consistent with the labels used by GitHub. Looking back, using "before / after" was probably not the best choice on my part.

Byron:
This looks fantastic now, amazing work!

Regarding Rust, I have rebased my refactor so now I'd think we are good on the Rust side.

CI seems to fail genuinely, but I wouldn't know how to fix it. But besides that, I think it's good to be merged.

> Commit generation doesn’t see image files (probably all binaries, see image below).

Let's debug this in a follow-up. It's probably something about the backend side being made with text files in mind, somewhere, but I didn't try to reproduce it yet.

PavelLaptev:
> The Swipe and Onion Skin modes are incredibly useful. I was thinking: for Swipe mode, would it be more intuitive to place the drag handle in the center of the image and use a semi-transparent center line to separate the before / after images? (similar to [Squoosh](https://squoosh.app/))

Good idea. Done

https://github.com/user-attachments/assets/f999fdb9-09a5-4d0e-b3d8-0885b56255b7

PavelLaptev:
> Also, I think replacing the "before / after" text with "deleted / added" would be more appropriate and consistent with the labels used by GitHub. Looking back, using "before / after" was probably not the best choice on my part.

"After"/"Before" works well when comparing two images. It clearly shows what changed when an image was replaced.
However, it's confusing when there's only one image.

"Removed"/"Added" is more descriptive for single images, but less intuitive than "After"/"Before" for replacements.
Use "After"/"Before" when an image is replaced, and "Removed"/"Added" when only one image exists, typically "Added" (since "Removed" implies there was a previous image to compare against).


https://github.com/user-attachments/assets/682e22f5-64f9-4732-9a1b-882ca33c5e13

PR #11465, opened 2025-12-05
Fix local relative upstream path

mathijs81:
## 🧢 Changes

Fix the active directory to be the repo root instead of `.git`

## ☕️ Reasoning

I used the following script to set up a quick test repo + local remote
```
#!/bin/bash
mkdir remote
cd remote
git init --bare
cd ..
mkdir local
cd local
git init
touch first_file.txt
git add first_file.txt
git commit -m "first commit"
git remote add origin ../remote
git push origin master

echo "local repo + local remote set up"
```

When loading `local` into GitButler there are errors that remote isn't a git repo because it's running git from within the `.git` dir and then the relative `../remote` doesn't exist.

I don't know if it's possible to use this setup script as the start of an integration test? If you think that's a good idea then I'd love a small pointer of where to add that.

### DIFF
--- a/crates/gitbutler-git/src/repository.rs
+++ b/crates/gitbutler-git/src/repository.rs
@@ -222,7 +222,13 @@ where
     );
 
     let cwd = match harness_env {
-        HarnessEnv::Repo(p) | HarnessEnv::Global(p) => p,
+        HarnessEnv::Repo(p) =>
+        // repo_path is the .git/ dir in the repository.
+        // Take the parent dir, so that e.g. an origin that's a relative path works.
+        {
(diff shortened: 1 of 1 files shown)


Byron:
This is a bug I introduced. I was under the assumption that Git sets its own workdir accordingly, but that apparently isn't the case.

Byron:
Thanks for the nudge! I have taken the fix upon myself in https://github.com/gitbutlerapp/gitbutler/pull/11469 and think this PR can be closed.

I'd love it if you could validate it with tomorrow's nightly maybe, in case I forget.
Thanks again.

PR #11473, opened 2025-12-05
Reimplement insert blank commit

Caleb-T-Owens:
Implements insert blank commit & has some refactors to the rebase engine (see commit message).

Insert blank commit hasn't been added to the but-api or hooked into the frontend

---
- <kbd>&nbsp;3&nbsp;</kbd> #11473 👈 
- <kbd>&nbsp;2&nbsp;</kbd> #11452 
- <kbd>&nbsp;1&nbsp;</kbd> #11195

### DIFF
--- a/crates/but-workspace/tests/fixtures/scenario/insert-blank-commit-three-commits copy.sh
+++ b/crates/but-workspace/tests/fixtures/scenario/insert-blank-commit-three-commits copy.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
+
+echo "/remote/" > .gitignore
+mkdir remote
+pushd remote
+git init
+popd
+
+git init
+
+git remote add origin ./remote
+
+git checkout -b one
+echo "foo" > one.txt
+git add .
+git commit -m "commit one"
+
+git checkout -b two
+echo "foo" > two.txt
+git add .
+git commit -m "commit two"
+
+git push -u origin two
+
+git checkout -b three
+echo "foo" > three.txt
+git add .
+git commit -m "commit three"
--- a/crates/but-workspace/tests/workspace/commit/insert_blank_commit.rs
+++ b/crates/but-workspace/tests/workspace/commit/insert_blank_commit.rs
@@ -1,0 +1,142 @@
+use anyhow::Result;
+use but_rebase::graph_rebase::GraphExt as _;
+use but_rebase::graph_rebase::mutate::InsertSide;
+use but_testsupport::{assure_stable_env, visualize_commit_graph_all};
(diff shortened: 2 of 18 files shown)


crates/but-workspace/tests/fixtures/scenario/insert-blank-commit-three-commits copy.sh
```diff
@@ -0,0 +1,30 @@
...
+pushd remote
+git init
+popd
+
+git init
+
+git remote add origin ./remote
+
+git checkout -b one
+echo "foo" > one.txt
```
Byron, line 16: How is files relevant as initial state to adding blank commits.
Maybe it's copied from a very old fixture from somewhere? I really thought we could do more with less by now, using function in `shared.sh`.

crates/but-workspace/tests/fixtures/scenario/insert-blank-commit-three-commits copy.sh
```diff
@@ -0,0 +1,30 @@
...
+
+echo "/remote/" > .gitignore
+mkdir remote
+pushd remote
+git init
+popd
+
+git init
+
+git remote add origin ./remote
```
Byron, line 13: I don't understand why there is a remote at all, it shouldn't be relevant.
If this is not copy-paste, I think it should be explained.

crates/but-workspace/tests/workspace/commit/insert_blank_commit.rs
```diff
@@ -0,0 +1,142 @@
...
+use but_rebase::graph_rebase::mutate::InsertSide;
+use but_testsupport::{assure_stable_env, visualize_commit_graph_all};
+use but_workspace::commit::insert_blank_commit;
+use but_workspace::commit::insert_blank_commit::RelativeTo;
+
+use crate::ref_info::with_workspace_commit::utils::named_writable_scenario_with_description_and_graph as writable_scenario;
+
+#[test]
+fn insert_below_commit() -> Result<()> {
+    assure_stable_env();
```
Byron, line 12: I will try to make clear (in another PR) that this function should be avoided, and will see what's needed to make it happen.
After all, modern code should be test-friendly to not rely on environment variables.
It seems it's mostly related to controlling the `GITBUTLER_CHANGE_ID`.

Byron (commented):
I was quite surprised to see `insert_blank_commit` testing the `InsertSide`, as the implementation appears to forward to the editor. Hence, I would expect these tests and more in `but-rebase`.
With that said, if there are tests in `but-workspace`, I'd also expect to see the workspace projection, before and after, which might lead to other interesting test cases.

The same comment could fit for the reword implementation it seems, I didn't take a look at it before.

PR #11476, opened 2025-12-06
Make `but.exe` as copied from `gitbutler-tauri.exe` print output on Windows

Byron:
Problem: as GUI application its output isn't initialised on Windows like one would expect for a terminal application, which is why it basically doesn't print anything to stdout.

The fix is to initialise the standard channels explicitly using the equivalent of

```c
void init_console()
{
    if (AllocConsole()) {
        freopen("CONOUT$", "w", stdout);
        freopen("CONOUT$", "w", stderr);
        freopen("CONIN$",  "r", stdin);
    }
}
```

### Tasks

* [x] research
* [x] ~~impl~~ - just doesn't work
* [ ] validate impl on Windows

### Related

* [ ] #11477

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3958,6 +3958,7 @@ dependencies = [
  "gitbutler-repo-actions",
  "gitbutler-watcher",
  "gix",
+ "libc",
  "log",
  "open",
  "parking_lot",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3958,7 +3958,6 @@ dependencies = [
  "gitbutler-repo-actions",
  "gitbutler-watcher",
  "gix",
- "libc",
  "log",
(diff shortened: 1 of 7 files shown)


Byron:
I am aborting this attempt as it requires special knowhow on Windows APIs that todays LLMs also don't have.

The only alternative I see is to ship `but.exe` separately there.
CC @Caleb-T-Owens .

Merging this PR anyway as what's left of it reduces the amount of compile warnings on Windows.

Byron:
Regarding the creation flags: on Windows this trait was marked as not needed for some reason, and apparently it does build fine everywhere.
Let me kick off a nightly to make sure. https://github.com/gitbutlerapp/gitbutler/actions/runs/19984115171

Issue #8293, opened 2025-04-28
Improve error handling during repository cloning


krlvi:
Currently we render [object Object]

![image](https://github.com/user-attachments/assets/029dd994-f391-4da8-ae3b-f2e647968d8c)

alexmadeathing:
+1 for this. Cloning into a folder which doesn't exist yields this.

PR #11479, opened 2025-12-06
fix(ui): show parsed error message in credential checks

imadx:
## 🧢 Changes

Import parseError and use it to display a user-friendly error message when awaiting credential check promises. 
This change ensures clearer, consistent error text in the UI.

## ☕️ Reasoning

Currently, the git credential check error is showing as `[object Object]`. This resolves that.


Byron:

This looks good to me, but let's wait for another set of eyes to be sure.

PS: there is another related issue: https://github.com/gitbutlerapp/gitbutler/issues/8293 , but it doesn't look like we have one for the one this PR is addressing.

Byron (approved):

krlvi (approved):
LGTM

PR #11480, opened 2025-12-06
feature: Allow commit title change on click

mathijs81:
## 🧢 Changes

Clicking the commit details title opens the edit dialog.

## ☕️ Reasoning

Currently, branch names can be inline-edited, but commit details not.
The UX is still a bit different as the implementing svelte classes are totally different (the branch name is rendered as an edit box, but for the commit details we have to pop up the little edit pane).

I chose to make the cursor a caret/text-pointer on hover to mimic the same look as with the branch editing, but I think `cursor: pointer` would also be good as it is not inline editing after all.

See video for before and after:

https://github.com/user-attachments/assets/000cfe79-b91b-42d7-bce2-96038fb590fb

### DIFF
--- a/apps/desktop/src/components/CommitTitle.svelte
+++ b/apps/desktop/src/components/CommitTitle.svelte
@@ -7,9 +7,10 @@
 		commitMessage: string;
 		className?: string;
 		editable?: boolean;
+		onclick?: () => void;
 	};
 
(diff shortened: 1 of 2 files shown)


PavelLaptev:
Thank you for splitting this up :-) I’ll try to check it this week.

PR #11495, opened 2025-12-09
Narrower `IdMap` constructor arguments, and unit tests

jonathantanmy2:
I've narrowed the constructor arguments, which made it possible to add unit tests that are even further from integration tests (so, no `Context` in a sandbox), generating input arguments directly.

I'm not sure if this is the direction we want to go in, but I'm leaning towards it. My main concerns are:
* This requires structs to have all public members. I think that having all public members is a good idea anyway (these are app-wide concepts so I prefer the data-oriented programming approach over OOP's encapsulation - if the nature of a `Stack` changes, then all code that uses it needs to be aware of the change anyway, so you might as well remove the unnecessary layer of indirection, but if the specifics of a database changes, then perhaps it can be isolated to its class, so OOP's encapsulation is good there). But if we move towards structs not having all public members, then we will need to undo my PR's approach.
* This requires custom setup code instead of reusing production code (`Context`) to set up the test. I think the custom setup code is not too much, though, when debugging `IdMap`, it's very useful to not have to go through `Context` to see where each piece of data ends up. 

If we continue in this direction, we'll probably need a new `TestDefault` or `DefaultTest` crate/module/something that contains reasonable defaults for all the structs we need.

### DIFF
--- a/crates/but/src/id/tests.rs
+++ b/crates/but/src/id/tests.rs
@@ -1,4 +1,4 @@
-use crate::legacy::id::{CliId, IdMap};
+use crate::{CliId, IdMap};
 use anyhow::bail;
 use bstr::BString;
 use but_testsupport::Sandbox;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -165,7 +165,7 @@ dependencies = [
  "anstyle-lossy",
  "anstyle-parse",
  "html-escape",
- "unicode-width 0.2.0",
+ "unicode-width",
 ]
 
 [[package]]
@@ -738,7 +738,6 @@ dependencies = [
  "cli-prompts",
  "colored",
  "command-group",
- "crossterm 0.29.0",
  "git2",
  "gitbutler-branch",
  "gitbutler-branch-actions",
@@ -751,20 +750,19 @@ dependencies = [
  "insta",
  "minus",
  "posthog-rs",
- "ratatui",
  "regex",
  "rmcp",
  "serde",
  "serde_json",
  "shell-words",
  "snapbox",
- "strum 0.27.2",
+ "strum",
  "terminal_size",
  "tokio",
  "tracing",
  "tracing-forest",
  "tracing-subscriber",
- "unicode-width 0.2.0",
+ "unicode-width",
 ]
 
 [[package]]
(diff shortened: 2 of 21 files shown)


crates/but/src/id/tests.rs
```diff
@@ -0,0 +1,184 @@
...
+        // TODO: fix this - should be read-only, but needs a `but-db` refactor to support read-only DB access.
+        let actual = id_map.parse_str(&commit_id.to_hex_with_len(2).to_string())?;
+        assert_eq!(actual.len(), 1, "The commit can be resolved");
+        assert!(
+            matches!(&actual[0], CliId::Commit { oid } if oid == commit_id,),
+            "The id refers to the right commit"
+        );
+    }
+    Ok(())
+}
```
Byron, line 27: Maybe this is something that should be removed in one of the next PRs as it's just an example, and it's clearly nicer to test assumptions with the DSL.
On the other hand, doing a real E2E test with real data might still be valuable, even though we don't really have a reason for that yet.
It's entirely up to you and what you find useful.

crates/but/src/id/tests.rs
```diff
@@ -0,0 +1,184 @@
...
+    );
+    Ok(())
+}
+
+mod util {
+    use bstr::BString;
+    use but_core::ref_metadata::StackId;
+    use but_hunk_assignment::HunkAssignment;
+    use but_workspace::branch::Stack;
+    use but_workspace::ref_info::{Commit, LocalCommit, Segment};
```
Byron, line 102: I put the utilities into their own module primarily to isolate their dependencies. Additionally, having the modules makes clear what is supposed to be used by tests, while allowing internal private functions.

Byron (approved):
That's a great step forward and, as it turns out, `IdMap` isn't dependent on legacy code or dependencies anymore 🎉.

I also love the testability of it, and one can chose between 'real' data from a sandboxed `Context`, or fully mocked data, depending on the usecase. However, I also think that the integration tests can be removed or pruned to only what we truly want to test. Having at least one E2E test with real data (and diverse data) seems like a good thing to have.

Additionally, I see their value as 'example' on how future utilities can be written and tested.

crates/but/src/id/tests.rs
jonathantanmy2: Thanks, removed in #11506. We are already doing E2E tests whenever an E2E test runs a command that either displays an ID (like `status`) or parses an ID (like `rub`), so I think we're covered there.

PR #11497, opened 2025-12-09
Fix bugs related to the image diff feature

nshcr:
This PR fixes several bugs introduced by that feature PR (https://github.com/gitbutlerapp/gitbutler/pull/11456):

1. Added a CSP exception rule to allow loading image size information in the commit preview. This issue only occurred in production builds.
2. Adjusted the frontend logic for reading before / after image files, removing the `commit` source type — only `blob` and `workspace` are needed.
3. Fixed an issue where image size changes were incorrectly shown for added or deleted files. This information should only appear when both the before and after images exist.

For the first two fixes, please refer to [the discussion on Discord](https://discord.com/channels/1060193121130000425/1204524097241878629/1447820731391016960). The third issue is shown in the image below.

### DIFF
--- a/crates/gitbutler-tauri/tauri.conf.json
+++ b/crates/gitbutler-tauri/tauri.conf.json
@@ -58,8 +58,8 @@
 		"security": {
 			"csp": {
 				"default-src": "'self'",
-				"img-src": "'self' asset: https://asset.localhost data: tauri://localhost https://avatars.githubusercontent.com https://*.gitbutler.com  https://gitbutler-public.s3.amazonaws.com https://*.gravatar.com https://io.wp.com https://i0.wp.com https://i1.wp.com https://i2.wp.com https://i3.wp.com https://github.com https://*.googleusercontent.com https://*.giphy.com/ blob: tauri://localhost",
-				"connect-src": "'self' ipc: http://ipc.localhost https://eu.posthog.com https://eu.i.posthog.com https://eu-assets.i.posthog.com https://app.gitbutler.com https://o4504644069687296.ingest.sentry.io ws://localhost:7703 https://github.com https://api.github.com https://api.openai.com https://api.anthropic.com https://app.staging.gitbutler.com https://*.gitlab.com https://gitlab.com wss://irc.gitbutler.com:8097",
+				"img-src": "'self' asset: https://asset.localhost data: tauri://localhost https://avatars.githubusercontent.com https://*.gitbutler.com  https://gitbutler-public.s3.amazonaws.com https://*.gravatar.com https://io.wp.com https://i0.wp.com https://i1.wp.com https://i2.wp.com https://i3.wp.com https://github.com https://*.googleusercontent.com https://*.giphy.com/ blob:",
+				"connect-src": "'self' ipc: http://ipc.localhost https://eu.posthog.com https://eu.i.posthog.com https://eu-assets.i.posthog.com https://app.gitbutler.com https://o4504644069687296.ingest.sentry.io ws://localhost:7703 https://github.com https://api.github.com https://api.openai.com https://api.anthropic.com https://app.staging.gitbutler.com https://*.gitlab.com https://gitlab.com wss://irc.gitbutler.com:8097 data: blob:",
 				"script-src": "'self' https://eu.posthog.com https://eu.i.posthog.com https://eu-assets.i.posthog.com",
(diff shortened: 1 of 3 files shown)


crates/gitbutler-tauri/tauri.conf.json
```diff
@@ -58,8 +58,8 @@
...
-				"img-src": "'self' asset: https://asset.localhost data: tauri://localhost https://avatars.githubusercontent.com https://*.gitbutler.com  https://gitbutler-public.s3.amazonaws.com https://*.gravatar.com https://io.wp.com https://i0.wp.com https://i1.wp.com https://i2.wp.com https://i3.wp.com https://github.com https://*.googleusercontent.com https://*.giphy.com/ blob: tauri://localhost",
-				"connect-src": "'self' ipc: http://ipc.localhost https://eu.posthog.com https://eu.i.posthog.com https://eu-assets.i.posthog.com https://app.gitbutler.com https://o4504644069687296.ingest.sentry.io ws://localhost:7703 https://github.com https://api.github.com https://api.openai.com https://api.anthropic.com https://app.staging.gitbutler.com https://*.gitlab.com https://gitlab.com wss://irc.gitbutler.com:8097",
+				"img-src": "'self' asset: https://asset.localhost data: tauri://localhost https://avatars.githubusercontent.com https://*.gitbutler.com  https://gitbutler-public.s3.amazonaws.com https://*.gravatar.com https://io.wp.com https://i0.wp.com https://i1.wp.com https://i2.wp.com https://i3.wp.com https://github.com https://*.googleusercontent.com https://*.giphy.com/ blob:",
```
nshcr, line 61: `tauri://localhost` appeared twice, so I removed one of them.

crates/gitbutler-tauri/tauri.conf.json
```diff
@@ -58,8 +58,8 @@
...
+				"connect-src": "'self' ipc: http://ipc.localhost https://eu.posthog.com https://eu.i.posthog.com https://eu-assets.i.posthog.com https://app.gitbutler.com https://o4504644069687296.ingest.sentry.io ws://localhost:7703 https://github.com https://api.github.com https://api.openai.com https://api.anthropic.com https://app.staging.gitbutler.com https://*.gitlab.com https://gitlab.com wss://irc.gitbutler.com:8097 data: blob:",
```
nshcr: I added `blob:` as well, just in case we need it in the future.

Byron: That seems like a good idea, at least under the assumption that this isn't anything exploitable. Is me not knowing how an indication that it's not possible, or that it's definitely possible 😅?

Byron: Also I was under the impression that we need `data: blob` to display the images in the first place? I am definitely missing something.

Byron (approved):
Thanks a million, this looks good to me!

However, I didn't try it locally for a lack of test repository for binary data (I should probably have that), but I also thought @PavelLaptev would be the one to actually validate and merge anyway.

PavelLaptev (approved):

PavelLaptev:
Looks good to me :-)

Byron:
@PavelLaptev Did you try it? Because I did not 😅.

crates/gitbutler-tauri/tauri.conf.json
nshcr: @Byron `blob:` and `data:` are mainly used for embedding media files, and if `blob:` isn't placed under `script-src`, there shouldn't be any risk of malicious exploitation, maybe.

https://github.com/gitbutlerapp/gitbutler/blob/91443fe382eb85bc2981391e15784ba0da3f8611/packages/ui/src/lib/components/ImageDiff.svelte#L65-L71

The reason `data:` needs to be added to `connect-src` is that this piece of code fetches a `data:` URL to get the image file's size. The exception for displaying images is already configured under `img-src`.

After thinking it through, `blob:` isn't necessary here, since we normally wouldn't fetch a blob. I've removed it — thanks for pointing that out!

PavelLaptev:
@Byron, I’ve just checked it. I can confirm this one:

> Fixed an issue where image size changes were incorrectly shown for added or deleted files. This information should only appear when both the before and after images exist.

The fix is working

Byron:
Thanks again to everyone, great work!

PR #11498, opened 2025-12-09
fix broken links in DEVELOPMENT.md

uwemaurer:
## 🧢 Changes

removed 404 links with working ones (or removed them)


Byron (approved):
This is a great catch, thanks a lot!

I double-checked the links are indeed correct now, they all work as expected.

PR #11506, opened 2025-12-09
Cleanup `IdMap` tests

jonathantanmy2:
All `IdMap` tests are unit tests now. (As for whether it is tested via integration tests, it is still indirectly tested by `but status` and `but rub` tests, which I think is the correct way to go - testing it via commands that use it.)


Byron:
Thank you, that's much better!

I also took the liberty to refactor the tests to make it clearer what's wrong when they fail (which they should during development). Further, and in a follow-up, I think it makes sense to enforce this utility type to be properly documented. I will let AI have a go and then tune it myself for familiarisation.

PR #11511, opened 2025-12-10
Fix default editor on debian-derived distros

mathijs81:
The current change still didn't do the right thing for me. I finally figured out how it works: debian-based distros use a symlink "editor" to point to the user's preferred editor and they compile git with the default fallback value replaced by that:
https://sources.debian.org/src/git/1%3A2.51.0-1/debian/rules#L17

This change detects that we're on a distro that has an "editor" binary/symlink and uses that if possible.

@Byron will you take a look?

### DIFF
--- a/crates/but/src/tui/get_text.rs
+++ b/crates/but/src/tui/get_text.rs
@@ -1,5 +1,6 @@
 //! Various functions that involve launching the editor.
 use anyhow::Result;
+use which::which;
 
 /// Launches the user's preferred text editor to edit some initial text,
 /// identified by a unique identifier (to avoid temp file collisions).
@@ -60,13 +61,40 @@ fn is_terminal_dumb() -> bool {
 /// - Fall back to platform defaults (vi on Unix, notepad on Windows)
 /// - Add comprehensive unit tests for all scenarios
 fn get_editor_command() -> Result<String> {
-    get_editor_command_impl(&|key| std::env::var(key), is_terminal_dumb())
+    get_editor_command_impl(
+        &|key| std::env::var(key),
+        &|command| which(command).is_ok(),
+        is_terminal_dumb(),
+    )
+}
+
+/// Get the platform fallback editor.
+fn get_platform_editor<F>(command_exists: &F) -> String
+where
+    F: Fn(&str) -> bool,
+{
+    if cfg!(windows) {
+        "notepad".to_owned()
+    } else if cfg!(target_os = "linux") && command_exists("editor") {
+        // Linux: git source code has "vi" as fallback editor, but debian-based distros override this to
+        // "editor" which is an alternatives-managed link to the user's preferred editor.
+        // See https://sources.debian.org/src/git/1%3A2.51.0-1/debian/rules#L17
+        // We check whether "editor" is in the path and return it if it is, otherwise we stay with "vi".
+        "editor".to_owned()
+    } else {
+        "vi".to_owned()
+    }
 }
 
 /// Internal get_editor_command_implementation that can be tested without modifying environment
-fn get_editor_command_impl<F>(env_var: &F, terminal_is_dumb: bool) -> Result<String>
+fn get_editor_command_impl<F, CE>(
+    env_var: &F,
+    command_exists: &CE,
+    terminal_is_dumb: bool,
+) -> Result<String>
 where
     F: Fn(&str) -> Result<String, std::env::VarError>,
+    CE: Fn(&str) -> bool,
 {
     // Try $GIT_EDITOR first
     if let Ok(editor) = env_var("GIT_EDITOR")
@@ -109,8 +137,7 @@ where
     }
 
     // Fallback to platform defaults (DEFAULT_EDITOR)
-    let default_editor = if cfg!(windows) { "notepad" } else { "vi" };
-    Ok(default_editor.to_string())
+    Ok(get_platform_editor(command_exists))
 }
 
 #[cfg(test)]
@@ -126,6 +153,9 @@ mod tests {
                 .ok_or(std::env::VarError::NotPresent)
         }
     }
+    fn doesnt_exist(_: &str) -> bool {
+        false
+    }
 
     #[test]
     fn git_editor_takes_precedence() {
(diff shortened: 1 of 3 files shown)


crates/but/src/tui/get_text.rs
```diff
@@ -60,13 +61,40 @@ fn is_terminal_dumb() -> bool {
...
+
+/// Get the platform fallback editor.
+fn get_platform_editor<F>(command_exists: &F) -> String
+where
+    F: Fn(&str) -> bool,
+{
+    if cfg!(windows) {
+        "notepad".to_owned()
+    } else if cfg!(target_os = "linux") && command_exists("editor") {
+        // linux: git source code has "vi" as fallback editor, but debian-based distros override this to
```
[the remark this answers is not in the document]
mathijs81: done

Byron:
Hey @mathijs81, thanks for making this happen, and sorry for the late response.
When doing my usual refactor I stopped at after this patch which got me thinking…

```patch
diff --git a/crates/but/src/tui/get_text.rs b/crates/but/src/tui/get_text.rs
index 2ece1674ee..cada5afbe5 100644
--- a/crates/but/src/tui/get_text.rs
+++ b/crates/but/src/tui/get_text.rs
@@ -51,14 +51,14 @@ fn is_terminal_dumb() -> bool {
         .unwrap_or(false)
 }
 
-/// Implement get_editor_command to match Git's C implementation of `git_editor`.
+/// Implement `get_editor_command` to match Git's C implementation of `git_editor`.
 ///
-/// - Check GIT_EDITOR environment variable first
+/// - Check `GIT_EDITOR` environment variable first
 /// - Check git config `core.editor` (editor_program)
 /// - Check VISUAL if terminal is not dumb
-/// - Check EDITOR
+/// - Check the `EDITOR` environment variable
 /// - Return error if terminal is dumb and no editor found
-/// - Fall back to platform defaults (vi on Unix, notepad on Windows)
+/// - Fall back to platform defaults (`vi` on Unix, or `editor` on Debian, notepad on Windows)
 /// - Add comprehensive unit tests for all scenarios
 fn get_editor_command() -> Result<String> {
     get_editor_command_impl(
```

The `editor` on Debian didn't seem quite right because we actually implement this behaviour for Linux in general. While Git has a specific patch on Debian, it's nothing we can do.

How do you think this should be reconciled?

Thank you 🙏

mathijs81:
I'm not sure I fully understand your question.
Do you mean that there might be a non-debian(ish) linux user with an 'editor' on the path but for whom this still would be the wrong choice?
I looked shortly at ways of detecting whether the distro seems to be debian(ish) but this seems more hassle than it's worth and just having a working "editor" on the path seems like a strong signal that that's a better choice than the final "vi" fallback.
Or do you have another suggestion of how to handle this?

mathijs81:
As for the comment / description of the algorithm, I think it's confusing to write `if debian -> use "editor"` because that's clearly not what it's doing, but rather we can just describe it as `if "editor" available -> use it` and comment that that is important for debian-based distros?

crates/but/src/tui/get_text.rs
```diff
@@ -60,13 +61,40 @@ fn is_terminal_dumb() -> bool {
...
+        // We check whether "editor" is in the path and return it if it is, otherwise we stay with "vi".
+        "editor".to_owned()
+    } else {
+        "vi".to_owned()
+    }
 }
 
 /// Internal get_editor_command_implementation that can be tested without modifying environment
-fn get_editor_command_impl<F>(env_var: &F, terminal_is_dumb: bool) -> Result<String>
+fn get_editor_command_impl<F, CE>(
```
krlvi, line 90: Thinking about the function signature here - maybe it would be nice to have a something that takes an `Option`? Seeing the tests defining `doesnt_exist()` bubbled this idea

crates/but/src/tui/get_text.rs
```diff
@@ -60,13 +61,40 @@ fn is_terminal_dumb() -> bool {
 /// - Fall back to platform defaults (vi on Unix, notepad on Windows)
 /// - Add comprehensive unit tests for all scenarios
 fn get_editor_command() -> Result<String> {
```
krlvi, line 63: Feels like the function docs can be updated somewhat to reflect the new behavior

crates/but/src/tui/get_text.rs
```diff
@@ -60,13 +61,40 @@ fn is_terminal_dumb() -> bool {
...
+
+/// Get the platform fallback editor.
+fn get_platform_editor<F>(command_exists: &F) -> String
+where
+    F: Fn(&str) -> bool,
+{
+    if cfg!(windows) {
+        "notepad".to_owned()
+    } else if cfg!(target_os = "linux") && command_exists("editor") {
+        // Linux: git source code has "vi" as fallback editor, but debian-based distros override this to
```
krlvi, line 79: I wonder, perhaps this comment is nicer as a function documentation, this way it gets surfaced in IDEs on hover etc

krlvi (commented):
Thanks! Left some general comments for consideration :)

Byron:
> I'm not sure I fully understand your question. Do you mean that there might be a non-debian(ish) linux user with an 'editor' on the path but for whom this still would be the wrong choice? I looked shortly at ways of detecting whether the distro seems to be debian(ish) but this seems more hassle than it's worth and just having a working "editor" on the path seems like a strong signal that that's a better choice than the final "vi" fallback. Or do you have another suggestion of how to handle this?

Yes, I agree. Let's just say that on Linux and if `editor` is available, we will use it, to get closer to what happens in Git.
Then the `which` dependency should also only be necessary on Linux.
Thank you.

mathijs81:
Thanks for the comments, I'm learning from this!

I realized the current path already contains a git invocation to handle the git config, so this can actually be simplified with a `git var` invocation that will remove the need for our code to replicate all the git logic, and then naturally handles the Debian specific behavior.

I propose that change in https://github.com/gitbutlerapp/gitbutler/pull/11516. If you agree that that's the better approach, let's go with that. If not, I will reopen this one and incorporate your comments.

PR #11514, opened 2025-12-10
skip serializableCheck for RTK Query state/cache

uwemaurer:
This serializableCheck is not needed since this data is loaded via JSON request and therefore always serializable.

This change makes the UI a lot snappier and
removes these  warnings:

`SerializableStateInvariantMiddleware took 1590ms, which is more than the warning threshold of 32ms. `


krlvi:
> SerializableStateInvariantMiddleware took 1590ms

Oh interesting! Nice catch. There is some lint error but other than that looks good to go!

krlvi (approved):

PR #11526, opened 2025-12-12
fix(gitlab): ensure gitlab state initializes per-project and cleans up

nshcr:
Fixes #10777

https://github.com/user-attachments/assets/106c1587-ebd4-4e2e-96bd-b7c8cf4570e9

The main reason for this issue is that each project's GitLab state subscription was not properly unsubscribed, which caused the current project to use the previous project's token together with the current project's ID when switching projects.

This PR refactors the subscription and cleanup mechanism for GitLab state to ensure that each project's GitLab state is properly maintained.

A bit of an aside:

Fixing bugs for GitButler has become exhausting for me, but since this bug severely disrupts my workflow, I still spent about an hour fixing it.

I like this software, but the error-prone software quality makes me constantly worry about when the workspace will break again. Sometimes, I spend more time dealing with GitButler's own issues than on the things I actually want to use it for.

I sincerely hope that at least the serious bugs that can be reproduced can be addressed proactively.

### DIFF
--- a/apps/desktop/src/lib/forge/gitlab/gitlabState.svelte.ts
+++ b/apps/desktop/src/lib/forge/gitlab/gitlabState.svelte.ts
@@ -12,6 +12,8 @@ export class GitLabState {
 	private _upstreamProjectId: Writable<string | undefined> | undefined;
 	private _instanceUrl: Writable<string | undefined> | undefined;
 	private _configured: Readable<boolean> | undefined;
+	private _currentProjectId: string | undefined;
+	private _unsubscribers: Array<() => void> = [];
 
 	constructor(private readonly secretService: SecretsService) {
 		this.token = writable<string | undefined>();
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks so much for helping out 🙏!

I could reproduce the bug that it fixes in nightly, and also see that it's fixed with this PR applied.

Could @Caleb-T-Owens take a look and merge it in so we have the fix soon?

> Fixing bugs for GitButler has become exhausting for me, but since this bug severely disrupts my workflow, I still spent about an hour fixing it.
>
>I like this software, but the error-prone software quality makes me constantly worry about when the workspace will break again. Sometimes, I spend more time dealing with GitButler's own issues than on the things I actually want to use it for.
>
> I sincerely hope that at least the serious bugs that can be reproduced can be addressed proactively.

Speaking for myself only, I am very reactive and spend a lot of time every day just to keep up with the inflow of user issues, with quite a backlog of bugs in new code to work through before finally getting to hopefully one day, again, working on API and CLI improvements.
Or in other words, maintaining the current state takes quite a bit away from getting off it.

@krlvi and I also noticed that we dropped the ball on the other conversation we were having and he will reach out to you soon.

PR #11532, opened 2025-12-13
Three-character `CliId`s and some code cleanup

jonathantanmy2:
As @Byron said in https://github.com/gitbutlerapp/gitbutler/pull/11508#issuecomment-3637803767 :

> The biggest thing I thought I found was the functioning of string_hash which is only good for 720 hashes. It's now called from a non-infinite loop, but it might be too limiting in practice as it's hard-coded to 720 iterations, something that seemed safer than a loop.
>
> However, it also shows that we can't scale up these short hashes, even though that might be needed when dealing with bigger workspaces. To my mind, the CLI ID limit can easily be reached when there is a workspace with many commits.
Is this something you can fix algorithmically? Maybe it can even be done efficiently so we don't allocate a new String each time just to throw it away while it's trying to find a match.

This now allows 3-character IDs (I also changed it to no longer be hashes). I've also made it more efficient in that we don't allocate new Strings when testing whether the string is a non-conflicting ID.

~You also mentioned proofreading the documentation - I'll do that later.~ I did this too.

### DIFF
--- a/crates/but/src/id/mod.rs
+++ b/crates/but/src/id/mod.rs
@@ -10,6 +10,7 @@
 
 #![forbid(missing_docs)]
 
+use anyhow::bail;
 use bstr::{BStr, BString, ByteSlice};
 use but_core::ref_metadata::StackId;
 use but_ctx::Context;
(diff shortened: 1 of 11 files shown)


crates/but/src/id/mod.rs
```diff
@@ -3,31 +3,119 @@
...
+        let mut result = String::new();
+
+        let (quo, rem) = divmod(self.0 as usize, 20);
+        result.push(Self::FIRST_CHARS[rem] as char);
+        let (quo, rem) = divmod(quo, 36);
+        result.push(Self::SUBSEQUENT_CHARS[rem] as char);
+        let (quo, rem) = divmod(quo, 37);
+        if quo > 0 {
+            // self is too big even for 3 characters.
+            return "00".to_string();
```
Caleb-T-Owens, line 52: I don't have any real context here, so feel free to disregard my comments here.

> // self is too big even for 3 characters.

I'm curious why we don't support arbitrary amounts of characters for the IDs if needed?

I've not been part of the conversation around how hunks are going to be represented, but I'm wondering if we start to include them in the ID pool if we could very quickly run out of even 3 identifiers.

Caleb-T-Owens:
If we are keen to have either 2 or 3 character CLIds, I'm still a little curious about the fallbacks to "00" in a couple of places. Could this cause some hard-to-detect bugs if we do run out of IDs?

crates/but/src/id/mod.rs
jonathantanmy2: The limit is 20 * 36 * 37 (26640) which I don't think we'll run out of, but if we think we will, we can change it to 4 characters. We can't support arbitrary amounts of characters for a (small) performance reason - if we extended it to 4 characters, for example, we would have to check every 4-character substring in every branch name to make sure we don't collide. (I guess we can do this at the very first point that we need 4 characters, so the performance hit will only occur when we need it, but it will slightly complicate the code.)

jonathantanmy2:
> If we are keen to have either 2 or 3 character CLIds, I'm still a little curious about the fallbacks to "00" in a couple of places. Could this cause some hard-to-detect bugs if we do run out of IDs?

The bug of an ID being "00" will be easy to detect, since the user will see it, try it, and then the CLI will interpret it as the unassigned area. Ideally we would enforce that it's impossible for a `UintId` to be greater than or equal to the limit, but I don't think it's worth it for something that's used in only one small part of the code.

crates/but/src/id/mod.rs
```diff
@@ -64,68 +152,67 @@ impl IdMap {
...
-                    let name = branch_name.to_string();
-                    let id = str::from_utf8(&pair)
-                        .expect("if we stored it, it's ascii-alphanum")
-                        .to_owned();
-                    ids_used.insert(id.clone());
-                    branch_name_to_cli_id.insert(branch_name, CliId::Branch { name, id });
-                    continue 'branch_name;
+        for branch_name in branch_names {
+            let name = branch_name.to_string();
+            let id = 'short_id: {
```
Byron, line 234: O…M…G…: after all these years, I never knew that labels can be used in non-loops. And it's so practical, to use them like a return label for a scope.
TIL

Byron (approved):
Thanks a lot, this looks so much better now, and the third-character expansion should get us pretty far.

As this definitely is an improvement, I'd love to merge it right away. Part of me keeps thinking that we don't necessarily have all the tests, but also I couldn't easily create one that I thought was missing.

Something that came to my mind recently is if it's possible to make files more stable as well. With that, rubbing multiple files in multiple invocations would be more feasible as they are relatively stable.
And while searching names for valid Cli-IDs is great, maybe here we'd need another mechanism if we were to pursue this at all. It's more like: "if it's easy, then why not"-kind of territory. 
When I thought about stability, I always thought about how JJ has real hashes and just highlights the unique prefixes based on what it prints to screen. Is this an approach we could *also* have at some point? And now that I write this, I don't think it would be very easy to implement as it needs specific support at print time, but then again, maybe some clever functions on `IdMap` could do the trick already.
Anyway, this is just rambling of lose thoughts at this point 😁.

jonathantanmy2:
> When I thought about stability, I always thought about how JJ has real hashes and just highlights the unique prefixes based on what it prints to screen. Is this an approach we could _also_ have at some point? And now that I write this, I don't think it would be very easy to implement as it needs specific support at print time, but then again, maybe some clever functions on `IdMap` could do the trick already.

This is actually possible right now, since we already have the list of files (and I already plumbed the `to_short_string` method, which we can include formatting information in and maybe rename it). The problem is that we have so many files (and later hunks) that users will probably need to check `status` output anyway to see how many characters they need to type, or type 4 characters all the time. In JJ, they only use it for changes, which are less numerous and more stable (if you're working on the same few changes, you know that their short IDs will remain consistent; it is extremely rare to not add files as you're working, which will add file CLI IDs that possibly conflict with existing ones).

Byron:
Thanks for sharing, particularly the note about hunks - I completely forgot about them. Yet, I am still tentatively optimistic that we haven't maxed out the CliID potential yet, but also think that as long as they work well enough and are solidly tested, we can always make them better later.

PR #11533, opened 2025-12-14
Do not use pager for bare help output

schacon:
Ensure that invoking `but help` or running `but` with no subcommand (which calls help) never uses the pager. 

Replace OutputChannel::new_with_pager with new_without_pager_non_json so help text is emitted directly (human-format non-JSON) for both the top-level grouped help and the push subcommand help.


Byron (commented):
Thanks for the fix!

Let's merge this.

Something that I am still having in my backlog is to see how the pager interacts with input, and how that should be controlled in future (i.e. once the app seeks input it needs to flush what's paged and turn off paging).

Also we'd want to know how users should interact with CLI-IDs - particularly when rubbing they'd need to memorise 2 of them. Or is there a way to re-print the paged output on demand? There doesn't seem to be a built-in way to do that, but the paging crate `minus` should be flexible enough to add custom handlers for that.

Issue #8411, opened 2025-05-06
Linux Distribution Compatibility


Byron:
Currently our distributables require Ubuntu 24 at the very least to have a chance to run, and by the look of https://github.com/tauri-apps/tauri/issues/9039 there are more caveats as to which distributable runs where.

The way to solve this puzzle may be a Matrix to keep track of what runs and what doesn't.

## Distributable Compatibility - Built on Ubuntu 22.04 / 24.04

| **Distribution** | **Package Format** | **Compatibility Status** | **Notes (Specific Issues/Caveats)** |
|-------------------|--------------------|--------------------------|--------------------------------------|
| **Debian**        | DEB-22             |                          |                                      |
|                   | DEB-24             |                          |                                      |
|                   | AppImage-22        |                          |                                      |
|                   | AppImage-24        |                          |                                      |
|                   | Flatpak-22         |                          |                                      |
|                   | Flatpak-24         |                          |                                      |
| **Ubuntu**        | DEB-22             |                          |                                      |
|                   | DEB-24             |                          |                                      |
|                   | AppImage-22        |                          |                                      |
|                   | AppImage-24        |                          |                                      |
|                   | Flatpak-22         |                          |                                      |
|                   | Flatpak-24         |                          |                                      |
| **Fedora**        | RPM-22             |                          |                                      |
|                   | RPM-24             |                          |                                      |
|                   | AppImage-22        |                          |                                      |
|                   | AppImage-24        |                          |                                      |
|                   | Flatpak-22         |                          |                                      |
|                   | Flatpak-24         |                          |                                      |
| **openSUSE**      | RPM-22             |                          |                                      |
|                   | RPM-24             |                          |                                      |
|                   | AppImage-22        |                          |                                      |
|                   | AppImage-24        |                          |                                      |
|                   | Flatpak-22         |                          |                                      |
|                   | Flatpak-24         |                          |                                      |
| **Mint**          | DEB-22             |                          |                                      |
|                   | DEB-24             |                          |                                      |
|                   | AppImage-22        |                          |                                      |
|                   | AppImage-24        |                          |                                      |
|                   | Flatpak-22         |                          |                                      |
|                   | Flatpak-24         |                          |                                      |
| **Pop!_OS**       | DEB-22             |                          |                                      |
|                   | DEB-24             |                          |                                      |
|                   | AppImage-22        |                          |                                      |
|                   | AppImage-24        |                          |                                      |
[...]

PR #11534, opened 2025-12-14
Add Linux compatibility document

slarse:
## 🧢 Changes
This is a draft for the official Linux compatibility policy of the GitButler project.

This has been discussed at some length in [Discord](https://discord.com/channels/1060193121130000425/1445541648686780587) already, and with this I'm trying to cement what we've discussed.

This is a _draft_, both in terms of content and in terms of location. Suggestions for changes in wording are welcome, nay, _expected_! The idea is that this is somehow communicated via the website as well, perhaps just with a link to this policy.

## ☕️ Reasoning
Linux compatibility is shaky, and above all, _unclear_. Linux is not an operating system. Ubuntu 22.04 is, or Fedora 39, or Arch Linux. They're all quite different, and no concerted effort has been made to provide a good experience on any of them.

The [downloads page](https://gitbutler.com/downloads) provides `deb`, `rpm` and `AppImage` packages, which implies support for "something Debian", "something Redhat" and "anything", yet no details are provided. It's all just rather confusing. The AppImage in particular works poorly in many places, and not at all in more.

There is significant risk that this lack of clarity causes people try out GitButler on some Linux distribution of their choice, have a downright crappy experience, and leave with a sour taste in their mouths. Or worse: spread said sour taste to others through word of mouth. Or Reddit.

My proposal is simple: let's narrow down to provide a _great_ experience on a few, mainstream distributions (here, Ubuntu 22.04 and 24.04). We might branch out from there in the future, or we may simply rely on community support for other Linux distros as popularity for GitButler grows.

## 🎫 Affected issues

https://github.com/gitbutlerapp/gitbutler/issues/8411

### DIFF
--- a/LINUX.md
+++ b/LINUX.md
@@ -1,0 +1,35 @@
+# Linux compatibility
+GitButler should run well on most modern Linux distributions, but there are too many software and hardware combinations for the core team to support on its own. As such, the GitButler project maintains official support for a select few Linux distributions, and otherwise relies on community-maintained packages.
+
+This document outlines the current level of commitment toward Linux distributions and packaging formats.
+
+> A more dynamic overview of the current state of Linux compatibility can be found in https://github.com/gitbutlerapp/gitbutler/issues/8411
+
+## Official support
+The officially supported way to install GitButler is with the `deb` package provided on the [downloads page](https://gitbutler.com/downloads). This package is regularly tested to work well with the following distributions.
+
+* Ubuntu 22.04 LTS (jammy)
+* Ubuntu 24.04 LTS (noble)
+
+On these distributions, we aim to provide as good a user experience as on Windows and macOS. Compatibility is routinely verified and compatibility issues are the domain of the core GitButler team. We know that the current user experience does not quite deliver on all fronts, and work on improvements is underway.
+
+This does not mean that the GitButler project is unconcerned with the woes of other Linux distributions. Reports of compatibility issues with _any_ distribution are welcome in the [issue tracker](https://github.com/gitbutlerapp/gitbutler/issues), but with the caveat that such issues are not guaranteed to get prioritized anytime soon. If there is a compatibility trade-off to be made, it will always be made in favor of the officially supported distributions.
+
+## Experimental distribution: `rpm`
+We provide an experimental `rpm` package alongside the `deb` package. It is fundamentally the same thing as the `deb` package and should have the same level of compatibility, but it is not regularly tested on any distribution as part of the development process. We provide it as it is no extra effort to build with the current toolchain, and have no reason to believe it would come with any particular compatibility caveats.
+
+## Experimental distribution: AppImage
+We provide an experimental AppImage that bundles the core dependencies required to run GitButler. The intention is that it should work on most Linux distributions, but experience shows that compatibility is relatively poor.
+
+The AppImage may be removed in the future if compatibility remains poor.
+
+## Community-maintained distributions
+There are several community-maintained distributions of GitButler. Issues with these distributions should typically be brought to the attention of their respective maintainers.
+
+> Know of one we missed? Submit a PR to keep us in the loop!
+
+* Arch Linux User Repository (AUR)
+    - [gitbutler](https://aur.archlinux.org/packages/gitbutler)
+    - [gitbutler-bin](https://aur.archlinux.org/packages/gitbutler-bin)
+* Flatpak
+    - [GitButler](https://flathub.org/en/apps/com.gitbutler.gitbutler)


LINUX_COMPATIBILITY_POLICY.md
```diff
@@ -0,0 +1,33 @@
+# Linux compatibility policy
+The GitButler project aims to provide as good a level of compatibility as possible with the various Linux distributions out there. This document outlines the current level of commitment toward different Linux distributions. GitButler _should_ run well on most modern Linux distributions if built specifically for them, but compatibility is only regularly monitored for the officially supported distributions.
```
Byron: "as good a level of compatibility as possible with" - would "best-effort" be clearer? Or is there another way of writing this that sounds less like a truism?

LINUX.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Official support
+The officially supported way to install GitButler is with the `deb` package provided on the [downloads page](https://gitbutler.com/downloads). This package is regularly tested to work well with the following distributions.
+
+* Ubuntu 22.04 LTS (jammy)
+* Ubuntu 24.04 LTS (noble)
+
+On these distributions, we aim to provide as good a user experience as on Windows and macOS. Compatibility is routinely verified and compatibility issues are the domain of the core GitButler team. We know that the current user experience does not quite deliver on all fronts, and work on improvements is underway.
+
+This does not mean that the GitButler project is unconcerned with the woes of other Linux distributions. Reports of compatibility issues with _any_ distribution are welcome in the [issue tracker](https://github.com/gitbutlerapp/gitbutler/issues), but with the caveat that such issues are not guaranteed to get prioritized anytime soon. If there is a compatibility trade-off to be made, it will always be made in favor of the officially supported distributions.
```
Byron, line 16: > Reports of compatibility issues with _any_ distribution are welcome in the [issue tracker](https://github.com/gitbutlerapp/gitbutler/issues), but with the caveat that such issues are not guaranteed to get prioritized anytime soon.

Is this where we'd want to suggest people to make PRs to help their distro? Is this a point where we should plug semi-official support for distro-maintainers?

Byron (commented):
Thanks again for getting this started!

Some thoughts I wanted to share below, along with the question if we should state that we may drop experimental support unannounced? Maybe it's implied, but it's to say nobody should rely on it.

*~~Maybe there is more, I am in a rush~~ Maybe I missed a lot, I was in a rush 😅*

LINUX_COMPATIBILITY_POLICY.md
slarse: I struggled with this part. My first attempt said even less, with more words :)

But I think this opening statement could perhaps be a good place to immediately make a reliance on community contributions clear. Something like:

```suggestion
GitButler should run well on most modern Linux distributions, but there are too many software and hardware combinations for the core team to support on its own. As such, the GitButler project maintains official support for a select few Linux distributions, and otherwise relies on community-maintained packages.

This document outlines the current level of commitment toward Linux distributions and packaging formats.
```

Wdyt, is that more meaningful?

LINUX.md
slarse: Probably, I'll find a way to shove it in there tonight!

LINUX_COMPATIBILITY_POLICY.md
Byron: This is fantastic, I wouldn't change a thing!

LINUX.md
Byron: The change above seems to already to it. Maybe we just merge this once #11412 is proven to work (better).

slarse: I'd tend to agree, sounds good! Do we want a document called `LINUX_COMPATIBILITY_POLICY.md` in the repository root? Should it be moved elsewhere? Should it be mentioned or even inlined in the README?

Byron: I'd probably call it `LINUX.md` or `LINUX_SUPPORT` and then remove mentions of "policy" to keep it 'more fun' if that makes sense :D.
Thus, I also wouldn't dare to edit the GH frontpage by moving it into the README, while also thinking for us it's enough to be able to refer to a document while having the name `LINUX` stick out in the file list.

I'd be ready to merge everything when you are.

slarse: > I'd probably call it LINUX.md

Let's go with that.

> Thus, I also wouldn't dare to edit the GH frontpage by moving it into the README

You have way more restraint than I do. I'm an "easier to ask for forgiveness than permission" kind of person :)

PR updated with the suggestions, it's ready for merge!

Byron (commented):
Awesome, let's have it!

PR #11538, opened 2025-12-15
Don't allow combos in lexical fields

### LINKED ISSUE #11536: "Command env_vars not found" when typing env into codegen
### Version

0.18.2

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

Whenever I type a word that starts with "env" in the codeGen text input I get a Unhandled Exception error "Command env_vars not found" and the "v" of "env" gets eaten. 
Its not a big deal but a touch annoying

1. go into codegen
2. type the word "env"

the word should just be there without any unhandled Exception messages

```shell

```

Caleb-T-Owens:
This adds a check for the `_lexicalHandled` property which seems to be 
an indicator as to whether an event was indeed handled by lexical.

* Closes https://github.com/gitbutlerapp/gitbutler/issues/11536#issuecomment-3654489173

### DIFF
--- a/apps/desktop/src/lib/utils/hotkeys.ts
+++ b/apps/desktop/src/lib/utils/hotkeys.ts
@@ -79,7 +79,8 @@ export function createKeybind(keybinds: KeybindDefinitions) {
 			if (
 				event.repeat ||
 				event.target instanceof HTMLInputElement ||
-				event.target instanceof HTMLTextAreaElement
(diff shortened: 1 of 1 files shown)


Byron:
Copilot is a little slow and often cannot bring in its review before it's auto-merged. In this case, it seems it found some interesting bits to consider. CC @Caleb-T-Owens.

PR #11555, opened 2025-12-16
Consolidate repeated metadata using workspace manifest

alerque:


Byron (approved):
I think this is very helpful and can simplify maintenance for us.

The one concern I have is that the `edition` field is now also shared, even though from experience it's useful to be able to do a couple of crates at a time as time permits. But… when the time comes, I think we can also change the field back to be a local edition, so I think this PR would be fine to bring in as is.

I validated every manifest change it looks like no error snuck in. It's also something Copilot is pretty good in detecting.

PR #11571, opened 2025-12-17
Restrict visibility of internal functions

krlvi:


Byron:
There a probably a lot of these in many places as I tend to make it pub as long as I know that the module they are in is not pub, to indicate it's for consumption of those who can naturally see it.
The upside of this is less noisy visibility, the downside is that someone making the module `pub` may expose functions that were not meant to be exported from the crate.
Do you think the latter is something we need to protect against? If so, it should probably be a guideline and something in `agent.md`.

PS: Rust does the right thing when it comes to its 'unused' checking even with `pub` functions as long as they aren't exported from the crate.

krlvi:
I pushed it mainly for the CI. Trying to weed out the actually problematic pub functions, which is likely not these ones

PR #11579, opened 2025-12-18
Do not invoke pager for `but status` output

schacon:
Avoid starting the pager for the output of the `but status` (and `stf`) command by passing an explicit flag when constructing the OutputChannel. Previously the code always created a pager-capable OutputChannel; now we determine if the pager should be used based on the invoked subcommand and call new_with_optional_pager accordingly.

This change adds a new constructor new_with_optional_pager(format, use_pager). The pager logic now also respects the use_pager flag so JSON/Non-human output or NOPAGER still disable paging, and callers can opt out for commands (like status) that should never invoke a pager.

### DIFF
--- a/crates/but/src/lib.rs
+++ b/crates/but/src/lib.rs
@@ -85,8 +85,13 @@ pub async fn handle_args(args: impl Iterator<Item = OsString>) -> Result<()> {
     } else {
         args.format
     };
-    // Set it so code past this point can assume it's set.;
-    let mut out = OutputChannel::new_with_pager(output_format);
+    // Determine if pager should be used based on the command
+    let use_pager = !matches!(
+        args.cmd,
+        Some(Subcommands::Status { .. }) | Some(Subcommands::Stf { .. })
+    );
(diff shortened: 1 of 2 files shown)


schacon:
@Byron I'm getting some weirdness with the legacy feature stuff here. Can you give me a hint what a better way to do this would be? I just don't want `but status` to ever use a pager.

Byron:
This should go through now, and I enabled auto-merge.

If I remember correctly, the real goal behind this PR is to be able to copy-paste multiple CliIDs. Another approach to make this happen would be to use a pager, but re-print the pager contents on when it's quit.

Is this something you'd want, @schacon?

Byron:
And now that I think about it: we have full programmatic control over the pager. So if building but-rub commands is the goal, we could help build these invocations, allowing the user to click CliIds, and show the respective `but rub` command in the bottom-most line. On exit, it could be printed to stdout so users can copy/paste it from there to execute. Or we even execute it for them.
With that the lines would blur between a CLI and a TUI, but suffice to say, it's all possible.

PR #11581, opened 2025-12-18
Test case for funky scenario

Caleb-T-Owens:


Byron:
Let's bring it in!

I removed the panic based on [this Discord comment](https://discord.com/channels/1060193121130000425/1204524097241878629/1451207955608502385).
From my point of view, the test doesn't show much as the complex bits aren't really related to the workspace, but maybe one day it can catch something so let's keep it.

It's also good as first demonstration of AI being able to recreate a Git commit graph using the primitives found here in the `scenario.sh` script.
I couldn't yet make codegen work for this, but I also never tried it in the actual codebase, but always asked for standalone scripts based on an SVG. The commit graph, commit by commit, seems to work better overall, despite its potential for ambiguity.

All in all, I like these vibed test-cases, and will try to make my own when the opportunity arises.

PR #11604, opened 2025-12-22
better base check (json and human)


schacon:
This is an update to the `but base check` command. Previously, `--json` was a no-op, this actually implements it. It also previously showed recent commits rather than upstream commits. Finally, I updated the style to match the simpler, ansi colored style preferred by the rest of the commands, rather than the emoji laden on this command used.

CleanShot 2025-12-22 at 20 47 51@2x

If `--json` is specified, it now works properly:

CleanShot 2025-12-22 at 20 48 35@2x


crates/but/src/command/legacy/base.rs
```diff
@@ -6,59 +6,176 @@ use gitbutler_branch_actions::upstream_integration::{
...
     StackStatuses::{UpToDate, UpdatesRequired},
 };
+use serde::Serialize;
 
 use crate::{args::base, utils::OutputChannel};
 
+/// JSON output for `but base check`
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct BaseCheckOutput {
```
Byron: All these would be better when placed in a `json` module (I think it's also done in `but status`).

Byron (commented):
There is just one comment from me, and maybe the request for a test as I don't think there is one yet. Pinning the output now to something you like will help to keep it that way later once the command is 'modernised'.

crates/but/src/command/legacy/base.rs
schacon: @Byron split this into `base/mod.rs` and `base/json.rs`

crates/but/src/command/legacy/base/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+                            writeln!(
+                                out,
+                                "❗️ There are uncommitted changes in the worktree that may conflict with
+                            the updates. Please commit or stash them and try again."
+                            )?;
+                        }
+                        None
+                    } else {
+                        if let Some(out) = out.for_human() {
+                            writeln!(out, "🔄 Updating branches...")?;
```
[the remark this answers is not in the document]
schacon: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/gitbutlerapp/gitbutler/pull/11604#discussion_r2645156405)

schacon:
Moved to #11632

PR #11609, opened 2025-12-23
id: ensure committed file info is sorted

jonathantanmy2:
I put this on #11588 because it depends on `id: split into modules`. (If that PR is rejected for whatever reason, I can make the same change in the old location of the code.)

---

Info about committed files will later go into a `BTreeSet` with specific ordering requirements; the caller in `mod.rs` assumes the invariants documented in `FileInfo`. In `from_workspace_commits_and_status`, sort the info so that we maintain those invariants.


Byron:
Thanks a lot! Once this is merged I will head over to the base of this PR to keep going there.

PR #11610, opened 2025-12-23
Misc doc adjustments that should have gone into #11588

Byron:


Byron:
Argh, Copilot you are right, I shouldn't have merged this automatically. Was in a rush, now I'd have to give it another round (but won't).

PR #11611, opened 2025-12-23
feat(gitlab): support rebase / squash merge method for MR

nshcr:
Hi! While using the GitLab integration, I noticed that the rebase and squash merge methods don't work.
After checking the relevant code, I found that GitButler only implements logic for direct merges and does not handle other merge methods.

I briefly reviewed the GitLab API documentation, then with the help of Claude Sonnet 4.5, I implemented support for rebase and squash when merging MRs.

All the code has gone through my own review and testing, and both the rebase and squash methods work correctly on the official GitLab instance.

Here are a few differences compared to GitHub that are worth noting:

1. Depending on the project configuration, GitLab always creates a merge commit for all merge methods, whereas a rebase merge on GitHub does not.
2. GitLab's rebase merge process consists of two asynchronous steps: first rebasing the MR branch, then performing the merge. In between, the API needs to be polled to check whether the rebase has completed.

For more details, please refer to the code comments I added.

And in advance, Merry Christmas🎄! This isn't an urgent request — feel free to leave this PR until after the holidays.

(Sorry for the delayed reply on Discord — the past two weeks have been unusually busy for me. I'll get back to it soon.)

### DIFF
--- a/apps/desktop/src/lib/forge/gitlab/gitlabPrService.svelte.ts
+++ b/apps/desktop/src/lib/forge/gitlab/gitlabPrService.svelte.ts
@@ -146,10 +146,18 @@ function injectEndpoints(api: GitLabApi) {
 				invalidatesTags: (result) => [invalidatesItem(ReduxTag.GitLabPullRequests, result?.number)]
 			}),
 			mergePr: build.mutation<undefined, { number: number; method: MergeMethod }>({
-				queryFn: async ({ number }, query) => {
+				queryFn: async ({ number, method }, query) => {
 					try {
 						const { api, upstreamProjectId } = gitlab(query.extra);
-						await api.MergeRequests.merge(upstreamProjectId, number);
+
+						// Note: GitLab's rebase is not supported here as it requires a two-step async process.
+						// See: https://github.com/gitbutlerapp/gitbutler/pull/11611
+
+						// For 'merge' and 'squash' methods, use the merge API directly
+						await api.MergeRequests.merge(upstreamProjectId, number, {
+							squash: method === 'squash',
+							shouldRemoveSourceBranch: true
+						});
 						return { data: undefined };
 					} catch (e: unknown) {
 						return { error: toSerializable(e) };
--- a/apps/desktop/src/components/MergeButton.svelte
+++ b/apps/desktop/src/components/MergeButton.svelte
@@ -1,5 +1,7 @@
 <script lang="ts">
+	import { DEFAULT_FORGE_FACTORY } from '$lib/forge/forgeFactory.svelte';
 	import { MergeMethod } from '$lib/forge/interface/types';
+	import { inject } from '@gitbutler/core/context';
 	import { persisted, type Persisted } from '@gitbutler/shared/persisted';
 
 	import { ContextMenuItem, ContextMenuSection, DropdownButton } from '@gitbutler/ui';
@@ -25,6 +27,9 @@
 		kind = 'outline'
 	}: Props = $props();
 
+	const forge = inject(DEFAULT_FORGE_FACTORY);
+	const isGitLab = $derived(forge.current.name === 'gitlab');
+
 	function persistedAction(projectId: string): Persisted<MergeMethod> {
 		const key = 'projectMergeMethod';
 		return persisted<MergeMethod>(MergeMethod.Merge, key + projectId);
@@ -32,14 +37,28 @@
 
 	const action = persistedAction(projectId);
 
+	// If GitLab and action is rebase, reset to merge
+	$effect(() => {
+		if (isGitLab && $action === MergeMethod.Rebase) {
+			$action = MergeMethod.Merge;
+		}
+	});
+
 	let dropDown: ReturnType<typeof DropdownButton> | undefined;
 	let loading = $state(false);
 
-	const labels = {
-		[MergeMethod.Merge]: 'Merge pull request',
+	// Available merge methods based on forge type
+	const availableMethods = $derived(
+		isGitLab
+			? [MergeMethod.Merge, MergeMethod.Squash]
+			: [MergeMethod.Merge, MergeMethod.Rebase, MergeMethod.Squash]
+	);
+
+	const labels = $derived({
+		[MergeMethod.Merge]: isGitLab ? 'Merge merge request' : 'Merge pull request',
 		[MergeMethod.Rebase]: 'Rebase and merge',
 		[MergeMethod.Squash]: 'Squash and merge'
-	};
+	});
 </script>
 
 <DropdownButton
@@ -62,7 +81,7 @@
 	{labels[$action]}
 	{#snippet contextMenuSlot()}
 		<ContextMenuSection>
-			{#each Object.values(MergeMethod) as method}
+			{#each availableMethods as method}
 				<ContextMenuItem
 					label={labels[method]}
 					onclick={() => {


apps/desktop/src/lib/forge/gitlab/gitlabPrService.svelte.ts
```diff
@@ -146,10 +146,53 @@ function injectEndpoints(api: GitLabApi) {
...
 						const { api, upstreamProjectId } = gitlab(query.extra);
-						await api.MergeRequests.merge(upstreamProjectId, number);
+
+						// Note: Unlike GitHub, GitLab's rebase is a two-step async process
+						if (method === 'rebase') {
+							// Rebase the source branch onto the target branch
+							// This is an async operation that returns immediately with 202 status
+							await api.MergeRequests.rebase(upstreamProjectId, number, {
+								skipCI: false
+							});
```
nshcr: Should the rebase operation skip CI? I'm a bit unsure here, since CI will definitely run again during the second merge step.

apps/desktop/src/lib/forge/gitlab/gitlabPrService.svelte.ts
```diff
@@ -146,10 +146,53 @@ function injectEndpoints(api: GitLabApi) {
...
+								// Check if rebase is complete
+								if (!mr.rebase_in_progress) {
+									rebaseComplete = true;
+								}
+								attempt++;
+							}
+
+							if (!rebaseComplete) {
+								throw new Error('Rebase operation timed out. Please try merging again later.');
+							}
```
nshcr: Is a 30-second polling interval reasonable? If the GitLab pipeline includes long-running CI jobs, this might cause the merge process to time out or be interrupted.

apps/desktop/src/lib/forge/gitlab/gitlabPrService.svelte.ts
```diff
@@ -146,10 +146,53 @@ function injectEndpoints(api: GitLabApi) {
...
+							// After rebase completes, perform the merge
+							await api.MergeRequests.merge(upstreamProjectId, number, {
+								shouldRemoveSourceBranch: true
+							});
+						} else {
+							// For 'merge' and 'squash' methods, use the merge API directly
+							await api.MergeRequests.merge(upstreamProjectId, number, {
+								squash: method === 'squash',
+								shouldRemoveSourceBranch: true
+							});
```
nshcr: Based on krlvi's reply in this PR (https://github.com/gitbutlerapp/gitbutler/pull/8222#issuecomment-2828490332), MRs created by GitButler explicitly specify deletion of the source branch, but this isn't necessarily true for MRs created elsewhere.

To ensure the stack branches feature works correctly, I think it's necessary to explicitly pass the `shouldRemoveSourceBranch` parameter when performing the merge.

apps/desktop/src/lib/forge/gitlab/gitlabPrService.svelte.ts
```diff
@@ -146,10 +146,53 @@ function injectEndpoints(api: GitLabApi) {
...
+								await sleep(1000);
+
+								const mr = await api.MergeRequests.show(upstreamProjectId, number, {
+									includeRebaseInProgress: true
+								});
+								// Check if rebase is complete
+								if (!mr.rebase_in_progress) {
+									rebaseComplete = true;
+								}
+								attempt++;
```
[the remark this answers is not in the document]
nshcr: Calling the rebase API and then waiting one second before checking whether the rebase has completed is reasonable — in practice, this operation will not finish within a single second.

apps/desktop/src/lib/forge/gitlab/gitlabPrService.svelte.ts
```diff
@@ -146,10 +146,53 @@ function injectEndpoints(api: GitLabApi) {
...
+						if (method === 'rebase') {
+							// Rebase the source branch onto the target branch
+							// This is an async operation that returns immediately with 202 status
+							await api.MergeRequests.rebase(upstreamProjectId, number, {
+								skipCI: false
+							});
+
+							// Poll for rebase completion before merging
+							// GitLab's rebase operation is asynchronous, so we need to wait
+							const maxAttempts = 30; // 30 seconds timeout
```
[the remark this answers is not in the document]
nshcr: A meaningless suggestion.

apps/desktop/src/lib/forge/gitlab/gitlabPrService.svelte.ts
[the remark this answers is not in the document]
nshcr: This suggestion makes sense. According to the official API documentation, a failed rebase will return a completed status along with failure message. I'll update the error-handling logic for this case a bit later.

https://docs.gitlab.com/api/merge_requests/#rebase-a-merge-request

apps/desktop/src/lib/forge/gitlab/gitlabPrService.svelte.ts
Byron: If this was `skipCI: true`, would that mean that after the merge, CI will run, to possibly detect that something went wrong when it's too late?
It's interesting as what CI should check is what would end up in the target branch if the desired strategy (i.e. merge, rebase, squash) is applied.

Because if `skipCI: false` then I am afraid that the rebase can't typically complete in time at least if CI needs to be completed for the rebase to be done. 

Can this be, that the `rebase_in_progress` flag is true even *after* the rebase was done just because CI is running?

Somehow I feel quite uneasy about this manual implementation of a feature that is best implemented/queued in the forge itself.

apps/desktop/src/lib/forge/gitlab/gitlabPrService.svelte.ts
Byron: To me it seems unreasonable to be forced to deal with that as a UI, as the only correct way to do it would be to have no timeout at all. And if that was the case, one would also have to make sure the operation is resumable, and then it gets complicated quickly.

Even with a 30 second timeout I wonder what happens if the user clicks to the branch view and back to the workspace. Will the UI elements still have the correct state, or will the UI be able to race against itself?

Byron:
Thanks a lot for giving this a shot, and for highlighting possible concerns.

For now, I only looked at those and added my thoughts to them. My fear is that due to the way the GitLab API has to be used, implementing the `rebase` feature correctly is very, very difficult. And even if it would work, it could break easily without exhaustive tests.

apps/desktop/src/lib/forge/gitlab/gitlabPrService.svelte.ts
nshcr: I've been thinking this problem as well. Letting the rebase process also run CI helps ensure the correctness of the rebase result, but skipping CI here is also reasonable and would allow the rebase to complete faster.

I haven't tested how CI affects the `rebase_in_progress` state yet. I didn't test this flow on a GitLab repository with pipelines configured, covering this scenario would likely require preparing several test cases.

Honestly, I can't say what the best approach is. Removing this parameter and keeping the default behavior to avoid explicitly setting it and follow the project's configuration (if there have) might also be a reasonable option.

apps/desktop/src/lib/forge/gitlab/gitlabPrService.svelte.ts
nshcr: I think there's at least one thing we can be confident about: if the rebase completes successfully, then even if the operation times out for some reason, running rebase again usually doesn't need to do anything.

In that case, a second click on the rebase-merge button could go straight to the merge step after a quick check. In practice, the situation shouldn't be that complex.

If we don't set a timeout, then given that GitButler already checks whether a PR is mergeable via polling, perhaps we could replace the 30-second polling within a single click with a UX that asks the user to click twice — once for rebase and once for merge.

nshcr:
Thank you very much for the detailed review!

I've responded to your comments, and I also have to admit that, from my perspective, I don't currently have a really good solution to these issues. They depend heavily on platform-specific rules and features of the code hosting service. I'm sorry that all I can really do here is make it "just work".

I'm not entirely sure what the "correct" way to handle rebase merges is. Compared to GitHub, GitLab always creates an extra merge commit, which makes the history look less clean, and it requires at least three API calls to complete the process (rebase, query result, merge). Still, the end result matches my expectations.

As for testing, merging MRs is indeed something that unit tests can't realistically cover — it requires fairly complex end-to-end test setups. I haven't seen such test coverage for merging GitHub PRs either, so I don't think GitLab needs to be held to a higher standard here. For me, having the feature available at all is more important.

Taking a step back, even if rebase doesn't work exactly as GitButler expects, the squash method is still reasonable and reliable. I don't actually rebase MRs very often anyway. Perhaps we should even hide the rebase option in the GitLab forge UI and only keep direct merge and squash merge?

After all, having all three buttons present but not behaving as expected would clearly be worse than simply not offering a partially broken option.

Byron:
Thanks for the detailed response!

I completely agree that testing this as part of automated tests is hard and nothing we'd have to strive for in this PR, particularly while no such setup exists for GitHub.

However, even manually testing and tuning this to do what's correct would be outside of what I can do here.

Nonetheless, I also agree that it's bad to show options that aren't properly working or aren't implemented, so as a PR that can be merged immediately, I'd prefer hiding unsupported or incorrect options instead.

If that's something you'd understandably find unsatisfactory, maybe the PR can be changed to something that only uses built-in one-shot features of the GitLab API so that no polling for a mutation  is necessary. I'd want to totally avoid implementing certain features in a forge using their API, and avoid creating new features in the process.

All that I say without knowing what GitLab offers in their own WebUI, or how it's implemented there. But I kind of doubt they implement anything with polling as closing the tab (or the lid of the laptop) certainly shouldn't affect the outcome of the button press.

nshcr:
So, when the current project is detected as using GitLab, would it be feasible to have the Merge button only show "Merge pull request" and "Squash and merge", hide "Rebase and merge"?
I think squash is essential in many merge scenarios anyway.

https://github.com/gitbutlerapp/gitbutler/blob/2e94b5d50bd3a2f5829718ef2f12fdbcd1966b49/apps/desktop/src/components/MergeButton.svelte#L38-L42

Additionally, I'm inclined to keep the code in this PR that handles the rebase branch condition. I can add comments indicating that this functionality is still incomplete, so it can serve as a reference in case the situation changes in the future or someone else has ideas on how to implement it properly.

Of course, if you feel the codebase shouldn't retain unreachable or inactive code, I'm completely fine with removing it.

If you agree with this change, I'll update the commit later.

Byron:
Yes, it would be great to only show what can be implemented natively (i.e. in one-shot) with the GitLab API and hide everything else. Also I don't think we should keep inactive code.

Thanks a lot, and sorry for the late response.

apps/desktop/src/lib/forge/gitlab/gitlabPrService.svelte.ts
```diff
@@ -146,10 +146,18 @@ function injectEndpoints(api: GitLabApi) {
...
-						await api.MergeRequests.merge(upstreamProjectId, number);
+
+						// Note: GitLab's rebase is not supported here as it requires a two-step async process.
+						// See: https://github.com/gitbutlerapp/gitbutler/pull/11611
+
+						// For 'merge' and 'squash' methods, use the merge API directly
+						await api.MergeRequests.merge(upstreamProjectId, number, {
+							squash: method === 'squash',
+							shouldRemoveSourceBranch: true
+						});
```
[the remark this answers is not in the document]
nshcr: This is the expected behavior, for all non-squash cases, the regular merge method is used.

apps/desktop/src/lib/forge/gitlab/gitlabPrService.svelte.ts
```diff
@@ -146,10 +146,18 @@ function injectEndpoints(api: GitLabApi) {
...
 			}),
 			mergePr: build.mutation<undefined, { number: number; method: MergeMethod }>({
-				queryFn: async ({ number }, query) => {
+				queryFn: async ({ number, method }, query) => {
 					try {
 						const { api, upstreamProjectId } = gitlab(query.extra);
-						await api.MergeRequests.merge(upstreamProjectId, number);
+
+						// Note: GitLab's rebase is not supported here as it requires a two-step async process.
+						// See: https://github.com/gitbutlerapp/gitbutler/pull/11611
```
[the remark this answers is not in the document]
nshcr: This PR provides sufficient context for the change, so I think the self-reference is reasonable.

Byron (approved):
Thanks a lot, this looks great to me!

Maybe @mtsgrd can take a final look?

mtsgrd (approved):
lgtm, thanks @nshcr!

PR #11615, opened 2025-12-23
fix: no-relative-imports fails with windows paths

Rettend:
I use Windows, and I just cloned the repo, install, ran the test script, and it failed.

So I fixed some hardcoded forward slashes by using node:path.

Now every test passes locally

### DIFF
--- a/packages/no-relative-imports/src/noRelativeImportPaths.ts
+++ b/packages/no-relative-imports/src/noRelativeImportPaths.ts
@@ -21,21 +21,21 @@ type TsConfig = {
  * Takes a file or directory path and tries to find the nearest tsconfig.json.
  */
 function findTsConfigDirectory(absolutePath: string): string | undefined {
-	const pathSegments = absolutePath.split('/').slice(1);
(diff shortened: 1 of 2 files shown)


Byron:

Assuming CI passes, I think it would be best if @Caleb-T-Owens took a look as well. We could also consider to run unit-tests on Windows additionally to protect this from regressing again.

Caleb-T-Owens:
@Rettend Thanks you for the contribution!

The parent traversal you introduced is probably preferable to do that way. I was however wondering, if we used the new `normalize` you introduced, could we keep the old parent traversal. What are your thoughts about that?

Rettend:
@Caleb-T-Owens we could, but it'll be ugly.

this always adds `/` at the front, so the path won't be correct for Windows.

```ts
return '/' + path.join(...pathSegments);
```

You would need to regex out the driver letter and then add it back if it's there.

Caleb-T-Owens:
Makes sense! Thank you for the contribution

PR #11622, opened 2025-12-24
fix: ctrl + a to select files on windows

Rettend:
When selecting all files with the keybind it only checked `metaKey`, so on Windows only `ctrl + win + a` worked.

This change makes `ctrl + a` work too.


krlvi (commented):
Thanks for this contribution!

krlvi (approved):

PR #11623, opened 2025-12-24
Unify uncommitted file and hunk CLI IDs

jonathantanmy2:
Copy and paste from commit message:

Teach `IdMap` to return the same data type regardless of whether the
given CLI ID represents a file or a hunk. (Internally, `IdMap` still
recognizes the difference.) This simplifies the `rub` code, which no
longer needs to handle both cases.

I made the design decision to generate uncommitted hunk IDs only if the
(stack assignment, path) pair has more than one hunk. This is because
operating on the lone hunk of a (stack assignment, path) pair is the
same as operating on the pair itself, so it would seem strange to me
as a user if I had 2 separate IDs for that concept. I've included a
describing mechanism to make it clear to the user that we are operating
on hunks, and when a hunk is the lone hunk of the (stack assignment,
path) pair.

The only other command that accepts CLI IDs as input (`absorb`) gains
the ability to operate on hunks "for free".

### DIFF
--- a/crates/but/src/command/legacy/absorb.rs
+++ b/crates/but/src/command/legacy/absorb.rs
@@ -12,7 +12,10 @@ use but_hunk_dependency::ui::HunkDependencies;
 use colored::Colorize;
 use gitbutler_project::Project;
 
-use crate::{CliId, IdMap, command::legacy::rub::parse_sources, utils::OutputChannel};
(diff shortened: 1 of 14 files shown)


krlvi (approved):
Nice simplification here, thanks

PR #11625, opened 2025-12-25
feat(settings): add img-src extra hosts and propagate to CSP

nshcr:
Before:

After:

This PR adds an internal setting to the application that allows users to manually add CSP exception rules for `img-src`.

The image resources shown in the PR description are very likely hosted at non-whitelisted addresses, and I think it's a reasonable requirement to let users configure these manually.

As for images in GitHub PRs, they currently redirect to S3 URLs, but I'm not sure whether GitHub uses a single bucket (most likely not). Adding exceptions for those domains one by one would be impractical, and I don't think we should add a domain exception for them.

By the way, wildcard configurations such as `https://github-production-user-asset-*.s3.amazonaws.com` did not work in my testing.

We probably need a better way to handle CSP rules in general. They introduce a lot of unexpected resource blocking issues that only show up in production builds, making them hard to notice during development.

That said, I don't really have a better solution either. Handing over control of some non-sensitive CSP settings to users seems like a reasonable compromise.

### DIFF
--- a/crates/but-settings/assets/defaults.jsonc
+++ b/crates/but-settings/assets/defaults.jsonc
@@ -34,7 +34,9 @@
 	// Allows for additional "connect-src" hosts to be included. Requires app restart.
 	"extraCsp": {
 		// Additional hosts that the application can connect to.
-		"hosts": []
+		"hosts": [],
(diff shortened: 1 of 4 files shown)


krlvi (approved):
Thank you!

PR #11632, opened 2025-12-26
Move base command into module, add JSON output and simplify human output

schacon:
Reorganize the legacy base command by moving base.rs into base/mod.rs
and adding a dedicated json.rs for JSON output structures. Provide
structured JSON output for `but base check` (BaseCheckOutput,
BaseBranchInfo, UpstreamInfo, UpstreamCommit, BranchStatusInfo) and wire
JSON and human output paths. Simplify human output by removing emoji
usage and standardizing status text, and update call site in lib.rs to
pass the legacy project correctly.


schacon:
Nevermind, we're going to move all the `base` functionality elsewhere - most of this will go to different structures in `status` output.

PR #11636, opened 2025-12-26
Refactor tree_change_diffs API to take Context

krlvi:


crates/but-api/src/legacy/diff.rs
```diff
@@ -17,14 +16,12 @@ use tracing::instrument;
...
-    project_id: ProjectId,
+    ctx: Context,
     change: TreeChange,
 ) -> anyhow::Result<Option<but_core::UnifiedPatch>> {
     let change: but_core::TreeChange = change.into();
-    let project = gitbutler_project::get(project_id)?;
-    let app_settings = AppSettings::load_from_default_path_creating_without_customization()?;
-    let repo = project.open_repo()?;
-    change.unified_patch(&repo, app_settings.context_lines)
+    let repo = ctx.open_repo()?;
```
Byron, line 23: With the context, we basically never want to call `open_repo()`, but use `ctx.repo.get()` instead.
That way we have a chance of reusing a prior instance, along with its caches.

But besides that, and just by looking at the types, the function could now be non-legacy.

krlvi: do you think we could at some point get rid of `open_repo` so that the consumers dont mix it up?

Byron: Yes, let's. I will see if that's a quick fix now.

Byron: https://github.com/gitbutlerapp/gitbutler/pull/11651

PR #11637, opened 2025-12-26
Implement non-legacy api for branch diffs

krlvi:


crates/but-api/src/branch.rs
```diff
@@ -56,3 +60,15 @@ pub fn apply(
...
+/// Gets the changes for a given branch.
+#[but_api(TreeChanges)]
+#[instrument(err(Debug))]
+pub fn branch_diff(ctx: &Context, branch: String) -> anyhow::Result<TreeChanges> {
+    let guard = ctx.shared_worktree_access();
+    let (repo, _meta, graph) =
+        ctx.graph_and_meta_and_repo_from_head(ctx.repo.get()?.clone(), guard.read_permission())?;
+    let reference = repo.find_reference(&branch)?;
+    let ws = graph.to_workspace()?;
+    but_workspace::ui::diff::changes_in_branch(&repo, &ws, reference.name())
```
Byron, line 73: I think this was marked as legacy because it still uses a UI type. While these are not marked as legacy, they are tuned to the needs of JSON and possible the UI.

With `but-api`, there is always 'clean' Rust types which are then transformed with the help of the `but_api` macro.

krlvi: Oh, I though it was legacy because it contained the "stack ID". I dont mind moving this to the legacy module in this case. Or... we should be able to make it use non-ui types too

Byron: It's a good idea to not rely on stack-ids, that's true, but ultimately it's the UI type which shouldn't appear in non-legacy but-apis.

Regarding stack-id, right now we have no universal way to refer to a stack if it has no name, except for by its stack id… if it has one.
The only universal thing is to identify a stack by its index in the parent list of the workspace merge commit, or its node-id in the but-graph. 

Maybe that's ok though, to assume ref-names to be present and help the user get a ref-name when that's not the case, as exception.

PR #11641, opened 2025-12-28
Move `but snapshot` to `but oplog snapshot`

schacon:
Group oplog functionality under a new "but oplog" platform with subcommands (list, snapshot) so the default "but oplog" runs the list action. 

I also added a `--snapshot` or `-s` filter so that it's easy to only show the manually made snapshots and implemented it with an `include` filter rather than pulling a bunch of oplog entries and filtering them after the fact.

I also updated the `--since` filter to actually look for the starting point and start the walk from there, rather than the very inefficient way we were doing it.


krlvi (approved):
Nice

krlvi:
@schacon force pushed to fix conflicts here

PR #11644, opened 2025-12-28
rename `but review` to `but pr`.

schacon:
there should be `but pr new (<branch-id>)` to open a new PR. just `but pr` should default to `but pr new` and prompt for which branch to open a pr on. if there is only one branch without a pr, it should prompt "do you want to open a new PR on branch [name]?".

currently it opens two tempfiles - one for the title and one for the description. reduce this to a single tempfile call and extract the first line as the title and the rest as the description. add comments into the tempfile to explain this

also made the output nicer:

CleanShot 2025-12-28 at 09 43 04@2x

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -754,6 +754,7 @@ dependencies = [
  "posthog-rs",
  "regex",
  "rmcp",
+ "schemars 1.1.0",
  "serde",
  "serde_json",
  "shell-words",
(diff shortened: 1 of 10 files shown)


krlvi:
Force pushed to resolve merge conflicts @schacon

krlvi (approved):

PR #11645, opened 2025-12-28
add `but squash` command with AI message support

schacon:
Introduce a new subcommand "squash" that lets users squash two or more contiguous commits in the same stack and uses an AI provider to generate a combined commit message.


krlvi:
implemented this here https://github.com/gitbutlerapp/gitbutler/pull/11994

PR #11647, opened 2025-12-28
Make 'modern' `but` builds work again

Byron:
The idea is to port more and more commands over explicitly. Some inconvenience on maintaining the modern build is expected and maybe even a useful motivation to get it done more quickly, or to make it an explicit effort.


Byron:
Hi @krlvi, I thought it's best if you review and merge this PR just to double-check this is still the direction we want to go. To me it's the right thing for sure, but maybe there are other ways?

Something I mentioned in another PR is that it should be fine to allow `but_core::ui` types in non-legacy but as long as we get it into `but-api`, which should be totally possible. Technically, the `ui` types shouldn't show up in `but-api`, but they are usually just a small refactor away from *not* being exposed in `but-api` directly in favor of standard Rust types.

PR #11650, opened 2025-12-28
Fix failing UI test for HardWrapPlugin

PavelLaptev:
Should fix this https://github.com/gitbutlerapp/gitbutler/actions/runs/20560964264/job/59051507942


Byron:
The linked test is a funky one and I noticed it before as well. What struck me is that the text that ends up in the textbox is garbled, and appears cut-off in the middle so 'additional' doesn't fully appear even though the sentence was typed in completely.

It's like the editor has a life on its own and rejects some input. This could be some sort of race as well.
Anyway, I hope this fixes it, but I have some doubts.

PR #11652, opened 2025-12-29
Make 'modern' `but` builds work again

Byron:
The idea is to port more and more commands over explicitly. Some inconvenience on maintaining the modern build is expected and maybe even a useful motivation to get it done more quickly, or to make it an explicit effort.

Follow-up on accidentally closed #11647


krlvi (approved):
lgtm

PR #11654, opened 2025-12-29
fix: prevent skipping single hunk assignments in uncommitted files

krlvi:
@jonathantanmy2 I am 99% sure this is a bug, so fixing it here. Let me know if i missed something

### DIFF
--- a/crates/but/src/id/mod.rs
+++ b/crates/but/src/id/mod.rs
@@ -276,7 +276,7 @@ impl IdMap {
         }
 
         for uncommitted_file in self.uncommitted_files.values() {
-            if uncommitted_file.hunk_assignments.len() > 1 {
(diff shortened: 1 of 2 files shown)


jonathantanmy2:
> @jonathantanmy2 I am 99% sure this is a bug, so fixing it here. Let me know if i missed something

Ah actually this was intended, although I admit that it might be confusing. I wrote my rationale in the commit message of the commit 0ecd45c96 (id: unify uncommitted file and hunk IDs, 2025-12-24).

krlvi:
Oh I see - i never saw that, it makes sense. 

I noticed this while implementing `but diff` since the hunk needed an id. We could modify the code to use the hunk from the file if it's a sole hunk.

PR #11658, opened 2025-12-29
fix but status files reporting wrong Cli IDs

krlvi:
fyi @jonathantanmy2 - it turns out that the issue related to but status --files showing incorrect ids is the fact the the files are somehow being added twice to the set


jonathantanmy2:
Ah, good catch!

PR #11659, opened 2025-12-29
but diff: improve hunk rendering

krlvi:


Byron:
Even though this command is still legacy, it's new and I would have expected a couple of tests to show its output. These should be easy to do with the current test support, and based on existing tests.

It would be good to learn what stopped you so this barrier can be removed.

krlvi:
Perhaps it's the fact that I am sure if this even the output format that we want to display here, i just wanted to test how this feels. I am not sure if i like showing the file name for each hunk for instance

Byron:
It's the early days for `but diff`, I see.
Then the next best time for testing is once it goes 'modern', and I am looking forward to that day, maybe I will make it happen even, I still have a feeling it shouldn't be too hard.

PR #11664, opened 2025-12-30
refactor(status): Remove mutable references to IdMap

krlvi:
Follow up from https://github.com/gitbutlerapp/gitbutler/pull/11658

### DIFF
--- a/crates/but/src/command/legacy/status/mod.rs
+++ b/crates/but/src/command/legacy/status/mod.rs
@@ -196,7 +196,7 @@ pub(crate) async fn worktree(
             review,
             ctx.legacy_project.id,
             &repo,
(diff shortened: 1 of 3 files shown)


crates/but/src/command/legacy/status/mod.rs
```diff
@@ -476,7 +476,6 @@ pub fn print_group(
             }
         }
     } else {
-        id_map.add_file_info_from_context(ctx, None)?;
```
krlvi, line 479: Seems like this was the extra mutation that was adding the same files multiple times. https://github.com/gitbutlerapp/gitbutler/pull/11658 made it so that no multiple entries are added to the set, but this also removes the additional mutation entrierly

PR #11685, opened 2026-01-04
demo of failure to create a branch with but-rebase

jonathantanmy2:
@Caleb-T-Owens do you know if it's possible to create a branch (with no local commits) with the rebase engine? I did what I thought was the most obvious way, first selecting the upstream commit, but it didn't work as you can see in the test.

DO NOT SUBMIT
until I figure out how to create a branch with the rebase engine

This is a demo of what happens when I try to create a branch with the rebase engine with what I expect should work. I can see that the rebase engine doesn't recognize the upstream commit. I thought of changing the code of the rebase engine to recognize the upstream commit myself, but doing this would mean that the rebase engine needs to distinguish between immutable commits (that can be inserted-above, but not replaced or inserted-below) and mutable commits (like the ones we have now), so I wanted to check if this is part of the intended design before proceeding. If not, how do I create a new branch that has no local commits?

The below part of the commit message is probably done:

This commit includes a new way to construct a testing sandbox that simulates a fresh clone of a repo. The setup is done in Rust code instead of Bash fixture scripts because Bash fixture scripts are per-crate (and thus a test writer/reader has new scripts to read whenever they work on a new crate). The abstraction chosen is a fresh clone, which is easily understood, and the most relevant information (upstream commit and workspace commit) is concisely explained in the Rustdoc.

As of this commit, this is only used to test `but-api`. Using it to test other crates probably should wait until `but-api` and/or `but-rebase` is more fully fleshed out - right now, setting up non-trivial commit graph scenarios is probably much more verbose than in a Bash fixture script.


crates/but-api/src/branch.rs
```diff
@@ -71,3 +72,25 @@ pub fn branch_diff(ctx: &Context, branch: String) -> anyhow::Result<TreeChanges>
...
+/// of a branch, the existing branch will be stacked above the new branch.
+#[but_api]
+#[instrument(err(Debug))]
+pub fn new_only(
+    ctx: &but_ctx::Context,
+    refname: gix::refs::FullName,
+    commit_id: gix::ObjectId,
+) -> anyhow::Result<()> {
+    let mut guard = ctx.exclusive_worktree_access();
+    let (repo, _, graph) = ctx.graph_and_meta_mut_and_repo_from_head(guard.write_permission())?;
```
Byron, line 86: One would have to use `but_workspace::branch::create_reference()` as it will handle meta-data correctly, too. Only with it will you currently be able to create actual virtual branches, i.e. workspace branches with defined order despite them pointing to the same commit.

crates/but-api/tests/branch.rs
```diff
@@ -0,0 +1,23 @@
+use but_api::branch::new_only;
+use but_testsupport::Sandbox;
+
+#[test]
+fn new_basic() -> anyhow::Result<()> {
+    let env = Sandbox::simulate_clone()?;
```
Byron, line 6: Even though one can probably say that more tests are better, I also think that testing in `but-api` *and* in `but-cli` leads to a lot of redundancy. 

This is because the plumbing crates (i.e. `but-workspace`) are well-tested, and `but-api` is just a thin wrapper around plumbing invocations with the occasional side-effect, like oplog or undo-queue handling.

Thus I wouldn't encourage to test on `but-api` level unless it's something unique to it, and something we deem valuable enough to pin down.

crates/but-testsupport/src/sandbox.rs
```diff
@@ -60,6 +61,53 @@ impl Sandbox {
...
+    /// upstream commit and workspace commit can be obtained by passing
+    /// `remotes/origin/main` and `gitbutler/workspace` respectively to
+    /// [gix::Repository::rev_parse_single].
+    pub fn simulate_clone() -> anyhow::Result<Sandbox> {
+        let env = Sandbox::empty()?;
+        env.invoke_git("init");
+        env.invoke_git("remote add origin ./fake/local/path");
+        env.invoke_git("config user.name User");
+        env.invoke_git("config user.email user@example.com");
+        let repo = env.open_repo()?;
```
Byron, line 74: I think this would be simpler if it was just a shell script, which then would also benefit from automatic caching, i.e. if it doesn't change, it's output can just be copied over into a writable location if that's requested.

Probably the reason for not using it here is that it's dependant on the crate that invokes the function to create/update the fixture. **However, what would already work is to pass an absolute path to the script, which could be made to always find the fixture script in `but-testsupport`.**

I'd be happy to provide an alternative, that could allow `gix-testtools` to receive a script as string as well.

Byron (commented):
Thanks a lot for sharing! 

My highlight from the comments below would be the question if `clone_from()` wouldn't be better as a simple shell script. `gix-testtools` would already support **absolute paths** so the invocation can be crate-independent, but I could also add the ability to *pass raw strings* as fixtures.

Generally, I think the way this *could* go is that there will be multiple test-styles - those that differentiate between read-only and writable, and those which don't. Those which don't will write their fixtures inline with Rust that even when only doing Git things will be more verbose than a shell script with `git` commands. Those that do differentiate only do touch-ups in Rust, for instance to set metadata, and do so in memory only so tests stay read-only if they are read-only.

Lastly, I fear that doing more than a touch-up of a fixture in Rust will lead to longer and wordier test cases which then get copy/pasted around, making it hard to recognise that these are all using the same fixture. Of course, this duplication can be organised away, but I am not sure that this will happen.

PS: I set this PR to draft mode based on the `DO NOT SUBMIT` note in the PR description.

jonathantanmy2:
> My highlight from the comments below would be the question if `clone_from()` wouldn't be better as a simple shell script. `gix-testtools` would already support **absolute paths** so the invocation can be crate-independent, but I could also add the ability to _pass raw strings_ as fixtures.

I'll answer this under your comment on `sandbox.rs`.

> Generally, I think the way this _could_ go is that there will be multiple test-styles - those that differentiate between read-only and writable, and those which don't. Those which don't will write their fixtures inline with Rust that even when only doing Git things will be more verbose than a shell script with `git` commands. Those that do differentiate only do touch-ups in Rust, for instance to set metadata, and do so in memory only so tests stay read-only if they are read-only.

What do you mean by "read-only" and "writable"? (E.g. a test like "status" that does not mutate the repository versus a test like "branch new" which mutates the repository?)

> Lastly, I fear that doing more than a touch-up of a fixture in Rust will lead to longer and wordier test cases which then get copy/pasted around, making it hard to recognise that these are all using the same fixture. Of course, this duplication can be organised away, but I am not sure that this will happen.

I think that the responsibility is on the programmer, if they copy and paste, to remove the parts that are irrelevant to the test, so hopefully the test cases won't be so wordy.

What is the benefit of recognizing that multiple tests use the same fixture? I would have thought that multiple tests using the same fixture may indicate that the fixture is one-size-fits-all, and contains details that relevant to only one of those multiple tests.

> PS: I set this PR to draft mode based on the `DO NOT SUBMIT` note in the PR description.

Thanks.

crates/but-api/tests/branch.rs
jonathantanmy2: OK, it makes sense to not test `but-api` if it's a thin wrapper, but then would it make sense to combine `but-workspace` and `but-api` into one crate? I only looked at `reword`, but for that, both `but-api` and `but-workspace` were thin wrappers around the rebase engine.

crates/but-api/src/branch.rs
jonathantanmy2: Hmm...shouldn't the rebase engine be able to handle ref metadata? From what I see of it, a ref can be a `Step` (i.e. a node of the graph), and such references can be the "from" and "to" of a replacement (`Editor::replace` takes a `Selector`, which is a `StepIndex`, and a `Step`, which can be a ref).

crates/but-testsupport/src/sandbox.rs
jonathantanmy2: By absolute path, do you mean a path from the root of the repo, as opposed to from the root of a crate? If yes, that would work (not sure if CI etc. is per-crate, but we can check).

A script as a string would work too, of course.

I was worried about slow test performance (spawning `git` processes to do every little thing instead of calling `gix` functions in-process) but the automatic caching will definitely more than compensate for it.

crates/but-api/src/branch.rs
Byron: It definitely should. But while it doesn't, do you think it's preferable to have an implementation in `but-api` that can't do a certain thing, knowing there is a plumbing function that is made to do that exact thing?
Here I'd wonder if that new `but-api` function can be used by the CLI to remove a legacy call, and then I'd want it to be correct.
There is already [an implementation in `but-testing`](https://github.com/gitbutlerapp/gitbutler/blob/091268b910cebb8196edd74e451e27b32c43e8bb/crates/but-testing/src/command/mod.rs#L608-L635) that calls the plumbing directly.

crates/but-api/tests/branch.rs
Byron: Yes, I also think that the rebase engine in `but-rebase` should have all the tests, and that thin wrappers should be going into `but-api` and completely bypass `but-workspace`. It's tests, the last time I checked, don't add anything that isn't tested in `but-rebase` yet.

crates/but-testsupport/src/sandbox.rs
Byron: > By absolute path, do you mean a path from the root of the repo, as opposed to from the root of a crate? If yes, that would work (not sure if CI etc. is per-crate, but we can check).

Yes, anything that starts with `/` basically. This works because `my_root.join("/foo")` will always be `/foo`.

> A script as a string would work too, of course.

Let's try to the absolute path version then.

> I was worried about slow test performance (spawning git processes to do every little thing instead of calling gix functions in-process) but the automatic caching will definitely more than compensate for it.

We already have an example for the limits of this technique in case of a [huge read-only](https://github.com/gitbutlerapp/gitbutler/blob/a91a245c485621cbb0c41846599a242821932bbd/crates/but-graph/tests/fixtures/scenarios.sh#L1-L0) scenario file that creates loads of initial states. It will take about 10 seconds to run if changes are made to it.
Writable fixtures now by API design are always single-repo-per-script, so the fixtures will be minimal, while read-only fixtures have to be split by hand once the pain-point is reached.

Byron:
> What do you mean by "read-only" and "writable"? (E.g. a test like "status" that does not mutate the repository versus a test like "branch new" which mutates the repository?)

Yes, that's it. Read-only tests re-use an existing fixture that is expected to not be altered. If tests were buggy and alter the fixture in observable ways, read-only tests might interfere with each other and fail, but I consider that a feature as it's clearly an implementation bug.

To me it's very important to keep fixtures read-only if they can for performance, but also to signal that we expect certain commands to not have observable side-effects. GitButler had issues with this, and I am looking at you, `list_virtual_branches` that would just create a branch when it thought it was necessary.

And what may seem like over-compensation really isn't, it's merely a reflection of `&` and `&mut` which we know is incredibly powerful in its own right, with plenty of immediate benefits like a chance of catching misbehaviour, and performance for fixture setup.

> What is the benefit of recognizing that multiple tests use the same fixture? I would have thought that multiple tests using the same fixture may indicate that the fixture is one-size-fits-all, and contains details that relevant to only one of those multiple tests.

It's true, it's what we have in `but-cli` right now, which probably indicates that the initial state doesn't matter too much as long as it's a workspace. The idea here would be to ask what initial states may affect the implementation, and create and use them accordingly.

For `but-cli`, I think it's fine to generalise such initial states, to say it's "a typical workspace" for example, or a "single-branch" situation. After all, everything `but-cli` calls is expected to work in all of these already so it's more about is adding more specialised initial states over time to reproduce specific bugs.

`but-graph` shows how it can also go, when only few initial states are re-used, and all are very specific to their tests. Here it's entirely unclear if these could be deduplicated, and that wouldn't be the goal. The goal for `but-graph` was to be specific and the test-suite reflects that.

> I think that the responsibility is on the programmer, if they copy and paste, to remove the parts that are irrelevant to the test, so hopefully the test cases won't be so wordy.

I'd hope that if the initial state isn't so important, that it can be setup in a line or two, so we don't end up copy-pasting a lot of fixture initialisation that clutters tests and gets copy-pasted.

With that said, in `but-cli` it really looks like it's "typical simple workspace", "typical workspace with worktree changes", and for the journeys, it's "from-unborn" and "from-bare" essentially, and we should be able to run typical journeys from there to completion. Right now we can't.

So to me the problem to fix would have been to add the needed fixtures to have worktree changes, and it probably was me who triggered this by adding a convenience `env.file(…)` method which forces tests to be read-write no matter what. I think it came from acknowledging the fact that most legacy commands currently need `but init` anyway, but it's really not supposed to be that way.

And now that I am writing this and remembering that at least one command now drops a file that contains CliID mappings in a read-only command, the whole idea of read-only goes out of the window for the `but` CLI. The test-suite must of course reflect that, but I don't know if that should be how we do things (CC @krlvi).

But that also nudges me towards accepting that `but` CLI is always writing, and the test-suite reflects that.

crates/but-api/src/branch.rs
jonathantanmy2: If we needed this ability in `but-api` right now, I agree that we should do what you suggest (use `but-workspace` directly) but my hope with this project was to understand the rebase engine more, so looks like I'll need to hold off on this work until the rebase engine understands ref metadata.

crates/but-api/tests/branch.rs
jonathantanmy2: Thanks, makes sense.

jonathantanmy2:
> > What is the benefit of recognizing that multiple tests use the same fixture? I would have thought that multiple tests using the same fixture may indicate that the fixture is one-size-fits-all, and contains details that relevant to only one of those multiple tests.
> 
> It's true, it's what we have in `but-cli` right now, which probably indicates that the initial state doesn't matter too much as long as it's a workspace. The idea here would be to ask what initial states may affect the implementation, and create and use them accordingly.
> 
> For `but-cli`, I think it's fine to generalise such initial states, to say it's "a typical workspace" for example, or a "single-branch" situation. After all, everything `but-cli` calls is expected to work in all of these already so it's more about is adding more specialised initial states over time to reproduce specific bugs.
> 
> `but-graph` shows how it can also go, when only few initial states are re-used, and all are very specific to their tests. Here it's entirely unclear if these could be deduplicated, and that wouldn't be the goal. The goal for `but-graph` was to be specific and the test-suite reflects that.

I think that the rebase engine tests will look more like `but-graph` than `but-cli` (each test has a different initial state).

How do you feel about the initial state being in a separate file (and not inline in the test itself)? I do find myself slightly annoyed by it (I have to cross-reference, and if I have to access an ID in the test, I presumably need to rev-parse it, when I could have assigned it to a variable if everything was in Rust), but if the reason is something like how it's less verbose to write in Bash, I guess I can live with that.

jonathantanmy2:
Besides the discussion above (thanks, by the way), I still have a question remaining about whether upstream commits should be selectable by the rebase engine. This is not only so that we can create new branches with no local commits, but also for things like rebasing/cherrypicking a commit directly onto an upstream commit.

Caleb-T-Owens:
Hey! Hopfully the following can shed some light on what is going on. Let me know if there are any portions I can clarify further. I'd be happy to hop on a call and go over any of it.

The rebase engine internally contains a graph of "steps" of all of the commits and references that are reachable by the workspace HEAD, similar to what a `git rebase -i --rebase-merges` might look like.

It's often not _all_ of the reachable commits from HEAD. The but-graph traversal is limited, and will stop after ~300ish commits deep to try and avoid doing _too_ much work. Because we use the step graph to infer where the parents should be, in this case we use the `preserved_parents` boolean on the `Step::Pick` step which tells the rebase engine to ingnore the parents (or lack there of) inferred by the graph and to use the origional parents.

When we call `.rebase()` we are going over that step graph from the parent-most steps, performing rewrites if the parents have changed.

The code that you wrote would work to add an additional reference to a commit that is reachable from HEAD. 

There are a couple options we can use to insert a reference to a commit outside of the `Editor`'s scope.

## Creating a rebase editor that does include these additional scopes

Currently the traversal is more limited in scope. The initial step-graph creation could be expanded to optionally pull more information from the `but-graph`. We could even have a way of taking a revwalker and turning that into an editor, though the but-graph should contain _everything_ in recent history.

Initially I chose to keep the scope limited since most operations are only manipulating commits 

## Add the ability to create a new nodes and edges

The rebase engine will try and rebase any nodes in it's step graph, so if we did want to add a reference to something not in the step graph, we could use an interface as follows:

```rs
// This would add a new pick step to the graph with `preserved_parents` set to true.
let commit_selector = editor.add_pick_to_unstepped_commit(target_commit_id)?;
editor.insert(
    commit_selector,
    Step::Reference { refname },
    InsertSide::Above,
)?;
```

Caleb-T-Owens:
I do like the idea of this `Sandbox` testing structure. Having worked with the bash based testing for a while now, I still struggle to find corresponding fixtures to the tests and still find bash code hard to read and write. I find some of the helper functions in the `but-workspace` quite hard to follow.

I'm quite a fan of having the setup code being in the test case itself so it's easy to see what the test is operating on easily.

With reguards to the read-only vs write thing, I always clear out the fixtures between test re-runs anyway to avoid stale fixture issues that I've had in the past, with a negligible performance impact.

Byron:
> How do you feel about the initial state being in a separate file (and not inline in the test itself)? I do find myself slightly annoyed by it (I have to cross-reference, and if I have to access an ID in the test, I presumably need to rev-parse it, when I could have assigned it to a variable if everything was in Rust), but if the reason is something like how it's less verbose to write in Bash, I guess I can live with that.

If I share my 'tricks' and it still isn't less annoying, then I think it's best to put that topic to rest and say do what you must to be happy 😁. What's there now makes me happy, but mileage varies.

What this system does is to give a place on disk with the fixture for read-only *and* (most) writable fixtures alike. This is really handy for manual investigation, but also to understand the initial state. Shell out a `git log` and capture the output, and instantly get an idea of the graph and the stable hashes that go with it. Not rerunning things that don't change is also very helpful, especially if there are many tests. Slow tests are an impediment and it's only acceptable if there aren't many tests (like is the case for most crates).

I haven't seen that ease-of-use with fixtures that all go into a tmp directory, and usually I had to add utilities or inline code to keep such tmp dirs to be able to inspect them myself. It's hard for me to grasp how it's possible to do this without such basic functionality.

With initial states, I'd never want to see how it's done, but I want to see what it is. Seeing what is done is a cognitive burden as one has to simulate in your head what the state is like - it really isn't a substitute for showing the initial state that matters to the code under test.

Lastly, if the implementation language or the lack of ease of 'jump to definition' that should be changed, then maybe something can be done.
The implementation language is already controlled by the shebang.
And adding native Rust support might also be possible with some macro-magic maybe. That way, the "run read-only/write-by-copy things only once until the script changes"-feature wouldn't be lost, while adding a jump-to capability in the IDE. Doing so doesn't sound easy or straightforward though 😅.

Byron:
One more note, now that I think about it: The `cargo` test-suite could be a reference. It's declarative, and the declaration of the initial state is basically the initial state. That's also perfectly fine, and the `Sandbox` approach would already help with that.

Maybe the way that is good mix of both approaches, skipping over the read-only capability as "performance optimisation", would be to have a bunch of high-level utilities on `Sandbox` to do the things we typically do. `Sandbox` already has a `debug()` function that allows to inspect the actual state, that's all I'd really want.

With that said, is there a conclusion to this PR that can be drawn so it could be merged? Something I'd still want to avoid is to do pure Git things in Rust that are just harder to read and maintain than a shell script.

jonathantanmy2:
> The code that you wrote would work to add an additional reference to a commit that is reachable from HEAD.

Thanks for your explanations. Ah, this was indeed what I missed - I didn't set HEAD to `gitbutler/workspace`.

jonathantanmy2:
> > How do you feel about the initial state being in a separate file (and not inline in the test itself)? I do find myself slightly annoyed by it (I have to cross-reference, and if I have to access an ID in the test, I presumably need to rev-parse it, when I could have assigned it to a variable if everything was in Rust), but if the reason is something like how it's less verbose to write in Bash, I guess I can live with that.
> 
> If I share my 'tricks' and it still isn't less annoying, then I think it's best to put that topic to rest and say do what you must to be happy 😁. What's there now makes me happy, but mileage varies.

OK, thanks for the discussion and yes, let's put the topic to rest. With absolute paths, I think that the large majority of my concerns are solved anyway.

I used:
```
format!(                                                 
    "/{}/../../crates/but/tests/fixtures/scenario/{}.sh",
    env!("CARGO_MANIFEST_DIR"),                          
    name                                                 
)
```
but there may be a better way.

jonathantanmy2:
> With that said, is there a conclusion to this PR that can be drawn so it could be merged? Something I'd still want to avoid is to do pure Git things in Rust that are just harder to read and maintain than a shell script.

I sent this PR to discuss 2 things (test strategy and why the rebase engine can't create the reference), both of which are resolved (the latter is because I didn't set HEAD to `gitbutler/workspace`), so I think the conclusion is that we can close this PR (without merging). I still have questions about the rebase engine, but that is separate from this PR and I'll ask them on Discord.

PR #11699, opened 2026-01-05
gerrit: fallback review URL when push has no new changes

### LINKED ISSUE #10188: Gerrit forge integration

godnight10061:
Fixes #10188

When pushing to Gerrit, if a commit was already pushed outside GitButler, Gerrit may reject with `(no new changes)`. GitButler treats this as success, but the push output contains no change URLs, so no Gerrit metadata was recorded and the UI kept the commit as 'unpushed' / missing the CR link.

This adds a fallback in `but-gerrit` to derive the Gerrit host from the repo push remote and generate a review URL from the commit's GitButler change UUID (the same `I<sha1(uuid)>` `Change-Id` used in trailers), e.g. `https://<gerrit-host>/q/<change-id>`.

**Tests**
- `cargo test -p but-gerrit`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1108,11 +1108,13 @@ dependencies = [
  "but-core",
  "but-ctx",
  "but-db",
+ "but-testsupport",
  "chrono",
  "gitbutler-commit",
  "gix",
  "serde",
  "sha-1",
+ "tempfile",
  "uuid",
 ]
 
--- a/crates/but-gerrit/Cargo.toml
+++ b/crates/but-gerrit/Cargo.toml
@@ -8,7 +8,6 @@ rust-version.workspace = true
 
 [lib]
 doctest = false
-test = false
 
 [dependencies]
 but-core.workspace = true
@@ -24,3 +23,11 @@ bstr.workspace = true
 chrono.workspace = true
 serde.workspace = true
 gix = { workspace = true }
+
+[dev-dependencies]
+
+but-testsupport.workspace = true
+
+gix = { workspace = true, features = ["revision"] }
+
+tempfile = "3"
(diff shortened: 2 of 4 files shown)


Byron (commented):

And while I would have a thing or two to say about the test (which I would probably just refactor myself), I lack experience with Gerrit to understand the problem the PR is solving in the first place.

Maybe @krlvi could see if this pans out (and maybe even try it) beforehand. Alternatively, and independently, maybe @godnight10061 could make a before and after video to show how the fix translates to the real world?

Thanks again.

godnight10061:
> Maybe @krlvi could see if this pans out (and maybe even try it) beforehand. Alternatively, and independently, maybe @godnight10061 could make a before and after video to show how the fix translates to the real world?
> 
> Thanks again.

https://github.com/user-attachments/assets/bc973da9-cb7a-4ef7-9620-5352f68d84eb

https://github.com/user-attachments/assets/abce04eb-fddc-4f8e-b908-63eb25637edf

How to reproduce:
  - In GitButler: Project Settings → General → “Gerrit Configuration” → toggle ON.
  - In GitButler: make a tiny change and create a commit in GitButler (so it has a GitButler change id).
      - Before (bug): push “succeeds”, but the commit still shows Local (circle) and Push stays enabled (still thinks
        there are unpushed commits).
      - After (fixed): commit flips to Local and remote (diamond) and Push disables (“No commits to push”).
        (Depending on URL format you may or may not also see a Gerrit pill/link.)
image

krlvi (approved):
Hey, thank you for this contribution. This fallback is indeed helpful

godnight10061:
Thanks for merging! I'm glad the fallback will be useful.

PR #11703, opened 2026-01-06
id: support filenames as CLI IDs

jonathantanmy2:
I wrote some TODOs in the code - if anyone has any ideas, let me know. I thought about them for a while but couldn't come up with solutions (except complicated solutions that are probably not worth the complication).

Copy and paste commit message:

Interpret them to refer to the uncommitted, unassigned file (uncommitted files that are already assigned to a stack are not included).

There are caveats to this approach, as documented in the TODOs, but this affects the rare case that a repo has files in its root directory whose names are short, so it seems worthwhile to implement this feature, especially since it's natural for a user to specify a filename where a CLI ID is required.


jonathantanmy2:
I think the failing test is because of something else other than my code - when it's fixed, I'll rebase and repush.

Byron:
I have restarted CI to see if it's flake. But even if it is, I am very surprised this database migrations runs into concurrency issues in the `but` CLI tests. This could point at something fundamentally wrong, but also, could lead to a fix that makes migrations as run from the UI much more likely to succeed.

[The log, copied for convenience](https://github.com/gitbutlerapp/gitbutler/actions/runs/20737352620/job/59537347900?pr=11703#step:5:174).
```
---- expected: tests/but/journey.rs:98:20
++++ actual:   stderr
   1      - Error: errors.projects.default_target.not_found
   2      - 
   3      - Caused by:
   4      -     there is no default target
        1 + Error: Failed to run migrations: Failed to run 2025-05-26-145725_hunk_assignments with: table `hunk_assignments` already exists
```

*If it fails again, maybe it's a broken migration, but that should be caught by tests*.

crates/but/src/id/tests.rs
```diff
@@ -567,6 +567,54 @@ fn ids_are_case_sensitive() -> anyhow::Result<()> {
...
+        Ok(if commit_id == id(1) && parent_id.is_none() {
+            vec![]
+        } else {
+            bail!("unexpected IDs {} {:?}", commit_id, parent_id);
+        })
+    };
+    let hunk_assignments = vec![hunk_assignment("foo", None)];
+    id_map.add_file_info(changed_paths_fn, hunk_assignments)?;
+
+    insta::assert_debug_snapshot!(id_map.resolve_entity_to_ids("foo")?, @r#"
```
Byron, line 589: I think it would help a lot to have more of an explanation what's under test there. What's important about the return value, what makes it special?

Unfortunately, `insta` doesn't support the 'third argument' in its assertions, so all we can do here is add an ordinary comment above it.

Byron (commented):
I don't anything to add to the TODOs, but regarding ambiguous ids, one can only ever ensure that IDs generated now don't clash with any IDs that were already generated. Order matters, and entities added first will claim their IDs no matter what.
Some of that might not matter though once we know the access patterns, so postponing this seems like the right choice.

Regarding tests, I hope that we can make them so that they carry as much information as possible about what they actually test, knowing the future you won't know the details anymore, nor does a 'current me'. The more it's spelled out, the more likely future changes will do the right thing once the test fails, having to determine if the failure is legit or needs its expectation changed. This goes in the direction of making the code maintainable by a growing set of people, while acknowledging tests as only 'hard' defence to regression.

crates/but/src/id/tests.rs
jonathantanmy2: Done.

jonathantanmy2:
Besides @Byron 's comment, I've also expanded the comment that Copilot mentioned.

PR #11709, opened 2026-01-06
move some output to progress so it's immediately printed

schacon:
This PR moves some of the output to the `progress` channel, so before network requests we can let the user know that we're making network requests.

https://github.com/user-attachments/assets/35999714-3065-402f-be34-5070f7db28a9


krlvi:
@schacon this is fucking cool

PR #11721, opened 2026-01-06
More changes to make CLI IDs less ambiguous

jonathantanmy2:
I'll try something new and will describe here why I didn't automerge this PR.

Why I didn't automerge
- In the first commit ("tests: replace TODOs with hunk ID verification") I decided to include the expected CLI output as an inline string instead of the typical SVG file. I'm bringing this up here because it's not usual practice, although in this case, I think it's warranted because a reader needs to know that a specific string is present (which will be used later in the test). I added `env_remove("CLICOLOR_FORCE")` to remove the color codes to make the CLI output easier to read.
- In the last commit ("id: rename unassigned area to zz") I decided to switch the ID of the unassigned area to "zz" (before this PR, it was "00"). I think this is a good change, but it may affect people's workflows, so I'm bringing this up here.

This PR reduces ID ambiguity by making branch IDs not collide with filenames, and by renaming the unassigned area to not collide with commits that start with "00".

### DIFF
--- a/crates/but/src/command/legacy/absorb.rs
+++ b/crates/but/src/command/legacy/absorb.rs
@@ -34,8 +34,8 @@ pub(crate) fn handle(
     out: &mut OutputChannel,
     source: Option<&str>,
 ) -> anyhow::Result<()> {
-    let mut id_map = IdMap::new_from_context(ctx)?;
-    id_map.add_file_info_from_context(ctx, None)?;
+    let mut id_map = IdMap::new_from_context(ctx, None)?;
+    id_map.add_committed_file_info_from_context(ctx)?;
     let source: Option<CliId> = source
         .and_then(|s| parse_sources(ctx, &id_map, s).ok())
         .and_then(|s| {
--- a/crates/but/src/command/legacy/branch/mod.rs
+++ b/crates/but/src/command/legacy/branch/mod.rs
@@ -68,8 +68,8 @@ pub fn handle(
             branch_name,
             anchor,
         }) => {
-            let mut id_map = IdMap::new_from_context(ctx)?;
-            id_map.add_file_info_from_context(ctx, None)?;
+            let mut id_map = IdMap::new_from_context(ctx, None)?;
+            id_map.add_committed_file_info_from_context(ctx)?;
             // Get branch name or use canned name
             let branch_name = branch_name.map(Ok).unwrap_or_else(|| {
                 but_api::legacy::workspace::canned_branch_name(ctx.legacy_project.id)
--- a/crates/but/src/command/legacy/commit.rs
+++ b/crates/but/src/command/legacy/commit.rs
@@ -20,8 +20,8 @@ pub(crate) fn insert_blank_commit(
     out: &mut OutputChannel,
     target: &str,
 ) -> Result<()> {
-    let mut id_map = IdMap::new_from_context(ctx)?;
-    id_map.add_file_info_from_context(ctx, None)?;
+    let mut id_map = IdMap::new_from_context(ctx, None)?;
+    id_map.add_committed_file_info_from_context(ctx)?;
 
     // Resolve the target ID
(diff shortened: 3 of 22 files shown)


krlvi:
I wonder... since we get to pick our "magical id", perhaps zz is more difficult to press (with the pinky finger). 
What if we used `uu` (as unassigned)? @schacon wdyt

jonathantanmy2:
`uu` would work too - let me know what you think.

schacon:
No, I think 'zz' is better than 'uu', I feel like it would be difficult to remember, especially if you're not a native english speaker. I honestly prefer '00' but I understand that the way you guys have approached it by using the first two chars of the SHA results in possible collisions. However, is it necessary to _always_ use the first two char of the SHA? if the first two char of the SHA is `00`, we can still assign a different short code, no? I think in my original implementation, I didn't use the SHA, I used a char range outside of 0-f for commits too. Do we have that hardcoded somewhere, that if it's in hex range we assume they're talking about a commit?

jonathantanmy2:
Thanks, Scott.

> However, is it necessary to _always_ use the first two char of the SHA? if the first two char of the SHA is `00`, we can still assign a different short code, no?

Right now we don't do any reassignment of commit CLI IDs, even if one collides with another. If we did that, yes we could un-collide commits that start with "00" by moving them to another ID, just like we would un-collide commits that start with the same 2 hex characters by moving all but one to another ID. But I think it's more natural to instead extend the length of the ID, which was one of the things I was planning to do (so a "00" commit would then be extended to 3 characters in order to un-collide it with the unassigned area).

Having said that, though, it seems strange to me to special-case "00" like this and would rather use "zz" or "uu".

> I think in my original implementation, I didn't use the SHA, I used a char range outside of 0-f for commits too. Do we have that hardcoded somewhere, that if it's in hex range we assume they're talking about a commit?

No, it's not hardcoded anywhere. (Previously it wasn't hardcoded since "00" looks like a SHA but is the unassigned area, and I didn't add any hardcoding in this PR.)

schacon:
If everyone feels strongly about it, my preference would be zz

krlvi (approved):
alright, solid arguments. From my perspective we are good to go, and zz is just fine :)

jonathantanmy2:
Thanks @krlvi . There is still the issue of how I wrote the test, but maybe people don't care about that :) (and if people do, feel free to mention it and I will change it).

Byron:
> * In the first commit ("tests: replace TODOs with hunk ID verification") I decided to include the expected CLI output as an inline string instead of the typical SVG file. I'm bringing this up here because it's not usual practice, although in this case, I think it's warranted because a reader needs to know that a specific string is present (which will be used later in the test). I added `env_remove("CLICOLOR_FORCE")` to remove the color codes to make the CLI output easier to read.

I think this is the way to go for most tests.

The SVG output isn't something we'd want to use everywhere, we just use it to pin 'graphics'. For everything else, we are better off with strings.

Maybe we'd want to invert how coloring is working as well to emphasise this, i.e. force color rather than remove it. The tests for coloring could even have their own names to make that clear and establish a pattern.

Does that resonate with you @jonathantanmy2? And if so, is this something you could help get done if it's quick enough? Otherwise I'd try to pick it up myself and try to get it done soon as SVGs as used now have been an impediment, so we'd want only those that really validate something that's important to us.

jonathantanmy2:
> Maybe we'd want to invert how coloring is working as well to emphasise this, i.e. force color rather than remove it. The tests for coloring could even have their own names to make that clear and establish a pattern.
> 
> Does that resonate with you @jonathantanmy2? And if so, is this something you could help get done if it's quick enough? Otherwise I'd try to pick it up myself and try to get it done soon as SVGs as used now have been an impediment, so we'd want only those that really validate something that's important to us.

Yes this resonates with me a lot. There weren't many to convert, so I've sent #11751 which I think does what you suggested - PTAL.

PR #11726, opened 2026-01-07
Add `but resolve` edit-mode workflow

schacon:
Introduce a new `but resolve <commit>` command and accompanying subcommands (status, finish, cancel) to enter an edit-mode resolution flow for conflicted commits. 

Status warns you of edit state:

CleanShot 2026-01-07 at 11 37 39@2x

`but resolve status` tells you which files are still outstanding:

CleanShot 2026-01-07 at 11 40 11@2x

Finishing a resolution:

CleanShot 2026-01-07 at 11 40 48@2x

Help output:

CleanShot 2026-01-07 at 11 41 10@2x

### DIFF
--- a/crates/but-api/src/legacy/modes.rs
+++ b/crates/but-api/src/legacy/modes.rs
@@ -76,6 +76,7 @@ pub fn abort_edit_and_return_to_workspace(project_id: ProjectId) -> Result<()> {
     Ok(())
 }
 
+// GUI-facing API that returns () for serialization compatibility
 #[but_api]
 #[instrument(err(Debug))]
 pub fn save_edit_and_return_to_workspace(project_id: ProjectId) -> Result<()> {
@@ -83,6 +84,14 @@ pub fn save_edit_and_return_to_workspace(project_id: ProjectId) -> Result<()> {
 
     gitbutler_edit_mode::commands::save_and_return_to_workspace(&ctx)?;
 
+    Ok(())
+}
+
+#[but_api]
+#[instrument(err(Debug))]
+pub fn save_edit_and_return_to_workspace_with_output(project_id: ProjectId) -> Result<()> {
+    let ctx = Context::new_from_legacy_project_id(project_id)?;
+    gitbutler_edit_mode::commands::save_and_return_to_workspace(&ctx)?;
     Ok(())
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -742,6 +742,8 @@ dependencies = [
  "gitbutler-branch",
  "gitbutler-branch-actions",
  "gitbutler-commit",
+ "gitbutler-edit-mode",
+ "gitbutler-operating-modes",
  "gitbutler-oplog",
  "gitbutler-project",
  "gitbutler-reference",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -723,6 +723,7 @@ dependencies = [
  "but-meta",
  "but-oxidize",
  "but-path",
+ "but-rebase",
  "but-rules",
  "but-secret",
  "but-serde",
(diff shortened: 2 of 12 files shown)


crates/but-api/src/legacy/modes.rs
```diff
@@ -76,14 +76,17 @@ pub fn abort_edit_and_return_to_workspace(project_id: ProjectId) -> Result<()> {
     Ok(())
 }
 
-#[but_api]
+// Note: Not using #[but_api] here because RebaseOutput doesn't implement Serialize
+// This function is only called from the CLI resolve command, not from the GUI
```
Caleb-T-Owens: This is not true?

We use this command to save and exit edit mode in the GUI.

```ts
			saveEditAndReturnToWorkspace: build.mutation<void, { projectId: string }>({
				extraOptions: { command: 'save_edit_and_return_to_workspace' },
				query: (args) => args,
				invalidatesTags: [
					invalidatesList(ReduxTag.WorktreeChanges),
					invalidatesList(ReduxTag.HeadSha),
					invalidatesList(ReduxTag.HeadMetadata)
				]
			}),
```

schacon:
```
error[E0277]: the trait bound `RebaseOutput: serde::Serialize` is not satisfied
   --> crates/but-api/src/legacy/modes.rs:79:1
    |
 79 | #[but_api]
    | ^^^^^^^^^^ the trait `github::_::_serde::Serialize` is not implemented for `RebaseOutput`
    |
    = note: for local types consider adding `#[derive(serde::Serialize)]` to your `RebaseOutput` type
    = note: for types from other crates check whether the crate offers a `serde` feature flag
    = help: the following other types implement trait `github::_::_serde::Serialize`:
              &'a T
              &'a mut T
              ()
              (T,)
              (T0, T1)
              (T0, T1, T2)
              (T0, T1, T2, T3)
              (T0, T1, T2, T3, T4)
            and 1637 others
note: required by a bound in `to_value`
   --> /Users/schacon/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/serde_json-1.0.148/src/value/mod.rs:997:8
    |
995 | pub fn to_value<T>(value: T) -> Result<Value, Error>
    |        -------- required by a bound in this function
996 | where
997 |     T: Serialize,
    |        ^^^^^^^^^ required by this bound in `to_value`
    = note: the full name for the type has been written to '/Users/schacon/projects/gitbutler/target/debug/deps/but_api-e1e09edc9ee3b90f.long-type-1328564852220009337.txt'
    = note: consider using `--verbose` to print the full type name to the console
    = note: this error originates in the attribute macro `but_api` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Caleb-T-Owens: We probably want to make a serializable type that contains the subset of information that you are interested in. I'm not hugely excited about having the RebaseOutput type going too far, especially since I'm trying to get rid of it at the minute 😅

schacon: it's because I wanted to get some result data from this for showing options to the user. perhaps there is a better way to do this without messing up the macro. let me look into it

Caleb-T-Owens: Let me know if you'd like me to make a patch tomorrow for this

schacon: I actually think I should not try to get the RebaseOutput directly, I can just let it happen and then look through the branch to see if anything is conflicted after the fact. No reason to modify the API.

Caleb-T-Owens: Sounds good. LMK if there is anything I can do to be helpful :+1:

schacon:
@Caleb-T-Owens I completely removed any changes to the API.

PR #11729, opened 2026-01-07
Changes list and diffs preview redesign

mtsgrd:


mtsgrd:
@PavelLaptev there's a few more things I need to do here, but could you have a quick look at this and provide feedback if you have any?

mtsgrd:
@krlvi could you give this a quick spin and share your impressions? Pavel is ready to work on it if we agree this is the right direction.

krlvi:
I am running this locally and I have to say, at least in the first 30 mins... I really enjoy this.

You are right that the visual distinction between files is not very clear, this can probably be improved by tweaking the design

mtsgrd:
Alright, let's move ahead. @PavelLaptev I'll add the same behaviour for the unassigned view!

PavelLaptev:
> Alright, let's move ahead. @PavelLaptev I'll add the same behaviour for the unassigned view!

@mtsgrd cool! I'll continue to work in this PR then.

PavelLaptev:
I’ll also convert it to draft for now, so Copilot can stop commenting for a while.

mtsgrd:
@PavelLaptev sorry for force pushing, but I wanted to fix one thing and add highlighting as an example for you.

PavelLaptev:
@mtsgrd, no problem. I’ll likely begin working on this today in the evening.

mtsgrd:
@PavelLaptev I just added a commit here, should make the virtual list work reliably with diffs. Since I rebase my stuff I wasn't able to keep this master merge commit you had pushed.. should I turn off force push for this collaboration to work better?

PavelLaptev:
> should I turn off force push for this collaboration to work better?

Thank you for the fix!
Probably to avoid that friction, switching to standard merges is probably easier for tracking the history. But it's up to you.
There are some changes in Master that I needed, I'll merge Master into the branch tomorrow again.

mtsgrd:
@PavelLaptev would you be able to try these latest changes? It should behave better with improved performance. 🤞

PavelLaptev:
> @PavelLaptev would you be able to try these latest changes? It should behave better with improved performance. 🤞

Yes, it feels better. We can check it together on my machine again.

The images now look fine and load once.

Here are the images and diffs when I scroll:

https://github.com/user-attachments/assets/4acd1cc6-ae04-4184-a416-5e388f0e8bc9

https://github.com/user-attachments/assets/aeea79df-3416-41e8-bb45-1b9772cc53ab

With images and diffs, I notice that the “loading” state appears almost always when I scroll. I wonder if we can load things earlier.

krlvi:
This is starting to look and feel really sleek. I tried it with dev, and im now making a nightly build from this branch to play with it some more (https://github.com/gitbutlerapp/gitbutler/actions/runs/21013803145) 

Perhaps as a future improvement (and I suppose you have already thought about this) - how difficult would it be to make the selection in the file list on the left "follow" what is currently in view?

krlvi:
I am really excited to see this in it's final form, because it already is really helpful in browsing through changes.

With the understanding that the PR is not yet ready (in terms of matching the design from the conversation on discord https://discord.com/channels/1060193121130000425/1224366824305463337/1455169946576486451 ) I was thinking about some design related ideas here. So while we are figuring stuff out, I would like to drop two design ideas into the "ideas hat"... Here we go:

1) Would it make sense to create bigger visual contrast for the file breaks so that it is more clear at a glance when a new file begins (the parts that i circled in red). 

2) And perhaps, if there is a way to make the list of files selection update while scrolling (https://github.com/gitbutlerapp/gitbutler/pull/11729#issuecomment-3752204648), would it make sense for the selection to somehow visually connect to the file that is in preview focus? (the part that i circled in blue).
Screenshot 2026-01-14 at 16 27 36

PavelLaptev:
> Would it make sense to create bigger visual contrast for the file breaks so that it is more clear

Yes, according to the design, there will be foldable drawers with distinctive headers.

image

---

> if there is a way to make the list of files selection update while scrolling

Yes, this is possible. We can add a pointer or a mark there how we do it the sidebar menu or branch header.

mtsgrd:
@PavelLaptev just fyi, with virtual list items that have async content it's important we have some initial skeleton of a reasonable height. If we have e.g. just the file header of 42px, it means for a split second we might e.g. fit 20 items on screen before it adjusts and potentially hides 19 of them.

PavelLaptev:
@krlvi, there’s a question for “unassigned” or “assigned” panels: how should we close panels that only contain files?

image

Previously, there was a close button in the header. However, now we have multiple headers, and the button should be for the entire section.

Of course, we can close the section by clicking on the selected row again or by pressing Esc. However, we also need a button for this purpose (To convey the same pattern we have in other closable panels). I’m currently using a floating cross button for this.

image

https://github.com/user-attachments/assets/7fa8d9f8-62d3-437a-9344-f299dcb466aa

PavelLaptev:
> @PavelLaptev just fyi, with virtual list items that have async content it's important we have some initial skeleton

@mtsgrd I don't mind skeletons :-)

mtsgrd:
@krlvi about making the selection follow what's on screen. I had an example that I showed in the eng meeting, but I presented a few corner cases and confusing ux to Pavel and we'd love to ship this first and then tackle that as a follow-up item. Wdyt?

Perhaps selection could stay as is, but we have a different way of indicating what diffs are in view? I like Pavel's suggestion of a pointer or a mark like in the left sidebar.

krlvi:
cool, i dont mind either way. btw do we wish to make an app release before merging this PR? Or do we feel confident about even releasing?

PR #11742, opened 2026-01-07
Make renaming a branch to its current name a NOOP

### LINKED ISSUE #11741: Renaming a branch to the same name causes error
### Version

master

### Operating System

Linux

### Distribution Method

N/A

### Describe the issue

Hello!

Renaming a branch to the same name causes an error. I discovered this as I started renaming a branch, and then butter-fingered the `Rename` button instead of `Cancel` when I changed my mind about the whole thing.

The error message is as follows:

```
command: update_branch_name
params: {"projectId":"14d30fed-99c2-4055-a1c5-e4b950bf05b3","stackId":"07f38968-c2e9-4e78-847a-1d435c3b6eb0","laneId":"07f38968-c2e9-4e78-847a-1d435c3b6eb0","branchName":"test","newName":"test"})

A patch reference with the name test exists
```

In any lane, click `Rename` from the context menu and then immediately click the `Rename` button to submit.

It also triggers if you rename a branch to something that, when normalized, results in the same name. For example, renaming the branch `test` to `test-` (the trailing `-` is stripped during normalization).

I think this should be a NOOP. It's the simplest solution from a technical standpoint, makes branch renaming idempotent and doesn't hurt the user experience.

```shell
2026-01-07T17:32:40.857965Z ERROR update_branch_name: crates/but-api/src/legacy/stack.rs:189: error=A patch reference with the name test exists project_id=b1230fbd-d622-40e4-
bb9e-42cf354aca52 stack_id=07f38968-c2e9-4e78-847a-1d435c3b6eb0 branch_name="test" new_name="test"
```

slarse:
## 🧢 Changes
Makes renaming a branch to its current name a NOOP.

## ☕️ Reasoning
This seems like the simplest solution to the issue, and has no particular downsides. It also makes branch renaming idempotent, which at least in web dev is usually rather desirable (unless the NOOP is itself indicative of an error, but I don't think it is in this case).

I fixed this at the lowest level possible.

Renaming a real reference is actually already a NOOP https://github.com/gitbutlerapp/gitbutler/blob/8fee6821a243ba63b8a6f51f15dc666c65c5215c/crates/gitbutler-stack/src/stack_branch.rs#L245-L248

But before we get there, we validate the name in `update_branch` by https://github.com/gitbutlerapp/gitbutler/blob/8fee6821a243ba63b8a6f51f15dc666c65c5215c/crates/gitbutler-stack/src/stack.rs#L571-L581

That validation, among other things, checks that the name doesn't already exist. Which of course it does in this scenario. I think that's a good check to have and don't want to complicate `validate_name()`, so this seems like a decent solution.

## 🎫 Affected issues

Fixes: #11741

## Test output before patch

I like to test that my test actually does something, here's the test output from before applying the production code patch.

```bash
$ cargo test -p gitbutler-stack
<snip>

failures:

---- update_branch_name_to_same_name_is_noop stdout ----

thread 'update_branch_name_to_same_name_is_noop' (138422) panicked at crates/gitbutler-stack/tests/mod.rs:395:5:
assertion failed: result.is_ok()
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    update_branch_name_to_same_name_is_noop

test result: FAILED. 49 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 30.77s
```

### DIFF
--- a/crates/gitbutler-stack/src/stack.rs
+++ b/crates/gitbutler-stack/src/stack.rs
@@ -569,7 +569,9 @@ impl Stack {
         let mut updated_heads = self.heads.clone();
 
         // Handle name updates
-        if let Some(name) = update.name.clone() {
+        if let Some(name) = update.name.clone()
+            && name != branch_name
+        {
             let head = updated_heads
                 .iter_mut()
                 .find(|h: &&mut StackBranch| *h.name() == branch_name);
--- a/crates/gitbutler-stack/tests/mod.rs
+++ b/crates/gitbutler-stack/tests/mod.rs
@@ -379,6 +379,26 @@ fn update_branch_name_fails_validation() -> Result<()> {
 }
 
 #[test]
+fn update_branch_name_to_same_name_is_noop() -> Result<()> {
+    let (ctx, _temp_dir) = command_ctx("multiple-commits")?;
+    let mut test_ctx = test_ctx(&ctx)?;
+    let branch_name = String::from("virtual");
+
+    let update = PatchReferenceUpdate {
+        name: Some(branch_name.clone()),
+        description: None,
+    };
+    let result = test_ctx
+        .stack
+        .update_branch(&ctx, branch_name.clone(), &update);
+
+    assert!(result.is_ok());
+    assert_eq!(test_ctx.stack.heads[0].name(), &branch_name);
+
+    Ok(())
+}
+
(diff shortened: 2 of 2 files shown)


krlvi (approved):
fine catch! thanks

PR #11745, opened 2026-01-08
fix: ensure first changed file is correctly selected for preview when switching between commits

nshcr:
https://github.com/user-attachments/assets/effcb600-ce3e-46b9-8ada-37678110c351

This PR fixes an issue in the commit history preview within the branch view, where switching between different commits did not correctly update the preview to the first changed file of the currently selected commit.

The root cause is likely a race condition between multiple Svelte effects in the codebase. The issue is reliably reproducible, but the resulting state can appear somewhat random.

Since race conditions caused by effects execution order in Svelte are complex to reason about, the approach taken here is to add additional logic to ensure that a fresh selection state is always used whenever the context changes, which guarantees correct behavior.

### DIFF
--- a/apps/desktop/src/components/ChangedFiles.svelte
+++ b/apps/desktop/src/components/ChangedFiles.svelte
@@ -6,6 +6,7 @@
 	import emptyFolderSvg from '$lib/assets/empty-state/empty-folder.svg?raw';
 	import { FILE_SELECTION_MANAGER } from '$lib/selection/fileSelectionManager.svelte';
 	import { readStableSelectionKey, stableSelectionKey, type SelectionId } from '$lib/selection/key';
+	import { debounce } from '$lib/utils/debounce';
 	import { inject } from '@gitbutler/core/context';
 	import { Badge, EmptyStatePlaceholder, LineStats } from '@gitbutler/ui';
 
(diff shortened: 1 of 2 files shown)


apps/desktop/src/components/ChangedFiles.svelte
```diff
@@ -57,13 +57,30 @@
...
+		if (selectionChanged || firstFileChanged) {
+			idSelection.clear(id);
+		}
+
+		previousSelectionKey = currentKey;
+		previousFirstPath = firstChangePath;
+
 		if (firstChangePath && autoselect && selection.entries.size === 0) {
-			idSelection.set(firstChangePath, selectionId, 0);
+			idSelection.set(firstChangePath, id, 0);
```
nshcr, line 89: Based on the description of this PR (https://github.com/gitbutlerapp/gitbutler/pull/10225), I kept the use of a stable selection key (L67), while also unifying the selectionId used to the deserialized ID.

nshcr:
I noticed that this approach would cause the file selection state from the previous commit to be reset, so I've temporarily marked this PR as a draft and will take another look at the issue later.

nshcr:
This commit should fix the bug where the selected file was not preserved when switching contexts. Preserving the previously selected file is the intended behavior. It also addresses the issue described in the original post.

Adding the debouncing logic here is a compromise on my part - the interactions between Svelte's reactive effects in this area are too complex. I'd appreciate it if someone could help review whether this approach has any issues.

Byron:
My apologies for the late response, and thanks so much for making GitButler better!

I'd be unable to review this, but hope @estib-vega will be able to take a look.

nshcr:
The issue this PR was meant to address no longer exists in the new ChangedFiles UI.
Since it also now has merge conflicts involving a file that were removed by upstream, so I'm going to close it.

PR #11751, opened 2026-01-08
tests: by default, do not use color in stdout

jonathantanmy2:
Why I'm not automerging this PR: see if @Byron (or someone else) has opinions about which tests should emit color and which shouldn't. I picked what I thought were reasonable choices (journey, `status`, `help` retain their colors and therefore SVGs, the rest don't).

---

For most tests, the color in stdout does not significantly matter, so flip the default to not emit color codes. This allows code readers to see inline the important details of the test without needing to either launch a SVG viewer or read the relatively cluttered SVG file themselves.

Journey tests, `status`, and `help` retain their colors.

Note that `assert_with_oplog_redactions` needs a new redaction; with colors, the hash appears in 2 different colors, but without, it's one long hash and must be redacted differently. This has been done in this commit.


crates/but-testsupport/src/sandbox.rs
```diff
@@ -168,6 +168,9 @@ impl Sandbox {
     /// Create an assert with custom redactions. Adapt as needed.
     pub fn assert_with_oplog_redactions(&self) -> Assert {
         let mut redactions = Redactions::new();
+        redactions
+            .insert("[HASH]", regex::Regex::new(r#"\b[a-f0-9]{12}\b"#).unwrap())
+            .unwrap();
```
Byron, line 173: I couldn't see where this is used in this PR (but that might be a browser-search issue), but I don't think I understand why this would be needed in the first place as Git hashes in fixtures are stable and we would want to see them.
What am I missing?

crates/but/tests/but/journey.rs
```diff
@@ -140,22 +140,29 @@ fn from_workspace() -> anyhow::Result<()> {
     env.setup_metadata(&["A", "B"])?;
 
     env.but("status")
+        .env("CLICOLOR_FORCE", "1")
```
Byron, line 143: Since this will be a staple of our testing, can we have an extension trait to just say: `with_color_for_svg()`? I'd love this extension trait to also grow with similar staples, as there is other occasions where specific environment variables are set or removed (I am looking at you, `BUT_OUTPUT_FORMAT`).

If you find this enticing, also feel free to do this in a separate PR.

Byron (commented):
Thanks so much for tackling this - it's so much better already!

I would have merged it already, but have this one question that I'd love to know the answer for.

Thanks again.

crates/but-testsupport/src/sandbox.rs
jonathantanmy2: Git hashes are stable, but operation hashes don't seem to be. (I tried to explain this in the commit message but maybe it wasn't clear there.)

crates/but/tests/but/journey.rs
jonathantanmy2: I'm OK either way. I can do it in a separate PR next week.

crates/but-testsupport/src/sandbox.rs
Byron: Ah, right. When `but…` is used and creates commits, it inserts a change-id which isn't pinned. And since we don't really control how this binary is built, we can't pin it down like usual.

Also sorry for not reading the commit message, which I should as you are actually putting thought into them, it's a UX problem :/.

I guess it's fine here as these redactions really are only used when needed, and they have to be used explicitly.

crates/but/tests/but/journey.rs
Byron: I'd love that if you could pick that up, thanks for offering.

jonathantanmy2: Sent #11793

PR #11752, opened 2026-01-09
fix: subscribe persisted store to prevent reused instances from sharing state for FileListMode

nshcr:
https://github.com/user-attachments/assets/05f7040f-49a5-4b9d-b49d-d69f58c8d88a

This PR fixes a bug I mentioned in that issue (https://github.com/gitbutlerapp/gitbutler/issues/11746) but didn't explain in detail.

In the branch tab, the persisted list mode state of the changed file preview for the selected branch did not correctly respond to branch switching, causing all branches to share the list mode state of the first selected branch (from the initial component initialization).

This is addressed by subscribing to the persisted store within the FileListMode component, ensuring that each branch maintains its own list mode state without affecting the others.

Although my proposal in that issue was to share state, I also believe that the state sharing fixed by this PR was clearly a bug. In this case, using isolated state aligns with the current expected behavior.

### DIFF
--- a/apps/desktop/src/components/FileListMode.svelte
+++ b/apps/desktop/src/components/FileListMode.svelte
@@ -13,11 +13,13 @@
 	let { persistId, mode = $bindable() }: Props = $props();
 
 	const userSettings = inject(SETTINGS);
-	let saved = persisted<Mode | undefined>(undefined, `file-list-mode-${persistId}`);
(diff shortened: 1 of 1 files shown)


Caleb-T-Owens:
Looks good to me, nice job spotting that :+1:

PR #11754, opened 2026-01-09
Don't sign workspace commits from rebase

Caleb-T-Owens:
* Related to https://github.com/gitbutlerapp/gitbutler/issues/11618

This PR makes use of the `sign_if_configured` property added in my previous PR. When creating an editor graph, if a workspace commit is detected, we now set that property to `false`.


crates/but-rebase/src/graph_rebase/mod.rs
```diff
@@ -206,3 +215,43 @@ impl RevisionHistory {
         self.mappings.push(mapping);
     }
 }
+
+/// I wanted to assert _somewhere_ the defaults for non-workspace & workspace commits. It doesn't feel like the right place to do it in integration tests because we should assert behaviour rather than details there.
```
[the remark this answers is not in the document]
Caleb-T-Owens: The main intent of this comment was to just indicate "ooh, this might be a bit of a weird test". I don't think this needs rewords

Byron, line 219: I think that if it's part of the public API, it should be tested in `<crate>/tests`. These can follow the module structure of the crate under test to make it easy to know where to place any test.

crates/but-rebase/tests/rebase/graph_rebase/workspace_commit_behaviour.rs
```diff
@@ -0,0 +1,217 @@
...
+    author author <author@example.com> 946684800 +0000
+    committer Committer (Memory Override) <committer@example.com> 946771200 +0000
+    gitbutler-headers-version 2
+    gitbutler-change-id 00000000-0000-0000-0000-000000000001
+
+    GitButler Workspace Commit
+    ");
+
+    // We expect "c" to remain signed
+    insta::assert_snapshot!(git_catfile(&repo, "c")?, @r"
```
Byron, line 142: While `catting` the whole commit object is convenient, I think it's worth focussing at what really matters. If it's signatures, these can easily be extracted with `repo.find_commit().extra_headers().pgp_signature()` and validated that way.

crates/but-rebase/tests/rebase/graph_rebase/workspace_commit_behaviour.rs
```diff
@@ -0,0 +1,217 @@
...
+    /Ixfyf+1YXRuiTVCSfZC1O37fo2FgH9yK4qqh8FHKknW2QAAAIEAvxU0LExhDbTamH4u7O
+    EIeCOR5S/O2MyI/aDbayxoHTx/AyTX01tGi973T+gSudh1OHUo0i4t2qXH0UtHivQjkwAx
+    O4j6sJyGoYu7FC5e6jnAfCoECNRLPxPCCDbyyGh5XekKoGCR4EnmAYDV9eFulwD8sAYhdE
+    4HaQysV/NhxaQAAACBAPzsVtQcdp9OCMtY+MK+nbKya4JajYe/vfOyPJLwCo4WX9xxdHrl
+    4782TSxxtPJegGDCFvbHRh2sY/ACMC3zqdRg2dPASCDTsP5Oob72L76Bg0zJjoSdSJwCjD
+    pMWS5gS0eNyZ5K2B1NgwjCYl6xKQFCeoixvZwaiDmzmJckAiVHAAAAgQDry0Jg/+k/N0LS
+    ek4l+CTdWiCDLH2YRpZf+uMHrHdleBF3cLU3tsvAztRHQbpZVBVol6RCX7kXZb4oiAXxBs
+    1+DtZaWWnCAll+lzPqxBYvbnNWinu+lmthvidXViV5keBDXwBz/B66inApVvq4PRSvIRti
+    /wNPSpUQOE43pLMhHwAAABB0ZXN0QGV4YW1wbGUuY29tAQ==
+    -----END OPENSSH PRIVATE KEY-----
```
Byron, line 43: Here (even) I came close to thinking that there is something like 'over-testing' 😅.

Byron (commented):
Thanks! Just a few small notes.

crates/but-rebase/tests/rebase/graph_rebase/workspace_commit_behaviour.rs
Caleb-T-Owens: Yesterday when we were discussing the "is signing stable" topic, didn't you suggest adding this assertion to indicate that we expected a consistent private key? Perhaps I misunderstood.

Byron: I remember nothing, but that doesn't mean anything with me 😅. With that said, as a stable key results in a stable signature which we do test, I'd remove this test which feels like adding noise.

Instead, one could put a note saying that hey, if the signature changed, either the key changed, or the commit hash changed. But we are testing the latter so people will be able to tell that all else being equal, somehow the key changed.

PR #11755, opened 2026-01-09
show staging and unstaged areas better

schacon:
Unstaged will now show as “Unstaged” rather than “Unassigned” and will say “no changes” when there are no changes.

Staged changes will now be shown above the stack, so we have a place to put the CliID when we have one.

CleanShot 2026-01-09 at 13 20 12@2x

---
- <kbd>&nbsp;2&nbsp;</kbd> #11755 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #11756


krlvi:

PR #11774, opened 2026-01-10
`but rub branch1 branch2` stacks branch1 on top of branch2

schacon:
Also adds a `but stack` command that does the same thing


krlvi:

Issue #11575, opened 2025-12-17
Stack is uninitialized


alvseek:
### Version

0.18.3

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

Stack is uninitialized

no matter what I do can't do anything

### How to reproduce (Optional)

update to ver 0.18.3

### Expected behavior (Optional)

I can back to workspace, rebase, or anything

### Relevant log output (Optional)

```shell
command: set_base_branch
params: {"projectId":"f5856ac0-7ae0-4244-8da0-a8d6578a2041","branch":"origin/develop"})

Stack is uninitialized
```

mikehuebner:
Seeing this exact thing. Have tried the following:
- Removed the project
- Disabled experimental features (attachment)
- Reinstalled gitbutler

Still nothing :( 

<img width="1144" height="938" alt="Image" src="https://github.com/user-attachments/assets/409c4a45-5f11-4fb2-95f1-b671b311eb01" />

mikehuebner:
Fixed it by wiping the repo from Gitbutler, recloning from Github.

Which is weird, I guess the local state got messed up and made this unusable.

Byron:
Thanks a lot for reporting, and for looking into it so it becoe clear that the metadata is probably at fault here.

This issue occurs when a `Head` stored in `.git/gitbutler/virtual-branches.toml` has an empty or possibly invalid `name` field.
Having tried it though, to me it seemed that it auto-fixes itself, but also while leaving a somewhat strangely convoluted virtual-branches.toml file. But that auto-fixed itself as well.

Yet, I couldn't reproduce that particular error, also when changing the target branch.

mikehuebner:
I just had it happen again after trying to apply a new remote since for some reason it lost it for one of my PRs. So I refreshed, removed the repo, re-added the local and it started doing this again. I'm re-cloning the repo again to see if that solves it for now.

Byron:
I am sorry to hear that. 
Instead of cloning the repository, you could also try to rename `.git/gitbutler` as this would make all metadata unavailable, while leaving the workspace mostly intract. If this would fix the issue, we'd know that the culprit is probably originating in `.git/gitbutler/virtual-branches.toml`.
Then also related: https://github.com/gitbutlerapp/gitbutler/issues/11546 .

DevAbdoTolba:
same problem here
commiting/pushing/rebasing

[...]

PR #11789, opened 2026-01-12
fix: make it less likely to reach unitialized stack error

krlvi:


Byron:
I so hope this works! And does that mean you have a reproduction? I never got one unfortunately.

Related to #11575 .

krlvi:
> I so hope this works! And does that mean you have a reproduction? I never got one unfortunately.
> 
> Related to #11575 .

I dont have a repro but the change here:

https://github.com/gitbutlerapp/gitbutler/pull/11789/changes/ae9d44ed726f2286584af78401271ef35ed7e9e7

now matches the semantic of what Stack::create does when creating a new stack. This way we shouldnt be able to reach an out of sync state

PR #11793, opened 2026-01-12
tests: extension trait to add color for SVG

jonathantanmy2:
This makes it clearer at the call site that callers want color for SVG output.

### DIFF
--- a/crates/but/tests/but/command/help.rs
+++ b/crates/but/tests/but/command/help.rs
@@ -1,6 +1,6 @@
 use snapbox::str;
 
-use crate::utils::Sandbox;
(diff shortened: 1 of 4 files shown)


crates/but/tests/but/command/help.rs
```diff
@@ -9,14 +9,14 @@ fn rub_looks_good() -> anyhow::Result<()> {
 
     // The help should be nice, as it's a complex command.
     env.but("rub --help")
-        .env("CLICOLOR_FORCE", "1")
+        .with_color_for_svg()
```
Byron, line 12: Wonderful!

PR #11794, opened 2026-01-12
Implements identifiers for stacks as well as the ability to diff and rub with stacks

krlvi:


crates/but/src/id/mod.rs
```diff
@@ -175,12 +177,25 @@ impl IdMap {
...
+            if let Some(id) = stack.id {
+                stack_ids.insert(
+                    id,
+                    CliId::Stack {
+                        id: id_usage.next_available()?.to_short_id(),
+                        stack_id: id,
+                    },
+                );
+            }
+        }
```
Byron, line 191: I think a test is missing for this one. Would @jonathantanmy2 be able to review it and add one?
Thanks so much for your help.

jonathantanmy2: Thanks for calling this to my attention. I'll take a look tomorrow morning (about 10 hours from now).

jonathantanmy2: Sent #11802. There's also the issue of whether stacks as obtained from `but_workspace::head_info` have a stable order, but a comment in `graph_to_ref_info` in `crates/but-workspace/src/ref_info.rs` seems to indicate that it is the case, at least for now (and if it ever becomes unstable, I think a lot of tests will start getting flaky, so I think we're fine).

krlvi: Sorry about the missing test. Thank you @jonathantanmy2 for adding it in #11802

Byron: Thanks @jonathantanmy2 !

> here's also the issue of whether stacks as obtained from but_workspace::head_info have a stable order

Yes, the order is stable.

PR #11798, opened 2026-01-13
status non-JSON: use `IdMap` for commit info

jonathantanmy2:
Why I didn't automerge: the number of Stack, Segment, and Commit data structures has increased by one, which is probably not what we want. I brainstormed some solutions:
 - create a single commit graph data structure in which different nodes can be annotated, e.g. `ButGraph<StackAnnotation, SegmentAnnotation, WorkspaceCommitAnnotation, RemoteCommitAnnotation>`. `IdMap` could take a `ButGraph<(), (), (), ()>` and turn it into a `ButGraph<StackInfo, SegmentInfo, WorkspaceCommitInfo, RemoteCommitInfo>`.
 - create one commit graph data structure that all CLI code uses, and that always contains CLI IDs (so every CLI command that operates on the commit graph must incur the performance hit of generating IDs, but I think that most if not all commands would want to parse or display them anyway, so it's fine). `IdMap` could fulfill that role, so if this is what we're going with, then the design in this PR would work and this PR is good to go.

There is also the idea that the commit graph and `IdMap` could remain separate if we always take a commit ID, for example, and then pass it to `IdMap` to get its CLI ID. This is the way things currently work, which I am trying to move away from - we can't use Rust's type system to guarantee that every entity that we operate on has an ID, and we have to code defensively by inserting fallbacks (like a `00` ID) just in case there's an edge case we missed.


jonathantanmy2:
I'll go ahead and merge this - we can revisit design decisions later, and it's nice to have `status` use `head_info` for its information and guarantee that everything it displays has an ID.

PR #11804, opened 2026-01-13
Use 32 characters in JJ-style change-id header

senekor:
Jujutsu uses 32 characters, so it's better if we generate change-ids of the same lenght. This makes it easier for gitbutler and Jujutsu to be compatible.

### DIFF
--- a/crates/but-core/src/change_id.rs
+++ b/crates/but-core/src/change_id.rs
@@ -7,8 +7,10 @@ use rand::Rng;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 /// How long a reverse hex change id should be if stored in actual reverse hex
-/// ASCII characters, ranging from `z` to `k`.
-const CHANGE_ID_REVERSE_HEX_LEN: usize = 40;
+/// ASCII characters, ranging from `z` to `k`. In accordance with the standard
+/// discussed together with the Jujutsu and Gerrit projects, it is 32.
+/// <https://lore.kernel.org/git/CAESOdVAspxUJKGAA58i0tvks4ZOfoGf1Aa5gPr0FXzdcywqUUw@mail.gmail.com>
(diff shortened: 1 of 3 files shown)


crates/but-core/src/change_id.rs
```diff
@@ -7,8 +7,9 @@ use rand::Rng;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 /// How long a reverse hex change id should be if stored in actual reverse hex
-/// ASCII characters, ranging from `z` to `k`.
-const CHANGE_ID_REVERSE_HEX_LEN: usize = 40;
+/// ASCII characters, ranging from `z` to `k`. It's 32 because that's the length
+/// Jujutsu uses. By using the same length, we are compatible with jj for free.
```
krlvi: I would suggest an update to the docs here. This has been discussed previously (the Git Merge conferece 2024) as well as over on the git mailing list https://lore.kernel.org/git/CAESOdVAspxUJKGAA58i0tvks4ZOfoGf1Aa5gPr0FXzdcywqUUw@mail.gmail.com/

It would be more helpful to reference that since this is not about matching JJ but about forming a standard.

krlvi (commented):
Thank you @senekor for this PR. I have a small documentation feedback, other than that looks good to me.

crates/but-core/src/change_id.rs
senekor: Good point, I adjusted the comment accordingly.

Caleb-T-Owens:
Thank you for highlighting this and submitting a PR. I ought to have double checked the length 😅.

Byron:
> Thank you for highlighting this and submitting a PR. I ought to have double checked the length 😅.

Me too! The PR spoke very confidently about 40 hex chars, I took that for granted 😅. And I just double-checked: my local LLM would have known that, so I think I will use this even to try amp up my workflow and just run fact-checks locally next time I get the chance.

PR #11808, opened 2026-01-14
status JSON: use `IdMap` for commit info

jonathantanmy2:
I'm on the fence about automerging this because it changes `status` output a bit (see the test), but I think the change is correct. I've explained this in the "status JSON: use IdMap for commit info" commit message. I'll merge this tomorrow if there are no comments.

This PR is same as #11798 but for the JSON part.


krlvi (approved):
Thanks. This looks reasonable wrt the base and the upstream bit in the test.

krlvi:
merging this to minimize the chance of conflicts :)

PR #11812, opened 2026-01-14
Remove legacy insert blank commit

krlvi:


Byron:
It's great that we have E2E but I haven't found a way to understand what the error actually is. Thanks to massive logs and unusable GitHub UI, the only way to see anything is to download the logs for local inspection. But even then it's often hard to find backend errors, which also seem to be absent here (excluding the 'standard' errors we always see).

apps/desktop/src/lib/stacks/stackService.svelte.ts
```diff
@@ -661,8 +661,7 @@ export class StackService {
 	}
 
 	get insertBlankCommit() {
-		const [legacyMutate, legacyStatus] = this.api.endpoints.legacyInsertBlankCommit.useMutation();
-		const [newMutate, newStatus] = this.api.endpoints.insertBlankCommit.useMutation();
+		const [newMutate, status] = this.api.endpoints.insertBlankCommit.useMutation();
```
Caleb-T-Owens: Hey! If we're getting rid of the old API, we should really update this function to directly return `this.api.endpoints.insertBlankCommit` and then update the call-sites to use the new API

krlvi: good point

krlvi:
Figured out the e2e issue - https://github.com/gitbutlerapp/gitbutler/pull/11812/changes/fc9af0370f96e75030a87d38fc4f0332d64ed389

I keep forgetting that the e2e tests dont use tauri... and finding the relevant error log was not trivial

PR #11813, opened 2026-01-14
remove legacy create commit code

krlvi:


Byron:
I really like how you rely on Copilot to fill in the PR notes for you, that saves time while being very usable from my point of view.
A named PR title would still be appreciated though for a quick overview, anything really 😅.

krlvi:
> I really like how you rely on Copilot to fill in the PR notes for you, that saves time while being very usable from my point of view. A named PR title would still be appreciated though for a quick overview, anything really 😅.

sorry this is still WIP - i just needed CI here :)

Byron:
Another technique I can imagine (but maybe you use it already), is to make public methods non-public, by a `cargo check --workspace`. If nothing fails to build, the warnings should show everything that can now be deleted.

PR #11817, opened 2026-01-14
fix: Parsing of the date time of stored revies

estib-vega:
The issue is that NaiveDateTime doesn't support the %+ format specifier,
which requires timezone information. The %+ specifier only works with 
timezone-aware DateTime types.

### DIFF
--- a/crates/but-forge/src/db.rs
+++ b/crates/but-forge/src/db.rs
@@ -43,7 +43,10 @@ impl TryFrom<but_db::ForgeReview> for ForgeReview {
     type Error = anyhow::Error;
     fn try_from(value: but_db::ForgeReview) -> anyhow::Result<Self, Self::Error> {
         fn to_iso_8601(datetime: &Option<chrono::NaiveDateTime>) -> Option<String> {
-            datetime.map(|dt| dt.format("%+").to_string())
+            datetime.map(|dt| {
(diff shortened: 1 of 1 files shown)


estib-vega:
cc @krlvi 
I was getting this error:

```bash
> but st

thread 'main' (3713367) panicked at /Users/estib/.rustup/toolchains/1.92-aarch64-apple-darwin/lib/rustlib/src/rust/library/alloc/src/string.rs:2891:14:
a Display implementation returned an error unexpectedly: Error
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It apparently could not properly display the string as specified and that panicked.
This made it work for me now

krlvi:
thanks for the fix! I wonder why i never ran into this earlier...

estib-vega:
Hmm, maybe me changing the some forgereview types are the thing that triggered the issue?

In any case, just wanted to raise that for context :)

Byron:
I'd really like to switch away from `chrono` to `jiff` - it's made by burntsushi and he knows his strings.
Maybe I give it a shot with Copilot.

@copilot Make this work by replacing `chrono` with `jiff`, see this commit for the fix that was needed: https://github.com/gitbutlerapp/gitbutler/pull/11817/commits/3981a17aa810bb22e8a6685159752bdd160acb54 . Ideally, the `but` crate does not use `chrono` anymore.

PR #11826, opened 2026-01-14
Make `prettier --check` actually run in CI when applicable

slarse:
## 🧢 Changes
I stumbled into the [LINUX.md file commit history](https://github.com/gitbutlerapp/gitbutler/commits/master/LINUX.md) and found that both @Byron and @jonathantanmy2 have cleaned up after my sorry butt forgetting to run formatting. I wondered how, as CI didn't fail on either of my PRs.

As it turns out, the `prettier` job _never_ runs, because there is no `docs` output from the `changes` job. So the condition to run `prettier` is trivially false. This then causes CI failure in [generate-ts-definitions](https://github.com/gitbutlerapp/gitbutler/actions/runs/20797364623/job/59734240631), as the script there runs format and then checks the output to detect changes in TS types. But that doesn't run if you only change docs, so doc-only PRs slip through.

This PR adds the missing output and makes `prettier` run if there are docs changes OR node changes (see checks on this PR for evidence that it runs!). Even though `generate-ts-definitions` will catch formatting errors in the case there are "node" changes, I think that's a rather obtuse way of showing it. It's better if it's clearly showcased here.

For reference, here's a failed run on my own fork showing that it actually functions: https://github.com/slarse/gitbutler/actions/runs/21006637327/job/60390596765

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -20,6 +20,7 @@ jobs:
       node: ${{ steps.filter.outputs.node }}
       rust: ${{ steps.filter.outputs.rust }}
       ui: ${{ steps.filter.outputs.ui }}
+      docs: ${{ steps.filter.outputs.docs }}
     steps:
       - uses: actions/checkout@v6
         with:
@@ -51,14 +52,14 @@ jobs:
 
   prettier:
     needs: changes
-    if: ${{ needs.changes.outputs.node == 'false' &&  needs.changes.outputs.docs == 'true'}}
+    if: ${{ needs.changes.outputs.node == 'true' || needs.changes.outputs.docs == 'true' }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
       - uses: ./.github/actions/init-env-node
-      - run: pnpm prettier
+      - run: pnpm prettier .
 
   generate-ts-definitions:
     needs: changes


krlvi (approved):

krlvi:
Thank you!

.github/workflows/push.yaml
```diff
@@ -51,14 +52,14 @@ jobs:
...
-    if: ${{ needs.changes.outputs.node == 'false' &&  needs.changes.outputs.docs == 'true'}}
+    if: ${{ needs.changes.outputs.node == 'true' || needs.changes.outputs.docs == 'true' }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
       - uses: ./.github/actions/init-env-node
-      - run: pnpm prettier
+      - run: pnpm prettier .
```
Byron, line 62: I couldn't believe my eyes as the `.` wanted to be the end of a sentence for me 🤦‍♂️.
With that said, if it's significant, is there a way to make clear why it's important? If removing it breaks CI I think that's enough of a sign that it indeed is important.

slarse: It's the target of the command, it's `.` as in "this directory". If you don't specify a target it just fails immediately. I guess you could make it `./` instead to make it more evident that it targets a directory, but that's not a very typical way of doing it in my experience.

I realize now though that the other similar commands put that `.` inlined in the the scripts in `package.json`, only the `prettier` script seems to be lacking it.

https://github.com/gitbutlerapp/gitbutler/blob/f31cbfb33e4118ed25018ad7fafc49b97c919398/package.json#L33-L36

I personally tend to leave the target out of the script as it makes the script more flexible in that I can specify a more narrow target, and TypeScript tooling is in general so fantastically slow that's often desirable. But for the sake of conformity we should probably inline the target in the `prettier` script as well.

Byron: That sounds like it would solve my problem. And maybe that's also the reason why I was so confused to see this `.` - I have never seen it before in this context.
Usually I just run `pnpm begood` and call it a day (after waiting excruciatingly long 32 seconds (and it takes 4GB of real mem to do the work apparently).

slarse: https://github.com/gitbutlerapp/gitbutler/pull/11875

PR #11843, opened 2026-01-15
UI: absorb method

estib-vega:
Add an absorb option to the file actions.

Screenshot 2026-01-15 at 16 22 28

On a selected file, open the modal displaying the absorb plan. On confirm, absorb the changes to the right commits


estib-vega:
- Needs design vetting
- Needs e2e tests
- Needs to be able to take in tree changes (right now it takes all worktree changes)

crates/but-api/src/legacy/absorb.rs
Byron, line 96: Maybe not for this PR but so that it's out there: So far I avoided putting implementations into the `but-api` and it's generally better to put it into plumbing crates where it's possible to test.
Passing a `ctx` to a function makes it hard to test as it then needs everything, and it's unclear what it actually uses.
While legacy, this is building debt.

PavelLaptev:
@estib-vega I made my changes. Feel free to merge

Screenshot 2026-01-19 at 15 41 15
Screenshot 2026-01-19 at 15 41 26
Screenshot 2026-01-19 at 15 41 44

PavelLaptev (approved):

PavelLaptev:
@Byron should we merge this

crates/but-hunk-assignment/src/lib.rs
```diff
@@ -149,6 +149,12 @@ pub enum AbsorptionTarget {
     HunkAssignments {
         assignments: Vec<HunkAssignment>,
     },
+    TreeChanges {
+        changes: Vec<but_core::ui::TreeChange>,
+        // Optionally, the stack to which the changes are assigned
+        #[ts(type = "string | null")]
```
Byron, line 154: It's interesting that it doesn't figure that out by itself, it seems like a problem we'd have a lot.

Byron (approved):
Right, now codeowners kicks in and `but-core` is among the changed crates.

Yes, let's merge this. Reasoning: all code changes are either in `ui` types, or in `legacy` modules, so anything goes.

PR #11844, opened 2026-01-15
Refactor Stack initialize & remove field 'name' from Stack

krlvi:
Take 2 on removing the name field

### DIFF
--- a/crates/but-meta/src/legacy.rs
+++ b/crates/but-meta/src/legacy.rs
@@ -154,16 +154,6 @@ impl Snapshot {
                 changed = true;
             }
 
-            if stack.name.is_empty() {
-                stack.name = stack
-                    .heads
-                    // experiments show this is the bottom-most branch
-                    .last()
-                    .map(|h| h.name.as_str())
-                    .unwrap_or_default()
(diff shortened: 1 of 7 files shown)


crates/but-meta/src/legacy.rs
```diff
@@ -997,6 +987,8 @@ fn branch_to_stack_branch(
 
 /// Deterministically compare two stacks by their `order` field, using `name` and `id` as a tiebreaker.
 fn order_then_name(a: &&Stack, b: &&Stack) -> Ordering {
+    #[allow(deprecated)]
+    // TODO: Need to remove the name ordering - no such field will be persisted.
```
krlvi, line 991: @Byron I am not sure why the meta code does this ordering... but the name is a value that is no longer persisted, so we should look into removing that check here. This is the **only** place where the value is being read.

Byron: Thanks for the ping. I took a look and came up with [this monster](https://github.com/gitbutlerapp/gitbutler/pull/11860), all this just makes no sense anymore, but that's OK.
The field is unused at least.

PR #11845, opened 2026-01-15
De-legicify but-hunk-dependencies

Caleb-T-Owens:


crates/but-action/src/simple.rs
```diff
@@ -98,14 +98,18 @@ fn handle_changes_simple_inner(
     }
 
     // Get any assignments that may have been made, which also includes any hunk locks. Assignments should be updated according to locks where applicable.
+    let repo = ctx.repo.get()?.clone();
+    let (_, graph) = ctx.graph_and_meta_from_head(perm.read_permission())?;
+    let workspace = graph.to_workspace()?;
```
Byron: On the latest `master` this will just be one line, as the workspace now owns the graph so it can be returned.

crates/but-api/src/legacy/diff.rs
Byron: When you see `project_id: ProjectId` you can always change it into whatever `ctx: &|&mut Context` you need. This will be handled by the `but_api` macro accordingly, and brings the API closer to being modern, or maybe even makes it modern like in this case.

crates/but-api/src/legacy/rules.rs
```diff
@@ -20,7 +20,11 @@ pub fn create_workspace_rule(
     request: CreateRuleRequest,
 ) -> Result<WorkspaceRule> {
     let ctx = &mut Context::new_from_legacy_project_id(project_id)?;
-    create_rule(ctx, request)
+    let guard = ctx.exclusive_worktree_access();
+    let repo = ctx.repo.get()?.clone();
```
Byron, line 22: This clone seems unnecessary. Do you mean `let repo = &*ctx.repo.get()?`?
In many cases this `repo` will deref to `&gix::Repository` automatically, so one could use it like `&repo` in places where that's needed, without the `&*` shenanigans.

Byron (commented):
I think there might be opportunities to make these parts of `but-api` officially modern (please see comments below).

Also, let's avoid `ctx.repo.get()?.clone()` as this prevents cache-reuse, which is really what we are after by reusing instances.
For now, Context isn't shared across calls from the clients, but the CLI can easily do that and benefit.

crates/but-api/src/legacy/rules.rs
Caleb-T-Owens: I only cloned the repository where it was required. This was primarily in contexts where a `&mut Context` was passed to other functions. Removing the clone meant that we couldn't have an exclusive mutable reference to the context because accessing the repo was holding an immutable reference

Byron: Ah, I see. I hope soon we can cut down on the `&mut Contxt` usage with `but-db` improvements and of course, not letting plumbing crates use `but-ctx` at all. One step at a time.

crates/but-action/src/simple.rs
```diff
@@ -98,14 +98,17 @@ fn handle_changes_simple_inner(
     }
 
     // Get any assignments that may have been made, which also includes any hunk locks. Assignments should be updated according to locks where applicable.
+    let repo = ctx.repo.get()?.clone();
+    let (_, workspace) = ctx.workspace_and_read_only_meta_from_head(perm.read_permission())?;
     let (assignments, _) = but_hunk_assignment::assignments_with_fallback(
         ctx,
+        &repo,
+        &workspace,
```
Byron, line 106: If something takes a `ctx`, they shouldn't take a `repo` anymore. After all, they can obtain it from the `ctx`.
Ideally, of course, `ctx` isn't passed at all and maybe that's the plan here, but I'd still not want that 'pattern' to proliferate.

Caleb-T-Owens: I considered whether to pass a repo or not because as you say, it can be obtained from the `ctx`.

The reason I chose to pass the `repo` along side  the `ctx` here, was because eventually my understanding is that we would like to stop passing the `ctx` all together, and in my mind, this is a little bit of a first step to actually getting there.

Byron: I see, and hope the last step to get rid of context won't be a large one. Should just be the DbHandle, I suppose.

PR #11846, opened 2026-01-15
Prepare but hunk dependency for optional stack ids

Caleb-T-Owens:


estib-vega (approved):
Absorption code looks good 👍

crates/but-hunk-dependency/src/ui.rs
```diff
@@ -93,9 +95,39 @@ impl HunkDependencies {
...
 }
+
+/// The target of a hunk lock. If a stack is identifiable, then it's StackId
+/// will be provided.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", tag = "type", content = "subject")]
+pub enum HunkLockTarget {
+    /// References a stack that has a StackId
+    Stack(StackId),
+    /// The hunk is locked to a stack that we can't reference
```
Byron: Can we add `.` to mark the end of sentences, and also, explain what it means that something can't be referenced?
Maybe this is a way of saying that there is no StackId, which would be enough for me to understand why it's unidentified.

PR #11847, opened 2026-01-15
demo of `status` output with local and remote commits

jonathantanmy2:
Here's a demo of `but status` output when local and remote commits diverge. The `git status` output in the .sh file is as I expect, but not the `but status` output. In particular, the "[origin/main]" is in the wrong place (it should be on 28baf9a), and looking at the code, I expect "upstream: on" (crates/but/src/command/legacy/status/mod.rs:640) to be printed, but it is not.

DO NOT SUBMIT


jonathantanmy2:
This question has been answered so I'll close it. (My subsequent work following this question is in #11899.)

PR #11871, opened 2026-01-17
Reduce UI large diff threshold from 2500 to 1000 lines

mtsgrd:


mtsgrd:
@krlvi I think this change is not very controversial, so I'm gonna go ahead and merge it. Let me know if you have any other thoughts? It would perhaps be better to show N lines by default, and allow the user to expand down, but this 1 line change is imho still an improvement.

krlvi:
makes sense. Does it chug above 1k?

PR #11874, opened 2026-01-18
Move update-check interval to AppSettings root

krlvi:


apps/desktop/cypress/e2e/support/mock/settings.ts
```diff
@@ -34,7 +34,7 @@ export const MOCK_APP_SETTINGS: AppSettings = {
 	},
 	ui: {
 		useNativeTitleBar: false,
-		cliIsManagedByPackageManager: false,
-		checkForUpdatesIntervalInSeconds: 3600
-	}
+		cliIsManagedByPackageManager: false
+	},
+	checkForUpdatesIntervalInSeconds: 3600
```
Byron, line 39: I see now that putting most settings into the `ui` section is wrong now that we have a CLI that would basically implement everything the UI can do, and will keep that in mind.

PR #11875, opened 2026-01-18
Inline target in prettier script

slarse:
## 🧢 Changes
Inlines the target of the `prettier` command into the script itself to conform with the rest of the TypeScript tooling scripts. Also updates outdated Copilot instructions for using the script.

See discussion in https://github.com/gitbutlerapp/gitbutler/pull/11826#discussion_r2692875038


Byron (approved):
Loving it, thanks!

PR #11876, opened 2026-01-18
Revert "Move update-check interval to AppSettings root"

krlvi:
Reverts gitbutlerapp/gitbutler#11874

### DIFF
--- a/apps/desktop/cypress/e2e/support/mock/settings.ts
+++ b/apps/desktop/cypress/e2e/support/mock/settings.ts
@@ -34,7 +34,7 @@ export const MOCK_APP_SETTINGS: AppSettings = {
 	},
 	ui: {
 		useNativeTitleBar: false,
-		cliIsManagedByPackageManager: false
-	},
-	checkForUpdatesIntervalInSeconds: 3600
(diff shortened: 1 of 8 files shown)


Byron:
That's interesting, did it the move cause failures somewhere and wasn't worth it after all?

krlvi:
> That's interesting, did it the move cause failures somewhere and wasn't worth it after all?

No. I removed this because I decided to give up on this field. I saw this code https://github.com/gitbutlerapp/gitbutler/blob/master/crates/gitbutler-tauri/src/main.rs#L51-L59 and decided that I'm not interested in this field anymore. 

While working on this pr https://github.com/gitbutlerapp/gitbutler/pull/11879 I introduced a comment here https://github.com/gitbutlerapp/gitbutler/pull/11879/changes#diff-fd2a3623a222353213ec5d70af6a08c0b963e9fed02168a8d02d8cc5e7dd35c5R51-R54 to highlight that this behavior is not desirable.

Moving forward we would like to decouple from the tauri update mechanism so this UI field can stay but it is now designated as legacy.

Byron:
I see, thanks 🙏.

Issue #4949, opened 2024-09-18
Cannot discard newly added "binary" files


ndom91:
### Version

latest

### Operating System

Linux

### Distribution Method

AppImage (Linux)

### Describe the issue

Cannot discard binary files.

Context:

1.
![image](https://github.com/user-attachments/assets/d83ea6ac-9716-4608-9480-72f4c5033af6)

2. 
![image](https://github.com/user-attachments/assets/45f6d38f-33b4-4bfd-85b4-4fd4fdaa923c)

ndom91:
The `Discard` option is no longer shown for binary files in the staging area.

PR #11877, opened 2026-01-18
Fix Discard change for added/deleted files

godnight10061:
## Summary
Fix Discard Change for added/deleted files by treating these as whole-file operations (send empty hunk headers) and hiding the line-selection discard action for those statuses.

## Background
The UI was sending hunk headers for file additions/deletions. The backend expects whole-file discard semantics for these, which can otherwise result in an error and a broken Discard flow.

## Testing
- pnpm prettier .
- pnpm lint
- pnpm check
- pnpm -C apps/desktop test
- pnpm build:desktop
- cd e2e && pnpm exec playwright test --config ./playwright/playwright.config.ts --tsconfig ./playwright/tsconfig.json playwright/tests/unifiedDiffView.spec.ts --workers=1 --timeout=300000
- Fork CI workflow_dispatch: Fork CI (discard fix) on sha ce369e8a28ac0f05c48ff52215aace7efaa53945 (run 21114126403)

## Before vs After

https://github.com/user-attachments/assets/208e7dbf-73ae-4a2a-a7a0-57ee6fe7d06c

https://github.com/user-attachments/assets/21b2e7fc-47ea-4847-bd77-a94d2ea46544

## Issue
Refs #4949

### DIFF
--- a/apps/desktop/src/components/HunkContextMenu.svelte
+++ b/apps/desktop/src/components/HunkContextMenu.svelte
@@ -75,13 +75,16 @@
 
 		unselectAllHunkLines(item.hunk);
 
+		const isWholeFileChange =
+			change.status.type === 'Addition' || change.status.type === 'Deletion';
+
 		await stackService.discardChanges({
 			projectId,
 			worktreeChanges: [
 				{
 					previousPathBytes,
 					pathBytes: change.pathBytes,
-					hunkHeaders: [item.hunk]
+					hunkHeaders: isWholeFileChange ? [] : [item.hunk]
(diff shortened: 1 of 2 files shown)


Byron:
Thanks so much for contributing :)!

This is a great fix, and I can confirm that this is indeed the intention. If no hunk is selected, then the UI should send empty hunks so the file can be treated as a whole. For discarding, this is definitely relevant.

estib-vega:
@godnight10061 Thanks so much for the fix & the test. This is a great contribution! ❤️

PR #11879, opened 2026-01-19
CLI: implement checking for updates

krlvi:


crates/but/src/args/update.rs
```diff
@@ -11,4 +11,14 @@ pub enum Subcommands {
...
     /// for your platform and release channel.
     Check,
+    /// Suppress update notifications for a specified duration, defaulting to 1 day.
+    ///
+    /// Temporarily hide update notifications for the CLI. The suppression
+    /// will automatically expire after the specified number of days (default: 1 day).
+    /// Maximum suppression duration is 30 days.
+    Suppress {
+        /// Number of days to suppress update notifications (1-30, default: 1)
+        #[clap(default_value = "1")]
```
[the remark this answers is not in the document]
krlvi: nice! learned something here

crates/but-update/src/cache.rs
```diff
@@ -0,0 +1,366 @@
...
+///
+/// This function handles:
+/// - Creating the cache directory if it doesn't exist
+/// - Atomic write-and-rename to prevent corruption
+/// - Cleanup of temporary files on failure
+///
+/// # Errors
+///
+/// Returns an error if the file cannot be written or renamed.
+fn write_cache_atomically(cached: &CachedCheckResult) -> anyhow::Result<()> {
```
Byron, line 60: I really don't think we should have to go back to files. While I understand it's easier to implement, I'd strongly prefer having a `Cache` version of the `but-db` (inside of that crate) that has 'wipe on error' semantics for migrations and which lives in the cache directory.

That way, there is one place for all meta-data we store with two levels of importance, with a proven way to maintain it, and tests that ideally will only get better.

I can setup such a cache quickly now that `but-db` has been refactored.

krlvi: Very happy to take this in. I mostly wanted to not do any new SQLite functionality while we overhaul the setup.
Just note that for the use case here, this particular piece of data is not for a repository, but rather for the whole app (in a cache directory)

Byron: The gist from Discord: Great, let's have an app-wide cache, and will be sure it's specific to the build-type, e.g. Nightly.

PR #11890, opened 2026-01-19
Determine `LocalAndRemote` commit status based on matching remote _and_ matching branch name

mtsgrd:
It seems we determine the `LocalAndRemote` status on the commit being reachable by any remote, rather than matching on the specific remote and branch.

I noticed the problem when I had two stacked branches, where the remote bottom branch included the commits of the local only top branch. They were both colored as pushed and matching upstream, and the push button was disabled as a consequence.

*edit: code is 100% generated by Claude, so most likely incorrect.

### DIFF
--- a/crates/but-workspace/src/branch_details.rs
+++ b/crates/but-workspace/src/branch_details.rs
@@ -80,7 +80,7 @@ pub fn branch_details(
 
     let mut authors = HashSet::new();
     let (mut commits, upstream_commits) = {
-        let commits = local_commits_gix(branch_id, integration_branch_id.detach(), &mut authors)?;
+        let commits = local_commits_gix(branch_id, integration_branch_id.detach(), remote_tracking_branch_id.map(|id| id.detach()), &mut authors)?;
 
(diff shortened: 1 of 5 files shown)


mtsgrd:
@Byron this PR was entirely generated by Claude so maybe disregard the specifics. I was debugging why the top branch had commits that appeared LocalAndRemote when in fact those same commits only existed on the remote in the bottom branch.

Claude seemed to confirm that we only check that the commit is reachable on any remote, while it feels to me the check should be for the specific remote, and reachable from a specific branch. Does this make sense to you?

image
image

Byron:
Thanks a lot for digging into this! I also think this should be fixed, and can be fixed. Will get to it right after #11887 .

Byron:
As getting into this could be a huge time-sink, I had to put it onto my backlog to be able to focus on [unapply()](https://github.com/gitbutlerapp/gitbutler/pull/10872). The only consolation I can give is that I let it jump the queue and put it second.

krlvi:

PR #11893, opened 2026-01-19
but-ollama

estib-vega:
- Create the ollama client in rust
- Create the LLMProvider and LLMClient abstraction
- Use the LLM capabilities through the LLMProvider interface, making room for different providers


Byron:
While being excited about the functionality this provides, I also think a lot of it is insta-tech-debt.
Without review, I wouldn't know how to make a difference, so I will go for categorising `but-` crates according to my understanding of their quality, giving them a chance to improve over time.

The worst offender here, to my mind, is the lack of any kind of documentation, something that expresses intent at least. So my suggestion for you is to fill in `lib.rs` with everything you know about the crate, and everything you think you want it to be. Maybe how it's going to be used, how it fits in. That way, future contributors will have something to go by.

PR #11900, opened 2026-01-20
But rub - handle ambiguity by prompting

krlvi:


crates/but/src/command/legacy/rub/mod.rs
```diff
@@ -183,45 +314,231 @@ fn makes_no_sense_error(source: &CliId, target: &CliId) -> String {
...
+
+    // Defensive check
+    if matches.is_empty() {
+        return Err(anyhow::anyhow!(
+            "Internal error: prompt_for_disambiguation called with empty matches"
+        ));
+    }
+
+    // Check if we're in an interactive environment
+    let is_interactive = std::io::stdin().is_terminal() && out.for_human().is_some();
```
Byron, line 347: I will make it easier to see if something is interactive, beyond providing only `out.prepare_for_terminal_input() -> Option<Prompt>`.

Byron (commented):
All I can do is to look for patterns and support them with `OutputChannel` and friends, maybe there is more you'd need though.

PR #11908, opened 2026-01-20
Improve the but-llm crate: On token callback

estib-vega:
- Improve the on-token callback signature so that the caller doesn’t
  have to care about creating an arc
- Move the ChatMessage-specific functions to the chat module
- Improve documentation

### Follow-ups
- Further pretty-fication of the code and documentation of the inner-workings of this
- Anthropic implementation
- LLM Studio implementation
- Read from the git config

### DIFF
--- a/crates/but-action/src/lib.rs
+++ b/crates/but-action/src/lib.rs
@@ -105,29 +105,24 @@ pub fn freestyle(
     }));
 
     // Now we trigger the tool calling loop.
-    let message_id_cloned = message_id.clone();
-    let project_id_cloned = project_id;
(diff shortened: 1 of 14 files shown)


crates/but-action/src/lib.rs
```diff
@@ -105,29 +105,24 @@ pub fn freestyle(
...
     let (response, _) = llm.tool_calling_loop_stream(
         system_message,
         internal_chat_messages,
         &mut toolset,
-        model,
-        on_token_cb,
+        &model,
+        {
+            let emitter = emitter.clone();
+            move |token: &str| {
```
Byron, line 116: Perfect 🤩!

PR #11910, opened 2026-01-20
but branch: Unapply tests

estib-vega:
Add some tests to the CLI but branch unapply command

### DIFF
--- a/crates/but/tests/but/command/branch/mod.rs
+++ b/crates/but/tests/but/command/branch/mod.rs
@@ -1,3 +1,4 @@
 mod apply;
 #[cfg(feature = "legacy")]
 mod new;
+mod unapply;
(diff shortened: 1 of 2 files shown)


Byron:
I really like these tests, particularly because they don't use unrelated `but` commands to visualise the state, but instead focus on the command under test.

With that said, I would refrain from trying to be exhaustive here as we can assume that 'unapply' is already exhaustively tested in plumbing. Ideally we only test what's important to us, and I suppose this would mostly be about the messaging to the user as well as JSON, with more user-messaging tests added as we improve the UX.

estib-vega:
> I really like these tests, particularly because they don't use unrelated `but` commands to visualise the state, but instead focus on the command under test.
> 
> 
> 
> With that said, I would refrain from trying to be exhaustive here as we can assume that 'unapply' is already exhaustively tested in plumbing. Ideally we only test what's important to us, and I suppose this would mostly be about the messaging to the user as well as JSON, with more user-messaging tests added as we improve the UX.

Hey thanks!

I get your point but still I see value in having this level of coverage. We didn't catch the issue with not being able to unapply without the force flag for around 2 months. This now tests for that (indirectly).

Byron:
I see - that's definitely a bug in the CLI and thus it must be tested there.

PR #11911, opened 2026-01-20
Add tests for CC hooks and fix additional deadlock

Caleb-T-Owens:


Byron (commented):
As the code-change is related to calls that directly originate in `but-claude`, I think testing this in `but` CLI is at the wrong layer of abstraction.

https://github.com/gitbutlerapp/gitbutler/blob/f24f295eb2cebc4c8ee568cc55b9d9028b2702bd/crates/but/src/lib.rs#L283-L297

If we can boost test coverage in plumbing crates, that's where we should do it.

Caleb-T-Owens:
We talked over this in person... is a call "in person"?... we talked about this over an audiovisual communication system and we both see value in having a high level "integration" test, and plumbing-crate level tests. We ended up concluding that we should leave these here for now, and that in the longer term we would love to add more crate-specific tests that assert that behaviours like assignments happen correctly.

PR #11912, opened 2026-01-20
Default creating a single PR to yes

Caleb-T-Owens:


crates/but/src/command/legacy/forge/review.rs
```diff
@@ -99,12 +99,12 @@ pub async fn create_pr(
...
                 .prompt(format!(
-                    "Do you want to open a new PR on branch '{}'? [y/n]",
+                    "Do you want to open a new PR on branch '{}'? [Y/n]",
                     branch_name
                 ))?
-                .context("Aborted.")?
-                .to_lowercase();
-            if response == "y" || response == "yes" {
+                .map(|s| s.to_lowercase());
+            if response.is_none() || response == Some("y".into()) || response == Some("yes".into())
```
Byron, line 106: This can be `response.is_none_or(|r| r == "y")`. We shouldn't allocate `String` with "".into() for comparisons.

Caleb-T-Owens: This was superseded and fixed by the PR which added functions that standardised y/n prompts

PR #11914, opened 2026-01-20
Add methods for specific y/n inputs

Caleb-T-Owens:


crates/but/src/utils/output_channel.rs
```diff
@@ -197,6 +219,66 @@ impl InputOutputChannel<'_> {
...
+        &mut self,
+        prompt: impl AsRef<str>,
+        default: ConfirmDefault,
+    ) -> anyhow::Result<Confirm> {
+        use std::fmt::Write;
+        let suffix = match default {
+            ConfirmDefault::Yes => "[Y/n]",
+            ConfirmDefault::No => "[y/N]",
+        };
+        write!(self, "{} {}: ", prompt.as_ref(), suffix)?;
```
Byron, line 236: I think there is value in unifying the prompt, which is why I think these functions should call `prompt()` internally with a customised prompt, followed by the parsing they need to do.

Right now, there are two prompting styles, and I don't think these are needed. I would be nice if you could pick one to make prompting consistent.

PR #11918, opened 2026-01-20
Fix hunk drag to respect selected lines

### LINKED ISSUE #2513: Move line-based changes around (instead of hunk-based changes)
Currently, git-butler works with hunks.

In a significant number of times, hunks are too coarse grained for me. Therefore, I like the line-based workings of `git gui`.

image

(Artificial example to show the context menu)

I would love if git-butler could support line-based drag'n'drop, too.

By that I mean:

Variant 1:

1. User drags a single line inside a hunk to another virtual branch
2. git-butler removes that change from the origin branch and places it to the target (virtual) branch

Variant 2:

1. User marks an area of text inside a hunk. That area is typically 2 to n lines long.
2. User drag'n'drops this area to another virtual branch
3. git-butler removes change touched by the area from the origin branch and places it to the target (virtual) branch

godnight10061:
Fixes gitbutlerapp/gitbutler#2513.

- Dragging a hunk now uses the selected line range (when present) instead of always using the full hunk.

Testing:
- Local (Windows):
  - pnpm prettier
  - pnpm lint
  - pnpm check
  - pnpm test
  - cargo fmt --check --all
  - cargo check --workspace --all-targets
  - cargo check --workspace --all-targets --features windows

Before/After:

https://github.com/user-attachments/assets/2533b28a-e401-475b-b3da-22e3bef776fe

https://github.com/user-attachments/assets/089263a7-8d7b-4a86-a8fb-a115097e533f

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,8 +1,5 @@
 name: 'Publish'
 on:
-  schedule:
-    # every day at 1am UTC, 3AM CEST or 2AM CET.
-    - cron: '0 1 * * *'
   workflow_run:
     workflows: ['Nightly build']
     types:
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -126,6 +126,8 @@ jobs:
           persist-credentials: false
       - name: Dependencies for 'keyring'
         run: sudo ./scripts/install-minimal-debian-dependencies.sh
+      - name: Dependencies for tauri
+        run: sudo ./scripts/install-tauri-debian-dependencies.sh
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2.8.2
         with:
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -118,14 +118,14 @@ jobs:
     needs: changes
     if: ${{ needs.changes.outputs.rust == 'true' }}
     runs-on: ubuntu-latest
-    container:
-      image: ghcr.io/gitbutlerapp/ci-base-image:latest
     env:
       CARGO_TERM_COLOR: always
     steps:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
+      - name: Dependencies for 'keyring'
+        run: sudo ./scripts/install-minimal-debian-dependencies.sh
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2.8.2
         with:
@@ -218,8 +218,6 @@ jobs:
     needs: changes
     if: ${{ needs.changes.outputs.rust == 'true' }}
     runs-on: ubuntu-latest
-    container:
-      image: ghcr.io/gitbutlerapp/ci-base-image:latest
     env:
       CARGO_TERM_COLOR: always
     defaults:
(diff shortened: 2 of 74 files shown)


Byron (changes requested):
Thanks a lot for tackling this, this seems like something we really want fixed as it's surprising if the selection isn't respected.

There I wonder if the visualisation should also be indicating that it's just the selected lines that are dragged, it still looks like it's the entire hunk.

Meanwhile, let's also put it back into draft while there are conflicts. Maybe @mtsgrd or @Caleb-T-Owens can chime in to see if this generally the way to go in the frontend.

**In the backend, I don't think any of the changes should be needed.**

mtsgrd:
Indeed, thanks for putting together a fix for this. It was never implemented, by reasonably should. I see you've just recently pushed some more commits, and there's a lint issue there.. but let me check this branch out in a couple of hours, and we'll get it merged as soon as possible. 🙏

godnight10061:
@Byron @mtsgrd Thanks! I simply added a "selected lines" section right after the dragging hunk header. Would appreciate any UI ideas to make it even clearer.
Video: 

https://github.com/user-attachments/assets/231697db-9da2-4bd5-8ef7-a6a33efb3a70

Byron:
Wow, I am loving it! Now it's not surprising anymore and the info-box one is dragging around has just the right size and information.

Something that my eye found less pleasing was the line break for 'lines'. Maybe removing the `selected` would already do it?
  
Screenshot 2026-01-22 at 06 45 35

@PavelLaptev would probably have some notes as well now.

PavelLaptev:
@godnight10061 thank you for he PR!

I noticed that we can select lines to drag if the change is assigned or unassigned. 
However, hunks in committed changes can also be draggable (see video).
To maintain consistency, we should provide the same functionality in all places where we can drag hunks.

https://github.com/user-attachments/assets/d1d73d5a-8591-4fa3-94c9-a61adfd3517c

Byron:
Thanks a lot @PavelLaptev! Is this something that can be generalised in code so it's applied generically?
If not, then it would probably still make sense to take these changes one 'location' at a time to keep this PR small.

PavelLaptev:
> Thanks a lot @PavelLaptev! Is this something that can be generalised in code so it's applied generically? If not, then it would probably still make sense to take these changes one 'location' at a time to keep this PR small.

I’m not sure if this can be generalized, but I believe it should be, or the behavior will be inconsistent.

PavelLaptev:
@Byron, as we discussed this offline, we can merge this. However, let’s not forget to implement the ability to drag hunk lines in other places where it’s possible to do so.

Byron:
I don't mean to destroy the vibe but I am getting a strange feeling about this PR. It keeps growing, CI is still not green, and most importantly, it even goes as far as to alter `but-core` test cases. These seem completely out of range for fixing `hunk-drag` to respect lines.

To make it more transparent what's going on here, I recommend to publish a task list in the PR description to show what's planned here.

Thanks for your consideration.

godnight10061:
> I don't mean to destroy the vibe but I am getting a strange feeling about this PR. It keeps growing, CI is still not green, and most importantly, it even goes as far as to alter `but-core` test cases. These seem completely out of range for fixing `hunk-drag` to respect lines.
> 
> To make it more transparent what's going on here, I recommend to publish a task list in the PR description to show what's 

OMG Sorry Byron, I honestly forgot to mirror a branch to get my fork CI green before pushing here—I really appreciate your patience. So sorry for the mess. PR is getting too noisy, i am going to revert the experimental changes and start a new one.

PR #11920, opened 2026-01-20
`but push` with just single branch pushes directly without prompting

krlvi:


crates/but/src/command/legacy/push.rs
```diff
@@ -837,8 +840,9 @@ fn handle_no_branch_specified(
         anyhow::bail!("No branches found in the workspace");
     }
 
-    // Check if we're in an interactive terminal
-    if !out.can_prompt() {
+    // Check if we're in an interactive terminal with human output format
+    let is_interactive = out.can_prompt() && out.for_human().is_some();
+    if !is_interactive {
```
Byron, line 845: Is this something that AI did because it stored a memory of previous edits? I noticed something like this happen as well.

If there is a reason for using this exact condition, then I think it should be backed inot `out.can_prompt()` instead.

krlvi: Yes it was generated.... but, based on the docs, I would have written the exact same code manually.
Based on the docs:

- can_prompt() - Return true if external prompt support like Selection can be used.
- for_human() - Provide a write implementation for humans, if the format setting permits.

What the code intends to check is: 1) is this human output, i.e. not JSON and 2) do we have the possibility to prompt. If both hold, then do the prompting

Byron: Ah, I understand now, and actually misread the meaning of the line as all I really noticed was that `is_interactive` is back 😅.

In any case, this is fixed [here](https://github.com/gitbutlerapp/gitbutler/pull/11935).

PR #11924, opened 2026-01-20
Support "branch:filename" format

jonathantanmy2:
I decided to go with `branch@{stack}`.

Previous description:

Why I'm not automerging: this PR is as I envisioned it before starting, but when I was writing this, I noticed that specifying a branch name when you actually mean its stack is not done throughout `but` - it is done as the target of `rub`, but elsewhere (e.g. `diff`, where I noticed it), it is treated not as a stack but as itself (to specify the stack, you have to look at the `status` output and then get the stack CLI ID from there). I finished writing it anyway and have sent it so that you can see that having these IDs improves scripting for end users, and end users here includes us (writing the tests). But we probably can't have branch mean stack - any ideas? I was thinking of something like in Git, `branch@{stack}` or `branch^{stack}`, or maybe we should think about what's suitable for a durable stack identifier.

### DIFF
--- a/crates/but/tests/but/command/rub.rs
+++ b/crates/but/tests/but/command/rub.rs
@@ -28,45 +28,55 @@ Rubbed the wrong way. Source 'nonexistent1' not found. If you just performed a G
     Ok(())
 }
 
-#[test]
-fn uncommitted_file_to_unassigned() -> anyhow::Result<()> {
+fn assigned_uncommitted_file_env() -> anyhow::Result<Sandbox> {
     let env = Sandbox::init_scenario_with_target_and_default_settings("two-stacks")?;
 
     env.setup_metadata(&["A", "B"])?;
-    commit_file_with_worktree_changes_as_two_hunks(&env, "A", "a.txt");
+    env.file("a.txt", "arbitrary text\n");
+    env.but("zz:a.txt A").assert().success();
+    Ok(env)
+}
 
-    // Assign the change to A and verify that the assignment happened.
-    env.but("i0 A").assert().success();
-    env.but("--json status -f")
-        .allow_json()
+#[test]
+fn assign_uncommitted_file() -> anyhow::Result<()> {
+    let env = assigned_uncommitted_file_env()?;
+    env.but("diff A@{stack}:a.txt")
         .assert()
         .success()
         .stderr_eq(snapbox::str![])
         .stdout_eq(snapbox::str![[r#"
-{
-  "unassignedChanges": [],
-  "stacks": [
-    {
-      "cliId": "l0",
-      "assignedChanges": [
-        {
-          "cliId": "i0",
-          "filePath": "a.txt",
-          "changeType": "modified"
-        }
-      ],
-...
+────────╮
+j0 a.txt│
+────────╯
+     1│+arbitrary text
 
 "#]]);
+    Ok(())
+}
 
-    env.but("i0 zz")
+#[test]
+fn uncommitted_file_to_unassigned() -> anyhow::Result<()> {
+    let env = assigned_uncommitted_file_env()?;
+    env.but("A@{stack}:a.txt zz")
         .assert()
         .success()
         .stdout_eq(snapbox::str![[r#"
-Unassigned all hunks in a.txt in a stack
+Unassigned the only hunk in a.txt in a stack
 
 "#]])
         .stderr_eq(str![""]);
+
+    env.but("diff zz:a.txt")
+        .assert()
+        .success()
+        .stderr_eq(snapbox::str![])
+        .stdout_eq(snapbox::str![[r#"
+────────╮
+j0 a.txt│
+────────╯
+     1│+arbitrary text
+
+"#]]);
 
     Ok(())
 }
--- a/crates/but/src/command/legacy/rub/mod.rs
+++ b/crates/but/src/command/legacy/rub/mod.rs
@@ -1061,6 +1061,7 @@ mod tests {
         CliId::Branch {
             name: "main".to_string(),
             id: "ef".to_string(),
+            stack_id: None,
         }
     }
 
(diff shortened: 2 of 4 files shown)


crates/but/tests/but/command/rub.rs
```diff
@@ -28,46 +28,56 @@ Rubbed the wrong way. Source 'nonexistent1' not found. If you just performed a G
...
 
 "#]]);
+    Ok(())
+}
 
-    env.but("i0 zz")
+#[test]
+fn uncommitted_file_to_unassigned() -> anyhow::Result<()> {
+    let env = assigned_uncommitted_file_env()?;
+    env.but("A:a.txt zz")
```
Byron: While `A:a.txt` is great to read in comparison to a CLI-Id, one still has to know that `A` is a branch that exists as part of the fixture.
To my mind, the most maintainable tests will show what's there, do something, assert specifically and if they are cautious, finish like they started with the result state so nothing weird and unrelated sneaks in.

I am happy to backoff from that position though as this is clearly an improvement already, and the tests seem easy enough (right now) to follow.

The underlying problem that led me to thinking about this is still valid though, and I suppose for now we have to live with it (and those who change the output of common commands will have to auto-update a lot of expectations, and do so blindly).

Byron (commented):
Thanks a lot for adding (what seems like) the first prefixed Cli-ID 🎉.

When reading the issue about branches and stacks, i.e. given a branch `A` that is in a workspace as well, what does the user mean? Is it `A:file` the assignment in the stack, or `A:file` the change in `A^{tree}`, or is it the changed file in `A~1..A`?

Is this something that can be determined by context? But even so, it feels like there is often ambiguity particularly in `but rub` that can do *everything*.

jonathantanmy2:
> Is this something that can be determined by context? But even so, it feels like there is often ambiguity particularly in `but rub` that can do _everything_.

In Git, the command determines the context (e.g. `git diff` vs `git diff-tree`, `git update-ref` vs `git symbolic-ref`) so there are very few cases where I have to think about whether an annotated tag would turn into a commit, or a commit would turn into a tree (branches are not a problem because in Git, the branch is the tip, not a set of commits). I was thinking a bit about whether we should readd such context to `but` and have `but rub` delegate to these commands. So interactive users would use CLI IDs (which unambiguously point to a single thing), and `but rub` would maybe print what command it ends up running (e.g. `but assign-hunks A:a.txt zz`) and runs it. Here, the first argument is unambiguously a set of hunks (maybe one hunk), and the second argument is unambiguously either a stack or the unassigned area, so we can interpret accordingly (the LHS of the colon must be a container of hunks). I haven't thought through all the possible ambiguities still remaining, though. But if this works, then interactive users would typically use `but rub`, and scripters (including us writing tests) would use the specific commands.

crates/but/tests/but/command/rub.rs
jonathantanmy2: > While `A:a.txt` is great to read in comparison to a CLI-Id, one still has to know that `A` is a branch that exists as part of the fixture. To my mind, the most maintainable tests will show what's there, do something, assert specifically and if they are cautious, finish like they started with the result state so nothing weird and unrelated sneaks in.
> 
> I am happy to backoff from that position though as this is clearly an improvement already, and the tests seem easy enough (right now) to follow.
> 
> The underlying problem that led me to thinking about this is still valid though, and I suppose for now we have to live with it (and those who change the output of common commands will have to auto-update a lot of expectations, and do so blindly).

The problem of needing to know that `A` is a branch still exists, but the problem of auto-updating expectations is better now, right (well, better after this PR gets merged in some form)? Here, I've replaced `status` with a command that outputs more expectations and less fluff (`diff`), which is much easier to auto-update.

Byron:
> and but rub would maybe print what command it ends up running (e.g. but assign-hunks A:a.txt zz) and runs it.
> [..] But if this works, then interactive users would typically use but rub, and scripters (including us writing tests) would use the specific commands.

I am in love with this 🩷. This would make `but rub` an official router, and a trainer as well for those who'd rather use the actual command. CC @krlvi.

krlvi:
Note from our chat yesterday we mentioned it may be nice to be able to use indexes too

Eg `ab:3` refers to the third file instead of having to spell out the whole file

krlvi:
@jonathantanmy2 what are the outstanding things for this PR? I am excited that this could help out with a performance issue that we observed https://discord.com/channels/1060193121130000425/1442962582444183593/1463927097704710279

jonathantanmy2:
> @jonathantanmy2 what are the outstanding things for this PR? I am excited that this could help out with a performance issue that we observed https://discord.com/channels/1060193121130000425/1442962582444183593/1463927097704710279

The main outstanding thing is whether the current ID is a good format. After some thought, I think I'm going with `branchname@{stack}` to unambiguously represent the stack of the branch (so we have something that scripters can use, but also something that doesn't block us into a corner when designing IDs - we can always add something easier to type later). This PR won't directly help with the performance issue (we need another PR to actually support a new way of committed file IDs) but yes, adding the colon (like in this PR) is the first step.

krlvi:
I will try out the code in a bit, but do you mean that `but diff mybranch:file` diffs the file in the branch where `but diff mybranch@{stack}:file` diffs the uncommitted file staged to that stack?

jonathantanmy2:
> I will try out the code in a bit, but do you mean that `but diff mybranch:file` diffs the file in the branch where `but diff mybranch@{stack}:file` diffs the uncommitted file staged to that stack?

No - right now `mybranch:file` doesn't match anything (the caller of the parsing function will report that the ID doesn't match anything). (We probably should shift some error reporting to `IdMap` instead of letting its callers deal with it, but that is not done in this PR.) `but diff mybranch@{stack}:file` does work, and does what you describe.

jonathantanmy2:
I'll go ahead and merge this and then work on committed file IDs (to hopefully help with the performance issue). If you notice anything when you try out the code, feel free to let me know and I'll update it in another PR.

PR #11937, opened 2026-01-21
but absorb: dry run

estib-vega:
Add the ability to dry run the absorb, to see the plan before executing it.

### Things that would be nicer
It would be nicer if we could actually see if there would be any conflicts if we were to absorb

### DIFF
--- a/crates/but/tests/but/command/absorb.rs
+++ b/crates/but/tests/but/command/absorb.rs
@@ -384,3 +384,74 @@ Hint: run `but help` for all commands
 
     Ok(())
 }
+
+#[test]
+fn dry_run_shows_plan_without_changes() -> anyhow::Result<()> {
+    let env = Sandbox::init_scenario_with_target_and_default_settings("two-stacks")?;
+
+    env.setup_metadata(&["A", "B"])?;
+    commit_file_with_worktree_changes_as_two_hunks(&env, "A", "a.txt");
+
+    // Get initial status
+    let initial_status = env.but("--json status -f").allow_json().output()?.stdout;
+
+    // Run absorb with dry-run flag
+    env.but("absorb i0 --dry-run")
+        .assert()
+        .success()
+        .stdout_eq(snapbox::str![[r#"
+Found 2 changed files to absorb:
+
+Absorbed to commit: f4ea7f8 a.txt
+  (files locked to commit due to hunk range overlap)
+    a.txt @1,4 +1,4
+    a.txt @6,4 +6,4
+
+Dry run complete. No changes were made.
+
+"#]])
+        .stderr_eq(str![""]);
+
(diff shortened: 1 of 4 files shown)


crates/but/tests/but/command/absorb.rs
```diff
@@ -384,3 +384,74 @@ Hint: run `but help` for all commands
...
+    first
+    line
+    line
+    line
+    line
+    line
+    line
+    line
+    last
+    ");
```
Byron, line 435: I think if the precondition would include the hash of the commit (or HEAD commit?) maybe, then we'd know that it didn't change if the post-condition hash is the same.

I really tried hard but I couldn't tell if this file is changed or not, based on what's visible in the test, even with the full file reproduced here.

crates/but/tests/but/command/absorb.rs
```diff
@@ -384,3 +384,74 @@ Hint: run `but help` for all commands
...
+        .stdout_eq(snapbox::str![[r#"
+{
+  "unassignedChanges": [
+    {
+      "cliId": "i0",
+      "filePath": "a.txt",
+      "changeType": "modified"
+    }
+  ],
+...
```
Byron, line 452: @jonathantanmy2 I thinks this is also a good use of `but status` that is less likely to change frequently. Maybe that's already it.
Besides, maybe I am living in the past where `but status` was used much more with human output.

jonathantanmy2: `but status` indeed works here because we can elide the unimportant parts in snapbox using `...`. Having said that, `but diff` would work too and might be better, because we can see the exact uncommitted changes. I think `but diff zz` works now, and hopefully something like `but diff zz:a.txt` would work soon.

Byron: I wonder if `but diff` works, why `git diff` wouldn't work in the same situations.
Using stable ways of presenting state would be preferable to me, so it's always pro-`git` and pro purpose written tooling from `but-testsupport`.

Last but not least: our `but` commands do more and thus are always slower than what the respective `git` invocation or purpose-written tool would be.

jonathantanmy2: I'm OK with both `git diff` and `but diff` - I think they're stable in different ways. `git diff`'s functionality is fixed, but if we end up changing how we use Git's staging area, its output would change too, but presumably `but diff`'s won't.

Byron: That's a good point - the staging area is handled differently and `but` basically merges it with the worktree changes.

In any case, shelling out is nothing I'd do, but what I see here is probably good enough and easy enough to use.
Knowing myself, if I get the chance to actually work on the CLI I'd just add what I think I need to `but-testsupport`.

PR #11939, opened 2026-01-21
but-llm: anthropic

estib-vega:
- add the anthropic module for the llm providers
- remove anthropic-api

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -262,8 +262,6 @@ tracing-forest = { version = "0.3.0" }
 sysinfo = { version = "0.37.2", default-features = false }
 shell-words = "1.1.1"
 nonempty = "0.12.0"
-
-[profile.release]
 # There are no overrides here as we optimise for fast release builds,
 # which are used when building locally and when making Nightly builds.
 # The Stable release builds override the relevant settings at release time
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -285,16 +283,5 @@ incremental = false
...
-gix-hash = { opt-level = 3 }
-gix-actor = { opt-level = 3 }
-gix-config = { opt-level = 3 }
-sha1-checked = { opt-level = 3 }
-zlib-rs = { opt-level = 3 }
 serde = { opt-level = 3 }
 serde_json = { opt-level = 3 }
-# This one is special as we can run into debug assertions otherwise.
-# They have a time, but let's chose the time instead of always being hit by it.
-gix-merge = { opt-level = 3, debug-assertions = false }
```
Byron, line 300: Is this accidental? If not I would have hoped for a separate commit with the reasoning.
`gix` alone isn't enough for fast debug builds, it needs these (and maybe more) crates. `zlib-rs` for instance should probably be added there.

Cargo.toml
```diff
@@ -262,8 +262,6 @@ tracing-forest = { version = "0.3.0" }
 sysinfo = { version = "0.37.2", default-features = false }
 shell-words = "1.1.1"
 nonempty = "0.12.0"
-
-[profile.release]
```
Byron, line 266: Without this header, the comment below doesn't make sense anymore.

Byron (commented):
There are changes to the `Cargo.toml` file that I'd want back 😅.

PR #11943, opened 2026-01-21
but llm: Use the AI provider from the git config

estib-vega:
Read the global git config and use the LLM provider specified in there.

This allows the user to use the same configuration of LLM provider in the frontend and the backend

### DIFF
--- a/crates/but-llm/Cargo.toml
+++ b/crates/but-llm/Cargo.toml
@@ -25,3 +25,4 @@ anyhow.workspace = true
 strum = { version = "0.27", features = ["derive"] }
 reqwest.workspace = true
 uuid.workspace = true
+git2.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1231,6 +1231,7 @@ dependencies = [
  "but-secret",
  "but-tools",
  "futures",
+ "git2",
  "ollama-rs",
  "reqwest 0.12.28",
  "schemars 1.2.0",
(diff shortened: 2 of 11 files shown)


crates/but-llm/Cargo.toml
```diff
@@ -25,3 +25,4 @@ anyhow.workspace = true
 strum = { version = "0.27", features = ["derive"] }
 reqwest.workspace = true
 uuid.workspace = true
+git2.workspace = true
```
Byron, line 28: While I agree that writing Git configuration was a bit cumbersome in the past, we seem to be reading only, and that is no problem with `gix`.

Thus I am very surprised to see `git2` used here.

PS: Even writing Git configuration has utilities now to make that easy, if this was a concern.

estib-vega: I didn't find a place in which we were writing to the global git config using gix, only local.
But if there's a way of doing it, I can gladly change that

Byron: Ah, I understand what's going on now, thanks for the heads-up.

I think it's fine the global writes (in legacy but-api) to stay as is as we can change it when porting it to be modern.
But with newly written code, I think here there is no reason to keep adding `git2` dependencies for this kind of thing.
And as discussed, using the repository configuration to access these settings would be the right way of doing it to allow users repo-local overrides.
Thanks for taking another stab at this, it's appreciated.

PR #11953, opened 2026-01-22
secrets: fail silently if the error is not found while deleting it

estib-vega:
If a secret is not found while deleting it, fail silently.


estib-vega:
We are getting a bunch of error popups of 'No matching entry found in secure storage' that seem to come from the delete API. We should swallow those errors.

Caleb-T-Owens (approved):

Byron (approved):

PR #11955, opened 2026-01-22
logout updates the settings correctly

estib-vega:
- Treat null as deletion when merging JSON objects: source null values remove the corresponding key from the target instead of being ignored.
- Update merge logic to recurse for non-null values and continue to avoid merging arrays.
- Add tests: new test it_unsets_values_into_null and updated existing test to reflect deletion behavior.
- Ensure logging out deletes user data (removes user id from storage).

### What changed
- merge_non_null_json_value now:
  - removes target keys when source value is null
  - performs recursive merging for non-null object values
  - still does not merge arrays
- Tests updated/added to assert null unsets target fields.
- Logout flow now deletes stored user id (telemetry cleanup).

### Risk and migration
- Behavior change: null in incoming JSON now deletes keys in target; callers relying on previous "ignore null" behavior must be aware.

### DIFF
--- a/crates/but-settings/src/json.rs
+++ b/crates/but-settings/src/json.rs
@@ -30,6 +30,7 @@ pub fn json_difference(
 
 /// Based on Zed `merge_non_null_json_value_into`
 /// Note: This doesn't merge arrays.
+/// Null values are treated as "remove this key".
 pub fn merge_non_null_json_value(source: serde_json::Value, target: &mut serde_json::Value) {
     use serde_json::Value;
     if let Value::Object(source_object) = source {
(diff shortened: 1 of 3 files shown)


crates/but-settings/src/json.rs
```diff
@@ -30,6 +30,7 @@ pub fn json_difference(
 
 /// Based on Zed `merge_non_null_json_value_into`
 /// Note: This doesn't merge arrays.
+/// Null values are treated as "remove this key".
 pub fn merge_non_null_json_value(source: serde_json::Value, target: &mut serde_json::Value) {
     use serde_json::Value;
```
estib-vega, line 35: This is the important change, please be aware of the changes to this function

estib-vega: @Byron @krlvi

estib-vega:
https://github.com/user-attachments/assets/f3b1ba1e-574c-41ea-855d-3d19589d27e3

Byron (approved):
I think this is a lovely minimal change to solve the problem, and since we don't use `null` anywhere in our files I don't think it will have unwanted side-effects to let `null` delete keys when merged in.

crates/but-settings/src/json.rs
krlvi: Now the function name is misleading. And i wonder if this change of behavior affects other things

Byron: Also strange this wasn't caught by Copilot.

Byron: Fixed with https://github.com/gitbutlerapp/gitbutler/pull/11963

Issue #5407, opened 2024-11-02
Changes in Project Files Not Detected Automatically in GitButler (Stable and Nightly)


cuervolu:
### Version

0.5.855

### Operating System

Linux

### Distribution Method

AppImage (Linux)

### Describe the issue

Since recent updates, both the stable (0.13.8) and Nightly (0.5.855) versions of GitButler are failing to detect changes made to project files automatically. I have to manually reload the view each time to see updates. Additionally, the toolbar is missing in the Nightly version, so I am forced to use CTRL + R to refresh. This behavior does not occur in the stable version, but even there, changes are not detected until a manual refresh.

I have provided a quick video example showing the issue, though this problem also occurs with larger file changes.

https://github.com/user-attachments/assets/3eb91e97-cb2b-4564-919a-61af82df3bd5

I have tested different file changes and reloaded multiple times, but the automatic change detection remains unreliable.

### How to reproduce

1. Open GitButler.
2. Make any changes to your project (small or large).
3. Observe that GitButler does not detect the changes automatically.

### Expected behavior

In previous versions, GitButler detected file changes immediately. I believe this should still be the case unless there is a specific reason for the new behavior (which I'd appreciate clarification on). It's inconvenient to reload the page manually for every minor update.

Byron:
Thanks a lot for reporting!

On MacOS, trying 0.13.7, 0.13.8 and nightly 0.5.855 I was unable to reproduce the issue on MacOS. Reeling in @ndom91 who will probably already have an idea what changed between these releases.

cuervolu:
Any updates on this? The error is still present in the latest stable and nightly versions, though it occurs less frequently.

ndom91:
Hey sorry missed this one. 

So I've been using the appimage for months and I've run into this issue with every version as far back as I can remember unfortunately. 

Do your changes always not get picked up? Or does it work for a while but then at some point stop picking up changes? In my case it works like the second example, the file watcher usually works for a bit, but then stops. 

Anyway, we're aware the file watcher could use some more love, we've put up an issue to track it generally here:

cuervolu:
Yes, it happens like the second option. Sometimes, when I make a commit or a pull request, only then does it realize there were other changes.

Anyway, I'll pay closer attention to that issue, although it's true that AppImages are often quite complex, so I'm not surprised this happens.

ndom91:
[...]

PR #11958, opened 2026-01-22
Update `gix` to latest version to include `gix status` fix to not fail if worktree files can't be read.

Byron:
Right now, when a file that shows up in the status check can't be diffed, the whole operation fails.

It turns out that on Windows, there are files that are locked and can't be read, so we should avoid failing in this case.

Unfortunately the exact error code is unknown, but maybe we can make it a general feature and log the ignored files for now.

Related to #5407

Edit: 

Actually this needs https://github.com/GitoxideLabs/gitoxide/blob/5c1bd0387f98eee37265a42ba4b6624c783c9a71/gix-status/src/index_as_worktree_with_renames/mod.rs#L530 to be altered to be able to survive not being able to read the file. So nothing to be done here.

This is now fixed, probably, with a fix in `gix`. If it still fails, I think Windows might come up with different error codes, as for now I constrain it to specific issues.

### Tasks

* [x] [fix this in `gix`](https://github.com/GitoxideLabs/gitoxide/pull/2395)
* [x] use it here (assuming it will work as actual diffs are already fallible)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -198,7 +198,7 @@ dependencies = [
  "objc2-foundation 0.3.2",
  "parking_lot",
  "percent-encoding",
- "windows-sys 0.60.2",
+ "windows-sys 0.59.0",
  "wl-clipboard-rs",
  "x11rb",
 ]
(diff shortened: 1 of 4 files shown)


Byron:
@krlvi The codeowners trigger is because I had to update `gix` to the latest release, which happened today.

Caleb-T-Owens (approved):
LGTM

PR #11960, opened 2026-01-22
but-llm remove the reference to git2 configuration

estib-vega:
Prefer using the gix way of reading the git global configuration.

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #11962 
- <kbd>&nbsp;1&nbsp;</kbd> #11960 👈

PR #11962, opened 2026-01-22
App Setting: Use ever only the new settings

estib-vega:
Don’t mess with the old settings anymore, use always the new ones

---
- <kbd>&nbsp;2&nbsp;</kbd> #11962 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #11960


Byron:
@estib-vega if you rebase everything onto `master`, CI will pass.

crates/but-settings/src/lib.rs
```diff
@@ -83,6 +84,7 @@ mod json;
 mod legacy_settings;
 mod persistence;
 mod watch;
+use ts_rs::TS;
 pub use watch::AppSettingsWithDiskSync;
```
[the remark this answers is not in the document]
Byron: I agree, and it's more than a nit as it interrupts the pub export below which should stay adjacent to the private module that defines it.

crates/but-settings/Cargo.toml
```diff
@@ -6,6 +6,9 @@ authors.workspace = true
 publish = false
 rust-version.workspace = true
 
+[features]
+export-ts = []
```
Byron, line 10: This should include ` = ["dep:ts-rs"]`, along with an optional `ts-rs`.
But given that this is an issue in all current uses of `ts-rs`, this can certainly [be done in a separate PR](https://github.com/gitbutlerapp/gitbutler/pull/11971). If it merges first, you could rebase.

packages/core/src/generated/index.ts
```diff
@@ -3,3 +3,4 @@ export * as WorkspaceRefInfo from './workspace/refInfo';
 export * as WorkspaceLegacy from './workspace/legacy';
 export * as Workspace from './workspace/index';
 export * as HunkAssignment from './hunkAssignment';
+export * as AppSettings from './settings';
```
mtsgrd: How about changing this to `Settings` we use `Settings.AppSettings` in the code rather than `AppSettings.AppSettings`?

estib-vega: Yup, 100%

Byron (approved):
LGTM!

I only really checked the backend code, and didn't run the GUI either.

Here is a PR that this one would ideally be rebased against: https://github.com/gitbutlerapp/gitbutler/pull/11971 when merged.

PR #11966, opened 2026-01-22
Make cmd/ctrl + a work in input fields

mtsgrd:
Looks like the Edit -> Select All menu item is necessary for supporting
select all in text boxes.

### DIFF
--- a/apps/desktop/src/components/BranchLabel.svelte
+++ b/apps/desktop/src/components/BranchLabel.svelte
@@ -60,10 +60,7 @@
 	}
 
 	function handleKeydown(e: KeyboardEvent) {
-		if ((e.metaKey || e.ctrlKey) && e.key === 'a') {
-			e.stopPropagation();
-			inputEl?.select();
-		} else if (e.key === 'Enter' || e.key === 'Escape') {
+		if (e.key === 'Enter' || e.key === 'Escape') {
 			e.stopPropagation();
 			inputEl?.blur();
 		}
--- a/apps/desktop/src/routes/+layout.svelte
+++ b/apps/desktop/src/routes/+layout.svelte
@@ -125,8 +125,25 @@
 	const uiState = inject(UI_STATE);
 	const idSelection = inject(FILE_SELECTION_MANAGER);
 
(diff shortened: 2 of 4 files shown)


mtsgrd:
@PavelLaptev I discovered this link between the edit menu and text boxes today.. haven't tested it fully yet, but just wanted to draw your attention to it.

PavelLaptev:
Looks like this breaks "Select all" (cmd+A) in lists.

With the branch applied:

https://github.com/user-attachments/assets/66c01836-36ba-4fa3-9162-7b1237f264d0

Without:

https://github.com/user-attachments/assets/8a33c0b0-fa95-43b2-83d6-9237d1cdc641

mtsgrd:
You're right.. this is the thing where we can't get it to behave like standalone chrome/safari.

mtsgrd:
I changed the approach to having an event listener on the document, and checking if currentTarget is an input element. This seems to work better, wdyt?

PavelLaptev:
> This seems to work better, wdyt?

Works for "Select all" but not for the text selection.

https://github.com/user-attachments/assets/b4e5a9ee-1d98-4818-b086-eed07d918674

---

That's interesting. Text selection works for me without this branch.

https://github.com/user-attachments/assets/ceff5002-5250-4aff-85e3-42890a5ece7b

It appears to me that this branch doesn’t add the ability to select text, but rather the opposite.

mtsgrd:
Argh, this was a target vs. currentTarget problem. I force pushed again if you could try that out?

PavelLaptev:
@mtsgrd I was thinking that cmd+A doesn't work. But It does for me in nightly. Is it not working for you?

https://github.com/user-attachments/assets/abd358c5-d9cf-43fe-9edc-d2adb458b807

mtsgrd:
Cmd + A works in nightly, but not for bare `<input>` elements. There's a few places where we manually add the ability to select all, because it's broken by default (one of them being the branch search on the branches page).

PavelLaptev (approved):
It seems to work now. Tested on branch filters, commit messages, and the feedback modal.

PR #11976, opened 2026-01-23
 Improve docs and Rust style around `do_capture` for telemetry

Byron:
This is a follow-up to https://github.com/gitbutlerapp/gitbutler/pull/11972 .


Byron:
I removed the logic changes, leaving only the docs and style changes, and am merging it to get it off the table, given it's no functional change.

Also, I am looking forward to seeing the docs improved more so it's clear how these fields interact and how they should be used.

krlvi (approved):

PR #11979, opened 2026-01-23
Fix CI ignoring specified toolchain

Caleb-T-Owens:
The `RUSTUP_TOOLCHAIN` takes precedence over anything in our rust-toolchain.toml. This commit removes the environment variables so CI should now build with the expected version of rust


Byron (approved):
Thank you very much!

Let's try that.

From the CI run I can see that it doesn't unnecessarily install anything, probably because the versions match. One might think that it will install what's need if the *do not* match.
The image includes `rustup` as well so I'd think it works then.

PR #11983, opened 2026-01-23
POC: Workspace comparison and events

estib-vega:
Concept of emitting events from the backend for whenever the workspace changes.

### A better way of propagating the update.
Currently, I just added the ISO datetime of whenever the event was emitted, but I’m not sure this is the best approach.

### A fine-grained update
We could propagate information about what changed, i.e number of stacks, content of single stacks, etc

### DIFF
--- a/crates/gitbutler-watcher/src/handler.rs
+++ b/crates/gitbutler-watcher/src/handler.rs
@@ -27,6 +27,8 @@ pub struct Handler {
     // need extra protection.
     /// A function to send events - decoupled from app-handle for testing purposes.
     send_event: Arc<dyn Fn(Change) -> Result<()> + Send + Sync + 'static>,
+    /// The last seen workspace, to detect changes.
+    last_seen_workspace: Option<but_graph::projection::Workspace>,
 }
 
 impl Handler {
@@ -34,13 +36,14 @@ impl Handler {
     pub fn new(send_event: impl Fn(Change) -> Result<()> + Send + Sync + 'static) -> Self {
         Handler {
             send_event: Arc::new(send_event),
+            last_seen_workspace: None,
         }
     }
 
     /// Handle the events that come in from the filesystem, or the public API.
     #[instrument(skip(self, app_settings), fields(event = %event), err(Debug))]
     pub(super) fn handle(
-        &self,
+        &mut self,
         event: InternalEvent,
         app_settings: AppSettingsWithDiskSync,
     ) -> Result<()> {
@@ -52,6 +55,7 @@ impl Handler {
                 let repo = ctx.repo.get()?.clone();
                 let (_, workspace) =
                     ctx.workspace_and_read_only_meta_from_head(guard.read_permission())?;
+                self.workpace_changed(&workspace, project_id)?;
                 self.project_files_change(paths, ctx, &repo, &workspace)
             }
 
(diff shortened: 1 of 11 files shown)


crates/gitbutler-watcher/src/handler.rs
```diff
@@ -27,20 +27,23 @@ pub struct Handler {
     // need extra protection.
     /// A function to send events - decoupled from app-handle for testing purposes.
     send_event: Arc<dyn Fn(Change) -> Result<()> + Send + Sync + 'static>,
+    /// The last seen workspace, to detect changes.
+    last_seen_workspace: Option<but_graph::projection::Workspace>,
```
Byron, line 31: I find storing the entire workspace (which now includes the Graph) too heavy-handed.
There is a crate-private `WorkspaceState` which could become public, and that could be used here.

Byron (commented):
Sorry for posting on a draft, but the codeowners review invite went out (somehow while this is in draft mode, or it started as non-draft), and I am actually glad I get to see it early.

Right now, I feel like I am lacking so much context around this PR and its motivation that I can't say much more than this: I don't have a good feeling.

While the approach of putting this into the watcher is *probably* alright for now, I can only think of death by a thousand bugs when having to turn two workspaces into a set of events that perfectly represent the difference between both, just so that the UI perfectly reproduces that result state on its end. It can go wrong in two places now.

estib-vega:
> Sorry for posting on a draft, but the codeowners review invite went out (somehow while this is in draft mode, or it started as non-draft), and I am actually glad I get to see it early.
> 
> Right now, I feel like I am lacking so much context around this PR and its motivation that I can't say much more than this: I don't have a good feeling.
> 
> While the approach of putting this into the watcher is *probably* alright for now, I can only think of death by a thousand bugs when having to turn two workspaces into a set of events that perfectly represent the difference between both, just so that the UI perfectly reproduces that result state on its end. It can go wrong in two places now.

Who said anything about a set of events for perfect reproduction?

This current implementation only diffs the workspaces and then emits. 

The idea of a future more granular approach would be to send along single stack ids to update in the frontend. I don't think that is either error prone nor worse than what we have at this moment. 

Do you think otherwise?

crates/gitbutler-watcher/src/handler.rs
estib-vega: Sounds good, I'll take a look

Byron: `Workspace` could get a conversion back to `WorkspaceState`, that should be perfectly fine to maintain as well.

Byron:
> The idea of a future more granular approach would be to send along single stack ids to update in the frontend. I don't think that is either error prone nor worse than what we have at this moment.

That's great to hear, I guess I was thinking more of it than it really is.

> Do you think otherwise?

It's a PoC and I shall wait and see. I shouldn't be too stuck on thinking the UI can just query the full workspace each time something changes. I see a future where even the expensive computation won't be expensive again because they are cached, so making this call won't be costly anymore (but rather, as cheap as it gets).

1. Something changes
2. The UI gets to know about it
3. asks for a new workspace
4. refreshes itself, doing its reconciliation accordingly

Maybe it's good to differentiate between `git status` changes and changes that can affect the workspace, to not unnecessarily fetch it. Maybe there is more as well, I haven't ever thought about it in great detail.

PR #11988, opened 2026-01-23
Be sure to actually find the lowest merge-base.

Byron:
Otherwise, the workspace stack traversal may 'sneak' past the lower bound, which all paths should go through, and basically pick up all commits to the beginning of time.

## Tasks

* [x] reproduce
* [ ] prototype fix (find the lowest merge-base, keep searching until all paths are exhausted)
    - Even though it's implemented now, it's not working. Look into this again, fresh
* [ ] test to reproduce the issue without the fix applied

## Problem

This merge-base has paths that go around it, hence the traversal doesn't stop.

Screenshot 2026-01-23 at 20 40 46

## Solution

It must find the lowest merge-base, which isn't :34, but :40`.

## While there is no test…

Test it in the real world.

```
 cargo run --bin but-testing -- -C /Users/byron/dev-gb/gitbutler-no-merge-base graph --no-debug-workspace
```

with this patch applied:

```patch
diff --git a/crates/but-graph/src/api.rs b/crates/but-graph/src/api.rs
index c1763bb485..fd7bd475e1 100644
--- a/crates/but-graph/src/api.rs
+++ b/crates/but-graph/src/api.rs
@@ -89,7 +89,7 @@ impl Graph {
         //       And yes, let's avoid `gix::Repository::merge_base` as we have free
         //       generation numbers here and can avoid work duplication.
         let mut segments_reachable_by_b = BTreeSet::new();
-        self.visit_all_segments_including_start_until(b, Direction::Outgoing, |s| {
+        self.visit_all_segments_including_start_until(dbg!(b), Direction::Outgoing, |s| {
             segments_reachable_by_b.insert(s.id);
             // Collect everything, keep it simple.
             // This is fast* as completely in memory.
@@ -98,7 +98,7 @@ impl Graph {
         });
 
         let mut candidate = None;
-        self.visit_all_segments_including_start_until(a, Direction::Outgoing, |s| {
+        self.visit_all_segments_including_start_until(dbg!(a), Direction::Outgoing, |s| {
             let prune = true;
             if candidate.is_some() {
                 return prune;
diff --git a/crates/but-graph/src/projection/workspace.rs b/crates/but-graph/src/projection/workspace.rs
index 7ce1bbb1b4..5690bd3bf8 100644
--- a/crates/but-graph/src/projection/workspace.rs
+++ b/crates/but-graph/src/projection/workspace.rs
@@ -592,7 +592,7 @@ impl Graph {
             }
         };
 
-        let (mut lower_bound, mut lower_bound_segment_id) = ws_lower_bound
+        let (mut lower_bound, mut lower_bound_segment_id) = dbg!(ws_lower_bound)
             .map(|(a, b)| (Some(a), Some(b)))
             .unwrap_or_default();
 
```

The output will be something like 

```
[crates/but-graph/src/api.rs:92:55] b = NodeIndex(9)
[crates/but-graph/src/api.rs:101:55] a = NodeIndex(10)
[crates/but-graph/src/api.rs:92:55] b = NodeIndex(8)
[crates/but-graph/src/api.rs:101:55] a = NodeIndex(34)
[crates/but-graph/src/api.rs:92:55] b = NodeIndex(7)
[crates/but-graph/src/api.rs:101:55] a = NodeIndex(34)
[crates/but-graph/src/api.rs:92:55] b = NodeIndex(5)
[crates/but-graph/src/api.rs:101:55] a = NodeIndex(34)
[crates/but-graph/src/api.rs:92:55] b = NodeIndex(4)
[crates/but-graph/src/api.rs:101:55] a = NodeIndex(34)
[crates/but-graph/src/api.rs:92:55] b = NodeIndex(1)
[crates/but-graph/src/api.rs:101:55] a = NodeIndex(34)
[crates/but-graph/src/api.rs:92:55] b = NodeIndex(3)
[crates/but-graph/src/api.rs:101:55] a = NodeIndex(34)
[crates/but-graph/src/projection/workspace.rs:595:61] ws_lower_bound = Some(
    (
        Sha1(eb4078b63c3a1fade00796da6ee8830f59bf4cbb),
        NodeIndex(34),
    ),
)
```

And you'd want the `34` become `:40`, to find the longest path by manipulating this portion of the code (probably):

https://github.com/gitbutlerapp/gitbutler/blob/4cbc93746110697d6cad1f5e2f0ea2f1762f9a2c/crates/but-graph/src/api.rs#L101-L111

### DIFF
--- a/crates/but-graph/src/api.rs
+++ b/crates/but-graph/src/api.rs
@@ -79,36 +79,90 @@ impl Graph {
     /// The segment representing the merge-base is expected to not be empty, as its first commit
     /// is usually what one is interested in.
     // TODO: should be multi, with extra segments as third parameter
-    // TODO: actually find the lowest merge-base, right now it just finds the first merge-base, but that's not
-    //       the lowest.
+    /// Find the lowest merge-base between two segments using a standard two-pass algorithm with bitflags.
+    ///
+    /// The algorithm:
+    /// 1. Mark all segments reachable from `b` with REACHABLE_FROM_B flag
+    /// 2. Walk from `a`, marking segments with REACHABLE_FROM_A
+    /// 3. When we find a segment with both flags (merge-base), stop traversing that path
+    /// 4. Among all merge-bases found at minimum distance, return the one with highest generation
+    ///
+    /// The segment representing the merge-base is expected to not be empty, as its first commit
+    /// is usually what one is interested in.
     pub fn first_merge_base(&self, a: SegmentIndex, b: SegmentIndex) -> Option<SegmentIndex> {
-        // TODO(perf): improve this by allowing to set bitflags on the segments themselves, to allow
-        //       marking them accordingly, just like Git does.
-        //       Right now we 'emulate' bitflags on pre-allocated data with two data sets, expensive
-        //       in comparison.
-        //       And yes, let's avoid `gix::Repository::merge_base` as we have free
-        //       generation numbers here and can avoid work duplication.
-        let mut segments_reachable_by_b = BTreeSet::new();
+        use crate::segment::SegmentFlags;
+        
+        // Clear all flags first
+        for segment in self.inner.node_weights() {
+            segment.flags.set(SegmentFlags::empty());
+        }
+        
+        // Pass 1: Mark all segments reachable from b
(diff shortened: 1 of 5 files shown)


Byron:
Closed in favor of #12002.

PR #11997, opened 2026-01-24
Investigate and try to fix 'bonkers' lane handling when inserting blank commit

Byron:
Best shown in a video.

But in short, the lanes are completely shuffled around, and as the algorithm to handle these shouldn't have changed, I assume the vb.toml file is fixed up incorrectly. *Turns out that assumption was false though.*

https://github.com/user-attachments/assets/6749766b-935e-4856-94fd-c27d8c741dd8

Here is the commit that introduces the insert-blank-commit issue:
https://github.com/gitbutlerapp/gitbutler/commit/5980d99c3ed467804fa92cbc1d7511b7ab2425ee

### Tasks

* [x] investigate what's causing this
    - [x] It's *not* reconciliation 
    - [x] bisect - the problem starts here: 5980d99c3ed467804fa92cbc1d7511b7ab2425ee
* [ ] fix it - wait for @Caleb-T-Owens who can probably fix it faster/better
    - Note that `cargo run --bin but-testing -- -C /test-repo graph` shows the worktree metadata, which also changes in unexpected ways.

#### Disable reconciliation of metadata

This didn't have any effect.

```patch
diff --git a/crates/but-meta/src/legacy.rs b/crates/but-meta/src/legacy.rs
index 0ea7982847..30a47095b8 100644
--- a/crates/but-meta/src/legacy.rs
+++ b/crates/but-meta/src/legacy.rs
@@ -185,6 +185,7 @@ impl Snapshot {
 
     #[instrument(level = "debug", skip(self, repo))]
     fn reconcile_and_fix_vb_toml(&mut self, repo: &gix::Repository) -> anyhow::Result<()> {
+        return Ok(());
         fn make_heads_match(ws_stack: &but_graph::projection::Stack, vb_stack: &mut Stack) -> bool {
             // Always leave extra segments.
 
```


Byron:
Closing as Caleb will take a first look.

PR #12000, opened 2026-01-24
Implement `but stack`

### LINKED ISSUE #11318: Implement `but stack`
Something like `but stack <my-branch> <other-branch>` that rebases my branch onto another into a stacked branch

imskr:
## 🧢 Changes

- Added `but stack` command to stack one branch onto another
- Implemented `stack_branch_onto_target` function that removes and recreates a branch with a new anchor
- Added validation to prevent stacking a branch onto itself
- Added a check to prevent stacking the last-named branch in a stack (to avoid anonymous segments)
- Preserved commit metadata when recreating the stacked branch

## ☕️ Reasoning

The `but stack` command provides a way to reorganize branch hierarchies by changing which branch another branch is stacked on top of. This is useful when you want to reorder your work or change dependencies between branches. The implementation safely removes the old branch reference while preserving its metadata, then recreates it with the target branch as its new anchor point. The validation ensures we don't create invalid states like anonymous stacks or circular dependencies.

## 🎫 Affected issues

Fixes: #11318

### DIFF
--- a/crates/but/src/args/metrics.rs
+++ b/crates/but/src/args/metrics.rs
@@ -50,6 +50,7 @@ pub enum CommandName {
     Resolve,
     Update,
     Merge,
+    Stack,
     #[default]
     Unknown,
 }
--- a/crates/but/src/args/mod.rs
+++ b/crates/but/src/args/mod.rs
@@ -814,6 +814,31 @@ pub enum Subcommands {
         branch: Option<String>,
     },
 
+    /// Rebase a branch onto another branch to create a stacked branch.
+    ///
+    /// This command rebases the source branch onto the target branch,
+    /// creating a stacked branch relationship where the source branch
+    /// depends on the target branch.
+    ///
+    /// ## Examples
+    ///
+    /// Stack a feature branch onto another branch:
+    ///
+    /// ```text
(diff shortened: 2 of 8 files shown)


imskr:
Hello 👋🏻 @krlvi, I've created this for `but stack`. Could you give it a review? 🙇🏻

Byron (changes requested):
HI! Thanks for giving it a shot!

Since I took a glimpse already, please let me point out a couple of things to be changed for the PR to have a chance to be merged. It would probably also be broken up into smaller pieces.

* analysis/scope - what does it mean to `stack` a branch. The original instruction "Something like but stack <my-branch> <other-branch> that rebases my branch onto another into a stacked branch" isn't enough to get going here.
* implement stacking in `but-workspace` and test it to death
* wire it up in `but` (CLI) with a couple of tests for typical user interaction

I recommend closing this PR and doing a detailed analysis in the respective issue.

imskr:
Thank you @Byron this helps. I would also like to get input from @krlvi on this PR, before working more to avoid wasting efforts.

krlvi:
this is now supported with `but move`

PR #12002, opened 2026-01-24
Add lowest_merge_base algorithm to find true dominator

mtsgrd:
The existing first_merge_base finds the first common ancestor encountered
during traversal, but this can be incorrect when there are multiple paths
between segments (e.g., via merge commits). It may return a node that has
a "path around" it.

The new lowest_merge_base algorithm finds a common ancestor where ALL paths
from both segments must converge - the true dominator with no bypass.

Algorithm:
- Use synchronized bidirectional BFS with colored walkers from both segments
- Track when nodes are visited by both origins ("merged" nodes)
- When all walkers have reached merged nodes, select the one with the
  largest generation (farthest from tips = oldest = true dominator)

Example where the algorithms differ:

        A                   B
        │                   │
      mid_a ─────────────── M (merge commit)
        │                 ╱   ╲
        │               ╱       ╲
        │             ╱           ╲
        │           ╱            mid_c
        │         ╱                │
        │       ╱                  │
      base ◄───────────────────────┘

From A: only path is A → mid_a → base
From B: two paths exist:
  - B → mid_a → base
  - B → mid_c → base  (path around mid_a!)

- first_merge_base(A, B) → mid_a (incorrect, has path around via mid_c)
- lowest_merge_base(A, B) → base (correct, all paths converge here)


Byron:
Thanks a lot!

While reviewing I found a 'convenience' in the test which I will investigate, and I will investigate. Overall, I think it's good to have a `lowest-merge-base` version of the function that other uses of the `first-merge-base` flavor can be handled one at a time, maybe they should use the `lowest-merge-base` one as well.

### Tasks

* [x] make test more difficult - it still works 🎉
* [x] review implementation and docs ❌ - it's too strange and at least needs serious cleanup.
* [x] compare this to the `gix` implementation, it also has logic for that just like Git
    - Git gives a natural merge base, but always the one with the shortest paths. This is usually preferable
    - Independently of that, we'd still pick up too many commits in our traversal unless we cut it off by generation number
* [ ] implement this 'like one normally would'
* [ ] validate each of the adjusted expectations
* [ ] figure out `virtual_branches::apply_virtual_branch::rebase_commit` failure (if it's still a problem then)
* [ ] merge if the above pans

mtsgrd:
Closing this as per conversation with @Byron - normal merge base will do.

Byron:
Actually I will salvage parts of it, but will bring it up in a new PR then.

PR #12014, opened 2026-01-25
Optimize a few more dependencies for debug builds

mtsgrd:
Switching back to the workspace was unbearably slow, taking about 20s on
my own machine. With these changes it now takes about 6s.


mtsgrd:
@Byron does this look ok?

Byron (approved):
Looks great, thank you!
This also includes the Cargo.toml part from #12013.

PR #12020, opened 2026-01-25
Improve branches page

mtsgrd:
- use new multidiff preview
- revamp selection mechanism
- drop complicated persisted selection state

image


PavelLaptev:
Nice! Thanks for the update.

I noticed a few remaining issues on the GH Pages site:
- Selection State: Clicking a commit or branch doesn't show any visual 'selected' indicator.
- Commit Interactivity: Commits don't seem clickable. The app appears stuck on the selected branch, and clicking individual commits has no effect.
- Changes List: The list of changes isn't appearing when a branch or commit is selected.


https://github.com/user-attachments/assets/452f4ce0-b7f9-434d-9c9a-6ed5029215d4

mtsgrd:
@PavelLaptev yeah I was a bit trigger happy in adding you as a reviewer last night - so I put the pr back to draft. But now it's ready for another pass.

One important thing though, I've currently disabled the "click branch header to show all changed files".. sometimes we have issues with determining what commits to include in the lists, and when that happens you end up with veeery long file lists. I would be ok leaving out that functionality for now, wdyt?

mtsgrd:
I had to do a bit of back and forth here, but I think I've crossed all the t's and dotted the i's. One change I made here you might object to is I replaced the custom header for the target branch with a regular `BranchCard`. Does that feel ok with your or shall I bring back the original custom structure?

PavelLaptev:
Thanks, for all the updates!

> One important thing though, I've currently disabled the "click branch header to show all changed files".. sometimes we have issues with determining what commits to include in the lists, and when that happens you end up with veeery long file lists. I would be ok leaving out that functionality for now, wdyt?

Yeah, let’s disable branch header clicks then.

There are a few minor UI things, but I can fix them later.

PavelLaptev (approved):

PR #12022, opened 2026-01-26
Fix cmd + a in commit title input

mtsgrd:


PavelLaptev:
Seems to work 🤞
Thanks!

PavelLaptev:
The only place where it doesn’t work is the project selector.

image

peaklabs-dev:
@mtsgrd while your are at it cmd + z is also not working in commit title input

mtsgrd:
> cmd + z is also not working in commit title input

Oh boy.. yeah, I know. This has to do with that same thing where Tauri doesn't let us use native features like select all, undo/redo, without preventing these keys from reaching the embedded webview.

image

If you look at this edit menu, you'll see we have native support for copy paste etc.. but this also means we would be unable to listen for `cmd + c` with javascript.

Now.. this is problematic because, the lexical text editor uses a custom undo stack. This means, we effectively have to choose between having undo available in `<input>` elements or in our commit message editors. We can't have both. 🤷 

Unless, we find a way of letting those keypresses reach the webview despite being handle by Tauri.

PR #12027, opened 2026-01-26
web: bring back the login buttons

estib-vega:
Display the login buttons in the header and the footer again, even if
there’s someone already logged in.

Screenshot 2026-01-26 at 15 16 18
Screenshot 2026-01-26 at 15 12 09


PavelLaptev:
@estib-vega can we have it only in the header?

estib-vega:
> @estib-vega can we have it only in the header?

Sure!

PavelLaptev (approved):
Thanks!

PR #12029, opened 2026-01-26
Replace dot based testing assertions with a git-log-like output

Caleb-T-Owens:
The dot based assertions have bothered me for a while, so I used generative tooling to create an algorithm similar to the `git log —graph —oneline` output.


Byron:
So much better 🩷!

Caleb-T-Owens:
Thanks, I was never happy with the `dot` format because it was hard to visually assert whether it's right. Especially since I was adding more tests for Pavel's issue, I thought I'd take the time to get this right

PR #12031, opened 2026-01-26
Installer in rust instead of bash

krlvi:


scripts/install.sh
```diff
@@ -1,521 +1,105 @@
...
-success "✓ GitButler CLI installation completed!"
-echo ""
-info "Usage:"
-echo "  but --help           Show available commands"
-echo "  but status           Show branch status"
-echo "  but commit           Create a commit"
-echo ""
-info "For more information, visit: https://docs.gitbutler.com"
+# Run installer with forwarded arguments
+exec "$INSTALLER_BIN" "$@"
```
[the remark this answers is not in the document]
krlvi: We cant guarantee that there is minisign installed on the machine...
Taking a look at how rustup does it, they also dont do verification in the bootstrap script https://github.com/rust-lang/rustup/blob/main/rustup-init.sh

apps/web/tests/install-script.spec.ts
```diff
@@ -14,22 +14,20 @@ test.describe('Install script endpoint', () => {
 		// Verify the script content
 		const scriptContent = await response.text();
 
-		// Check for bash shebang
-		expect(scriptContent).toContain('#!/bin/bash');
+		// Check for shell shebang
```
Byron, line 17: It's definitely interesting to see the hosting webserver check the content of what's hosted.

crates/but-installer/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+serde_json.workspace = true
+tempfile.workspace = true
+dirs.workspace = true
+url.workspace = true
+
+# HTTP client using system libcurl (zero binary size impact)
+curl = { version = "0.4.49", default-features = false }
+
+# For archive extraction
+flate2 = "1.0"
```
Byron, line 30: `flate2` - good choice (I maintain it ;)). Default features are fine and good for a minimal ZLIB implementation in pure Rust. Not the fastest, but hopefully the smallest. Speed can be fixed with `zlib-rs` if needed.

crates/but-installer/src/config.rs
```diff
@@ -0,0 +1,290 @@
+//! Configuration and platform detection
+
+use anyhow::{Result, anyhow, bail};
+use std::env;
+use std::path::PathBuf;
+
+/// Channel type for the installation
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
Byron, line 8: Because it should be small, `Debug` could be something like: `#[cfg_attr(debug_assertions, derive(Debug))]`. Then the debug trait code will only be there when not building for releases.

crates/but-installer/src/config.rs
```diff
@@ -0,0 +1,290 @@
...
+            Some(version_str) => {
+                let version = Version::new(version_str.to_string())?;
+                Ok(VersionRequest::Specific(version))
+            }
+        }
+    }
+}
+
+/// Configuration for the installer
+pub struct InstallerConfig {
```
Byron, line 124: In a crate where we don't want dead code, I'd probably go for no `lib.rs` so it *should* bark if anything is unused even if it's `pub`.

crates/but-installer/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+[dependencies]
+anyhow.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+tempfile.workspace = true
+dirs.workspace = true
+url.workspace = true
+
+# HTTP client using system libcurl (zero binary size impact)
+curl = { version = "0.4.49", default-features = false }
```
Byron, line 27: Probably good for MacOS, but likely needs changing for other platforms to bundle everything.

crates/but-installer/src/install.rs
```diff
@@ -0,0 +1,374 @@
...
+
+    info(&format!(
+        "Installing {} channel to {}...",
+        channel.display_name(),
+        install_app.display()
+    ));
+
+    // Clean up any leftover temp files from previous failed installations
+    let _ = fs::remove_dir_all(&install_app_new);
+    let _ = fs::remove_dir_all(&install_app_backup);
```
Byron, line 81: Not sure about that. Should be very conservative when deleting files outside a temp directory. But probably OK also because it's in `~/Applications`.

crates/but-installer/src/install.rs
```diff
@@ -0,0 +1,374 @@
...
+        } else if ty.is_dir() {
+            copy_dir_all(&src_path, &dst_path)?;
+        } else {
+            fs::copy(&src_path, &dst_path)?;
+        }
+    }
+    Ok(())
+}
+
+/// Recursively remove the macOS quarantine attribute from a directory
```
Byron, line 358: Are you sure this is needed? I was quite convinced that macOS treats apps as files, so it won't spam attributes like that.

crates/but-installer/src/install.rs
```diff
@@ -0,0 +1,374 @@
...
+        }
+
+        // Verify final installation
+        let final_status = Command::new(&but_symlink)
+            .arg("--version")
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
+            .status();
+
+        if !final_status.map(|s| s.success()).unwrap_or(false) {
```
Byron, line 285: We should probably have a clippy lint for this: `final_stsatus.is_err_or(|s| !s.is_success())`.

crates/but-installer/src/lib.rs
```diff
@@ -0,0 +1,184 @@
+//! GitButler installer library
+//!
+//! This library provides functionality for installing GitButler on macOS systems.
+//! It handles downloading releases, verifying signatures, extracting archives,
+//! and setting up the application and CLI tools.
```
Byron, line 5: Ah, I see. You also plan to use this in `but` CLI, hence the library.

crates/but-installer/src/release.rs
```diff
@@ -0,0 +1,205 @@
...
+    pub platforms: HashMap<String, PlatformInfo>,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct PlatformInfo {
+    pub url: String,
+    pub signature: String,
+}
+
+pub(crate) fn fetch_release(config: &InstallerConfig) -> Result<Release> {
```
Byron, line 23: Took me a moment to realise this is `fetch_release_metadata` rather than the release itself.

crates/but-installer/src/shell.rs
```diff
@@ -0,0 +1,528 @@
...
+}
+
+pub(crate) fn setup_path(home_dir: &Path) -> Result<()> {
+    let bin_dir = home_dir.join(".local/bin");
+
+    // Canonicalize bin_dir for comparison (resolves symlinks, removes .., etc.)
+    let bin_dir_canonical = fs::canonicalize(&bin_dir).unwrap_or_else(|_| bin_dir.clone());
+
+    // Check if already in PATH with normalized comparison
+    let current_path = env::var("PATH").unwrap_or_default();
```
Byron, line 44: Must always be `var_os`. Can then be interpreted as bytes and worked with ASCII assumption for characters of interest.

crates/but-installer/src/shell.rs
```diff
@@ -0,0 +1,528 @@
...
+        } else if p.starts_with("$HOME/") {
+            home_dir.join(p.trim_start_matches("$HOME/"))
+        } else {
+            PathBuf::from(p)
+        };
+
+        // Normalize by removing trailing slashes and canonicalizing
+        let normalized = fs::canonicalize(&expanded).unwrap_or_else(|_| {
+            // If path doesn't exist yet, at least normalize the string
+            let path_str = expanded.to_string_lossy();
```
Byron, line 62: All this stringification should really be avoided.
Probably OK in most cases, but it's ultimately something to improve.

crates/but-installer/README.md
```diff
@@ -0,0 +1,42 @@
+# but-installer
+
+A lightweight Rust installer for GitButler. Currently supports macOS only.
```
Byron, line 3: I like these READMEs!

crates/but-installer/README.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+## Building
+
+```bash
+cargo build --release -p but-installer
+cargo test -p but-installer
+```
+
+The binary is optimized for size (~1.0MB) using system libcurl instead of bundling an HTTP client. Keeping this installer slim is a priority since it is downloaded before every installation.
```
Byron, line 42: I would remove the size (1MB), or maybe call it aspirational.

Byron (approved):
This is massive, in many ways :)!

And strange I'd say this, it's nice to have less bash!

Good to go, but… I didn't even run it, and didn't test it, and skipped over `shell.rs` which might be the thing that edits (and potentially destroys) user's configuration files.

#### Some *Flags*

* if the installation fails, I wouldn't trust the auto-cleanup there. Too easy to forget what was changed. A lot of code that deals with this could probably be removed if that cleanup handling was a little more 'automated'
* should really use `std::env::var_os` and not stringify paths

#### Some notes

* skimmed through `shell.rs` really fast, I probably missed everything that I shouldn't have.
* it takes about 1m to run CI of it, that's great!
* the installer can be smaller with [more build flags](https://github.com/johnthagen/min-sized-rust). We can put more compile wallclock time in it as well without slowing down CI  in general.
* the installer code is probably good enough, and I will refrain from giving it a refactor as well 😅
* `curl` is definitely a good choice for size. But I don't know if relying on system installed libraries is good for an installer. Probably not.

PR #12042, opened 2026-01-27
id: refactor to reduce number of structs

jonathantanmy2:
Why I'm not automerging: this introduces a new direct use of a crate (`self_cell`). It is already used indirectly by `rusqlite`, so adding it does not increase binary size or anything like that. It also introduces self-referential structs, which is something I think is really needed by `IdMap`, but I wanted to check with others to see what they think about it. (The main alternative design I see people talking about online is to use `Rc` anywhere, but that is a lot of clutter, and is also incompatible with the current algorithm to generate the stacks; the current algorithm generates a skeleton and then fills it in, which is incompatible with `Rc`, because once you create an `Rc`, you cannot modify it.)

This PR came about because I found #12040 much harder to write than expected due to some technical debt, so I thought I had better do some cleanups to avoid this problem in the future.


Byron (approved):
This is pretty neat, thanks for introducing this technique. In my 11 (?) years with Rust I never did self-referential directly, but if I'd do it then `self_cell` would be the way. It seems trustworthy.

Also, besides a manageable API, `IndexedStacks` acts like a normal struct that can be moved safely.

jonathantanmy2:
Thanks! Yes, `IndexedStacks` can be moved safely (all the "important" stuff is on the heap and must be accessed through helper methods).

PR #12049, opened 2026-01-27
deduplicate storage of access token

estib-vega:
The token is already set by the login process, so there’s no need to store it twice


estib-vega:
cc @Caleb-T-Owens 
Just as a heads-up, since we already store the user access one way, we can completely forgo storing it separately

PR #12054, opened 2026-01-28
Provide SKILL.md for the `but` CLI as well as a `but skill install` command

krlvi:
The idea here is to have a coding agent skill that we can keep up to date as the CLI evolves with new commands. 
The `crates/but/skill` folder should contain everything one needs to get started manually, but also for convenience, the app provides a `but skill install` command which installs the most up-to-date skill.

@aheritier i saw your project here https://github.com/aheritier/boost-your-ai/ (really cool btw ❤️) and wanted to give you a ping on this PR. 

I am not fully sure what the exact good practices are for creating a good `SKILL` - feedback is appreciated. By doing it here, we will be able to keep it up to date. Feel free to pull code from here for https://github.com/aheritier/boost-your-ai if it is useful and let me know if we can do things in a better way

### DIFF
--- a/crates/but/src/lib.rs
+++ b/crates/but/src/lib.rs
@@ -250,6 +250,27 @@ async fn match_subcommand(
                 }
             }
         }
+        Subcommands::Skill(args::skill::Platform { cmd }) => {
+            // For global installs or absolute paths, we don't need to be in a git repository
+            // For --infer without --global, we try to get repo context but don't require it
+            let needs_repo = match &cmd {
+                args::skill::Subcommands::Install {
+                    global,
+                    path,
+                    infer,
+                } => {
+                    !global
+                        && !infer
+                        && path
+                            .as_ref()
+                            .is_none_or(|p| !std::path::Path::new(p).is_absolute())
+                }
+            };
+
+            let ctx = but_ctx::Context::discover(&args.current_dir);
+            let mut ctx = if needs_repo { Some(ctx?) } else { ctx.ok() };
+            command::skill::handle(ctx.as_mut(), out, cmd).emit_metrics(metrics_ctx)
+        }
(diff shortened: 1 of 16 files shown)


crates/but/src/lib.rs
```diff
@@ -250,6 +250,36 @@ async fn match_subcommand(
...
+                command::skill::handle(Some(&mut ctx), out, cmd).emit_metrics(metrics_ctx)
+            } else {
+                // For global installs, try to get context but don't fail if not in a repo
+                match but_ctx::Context::discover(&args.current_dir) {
+                    Ok(mut ctx) => {
+                        command::skill::handle(Some(&mut ctx), out, cmd).emit_metrics(metrics_ctx)
+                    }
+                    Err(_) => command::skill::handle(None, out, cmd).emit_metrics(metrics_ctx),
+                }
+            }
```
penso:
```Rust
let ctx = but_ctx::Context::discover(&args.current_dir);
let mut ctx = if needs_repo { Some(ctx?) } else { ctx.ok() };
command::skill::handle(ctx.as_mut(), out, cmd).emit_metrics(metrics_ctx)
```

krlvi: fixed :)

aheritier:
perfect @krlvi exactly what we needed

aheritier:
FYI I added a deprecation warning : https://github.com/aheritier/boost-your-ai/pull/8

krlvi:
@aheritier If you observe the skill misbehaving or in need of tweaking, let us know. Since publishing this PR, I noticed that the agents were getting confused with `but absorb` causing it to absorb _everything_ which may not be desirable while running multiple agents in parallel. 

Overall I expect that as we experiment and learn, there will be more learnings and tweaks to the skill

PR #12061, opened 2026-01-28
logged in page

estib-vega:
If there’s a build type provided, allow the user to directly open their
client through deeplinks, removing the need to copy the access token.

The button is only shown when logging in from the Release or Nightly clients. If that's not the case, the 'Copy access token' button is shown

### DIFF
--- a/apps/web/src/routes/(app)/loggedin/+page.svelte
+++ b/apps/web/src/routes/(app)/loggedin/+page.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
 	import { goto } from '$app/navigation';
+	import { page } from '$app/state';
 	import FullscreenUtilityCard from '$lib/components/service/FullscreenUtilityCard.svelte';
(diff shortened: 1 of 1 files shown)


PavelLaptev:
@estib-vega what will happen if I click "Open client" button?
What will happen with this section 
image

estib-vega:
> @estib-vega what will happen if I click "Open client" button?

In macos, you'll get an OS confirmation modal asking to open the GitButler application.
If accepted, the client will open and you'll already be logged in. No need to copy and paste anything

estib-vega:
> > @estib-vega what will happen if I click "Open client" button?
> 
> In macos, you'll get an OS confirmation modal asking to open the GitButler application. If accepted, the client will open and you'll already be logged in. No need to copy and paste anything

That section will display the logged in user

PavelLaptev (approved):
@estib-vega thanks, lgtm.

PR #12071, opened 2026-01-28
Add Claude Rust SDK integration with feature flag

mtsgrd:
Add experimental support for using the Rust SDK (claude-agent-sdk-rs)
instead of spawning the Claude binary. The SDK implementation handles
session management, MCP server configuration, system prompts, streaming
responses, and user cancellation. A toggle in project Agent Settings
controls which backend is used.

---
- <kbd>&nbsp;2&nbsp;</kbd> #12084 
- <kbd>&nbsp;1&nbsp;</kbd> #12071 👈

### DIFF
--- a/crates/but-rebase/src/graph_rebase/creation.rs
+++ b/crates/but-rebase/src/graph_rebase/creation.rs
@@ -166,7 +166,11 @@ impl GraphExt for Graph {
                     {
                         // The commit has preserved parents, so the parent
                         // connections don't really matter.
-                    } else {
+                        // Further, the but graph does some _fancy_ stuff with
+                        // the workspace commit so we can ignore divergance.
+                        // It will result in the workspace commit getting
+                        // rewritten, but that is _fine_.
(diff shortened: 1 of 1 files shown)


crates/but-claude/src/bridge.rs
```diff
@@ -320,6 +327,323 @@ impl Claudes {
...
+
+            let rule = {
+                list_claude_assignment_rules(&ctx)?
+                    .into_iter()
+                    .find(|rule| rule.stack_id == stack_id)
+            };
+
+            let session_id = rule.map(|r| r.session_id).unwrap_or(uuid::Uuid::new_v4());
+            let session = upsert_session(&mut ctx, &repo, &workspace, session_id, stack_id)?;
+            let project_workdir = sync_ctx.legacy_project.worktree_dir()?.to_owned();
```
Byron: Let's not let it use `legacy_project` to get `worktree_dir()`, there is a function on the `ctx` itself.

Also, I am working on getting rid of these old things now as AI keeps making new code with them.

PR #12074, opened 2026-01-29
Questions about teardown and setup workflows

jonathantanmy2:
I was looking at teardown and setup workflows and have these questions:

 - Is it correct to have `gitbutler/workspace` and `gitbutler/target` refs still around even after teardown? See `multiple_branches_preserves_state`. I was expecting them to disappear. (All the other branches still exist, which is good.)
 - When running setup after teardown, a message "GitButler project is already set up!" is printed. See `setup_after_teardown_and_new_commit`. Is this expected? There was a change to the `gitbutler/workspace` branch incorporating the new commit  (which was what I hoped for, which is good).

I can take a look at fixing any of these if need be.

### DIFF
--- a/crates/but/tests/but/command/setup.rs
+++ b/crates/but/tests/but/command/setup.rs
@@ -596,3 +596,91 @@ More info: https://docs.gitbutler.com/workspace-branch
 
     Ok(())
 }
+
+#[test]
+fn setup_after_teardown_and_new_commit() -> anyhow::Result<()> {
+    let env = Sandbox::init_scenario_with_target_and_default_settings("two-stacks")?;
+    env.setup_metadata(&["A", "B"])?;
+
+    // Run teardown
+    env.but("teardown").assert().success();
+
+    // Create a new commit on branch A, and verify that the commit graph is what
+    // we expect
+    std::process::Command::new("git")
+        .env("GIT_AUTHOR_DATE", "1675176957 +0100")
+        .env("GIT_COMMITTER_DATE", "1675176957 +0100")
+        .arg("-C")
+        .arg(env.projects_root())
+        .arg("commit")
+        .arg("--allow-empty")
+        .arg("-m")
+        .arg("New commit on branch A")
+        .output()?;
+    insta::assert_snapshot!(env.git_log()?, @r"
+    * 3b447f7 (HEAD -> A) New commit on branch A
+    | *   c128bce (gitbutler/workspace) GitButler Workspace Commit
+    | |\  
+    | |/  
+    |/|   
+    * | 9477ae7 add A
+    | * d3e2ba3 (B) add B
+    |/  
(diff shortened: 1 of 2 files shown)


jonathantanmy2:
Hmm..."New commit on branch A" doesn't appear when the CI runs the test, although it appears on mine. In any case, I think that my questions still stand, so I'll leave this alone for now.

crates/but/tests/but/command/setup.rs
```diff
@@ -592,6 +592,94 @@ To undo these changes and return to normal Git mode, either:
...
+fn setup_after_teardown_and_new_commit() -> anyhow::Result<()> {
+    let env = Sandbox::init_scenario_with_target_and_default_settings("two-stacks")?;
+    env.setup_metadata(&["A", "B"])?;
+
+    // Run teardown
+    env.but("teardown").assert().success();
+
+    // Create a new commit on branch A, and verify that the commit graph is what
+    // we expect
+    std::process::Command::new("git")
```
Byron, line 610: Isn't there `but_testsupport::git()` for this? If it doesn't work for you, can it be made to work?

Byron (commented):
It really does sound a test is missing to setup after a teardown, so fixing issues that arise there should definitely be fixed.

And indeed, I'd also think that no gitbutler refs remain, particularly if `.git/gitbutler` doesn't remain.

crates/but/tests/but/command/setup.rs
jonathantanmy2: Ah yes there is. I've used it in #12085

jonathantanmy2:
> It really does sound a test is missing to setup after a teardown, so fixing issues that arise there should definitely be fixed.
> 
> And indeed, I'd also think that no gitbutler refs remain, particularly if `.git/gitbutler` doesn't remain.

`.git/gitbutler` remains, probably by design (we make a snapshot upon `teardown`), but I still think that no gitbutler refs should remain. I've sent a PR for that (#12085) and will close this PR.

Byron:
> .git/gitbutler remains, probably by design (we make a snapshot upon teardown), but I still think that no gitbutler refs should remain.

I don't even know if it's wise to leave data there to rot (unless the dir is empty), so intuitively I'd want to delete that as well.
Thinking about it: doing so may need some hygiene on Windows as we have to close all files in there (e.g. database). And even then Windows actually cleans up its locks with delay so immediate deletion attempts can still fail 🤦‍♂️. Anyway, I'd hope for the best there 😅.

schacon:
I addressed this in the other PR, but the general assumption should be that people setup and teardown multiple times, not once each. Teardown shouldn't mean obliterate and abandon, it should mean "go back to the Git way until I want to come back to the GitButler way", but we shouldn't delete possibly useful information (oplog, sqlite data, etc) without reason.

PR #12085, opened 2026-01-30
Delete GitButler branches upon `teardown`

jonathantanmy2:
Why I'm not automerging: this might be a workflow change, although I think it's misleading for the `gitbutler/workspace` and `gitbutler/target` branches to stick around, so I think it's better for them to be removed. (I've included a test to show that `setup` after `teardown` still works even if we remove the branches.)

This is a followup from #12074

### DIFF
--- a/crates/but/src/command/legacy/teardown.rs
+++ b/crates/but/src/command/legacy/teardown.rs
@@ -2,7 +2,7 @@ use anyhow::Context as _;
 use but_ctx::Context;
 use but_workspace::legacy::StacksFilter;
 use colored::Colorize;
-use gix::refs::transaction::PreviousValue;
+use gix::refs::transaction::{Change, PreviousValue, RefEdit, RefLog};
 use serde::Serialize;
 
 use crate::utils::OutputChannel;
@@ -177,6 +177,41 @@ pub(crate) fn teardown(ctx: &mut Context, out: &mut OutputChannel) -> anyhow::Re
             out,
             "  {}",
             format!("✓ Checked out: {}", target_branch_name).green()
+        )?;
+        writeln!(out)?;
+    }
+
+    // Remove gitbutler branches
+    if let Some(out) = out.for_human() {
+        writeln!(
+            out,
+            "{}",
+            "→ Removing GitButler workspace artifacts...".dimmed()
+        )?;
+    }
+    repo.edit_references([
+        RefEdit {
+            change: Change::Delete {
+                expected: PreviousValue::Any,
+                log: RefLog::AndReference,
+            },
+            name: "refs/heads/gitbutler/workspace".try_into()?,
+            deref: false,
+        },
+        RefEdit {
+            change: Change::Delete {
+                expected: PreviousValue::Any,
+                log: RefLog::AndReference,
+            },
+            name: "refs/heads/gitbutler/target".try_into()?,
+            deref: false,
+        },
+    ])?;
+    if let Some(out) = out.for_human() {
+        writeln!(
+            out,
+            "  {}",
+            "✓ Removed GitButler workspace artifacts".green()
         )?;
         writeln!(out)?;
     }
--- a/crates/but/tests/but/command/setup.rs
+++ b/crates/but/tests/but/command/setup.rs
@@ -602,6 +602,10 @@ fn setup_after_teardown_and_new_commit() -> anyhow::Result<()> {
     let env = Sandbox::init_scenario_with_target_and_default_settings("two-stacks")?;
     env.setup_metadata(&["A", "B"])?;
 
+    // Add a snapshot so that we can verify later that it is preserved across
+    // teardown/setup
+    env.but("but oplog snapshot -m foo").assert().success();
+
     // Run teardown
     env.but("teardown").assert().success();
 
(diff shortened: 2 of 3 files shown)


schacon:
I think it makes sense to remove them. :+1:

crates/but/src/command/legacy/teardown.rs
```diff
@@ -181,6 +181,32 @@ pub(crate) fn teardown(ctx: &mut Context, out: &mut OutputChannel) -> anyhow::Re
         writeln!(out)?;
     }
 
+    // Remove gitbutler branches
+    if let Some(out) = out.for_human() {
+        writeln!(
+            out,
+            "{}",
+            "→ Removing GitButler branches...".to_string().dimmed()
```
schacon: we might want to make this more explicit. when I first read it, I thought it meant removing the applied branches I had been working on. maybe something like "Removing GitButler workspace artifacts" or something that doesn't make people think we're deleting branches.

crates/but/src/command/legacy/teardown.rs
```diff
@@ -181,6 +181,32 @@ pub(crate) fn teardown(ctx: &mut Context, out: &mut OutputChannel) -> anyhow::Re
...
+        )?;
+    }
+    std::process::Command::new("git")
```
schacon: Do we need to shell out to Git for this? I would love to just use Gix when possible. Should be pretty easy to remove a few references.

One possible side effect is that we hide the oplog head in the `gitbutler/target` reflog so those objects aren't GC'd. Removing this may remove the reflog associated. It would be nice to try to keep the reflog while removing the `refs/heads/gitbutler/*` references.

crates/but/src/command/legacy/teardown.rs
jonathantanmy2: Technically we are deleting branches, but your wording sounds good. I don't know offhand whether the GitButler product uses American or British English throughout, so I'll use the American spelling ("artifact") as you have written instead of "artefact".

crates/but/src/command/legacy/teardown.rs
jonathantanmy2: It's more verbose to delete a reference in Gix due to its more careful design (which I agree with) - we have to specify that we don't care about the previous value of the ref, and that we want to delete both the ref and the reflog. But yes, I can switch to that.

There is the issue that I don't think we can keep the reflog if the ref is deleted, or at least, I don't know how. `git branch -D` is documented to delete the reflog:

> With a -d or -D option, <branch-name> will be deleted. You may specify more than one branch for deletion. If the branch currently has a reflog then the reflog will also be deleted.

and in Gix (https://docs.rs/gix-ref/0.58.0/gix_ref/transaction/enum.Change.html), the `RefLog` enum has only the options to delete the reflog, or the reference and the reflog (i.e. no "delete reference only").

If it's important to keep the reflog, should we move `gitbutler/target` to `refs/gitbutler-target-backup` (no `heads`, so not a branch) or something like that, and make sure that `setup` restores from this ref?

jonathantanmy2: Another design direction we could take is to make `teardown` truly switch to a regular Git repository, and thus it should not only delete `gitbutler/target` (including reflog) but also the `.git/gitbutler` directory (@Byron mentions this as a possibility here https://github.com/gitbutlerapp/gitbutler/pull/12074#issuecomment-3821692802 ).

schacon: I would rather not do this, because we _do_ want to make sure that switching back and forth is seamless. If we delete `.git/gitbutler` then we gain nothing because Git doesn't care about it but we simultaneously lose oplog history and any data in sqlite, etc. There is no reason to lose that as there is no gain - git does not complain about it (or know about it). If someone tears down and then runs setup again, it would be nice if we could rather seamlessly pick up where we left off. I just want teardown to switch to a normal branch.

Deleting the `gitbutler/*` references is a _little_ cleaner without losing data, so that's fine, but deleting useful data that is otherwise not obtrusive doesn't make much sense to me.

crates/but/src/command/legacy/teardown.rs
Byron: OMG, I never knew there was a difference, but that finally explains my constant confusion on how to write it. TIL

crates/but/src/command/legacy/teardown.rs
jonathantanmy2: > I would rather not do this, because we _do_ want to make sure that switching back and forth is seamless. If we delete `.git/gitbutler` then we gain nothing because Git doesn't care about it but we simultaneously lose oplog history and any data in sqlite, etc. There is no reason to lose that as there is no gain - git does not complain about it (or know about it). If someone tears down and then runs setup again, it would be nice if we could rather seamlessly pick up where we left off. I just want teardown to switch to a normal branch.

Makes sense.

> Deleting the `gitbutler/*` references is a _little_ cleaner without losing data, so that's fine, but deleting useful data that is otherwise not obtrusive doesn't make much sense to me.

I don't think we can delete the references without also deleting their reflogs (as mentioned above in https://github.com/gitbutlerapp/gitbutler/pull/12085#discussion_r2746964459), so we might lose data if the user GCs outside whatever grace period is configured in Git (default is 2 weeks). If we proceed with deleting the references, I see a few ways to proceed:
 - delete them, and make sure that `setup` works even if the oplog heads etc. were GCed (I can add a test)
 - upon *teardown*, move `gitbutler/target` to `refs/gitbutler-target-backup` instead of deleting it, forcibly overwriting the existing reference ("forcibly" so that a second teardown will still succeed)
 - upon *setup*, forcibly delete `refs/gitbutler-target-backup` and upon teardown, move `gitbutler/target` to `refs/gitbutler-target-backup` first checking that it doesn't exist

I don't really like the options that forcibly delete data because they might end up deleting data that the end user has created. We could store what the ref target is supposed to be (in `.git/gitbutler` somewhere) and only delete the ref if it matches, but this might be too complicated for such a corner case.

I'm personally inclined to doing the first option above (deleting the branch outright and making sure we can recover from it) at least because I don't think that the end user would `teardown` and then, after a significant period of time has elapsed, `setup` (possibly with a new version of the CLI, and possibly with format changes in `.git/gitbutler`, which means that we need to ignore any existing information anyway). What do you think?

jonathantanmy2:
I've included a demo of what the oplog looks like after teardown-gc-setup (it's not preserved, but things seem to work).

jonathantanmy2:
I think there is still an issue outstanding about how we should preserve the reflog of `gitbutler/target` if we do delete that branch, but in any case, I now think that deleting the GitButler branches is less useful (from https://github.com/gitbutlerapp/gitbutler/pull/12146#issuecomment-3845124241 , it seems that the GitButler metadata takes precedence over the current commit graph even after `teardown`, so the commit graph manipulations we can do are limited anyway), so I'll close this PR.

PR #12091, opened 2026-01-30
Add a clean script

Caleb-T-Owens:
Introduce a script for cleaning out any build artifacts or generated artifacts.

This is not a safe script, but it warns the user and asks for confirmation so I think there is sufficient warning.


Caleb-T-Owens:
The globstar option doesn't seem to exist? This works fine on my computer.

PR #12102, opened 2026-01-31
Handle heredocs when splitting bash commands

mtsgrd:
Extend split_bash_commands to recognize << and <<- heredocs, preventing
the parser from incorrectly splitting on newlines or operators within
heredoc content.

The implementation tracks the heredoc delimiter and current line in an
InHeredoc state, matching the closing delimiter at line boundaries or
before operators. For <<- heredocs, leading tabs on the closing line
are stripped when matching.

### DIFF
--- a/crates/but-claude/src/permissions/bash.rs
+++ b/crates/but-claude/src/permissions/bash.rs
@@ -1,5 +1,5 @@
-/// Splits a bash command string into individual commands, respecting quotes and escape sequences.
-/// Splits on: &&, ||, ;, |, &, and newlines (when not inside quotes)
+/// Splits a bash command string into individual commands, respecting quotes, escape sequences,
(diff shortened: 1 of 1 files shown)


mtsgrd:
@Caleb-T-Owens I've had this applied in my dev instance, and here's some new permissions that have been added:
```json
{
  "permissions": {
    "allow": [
      "Bash(cargo:*)",
      "Bash(1:*)",
      "Bash(/Users/mattias/c/gitbutler/target/debug/but:*)",
      "Bash(pnpm check 2>)",
      "Bash(1)",
      "Bash(head -50)",
      "Bash(pnpm check 2>)",
      "Bash(1)",
      "Bash(tail -100)"
    ]
  }
}
```
Two things stand out, `Bash(1)` and `Bash(pnpm check 2>)`. The origin of the first one is confusing, and the latter should probably cut the mangled redirect. Wdyt?

Caleb-T-Owens (approved):

PR #12112, opened 2026-02-01
Log claude commands only after clearing its env

Byron:
They may contain confidential information like ANTHROPIC_AUTH_TOKENs.


Byron:
CC @krlvi and @Caleb-T-Owens.

Also, is there a cure for this issue?

Screenshot 2026-02-01 at 16 40 33

PR #12115, opened 2026-02-01
Make restore an oplog subcommand

krlvi:
Change the CLI and docs so restore is invoked as `but oplog restore`
instead of the top-level `but restore`. This updates help text,
reference docs, examples, argument parsing, command dispatch, legacy
oplog messages, and metrics so that the restore action is nested under
the oplog subcommand and routed to the existing oplog restore
implementation.


krlvi:
@schacon wdyt?

PR #12120, opened 2026-02-01
Deduplicate commit parents to handle malformed merges

mtsgrd:
Handle cases where commits contain duplicate parent IDs (produced by
older GitButler code) by filtering parent_ids through a HashSet before
constructing Commit values. This prevents mismatched parent lists that
caused errors like "Parents for commit ... do not match" during
operations such as commit_move_changes_between. The change applies
deduplication in two places: when building a Commit fallback in walk.rs
and when assembling parent_ids for newly created Commit values.

### DIFF
--- a/crates/but-graph/src/init/walk.rs
+++ b/crates/but-graph/src/init/walk.rs
@@ -585,12 +585,22 @@ impl TraverseInfo {
             .collect();
         Ok(match self.commit {
             Some(commit) => Commit { refs, flags, ..commit },
-            None => Commit {
-                id: self.inner.id,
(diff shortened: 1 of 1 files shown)


Byron (changes requested):
In general, we can't change the order of the commits even if deduplicated.
I think a test should be very possible as well, and if Git doesn't want to do it, `git hash-object` will.

Lastly, while pondering over this a little, I came up with "it should be good to change this here as it should never be intended to have such duplicates.

Blocking merge while test is missing, and while I would hope that the deduplication might only have to happen in one of these spots.

krlvi:
do we still need this PR?

PR #12121, opened 2026-02-01
Check remote tracking branches when generating canned names

mtsgrd:
Ensure generated canned branch names avoid conflicts with remote
tracking refs as well as local branches. Add new functions in but-core
to find unique refnames with an explicit remote check and a canned-ref
wrapper that uses it. Update but-api to consult the project's default
push remote and use the remote-aware generation when available. Include
tests covering remote-aware behavior to ensure names that exist on
remotes are skipped.

### DIFF
--- a/crates/but-api/src/legacy/workspace.rs
+++ b/crates/but-api/src/legacy/workspace.rs
@@ -593,10 +593,22 @@ pub fn stash_into_branch(
 
 /// Returns a new available branch name based on a simple template - user_initials-branch-count
 /// The main point of this is to be able to provide branch names that are not already taken.
+/// This checks both local branches and remote tracking branches to avoid conflicts with branches
+/// that exist on the remote but haven't been checked out locally.
 #[but_api]
 #[instrument(err(Debug))]
 pub fn canned_branch_name(ctx: &Context) -> Result<String> {
-    let rn = but_core::branch::unique_canned_refname(&*ctx.repo.get()?)?;
+    let repo = ctx.repo.get()?;
+    let state = VirtualBranchesHandle::new(ctx.project_data_dir());
+
+    // Try to get the push remote name from the default target, fall back to just checking local refs
+    let rn = match state.get_default_target() {
+        Ok(target) => {
+            let remote_name = target.push_remote_name();
+            but_core::branch::unique_canned_refname_with_remote_check(&repo, &remote_name)?
+        }
+        Err(_) => but_core::branch::unique_canned_refname(&repo)?,
+    };
     Ok(rn.shorten().to_string())
 }
 
--- a/crates/but-core/src/branch/generate.rs
+++ b/crates/but-core/src/branch/generate.rs
@@ -74,6 +74,71 @@ pub fn unique_canned_refname(repo: &gix::Repository) -> anyhow::Result<gix::refs
     find_unique_refname(repo, name.as_ref())
 }
 
+/// Like [`find_unique_refname`], but also checks that the name doesn't exist as a remote tracking branch.
+///
+/// The `remote_name` should be the name of the remote to check against (e.g., "origin").
+/// This checks both local refs (`refs/heads/<name>`) and remote tracking refs (`refs/remotes/<remote>/<name>`).
+pub fn find_unique_refname_with_remote_check(
+    repo: &gix::Repository,
+    template: &gix::refs::FullNameRef,
+    remote_name: &str,
+) -> anyhow::Result<gix::refs::FullName> {
+    let short_name = template.shorten();
+    let category = template
+        .category()
+        .with_context(|| format!("Input branch {template} could not be categorized"))?;
+
+    let ref_exists = |name: &bstr::BStr| -> anyhow::Result<bool> {
+        let local_ref = category.to_full_name(name)?;
+        if repo.try_find_reference(&local_ref)?.is_some() {
+            return Ok(true);
+        }
+        let remote_ref = format!("refs/remotes/{remote_name}/{}", name.to_str_lossy());
+        Ok(repo.try_find_reference(remote_ref.as_str())?.is_some())
+    };
+
+    // Check if the original name is available
+    if !ref_exists(short_name)? {
+        return Ok(template.to_owned());
+    }
+
+    // Extract base name and number if it already has a numerical suffix
+    let trailing_number =
+        short_name
+            .rfind("-")
+            .map(|pos| (&short_name[pos + 1..], pos))
+            .and_then(|(maybe_num, pos)| {
+                let num = maybe_num.to_str().ok()?.parse::<usize>().ok()?;
+                Some((&short_name[..pos], num + 1))
+            });
+    let (base, start_num) = trailing_number.unwrap_or((short_name, 1));
+
+    // Try incrementing numbers until we find one that doesn't exist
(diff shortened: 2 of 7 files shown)


crates/but-core/src/branch/generate.rs
```diff
@@ -74,6 +74,71 @@ pub fn unique_canned_refname(repo: &gix::Repository) -> anyhow::Result<gix::refs
...
 }
 
+/// Like [`find_unique_refname`], but also checks that the name doesn't exist as a remote tracking branch.
+///
+/// The `remote_name` should be the name of the remote to check against (e.g., "origin").
+/// This checks both local refs (`refs/heads/<name>`) and remote tracking refs (`refs/remotes/<remote>/<name>`).
+pub fn find_unique_refname_with_remote_check(
+    repo: &gix::Repository,
+    template: &gix::refs::FullNameRef,
+    remote_name: &str,
```
Byron, line 84: I think there is a type for that, probably `gix::remote::Name`.

crates/but-core/src/branch/generate.rs
```diff
@@ -74,6 +74,71 @@ pub fn unique_canned_refname(repo: &gix::Repository) -> anyhow::Result<gix::refs
...
+    let category = template
+        .category()
+        .with_context(|| format!("Input branch {template} could not be categorized"))?;
+
+    let ref_exists = |name: &bstr::BStr| -> anyhow::Result<bool> {
+        let local_ref = category.to_full_name(name)?;
+        if repo.try_find_reference(&local_ref)?.is_some() {
+            return Ok(true);
+        }
+        let remote_ref = format!("refs/remotes/{remote_name}/{}", name.to_str_lossy());
```
Byron, line 96: Can't use `to_str_lossy()` because we can't loose information here.

crates/but-core/src/branch/generate.rs
```diff
@@ -74,6 +74,71 @@ pub fn unique_canned_refname(repo: &gix::Repository) -> anyhow::Result<gix::refs
...
+    repo: &gix::Repository,
+    template: &gix::refs::FullNameRef,
+    remote_name: &str,
+) -> anyhow::Result<gix::refs::FullName> {
+    let short_name = template.shorten();
+    let category = template
+        .category()
+        .with_context(|| format!("Input branch {template} could not be categorized"))?;
+
+    let ref_exists = |name: &bstr::BStr| -> anyhow::Result<bool> {
```
Byron, line 91: I think it will help the code to take a `FullNameRef` here.

crates/but-core/src/branch/generate.rs
```diff
@@ -74,6 +74,71 @@ pub fn unique_canned_refname(repo: &gix::Repository) -> anyhow::Result<gix::refs
...
+        let remote_ref = format!("refs/remotes/{remote_name}/{}", name.to_str_lossy());
+        Ok(repo.try_find_reference(remote_ref.as_str())?.is_some())
+    };
+
+    // Check if the original name is available
+    if !ref_exists(short_name)? {
+        return Ok(template.to_owned());
+    }
+
+    // Extract base name and number if it already has a numerical suffix
```
Byron, line 105: This looks like a duplicate, let's let there be only one way of generating these names.

crates/but-api/src/legacy/workspace.rs
```diff
@@ -593,10 +593,22 @@ pub fn stash_into_branch(
...
 /// Returns a new available branch name based on a simple template - user_initials-branch-count
 /// The main point of this is to be able to provide branch names that are not already taken.
+/// This checks both local branches and remote tracking branches to avoid conflicts with branches
+/// that exist on the remote but haven't been checked out locally.
 #[but_api]
 #[instrument(err(Debug))]
 pub fn canned_branch_name(ctx: &Context) -> Result<String> {
-    let rn = but_core::branch::unique_canned_refname(&*ctx.repo.get()?)?;
+    let repo = ctx.repo.get()?;
+    let state = VirtualBranchesHandle::new(ctx.project_data_dir());
```
Byron, line 602: Seeing this showing up just tells me I have to work on getting rid of these handles so it won't be picked up anymore.

Technically it's legacy, but that doesn't mean we can keep writing insta-legacy code. Instead it means it should become modern.

Here one would use `let (_guard, repo, ws, _) = ctx.workspace_and_db()?` and ask `ws` for the target information.

crates/but-core/tests/core/branch/find_unique_refname.rs
```diff
@@ -55,3 +55,91 @@ fn returns_original_if_unique() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+mod with_remote_check {
+    use super::*;
+
+    #[test]
+    fn avoids_remote_tracking_branches() -> anyhow::Result<()> {
+        let (repo, _tmp) = writable_scenario("unborn-empty");
```
Byron, line 64: It's my own fault to see this repeated, but in the final review I'd definitely change the pattern of using writable fixtures for read-only functions, and instead create a fixture just for it which is presented before beforehand via `but-testsupport`.

crates/but-rules/src/handler.rs
```diff
@@ -180,9 +185,21 @@ fn create_stack(
...
     meta: &mut impl but_core::RefMetadata,
     _perm: &mut RepoExclusive,
+    project_data_dir: &Path,
 ) -> anyhow::Result<(StackId, but_graph::projection::Workspace)> {
     use anyhow::Context;
-    let branch_name = but_core::branch::unique_canned_refname(repo)?;
+    use gitbutler_stack::VirtualBranchesHandle;
+
+    // Try to get the push remote name to avoid conflicts with remote branches
+    let state = VirtualBranchesHandle::new(project_data_dir);
```
Byron, line 194: We must use `ws` to check the target information.

crates/but-rules/Cargo.toml
```diff
@@ -17,6 +17,7 @@ but-graph.workspace = true
 but-workspace = { workspace = true, features = ["legacy"] }
 but-hunk-dependency.workspace = true
 but-ctx.workspace = true
+gitbutler-stack.workspace = true
```
Byron, line 20: Just yesterday I removed this 😭🤣.

Byron (changes requested):
It turns out I have a lot of comments, but also learned a lot on what still needs to be done to have an "AI-safe" codebase. `VirtualBranchesHandle`, I am looking at you :P.

A couple of things I think could address before I give it the finishing touches.

PS: Something I don't understand it where conflicts are coming from. The unique names are unique, so `refs/heads/feat-1` will not conflict. And if the problem is that there may be `refs/remotes/origin/feat-1` which then seems to be the remote tracking branch of `refs/heads/feat-1`, which will then be inferred by `but-graph` logic as local/remote tracking branch pair… then I think at least at some point in the future we should consider to remove that `but-graph` logic. After all, it was just added to deal with GitButler not setting up tracking previously, but when the new `apply` is used, that will now happen as well. So I think this context I'd also add to the docs of the function to mark it clearly as "shoudln't be needed once `but-graph` changed its ways.

mtsgrd:
Thanks @Byron, I'll address these comments and mark it ready for review when I get there!

krlvi:
do we still need this PR?

PR #12125, opened 2026-02-02
add a little ascii output on setup

schacon:
and a mini cheat-sheet. only when setting up a project.
also skip the pager

CleanShot 2026-02-02 at 11 30 09@2x


krlvi:
X BUT :D

Byron:
XBut?
I only ever see XBut there 😭.

schacon:
yeah, I wanted to have it in there for testing. we can replace the ascii art with a better bowtie when someone has some

PR #12136, opened 2026-02-02
Add network status monitoring service and UI

mtsgrd:
Introduce NetworkStatusService that monitors browser online/offline events and exposes reactive network status. Add NetworkStatus component that displays an offline banner when connectivity is lost. This helps users understand when network-dependent features may be unavailable.


mtsgrd:
@estib-vega this is the thing I mentioned earlier today, feel free to hijack if useful. I'll leave the pr in draft mode.

estib-vega:
Thanks!

krlvi:
do we still need this PR?

PR #12138, opened 2026-02-02
Also rename the frontend parameterto match the backend.

Byron:
This fixes a bug that was introduced yesterday when the backend paramete was unilaterally renamed.


Byron:
This was fixed in the meantime, and I wasn't aware this one didn't merge.

PR #12160, opened 2026-02-03
skill-version-check

krlvi:


crates/but/src/command/skill.rs
```diff
@@ -197,8 +225,266 @@ fn is_gitbutler_skill(skill_md_path: &std::path::Path) -> bool {
     }
 }
 
-/// Infer installation path by detecting existing skill installation
-fn infer_install_path(ctx: Option<&mut Context>, global: bool) -> Result<PathBuf> {
+/// Extract the version from an installed SKILL.md file's YAML frontmatter.
+/// Returns None if the file doesn't exist, isn't readable, or has no valid version.
+fn extract_installed_version(skill_md_path: &std::path::Path) -> Option<String> {
+    let content = std::fs::read_to_string(skill_md_path).ok()?;
```
Byron, line 231: Instead of testing on input strings directly, of course it has to read everything from disk into a string and test that instead 🤦‍♂️.
😭

Byron: I should have known better than to look at this 😁.

krlvi: some cleanup here https://github.com/gitbutlerapp/gitbutler/pull/12171

PR #12166, opened 2026-02-03
fix manpage generation

schacon:
In order to generate the manpages from the comments, we need the `but-clap` binary in Cargo and in order for the output to be valid MDX content, anything with backticks needs to be verbatium flagged.


Byron (approved):
Sorry for that, that was accidental, but it's also good to re-add what we really need.

PR #12168, opened 2026-02-03
Handle null platform URLs in releases

krlvi:
This fixes an issue with `curl -sSL https://gitbutler.com/install.sh | bash -s nightly`


krlvi:
@slarse FYI - I made this fix

PR #12189, opened 2026-02-04
fix(skill): Update GitHub Copilot skill path to correct location

nshcr:
https://docs.github.com/en/copilot/concepts/agents/about-agent-skills

Adjust the Copilot SKILL installation path to the correct location according to the official documentation.

`.github/copilot/skills/gitbutler` -> `.github/skills/gitbutler`

### DIFF
--- a/crates/but/src/command/skill.rs
+++ b/crates/but/src/command/skill.rs
@@ -98,7 +98,7 @@ const SKILL_FORMATS: &[SkillFormat] = &[
     SkillFormat {
         name: "GitHub Copilot",
         description: "GitHub Copilot skill format",
-        path: ".github/copilot/skills/gitbutler",
+        path: ".github/skills/gitbutler",
     },
     SkillFormat {
         name: "Cursor",


Byron:
Thanks a lot, great catch!

I validated this and… second Copilot here :D.

nshcr:
@Byron Thanks for merging this so quickly!

In fact, this PR only fixes the project-level Copilot skill installation path (`.github/skills`). For global installation, Copilot uses a different location (`.copilot/skills`), while `but skill install` currently doesn't provide separate path for installing to the global scope (it only appends to the home path for now).



So unfortunately, the global Copilot installation path is still incorrect. Since it involves more extensive changes, I may not be able to address it :(.

krlvi:
Thanks for this!

krlvi:
> @Byron Thanks for merging this so quickly!
> 
> In fact, this PR only fixes the project-level Copilot skill installation path (`.github/skills`). For global installation, Copilot uses a different location (`.copilot/skills`), while `but skill install` currently doesn't provide separate path for installing to the global scope (it only appends to the home path for now).
> 
> 
> So unfortunately, the global Copilot installation path is still incorrect. Since it involves more extensive changes, I may not be able to address it :(.

I will take care of this

krlvi:
https://github.com/gitbutlerapp/gitbutler/pull/12192

PR #12196, opened 2026-02-04
feat(gitbutler-git): Make askpass-machinery opt-in

slarse:
Using the askpass machinery for CLI is redundant as the terminal IO can simply be (and is by default) handed over to the child process that requires credentials. So we make it opt-in, s.t. the CLI can be compiled without it, and we can then distribute the CLI as a single, statically linked binary that does not depend on the `gitbutler-git-setsid` and `gitbutler-git-askpass` binaries.

Note to self: I need to double check how Windows uses the CLI, as I recall it doesn't use built-in but.

### DIFF
--- a/crates/gitbutler-git/Cargo.toml
+++ b/crates/gitbutler-git/Cargo.toml
@@ -25,6 +25,7 @@ serde = ["dep:serde"]
 tokio = ["dep:tokio"]
 ## a flag to indicate benchmarks are run
 benches = []
+## a flag to turn on the askpass machinery for Git commands (should be enabled for GUI, disabled for CLI)
 askpass = []
 
 [dependencies]
--- a/crates/gitbutler-git/Cargo.toml
+++ b/crates/gitbutler-git/Cargo.toml
@@ -25,6 +25,7 @@ serde = ["dep:serde"]
 tokio = ["dep:tokio"]
 ## a flag to indicate benchmarks are run
 benches = []
+askpass = []
 
 [dependencies]
 tracing.workspace = true
--- a/crates/gitbutler-git/src/repository.rs
+++ b/crates/gitbutler-git/src/repository.rs
@@ -1,12 +1,18 @@
-use std::{collections::HashMap, path::Path, time::Duration};
+use std::{collections::HashMap, path::Path};
 
-use futures::{FutureExt, select};
 use gix::bstr::ByteSlice;
-use rand::{Rng, SeedableRng};
 
 use super::executor::{AskpassServer, GitExecutor, Pid, Socket};
 use crate::RefSpec;
 
+#[cfg(feature = "askpass")]
+use futures::{FutureExt, select};
+#[cfg(feature = "askpass")]
+use rand::{Rng, SeedableRng};
+#[cfg(feature = "askpass")]
+use std::time::Duration;
+
+#[cfg(feature = "askpass")]
 /// The number of characters in the secret used for checking
 /// askpass invocations by ssh/git when connecting to our process.
 const ASKPASS_SECRET_LENGTH: usize = 24;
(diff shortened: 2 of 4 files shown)


crates/gitbutler-git/src/repository.rs
```diff
@@ -70,6 +76,31 @@ enum HarnessEnv<P: AsRef<Path>> {
 
 #[cold]
 async fn execute_with_auth_harness<P, F, Fut, E, Extra>(
+    harness_env: HarnessEnv<P>,
+    executor: &E,
+    args: &[&str],
+    envs: Option<HashMap<String, String>>,
+    // below arguments only used if askpass is enabled
+    mut _on_prompt: F,
+    _extra: Extra,
```
slarse, line 85: This is terrible design, as the callers of this will pass in arguments that are then just dropped if askpass is not enabled. This applies transitively through multiple crates.

The other solution I can come up with is to propagate the askpass feature flag through all of the crates all the way to the top. IMO this is better design and easy to delegate to an LLM to implement (i.e. costs me little time), but I'm unsure if it's worth duplicating every function along that path to account for the argument discrepancy. Thoughts?

slarse (commented):
Need some input :)

krlvi:
I think @Byron could have some good input here

crates/gitbutler-git/src/repository.rs
```diff
@@ -296,6 +298,84 @@ where
     }
 }
 
+#[cfg(not(feature = "askpass"))]
+/// Directly execute the Git command without invoking the askpass pipe machinery. This is useful
+/// for the CLI, as the child process simply inherits stdin from the parent process and therefore
+/// doesn't need the askpass mechanism.
+///
+/// It's doubly useful in the sense that the CLI then does not need the askpass and setsid
+/// binaries.
```
Byron, line 307: Loving that!

crates/gitbutler-git/src/repository.rs
Byron: I'd prefer to keep it easy and 'contained', as `gitbutler-git` is on my rewrite list of for a long time now and just worked well enough not become a problem, yet.
So putting a dead-horse in an ugly dress isn't really that bad 😁.

crates/gitbutler-git/Cargo.toml
```diff
@@ -25,6 +25,7 @@ serde = ["dep:serde"]
 tokio = ["dep:tokio"]
 ## a flag to indicate benchmarks are run
 benches = []
+askpass = []
```
Byron, line 29: Docs are amiss.

Byron (approved):
I think it's OK and I like its self-contained nature.

crates/gitbutler-git/Cargo.toml
slarse: Sorry, you mean that docs are missing? As in the askpass feature flag should be documented in what it does? :)

crates/gitbutler-git/src/repository.rs
slarse: I figured something like that might be the case, which is why I went with the ugly dress for a first stab at it :)

crates/gitbutler-git/Cargo.toml
slarse: Added docs, assuming that's what you meant!

krlvi:
:shipit:

krlvi:
>  I need to double check how Windows uses the CLI, as I recall it doesn't use built-in but.

Not sure if the windows cli works at all

slarse:
> > I need to double check how Windows uses the CLI, as I recall it doesn't use built-in but.
> 
> Not sure if the windows cli works at all

It does! Just not the Git command stuff as it's built standalone, so it hits the askpass not initialized panic. But with this fix that should also now work!

crates/gitbutler-git/Cargo.toml
Byron: Yes, I tried to write with fancy words 🤦‍♂️.
Actually, I tried to mimic you and do some creative writing, but… "ein Meister ist noch nicht vom Himmel gefallen" :D.

PR #12205, opened 2026-02-04
fix: restore remote tracking branch check in canned_branch_name

krlvi:
The canned_branch_name function was changed in 3b84179150 to use
but_core::branch::unique_canned_refname which only checks local
refs. This caused collisions with branches that exist on the remote.

This restores the original behavior by using Stack::next_available_name
which properly checks:
- GitButler patch references
- Local git references  
- Remote tracking branches

Also adds a test to prevent this regression in the future.

### DIFF
--- a/crates/but-api/src/legacy/workspace.rs
+++ b/crates/but-api/src/legacy/workspace.rs
@@ -14,7 +14,7 @@ use gitbutler_oplog::{
     entry::{OperationKind, SnapshotDetails},
 };
 use gitbutler_reference::{LocalRefname, Refname};
-use gitbutler_stack::{StackId, VirtualBranchesHandle};
+use gitbutler_stack::{Stack, StackId, VirtualBranchesHandle};
 use tracing::instrument;
 
 use crate::json::HexHash;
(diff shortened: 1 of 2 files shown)


krlvi:
@Byron FYI - the new function was incomplete

Byron:
Thanks for the heads-up, and sorry for that.

I think the issue was also solved separately by @mtsgrd, and I wasn't aware how much of an issue it is, also lost track of what's going on there.

My finding is that the problem is us picking up remote tracking branches for local tracking branches automatically, even without Git configuration.
That way, a newly created branch can wrongly be associated with an old and stale remote tracking branch, causing all kinds of funkiness. In future, with the new apply, we'd probably want to rely on Git exclusively, but until then… I have brought up a follow-up https://github.com/gitbutlerapp/gitbutler/pull/12227 just so the code doesn't have to take a step back.

krlvi:
what alternate solution is there?

Byron:
~~The 'modern' version fix is upcoming here: https://github.com/gitbutlerapp/gitbutler/pull/12227~~

Just realised I already wrote that.

The alternative is to make the remote-ref check in modern code and document why it's important right now. I'd think that won't always be the case when we become more Gitizen-like.

PR #12212, opened 2026-02-04
fix: use a unified file list sorting rule in both list and tree views

nshcr:
The previous file list component used different filename sorting rules in the tree view and list view.
This wasn't a major issue in the earlier changed files view, but it has become much more noticeable with the recently introduced all-in-one continuous files diff preview, which follows the same sorting order as the list view.

This PR fixes the issue by making both views use a unified sorting rule.

### DIFF
--- a/apps/desktop/src/lib/files/filetreeV3.ts
+++ b/apps/desktop/src/lib/files/filetreeV3.ts
@@ -7,6 +7,17 @@
 
 import type { TreeChange } from '$lib/hunks/change';
 
+/**
+ * Sort configuration for file lists.
+ */
+const FILE_SORT_CONFIG: Intl.CollatorOptions = {
+	numeric: true,
+	caseFirst: 'lower',
+	sensitivity: 'base'
+} as const;
+
+const FILE_SORT_LOCALE = 'en';
+
 export type TreeNode = {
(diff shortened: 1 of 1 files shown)


krlvi:
cc @mtsgrd

mtsgrd:
Amazing. Thank you! 🎉

mtsgrd:
Just one question, would FILE_SORT_LOCALE = undefined make it use the browsers locale? It might be better than explicitly setting `en`.

mtsgrd (approved):

mtsgrd:
@PavelLaptev fyi!

PavelLaptev:
@nshcr awesome :-) Thanks!

nshcr:
> Just one question, would FILE_SORT_LOCALE = undefined make it use the browsers locale? It might be better than explicitly setting `en`.

@mtsgrd Thanks for merging it!

I kept the locale set to `en` to avoid changing the existing code logic.
Yes, I think removing `FILE_SORT_LOCALE` and letting this parameter be `undefined` would be better.

For reference, in this PR (https://github.com/gitbutlerapp/gitbutler/pull/10517), I also used formatting rules that respect the user's locale, which might be helpful for the software's localization in the future.

PR #12223, opened 2026-02-04
but: removed `num_args` from `--changes`

rsteube:
## 🧢 Changes

Removed `num_args`.

## ☕️ Reasoning

This consumes multiple args until another flag is encountered. Thus mininterprets a following positional argument (branch) as change unless separated with dash `--`.

see https://docs.rs/clap/latest/clap/struct.Arg.html#method.num_args

```
esh ❯ but
GitButler uses metrics to help us know what is useful and improve it. Configure with `but config metrics`.
╭┄zz [unstaged changes]
┊   g0 A fileA
┊   h0 A fileB
┊
┊╭┄br [branch] (no commits)
├╯
┊
┴ 791af90 (common base) [gb-local/master] 2026-02-04 Initial empty commit (checked 4 minutes ago)

Hint: run `but diff` to see uncommitted changes and `but stage <file>` to stage them to a branch
```

```
esh ❯ but commit --changes fileA branch 
Error: Invalid file ID(s):
  'branch' is a branch but must be an uncommitted file or hunk
Exception: but exited with 1
  [tty 35]:1:1-34: but commit --changes fileA branch
```


krlvi:
hey @rsteube the idea for this was to be able to do `but commit --changes c1 c2 c3` - thus it always has to be the last arg (though it also supports `--changes="c1,c2,c3"`
Do you think another approach can be taken here?

rsteube:
@krlvi 
Using comma separated values or repeating the flag is the most common approach.
```sh
but commit --changes fileA,fileB
but commit --changes fileA --changes fileB
```
I think `num_args` (flag consuming multiple arguments) is generally not a good idea due to the aformentioned problems.
It's also rarely seen in commands (nix, some based on pythons argparse).

Some ideas below.
I think I'd keep the `--changes` flag (comma seperated) and additionally use dash args for changes(4).

1.
Use positional arguments for changes/target:
```sh
# but commit [change]... target
but commit branch
but commit fileA branch
but commit fileA fileB branch
```

In the completions both changes and branches would appear for each position though.

2.
Invert the usage of flags:
```sh
# but commit [-t target] [change]...
but commit -t branch
but commit -t branch fileA
but commit -t branch fileA fileB
```

3.
Use a bool flag to modify the behaviour of parsing positional arguments (as in 1).
Git does this overloading quite a bit and its rarely a good idea.

4.
Explicitly use dash args for files.
```sh
# but commit [target] [-- change...]
but commit branch
but commit branch -- fileA
but commit branch -- fileA fileB
but commit -- fileA fileB
```

krlvi (approved):
Okay now i understand the considerations. Thank you for this contribution

PR #12232, opened 2026-02-05
git settings: Add gitlab project settings + export type

estib-vega:
- Add the GitLab project and upstream project IDs to the git settings
- Export the git config type and use it in the git config service


crates/but-core/src/settings.rs
```diff
@@ -20,16 +22,27 @@ pub mod git {
...
         #[derive(Debug, PartialEq, Clone, Default, serde::Serialize, serde::Deserialize)]
+        #[cfg_attr(feature = "export-ts", derive(ts_rs::TS))]
         #[serde(rename_all = "camelCase")]
+        #[cfg_attr(feature = "export-ts", ts(export, export_to = "./settings/gitConfigSettings.ts"))]
         #[expect(missing_docs)]
         pub struct GitConfigSettings {
             #[serde(rename = "signCommits")]
             pub gitbutler_sign_commits: Option<bool>,
             pub gitbutler_gerrit_mode: Option<bool>,
+            #[cfg_attr(feature = "export-ts", ts(type = "string | null"))]
```
Byron, line 33: ~~While having seen that before, it's strange to me that `ts-rs` won't do this automatically. Are we sure it' can't turn an option into `null` - that should be the point of it.~~

Edit: I think I understand now that the problem is not the Option, the problem is the `BStringForFrontend`. It would be nice if there was a way to configure ts-rs more centrally, and configure it once.

Byron (approved):
Thanks a lot!

I added a refactor on top which simplifies the type we store for Gitlab project ids to match what it's storing. That way, everything gets a lot simpler and we don't just proliferate the crutch that is `BStringForFrontend`.

And seeing these ugly annotations, I was hoping that maybe there is some central place to teach `ts-rs` basic type mappings, on a project level. So from that point of you, these `cfg_attr` annotations do exactly what they are supposed to - make us think of ways to have less of them 😁.

In any case, good to go.

estib-vega:
Thanks @Byron!
Noted, I'll look into configuring it somewhere similar to the things we have for serde.

Byron:
Thanks @estib-vega, I hope this works out.

I also realise that `string_or_ignore` should probably at least log the ignored value, or something, so we don't have to track this down 10 years from now because there is this one user… then again, how would they ever have *valid* project ids that contain invalid UTF8? Never mind :D.

PR #12234, opened 2026-02-05
chore: Fix ts definitions check

slarse:
## 🧢 Changes
scripts/generate-ts-definitions-from-rust.sh script is failing on master because it updates Cargo.lock (a bit of a false positive but IMO not really worth investing time into fixing atm).

Hilariously, the script also has `git add .` in backticks in the echo at the end of the script, which actually executes `git add .`, staging the changes and makes the error message `<snip> This error will go away with ,<snip>` instead of `<snip>This error will go away with 'git add .'<snip>`.

### DIFF
--- a/scripts/generate-ts-definitions-from-rust.sh
+++ b/scripts/generate-ts-definitions-from-rust.sh
@@ -16,6 +16,6 @@ pnpm format
 
 if ! git diff --quiet; then
   git status
-  echo "There are changed TS bindings. This error will go away with `git add .`, and on CI when the changes are committed."
+  echo "There are changed TS bindings. This error will go away with 'git add .', and on CI when the changes are committed."
   exit 2
 fi
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1911,7 +1911,7 @@ dependencies = [
  "pin-project",
  "serde",
  "serde_json",
- "thiserror 2.0.17",
+ "thiserror 2.0.18",
  "tokio",
  "tracing",
  "typed-builder",


slarse:
@Byron  I'm just tagging you here as an FYI, I'll merge this as soon as CI passes as it's currently messing up master :)

scripts/generate-ts-definitions-from-rust.sh
```diff
@@ -16,6 +16,6 @@ pnpm format
 
 if ! git diff --quiet; then
   git status
-  echo "There are changed TS bindings. This error will go away with `git add .`, and on CI when the changes are committed."
```
Byron, line 19: Oh no! Bash, you got me again!

I mean, really, I wasn't aware of this "feature" 🤦‍♂️.

Thanks for tagging me for this gem.

slarse: When debugging the error, I was initially _very_ confused as to why I had staged changes in my repository after running the script :D

But to make you feel better, I just got sniped by bash myself in that I never knew that `&&` and `||` have the same precedence, meaning that they'll just do normal left-to-right associative execution. So, for example:

```bash
if [ 1 == 1 ] || [ 1 == 0 ] && [ 1 == 0];
    echo 'Yay!'
else
    echo 'Nay!'
fi
```

With any modern programming language, the precedence of `&&` is higher than `||` so this would be executed as `1 == 1 || (1 == 0 && 1 == 0)`, which ultimately evaluates to true. However, in bash (or any POSIX shell), `&&` and `||` having the same precedence means that this instead just binds left-to-right and evaluates as `((( 1 == 1 ) || 1 == 0 ) && 1 == 0)`, which is of course false.

Cost me an hour of absolute confusion, as I blindly trusted Opus that it was correctly written (I mean, it wouldn't get _that_ basic a thing wrong, RIGHT!!!).

Byron: Opus 4.6 is already *such* a disappointment 😁.
I just try to be funny here in the face of the sadness that often is incurred by bash.

slarse: Even more infuriatingly, Copilot actually reinforced the idea that the operator precedence was as I expected with this stellar review comment:

> The mixed use of || and && operators without explicit parentheses relies on bash operator precedence (where && binds tighter than ||). While the logic is technically correct and produces the intended behavior, consider adding explicit parentheses for clarity.

So even though its suggestion would have fixed the issue, the incorrect justification had the opposite effect: I ignored that line of code when trying to figure out the problem :D

PR #12236, opened 2026-02-05
Revert "chore(publish): Build and publish standalone but CLI for Linux"

slarse:
Reverts gitbutlerapp/gitbutler#12230

Just in case I need it.


slarse:
Not needed.

PR #12251, opened 2026-02-06
Try to avoid rust analyzer dependency issue

Caleb-T-Owens:
For some really strange reason, I was getting the error `Dependency opener is not an optional dependency` in my code editor on  line six of `gitbutler-tauri/src/lib.rs`. 

Making opener optional seems to help, but I really don’t know why.

### DIFF
--- a/crates/gitbutler-tauri/Cargo.toml
+++ b/crates/gitbutler-tauri/Cargo.toml
@@ -78,7 +78,7 @@ tracing-forest.workspace = true
 open.workspace = true
 ## This is only used on macOS when it's the stable build to open `com.gitbutler.app` folders.
 ## Ideally, we could only use the dependency then.
-opener = { version = "0.8.3", features = ["reveal"] }
+opener = { version = "0.8.3", features = ["reveal"], optional = true }
 url = "2.5.4"
 but-path.workspace = true
 uuid.workspace = true


Caleb-T-Owens:
I don't know enough about what is going wrong here to fix this issue

Byron:
That's interesting. Zed also uses the Rustanaliser and it doesn't show anything.

Screenshot 2026-02-06 at 15 18 05

As I am new to the LSP world, I also have no experience with related errors yet.

Caleb-T-Owens:
Screenshot 2026-02-06 at 15 28 59

This is what I'm seeing @Byron.

My rust analyser setup looks like:

```
	"rust-analyzer.check.command": "clippy",
	"rust-analyzer.check.allTargets": true,
	"rust-analyzer.check.features": "all",
	"rust-analyzer.check.extraArgs": ["--target-dir", ".rust-analyzer/target"],
```

in vscode.

It's been this way for a long time now and I'm not super inclined to change it

Caleb-T-Owens:
Other than, changing .rust-analyzer/target to be `target/rust-analyzer` - but the clippy checking stuff hasn't been a problem before which is a little strange to me

Byron:
I see. Indeed I have just standard settings for the RA, probably that's related.

PR #12300, opened 2026-02-09
fix: convert filesystem paths to file:// URLs for Linux file manager (#12272)

### LINKED ISSUE #12272: 'Show in File Manager' throws 'Invalid path format' error
### Version

0.19.0

### Operating System

Linux

### Distribution Method

deb (Linux)

### Describe the issue

```

Caused by:

```

_No response_

_No response_

```shell

```

themavik:
## Summary

Fixes #12272.

**Root cause:** `show_in_finder()` on Linux passes bare filesystem paths (e.g., `/home/me/Projects/Launcher/Launcher`) directly to `open_that()`. The `open_that()` function parses its input with `Url::parse()`, which requires a valid URL with a scheme. Bare filesystem paths lack a scheme and fail with the error: `Invalid path format: '...' → relative URL without a base`.

Note: This only affects Linux. On macOS and Windows, `show_in_finder()` uses `Command::new("open")` and `Command::new("explorer")` respectively, bypassing `open_that()` entirely.

**Fix:** Two minimal changes:
1. Convert filesystem paths to `file://` URLs using `Url::from_file_path()` in the Linux block of `show_in_finder()`, so `open_that()` receives a valid URL.
2. Add `"file"` to the allowed URL schemes in `open_that()` so `file://` URLs pass the scheme validation.

## Changes

- `crates/but-api/src/legacy/open.rs`:
  - `open_that()`: Added `"file"` to the allowed URL schemes whitelist.
  - `show_in_finder()` (Linux block): Refactored to convert the target directory path to a `file://` URL via `Url::from_file_path()` before passing to `open_that()`. This ensures `xdg-open` receives a properly formatted URL.

## Root Cause Analysis

The `open_that()` function was designed for opening URLs (http, https, vscode://, etc.) and performs URL parsing + scheme validation. When `show_in_finder()` on Linux was implemented, it reused `open_that()` for filesystem paths without converting them to URLs first. macOS and Windows code paths use direct `Command::new()` calls and don't go through `open_that()`, which is why this bug is Linux-specific.

## Risk Assessment

- **Low risk:** The `file` scheme is safe — it only opens the system file manager via `xdg-open`.
- **No behavior change** on macOS or Windows, since they don't use `open_that()` for file manager operations.
- **AppImage env cleaning** is preserved since `open_that()` still handles the command execution.

### DIFF
--- a/crates/but-api/src/legacy/open.rs
+++ b/crates/but-api/src/legacy/open.rs
@@ -12,6 +12,7 @@ pub(crate) fn open_that(path: &str) -> anyhow::Result<()> {
         "http",
         "https",
         "mailto",
+        "file",
         "vscode",
         "vscodium",
         "vscode-insiders",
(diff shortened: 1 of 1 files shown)


Caleb-T-Owens:
Thanks for the PR. I ended up tweaking it a little bit here: https://github.com/gitbutlerapp/gitbutler/pull/12352

PR #12304, opened 2026-02-09
Fix SKILL.md frontmatter to comply with Agent Skills spec

rkondratowicz:
## 🧢 Changes

- Updated SKILL.md frontmatter to move `version` and `author` fields under `metadata` section
- Modified `inject_version()` function to replace `metadata.version` instead of top-level `version` field
- Updated `extract_installed_version_from_content()` to parse version from the `metadata.version` field
- Added YAML parsing logic to detect and navigate the metadata section in frontmatter

## ☕️ Reasoning
The current `SKILL.md` frontmatter uses non-standard top-level fields (`version` and `author`) that don't comply with the [Agent Skills specification](https://agentskills.io/specification).  

This change makes the GitButler CLI skill compliant with the Agent Skills spec, which is important for proper integration with AI coding assistants that implement this specification (Claude Code, Cursor, GitHub Copilot, etc.).

Testing:

✅ All 42 unit tests passing
✅ All 2 integration tests passing
✅ Builds successfully with `cargo build -p but --release`

This is a non-breaking change - the skill functionality remains identical, only the **frontmatter** structure is updated.

### DIFF
--- a/crates/but/skill/SKILL.md
+++ b/crates/but/skill/SKILL.md
@@ -1,8 +1,9 @@
 ---
 name: but
-version: 0.0.0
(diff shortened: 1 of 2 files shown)


Byron:
Thanks for the suggestion!

@krlvi would know best if this change is desirable. Meanwhile, I have put this PR back into draft as it doesn't pass CI and Copilot would probably deserve attention, even if just by discarding its comments.

PR #12305, opened 2026-02-10
tests commit: use filenames instead of autogen IDs

jonathantanmy2:
Make clearer what files are being operated on by using filename CLI IDs instead of the two-character automatically generated ones.

@krlvi PTAL and see if this change is reasonable. I think there might be some misunderstanding about what a filename CLI ID represents - it either represents an unstaged file, a staged file to a specific stack, or a file in a specific commit (it cannot represent two or more at once). In the original test, `file_id` is originally an unstaged file and coincidentally became a staged file, which might have caused the confusion.

I'm going to automerge this to avoid blocking a subsequent PR, but wanted to call this out separately in case we need to revert this (or design CLI IDs so that we support this use case).


krlvi:
looks good

PR #12306, opened 2026-02-10
Stable IDs for unstaged files

jonathantanmy2:
Why I'm not automerging: I need code owner review.

As described in the "id: use filename-derived CLI IDs for unstaged files" commit message, this is limited to unstaged files. I plan to add support for the others in subsequent PRs.


krlvi (approved):
Thanks! Looks good, and im excited to use it

PR #12321, opened 2026-02-11
fix: count unique files instead of hunks in `but absorb` summary

themavik:
## Summary

Fixes #12298

`but absorb` displays the wrong file count when a file has multiple modified hunks. It counts the total number of hunk entries (`FileAbsorption` entries) instead of unique file paths.

**Before (bug):** 2 files with 6 total hunks shows `Found 6 changed files to absorb:`
**After (fix):** Same scenario correctly shows `Found 2 changed files to absorb:`

## Root Cause

In `display_absorption_plan()`, the `total_files` count was computed as:
```rust
let total_files: usize = commit_absorptions.iter().map(|c| c.files.len()).sum();
```

Each `CommitAbsorption.files` entry is a `FileAbsorption` representing a **single hunk**, not a unique file. A file with 4 hunks produces 4 `FileAbsorption` entries, all with the same `path`.

## Fix

Collect unique file paths into a `HashSet` before counting:
```rust
let total_files: usize = commit_absorptions
    .iter()
    .flat_map(|c| c.files.iter().map(|f| f.path.as_str()))
    .collect::<std::collections::HashSet<_>>()
    .len();
```

This also fixes the `total_files` field in the JSON output, which uses the same variable.

## Changes

- `crates/but/src/command/legacy/absorb.rs` — Fix `total_files` to count unique file paths via `HashSet`
- `crates/but/tests/but/command/absorb.rs` — Update 5 snapshot tests: all test scenarios use a single file (`a.txt`) with multiple hunks, so the expected count changes from `2` to `1`

## Test Plan

- [x] `cargo check -p but` — compiles clean
- [x] `cargo clippy -p but` — no warnings
- [x] `cargo fmt -p but -- --check` — properly formatted
- [ ] CI snapshot tests should pass with the updated expectations

### DIFF
--- a/crates/but/src/command/legacy/absorb.rs
+++ b/crates/but/src/command/legacy/absorb.rs
@@ -187,8 +187,12 @@ fn display_absorption_plan(
     new: bool,
     write_json: bool,
 ) -> anyhow::Result<Option<JsonAbsorbOutput>> {
-    // Count total files
-    let total_files: usize = commit_absorptions.iter().map(|c| c.files.len()).sum();
+    // Count unique files (not hunks — a file with multiple hunks is still one file)
+    let total_files: usize = commit_absorptions
+        .iter()
+        .flat_map(|c| c.files.iter().map(|f| f.path.as_str()))
+        .collect::<std::collections::HashSet<_>>()
+        .len();
 
     // Handle empty case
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot, and sorry, I saw this too late and superseded it with #12362 .

PR #12322, opened 2026-02-11
fix: preserve `#` in branch names during frontend slugification

themavik:
## Summary

Fixes #9076

Branch names containing `#` (e.g. `feat/#1234-ticket-name`) have the `#` silently stripped when creating or renaming branches. This is because the frontend's `slugify` function uses a character allowlist that doesn't include `#`, even though `#` is a valid character in git ref names.

The backend's `normalize_short_name` correctly preserves `#` (confirmed by existing Rust tests: `foo#branch` → `foo#branch`). The issue is purely in the frontend `BranchNameTextbox` component.

## Root Cause

`packages/ui/src/lib/utils/string.ts`:
```typescript
// This regex strips # because it's not in the allowed set
.replace(/[^A-Za-z0-9._/ -]/g, '')
```

This `slugify` is also used for non-branch contexts (org slugs, markdown headings) where `#` should be stripped, so modifying it directly would break those contexts.

## Fix

1. **New function `slugifyBranchName`** in `packages/ui/src/lib/utils/string.ts` — a branch-name-aware variant that additionally allows `#` and `@` (both valid in git refs)
2. **`BranchNameTextbox.svelte`** now imports `slugifyBranchName` instead of the generic `slugify`
3. The generic `slugify` remains unchanged for non-branch contexts (`OrganizationEditModal`, `Heading.svelte`)

## Changes

- `packages/ui/src/lib/utils/string.ts` — Add `slugifyBranchName` function
- `packages/ui/src/lib/utils/string.test.ts` — Add 7 tests for the new function
- `apps/desktop/src/components/BranchNameTextbox.svelte` — Use `slugifyBranchName`

## Test Plan

- [x] New vitest tests cover `#`, `@`, slashes, periods, whitespace, git-invalid chars
- [x] Existing `slugify` tests unmodified and still pass
- [x] Backend already preserves `#` (Rust test: `normalize_short_name("foo#branch") == "foo#branch"`)
- [ ] Manual: type `feat/#1234-ticket` in branch name input → should show the `#` preserved

### DIFF
--- a/apps/desktop/src/components/BranchNameTextbox.svelte
+++ b/apps/desktop/src/components/BranchNameTextbox.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 	import { Textbox } from '@gitbutler/ui';
-	import { slugify } from '@gitbutler/ui/utils/string';
+	import { slugifyBranchName } from '@gitbutler/ui/utils/string';
 
 	type Props = {
 		value?: string;
@@ -19,7 +19,7 @@
 
 	let textbox = $state<ReturnType<typeof Textbox>>();
 
-	const slugifiedName = $derived(value && slugify(value));
+	const slugifiedName = $derived(value && slugifyBranchName(value));
 	const namesDiverge = $derived(!!value && slugifiedName !== value);
 	const computedHelperText = $derived(
 		namesDiverge ? `Will be created as '${slugifiedName}'` : helperText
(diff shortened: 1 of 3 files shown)


Byron (approved):
Thanks a lot for fixing this!

The problem is indeed resolved.

https://github.com/user-attachments/assets/407069cd-5a04-4a7d-bc2e-6073e00fc11f

What I wonder though is if the frontend shouldn't just use the backend function instead of implementing its own?

Could @estib-vega decide if this can be merged, and if it makes sense to have a follow-up that relies completely on the backend instead?

Byron:
Maybe @mtsgrd could take a look then?

Byron:
Superseded by #12488 .

PR #12323, opened 2026-02-11
fix(cli): allow `but resolve` commands to work on gitbutler/edit branch

### LINKED ISSUE #12283: but resolve finish fails with 'No default target branch set' while on gitbutler/edit
## Summary

`but resolve finish` (and `but resolve status`) fail with "Setup required: No default target branch set" while on the `gitbutler/edit` branch, even though the default target is correctly configured in `virtual_branches.toml`.

The `gitbutler/edit` branch is the one that `but resolve <commit>` itself checks out — so the user is put into a state where the resolve workflow cannot be completed through normal means.

## Version

- `but` 0.19.0 (latest as of 2026-02-08)
- macOS (Darwin 25.3.0)

## Steps to Reproduce

1. Have a GitButler workspace with a stacked branch that has a conflicted commit (e.g. after `but absorb` triggers a rebase conflict)
2. Enter conflict resolution: `but resolve <conflicted-commit-sha>`
   - GitButler checks out `gitbutler/edit` as expected
   - `edit_mode_metadata.toml` is written with the commit and stack IDs
3. Resolve the conflict in the working tree
4. Stage the resolved file: `git add <file>`
5. Run `but resolve finish`

**Expected:** Conflict resolution completes, workspace is updated with the resolved commit, returns to `gitbutler/workspace`

**Actual:**
```
Error: Setup required: No default target branch set.
```

## Key Detail: the `gitbutler/edit` branch context

The error occurs specifically while on the `gitbutler/edit` branch — the branch that `but resolve` itself creates and checks out. This means the resolve workflow is broken end-to-end: `but resolve` puts you on `gitbutler/edit`, but `but resolve finish` cannot operate from that branch.

The `virtual_branches.toml` has the target correctly configured:

```toml
[default_target]
branchName = "main"
remoteName = "origin"
remoteUrl = "git@github.com:example/repo.git"
sha = "e95b8d100bc1f39cf25a6fc9c3fcfcb3c1850c37"
pushRemoteName = "origin"
```

Other commands like `but setup` read this target fine when run from `gitbutler/workspace` or `main`. The resolve subcommand's target lookup appears to use a different codepath that fails when the current branch is `gitbutler/edit`.

`but resolve status` also fails with the identical error, suggesting the issue is in the resolve subcommand's shared target resolution logic rather than being specific to `finish`.

## Impact

This effectively makes conflict resolution impossible through the normal `but resolve` workflow. The user is stranded on `gitbutler/edit` with no way to complete the resolution via `but` commands.

## Workaround

We were able to work around this by:
1. Manually creating a replacement git commit object (same tree from the "ours" side of the conflict, same parent/headers, but without the `gitbutler-conflicted 1` header)
2. Using `git replace <conflicted-sha> <clean-sha>` so GitButler transparently sees the clean commit
3. Restoring the original `virtual_branches.toml` with the original commit SHAs
4. Running `but teardown` + `but setup` to rebuild the workspace

This required significant git plumbing knowledge and is not a realistic path for most users.

## Environment

```
$ but --version
but 0.19.0

$ but update check
✓ You're running the latest version (0.19.0)

$ uname -a
Darwin 25.3.0
```

themavik:
## Summary

- `but resolve finish`, `but resolve status`, and `but resolve cancel` all fail with "Setup required: No default target branch set" when running on the `gitbutler/edit` branch
- Root cause: `check_project_setup()` builds a workspace graph from HEAD, but on `gitbutler/edit` the graph doesn't expose the target ref from the workspace metadata — causing the target ref check to bail
- Fix: move the HEAD branch check before the target/remote checks and early-return `Ok(true)` when on `gitbutler/edit`, since the project was already fully validated when entering edit mode via `but resolve <commit>`

## Problem

When a user enters conflict resolution mode with `but resolve <commit>`, GitButler checks out the `gitbutler/edit` branch. All subsequent `but resolve` subcommands (`finish`, `status`, `cancel`) then fail because `init_ctx()` calls `check_project_setup()` which:

1. Builds a workspace projection from `gitbutler/edit` HEAD
2. The edit branch has no workspace metadata, so `target_ref` is `None`
3. Bails with "No default target branch set"

This makes the entire resolve workflow broken end-to-end: `but resolve` puts you on `gitbutler/edit`, but no resolve subcommand can run from that branch.

## Fix

In `check_project_setup()`:
1. Check the HEAD branch name **first** (moved before target/remote checks)
2. If HEAD is `gitbutler/edit`, return `Ok(true)` immediately — the project was already set up when entering edit mode, and the target configuration exists in `virtual_branches.toml` even though the workspace graph from the edit branch doesn't expose it
3. Otherwise, proceed with the existing target ref and remote checks

## Test plan

- [x] \`cargo check --package but\` passes
- [x] \`cargo clippy --package but\` passes with no warnings
- Verify manually:
  1. Set up a workspace with a conflicted commit
  2. Run \`but resolve <commit-id>\` to enter edit mode
  3. Run \`but resolve status\` — should succeed (previously failed)
  4. Run \`but resolve finish\` — should complete resolution (previously failed)
  5. Run \`but resolve cancel\` on a fresh edit session — should cancel (previously failed)

Fixes #12283
Fixes #12312

### DIFF
--- a/crates/but/src/command/legacy/setup.rs
+++ b/crates/but/src/command/legacy/setup.rs
@@ -396,9 +396,28 @@ More info: https://docs.gitbutler.com/workspace-branch
 /// - if the project is registered in GitButler
 /// - if there is a remote
 /// - if there is a default target branch set
-/// - if we're on gitbutler/workspace
+/// - if we're on a gitbutler/* branch
 pub fn check_project_setup(ctx: &Context, perm: &RepoShared) -> anyhow::Result<bool> {
     let (repo, ws, _) = ctx.workspace_and_db_with_perm(perm)?;
+
+    // check if we're on a gitbutler/* branch
+    let head = repo.head()?;
+    let head_name = head
+        .referent_name()
+        .map(|n| n.shorten().to_string())
+        .unwrap_or_default();
+    if !head_name.starts_with("gitbutler/") {
+        anyhow::bail!("Not currently on a gitbutler/* branch.");
+    }
+
+    // When on gitbutler/edit, the project was already set up when entering edit mode.
+    // The workspace graph built from gitbutler/edit doesn't expose the target ref or
+    // remote configuration, but both are still configured in virtual_branches.toml
+    // and will be accessible when returning to gitbutler/workspace.
+    if head_name == "gitbutler/edit" {
+        return Ok(true);
+    }
+
     // TODO(legacy): it's fine to have no target.
     if ws.target_ref.is_none() {
         anyhow::bail!("No default target branch set.");
@@ -410,16 +429,6 @@ pub fn check_project_setup(ctx: &Context, perm: &RepoShared) -> anyhow::Result<b
             "Neither found push remote found in workspace nor unambiguously in the Git repository configuration."
         )
     };
-
-    // check if we're on gitbutler/workspace
-    let head = repo.head()?;
-    let head_name = head
-        .referent_name()
-        .map(|n| n.shorten().to_string())
-        .unwrap_or_default();
-    if !head_name.starts_with("gitbutler/") {
-        anyhow::bail!("Not currently on a gitbutler/* branch.");
-    }
 
     Ok(true)
 }


krlvi:
~~Thanks a lot of this PR! This is a legit fix, and the original code was written in a sub-optimal way, because there is dedicated functionality for this (no need for manual parsing)~~

~~Fixing it here~~
https://github.com/gitbutlerapp/gitbutler/pull/12331

krlvi:
thanks!

krlvi (approved):

PR #12325, opened 2026-02-11
fix(ui): prevent duplicate revert operations when restoring a snapshot

themavik:
## Summary

- Reverting a snapshot creates duplicate RestoreFromSnapshot entries in the history because `restoreSnapshot()` was not awaited before `location.reload()` fired
- Fix: await the restore mutation before reloading, and disable all Revert buttons during the operation to prevent double-clicks
- Added loading spinner feedback on the Revert button while restoring

## Problem

When clicking "Revert" on a snapshot:

1. `historyService.restoreSnapshot(projectId, entry.id)` is called but **not awaited**
2. `location.reload()` fires immediately, before the async IPC call completes
3. This creates a race condition where:
   - The reload may interrupt the in-flight restore
   - The button could be clicked again before the first restore finishes
   - Both scenarios can produce duplicate RestoreFromSnapshot entries

## Fix

1. **Await the mutation**: The `onRestoreClick` handler now awaits `restoreSnapshot()` before reloading
2. **Prevent double-clicks**: A `restoring` state flag disables all Revert buttons while a restore is in progress
3. **Loading feedback**: The active Revert button shows a spinner via the existing `loading` prop

## Test plan

- [ ] Open a project with snapshot history
- [ ] Click "Revert" on any snapshot entry
- [ ] Verify only ONE RestoreFromSnapshot entry appears in history after the page reloads
- [ ] Verify the Revert button shows a spinner and all Revert buttons are disabled during the operation
- [ ] Verify the page still reloads correctly after restore completes

Fixes #12282

### DIFF
--- a/apps/desktop/src/components/SnapshotCard.svelte
+++ b/apps/desktop/src/components/SnapshotCard.svelte
@@ -14,12 +14,20 @@
 	interface Props {
 		entry: Snapshot;
 		isWithinRestore?: boolean;
+		restoring?: boolean;
 		onRestoreClick: () => void;
 		onDiffClick: (filePath: string) => void;
 		projectId: string;
 	}
 
-	const { projectId, entry, isWithinRestore = true, onRestoreClick, onDiffClick }: Props = $props();
+	const {
+		projectId,
+		entry,
+		isWithinRestore = true,
+		restoring = false,
+		onRestoreClick,
+		onDiffClick,
+	}: Props = $props();
 
 	function getShortSha(sha: string | undefined) {
 		if (!sha) return "";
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot for attempting the fix!

I can confirm that it works!

https://github.com/user-attachments/assets/15c987a6-f034-4637-97f8-44b94edd6085

Could @estib-vega bring this PR to completion? I think it's quite a high-value fix.
Thanks

Byron:
Maybe @mtsgrd could take [a look](https://github.com/gitbutlerapp/gitbutler/pull/12325#issuecomment-3907122725) then?

estib-vega (approved):
Thanks for the contribution!

### More context
We need to keep an eye on having too many state variables that mark the state of an API call.

I'll create a follow-up to improve that a bit. Otherwise, this PR looks good.

estib-vega:
The changes need to be signed, and re-formatted.

PR #12326, opened 2026-02-11
fix(linux): 'Show in File Manager' fails with 'Invalid path format' error

### LINKED ISSUE #12272: 'Show in File Manager' throws 'Invalid path format' error

themavik:
## Summary

- On Linux, right-clicking a file and selecting "Show in File Manager" fails with `Invalid path format` / `relative URL without a base` error
- Root cause: `show_in_finder` called `open_that()` with a plain filesystem path, but `open_that()` validates input as a URL via `Url::parse()` which rejects paths lacking a URL scheme
- Fix: refactored to separate URL validation from system-handler invocation, so Linux file manager opening bypasses URL validation while keeping AppImage env var cleaning

## Problem

The `open_that()` function was designed for URLs (http/https/vscode/etc.) and uses `Url::parse()` for validation. When `show_in_finder()` on Linux passed a filesystem path like `/home/user/Projects/Launcher/Launcher` to it:

1. `Url::parse("/home/user/Projects/...")` fails — plain paths aren't valid URLs
2. Error: `Invalid path format: '/home/user/Projects/Launcher/Launcher'` caused by `relative URL without a base`

macOS and Windows were unaffected because they use direct `Command::new("open")` / `Command::new("explorer")` calls that accept filesystem paths natively.

## Fix

Extracted the common AppImage-safe process spawning logic into `open_with_cleaned_env()`:

- `open_that()` — validates URL scheme, then delegates to `open_with_cleaned_env()` (for `open_url` command)
- `open_with_cleaned_env()` — opens any target (URL or path) via the `open` crate with cleaned environment variables
- Linux `show_in_finder` now calls `open_with_cleaned_env()` directly, bypassing URL validation

## Test plan

- [ ] On Linux: right-click a file in changed files list → "Show in File Manager" should open the parent directory
- [ ] On Linux: verify URLs still open correctly (e.g. clicking PR links, opening in editor)
- [ ] Verify no regressions on macOS/Windows file manager opening

Fixes #12272

### DIFF
--- a/crates/but-api/src/legacy/open.rs
+++ b/crates/but-api/src/legacy/open.rs
@@ -6,8 +6,49 @@ use but_api_macros::but_api;
 use tracing::instrument;
 use url::Url;
 
-pub(crate) fn open_that(path: &str) -> anyhow::Result<()> {
-    let target_url = Url::parse(path).with_context(|| format!("Invalid path format: '{path}'"))?;
+/// Environment variable names that need cleaning when spawning subprocesses
+/// on Linux to avoid AppImage path contamination.
+const APPIMAGE_ENV_VARS: &[&str] = &[
+    "APPDIR",
+    "GDK_PIXBUF_MODULE_FILE",
+    "GIO_EXTRA_MODULES",
+    "GIO_EXTRA_MODULES",
+    "GSETTINGS_SCHEMA_DIR",
+    "GST_PLUGIN_SYSTEM_PATH",
+    "GST_PLUGIN_SYSTEM_PATH_1_0",
+    "GTK_DATA_PREFIX",
+    "GTK_EXE_PREFIX",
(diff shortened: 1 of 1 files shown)


Byron (changes requested):
Thanks a lot for giving it a shot!

#12352 seems to be another attempt at a fix. And while refactors are appreciated, I think they should definitely get their own commit so it's easier to know what to review more closely.

In any case, let's check with @Caleb-T-Owens on how to proceed here, while putting a ping on @slarse radar as it looks like a test on Linux would be good.

Byron:
Closed as duplicate of [12352](https://github.com/gitbutlerapp/gitbutler/pull/12352).

PR #12331, opened 2026-02-11
feat(setup): Improve branch mode validation in project setup check

### LINKED ISSUE #12283: but resolve finish fails with 'No default target branch set' while on gitbutler/edit

krlvi:
Related to https://github.com/gitbutlerapp/gitbutler/pull/12323
Fixes #12283
Fixes #12312

### DIFF
--- a/crates/but/src/command/legacy/setup.rs
+++ b/crates/but/src/command/legacy/setup.rs
@@ -6,6 +6,7 @@ use but_core::{
 };
 use but_ctx::Context;
 use colored::Colorize;
+use gitbutler_operating_modes::OperatingMode;
 use serde::Serialize;
 
 use crate::utils::OutputChannel;
@@ -396,8 +397,17 @@ More info: https://docs.gitbutler.com/workspace-branch
 /// - if the project is registered in GitButler
 /// - if there is a remote
 /// - if there is a default target branch set
-/// - if we're on gitbutler/workspace
+/// - if we're not in outside-workspace mode
 pub fn check_project_setup(ctx: &Context, perm: &RepoShared) -> anyhow::Result<bool> {
+    let mode = gitbutler_operating_modes::operating_mode(ctx);
+    if matches!(mode, OperatingMode::OutsideWorkspace(_)) {
+        anyhow::bail!("Not currently on a gitbutler/* branch.");
+    }
+
+    if matches!(mode, OperatingMode::Edit(_)) {
+        return Ok(true);
+    }
+
     let (repo, ws, _) = ctx.workspace_and_db_with_perm(perm)?;
     // TODO(legacy): it's fine to have no target.
     if ws.target_ref.is_none() {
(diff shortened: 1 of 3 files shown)


krlvi:
Okay the function `gitbutler_operating_modes::operating_mode` is kinda flawed, cant use here gitbutler_operating_modes::operating_mode

PR #12335, opened 2026-02-11
Use gix in operating_mode and avoid panic

krlvi:
### Motivation
- Avoid panics from unwrapping the legacy `git2` repository.
- Unify HEAD inspection to `gix` so `operating_mode` can fail gracefully.

### Description
- `gitbutler_operating_modes::operating_mode` now uses `ctx.repo.get()` (gix) instead of `ctx.git2_repo.get().unwrap()`.
- All failure paths return `OperatingMode::OutsideWorkspace` (no panics). If the repo handle can't be opened/borrowed, we return `OutsideWorkspaceMetadata::default()` directly.
- HEAD/ref inspection uses `gix` (`head().referent_name()`) and compares against `OPEN_WORKSPACE_REFS` / `EDIT_BRANCH_REF` via byte-wise comparisons.
- Minor cleanup: factor repeated OutsideWorkspace construction into a local helper and avoid redundant metadata computation when repo access already failed.

Files:
- `crates/gitbutler-operating-modes/src/lib.rs`

Regression coverage added:
- Fallback when the repo cannot be opened (`.git` removed).
- Fallback when on `gitbutler/edit` without edit-mode metadata.

### DIFF
--- a/crates/gitbutler-operating-modes/src/lib.rs
+++ b/crates/gitbutler-operating-modes/src/lib.rs
@@ -79,18 +79,26 @@ pub enum OperatingMode {
 }
 
 pub fn operating_mode(ctx: &Context) -> OperatingMode {
-    let repo = ctx.git2_repo.get().unwrap();
+    let outside = || OperatingMode::OutsideWorkspace(outside_workspace_metadata(ctx).unwrap_or_default());
+
+    let Ok(repo) = ctx.repo.get() else {
+        // If we can't even open/borrow the repo handle, attempting to compute metadata would just
+        // try (and fail) to open it again.
+        return OperatingMode::OutsideWorkspace(OutsideWorkspaceMetadata::default());
+    };
     let Ok(head_ref) = repo.head() else {
-        return OperatingMode::OutsideWorkspace(outside_workspace_metadata(ctx).unwrap_or_default());
+        return outside();
+    };
+    let Some(head_ref_name) = head_ref.referent_name().map(|name| name.as_bstr()) else {
+        return outside();
     };
 
-    let Some(head_ref_name) = head_ref.name() else {
-        return OperatingMode::OutsideWorkspace(outside_workspace_metadata(ctx).unwrap_or_default());
-    };
-
-    if OPEN_WORKSPACE_REFS.contains(&head_ref_name) {
+    if OPEN_WORKSPACE_REFS
+        .iter()
+        .any(|workspace_ref| workspace_ref.as_bytes() == head_ref_name)
+    {
         OperatingMode::OpenWorkspace
-    } else if head_ref_name == EDIT_BRANCH_REF {
+    } else if EDIT_BRANCH_REF.as_bytes() == head_ref_name {
         let edit_mode_metadata = read_edit_mode_metadata(ctx);
 
         match edit_mode_metadata {
@@ -100,11 +108,11 @@ pub fn operating_mode(ctx: &Context) -> OperatingMode {
                     "Failed to open in edit mode, falling back to outside workspace {}",
                     error
                 );
-                OperatingMode::OutsideWorkspace(outside_workspace_metadata(ctx).unwrap_or_default())
+                outside()
             }
         }
     } else {
-        OperatingMode::OutsideWorkspace(outside_workspace_metadata(ctx).unwrap_or_default())
+        outside()
     }
 }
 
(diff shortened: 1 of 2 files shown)


Caleb-T-Owens:
@krlvi I just want to make sure I understand the problem we're solving here.

It looks like the function may be getting called in cases where there isn’t a git repository. If that’s the case, I’m not sure returning `OperatingMode::OutsideWorkspace` makes the most sense.

Also, with hindsight, unwrapping here seems like a bad idea, so we should probably refactor it to be a `Result` type.

Would it make sense to update the return type to `Result<Option<OperatingMode>>`, where `Ok(None)` represents the absence of a repository? Or would it be cleaner to introduce a `NoRepository` operating mode instead?

WDYT?

krlvi:
> @krlvi I just want to make sure I understand the problem we're solving here.
> 
> It looks like the function may be getting called in cases where there isn’t a git repository. If that’s the case, I’m not sure returning `OperatingMode::OutsideWorkspace` makes the most sense.
> 
> Also, with hindsight, unwrapping here seems like a bad idea, so we should probably refactor it to be a `Result` type.
> 
> Would it make sense to update the return type to `Result<Option<OperatingMode>>`, where `Ok(None)` represents the absence of a repository? Or would it be cleaner to introduce a `NoRepository` operating mode instead?
> 
> WDYT?

My objectives were:
- remove the panic
- remove the usage of git2

Happy to change the api, but im wondering if we want to go from `OperatingMode` to `Result<Option<OperatingMode>>` 🤷‍♂️
Maybe just a `Option<OperatingMode>` to represent not having a repo?

But I am also thinking, since "operating mode" is generic enough, maybe there could be a variant for "not a git repo" ?

Caleb-T-Owens:
> > @krlvi I just want to make sure I understand the problem we're solving here.
> > It looks like the function may be getting called in cases where there isn’t a git repository. If that’s the case, I’m not sure returning `OperatingMode::OutsideWorkspace` makes the most sense.
> > Also, with hindsight, unwrapping here seems like a bad idea, so we should probably refactor it to be a `Result` type.
> > Would it make sense to update the return type to `Result<Option<OperatingMode>>`, where `Ok(None)` represents the absence of a repository? Or would it be cleaner to introduce a `NoRepository` operating mode instead?
> > WDYT?
> 
> My objectives were:
> 
> * remove the panic
> * remove the usage of git2
> 
> Happy to change the api, but im wondering if we want to go from `OperatingMode` to `Result<Option<OperatingMode>>` 🤷‍♂️ Maybe just a `Option<OperatingMode>` to represent not having a repo?
> 
> But I am also thinking, since "operating mode" is generic enough, maybe there could be a variant for "not a git repo" ?

I think it would be helpful to refactor this to return a result while we're looking at it. I'm happy to take this on if you'd like.

After giving it some thought, I like the idea of using an `OperatingMode` variant to represent a missing repo.

krlvi:
> > > @krlvi I just want to make sure I understand the problem we're solving here.
> > > It looks like the function may be getting called in cases where there isn’t a git repository. If that’s the case, I’m not sure returning `OperatingMode::OutsideWorkspace` makes the most sense.
> > > Also, with hindsight, unwrapping here seems like a bad idea, so we should probably refactor it to be a `Result` type.
> > > Would it make sense to update the return type to `Result<Option<OperatingMode>>`, where `Ok(None)` represents the absence of a repository? Or would it be cleaner to introduce a `NoRepository` operating mode instead?
> > > WDYT?
> > 
> > 
> > My objectives were:
> > 
> > * remove the panic
> > * remove the usage of git2
> > 
> > Happy to change the api, but im wondering if we want to go from `OperatingMode` to `Result<Option<OperatingMode>>` 🤷‍♂️ Maybe just a `Option<OperatingMode>` to represent not having a repo?
> > But I am also thinking, since "operating mode" is generic enough, maybe there could be a variant for "not a git repo" ?
> 
> I think it would be helpful to refactor this to return a result while we're looking at it. I'm happy to take this on if you'd like.
> 
> After giving it some thought, I like the idea of using an `OperatingMode` variant to represent a missing repo.

Okay feel free to take over this thing but only if you have the bandwidth for it (i.e. it's not interfering with the rebase engine stuff). If you wanna take it, just close this PR

krlvi:
ping @Caleb-T-Owens how should we do here?

PR #12337, opened 2026-02-12
feat(but): add Gemini CLI support for skill installation

vikks:
### Add Gemini CLI support for skill installation

```sh
# supports gemini-cli now
but skills install
```

**Changes**
   - Added Gemini CLI to the supported platforms for but skill install.
   - Set default installation path to .gemini/skills/gitbutler.
   - Updated CLI help documentation and AI integration sections to include Gemini CLI.

**Reasoning**
Adds support for users of the Gemini CLI to easily install and manage GitButler skills, bringing it on parity with other cli agents.


Byron:
Thanks a lot!

Could you take a look at the conflict and resolve it?
Meantime, I think @krlvi would be interested in this generally.

krlvi:
this is already supported in the latest version

PR #12344, opened 2026-02-12
fix(graph): solve commit graph sort transitivity issue 

foxfirecodes:
**DISCLAIMER:** this fix was written by Claude Code. it looks *reasonable* to me but i am severely lacking in the context of how the commit graph works to know if this solution is right. just sharing this to maybe save you some time, take with a grain of salt

## 🧢 Changes

this PR solves a sort transitivity issue when constructing the commit graph. it also adds some unit tests to avoid this issue cropping up in the future

## ☕️ Reasoning

see https://github.com/gitbutlerapp/gitbutler/issues/12343 for more context, but tl;dr after a force push my commit graph ended up in a state where Git Butler panicked trying to parse. this boils down to a violation of the rules of transitivity when it comes to sorting. i will share the explanation that Claude Code gave me as it does a better job explaining than me lol:

```
  Root Cause: GenThenTime::cmp() violates transitivity

  The Ord impl at crates/but-graph/src/init/walk.rs:555-563:

  fn cmp(&self, other: &Self) -> Ordering {
      let time_cmp = self.committer_time.cmp(&other.committer_time).reverse();
      match (self.generation, other.generation) {
          (Some(a), Some(b)) => a.cmp(&b).reverse().then(time_cmp),
          _ => time_cmp,  // BUG: ignores generation when only one side has it
      }
  }

  Concrete transitivity violation:
  - A: gen=Some(3), time=200
  - B: gen=None, time=150
  - C: gen=Some(5), time=100
  ┌────────────┬───────────────────────┬──────────────────────────────────┐
  │ Comparison │         Path          │              Result              │
  ├────────────┼───────────────────────┼──────────────────────────────────┤
  │ A vs B     │ mixed gen → time only │ A < B (200 > 150, younger first) │
  ├────────────┼───────────────────────┼──────────────────────────────────┤
  │ B vs C     │ mixed gen → time only │ B < C (150 > 100, younger first) │
  ├────────────┼───────────────────────┼──────────────────────────────────┤
  │ A vs C     │ both Some → gen first │ A > C (gen 5 > 3, younger first) │
  └────────────┴───────────────────────┴──────────────────────────────────┘
  A < B < C but A > C — transitivity broken. Newer Rust sort implementations
  detect this and panic.

  The issue is that when both have generations, generation is primary. But when
  one is None, it falls back to time-only ordering, and these two orderings can
  disagree.

  Fix

  The fix is to give None a consistent sentinel value so generation ordering is
  always total. Using u32::MAX (matching git's GENERATION_NUMBER_INFINITY
  convention) treats unknown-generation commits as "assume youngest, process
  first" — conservative and safe since the sort is an optimization, not a
  correctness requirement.
```

## 🎫 Affected issues

Fixes: https://github.com/gitbutlerapp/gitbutler/issues/12343

### DIFF
--- a/crates/but-graph/Cargo.toml
+++ b/crates/but-graph/Cargo.toml
@@ -9,7 +9,6 @@ rust-version.workspace = true
 
 [lib]
 doctest = false
-test = false
 
 [dependencies]
 but-core.workspace = true
--- a/crates/but-graph/src/init/walk.rs
+++ b/crates/but-graph/src/init/walk.rs
@@ -1088,3 +1088,103 @@ impl crate::RefInfo {
         Self { ref_name, worktree }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn gtt(generation: Option<u32>, committer_time: u64) -> GenThenTime {
+        GenThenTime {
+            generation,
+            committer_time,
+        }
+    }
+
+    #[test]
+    fn gen_then_time_total_ordering_is_transitive_with_mixed_generations() {
+        // This is the exact scenario that previously caused a panic:
+        //   "user-provided comparison function does not correctly implement a total order"
+        //
+        // With the old implementation (fall back to time-only when generations are mixed):
+        //   A < B (by time: 200 > 150)
+        //   B < C (by time: 150 > 100)
+        //   A > C (by generation: 5 > 3)  — transitivity violation!
+        let a = gtt(Some(3), 200);
+        let b = gtt(None, 150);
+        let c = gtt(Some(5), 100);
+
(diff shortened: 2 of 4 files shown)


Caleb-T-Owens:
This looks like a reasonable change to me. I've started CI on it.

Out of an abundance of caution I would also like to defer to @Byron before merging since he has the most knowledge of this part of the code.

Byron:
Thanks a lot for setting up the fix!

I will merge it shortly after a quick hands-on.

Byron (approved):
Thanks so much for the fix - I think it's good, too.

Nothing else breaks, the tests actually test for this property and the old version fails them.

Let's bring it in.

PR #12347, opened 2026-02-13
Add conflict resolution recipes to agent skill

jasmeet107:
## Context

I use the `but` CLI daily as my primary version control interface (via AI coding agents with the skill file). The 0.19.1 skill rewrite is a huge improvement — the non-negotiable rules and `--status-after` pattern especially.

One gap I've noticed: conflict resolution isn't covered in SKILL.md. The reference docs (`references/reference.md`, `references/concepts.md`, `references/examples.md`) have good coverage of the `but resolve` commands, but agents that primarily work from the top-level SKILL.md don't see any of that.

In practice, my agents consistently struggle when `but pull` introduces conflicts — they don't know the `but resolve` workflow and sometimes fall back to git commands or get stuck.

## What this adds

Two new task recipes in SKILL.md:

1. **Resolve merge conflicts after pull** — the full `but resolve` workflow including zdiff3 markers, cascading conflicts, and the `but gui` fallback
2. **Handle hunk lock rejections** — the `AssignmentRejection`/`HunkLock` pattern when multiple branches touch the same file, and the `but rub` fix

Both follow the existing recipe format and keep the same concise style as the other recipes.

## Open to discussion

This is based on my experience as a daily user — happy to adjust the approach, wording, or structure. If there's a better way to surface conflict resolution to agents (maybe the references are meant to be the canonical source?), I'm totally open to that. Just want to find a good solution for this gap!

### DIFF
--- a/crates/but/skill/SKILL.md
+++ b/crates/but/skill/SKILL.md
@@ -91,6 +91,34 @@ Do not run `git push`, even if `but push` reports nothing to push.
 5. This two-step sequence is the safe default for reorder requests.
 6. Never use `git rebase` for this.
 
+### Resolve merge conflicts after pull
+
+When `but pull` introduces conflicts, affected commits are marked CONFLICTED.
+
+1. `but status --json` — identify commits with conflicted status
+2. `but resolve <commit-id>` — enter resolution mode for a specific commit
+   (or `but resolve` with no args to interactively select the first conflicted commit)
+3. Edit files to remove conflict markers (GitButler uses zdiff3-style: `<<<<<<<` ours, `|||||||` ancestor, `=======` theirs, `>>>>>>>`)
+4. `but resolve status` — check remaining conflicted files
+5. `but resolve finish` — finalize and return to workspace mode
+   (or `but resolve` again to proceed to the next conflicted commit)
+6. To abort: `but resolve cancel`
+
+**Important:**
+- Resolving one commit may cascade new conflicts in commits above it (rebase chain). Check `but status --json` after each resolution.
+- During resolution mode, other `but` mutation commands are limited. Finish or cancel before resuming normal work.
+- Conflicts can be resolved in any order — you are not forced to resolve them sequentially.
+- If `but resolve finish` fails unexpectedly, `but gui` can be used as a fallback to complete resolution through the desktop client.
+
+### Handle hunk lock rejections
+
+When two branches modify the same file, GitButler locks hunks to the branch that first touched those line ranges. This is not a merge conflict — it is a branch ownership constraint.
+
+**Symptoms:** `but stage` returns an `AssignmentRejection` with reason `HunkLock`. Commits targeting the locked lines silently produce empty commits.
+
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot for the update!

Let me summon @krlvi to see what can be done here.

jasmeet107:
even after using this update a bit, i've still run into a couple issues with resolving. claude usually ends up figuring it out, but i asked it to give me a summary of some of the issues its run into and solutions it came up with in this area. in case thats helpful!

```

CLI IDs shift after every operation

  CLI IDs (g0, h0, etc.) reshuffle after any but operation — commits, absorbs, amends, branch creation, even background syncs. Running but status --json, grabbing an ID, doing an
  operation, then using a previously-fetched ID grabs a completely different file. This has caused wrong files getting committed/amended into the wrong branch multiple times. The
  workaround is to always re-run but status --json before every single operation.

  but undo has workspace-wide blast radius

  but undo reverts to a workspace-wide snapshot, not per-branch. Chaining multiple undos to fix a bad amend caused unrelated branches to all show as "unpushed" even though nothing
  changed in them. Workaround: instead of chaining undos, restore the original file content (git show <merge-base>:<path> > <path>) and amend the restoration into the commit to cancel
  out the unwanted diff.

  Stuck in resolve mode (gitbutler/edit branch)

  but resolve <commit> checks out a gitbutler/edit branch for conflict resolution. If you exit incorrectly, all but commands fail with "No default target branch set." Clearing
  edit_mode_metadata.toml alone isn't enough — you need to git checkout gitbutler/workspace first, then clear the metadata file. Order matters.

  Hunk locks block cross-branch staging

  When two virtual branches modify the same file, hunks get locked to the branch that first touched those line ranges. but stage returns AssignmentRejection with HunkLock, and commits
  silently create empty. Workaround: but rub <commit> <target-branch> to move the locking commit, or the nuclear option — restore file to merge-base, absorb the revert (releases the
  lock), then re-apply changes fresh.

  Squash-merged branches not detected by but pull

  but pull can't detect squash-merged PRs as integrated because the squash commit doesn't match the original branch commits. Branches that are fully merged show as {conflicted} after
  pull. Workaround: but unapply <branch-name> to remove the stack.

  "Anonymous segment" error blocks branch deletion

  but branch delete refuses to delete if it's the only named branch in a stack. --force doesn't help. Workaround: but unapply <branch-name> to remove the stack from workspace instead.
  
```

so maybe my change is not enough 😅 not sure, haha

krlvi:
Thanks for the PR @jasmeet107 
One thing that we recently discovered is that helps with these multi turn commands is if in their output they provide "next steps" type of hints. I will try adding that

jasmeet107:
> Thanks for the PR @jasmeet107 One thing that we recently discovered is that helps with these multi turn commands is if in their output they provide "next steps" type of hints. I will try adding that

i'm happy to experiment with this myself and push changes, or i can close this pr and let you handle it :) whatever works!

Byron:
*(I just put it back intro draft while the PR is still in flux)*

PR #12366, opened 2026-02-16
Global clippy setup for all crates

Byron:
This helps to get lints for low-effort-high-value lints that improve readability.

Motivated by wanting more readable format strings, and help AI to avoid the less readable variants.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -276,6 +276,12 @@ crossterm = "0.28"
 self_cell = "1.2.2"
 unicode-segmentation = "1.12.0"
 
+[workspace.lints.clippy]
+all = "deny"
+perf = "deny"
+correctness = "deny"
+uninlined_format_args = "deny"
+
 [profile.release]
 # There are no overrides here as we optimise for fast release builds,
 # which are used when building locally and when making Nightly builds.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -278,8 +278,8 @@ unicode-segmentation = "1.12.0"
 
 [workspace.lints.clippy]
 all = "deny"
-perf = "deny"
-correctness = "deny"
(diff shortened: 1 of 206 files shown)


Byron (commented):
I didn't look at all changes, but what's really changed is the Cargo.toml updates needed to get Clippy workspace configuration going, and the new lint auto-fix to encourage inline variables where possible.

Cargo.toml
```diff
@@ -276,6 +276,12 @@ crossterm = "0.28"
 self_cell = "1.2.2"
 unicode-segmentation = "1.12.0"
 
+[workspace.lints.clippy]
+all = "deny"
+perf = "deny"
+correctness = "deny"
+uninlined_format_args = "deny"
```
Caleb-T-Owens, line 283: Could you clarify what `all` followed by `perf` and `correctness` means? Does `all` already encompass those?

Caleb-T-Owens:
@Byron, while we're putting together this PR that formats the codebase in a fairly widespread way, could you please revert the line length to the default?

Byron:
Oh right, we didn't talk about it during the Summit at all.
I'd like to keep that separate from this PR though.

Regarding line length, I'd like to hand the decision on how to best proceed over to @krlvi. Something I want to avoid is to have reversal based on principle, which I support, just to later find it should be brought back based on merit. Also, I want to be done with it, so resolving this sooner rather than later would be my preference.

Cargo.toml
Byron: Thanks for the review!
I didn't question this as it was already present, but now that I did (A)I could verify that `all` includes `perf` and `correctness`.
Now there is another note to show how to see more lints if one wanted to.

Caleb-T-Owens (approved):

PR #12367, opened 2026-02-16
Stack Entry: Carry associated review information

estib-vega:
Stack entries also know the associtated forge review ID.

This makes it easier to check whether a branch has an associated PR already.

### DIFF
--- a/crates/but-workspace/src/legacy/stacks.rs
+++ b/crates/but-workspace/src/legacy/stacks.rs
@@ -52,6 +52,15 @@ fn id_from_name_v2_to_v3_opt(
     }))
 }
 
+/// Get the associated forge review information out of the metadata, if any.
+fn review_id_from_meta(
+    name: &gix::refs::FullNameRef,
+    meta: &VirtualBranchesTomlMetadata,
+) -> anyhow::Result<Option<usize>> {
+    let pull_request = meta.branch_opt(name)?.and_then(|ref_meta| ref_meta.review.pull_request);
+    Ok(pull_request)
+}
+
 /// Returns the list of branch information for the branches in a stack.
 pub fn stack_heads_info(stack: &Stack, repo: &gix::Repository) -> anyhow::Result<Vec<StackHeadInfo>> {
     let branches = stack
@@ -62,6 +71,7 @@ pub fn stack_heads_info(stack: &Stack, repo: &gix::Repository) -> anyhow::Result
             let tip = branch.head_oid(repo).ok()?;
             Some(StackHeadInfo {
                 name: branch.name().to_owned().into(),
+                review_id: branch.pr_number,
                 tip,
                 is_checked_out: false,
             })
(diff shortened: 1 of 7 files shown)


estib-vega:
Sorry for the reviewers, this was supposed to be a draft

crates/but-workspace/src/legacy/ui.rs
```diff
@@ -18,6 +18,8 @@ pub struct StackHeadInfo {
     #[serde(with = "but_serde::object_id")]
     #[cfg_attr(feature = "export-ts", ts(type = "string"))]
     pub tip: gix::ObjectId,
+    /// The associated forge review with this branch, e.g. GitHub PRs or GitLab MRs
+    pub review_id: Option<usize>,
     /// If `true`, then this head is checked directly so `HEAD` points to it, and this is only ever `true` for a single head.
     /// This is `false` if the worktree is checked out.
     pub is_checked_out: bool,
```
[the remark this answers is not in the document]
estib-vega: It's an optional field.

crates/but/src/command/legacy/forge/review.rs
```diff
@@ -169,19 +169,25 @@ async fn ensure_forge_authentication(ctx: &mut Context) -> Result<(), anyhow::Er
 fn get_branches_without_prs(
     review_map: &std::collections::HashMap<String, Vec<but_forge::ForgeReview>>,
     applied_stacks: &[but_workspace::legacy::ui::StackEntry],
-) -> Vec<String> {
+) -> anyhow::Result<Vec<String>> {
     let mut branches_without_prs = Vec::new();
```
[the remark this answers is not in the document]
estib-vega: doesn't make sense

crates/but-workspace/src/legacy/stacks.rs
```diff
@@ -52,6 +52,15 @@ fn id_from_name_v2_to_v3_opt(
     }))
 }
 
+/// Get the associated forge review information out of the metadata, if any.
+fn review_id_from_meta(
+    name: &gix::refs::FullNameRef,
+    meta: &VirtualBranchesTomlMetadata,
+) -> anyhow::Result<Option<usize>> {
+    let ref_meta = meta.branch(name)?;
```
Caleb-T-Owens: Should this use `meta.branch_opt`. just `.branch` will error if there isn't a corresponding `virtual_branches.toml` entry.

Since it seems like the only time we call this function is in the case where we haven't already found the `virtual_branches.toml` entry, it seems quite likely that this will cause unexpected errors.

estib-vega: Oooh didn't know that, thanks!

estib-vega: Fixed 👍

Caleb-T-Owens (approved):

PR #12369, opened 2026-02-16
chore: make changes to install.sh trigger vercel deploy

slarse:
## 🧢 Changes
Includes the `install.sh` script in the paths that will trigger a deploy on Vercel.

Also took the opportunity to clean up a bit, removing paths that don't exist in our project and consolidating to a single loop (no reason to check files and directories separately).

### DIFF
--- a/ignore-vercel.sh
+++ b/ignore-vercel.sh
@@ -8,10 +8,16 @@
 echo "🔍 Checking for changes in monitored directories..."
 
 # Directories to monitor for changes
-MONITORED_DIRS=(
+MONITORED_PATHS=(
+  # webapp source
   "apps/web"
   "packages/shared" 
   "packages/core"
+  # dependency changes that may affect webapp
+  "package.json"
+  "pnpm-lock.yaml"
+  # the install script is bundled in the webapp
+  "scripts/install.sh"
 )
 
 # Check if we're in a git repository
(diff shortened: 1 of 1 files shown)


ignore-vercel.sh
```diff
@@ -40,37 +46,19 @@ fi
...
-  "yarn.lock"
-  "pnpm-lock.yaml"
-)
-
-for file in "${ROOT_FILES[@]}"; do
-  if [ -f "$file" ]; then
-    if ! git diff --quiet "$PREVIOUS_COMMIT" HEAD -- "$file"; then
-      echo "✅ Changes detected in root: $file"
-      HAS_CHANGES=true
-    fi
```
slarse, line 73: Could not find a good reason to separate checking files from directories, it's the exact same command to check regardless.

Also removed `yarn.lock` and `package-lock.json` as they aren't applicable to our project.

ignore-vercel.sh
```diff
@@ -8,10 +8,16 @@
...
+MONITORED_PATHS=(
+  # webapp source
   "apps/web"
   "packages/shared" 
   "packages/core"
+  # dependency changes that may affect webapp
+  "package.json"
+  "pnpm-lock.yaml"
+  # the install script is bundled in the webapp
+  "scripts/install.sh"
```
slarse, line 20: This is the meaningful addition - as the install script is bundled at compile time we need to rebuild and redeploy every time it changes.

PR #12370, opened 2026-02-16
fix(stack-service): normalize commit stats shape

PavelLaptev:
Screenshot 2026-02-16 at 15 24 47

The statistics for commits were not visible because the backend keys were in snake_case format, while the UI expected camelCase.

### DIFF
--- a/apps/desktop/src/lib/stacks/stackService.svelte.ts
+++ b/apps/desktop/src/lib/stacks/stackService.svelte.ts
@@ -1175,10 +1175,11 @@ function injectEndpoints(api: ClientState['backendApi'], uiState: UiState) {
 				],
 				transformResponse(rsp: CommitDetails) {
 					const changes = changesAdapter.addMany(changesAdapter.getInitialState(), rsp.changes);
+					const stats = rsp.stats;
 					return {
 						changes: changes,
 						details: rsp.commit,
-						stats: rsp.stats,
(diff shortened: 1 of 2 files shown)


apps/desktop/src/lib/stacks/stackService.svelte.ts
```diff
@@ -1175,10 +1175,11 @@ function injectEndpoints(api: ClientState['backendApi'], uiState: UiState) {
 				],
 				transformResponse(rsp: CommitDetails) {
 					const changes = changesAdapter.addMany(changesAdapter.getInitialState(), rsp.changes);
+					const stats = rsp.stats as TreeStats;
```
Caleb-T-Owens: Is it required to say `as TreeStats` here?

PavelLaptev: nope

PavelLaptev: should I remove it?

Caleb-T-Owens: Yes please

PavelLaptev: fixed. Thanks :-)

Caleb-T-Owens (approved):

PR #12384, opened 2026-02-17
Track visible items in VirtualList and highlight corresponding file list entries

mtsgrd:
Separate the VirtualList's render range (includes buffer for smooth scrolling) from the actual visible range (what the user sees in the viewport). Expose an onVisibleChange callback that fires when the visible range updates, and thread it through the component hierarchy so the file list sidebar can highlight which files are currently in view.


mtsgrd:
@PavelLaptev here's a PoC for being able to highlight the visible range of a virtual list.

PavelLaptev:
@mtsgrd thanks! Should I jump in with my edits, or are you still tweaking the draft?

mtsgrd:
You can go ahead!

PR #12387, opened 2026-02-17
Fix Claude hooks failing when tools access files outside the repository

mtsgrd:
Context::discover() walks up from the file path to find a .git
directory, which fails for files outside the repo (e.g. ~/.cargo).
Instead, pass the Context from the caller — either from init_ctx
in the CLI or from ThreadSafeContext in the SDK.

Also deduplicate the SDK/CLI hook variants into single functions,
and rename them for clarity:
- lock_file_for_tool_call (pre-tool file locking)
- assign_hunks_post_tool_call (post-tool hunk assignment)
- handle_session_stop (session end / commit creation)

### DIFF
--- a/crates/but-claude/src/hooks/mod.rs
+++ b/crates/but-claude/src/hooks/mod.rs
@@ -1,6 +1,6 @@
 use std::{collections::HashMap, path::Path, str::FromStr};
 
-use anyhow::{Context as _, Result, anyhow};
+use anyhow::{Context as _, Result};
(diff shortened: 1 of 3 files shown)


Byron (commented):
That's a great catch! I'd hope we can replace all the `::discover()` calls there as well.
(*and sorry for being here so early 😅, just wanted to share excitement*).

mtsgrd:
Thanks @Byron - your input is always valuable :)

I got rid of the other discover calls as well. Let me know what you think?

Byron (commented):
Less code is more, and less `::discovery()` as well 🎉.

PR #12388, opened 2026-02-17
but-sdk

estib-vega:
Bring it all together.

Hook up everything and generate bindings for:
- Listing stacks
- Fetching stack details
- Listing projects

### More context
At the top of this stack we have the full pipeline:
- Generate bindings and functions signature types using the `but_api` macros.
- Generate complex parameter/return types from their JSON schema.
- Gather everything together in an NPM package.

### Next steps:

- [x] Document the setup and generation of the types and bindings.
- [ ] Setup CI checks, similarly to the ts-rs type generation and tests.
    - [x] Generate the bindings and types on CI. Check the validity of the types.
    - [ ] Compile node plugins for multiple platforms (macOS types are the only ones available at the moment).
- [x] Setup an Electron scaffold.
- [ ] Migrate away from ts-rs.

### Intentionally omitted

* E2E for Electron - let's wait and see

---
This is **part 4 of 5 in a stack** made with GitButler:
- <kbd>&nbsp;5&nbsp;</kbd> #12405 
- <kbd>&nbsp;4&nbsp;</kbd> #12388 👈 
- <kbd>&nbsp;3&nbsp;</kbd> #12398 
- <kbd>&nbsp;2&nbsp;</kbd> #12396 
- <kbd>&nbsp;1&nbsp;</kbd> #12395


Byron:
Why aren't we using the `#[but_api]` macro for this? With it, the `napi` bindings could be feature toggled, similar to what's happening with the tauri bindings. I'd really want to avoid the duplication.

estib-vega:
> Why aren't we using the `#[but_api]` macro for this? With it, the `napi` bindings could be feature toggled, similar to what's happening with the tauri bindings. I'd really want to avoid the duplication.

Short answer is that I am at the 'just get it to work' phase.
But if we can cram this into the `but_api` macro, I agree, that's the better approach 👍

Byron:
Great to hear you see this could work.
With this PoC carved out, it will become clearer what that macro would have to do, and from all I saw (which isn't everything) that is very doable. Let's hope it stays that way, having another parallel instantiation of the API surface would be a maintenance nightmare.

Byron (approved):
It's Ok and we can follow-up changes in future PRs.
I am pretty sure this isn't going to break anything, but also did nothing more than an armchair review alongside a very pleasant call.

### Some notes

* please make sure the pipeline runs on CI.
* high-level docs would be appreciated. Educate the team.
* clearly mark generated files, ideally by putting them into an obnoxiously named folder.
* try to reduce the amount of AI generated code, can it be more concise (if needed).

PR #12408, opened 2026-02-18
Saver UX around exiting edit mode.

### LINKED ISSUE #12310: Editmode removed a day of work
### Version

0.19.1

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

I think i was in resolve mode, trying to resolve a conflict. I logged off and the next day i forgot about it. Did a lot of work and then opened gitbutler and saw edit mode. I didnt know what it was and clicked it away. All my work was gone and i couldnt even see it in the history. 

1. Create a change
2. Make sure it is a conflict with the server
3. Go in resolve mode
4. Make extra changes and then go out of resolve mode

Im not a 100% sure that is what happened, but that is my best guess

_No response_

```shell

```

Caleb-T-Owens:
* Resolves https://github.com/gitbutlerapp/gitbutler/issues/12310

When exiting edit mode, the intended behaviour is that any changes made are dropped.

This, however, introduced a UX issue where users were not made aware this was the case, and it unfortunately resulted in one user losing work.

This commit introduces flows that inform users of the potential risk, suggest that saving and exiting may be the preferred action, and then provide a safety override.

I have not switched over to `safe_checkout` yet because I am not sure it is doing exactly what we need here. In general, I am keen to refactor this code to use the new commit engine and move away from the older usage of stacks. This code has not seen many updates over its lifetime.

---

Desktop UX:
If it's safe, abort is immediately performed, otherwise the following modal is shown:

Screenshot 2026-02-18 at 13 01 12

CLI UX:
If it's safe, abort is immediately performed, otherwise this message is shown:

```bash
X why > /Users/calebowens/Sherman/projects/gitbutler-client/target/debug/but resolve cancel
Failed to handle conflict resolution. There are changes that differ from the original commit you were editing. Canceling will drop those changes.

If you want to go through with this, please re-run with `--force`.

If you want to keep the changes you have made, consider finishing the resolution and then moving the changes with the rub command
```

### DIFF
--- a/crates/but/tests/but/command/resolve.rs
+++ b/crates/but/tests/but/command/resolve.rs
@@ -105,12 +105,45 @@ fn resolve_cancel_works_in_edit_mode() -> anyhow::Result<()> {
     let env = Sandbox::init_scenario_with_target_and_default_settings("one-stack")?;
     enter_edit_mode_with_conflicted_commit(&env)?;
 
-    let cancel_output = env.but("resolve cancel").output()?;
+    let cancel_output = env.but("resolve cancel --force").output()?;
     let cancel_stderr = String::from_utf8_lossy(&cancel_output.stderr);
     anyhow::ensure!(cancel_output.status.success(), "resolve cancel should succeed");
     anyhow::ensure!(
         !cancel_stderr.contains("Setup required:"),
         "resolve cancel should not fail setup checks"
+    );
+
+    assert_eq!(current_branch_name(&env)?, "gitbutler/workspace");
+    Ok(())
+}
+
+#[test]
+fn resolve_cancel_requires_force_when_changes_were_made() -> anyhow::Result<()> {
+    let env = Sandbox::init_scenario_with_target_and_default_settings("one-stack")?;
+    enter_edit_mode_with_conflicted_commit(&env)?;
+
+    env.file("test-file.txt", "resolved content with additional edits\n");
+
+    let cancel_output = env.but("resolve cancel").output()?;
+    let cancel_stderr = String::from_utf8_lossy(&cancel_output.stderr);
+    anyhow::ensure!(
+        !cancel_output.status.success(),
+        "resolve cancel should fail without force"
+    );
+    anyhow::ensure!(
+        cancel_stderr.contains("--force"),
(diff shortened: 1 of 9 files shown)


crates/but/tests/but/command/resolve.rs
```diff
@@ -116,3 +116,36 @@ fn resolve_cancel_works_in_edit_mode() -> anyhow::Result<()> {
...
+    let force_cancel_output = env.but("resolve cancel --force").output()?;
+    let force_cancel_stderr = String::from_utf8_lossy(&force_cancel_output.stderr);
+    anyhow::ensure!(
+        force_cancel_output.status.success(),
+        "resolve cancel --force should succeed"
+    );
+    anyhow::ensure!(
+        !force_cancel_stderr.contains("Setup required:"),
+        "resolve cancel --force should not fail setup checks"
+    );
```
Byron, line 147: I noticed that AI really doesn't like Snapbox - it did this to me too and it's trivial to use snapbox here.

Can you fix it and maybe let it write instructions while at it?

crates/but/tests/but/command/resolve.rs
```diff
@@ -116,3 +116,36 @@ fn resolve_cancel_works_in_edit_mode() -> anyhow::Result<()> {
...
+        "resolve cancel should fail without force"
+    );
+    anyhow::ensure!(
+        cancel_stderr.contains("--force"),
+        "resolve cancel without force should explain how to proceed; stderr was: {cancel_stderr}"
+    );
+
+    let force_cancel_output = env.but("resolve cancel --force").output()?;
+    let force_cancel_stderr = String::from_utf8_lossy(&force_cancel_output.stderr);
+    anyhow::ensure!(
```
Byron, line 140: `anyhow::ensure!` is new to me as tests should panic, for the line numbers.

Byron (approved):
Thanks a lot, so much better!

Despite the approval for convenience, I'd really love that `but` test to be improved before a merge.
Thank you

Caleb-T-Owens:
The tests were written using the existing pattern in that test file. I'll make a follow up PR that updates the testing tooling for this file.

Byron:
Nevermind then, and thanks for making me aware. I will take a look and see if Agents.md can be adjusted to nudge future agents.

PR #12414, opened 2026-02-18
feat: gate tokio-console instrumentation behind an environment variable

smitmartijn:
Hi,

Happy GitButler user since ~day one, first time contributor. 👋

I noticed that `~/Library/Logs/com.gitbutler.app/tokio-console` was growing very large (16968 lines per 5 seconds) and a few GBs per day. It was cleared each time GB was restarted, but it's still a lot on unnecessary disk I/O. It incrementally gets worse when you have more repos added.

After digging into the code, I found that tokio-console instrumentation was unconditionally enabled in all builds. I understand having it for troubleshooting and tracing issues, but it seems excessive to have it enabled for all users by default.

So..I added a new "Tokio runtime tracing" toggle under Experimental settings to control whether the `ConsoleLayer` is registered at startup. When off, no `ConsoleLayer` is registered and no tokio-console file is written. I figured you'd want people to be able to enable it, so I also added a setting for it. 

## 🧢 Changes

Adds a "Tokio runtime tracing" toggle to the Experimental settings page that controls whether the ConsoleLayer is registered at startup. Disabled by default, eliminating unnecessary disk I/O and gRPC overhead for all users. Shows a restart-required warning toast when toggled.

## NOTE:

Changed enabling the tokio-console logging to use an environment variable, as requested. `GITBUTLER_TOKIO_DEBUG=1` makes it install the logging now. Without that, it's not installed and not generating logs.

### DIFF
--- a/crates/gitbutler-tauri/src/logs.rs
+++ b/crates/gitbutler-tauri/src/logs.rs
@@ -5,7 +5,7 @@ use tracing::{Level, instrument, metadata::LevelFilter, subscriber::set_global_d
 use tracing_appender::rolling::{RollingFileAppender, Rotation};
 use tracing_subscriber::{Layer, filter::filter_fn, fmt::format::FmtSpan, layer::SubscriberExt};
 
-pub fn init(app_handle: &AppHandle, logs_dir: &Path, performance_logging: bool) {
+pub fn init(app_handle: &AppHandle, logs_dir: &Path, performance_logging: bool, tokio_console_enabled: bool) {
     fs::create_dir_all(logs_dir).expect("failed to create logs dir");
 
     let log_prefix = "GitButler";
@@ -40,9 +40,8 @@ pub fn init(app_handle: &AppHandle, logs_dir: &Path, performance_logging: bool) 
     let log_level = log_level_filter.into_level();
 
     let use_colors_in_logs = cfg!(not(feature = "windows"));
-    let subscriber = tracing_subscriber::registry()
-        .with(
-            // subscriber for https://github.com/tokio-rs/console
+    let console_layer = if tokio_console_enabled {
+        Some(
             console_subscriber::ConsoleLayer::builder()
                 .server_addr(get_server_addr(app_handle))
                 .retention(Duration::from_secs(3600)) // 1h
@@ -50,15 +49,18 @@ pub fn init(app_handle: &AppHandle, logs_dir: &Path, performance_logging: bool) 
                 .recording_path(logs_dir.join("tokio-console"))
                 .spawn(),
         )
-        .with(
-            // subscriber that writes spans to a file
-            tracing_subscriber::fmt::layer()
-                .event_format(format_for_humans.clone())
-                .with_ansi(false)
-                .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
-                .with_writer(file_writer)
-                .with_filter(filter_fn(move |meta| should_log(log_level, meta))),
-        );
+    } else {
+        None
+    };
+    let subscriber = tracing_subscriber::registry().with(console_layer).with(
+        // subscriber that writes spans to a file
+        tracing_subscriber::fmt::layer()
+            .event_format(format_for_humans.clone())
+            .with_ansi(false)
+            .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
+            .with_writer(file_writer)
+            .with_filter(filter_fn(move |meta| should_log(log_level, meta))),
+    );
     if performance_logging {
         set_global_default(
             subscriber.with(
--- a/crates/gitbutler-tauri/src/logs.rs
+++ b/crates/gitbutler-tauri/src/logs.rs
@@ -5,7 +5,7 @@ use tracing::{Level, instrument, metadata::LevelFilter, subscriber::set_global_d
 use tracing_appender::rolling::{RollingFileAppender, Rotation};
 use tracing_subscriber::{Layer, filter::filter_fn, fmt::format::FmtSpan, layer::SubscriberExt};
 
-pub fn init(app_handle: &AppHandle, logs_dir: &Path, performance_logging: bool, tokio_console_enabled: bool) {
+pub fn init(app_handle: &AppHandle, logs_dir: &Path, performance_logging: bool, enable_tokio_console_log: bool) {
     fs::create_dir_all(logs_dir).expect("failed to create logs dir");
 
     let log_prefix = "GitButler";
@@ -40,7 +40,7 @@ pub fn init(app_handle: &AppHandle, logs_dir: &Path, performance_logging: bool, 
     let log_level = log_level_filter.into_level();
 
     let use_colors_in_logs = cfg!(not(feature = "windows"));
-    let console_layer = if tokio_console_enabled {
+    let console_layer = if enable_tokio_console_log {
         Some(
             console_subscriber::ConsoleLayer::builder()
                 .server_addr(get_server_addr(app_handle))
--- a/crates/gitbutler-tauri/src/main.rs
+++ b/crates/gitbutler-tauri/src/main.rs
@@ -78,8 +78,7 @@ fn main() -> anyhow::Result<()> {
     std::fs::create_dir_all(&app_cache_dir).context("failed to create app cache dir")?;
     std::fs::create_dir_all(&app_log_dir).context("failed to create app log dir")?;
 
-    let tokio_console_enabled = std::env::var_os("GITBUTLER_TOKIO_DEBUG").is_some();
-
+    let tokio_debug = std::env::var_os("GITBUTLER_TOKIO_DEBUG").is_some();
     let app_settings_for_menu = app_settings.clone();
     runtime.block_on(async {
         tauri::async_runtime::set(tokio::runtime::Handle::current());
@@ -111,7 +110,7 @@ fn main() -> anyhow::Result<()> {
 
                 let app_handle = tauri_app.handle();
 
-                logs::init(app_handle, &app_log_dir, performance_logging, tokio_console_enabled);
+                logs::init(app_handle, &app_log_dir, performance_logging, tokio_debug);
 
                 but_action::cli::auto_fix_broken_but_cli_symlink();
                 inherit_interactive_login_shell_environment_if_not_launched_from_terminal();
--- a/crates/gitbutler-tauri/src/main.rs
+++ b/crates/gitbutler-tauri/src/main.rs
@@ -78,6 +78,8 @@ fn main() -> anyhow::Result<()> {
     std::fs::create_dir_all(&app_cache_dir).context("failed to create app cache dir")?;
     std::fs::create_dir_all(&app_log_dir).context("failed to create app log dir")?;
 
+    let tokio_console_enabled = std::env::var_os("GITBUTLER_TOKIO_DEBUG").is_some();
+
     let app_settings_for_menu = app_settings.clone();
     runtime.block_on(async {
         tauri::async_runtime::set(tokio::runtime::Handle::current());
@@ -109,7 +111,7 @@ fn main() -> anyhow::Result<()> {
 
                 let app_handle = tauri_app.handle();
 
-                logs::init(app_handle, &app_log_dir, performance_logging);
+                logs::init(app_handle, &app_log_dir, performance_logging, tokio_console_enabled);
 
                 but_action::cli::auto_fix_broken_but_cli_symlink();
                 inherit_interactive_login_shell_environment_if_not_launched_from_terminal();
--- a/crates/gitbutler-tauri/src/main.rs
+++ b/crates/gitbutler-tauri/src/main.rs
@@ -78,7 +78,7 @@ fn main() -> anyhow::Result<()> {
     std::fs::create_dir_all(&app_cache_dir).context("failed to create app cache dir")?;
     std::fs::create_dir_all(&app_log_dir).context("failed to create app log dir")?;
 
-    let tokio_debug = std::env::var_os("GITBUTLER_TOKIO_DEBUG").is_some();
+    let tokio_debug = matches!(std::env::var("GITBUTLER_TOKIO_DEBUG").as_deref(), Ok("1"));
     let app_settings_for_menu = app_settings.clone();
     runtime.block_on(async {
         tauri::async_runtime::set(tokio::runtime::Handle::current());


apps/desktop/src/components/profileSettings/ExperimentalSettings.svelte
```diff
@@ -113,6 +116,35 @@
...
+			events to disk and starts a local gRPC server. Requires app restart.
+		{/snippet}
+		{#snippet actions()}
+			<Toggle
+				id="tokio-console"
+				checked={$settingsStore?.debug.tokioConsoleEnabled}
+				onclick={() => {
+					settingsService.updateDebug({
+						tokioConsoleEnabled: !$settingsStore?.debug.tokioConsoleEnabled
+					});
```
Caleb-T-Owens: Although the update call should always finish before the user has time to react to the pop-up, it would be best to await this call before showing the warning.

apps/desktop/src/components/profileSettings/ExperimentalSettings.svelte
```diff
@@ -113,6 +116,35 @@
...
+					showWarning(
+						'Restart required',
+						'Tokio runtime tracing change takes effect after restart.',
+						{
+							label: 'Restart Now',
+							onClick: () => updaterService.relaunchApp()
```
Caleb-T-Owens: Could we inject `BACKEND` and call `backend.relaunch()` instead? I'm not sure restarting the app via the updater is the best fit.

apps/desktop/src/components/profileSettings/ExperimentalSettings.svelte
```diff
@@ -113,6 +116,35 @@
...
+						}
+					);
```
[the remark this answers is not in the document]
smitmartijn: done, it awaits the setting update now.

crates/but-api/src/legacy/settings.rs
```diff
@@ -94,3 +94,13 @@ pub struct UpdateUiParams {
...
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct UpdateDebugParams {
+    pub update: DebugUpdate,
+}
+
+pub fn update_debug(app_settings_sync: &AppSettingsWithDiskSync, params: UpdateDebugParams) -> Result<()> {
+    app_settings_sync.update_debug(params.update)
+}
```
Caleb-T-Owens: I see this has been added to gitbutler-tauri, can we also make sure it's provided to the `but-server`?

Caleb-T-Owens (changes requested):
Thank you for the contribution! This seems like a good change to make, especially since we do not use the tokio-console information very often.

I have left some feedback, and I would appreciate it if you could take a look when you have a chance.

apps/desktop/src/components/profileSettings/ExperimentalSettings.svelte
smitmartijn: Done. Copilot suggested the same.

apps/desktop/src/components/profileSettings/ExperimentalSettings.svelte
smitmartijn: Done! I wasn't aware of the `BACKEND` having the same function

crates/but-api/src/legacy/settings.rs
smitmartijn: Done, I believe this is enough? https://github.com/gitbutlerapp/gitbutler/pull/12414/changes#diff-c779d05ad4fb56f37c8618735f347e6a720ca11742d2c8b64edb20e0dfdbdfee 

If not, please advise

Byron:
Thanks a lot - this looks promising!

Here are some observations.

* The `restart` functionality doesn't seem to work.
    - My suggestion is to cut it entirely instead of trying to figuring it out.
    - This could be an artefact of the dev mode of the app I was using, and a local nightly build might not have that problem. But I don't know if it's worth it to experiment more.

https://github.com/user-attachments/assets/e3fe0c72-d9cb-48dd-b397-bfb1621261d9

* It generally works, so the log won't be created by default.

https://github.com/user-attachments/assets/b0a0c9c5-3f81-453d-994e-d294bad0aabf

* And if the checkbox is turned on, it will start writing the log like before.

https://github.com/user-attachments/assets/8c0b549e-94a6-4182-aa8c-dbb2f2e991c4


So I think overall it's good, but I'd trim some UI complexity.

Also I want to note that there is more here: Why would it have 16k loggable events per second continuously in the first place? I can't reproduce it, and this is what I get per second in a repository of any size.

```

{"Waker":{"id":4507173040160792,"op":{"Wake":{"self_wake":false}},"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356120750}}}
{"Enter":{"id":4507173040160792,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356144417}}}
{"Enter":{"id":6757598464311299,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356148792}}}
{"Enter":{"id":36030996042219546,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356150292}}}
{"Enter":{"id":2199023255563,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356154709}}}
{"Enter":{"id":2199023255564,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356155750}}}
{"Waker":{"id":4507173040160792,"op":"Clone","at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356167834}}}
{"Exit":{"id":2199023255564,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356172625}}}
{"Exit":{"id":2199023255563,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356173750}}}
{"Exit":{"id":36030996042219546,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356177542}}}
{"Exit":{"id":6757598464311299,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356178042}}}
{"Close":{"id":36030996042219547,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356178750}}}
{"Close":{"id":36030996042219546,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356180542}}}
{"Enter":{"id":6757598464311299,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356447125}}}
{"Enter":{"id":38282795855904795,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356454834}}}
{"Exit":{"id":38282795855904795,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356455667}}}
{"Exit":{"id":6757598464311299,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356456042}}}
{"Enter":{"id":6757598464311299,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356456500}}}
{"Enter":{"id":38282795855904795,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356456667}}}
{"Enter":{"id":2199023255563,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356457084}}}
{"Enter":{"id":2199023255564,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356457334}}}
{"Waker":{"id":4507173040160792,"op":"Clone","at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356460125}}}
{"Waker":{"id":4507173040160792,"op":"Drop","at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356460834}}}
{"Exit":{"id":2199023255564,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356462250}}}
{"Exit":{"id":2199023255563,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356462459}}}
{"Exit":{"id":38282795855904795,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356463500}}}
{"Exit":{"id":6757598464311299,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356463875}}}
{"Exit":{"id":4507173040160792,"at":{"secs_since_epoch":1771475186,"nanos_since_epoch":356464000}}}

```

The reason I bring this up is that removing this log by default will remove a standard debug option as well: We can't ask to check for panics anymore, which only show up there, unfortunately, or when the application is launched from a terminal.

If in doubt, I'd want avoid crazy IO like this, after all, we can still ask people to turn on the log or launch from the terminal. The latter is admittedly what I'd do, I'd never ask to turn it on as it's always noisy compared to what we'd get from a terminal-launch.

**This makes me wonder if we don't just want to turn it off permanently?**

After all, the logs also get cleared after a restart and thus its contents is quite fragile.

smitmartijn:
Hi Byron,

> The restart functionality doesn't seem to work.

Yes, that is because of the dev build - I saw the same, tried it with a release build and that worked as expected. My thought it that `BACKEND` restarts the app, but doesn't keep the debug environment or something.

> Why would it have 16k loggable events per second continuously in the first place? I can't reproduce it, and this is what I get per second in a repository of any size.

I think it's related to the number of repositories, not per se the size. I've got 79 repos in my production app, where it does this in the extreme. The debug build wasn't doing it so heavy, which only had 1 repo. 

> We can't ask to check for panics anymore
> This makes me wonder if we don't just want to turn it off permanently?

I honestly think tokio-rs is the wrong approach if you're just interested in panics. Tokio is designed for diagnostics tracing and debugging info for sessions, not for always-on crash monitoring. Using something like Sentry, [human-panic](https://github.com/rust-cli/human-panic),  might be better.

Either way, the restart prompt was to improve UX to make sure someone restarts the app. Same for the interface setting, just to make it easier for you guys to ask for the trace logs. If you'd rather have a environment variable, that could be a pretty easy refactor.

Byron:
> > Why would it have 16k loggable events per second continuously in the first place? I can't reproduce it, and this is what I get per second in a repository of any size.
> 
> I think it's related to the number of repositories, not per se the size. I've got 79 repos in my production app, where it does this in the extreme. The debug build wasn't doing it so heavy, which only had 1 repo.

That definitely shouldn't be possible and it would be a serious bug. Maybe something worth investigating. There is an actual `tokio-console` application that shows what's going on in `tokio`, maybe that would reveal what kind of tasks are spawned there. Filesystem watchers are only installed for the active project and torn down otherwise.

> I honestly think tokio-rs is the wrong approach if you're just interested in panics. Tokio is designed for diagnostics tracing and debugging info for sessions, not for always-on crash monitoring. Using something like Sentry, [human-panic](https://github.com/rust-cli/human-panic), might be better.

These panics happen in tokio-tasks, which `tauri` spawns for each command that the frontend invokes. By now we have our own proc macro which would allow to catch panics ourselves, and handle them separately and possibly even without leaving the frontend waiting for a response that never happens. I think this is the way to go and we can disregard the tokio-console log more even.

> Either way, the restart prompt was to improve UX to make sure someone restarts the app. Same for the interface setting, just to make it easier for you guys to ask for the trace logs. If you'd rather have a environment variable, that could be a pretty easy refactor.

Now that I think about it… I think we can slim this PR down to allowing the console log to be enabled with an environment variable. Let's call it `GITBUTLER_TOKIO_DEBUG=1` or something, and we can upgrade this to one day also support the interactive tokio-console for realtime inspection.

Then we can find a better way to handle panics, logging them to the official log for starters.

Thanks again for getting this conversation started, and for your help here - it's much appreciated.

smitmartijn:
> That definitely shouldn't be possible and it would be a serious bug. Maybe something worth investigating. There is an actual `tokio-console` application that shows what's going on in `tokio`, maybe that would reveal what kind of tasks are spawned there. Filesystem watchers are only installed for the active project and torn down otherwise.

I took a small swing, but failed. I can't get the dev build to read my release builds' repos to reproduce it with logging which functions are calling tokio-consoles' event emitters. Still, the base for these logs is like 80-100 per second on a clean install, which is too much to have always on. It'd be hard to get panics from it, I think.

> Now that I think about it… I think we can slim this PR down to allowing the console log to be enabled with an environment variable. Let's call it `GITBUTLER_TOKIO_DEBUG=1` or something, and we can upgrade this to one day also support the interactive tokio-console for realtime inspection.

Done! I'll leave the rest of the discussion on what to do with panic logging to you guys.

> Thanks again for getting this conversation started, and for your help here - it's much appreciated.

No worries, I ❤️ the app and happy to contribute in some way.

Byron:
That's fantastic, thanks!

I am going to take this PR over and merge it, now that we will see (most) panics propagated into the frontend.

Byron (approved):
Now there is the `GITBUTLER_TOKIO_DEBUG` environment variable that can be used to enable the tokio log.
This can be relevant for panics that happen outside of frontend calls, like in the filesystem handler.
Fortunately, if that happens, it's related to the status or workspace projection, and that can also be triggered by the user by default, so we wouldn't really miss anything.

Chances are that we will not use the tokio console log again.

crates/gitbutler-tauri/src/logs.rs
```diff
@@ -5,7 +5,7 @@ use tracing::{Level, instrument, metadata::LevelFilter, subscriber::set_global_d
 use tracing_appender::rolling::{RollingFileAppender, Rotation};
 use tracing_subscriber::{Layer, filter::filter_fn, fmt::format::FmtSpan, layer::SubscriberExt};
 
-pub fn init(app_handle: &AppHandle, logs_dir: &Path, performance_logging: bool) {
+pub fn init(app_handle: &AppHandle, logs_dir: &Path, performance_logging: bool, enable_tokio_log: bool) {
```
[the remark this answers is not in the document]
Byron: Good point

crates/gitbutler-tauri/src/main.rs
```diff
@@ -78,6 +78,7 @@ fn main() -> anyhow::Result<()> {
     std::fs::create_dir_all(&app_cache_dir).context("failed to create app cache dir")?;
     std::fs::create_dir_all(&app_log_dir).context("failed to create app log dir")?;
 
+    let tokio_debug = std::env::var_os("GITBUTLER_TOKIO_DEBUG").is_some();
```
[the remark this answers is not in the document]
Byron: Right, better

crates/gitbutler-tauri/src/logs.rs
```diff
@@ -5,7 +5,7 @@ use tracing::{Level, instrument, metadata::LevelFilter, subscriber::set_global_d
 use tracing_appender::rolling::{RollingFileAppender, Rotation};
 use tracing_subscriber::{Layer, filter::filter_fn, fmt::format::FmtSpan, layer::SubscriberExt};
```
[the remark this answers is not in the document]
Byron: It's a trait, needed.

apps/desktop/src/components/profileSettings/ExperimentalSettings.svelte
Caleb-T-Owens: Ahh! I've been replaced by AI already 😅

Caleb-T-Owens (approved):
Thank you for working on this!

PR #12416, opened 2026-02-18
feat(but): make `but pr new` more markdown friendly

slarse:
<details>
<summary>First attempt, kept for historical context</summary>

Previously, `but pr new` would strip any line starting with `#` (as they were considered comments). This meshes poorly with Markdown, as that's how we denote a heading, and those would just be stripped. Rather unexpected.

This PR changes the behavior s.t. We have a _marker line_ instead, after which instructions follow.

> Note: If the user removes the marker line, `but` gets grumpy and complains.

In addition, I've also taken the liberty of:

* Making the temporary file created for the PR message a `.md` file, letting the editor provide syntax highlighting if available.
* Adding in information about how a file was changed (addition, deletion, modification, rename)
    - I think we could do more with this

The editing experience (for _me_) changes as follows:

Before:
before

After:
> Note: there is an extra blank line between the template and instructions marker line in this image. That's been removed
after

## :warning: Concerns/requests for comments
* Inconsistent with the other edit experiences (e.g. commit, where `#` is a comment)
* It's less clear now what is comments and what isn't.
* The instructions marker is uuuuugly
* People may be tempted to remove the instructions marker and that will fail the submit. I'm not sure what's better - risk an accidental remove with a bunch of instructions going into the PR description, or frustrate users that intentionally removed all instructions.

</details>

Changes the editing experience in the following way:

* Uses HTML comments for comments instead of lines starting with `#`, which means that you can actually fully utilize Markdown. The comments are stripped before submit.
* The temporary file gets a `.md` suffix, which will make editors add syntax highlighting (if available, but it pretty much _always_ is for Markdown).
* Adds the change status to the files list of commits (i.e. A, M etc)

Before:
before

After:

but_pr_new

The PR that was created with the above submit can be found here: https://github.com/gitbutlerapp/gitbutler/pull/12423

### DIFF
--- a/crates/but/src/command/legacy/forge/review.rs
+++ b/crates/but/src/command/legacy/forge/review.rs
@@ -698,15 +698,17 @@ fn get_pr_title_and_body_from_editor(
     }
 
     // Add instructions as comments
-    template.push_str("\n# PR Title and Description for branch: ");
-    template.push_str(branch_name);
-    template.push_str("\n#\n");
-    template.push_str("# The FIRST LINE of this file will be the PR title.\n");
-    template.push_str("# Everything AFTER the first line will be the PR description.\n");
-    template.push_str("#\n");
-    template.push_str("# Lines starting with '#' will be ignored.\n");
-    template.push_str("# An empty title (first line) aborts the operation.\n");
-    template.push_str("#\n");
+    let mut instructions = format!(
+        "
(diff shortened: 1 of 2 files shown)


Byron:
@codex review

slarse:
@codex review

slarse:
Gosh darn it...

PR #12421, opened 2026-02-19
Use `sapling-streampager` instead of `minus` for paging

Byron:
This makes it possible to keep the visible content on screen, probably because it doesn't use an alaternate screen.

However, this still won't reprint the lines that are 'above' the screen.

https://github.com/user-attachments/assets/abce6b30-289d-485f-8f63-d353488df1f7

### *Benefits* of `streampager` over `minus`

- It seems to be better with large content as it doesn't buffer everything, somehow.

### *Shortcomings* of `streampager` over `minus`

- It has more dependencies, and bundles `fancy-regex`

### Tasks

* [ ] repeat lines above the screen without duplicating content. This might need a patch in the streampager.
    - if we'd have to patch it, we might as well patch and vendor `minus`, which was unresponsive to my previous PR
    - maybe `streampager`, the original, would be more responsive to PRs
* [ ] thorough testing - how does it work on Windows?
* [ ] use the pager more, but validate how it interacts with input. Probably it's an opt-in affair anyway.
* [ ] Thorough review and refactor (the PR is fully vibed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -153,7 +153,7 @@ dependencies = [
  "anstyle-lossy",
  "anstyle-parse",
  "html-escape",
- "unicode-width",
+ "unicode-width 0.2.2",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


Byron:
A lot has changed (including priorities)

PR #12422, opened 2026-02-19
Add the ability to create a draft review to the CLI

estib-vega:
- Add a draft flag to the review creation command
- Update the review terminology

### Usage
```sh
but pr -d
but pr --draft
but pr new -d
but pr new --draft
```

### DIFF
--- a/crates/but/src/lib.rs
+++ b/crates/but/src/lib.rs
@@ -798,7 +798,10 @@ async fn match_subcommand(
                 .emit_metrics(metrics_ctx)
         }
         #[cfg(feature = "legacy")]
-        Subcommands::Pr(forge::pr::Platform { cmd }) => {
+        Subcommands::Pr(forge::pr::Platform {
+            cmd,
+            draft: top_level_draft,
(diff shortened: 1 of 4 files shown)


crates/but/src/lib.rs
```diff
@@ -816,7 +819,9 @@ async fn match_subcommand(
                     with_force,
                     run_hooks,
                     default,
+                    draft,
                 }) => {
+                    let draft = top_level_draft || draft;
```
estib-vega, line 824: If either the top level one or the specific one is set to true, consider it as true.

jimpriest:
Thanks for adding this!

lkuoch:
@estib-vega is this also available in the ui?

estib-vega:
> @estib-vega is this also available in the ui?
Screenshot 2026-03-05 at 09 25 19

@lkuoch Yes! ☝️ 
When creating a PR in the UI, you have the option to select the type of PR that will be opened through the dropdown.
The GUI will remember the last choice you selected.

PR #12423, opened 2026-02-19
Add hello to README

slarse:
## 🧢 Changes
This is just a test of the Markdown-friendly changes to `but pr new`

## ☕️ Reasoning
:)


slarse:
This is what it looked like in my editor (VS Code stock looks almost identical).

but_pr_new

PR #12396, opened 2026-02-17
setup but-sdk

estib-vega:
Setup the but sdk.

### More context
At this point we have the main two pillars in place:
- The way to arbitrarily generate TS types out of Json Schema-enabled types.
- The way to generate the bindings for Node to access the Rust APIs.

This PR doesn’t contain the actual types generated since we haven’t at this point created the mechanism for generating bindings out of the `but-api` functions. That will come later

---
This is **part 2 of 5 in a stack** made with GitButler:
- <kbd>&nbsp;5&nbsp;</kbd> #12405 
- <kbd>&nbsp;4&nbsp;</kbd> #12388 
- <kbd>&nbsp;3&nbsp;</kbd> #12398 
- <kbd>&nbsp;2&nbsp;</kbd> #12396 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #12395

PR #12427, opened 2026-02-19
fix(but-sdk): align entrypoints/files with generated output

Byron:
Fixes unresolved Copilot packaging breakage review threads.

Refs: https://github.com/gitbutlerapp/gitbutler/pull/12396


estib-vega:
@Byron I think this is already the case in master

PR #12428, opened 2026-02-19
fix(but-api-macros): validate i64->usize/isize remap bounds

Byron:
Fixes unresolved Copilot overflow/truncation review thread.

### DIFF
--- a/crates/but-api-macros/src/lib.rs
+++ b/crates/but-api-macros/src/lib.rs
@@ -765,12 +765,36 @@ fn build_napi_params<'a>(
                 }
                 _ => {
                     // For all other types: check for napi-incompatible types first
-                    if let Some((napi_ty, cast_expr)) = napi_type_remap(base_ty) {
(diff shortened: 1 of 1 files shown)


Byron:
I was pondering about this and think it's worth doing. The reason here is that we also seem to convert from `isize` to `usize`, and that can easily be failing if for some reason it's negative. Then we don't want to accidentally have some very high `usize` in there, but fail instead.

Byron:
@estib-vega I merged this because I think it's right, but I didn't test it. Maybe run it through its paces.

estib-vega:
@Byron please don't merge this changes without testing. Or please ping me to test this before merging.

estib-vega:
It seems to work 👍 
I'm still working on this, so ideally we don't merge changes to this while I finish the follow-up changes

Byron:
Sorry for the hassle, I was too quick on the trigger. Glad it worked, will wait next time.

PR #12429, opened 2026-02-19
fix(installer): require explicit env var to skip Linux signature verification

Byron:
Fixes unresolved Copilot security-risk review thread.

Refs: https://github.com/gitbutlerapp/gitbutler/pull/12393

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -398,10 +398,14 @@ jobs:
           path: .
       - run: chmod +x ./but-installer
       - name: Install release
+        env:
+          GITBUTLER_SKIP_SIGNATURE_VERIFICATION: ${{ startsWith(matrix.platform, 'ubuntu') && '1' || '0' }}
         run: ./but-installer
       - name: Validate release version
         run: but --version | grep "but $LATEST_RELEASE_VERSION"
       - name: Install nightly
+        env:
+          GITBUTLER_SKIP_SIGNATURE_VERIFICATION: ${{ startsWith(matrix.platform, 'ubuntu') && '1' || '0' }}
         run: ./but-installer nightly
       - name: Validate nightly version
         run: but --version | grep "but $LATEST_NIGHTLY_VERSION"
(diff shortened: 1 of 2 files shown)


Byron:
@slarse I came up with a prompt after [this thought occurred](https://discord.com/channels/1060193121130000425/1272844201914142834/1473980155197390941) and this came out of it. I like the specificity of this special feature, but is it a keeper?
Please feel free to merge or close or make changes. Thank you.

slarse:
@Byron I ignored this Copilot comment on purpose, but I forgot to resolve the conversation. I just commented on it like a doofus :). I keep talking to bots ...

Here's the reasoning for why I made it "insecure" (it'll only be there until we get a proper release with a signature for the CLI): https://github.com/gitbutlerapp/gitbutler/pull/12393#discussion_r2816958614

But I like the idea of automatically going over Copilot comments.

slarse:
Closing as "overkill" as the signature skip will be removed as soon as the next release is published (which is today).

Fix PR here: https://github.com/gitbutlerapp/gitbutler/pull/12437

slarse:
@Byron Perhaps we should do this for _unengaged_ Copilot comments? I'm 97% sure I'll keep just reponding to Copilot as if it were a human. It's not _as_ insane as it sounds as I like to put justifications in writing.

Another idea is that we could also add a "no unresolved conversations" rule to merging. That would however completely block the auto-merge flow where you don't wait for Copilot to finish, so perhaps that's too disruptive.

But to be completely honest, I think you _should_ wait the few minutes that are necessary for code review to hit, so prohibiting merge with unresolved conversations is IMO not a bad idea. Speed now is paid for by bugs later. But others may disagree, it's definitely a conversation to have.

Byron:
> Another idea is that we could also add a "no unresolved conversations" rule to merging. That would however completely block the auto-merge flow where you don't wait for Copilot to finish, so perhaps that's too disruptive.

This is fantastic! I didn't know that was possible, and it would solve the Copilot review problem. Would you be the one to start a discussion on this? Or should it be me? I think you are good at this, which is why I am asking 😅.

slarse:
> Would you be the one to start a discussion on this?

Sure, I don't mind rocking the boat a little bit. It's what I get paid for. Or maybe it's something else, I forget. It's on tomorrow's TODO list :)

slarse:
Just to ensure I wasn't hallucinating, I checked branch protection rules and the setting does indeed exist outside of my tiny skull:

require_resolution

PR #12430, opened 2026-02-19
fix(but-claude): save commit notifications using resolved session id

Byron:
Fixes unresolved Copilot session-association bug thread.

Refs: https://github.com/gitbutlerapp/gitbutler/pull/12394


Byron:
@mtsgrd This came up as part of [this idea](https://discord.com/channels/1060193121130000425/1272844201914142834/1473980155197390941). However, I think if one wanted to clean that up, more work is needed. Maybe it's not worth it though if it just keeps working, and I prevented myself from being sucked into this 😅.


I validated that it's still working locally, existing sessions can be found, new sessions can be created.

PR #12395, opened 2026-02-17
but-schema-gen: Yet another way of creating TS types

estib-vega:
Create a new create, that exports ts types from JSON schemas.

### More context
This is part of the effort of creating a TS SDK using NAPI bindings.

In order to do that we’ll need a couple of pieces:
- A way to create the actual bindings (napi-rs)
- A way to type the whole thing in a safe way (napi-rs, our own typing sauce)
- A way to generate the bindings out of the but-api, ideally without having to maintain multiple sources of truth (changes to the `but_api` macro)
-  A NPM package we can use, and later publish (napi-rs)

### The changes in this PR
Here we’re only adding our own typing sauce to the mix.
This allows for the on-demand creation of TypesScript types into the file of our choosing, out of structs (and other types) that derive `schemers::JsonSchema`.

This accomplishes two things:
1. A more flexible, in house alternative to ts-rs.
2. A mechanism to export types that will be needed in the napi-generated TS bindings.

---
This is **part 1 of 5 in a stack** made with GitButler:
- <kbd>&nbsp;5&nbsp;</kbd> #12405 
- <kbd>&nbsp;4&nbsp;</kbd> #12388 
- <kbd>&nbsp;3&nbsp;</kbd> #12398 
- <kbd>&nbsp;2&nbsp;</kbd> #12396 
- <kbd>&nbsp;1&nbsp;</kbd> #12395 👈

PR #12431, opened 2026-02-19
fix(but-schema-gen): overwrite output to make generation idempotent

Byron:
Fixes unresolved Copilot duplicate-output review thread.

Refs: https://github.com/gitbutlerapp/gitbutler/pull/12395


Byron:
Yeah, this seems wrong.

PR #12440, opened 2026-02-19
but-sdk: Run checks on CI

estib-vega:
Build and run the checks for but-sdk on CI when changing rust files or but-sdk itself.

### Why is this not run as part of the other npm packages checks?

Funny you ask. Was just thinking about that.
We can't generate all the types for the but-sdk, without explicitely compiling the Node addons. So packaging the but-sdk if a bit more chunky and requires more deps than the other NPM packages in this repo.

That's why I chose to make this a separate step in the `rust-checks` suite, rather than making it part of the turbo `package` jobs. That way we don't need to run it always.

In the future, we'll also need to compile the addons for different targets (macOS, windows, other linux targets). We'll deal with that as we go on.

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -19,6 +19,7 @@ jobs:
       node: ${{ steps.filter.outputs.node }}
       rust: ${{ steps.filter.outputs.rust }}
       ui: ${{ steps.filter.outputs.ui }}
+      but_sdk: ${{ steps.filter.outputs.but_sdk }}
       docs: ${{ steps.filter.outputs.docs }}
       but_installer: ${{ steps.filter.outputs.but_installer }}
       install_script: ${{ steps.filter.outputs.install_script }}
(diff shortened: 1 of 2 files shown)


.github/workflows/push.yaml
```diff
@@ -122,6 +131,38 @@ jobs:
...
+      CARGO_TERM_COLOR: always
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - name: Cache apt archives
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5
+        with:
+          path: .apt-cache
+          key: ${{ runner.os }}-apt-but-sdk-${{ hashFiles('.github/ci/but-sdk-apt-packages.txt') }}
```
Byron, line 148: This is very clever, and I think it would be even better if it could use 

https://github.com/gitbutlerapp/gitbutler/blob/792a42633622960a23cf24e760f5b5b0969a4a1d/scripts/install-minimal-debian-dependencies.sh#L7

I think the hashing should still be fine, as the script really doesn't change unless deps change.

I'd really want to avoid keeping the same 'intent' behind different mechanisms.
The `-o dir::cache::archives="$PWD/.apt-cache" ` could just be added as optional parameter to the script.

WDYT @estib-vega ?

estib-vega: Yeah, sounds good to me 👌

Byron: @copilot implement https://github.com/gitbutlerapp/gitbutler/pull/12440#discussion_r2828882058

Edit: apparently it doesn't listen here 😅

Byron: Follow up: https://github.com/gitbutlerapp/gitbutler/pull/12453

PR #12443, opened 2026-02-19
increase installer + curl download timeout from 30s to 300s

craigmayhew:
## 🧢 Changes

 - [30db885032895af1d8b14c6addd72d6b3d30e917] Increase curl timeout from 30s to 300s in install sh script
 - [9965bccc5dd727ab2b46ad72621ccbd648797143] Modified rust installer to wait 300 seconds

## ☕️ Reasoning

Slow conference wifi prevented me installing `but`!

## 🎫 Affected issues

n/a

### DIFF
--- a/crates/but-installer/src/http.rs
+++ b/crates/but-installer/src/http.rs
@@ -5,7 +5,7 @@ use std::time::Duration;
 use anyhow::{Context, Result};
 use curl::easy::Easy;
 
-const REQUEST_TIMEOUT_SECS: u64 = 30;
+const REQUEST_TIMEOUT_SECS: u64 = 300;
 const CONNECT_TIMEOUT_SECS: u64 = 10;
 const MAX_REDIRECTS: u32 = 5;
 const USER_AGENT: &str = concat!("GitButler-Installer/", env!("CARGO_PKG_VERSION"), " (Rust installer)");
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -46,7 +46,7 @@ trap 'rm -f "$INSTALLER_JSON" "$INSTALLER_BIN"' EXIT INT TERM
 INSTALLER_API_URL="https://app.gitbutler.com/installers/info/$INSTALLER_OS/$INSTALLER_ARCH"
 echo "Fetching installer information..."
 
-EFFECTIVE_URL=$(curl --fail --silent --show-error --location --max-redirs 5 --max-time 120 \
+EFFECTIVE_URL=$(curl --fail --silent --show-error --location --max-redirs 5 --max-time 300 \
     -o "$INSTALLER_JSON" -w '%{url_effective}' "$INSTALLER_API_URL") || {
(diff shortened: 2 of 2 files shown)


krlvi:
Thanks for this! <3 The conference wifi is a gift that keeps on giving ;p (who would have thought that downloading 1MB takes more than 30s :)

craigmayhew:
I also needed to modify the rust installer binary itself as that has a hard 30s timeout.
24MB on MacOS and that was the main culprit.

Have compiled and run ./target/debug/but-installer on very slow wifi to prove it now works.

crates/but-installer/src/http.rs
```diff
@@ -5,7 +5,7 @@ use std::time::Duration;
 use anyhow::{Context, Result};
 use curl::easy::Easy;
 
-const REQUEST_TIMEOUT_SECS: u64 = 30;
+const REQUEST_TIMEOUT_SECS: u64 = 300;
```
slarse, line 8: Oof, this timeout was very tight! The binaries can be pretty large (nightly Linux is 66 MB, for example). This was an oversight for sure!

scripts/install.sh
```diff
@@ -46,7 +46,7 @@ trap 'rm -f "$INSTALLER_JSON" "$INSTALLER_BIN"' EXIT INT TERM
 INSTALLER_API_URL="https://app.gitbutler.com/installers/info/$INSTALLER_OS/$INSTALLER_ARCH"
 echo "Fetching installer information..."
 
-EFFECTIVE_URL=$(curl --fail --silent --show-error --location --max-redirs 5 --max-time 30 \
+EFFECTIVE_URL=$(curl --fail --silent --show-error --location --max-redirs 5 --max-time 300 \
```
slarse, line 49: While I don't think this is where the issue was (the app is ~25-65x the size of the installer, so the app download timeout is still by far the most limiting factor), I don't see any particular harm in increasing this timeout as well.

As long as there is _some_ timeout!

slarse (approved):
Thanks, looks good!

@craigmayhew Unfortunately you have one unsigned commit. We require signed commits as of... yesterday I think. Could you please sign it?

slarse (changes requested):
Actually, I'll request changes to make it clear. Sorry about the confusion, I missed that detail.

@craigmayhew Unfortunately you have one unsigned commit. We require signed commits as of... yesterday I think. Could you please sign [12f49e4](https://github.com/gitbutlerapp/gitbutler/pull/12443/commits/12f49e4364baa892607df617fcbbd6e8a859d77f)?

craigmayhew:
> @craigmayhew Unfortunately you have one unsigned commit. We require signed commits as of... yesterday I think. Could you please sign [12f49e4](https://github.com/gitbutlerapp/gitbutler/pull/12443/commits/12f49e4364baa892607df617fcbbd6e8a859d77f)?

@slarse An oversight on my part - I have signed and pushed :)

slarse (approved):
Thanks @craigmayhew!

PR #12446, opened 2026-02-19
Use nucleo-matcher for VSCode-style file search

mtsgrd:
Replace the previous skim-based fuzzy matcher with nucleo-matcher to
provide path-aware matching similar to VSCode's Cmd+P. The new
implementation:

- swaps fuzzy-matcher for nucleo-matcher in Cargo.toml
- bonuses for matches after '/', word boundaries and camelCase
  transitions
- updates compute_match_score to combine nucleo's path score with a
  filename bonus and a mild depth penalty

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3457,15 +3457,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "fuzzy-matcher"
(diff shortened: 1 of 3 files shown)


krlvi:
do we still need this PR?

PR #12452, opened 2026-02-20
Restore required token permissions in publish workflow

Byron:
## Summary
This PR addresses an unresolved Copilot review comment from source PR #12441.

## Commits
- `851ab7825b` Restore publish workflow token scopes
  - Replaces top-level `permissions: {}` in `.github/workflows/publish.yaml` with minimal required permissions:
    - `contents: read`
    - `actions: write`
  - This keeps least-privilege defaults while allowing checkout/artifact/cache operations used by the workflow.
  - Fixes up: https://github.com/gitbutlerapp/gitbutler/pull/12441#discussion_r2828390526

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -29,7 +29,9 @@ on:
           - minor
           - major
 
-permissions: {}
+permissions:
+  contents: read
+  actions: write
 
 jobs:
   build-sveltekit:


Byron:
This seems wrong - empty permissions means no permissions, which is the amount of permissions needed to run this (I think).

slarse:
@Byron Yep, Copilot was wrong about that initial comment (but I did resolve that conversation so I'm curious as to why this PR was opened?). The `permissions` attribute sets permissions for the `GITHUB_TOKEN`, but `actions/upload-artifact` and all related actions actually use an undocumented `ACTIONS_RUNTIME_TOKEN` for permissions. Over which I don't think we have _any_ control. Found some breadcrumbs about it [over here](https://github.com/actions/upload-artifact/issues/197#issuecomment-832279436).

PR #12453, opened 2026-02-20
Reuse minimal Debian deps script in but-sdk CI job

Byron:
The but-sdk workflow duplicated apt install intent that already exists in scripts/install-minimal-debian-dependencies.sh. Keeping the same dependency intent in two places increases drift risk and makes updates easy to miss.

Refactor the but-sdk build-check job to call the shared script instead of installing from a separate package list file. Add an optional cache-archives directory argument to the script so CI can still use actions/cache with dir::cache::archives.

Also switch the apt cache key hash to the script and remove the now-redundant .github/ci/but-sdk-apt-packages.txt file.

Motivation: make dependency installation a single source of truth while preserving CI apt caching behavior.

### DIFF
--- a/.github/ci/but-sdk-apt-packages.txt
+++ b/.github/ci/but-sdk-apt-packages.txt
@@ -1,2 +1,0 @@
-libdbus-1-dev
-pkg-config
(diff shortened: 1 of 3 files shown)


Byron:
I think this works and is correct, CC @estib-vega .

estib-vega:
Awesome thanks!

PR #12454, opened 2026-02-20
feat: enable force push protection by default

slarse:
## 🧢 Changes
Enables force push protection by default. I feel like we want this?

[Suggested in Discord](https://discord.com/channels/1060193121130000425/1073202153163857920/1473661110954496070)


slarse:
No time to finish this right now.

PR #12455, opened 2026-02-20
Landing: Update download links

PavelLaptev:
image
image
image

---

### Changes:
- new download cards
- updated CLI button for Linux

---

### TODO:
- add valid CLI Link for Linux. The current on is just a placeholder.


krlvi:
Nice! I am wondering, would it be possible to also make the curl command text selectable?

PavelLaptev:
> Nice! I am wondering, would it be possible to also make the curl command text selectable? 

Why? You copy it on click

PavelLaptev:
I wouldn't do it because the whole thing it's a button. Remember the blog situation, it will be hard to select the text and not click on the button.

PR #12457, opened 2026-02-20
ci: turn on optimizations for nightlies as well

slarse:
## 🧢 Changes
Turns on all optimizations for all builds (which in this case means nightlies also get optimizations).

## ☕️ Reasoning
I've observed that there isn't much of a difference in total CI time between release and nightly, so I want to try optimizing all builds. I don't think we'll notice much of a difference.

The spark of this was that I observed the nightly CLI binary for Linux to be 67 MB large, compared to the 34 MB of the release binary.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -189,13 +189,6 @@ jobs:
         with:
           name: sveltekit-build
           path: ./apps/desktop/build/
-      - name: Set release build optimization env vars
-        shell: bash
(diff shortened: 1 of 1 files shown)


Byron (approved):
Let's do it!

I'd also be interested in seeing the effects on release performance (CI wallclock).

slarse:
> I'd also be interested in seeing the effects on release performance (CI wallclock).

Let's let it run over the weekend so we have a few builds to compare to, and then decide if it was a good idea or not!

PR #12462, opened 2026-02-20
CI check for commit signing

krlvi:


.github/workflows/verify-commit-signatures.yml
```diff
@@ -0,0 +1,26 @@
+name: Verify Commit Signatures
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
```
slarse, line 8: Why these types specifically?

.github/workflows/verify-commit-signatures.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Check commits are signed and Verified
+        uses: 1Password/check-signed-commits-action@ed2885f3ed2577a4f5d3c3fe895432a557d23d52 # v1.2.0
+        with:
+          comment: |
+            This pull request contains commits that are not **Verified** on GitHub.
+
+            GitButler requires signed commits for all pull requests. Please re-sign the commits and force-push your branch.
+
+            If you need help setting up signing, start here:
+            https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification
```
slarse, line 26: Do you intend for this to publish PR comments about there being unsigned commits? That won't work for external contributors (the `pull_request` event from a fork does not allow write access to the repository, not even PR comments). Only `pull_request_target` allows that, but that should itself _never be used_ because it allows PRs from forks access to all kinds of things, including secrets.

So, could you elaborate on why we need this workflow? It would only work for internal contributors, which I don't feel is necessarily all that useful?

slarse (commented):
@krlvi I'm actually a bit confused about what the point of this is? The `Require signed commits` check already requires the author to have verified signatures on all commits.

require_signed_commits

.github/workflows/verify-commit-signatures.yml
krlvi: It was indeed for external contributors. And you are raising a good point. In this case this workflow is worthless indeed

krlvi:
During review we learned that this is not actually valuable

.github/workflows/verify-commit-signatures.yml
slarse: I think `pull_request_target` can be used responsibly, but it's a very scary thing...

For these kinds of tasks, I personally prefer having bots do the work rather than the actions themselves. We could have a triage bot that does these kinds of things. Way less scary.

PR #12463, opened 2026-02-20
fix(but): never use askpass broker for cli

### LINKED ISSUE #12279: `but pull` fails: `askpass broker not initialized`
### Version

0.19.0

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

i have a password protected ssh key which is marked as the required identity for the repo's configured remote. with git, this results in a prompt that asks for the password to my ssh key. i have `ssh-askpass` installed off of brew. for the terminal emulator, im using `ghostty`.

calling `but pull` results in an error:

```
RUST_BACKTRACE=full but pull
Fetching newest data from remotes...

thread '<unnamed>' (24011034) panicked at crates/gitbutler-repo-actions/src/askpass.rs:29:45:
askpass broker not initialized
stack backtrace:
   0:        0x105990338 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h8c5ab870ebd4c242
   1:        0x1052bbf38 - core::fmt::write::h64810b21425781ec
   2:        0x10598f02c - std::io::Write::write_fmt::h947344a43c638468
   3:        0x10598f29c - std::panicking::default_hook::h7aae5c8c53b8a669
   4:        0x10598ec24 - std::panicking::panic_with_hook::hb885411bb59bebf4
   5:        0x1059ab9cc - std::panicking::panic_handler::{{closure}}::hb04320f9e43192c5
   6:        0x1059ab988 - std::sys::backtrace::__rust_end_short_backtrace::ha3c55a1c4767eecf
   7:        0x1059ab97c - __rustc[d131491b17107b07]::rust_begin_unwind
   8:        0x105f27668 - core::panicking::panic_fmt::h6f8816b337451651
   9:        0x105f27994 - core::option::expect_failed::h0c7474fae24b826b
  10:        0x1053c2784 - gitbutler_git::repository::fetch::{{closure}}::h994db49f37059f9b
  11:        0x1053c05c4 - <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll::h3d7f4ed57e45df59
  12:        0x1053bee40 - std::sys::backtrace::__rust_begin_short_backtrace::h9db85d95e5434dd8
  13:        0x1053c31ec - core::ops::function::FnOnce::call_once{{vtable.shim}}::h8a50b28e03a63c50
  14:        0x1059b4c3c - std::sys::thread::unix::Thread::new::thread_start::hd42ed9e591878587
  15:        0x19aba2c08 - __pthread_cond_wait

thread 'main' (24010968) panicked at crates/gitbutler-repo-actions/src/repository.rs:290:14:
called `Result::unwrap()` on an `Err` value: Any { .. }
stack backtrace:
   0:        0x105990338 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h8c5ab870ebd4c242
   1:        0x1052bbf38 - core::fmt::write::h64810b21425781ec
   2:        0x10598f02c - std::io::Write::write_fmt::h947344a43c638468
   3:        0x10598f29c - std::panicking::default_hook::h7aae5c8c53b8a669
   4:        0x10598ec24 - std::panicking::panic_with_hook::hb885411bb59bebf4
   5:        0x1059ab9cc - std::panicking::panic_handler::{{closure}}::hb04320f9e43192c5
   6:        0x1059ab988 - std::sys::backtrace::__rust_end_short_backtrace::ha3c55a1c4767eecf
   7:        0x1059ab97c - __rustc[d131491b17107b07]::rust_begin_unwind
   8:        0x105f27668 - core::panicking::panic_fmt::h6f8816b337451651
   9:        0x105f279dc - core::result::unwrap_failed::h1f9c817951eee5c5
  10:        0x1053c7f94 - <but_ctx::Context as gitbutler_repo_actions::repository::RepoActionsExt>::fetch::hd5bf823b370bfb64
  11:        0x105305a88 - core::ops::function::impls::<impl core::ops::function::FnMut<A> for &mut F>::call_mut::h33ebffe3749df2e7
  12:        0x104edd42c - but_api::legacy::virtual_branches::fetch_from_remotes::hdd2e6aa2ffa1cb44
  13:        0x1048b85f4 - but::command::legacy::pull::handle_pull::{{closure}}::h2f29832255ab93a9
  14:        0x1048bb5ec - but::command::legacy::pull::handle::{{closure}}::h916675101cc87da4
  15:        0x10489f8a0 - but::match_subcommand::{{closure}}::hb5dd3e508a9d4afb
  16:        0x104c88320 - <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll::hcbea72921b7a2fb7
  17:        0x1047bea44 - gitbutler_tauri::main::hfe207b0db0d013f8
  18:        0x1048cf8cc - std::sys::backtrace::__rust_begin_short_backtrace::hee64352adb6a0a81
  19:        0x104cbd4c8 - _main
```

from the gui however, it correctly asked for my password through a dialog box as one would expect.

slarse:
## 🧢 Changes
Removes the askpass feature flag and instead figures out at runtime if the askpass broker is active or not.

See the commit message and comments in the code for details.

## ☕️ Reasoning
When running `but` as a symlink to `gitbutler-tauri`, the feature flag will obviously have been set to include the askpass broker. However, the codepath never initializes it. And it doesn't really matter if it does because there's no good mechanism to respond. The CLI should delegate credentials input to whatever the terminal is configured to use (unless we want to build a full on TUI for everything, but we're not there yet).

Fixes #12279


crates/but-api/src/legacy/repo.rs
```diff
@@ -37,6 +42,7 @@ pub async fn git_clone_repository(repository_url: String, target_dir: PathBuf) -
 async fn handle_git_prompt_clone(prompt: String, url: String) -> Option<String> {
     tracing::info!("received prompt for clone of {url}: {prompt:?}");
     askpass::get_broker()
+        .expect("failed to get askpass broker")
```
Byron: With the recent improvements, panics (that this line could introduce) are safer as they are turned into errors.
We still have to be careful about them if they could run outside of GUI triggers, like in the filewatcher, but this isn't the case here.
(no action needed, just sharing some insights)

crates/but/Cargo.toml
```diff
@@ -63,6 +63,9 @@ but-api = { workspace = true, features = ["path-bytes"] }
 but-graph.workspace = true
 but-llm.workspace = true
 
+# this crate is only required so we can disable the askpass broker
+gitbutler-repo-actions.workspace = true
```
Byron: Legacy crates have to be behind the `legacy` feature toggle, and be optional dependencies.
Should also obey the grouping and move down to the reset of 'em.

crates/gitbutler-git/src/repository.rs
```diff
@@ -377,24 +367,23 @@ where
...
+/// properly initialized. IT should return the user's response or `None` if the operation should be
+/// aborted, in which case an `Err` value is returned from this function.
+pub async fn fetch<P, F, Fut, E>(
     repo_path: P,
     executor: E,
     remote: &str,
     refspec: RefSpec,
-    on_prompt: F,
-    extra: Extra,
+    on_prompt: Option<F>,
```
Byron, line 376: I like this.

crates/gitbutler-git/src/repository.rs
```diff
@@ -435,27 +424,26 @@ where
 }
 
 /// Pushes a refspec to the given remote in the repository at the given path.
-/// Any prompts for the user are passed to the asynchronous callback `on_prompt`,
-/// which should return the user's response or `None` if the operation should be
+///
+/// If `on_prompt` is provided, the custom askpass broker machinery is used, which MUST have been
+/// properly initialized. It should return the user's response or `None` if the operation should be
```
Byron: How would it be initialised? You could link the initialisation function via `[crate::init()]` or something like it.

crates/gitbutler-repo-actions/src/askpass.rs
```diff
@@ -1,32 +1,46 @@
...
+        .set(Some(AskpassBroker::init(submit_prompt)))
+        .unwrap_or_else(|_| panic!("broker already configured"))
+}
+
+/// Explicitly disable the global askpass broker.
 ///
-/// This function is **NOT** thread safe.
-#[expect(static_mut_refs)]
-pub unsafe fn init(submit_prompt: impl Fn(PromptEvent<Context>) + Send + Sync + 'static) {
-    unsafe {
```
Byron, line 18: Great!

crates/gitbutler-repo-actions/src/repository.rs
```diff
@@ -335,6 +345,7 @@ async fn handle_git_prompt_push(prompt: String, askpass: Option<Option<StackId>>
     if let Some(branch_id) = askpass {
         tracing::info!("received prompt for branch push {branch_id:?}: {prompt:?}");
         askpass::get_broker()
+            .expect("failed to get askpass broker")
```
Byron: Rather than stating the obvious, it would be better to state what would be expected here. Maybe `init() was called`. In other words, call out the requirement or expectations as it would help fixing the panic if we'd hit it.

Byron (commented):
I took a first look and like where this is going! Thanks for tackling this.

There are a few actionable items as well.

crates/but-api/src/legacy/repo.rs
slarse: Those are good insights. I basically just moved this panic one step up the stack, it was originally inside the `get_broker()` function. At that point I felt like that was better than no improvement and left it there :sweat_smile:.

crates/but/Cargo.toml
slarse: Thanks! Now the optional dependencies make sense to me, it just didn't compute before I read your comment.

crates/gitbutler-git/src/repository.rs
slarse: Thanks for pointing this out, I don't know which one of my braincells wrote this but this isn't even remotely accurate. This crate knows _nothing_ about the askpass broker and does not care whether it's initialized or not, that's a higher level concern for the one passing `on_prompt` and using the broker in the callback.

I've updated this to reflect what actually happens if you specify on_prompt: we override SSH_ASKPASS to point to our own askpass command.

crates/gitbutler-repo-actions/src/repository.rs
slarse: That is a fair observation, fixed.

crates/but-server/Cargo.toml
```diff
@@ -29,6 +29,7 @@ but-ctx.workspace = true
 
 gitbutler-project.workspace = true
 gitbutler-watcher.workspace = true
+gitbutler-repo-actions.workspace = true
```
slarse, line 32: @Byron As no legacy feature flag is used in `but-server`, I wasn't sure if it was considered legacy itself and so did not add a legacy flag. Or if we just don't care much about the details of `but-server` as it's not released as a product.

Just respond here if you think we should have a legacy gate for these deps in `but-server` and I'll add it in (even if this PR is merged already when you see this).

Byron: That's the right choice, to not add a legacy flag. The `but-cli` has that to allow building a 'modern' version, but there is no notion for this in the GUI of any kind.

PR #12467, opened 2026-02-20
Reduce build deps

craigmayhew:
## 🧢 Changes

This PR was the result of using `cargo machete` + some manual testing to remove potentially unused dependencies.

c5a403dd9bd0c681142e49d5d8ddadc1770e9f90..a257dcab193e5ad640a51cdf39ce0b7c7d9c99ac - remove many potentially unused deps
65b079d014d90cd0881af38edab24108e61de623 revert to fix CI error 'gitbutler-tauri' does not contain this feature: error-context
f55970681359e29041412130ae2c1cd1ab22d816 fix CI error as but_meta no a dep and no longer linkable in docs
7a39cd9 fix CI error caused by napi being removed when still needed. Thanks @estib-vega for the help on this one
f0565b63fac668bacda1b78f74f489df51cbaef5 fix CI error as it's better to use the correct Cargo.lock :p 

### Notes for reviewers
 - `error-context = ["dep:backtrace"]`, is this important? Also note the comment `Needed to make tests work with stable change ids`

### Further discussion
 - Is the feature `testing` still used? It has a note saying it fights flaky tests, but I am unsure if this is still in use.
 - Adding a `cargo machete` github action could be a good follow up PR once unused deps have been removed. 

## ☕️ Reasoning

Faster CI. Faster builds.

PR #12470, opened 2026-02-20
Add CI workflow to check for unused deps

craigmayhew:
## 🧢 Changes

Introduce cargo machete to block builds on unused dependencies.

088a0cfefaafdf261a65e9638b4d6b1aff73fbfc introduce machete CI workflow to prevent unused crates, and  ignore current machete false positives

## ☕️ Reasoning

Cargo.toml files had unused crates as seen in #12467 

## 📌 Todos

 - [x] Fix the inevitable false positives!
 - [x] Blocked by #12467

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -207,6 +207,7 @@ jobs:
           rustup component add clippy
           cargo clippy --workspace --all-targets -- -D warnings
 
+  # Find unused cargo dependencies
   cargo-machete:
     needs: changes
     if: ${{ needs.changes.outputs.rust == 'true' }}
@@ -218,11 +219,10 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
+      - name: Install cargo-machete
+        run: cargo install --locked cargo-machete --version 0.9.1
       - name: Run cargo-machete on workspace
-        uses: bnjbvr/cargo-machete@7959c845782fed02ee69303126d4a12d64f1db18 # v0.9.1
-        with:
-          # Point machete at the Rust workspace directory
-          args: .
+        run: cargo machete
 
   cargo-deny:
     needs: changes
@@ -534,6 +534,7 @@ jobs:
       - rust-test
       - rust-test-tauri
       - rust-lint
+      - cargo-machete
       - cargo-doc
       - build-installer
       - but-sdk-build-check
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -207,6 +207,23 @@ jobs:
           rustup component add clippy
           cargo clippy --workspace --all-targets -- -D warnings
 
+  cargo-machete:
+    needs: changes
+    if: ${{ needs.changes.outputs.rust == 'true' }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - name: Run cargo-machete on workspace
+        uses: bnjbvr/cargo-machete@7959c845782fed02ee69303126d4a12d64f1db18 # v0.9.1
+        with:
+          # Point machete at the Rust workspace directory
+          args: .
+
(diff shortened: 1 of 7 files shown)


.github/workflows/machete.yml
```diff
@@ -0,0 +1,29 @@
...
+  detect-unused-dependencies:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@631a55b12751854ce901bb631d5902ceb48146f7 # stable
```
Byron: I don't think we want to re-introduce this.

Byron (changes requested):
Actually this looks great! Thanks a lot for contributing!

Also, I wasn't aware of `cargo machete` and love it.

Is there a reason it's not in the main `push.yaml` workflow?

If you don't feel the need for further back and forth, I could also take it from here, just let me know.

Thanks again.

.github/workflows/machete.yml
craigmayhew: An oversight on my part, I have now aligned it to the container image used in other workflows. This job is no longer an exception!

I have also moved the job to push.yaml as I see now that's a much better home for it.

Let me know any other feedback please! This is a great way for me to be more familiar with the codebase

.github/workflows/push.yaml
```diff
@@ -207,6 +207,25 @@ jobs:
...
 
+  cargo-machete:
+    needs: changes
+    if: ${{ needs.changes.outputs.rust == 'true' }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: read
+    container:
+      image: ghcr.io/gitbutlerapp/ci-base-image@sha256:3eaeae1b07072796a53dc1e7585cdc1f462d0b11d10ecd3685c6fa8082d647dd
```
Byron: Is this container needed for the run to be successful? I thought `machete` only checks metadata?

Byron (changes requested):
This looks great, thanks for the fixes!

CI might be failing as the check job needs to learn about the new job. CI is stuck otherwise.

Could you also bring everything into a single commit? If I do it, I'd probably break your signature.

Thanks again.

PS: also putting it back to draft and you can bring it back to non-draft once you think it's ready.

.github/workflows/push.yaml
craigmayhew: Good point. I tested in another repo and it does not need the rust toolchain (which was why I unnecessarily specified the container). The bnjbvr/cargo-machete comes with the toolchain and builds cargo-machete if required.

Removed! Thank you!

Byron:
Right, it turns out that CI doesn't allow external repositories anymore unless they are trusted.

Instead, let's try this:

```patch
diff --git a/.github/workflows/push.yaml b/.github/workflows/push.yaml
index 318eeee6b9..8d74175597 100644
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -207,6 +207,22 @@ jobs:
           rustup component add clippy
           cargo clippy --workspace --all-targets -- -D warnings
 
+  cargo-machete:
+    needs: changes
+    if: ${{ needs.changes.outputs.rust == 'true' }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+      - name: Install cargo-machete
+        run: cargo install --locked cargo-machete --version 0.9.1
+      - name: Run cargo-machete on workspace
+        run: cargo machete
+
   cargo-deny:
     needs: changes
     if: ${{ needs.changes.outputs.rust == 'true' }}
@@ -517,6 +533,7 @@ jobs:
       - rust-test
       - rust-test-tauri
       - rust-lint
+      - cargo-machete
       - cargo-doc
       - build-installer
       - but-sdk-build-check
```

Byron:
It's quite amazing how the failure of that job isn't surfaced at all here, so I was left thinking it's the usual deadlock even though it's deadlock + startup failure.

If this works now, I think we can merge.

Thanks for bearing with me here, I was on the wrong boat for a while.

Byron (approved):

PR #12471, opened 2026-02-20
refactor: remove the bundled setsid binary

slarse:
## 🧢 Changes
I'm experimenting with simplifying what we ship, and I think the `gitbutler-git-setsid` binary is redundant. From what I can tell, the only reason we have it is to disconnect a Git command child process from the controlling terminal, s.t. the SSH client will invoke the askpass program specified in the `SSH_ASKPASS` environment variable. See the commit message for [66c1e91](https://github.com/gitbutlerapp/gitbutler/pull/12471/commits/66c1e91dcbe7879c94f7ed34f747f763bccf168b) for details on that.

From the ssh man pages on my machine:

```bash
SSH_ASKPASS           If ssh needs a passphrase, it will read the passphrase  from  the
                      current  terminal if it was run from a terminal.  If ssh does not
                      have a terminal associated with it but  DISPLAY  and  SSH_ASKPASS
                      are set, it will execute the program specified by SSH_ASKPASS and
                      open  an X11 window to read the passphrase.  This is particularly
                      useful when calling ssh  from  a  .xsession  or  related  script.
                      (Note  that  on some machines it may be necessary to redirect the
                      input from /dev/null to make this work.)
```

Since SSH 8.4 (released in 2020), we can instead use `SSH_ASKPASS_REQUIRE=force` to force the use of `SSH_ASKPASS` regardless of if there is a connected terminal or not.

> Note: The removal of the binary means nothing for Windows as it isn't used there (`setsid` does not exist on Windows), but the addition of `SSH_ASKPASS_REQUIRE=force` may benefit Windows 11 as it seems to require that to be set.

## ☕️ Reasoning
This simplifies our deployed artifacts.


crates/gitbutler-git/src/bin/setsid/windows.rs
```diff
@@ -1,3 +0,0 @@
-pub fn main() {
-    panic!("This binary is only meant to be run on Unix-like systems. It exists on Windows only because Cargo cannot switch off bins based on target platform.");
-}
```
slarse, line 3: Hilariously, we actually shipped this binary on Windows.

crates/gitbutler-git/src/repository.rs
```diff
@@ -193,33 +184,30 @@ where
...
+    // Windows 11 does not have a concept of a controlling terminal, but appears to require
+    // SSH_ASKPASS_REQUIRE=force to look at SSH_ASKPASS: https://github.com/PowerShell/Win32-OpenSSH/issues/2115
+    //
+    // Note that when setting SSH_ASKPASS_REQUIRE to force, we do NOT need to set DISPLAY and we do
+    // NOT need to disconnect from the controlling terminal, as is otherwise the case for SSH to
+    // consider having a peek at the SSH_ASKPASS variable.
+    //
+    // See the OpenSSH client manual for more info.
+    envs.insert("SSH_ASKPASS".into(), askpath_path.display().to_string());
+    envs.insert("SSH_ASKPASS_REQUIRE".into(), "force".into());
```
slarse, line 207: This is the key part of being able to remove the `setsid` binary. It means we rely on behavior from OpenSSH 8.4 when running the GUI, which we previously didn't. I'll see to adding a check for that somewhere so we can emit a warning if the OpenSSH version is too old.

PR #12474, opened 2026-02-20
Commit creation

Caleb-T-Owens:
Introduce a non-legacy version of `commit_and_update_refs`. Instead of the old reference frame and other complicated setup, this now takes the standard relative_to selector and an insert side.

### DIFF
--- a/crates/but-api/src/commit.rs
+++ b/crates/but-api/src/commit.rs
@@ -1,8 +1,9 @@
 use std::collections::HashSet;
 
+use anyhow::bail;
 use bstr::{BString, ByteSlice};
 use but_api_macros::but_api;
-use but_core::sync::RepoExclusive;
+use but_core::{DiffSpec, sync::RepoExclusive};
 use but_hunk_assignment::HunkAssignmentRequest;
 use but_oplog::legacy::{OperationKind, SnapshotDetails};
 use but_rebase::graph_rebase::{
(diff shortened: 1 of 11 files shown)


Byron:
@codex review

crates/but-api/src/commit.rs
```diff
@@ -158,6 +159,102 @@ pub fn commit_insert_blank(
     res
 }
 
+/// Creates and inserts a commit relative to either a commit or a reference.
+#[but_api]
+#[instrument(err(Debug))]
+pub fn commit_create_only(
+    ctx: &mut but_ctx::Context,
+    relative_to: ui::RelativeTo,
```
Byron, line 171: I now realise that we have a `ui::` type outside of legacy, even though that JSON specific types would have to be provided separately so we can have a non-UI type for the bare Rust version.
Also, I think we would not use `ui` anymore in `but-api` as all it knows is JSON, and the `but-api` turns this into whatever we need.

With that said, I think this is a 'note to self' that there needs to be a parameter mapping so in the `but-api` macro one can specify which JSON type a parameter should use.

And I am aware that this already slipped in for `insert_blank_commit` and I wonder if I missed that.

crates/but-api/src/commit.rs
```diff
@@ -158,6 +159,102 @@ pub fn commit_insert_blank(
...
+        message,
+        context_lines,
+        guard.write_permission(),
+    )
+}
+
+/// Creates and inserts a commit relative to either a commit or a reference.
+pub(crate) fn commit_create_only_impl(
+    ctx: &mut but_ctx::Context,
+    relative_to: ui::RelativeTo,
```
Byron, line 192: The docs of `RelativeTo` would need an update now. I'd also know how to extend them to be less generic if you want me to take a stab.

crates/but-api/src/commit.rs
```diff
@@ -158,6 +159,102 @@ pub fn commit_insert_blank(
...
+    let new_commit = match (rebase, commit_selector) {
+        (Some(rebase), Some(commit_selector)) => {
+            let materialized = rebase.materialize()?;
+            Some(materialized.lookup_pick(commit_selector)?)
+        }
+        (None, None) => None,
+        (Some(_), None) | (None, Some(_)) => bail!("BUG: inconsistent commit outcome"),
+    };
+
+    ws.refresh_from_head(&repo, &meta)?;
```
Byron, line 227: As a note: I hope with the next go at the frontend we can actually make use of this and just provide the new workspace to the frontend to figure out and display.

crates/but-api/src/commit.rs
```diff
@@ -158,6 +159,102 @@ pub fn commit_insert_blank(
...
+
+    Ok(json::UICommitCreateResult {
```
Byron, line 229: There would have to be a non-UI type for this, and the `but-api` macro already has a way to specify the JSON type for return value, I think it's `but_api(json::Type)`, as this is non-fallible.

crates/but-workspace/src/commit/commit_create.rs
```diff
@@ -0,0 +1,88 @@
...
+        pub rebase: Option<SuccessfulRebase>,
+        /// Selector pointing to the newly created commit, if one was created.
+        pub commit_selector: Option<Selector>,
+        /// Rejected diff specs from commit creation, matching legacy behavior.
+        pub rejected_specs: Vec<(but_core::tree::create_tree::RejectionReason, DiffSpec)>,
+        /// The intermediate tree before cherry-picking back onto the target tree.
+        pub changed_tree_pre_cherry_pick: Option<gix::ObjectId>,
+    }
+
+    /// Create a commit from `changes` and insert it relative to `relative_to` on `side`.
```
Byron, line 25: The docs are severely lacking here, but I'd be happy to write them when taking a closer look.

crates/but-workspace/src/commit/commit_create.rs
```diff
@@ -0,0 +1,88 @@
...
+            (Step::Pick(Pick { id, .. }), InsertSide::Above) => Some(id),
+            (Step::Pick(Pick { id, .. }), InsertSide::Below) => {
+                let commit = editor.find_commit(id)?;
+                commit.parents.first().copied()
+            }
+            (Step::Reference { refname }, _) => {
+                let mut reference = editor.repo().find_reference(refname.as_ref())?;
+                let target_id = reference.peel_to_id()?.detach();
+                Some(target_id)
+            }
```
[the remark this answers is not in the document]
Byron: I don't think the editor has any changes at this time, or is expected to have any. Otherwise it's a good point.
Maybe this is a chance to find a guideline, like "if you have an editor, always use the editor to resolve references"?

crates/but-workspace/tests/workspace/commit/commit_create.rs
```diff
@@ -0,0 +1,215 @@
...
+    Ok(but_core::diff::worktree_changes(repo)?
+        .changes
+        .into_iter()
+        .map(DiffSpec::from)
+        .collect())
+}
+
+#[test]
+fn commit_above_commit() -> Result<()> {
+    let (_tmp, graph, repo, mut _meta, _description) = writable_scenario("reword-three-commits", |_| {})?;
```
Byron: There are new primitives that allow to alter a scenario with Rust post creation. This is a good motivation to get it there even if it's not yet possible, as this typically makes it much clearer which changes belong to the scenario.

crates/but-workspace/tests/workspace/commit/commit_create.rs
Byron, line 1: The tests seem like they are duplicating tests from the editor and the actual commit creation. And while I believe to remember similar tests to exist for inserting an empty commit as well, even back then I didn't understand it.

So here I think it's fine to have a smoke test only that would only detect major breakage, knowing that the individual parts are well tested.

Byron (commented):
Thanks for looping me in.
Please note that I only performed an armchair review to get first notes out quickly.

While Copilot and Codex might have their own points, I also found quite a few that ideally will be addressed before merging.
However, if there is any need for rushing this, @krlvi can also approve it.

crates/but-api/src/commit.rs
Caleb-T-Owens: Can you help me understand what you're looking for here?

If we're making a type focused for frontend consumption, where would you like it to be put?

crates/but-api/src/commit.rs
Caleb-T-Owens: What do you mean here?

crates/but-workspace/tests/workspace/commit/commit_create.rs
Caleb-T-Owens: Can you point me to an example of the new primitives?

crates/but-workspace/tests/workspace/commit/commit_create.rs
Caleb-T-Owens: The rebase engine is only facilitating the history manipulation. Testing at this level is important to ensure that we've manipulated the history correctly.

crates/but-workspace/src/commit/commit_create.rs
Caleb-T-Owens: If you make a PR, please tag me in it.

What details are missing for you?

crates/but-workspace/src/commit/commit_create.rs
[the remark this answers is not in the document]
Caleb-T-Owens: I've updated the behaviour since this could be encountered in the longer term.

crates/but-workspace/src/commit/commit_create.rs
Byron: All parameters need to be documented.

crates/but-api/src/commit.rs
Byron: I mean the struct, it talks about `blank commits`:

https://github.com/gitbutlerapp/gitbutler/blob/10144aea72b9755f81297647f5e28fde643a39fd/crates/but-api/src/commit.rs#L67-L78

crates/but-api/src/commit.rs
Byron: To sum up some notes from our call: While the types consumable through the `but_api` API should not be degenerating information, if you think that this non-composable commit function is only good for the frontend and doesn't belong in "The API layer[..]" that "[..] can be used to create GitButler applications.", then we should find a new home specific to our current frontend.
There the types should be what the frontend needs, along with the API itself.

crates/but-workspace/tests/workspace/commit/commit_create.rs
Byron: I think this comment is crucially about the test pyramid which to me is leveraging knowledge about the code to test only what we have to, and at the layer that make it the easiest to put our little tentpoles down, each representing something that is important and shouldn't regress.

I argue that the implementation here shouldn't repeat anything that is known (or assumed) to be tested in the methods that the implementation is calling. Seeing these tests alone makes me think there should probably be more because they are more than a single happy path, but not enough to be exhaustive.

What it should be is to show the implementation is generally working, probably even with a note saying that the exhaustive tests are done in `but-rebase` and `commit::create()` respectively.

The alternative can only be repeat all tests of lower levels in higher levels because we don't want to trust them, which also isn't really done here, and which is something I would want to avoid as I still hope we will be reading this code to maintain it.

Byron:
I also realised that maybe this PR should also enable these functions, insert blank and create commit, in single branch mode?
After all, they all can now work. That way additional value would be enabled.

Byron (approved):
Let's get this going, but please note that I did not test it at all or tried the code.

I will try an async workflow for certain fixes that I think are needed, and I will CC you when/if I make them.

PR #12475, opened 2026-02-20
Add project drop zone 

PavelLaptev:
This issue is addressed here: https://github.com/gitbutlerapp/gitbutler/issues/4084

https://github.com/user-attachments/assets/b70da90e-ca21-4736-af37-1e3a590a43a7

This pull request introduces a new drag-and-drop project addition feature to the desktop app, allowing users to add projects by dropping a folder onto the app window. It includes a new `ProjectDropZone` Svelte component for the UI and logic, integrates this component at the app layout level, and updates the Tauri window configuration to enable custom drag-and-drop handling. There are also minor UI text improvements.

**New drag-and-drop project addition:**

* Added a new `ProjectDropZone.svelte` component that listens for drag-and-drop events at the window level, displays a visual overlay when a folder is dragged over the app, and handles adding the dropped folder as a project using the projects service. This includes custom SVG artwork and animated effects for user feedback.
* Integrated the `ProjectDropZone` component into the main app layout (`+layout.svelte`) so it is always active and ready to handle project drops. [[1]](diffhunk://#diff-86c85f1cf669f26cf476d52d5dcac60bf770423d1fddc638e9d7564de1980f43R12) [[2]](diffhunk://#diff-86c85f1cf669f26cf476d52d5dcac60bf770423d1fddc638e9d7564de1980f43R211)

**Tauri window configuration:**

* Removed `.disable_drag_drop_handler()` from the Tauri window builder to allow the app to receive drag-and-drop events, enabling the new custom drop zone functionality. [[1]](diffhunk://#diff-aafeffcc3d3842c5cf13d1d40ce61a00c9983e28d8a5f2ed39dd4d5c836a7c4eL246) [[2]](diffhunk://#diff-aafeffcc3d3842c5cf13d1d40ce61a00c9983e28d8a5f2ed39dd4d5c836a7c4eL271)

**UI/UX improvements:**

* Updated the message for the "project already exists" modal to end with a period for consistency and improved user experience.

PR #12498, opened 2026-02-22
deleteme

estib-vega:


estib-vega:
Just created this PR to verify that the e2e were not completely broken. 

There's something weird with #12475

PR #12506, opened 2026-02-22
chore(rust): bump deps and remove rustls-pemfile chain

krlvi:
- bump `rmcp` to `0.16.0` and `async-openai` to `0.33.0`
- switch `but` telemetry back to `posthog-rs`, using `gitbutlerapp/posthog-rs` pinned to commit `664c9245f79aaedcdd025b9f3c141b62bb866b8e`
- update metrics client configuration for the newer SDK API (`host(...)` instead of `api_endpoint(...)`)
- fix MCP internal handler request-param imports to match `rmcp 0.16.0`
- refresh `Cargo.lock` for dependency changes

This keeps the `rustls-pemfile` chain out of the `but` dependency graph while preserving telemetry behavior via the SDK.

### DIFF
--- a/crates/but/Cargo.toml
+++ b/crates/but/Cargo.toml
@@ -88,7 +88,7 @@ gitbutler-repo = { workspace = true, optional = true }
 
 git2.workspace = true
 schemars.workspace = true
-posthog-rs = { version = "0.3.7" }
+posthog-rs = { git = "https://github.com/gitbutlerapp/posthog-rs", rev = "664c9245f79aaedcdd025b9f3c141b62bb866b8e" }
 serde.workspace = true
 shell-words.workspace = true
 tokio = { workspace = true, features = ["rt-multi-thread", "io-std"] }
--- a/crates/but/src/utils/metrics.rs
+++ b/crates/but/src/utils/metrics.rs
@@ -364,7 +364,7 @@ fn posthog_client(app_settings: AppSettings) -> Option<impl Future<Output = post
     {
         let options = posthog_rs::ClientOptionsBuilder::default()
             .api_key(api_key.to_string())
-            .api_endpoint("https://eu.i.posthog.com/i/v0/e/".to_string())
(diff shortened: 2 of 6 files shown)


crates/but/src/utils/metrics.rs
```diff
@@ -315,28 +329,33 @@ impl BackgroundMetrics {
...
-    }
-    client.capture(posthog_event)
+    let mut properties = event.props;
+    properties.insert("$lib".to_string(), serde_json::Value::String("but".to_string()));
+    properties.insert(
+        "$lib_version".to_string(),
+        serde_json::Value::String(option_env!("VERSION").unwrap_or_default().to_string()),
+    );
+    let body = PosthogCaptureRequest {
+        api_key: client.api_key,
```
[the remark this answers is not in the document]
krlvi: Keeping this intentionally best-effort to match the previous `posthog-rs` capture behavior (transport failures only). Added an inline code comment next to the send call to document that rationale.

crates/but/Cargo.toml
```diff
@@ -88,7 +88,7 @@ gitbutler-repo = { workspace = true, optional = true }
 
 git2.workspace = true
 schemars.workspace = true
-posthog-rs = { version = "0.3.7" }
+posthog-rs = { git = "https://github.com/gitbutlerapp/posthog-rs", rev = "664c9245f79aaedcdd025b9f3c141b62bb866b8e" }
```
Byron, line 91: I think it would be good to know why it must be this particular commit. Maybe add a comment to inform about the requirements covered by the specific hash.

From the commit message, it's also not clear (to me) *why* that pemfile should not be pulled in anymore.

Byron (approved):
LGTM!

crates/but/Cargo.toml
krlvi: i hope to get this merged upstream. it's an old version

PR #12545, opened 2026-02-23
makefile to help run CI locally

schacon:
I can never remember everything, so this lets me just run `make` to
see if CI is going to pass or not.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,12 @@
+.PHONY: check fmt clippy all
+
+all: fmt fmt-check clippy
+
+fmt:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,12 @@
+.PHONY: check fmt clippy all
+
+all: fmt check clippy
+
+fmt:
+	cargo fmt --check --all
+
+check:
+	cargo check --workspace --all-targets
+
```
Byron, line 10: This can certainly be removed in favor of just `clippy`

PR #12547, opened 2026-02-23
fix: enable askpass feature for tauri app again

slarse:
## 🧢 Changes
It was accidentally disabled in pruning deps.

This feature flag will go away soon-ish so this workaround can then go away as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4172,6 +4172,7 @@ dependencies = [
  "console-subscriber",
  "document-features",
  "git2",
+ "gitbutler-git",
  "gitbutler-oplog",
  "gitbutler-project",
  "gitbutler-repo-actions",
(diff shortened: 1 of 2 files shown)


Byron:
I am always a bit saddened if the PR descriptions contain more information than the commit, as it deepens our GitHub dependence.

slarse:
I agree, allow me to cheer you up with [this commit](https://github.com/gitbutlerapp/gitbutler/pull/12471/changes/66c1e91dcbe7879c94f7ed34f747f763bccf168b).

That said, I think we should perhaps make the default merge commit message just copy the body of the PR description? Right now we copy the title, which is hardly very useful.

Byron:
> I agree, allow me to cheer you up with [this commit](https://github.com/gitbutlerapp/gitbutler/pull/12471/changes/66c1e91dcbe7879c94f7ed34f747f763bccf168b).

You know what I like :)!

> That said, I think we should perhaps make the default merge commit message just copy the body of the PR description?

Yes, that's easy to do and better than just the subject. I tend to be lazy enough to not even break lines in my commit messages when I copy markdown into them. But I just keep doing that until I am called out 🤞.

PR #12553, opened 2026-02-23
fix(but): checkout to default target as fallback

slarse:
## 🧢 Changes
Makes `but teardown` check out to the default target as a fallback if there are no active branches.

Previously it would error out like so if you just setup/teardown on a fresh repository:

```bash
$ but setup && but teardown
<snip>
Exiting GitButler mode...

→ Creating snapshot...
  ✓ Snapshot created: f8bbb6f

→ Finding active branch to check out...
Error: Failed to teardown GitButler project.

Caused by:
    No active branches found
```

## TODO
This is a naive fix for this very particular issue, just wanted to see if it'd pass CI. Also want to handle broken states (e.g. top stack has no active branches).

### DIFF
--- a/crates/but/src/command/legacy/teardown.rs
+++ b/crates/but/src/command/legacy/teardown.rs
@@ -90,16 +90,19 @@ pub(crate) fn teardown(ctx: &mut Context, out: &mut OutputChannel) -> anyhow::Re
     // Sort by order to ensure we get the leftmost (lowest order) stack first
     stacks.sort_by_key(|s| s.order.unwrap_or(usize::MAX));
 
-    let target_stack = stacks
-        .first()
-        .ok_or_else(|| anyhow::anyhow!("No active branches found"))?;
-
-    // Get the name of the top branch in the stack
-    let target_branch_name = target_stack
-        .heads
-        .first()
-        .map(|h| h.name.to_string())
-        .ok_or_else(|| anyhow::anyhow!("Stack has no branches"))?;
+    let target_branch_name = if let Some(target_stack) = stacks.first() {
+        target_stack
+            .heads
+            .first()
+            .map(|h| h.name.to_string())
+            .ok_or_else(|| anyhow::anyhow!("Stack has no branches"))?
+    } else {
+        gitbutler_stack::VirtualBranchesHandle::new(ctx.project_data_dir())
+            .get_default_target()?
+            .branch
+            .branch()
+            .to_string()
+    };
 
     if let Some(out) = out.for_human() {
         writeln!(
--- a/crates/but/tests/but/command/teardown.rs
+++ b/crates/but/tests/but/command/teardown.rs
@@ -464,3 +464,34 @@ fn json_output_with_dangling_commits() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+/// Test: Teardown checks out target branch when there is no active branch
+#[test]
+fn teardown_checks_out_target_branch_when_there_is_no_active_branch() -> anyhow::Result<()> {
+    let env = Sandbox::init_scenario_with_target_and_default_settings("repo-with-remote-and-head")?;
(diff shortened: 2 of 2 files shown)


crates/but/src/command/legacy/teardown.rs
```diff
@@ -90,16 +90,19 @@ pub(crate) fn teardown(ctx: &mut Context, out: &mut OutputChannel) -> anyhow::Re
...
-        .map(|h| h.name.to_string())
-        .ok_or_else(|| anyhow::anyhow!("Stack has no branches"))?;
+    let target_branch_name = if let Some(target_stack) = stacks.first() {
+        target_stack
+            .heads
+            .first()
+            .map(|h| h.name.to_string())
+            .ok_or_else(|| anyhow::anyhow!("Stack has no branches"))?
+    } else {
+        gitbutler_stack::VirtualBranchesHandle::new(ctx.project_data_dir())
```
Byron, line 100: Could you try using `ctx.ws.get()?`. It has all kinds of information, and while the target there is optional, it would typically be set.
Thank you

slarse: Thanks! I was looking around for alternative ways to get that information, this seemed "heavy".

While I've got you on the line, I feel like we should also have "Stack has no branches" situations fall back on the default target. Don't you think? That seems like a completely broken state to me, but all the more reason not to leave the user stuck in a broken GitButler workspace. Right now it just errors out.

Byron: The `ws` data structure should always have one stack, but it might be unnamed.
To my mind, to actually do it right we should go back to the branch that the user used before switching onto `gitbutler/workspace`. One could check the reflog for `HEAD` to figure this out.
Meantime, doing better here without being perfect seems acceptable.

Byron: What was I thinking - of course `stacks` can be empty. That's a completely acceptable state. In single-branch mode, i.e. when `main` is checked out, it won't ever be empty, but in full workspace mode, it can be.

slarse: > What was I thinking - of course stacks can be empty.

Sorry, that's not what I meant. This state is the entire reason this PR is necessary, as if you have no active branches in your workspace there are also no stacks (e.g. just after running `but setup`) and then we just fail the teardown.

What I was considering is the case where we _have_ stacks (i.e. the `if` branch above), where there is a possibility that said stack has no branches. At least in terms of code, that can happen. How that would happen in reality I'm not sure, it seems like an error state.

Somewhat simplified, the current code flow of this conditional is as follows:

* There is at least one stack?
    - Yes: There is at least one branch in that stack?
        - Yes: Check out to that branch
        - No: Error (<---- this is where I feel we should _also_ try to check out to default target instead of erroring)
    - No: There is a default target?
        - Yes: Check out to default target
        - No: Error (<----- can we end up here? Is there a case where we don't have a default target?)

Or maybe, as you say, we should just completely abandon the above and just utilize the reflog instead. Find the last branch and check out to that. But what if that branch has been deleted since switching to the gitbutler workspace? Perhaps then we should check out to the default target. So it would go something like this:

* Previously checked out branch exists?
    - Yes: Checkout to that branch
    - No: Checkout to default target

WDYT?

Byron: > No: Error (<---- this is where I feel we should also try to check out to default target instead of erroring)

I also think we should fall back here, it's then the only branch name we have. And if the code should be perfect, it would have to deal with its absence which is perfectly fine.
Git also has the notion of the default branch name, which is configured and can be queried. That one can be checked out if nothing else helps, I think.
As disclaimer: my comment is very technical and doesn't optimise for UX, it optimises for "find a reasonable branch name" 😅.

slarse:
Replaced with #13793

PR #12562, opened 2026-02-24
Update but-api to use higher level commits

Caleb-T-Owens:
Update the but-api to use higher level types. This means rust consumers don’t need to deal with lower-level types like a `HexHash`.

By making use of the `but_api(type)` paramater we can specify a JSON specific type.

---

This commit also ended up containing a unrelated change where I dropped the “changed_tree_pre_cherry_pick” since It’s not used at the minute.

---
- <kbd>&nbsp;3&nbsp;</kbd> #12570 
- <kbd>&nbsp;2&nbsp;</kbd> #12569 
- <kbd>&nbsp;1&nbsp;</kbd> #12562 👈

### DIFF
--- a/crates/but-api/src/commit.rs
+++ b/crates/but-api/src/commit.rs
@@ -3,7 +3,7 @@ use std::collections::HashSet;
 use anyhow::bail;
 use bstr::{BString, ByteSlice};
 use but_api_macros::but_api;
-use but_core::{DiffSpec, sync::RepoExclusive};
+use but_core::{DiffSpec, sync::RepoExclusive, tree::create_tree::RejectionReason};
 use but_hunk_assignment::HunkAssignmentRequest;
 use but_oplog::legacy::{OperationKind, SnapshotDetails};
 use but_rebase::graph_rebase::{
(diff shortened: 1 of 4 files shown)


krlvi (approved):
@Caleb-T-Owens  Just flagging that the api name `commit_uncommit_changes` is a bit ambiguous. 
Sounds like a toggle - an api that can commit and uncommit changes with the same api.
Maybe `commit_uncommited_changes` helps?

Byron (commented):
Yes, that's it from a `but-api` point of view! 🙏

crates/but-api/src/commit.rs
```diff
@@ -422,23 +428,23 @@ pub fn commit_uncommit_changes_only(
...
-        replaced_commits: vec![(commit_id, new_commit_id.into())],
+    Ok(MoveChangesResult {
+        replaced_commits: vec![(commit_id, new_commit_id)],
     })
 }
 
 /// Uncommits changes from a commit, with oplog and optional assign_to support
 ///
 /// If `assign_to` is provided, the newly uncommitted changes will be assigned
 /// to the specified stack.
```
Byron, line 439: With `but-api`, just like in plumbing crates, every parameter must be documented. Something to think about is `ctx` as it's always the first and I don't think it needs docs.
A viable way to not be repetitive is to link to the functions it calls, if viable.

PR #12570, opened 2026-02-24
For properly chaining create_commit we always want a SuccessfulRebase returned

Caleb-T-Owens:
Calling `rebase` here will have a negligable performance impact since all the 3wms are identity no-ops


krlvi (approved):
to be honest - i havent properly reviewed this, but i trust its fit for purpose - there is a test

PR #12571, opened 2026-02-24
fix(but): invalidate update cache on install

slarse:
As the cache includes not only the latest version at the time, but also whether the client is up-to-date with that version, installing a new version invalidates the result of the check. This could cause a confusing "Update available: <newer version> -> <older version>" to show when a new version was installed, and the cache was populated before the version now installed was even released. It goes away only when the cache is refreshed.

This happens to me pretty much daily with the nightly builds.

This PR invalidates (deletes) the cache when a new version is installed s.t. we don't have that outdated cache.

### DIFF
--- a/crates/but-db/src/cache/table/update.rs
+++ b/crates/but-db/src/cache/table/update.rs
@@ -17,6 +17,9 @@ pub(crate) const M: &[M<'static>] = &[M::up(
 );",
 )];
 
+// This table currently only has a single row representing the latest update check.
+const SINGLETON_RECORD_ID: u8 = 1;
+
 /// A utility for accessing the update checks.
 pub struct UpdateCheckHandle<'conn> {
     conn: &'conn rusqlite::Connection,
@@ -179,7 +182,7 @@ impl UpdateCheckHandleMut<'_> {
               suppressed_at, suppress_duration_hours)
              VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9)",
             rusqlite::params![
-                1, // Singleton record
+                SINGLETON_RECORD_ID,
                 result.checked_at.naive_utc(),
                 result.status.up_to_date,
                 result.status.latest_version,
@@ -192,6 +195,22 @@ impl UpdateCheckHandleMut<'_> {
         )?;
 
         sp.commit()?;
+        Ok(())
+    }
+
+    /// Deletes the update check, which is equivalent to cache invalidation.
+    ///
+    /// This operation is idempotent and is safe to perform regardless of if an update check record
+    /// exists or not.
+    pub fn delete(self) -> rusqlite::Result<()> {
+        let sp = self.sp;
+
+        sp.execute(
+            "DELETE FROM `update-check` WHERE id = ?1;",
+            rusqlite::params![SINGLETON_RECORD_ID],
+        )?;
+        sp.commit()?;
+
         Ok(())
     }
 
@@ -222,8 +241,8 @@ impl UpdateCheckHandleMut<'_> {
         sp.execute(
             "UPDATE `update-check`
              SET suppressed_at = ?1, suppress_duration_hours = ?2
-             WHERE id = 1",
-            rusqlite::params![Utc::now().naive_utc(), hours],
+             WHERE id = ?3",
+            rusqlite::params![Utc::now().naive_utc(), hours, SINGLETON_RECORD_ID],
         )?;
 
         sp.commit()?;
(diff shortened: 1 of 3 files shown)


crates/but/src/command/update.rs
```diff
@@ -161,5 +163,24 @@ fn install(out: &mut OutputChannel, target: Option<String>) -> Result<()> {
...
+        }
+    }
+
+    Ok(())
+}
+
+fn invalidate_update_cache(cache: &mut but_db::AppCacheHandle) -> anyhow::Result<()> {
+    let mut trans = cache
+        .immediate_transaction_nonblocking()?
+        .context("could not acquire transaction")?;
```
Byron: I think we should add docs to these functions to say they are only to be used when multiple ORM calls are made in a row.

crates/but-db/src/cache/table/update.rs
```diff
@@ -17,6 +17,9 @@ pub(crate) const M: &[M<'static>] = &[M::up(
 );",
 )];
 
+// This table currently only has a single row representing the latest update check.
+const SINGLETON_RECORD_ID: u8 = 1;
```
Byron, line 21: Yes, that's better.

crates/but-db/src/cache/table/update.rs
```diff
@@ -195,6 +198,19 @@ impl UpdateCheckHandleMut<'_> {
         Ok(())
     }
 
+    /// Deletes the update check (lazy way to implement cache invalidation).
```
Byron: Method documentation isn't snappy ;).

```suggestion
    /// Deletes the update check, which also acts as cache invalidation.
```

Byron:
@codex review

Byron (commented):
I think this looks good besides that transaction usage which you already have fixed locally.
Regarding that, `immediate_transaction_nonblocking` shouldn't be used unless you will gracefully handle not getting it. Otherwise blocking ones should be used (which are the ones that are nice to call).

crates/but-db/src/cache/table/update.rs
slarse: Fixed.

crates/but/src/command/update.rs
slarse: Perhaps, but even more so a document explaining how the database connection is _meant_ to be used, why that is, and all of the tradeoffs you and I discussed would be a very welcome read. What if I put together a little document that makes sense to me, link to it in the docs for these functions and you can have a read through it and see if our mental models are actually aligned on the topic?

That also expose the back-and-forth you and I have been having to a much broader audience, human and machine alike.

Byron: Yes, that sounds good. But I'd prefer it to be processed and and in `but-db/src/lib.rs` to serve as entrypoint. 

#12580 makes some improvements as well, and I think there is enough 'local' docs to understand how this works. Maybe a landing pad in `lib.rs` makes sense beyond that and I think you would be destined to write that.

Maybe there are other good places, too, but I don't like to scatter things beyond the crate.

slarse: > Maybe there are other good places, too, but I don't like to scatter things beyond the crate.

Absolutely agree, I just want kind of a holistic "how to database" piece of text that I can read through to explain what decisions have been made overall and the justification for them. As opposed to having to go through all of the docs for individual methods and piece together the bigger picture.

If that text is co-located with the code - all the better!

slarse:
@Byron I've updated it according to your comments and what we discussed, but as this is an owned crate I need an approval :)

crates/but-db/src/cache/table/update.rs
```diff
@@ -195,6 +198,22 @@ impl UpdateCheckHandleMut<'_> {
         Ok(())
     }
 
+    /// Deletes the update check, which is equivalent to cache invalidation.
+    ///
+    /// This operation is idempotent and is safe to perform regardless of if an update check record
+    /// exists or not.
```
Byron, line 204: Nice!

crates/but/src/command/update.rs
```diff
@@ -161,5 +161,17 @@ fn install(out: &mut OutputChannel, target: Option<String>) -> Result<()> {
         writeln!(writer)?;
     }
 
+    tracing::info!("Invalidating update check cache");
```
Byron: To me this isn't an `info` level trace, it's more like a `debug` or even `trace`.

Byron (approved):
Looks good to me, even with the nit I left there.

crates/but/src/command/update.rs
slarse: That was quite literally debug that accidentally was left in there. I don't think it needs to be there at all.

```suggestion
```

Byron: Then I recommend using `dbg!()` more generously. It's ugly and not quite the same, but that has its benefits, too :).

PR #12576, opened 2026-02-24
cli: wire commit command to commit_create API

krlvi:
@Caleb-T-Owens so... the semantic of the new API is every so slightly different. Take a look at the changes to the test.

The previous api (create_commit_from_worktree_changes) also carried stack_id + stack_branch_name, which scoped which branch ref gets advanced.

With commit_create, targeting a commit node can rewire incoming refs at that node, so multiple refs pointing at the same commit may move. Thats why im trying here to use RelativeTo::Reference(target_branch.reference) + InsertSide::Below: it keeps parent behavior equivalent while scoping ref movement to the selected branch reference.

Edit:

This is now based on branch-creation-2.0

### DIFF
--- a/crates/but/src/command/legacy/commit.rs
+++ b/crates/but/src/command/legacy/commit.rs
@@ -3,8 +3,7 @@ use std::{collections::BTreeMap, fmt::Write as _};
 use anyhow::{Context, Result, bail};
 use bstr::{BString, ByteSlice};
 use but_api::{
-    commit::commit_insert_blank,
-    json::HexHash,
+    commit::{commit_create, commit_insert_blank},
     legacy::{diff, repo, workspace},
 };
 use but_core::{DiffSpec, ui::TreeChange};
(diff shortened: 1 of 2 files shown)


Caleb-T-Owens (approved):
LGTM.

There is one bug fix in the rebase engine WRGT multiple empty parallel lanes that needs to be landed before we can make a release with this though.

I'm hoping to put out the fix tomorrow so it shouldn't be a blocker to merging

krlvi:
> LGTM.
> 
> There is one bug fix in the rebase engine WRGT multiple empty parallel lanes that needs to be landed before we can make a release with this though.
> 
> I'm hoping to put out the fix tomorrow so it shouldn't be a blocker to merging

@Caleb-T-Owens btw i updated the PR description - PTAL. Basically im still not 100% sure this is correct. It seems correct tho

Caleb-T-Owens:
Sorry - I assumed you were happy with the test change and skipped it.

I do agree that it looks off. It seems related to the empty-branch issue I mentioned. Let's hold off on merging and see if the  test can be reverted after my fix for the problem is resolved.

PR #12577, opened 2026-02-24
fix(resolve): Prevent submodule deletion

dkattan:
## Summary
This PR is now scoped to the submodule-preservation fix in resolve/edit mode. The recovery guidance/UI changes and overwrite diagnostics that were previously on this branch have been split out.

## What changed
- Added shared helpers in `gitbutler-workspace::submodules` to detect configured submodules, identify submodule-related paths, and remove untracked files without deleting populated submodule worktrees.
- Updated `update_uncommitted_changes_with_tree(...)` to stop using broad `remove_untracked(true)` cleanup and instead perform targeted cleanup via `remove_untracked_excluding_submodule_paths(...)`.
- Updated edit-mode enter/abort flows to preserve submodule worktrees during checkout/cleanup, keep the index in sync after forced cleanup, and surface a clearer forced-abort error when submodule/gitlink changes would be reverted.
- Added regression coverage for submodule detection, targeted cleanup, and edit-mode save/abort behavior with populated, dirty, and diverged submodule worktrees.

## Key files touched
- `crates/gitbutler-workspace/src/submodules.rs`
- `crates/gitbutler-workspace/src/branch_trees.rs`
- `crates/gitbutler-workspace/src/lib.rs`
- `crates/gitbutler-edit-mode/src/lib.rs`
- `crates/gitbutler-edit-mode/tests/edit_mode.rs`
- `crates/gitbutler-edit-mode/tests/fixtures/edit_mode.sh`
- `crates/gitbutler-workspace/Cargo.toml`
- `Cargo.lock`

## Validation run
- `cargo test -p gitbutler-edit-mode -p gitbutler-workspace`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4300,6 +4300,7 @@ dependencies = [
  "gitbutler-repo",
  "gitbutler-stack",
  "gix",
+ "tempfile",
 ]
 
 [[package]]
--- a/crates/gitbutler-edit-mode/src/lib.rs
+++ b/crates/gitbutler-edit-mode/src/lib.rs
@@ -21,7 +21,12 @@ use gitbutler_repo::{RepositoryExt as _, SignaturePurpose, signature};
 use gitbutler_stack::VirtualBranchesHandle;
 use gitbutler_workspace::{
     branch_trees::{WorkspaceState, update_uncommitted_changes_with_tree},
-    submodules::has_submodules_configured,
+    submodules::{
+        configured_submodule_paths,
+        has_submodules_configured,
+        is_submodule_related_path,
+        remove_untracked_excluding_submodule_paths,
+    },
 };
 use serde::Serialize;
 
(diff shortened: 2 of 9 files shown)


dkattan:
Addressed the Copilot review feedback in follow-up commits:

- switched git invocation to the codebase pattern: `Command::from(gix::command::prepare(gix::path::env::exe_invocation()))`
- added tracing logs for submodule repopulation outcomes (success/failure/error)
- added corresponding lockfile update for the new `tracing` dependency

Validation rerun: `cargo test -p gitbutler-edit-mode --test edit_mode` (passes).

Caleb-T-Owens:
Hi! Thanks for submitting this PR.

Please can you sign your PR's commits.

---

It's an interesting question whether we want to manually restore submodules like this. I'm not super familiar with with submodules so I'm hoping you can help me understand some of the following questions:

Does it always make sense to do a `--recursive` type update? Are there user setups where this could actually be destructive or undesirable?

What is git's behaviour after a checkout with submodules - can we mirror that exactly?

---

I'm also curious about your reasoning to add this call in the `branch_trees.rs` file. If your scope is to just target edit mode - should we have the calls specifically be in the edit mode code?

Otherwise, is this a sustainable approach to trying to better support submodules? Are there downsides to calling this after every update of HEAD?

If we do want to take this as a more wide-spread approach of maintaining submodules, could you help me understand how this fits into the current state of play with our submodule support.

Many thanks, Caleb

dkattan:
Hi Caleb - thanks again for the detailed feedback. You were looking at an earlier approach and your concerns were spot on.\n\nI've now updated the implementation and force-pushed as a single signed commit (). The current approach no longer runs  and no longer does filesystem shuffling.\n\nWhat changed:\n- scoped behavior to edit-mode checkout transitions using path-filtered checkout in \n- kept  changes because  still goes through , so this remains on the edit-mode lifecycle path\n- deduped  into  and reused it from both call-sites\n- updated PR description to match this implementation\n\nOn your sustainability questions: I agree broad recursive submodule updates after every HEAD movement are too risky/surprising. This revision avoids that and keeps the fix narrowly targeted to preventing destructive checkout behavior in the edit-mode flow covered by the regression test.\n\nRe signing: commit is now signed locally (SSH signing), but GitHub currently reports  because this environment's token can't register SSH signing keys ( scope unavailable). I’ll add the signing key in GitHub settings and refresh verification next.

dkattan:
@Caleb-T-Owens 
I've now updated the implementation and force-pushed as a single signed commit (`bad49bc2`). The current approach no longer runs `git submodule update --init --recursive`. The approach now is to prevent the submodule deletion in the first place rather than attempt to fix it whenever it happened.

What changed:
- scoped behavior to edit-mode checkout transitions using path-filtered checkout in `gitbutler-edit-mode/src/lib.rs`
- kept `branch_trees.rs` changes because `save_and_return_to_workspace` still goes through `update_uncommitted_changes_with_tree`, so this remains on the edit-mode lifecycle path
- deduped `has_submodules_configured` into `gitbutler-workspace/src/submodules.rs` and reused it from both call-sites
- updated PR description to match this implementation

On your sustainability questions: I agree broad recursive submodule updates after every HEAD movement are too risky/surprising. This revision avoids that and keeps the fix narrowly targeted to preventing destructive checkout behavior in the edit-mode flow covered by the regression test.

Human edit: Apologies for the AI slop in the responses. Should be cleaned up now. I swear I'm a human.

dkattan:
Addressed Copilot feedback on PR/body mismatch by updating the PR description to match the current implementation (path-filtered checkout behavior, guarded remove_untracked usage, shared submodule detection helper, and targeted configured-submodule sync).

Byron:
Thanks @dkattan. Could you also hit the resolve button for everything you addressed, or comment on it before resolving if it wasn't addressed? Lastly, the PR needs rebasing due to conflicts, and I am putting it back to draft now so you can put it back in 'ready for review' when it's truly ready.
CI fails too.

dkattan:
@copilot please review this PR.

Byron:
@codex review

Byron:
Closing as it seems to have stalled. There is lots of unresolved conflicts as well.

PR #12594, opened 2026-02-25
feat(but): make less default pager and add pager override

slarse:
This changes the default pager from the built-in minus pager to using `less` instead. `less` is ubiquitous on UNIX-like systems and a really good pager.

We also now respect the PAGER environment variable (which is good practice) along with the more specific BUT_PAGER variable.

The built-in pager is still used as a fallback in case `less` is not available or the user-specified pager is not working for some reason.

### DIFF
--- a/crates/but/src/utils/output_channel.rs
+++ b/crates/but/src/utils/output_channel.rs
@@ -1,9 +1,14 @@
 use std::io::{IsTerminal, Write};
 
 use crossterm::event::{self, Event, KeyCode, KeyEvent, KeyEventKind, KeyModifiers};
-use minus::ExitStrategy;
 
-use crate::{args::OutputFormat, utils::json_pretty_to_stdout};
+use crate::{
+    args::OutputFormat,
+    utils::{
+        json_pretty_to_stdout,
+        pager::{self, Pager},
+    },
+};
 
 /// Default value for a confirmation prompt.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -34,7 +39,7 @@ pub struct OutputChannel {
     /// The output to use if there is no pager.
     stdout: std::io::Stdout,
     /// Possibly a pager we are using. If `Some`, the pager itself is used for output instead of `stdout`.
-    pager: Option<minus::Pager>,
+    pager: Option<Pager>,
     /// When `Some`, JSON values written via `write_value` are captured here instead of going to stdout.
     /// Used by `--status-after` to buffer mutation JSON before combining with status JSON.
     json_buffer: Option<serde_json::Value>,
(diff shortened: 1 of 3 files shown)


crates/but/src/utils/output_channel.rs
```diff
@@ -224,19 +229,32 @@ impl std::fmt::Write for OutputChannel {
...
+                    None => {
+                        self.stdout.write_all(s.as_bytes()).or_else(|err| {
+                            if err.kind() == std::io::ErrorKind::BrokenPipe {
+                                // Ignore broken pipes and keep writing.
+                                // This allows the caller to use `?` without having to think
+                                // about ignoring errors selectively.
+                                Ok(())
+                            } else {
+                                Err(std::fmt::Error)
+                            }
```
slarse, line 255: I'm not confident in this "ignore broken pipes" behavior. It feels off to me. I'll let the behavior stay for now but if someone has some wise words on it (@Byron ??) I'd appreciate it.

Byron: Byron the wise has been summoned?
This is to provide a nice UX around `but status | head -5` and friends, without which there is always an error on stderr. Git, for instance, won't have issues there either which I suppose is a baseline.

slarse: Oh, okay, now I see the intent. I don't think this is entirely the right way to go about it, I find it odd for us to just keep on writing when the pipe is clearly not there anymore. It entails that we'll keep on doing work for which we have no way of reporting the output.

Git does not behave like this, [it will exit silently with exit code 141 when it encounters a broken pipe](https://github.com/git/git/blob/7b2bccb0d58d4f24705bf985de1f4612e4cf06e5/run-command.c#L1139-L1150). Just as a quick demonstration of that behavior, here's what happens if you pipe a command that iteratively produces output and close the pipe:

```bash
[gitbutler]$ time git log --oneline > /dev/null

real    0m0.154s
user    0m0.141s
sys     0m0.013s
[gitbutler]$ time git log --oneline | head -1
ced58d916 GitButler Workspace Commit

real    0m0.002s
user    0m0.000s
sys     0m0.003s
[gitbutler]$ git log --oneline | head -1
ced58d916 GitButler Workspace Commit
[gitbutler]$ echo "${PIPESTATUS[0]} ${PIPESTATUS[1]}"
141 0
```

Exiting feels generally more well-behaved to me.

As `but` does a lot of network calls, and we're adding in more, I think this might become important. Does this make sense to you?

Byron: Thanks for researching this, and for making me aware!

Yes, this does make sense!
It looks like the way to do this right is to instead of ignoring, just let it fail, and instead handle this case specifically here:

https://github.com/gitbutlerapp/gitbutler/blob/40147d6a24e2fe11a4fdd46a1785bbdcff84b45f/crates/but/src/main.rs#L3

In any case, I will be looking forward for you to fix this :).

slarse: Cool! Will fix it today or tomorrow, have some more urgent things first :)

PR #12604, opened 2026-02-26
CI: Unit tests should not always install playwright

estib-vega:
~Turbo had marked the installation of the playwright tests as a dependency for the unit-tests.~
~They are not needed, so remove them.~

So here's the deal gang:

The unit tests and the component tests in the UI package use playwright and require the binary installation.

This was not being cached correctly, it seems. At least for the unit-tests.
Further more, we were installing both chromium AND webkit, that sometimes led to wait times of up to 60 min.

**And webkit is not even used for the unit-tests**

### Fixes

- ~Install only chromium on CI~
- Install chromium only for unit-tests
- Install both chromium and webkit for component tests
- Cache the binaries

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -133,6 +133,17 @@ jobs:
         with:
           persist-credentials: false
       - uses: ./.github/actions/init-env-node
+      - id: get_playwright_version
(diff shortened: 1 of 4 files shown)


Byron:
Oh I so want to merge this 😅.

estib-vega:
Soooo.... Apparently the unit tests for the UI package **do** use playwright, they just don't cache the binaries at all and they install chromium & webkit

PR #12606, opened 2026-02-26
More robust auto-pruning of but-graph display from GUI

Byron:
Now it uses the lower bound to ensure interesting segments will be kept.


Byron:
#### Before

Screenshot 2026-02-26 at 10 42 56


#### After

Screenshot 2026-02-26 at 10 42 10

PR #12607, opened 2026-02-26
WIP: Handle hunk-based operations for file additions/removals

mtsgrd:
Allow hunk-based uncommit operations to work on files that were added or
removed by treating a missing "before" entry as an empty blob and None
state. This lets unified-patch computation proceed for additions,
applies hunks against an empty before state, and correctly handles the
case where removing all hunks from an added file should delete the file
from the tree. Also preserve the correct file mode when the resulting
contents match the before or after state and clean up builder
upsert/remove logic accordingly.


mtsgrd:
@Caleb-T-Owens here it is

PR #12610, opened 2026-02-26
Expose forge_provider command in but-server

mtsgrd:
Add an HTTP route for the existing but-api command forge_provider so the
command is reachable via but-server. This change plugs the
legacy::forge::forge_provider_cmd handler into the router at
"/forge_provider", making the API consistent with other forge endpoints
and enabling clients to call the provider functionality through the
server.


estib-vega (approved):
Dang, nice catch. Thanks!

PR #12612, opened 2026-02-26
fix(but): prevent panic on short file paths that are their own reverse hex

slarse:
If a file path happened to be in the space of a reverse hex (k-z), the reverse hex would be the path itself. This had multiple issues:

1. A path of just 1-2 characters caused an immediate panic
2. There was a panic due to a "self-collision" between the reverse hex and the short ID of the path due to all short IDs being added to the reverse hexes slice in order to prevent collisions with branch substrings
3. Even with 2. fixed, it could become impossible to disambiguate the reverse hex properly against e.g. a branch if the branch ID was a substring of path

It's probably possible to fix 3 as well and we can have "paths thatare their own reverse hexes" back, but I see little value in the feature and need to move on with my day. So I'm simply making the reverse hex always be a proper hash (20 bytes) so it can always be disambiguated against other IDs.

### DIFF
--- a/crates/but/src/id/mod.rs
+++ b/crates/but/src/id/mod.rs
@@ -36,28 +36,32 @@ const UNASSIGNED: &str = "zz";
 
 /// Create a CLI ID for the given staged file (if `stack_id` is `Some`) or the
 /// given unstaged file or committed file (if `stack_id` is `None`).
+///
+/// A reverse hex is always 32 bytes long.
 fn create_reverse_hex_id(
     path_bytes: &[u8],
     stack_id: Option<&StackId>,
 ) -> anyhow::Result<ChangeId> {
-    Ok(
-        if stack_id.is_none() && path_bytes.iter().all(|c| b'k' <= *c && *c <= b'z') {
-            ChangeId::from(BString::from(path_bytes))
-        } else {
-            let mut hasher = gix::hash::hasher(gix::hash::Kind::Sha1);
-            hasher.update(path_bytes);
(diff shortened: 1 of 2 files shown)


crates/but/src/id/mod.rs
slarse, line 467: These indexings _do_ cause panics if the reverse hex is too short, but given the current invariant I think this is fine. We should blow up bad if the reverse hex is shorter than 3 characters, because that's just way too short for disambiguation.

jonathantanmy2:
We really need reverse hex filenames to participate as themselves in the short ID calculation (if not, if a regular filename `foo.txt` ends up having ID `kl`, the presence of a filename `kl` would mean that `foo.txt` is unselectable by its short ID). But what we can do is to append a hash to it, which also meets your requirement of making the reverse hexes long - I've written #12613 , which also passes your test (included in that PR).

There's a corner case in which let's say `foo.txt` has a reverse hex of `klzz`, and `kl` has a reverse hex of `klzy` (after prepending itself to its reverse hex). What I want to happen is that we drop `klzy` so that we can refer to `foo.txt` by `klz` instead of needing to write `klzz`. But I think that this is really a rare corner case, and the consequence is so minor, that we can leave it alone.

slarse:
> We really need reverse hex filenames to participate as themselves in the short ID calculation (if not, if a regular filename foo.txt ends up having ID kl, the presence of a filename kl would mean that foo.txt is unselectable by its short ID).

Ah, there we go, that's the thing I missed. Closing in favor of #12613

PR #12613, opened 2026-02-26
id: lengthen IDs of reverse hex filenames

jonathantanmy2:
Alternative fix to #12612

### DIFF
--- a/crates/but-core/src/change_id.rs
+++ b/crates/but-core/src/change_id.rs
@@ -1,6 +1,9 @@
 //! A type representing ChangeIDs in commit headers.
 
-use std::{fmt, ops::Deref};
+use std::{
+    fmt,
+    ops::{Deref, DerefMut},
+};
 
 use bstr::{BStr, BString};
 use rand::Rng;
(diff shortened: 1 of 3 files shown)


jonathantanmy2:
Hi @slarse , this is an alternative fix to #12612 . Feel free to review this or incorporate any ideas here into your PR. (If we're going with this PR, I'll also need code owner review.)

slarse (approved):
This is a better way to go about it. The only thing I'm thinking is that e.g. `kl` gets a 3 character ID of `klx` only because the reverse hash collides with the very same path it's derived from. I think it'd be fine if `path = kl, id = kl` as they're both selecting the same thing, and it would look better in the terminal.

It's not _that_ hard to fix but it's also such a niche case that it might just not be worthwhile. It's a corner case that only affects paths that are 1-3 characters, and you can always just select those paths by the path itself. It'll just look a bit odd in some cases and hardly anyone will ever notice.

slarse:
@Caleb-T-Owens @Byron @krlvi Please have a look at this at earliest convenience, my repo breaks all the time because I have an incorrigible habit of adding an untracked file called `out` :s

krlvi (approved):
lgtm!

slarse:
This is what I meant by my above comment on the reverse hex colliding with its own filepath, if that filepath is short enough:
outk

I think it'd be totally valid for `out` to be both the ID and the path in this case. But at the same time it doesn't seem worth fixing, there's nothing broken here and it's an edge case of edge cases. Just wanted it documented :)

PR #12619, opened 2026-02-27
Defer DB deserialization until after TOML freshness decision

Byron:
Follow-up on https://github.com/gitbutlerapp/gitbutler/pull/12263

Previously:
This eagerly deserializes the DB snapshot before evaluating whether
TOML should overwrite it, so a malformed DB row (for example an invalid SHA
from manual edits or a prior bad write) causes an early error and prevents
the newer-valid-TOML recovery path from running. In that case reads fail
even though the synchronization rules intend newer TOML to repair DB state.

### DIFF
--- a/crates/but-meta/src/legacy/storage.rs
+++ b/crates/but-meta/src/legacy/storage.rs
@@ -94,10 +94,9 @@ fn ensure_vb_storage_in_sync(
         };
     }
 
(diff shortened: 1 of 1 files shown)


Caleb-T-Owens (approved):
Seems like a good change to me 👍

PR #12621, opened 2026-02-27
Fix workspace file path traversal by enforcing worktree boundary

krlvi:


crates/gitbutler-repo/src/commands.rs
```diff
@@ -243,26 +243,39 @@ impl RepoCommands for Context {
     }
 
     fn read_file_from_workspace(&self, probably_relative_path: &Path) -> Result<FileInfo> {
+        let workdir = self.workdir_or_fail()?;
+        let canonical_workdir = gix::path::realpath(&workdir)?;
```
Byron, line 252: I am wondering why we'd fiddle with the `probably_relative_path` so much as it should always be relative. 
Probably one could run `gix::path::normalize()` on it to remove possible attempts to do `../` and if it's not the same afterwards, someone tries to deceive you.

If we use `workdir.join()` it will possibly override the workdir if it's absolute, so we'd have to strip the workdir afterward to be sure. But it might be enough to check with is_relative() because it should always be relative.

Maybe also document what `probably_relative_path` is supposed to be - to my mind it should only ever be a path relative to the repository as it would be in the `.git/index`, and everything else the above *should* catch.

PR #12645, opened 2026-03-02
feat: add Gitea forge integration (#2904)

hhhcccbbb:
${PR_BODY}


Byron:
Thanks a lot for giving the Gitea implementation a try (even if it's only fuelled by the bounty that someone put on it).

However, with `${PR_BODY}` as PR body and an implementation that is frontend-only (even though by now we let the backend perform forge communication) there is no chance this PR could be suitable.

If you'd like to keep trying, I recommend validating an implementation plan first before submitting a PR, to to create draft PR that can take things in small steps.

PR #12652, opened 2026-03-02
but: fix a bug where the commiting renamed files only commits the "added" side

krlvi:


crates/but/src/command/legacy/commit.rs
```diff
@@ -156,6 +156,37 @@ fn generate_unified_diff(
...
+                    _ => None,
+                });
+
+            DiffSpec {
+                previous_path,
+                path: fa.path.clone(),
+                hunk_headers,
+            }
+        })
+        .collect()
```
slarse, line 187: This loop is `O(files_to_commit.len() * changes.len())`. Could this perhaps be a performance issue if you e.g. rename a close-to-root directory?

PR #12662, opened 2026-03-03
fix(but): add dummy setsid binary for old macos installer compat

slarse:
The macos installer in version <=0.19.3 validates that the gitbutler-git-setsid binary is present, so we're re-adding it as a dummy binary to ensure that old installers keep working.

### DIFF
--- a/crates/gitbutler-git/Cargo.toml
+++ b/crates/gitbutler-git/Cargo.toml
@@ -14,6 +14,11 @@ name = "gitbutler-git-askpass"
 path = "src/bin/askpass.rs"
 test = false
 
+[[bin]] # we need this binary only for backwards compatibility with old bundled CLI installers that validate that it's there
+name = "gitbutler-git-setsid"
+path = "src/bin/noop.rs"
+test = false
+
 [features]
 default = ["tokio"]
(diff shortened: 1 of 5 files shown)


Byron:
Is there a plan to phase out the dummy? Or will it be there for life now?
I also don't understand how old installers have any business with new releases.

slarse:
> Is there a plan to phase out the dummy?

Yes! We'll phase it out when the telemetry shows ~90% adoption of compatible releases. We haven't agreed on an exact number but in terms of time it should be ~14-30 days.

> I also don't understand how old installers have any business with new releases.

The installer is bundled with the app, so old releases have old installers and they need to be able to install new releases. If you get the latest installer from the website there's no issue, but if you use the one bundled with the app, there is.

The way it's currently setup for macOS specifically, we can't do any breaking changes to the release "structure" as this is verified pre-install, which is of course quite silly. So there's a very tight coupling between the installer and the release structure, and we should decrease that, or we'll run into issues where old installers can't install new releases or new installer can't install old releases.

Byron:
Aaaah, I see, thanks a lot for elaborating 🙏.

PR #12669, opened 2026-03-03
head-info for but-sdk

Caleb-T-Owens:
Added and done manual verification of the types.

The only type I’ve not managed to verify is the two instances of “GixTime”. These were null for me so I wasn’t able to verify it was right.

---

Adding this to the electron IPC thing felt a bit odd, so I’m wondering if that can be made a little bit tider. And, in general, I’m wondering if we should come up with some organisation system for it.


krlvi (approved):
thanks!!

Byron:
By the way, before this type spreads: it was made not made for the new efforts, it was made to improve the current UI and I am sure it carries debt.

If we have a clean start, I'd prefer to stick much closer to the Rust types, i.e. `but_graph::projection::Workspace` directly, and from there we should find a way to get 'enriched' versions of it or 'enrichments' which provide more information that is expensive to compute (i.e. integration checks, dependencies, etc).

This is a very foundational type but-lite, and it deserves more thought and care.

PR #12686, opened 2026-03-04
ci: add smoke test for version compatibility

slarse:
## 🧢 Changes
Adds in a smoke test that runs on publish to validate compatibility with the last mainline release. See comments in the code for details.

Edit: I decided to run the smoke tests for the `push.yaml` workflow as well, as we really don't keep good track of the nightly builds right now. [The build yesterday failed](https://github.com/gitbutlerapp/gitbutler/actions/runs/22651021691), for example (spurious BS error), and no one brought it up.

On push the smoke test isn't actually checking the pushed code, it's still checking latest nightly <-> latest release. So if it fails it cannot possibly have anything to do with the push that triggered the failure, hence the check is not required.

[Here is an example run](https://github.com/gitbutlerapp/gitbutler/actions/runs/22714105179/job/65859288555?pr=12686), more can be found in the checks below!

### DIFF
--- a/scripts/validate-version-compatibility.sh
+++ b/scripts/validate-version-compatibility.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+# Validate that two versions of the CLI are compatible with a simple smoke test.
+#
+# We make few assertions here, we rely mostly on commands exiting non-zero if something goes wrong.
+#
+# Note that every `but` command is piping to cat. This is a li'l backwards compatible hack to ensure
+# that we don't get any prompts, as the pipe is detected as not being a tty even in the very
+# earliest versions of the CLI.
+
+set -o errexit
+set -o pipefail
+
+old_version="$1"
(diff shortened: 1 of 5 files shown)


scripts/validate-version-compatibility.sh
```diff
@@ -0,0 +1,72 @@
...
+old_version="$1"
+new_version="$2"
+
+if [ -z "$old_version" ] || [ -z "$new_version" ]; then
+  echo "usage: validate-version-compatibility.sh <old_version> <new_version>"
+  exit 1
+fi
+
+function smoke_test() {
+  but setup | cat # we currently need a call to setup when moving between release and nightly, not entirely sure why
```
slarse, line 22: I have on my TODO list to investigate why this is necessary, it's something to do with one of the shared data stores.

estib-vega (approved):
Nice!

PR #12696, opened 2026-03-05
move branches, but make it *graphic*

estib-vega:
Implement the new way of moving branches by using the graph and new rebase engine.

- Add mutation primitives for disconnecting and inserting segments in the graph
- Add an API that enables moving branches using the graph (and mark the previous version as legacy)
- Add a but command for moving branches

### Usage of the but command
```
but branch move [branch-to-move] [target-branch]
# or 
but stack [brach-to-move] [target-branch]
```

### Tests
- Test at the graph operation level
- Test at the move-branch API level
- Test at the but CLI level

fixes GB-1089

### DIFF
--- a/crates/but-api/src/branch.rs
+++ b/crates/but-api/src/branch.rs
@@ -108,16 +108,39 @@ pub fn move_branch(
     subject_branch: &gix::refs::FullNameRef,
     target_branch: &gix::refs::FullNameRef,
 ) -> anyhow::Result<()> {
-    let mut meta = ctx.meta()?;
-    let (_guard, repo, mut ws, _) = ctx.workspace_mut_and_db()?;
-    let editor = ws.graph.to_editor(&repo)?;
+    let maybe_oplog_entry = but_oplog::UnmaterializedOplogSnapshot::from_details(
+        ctx,
+        SnapshotDetails::new(OperationKind::MoveBranch),
+    )
+    .ok();
 
+    let move_branch_result = move_branch_impl(ctx, subject_branch, target_branch);
+    if let Some(snapshot) = maybe_oplog_entry.filter(|_| move_branch_result.is_ok()) {
+        let mut guard = ctx.exclusive_worktree_access();
+        snapshot.commit(ctx, guard.write_permission()).ok();
+    }
+    move_branch_result
+}
+
+/// Move the branch, updating the workspace and the metadata.
+fn move_branch_impl(
+    ctx: &mut but_ctx::Context,
+    subject_branch: &gix::refs::FullNameRef,
+    target_branch: &gix::refs::FullNameRef,
+) -> anyhow::Result<()> {
+    let mut meta = ctx.meta()?;
+    let (_guard, repo, mut workspace, _) = ctx.workspace_mut_and_db()?;
+    let editor = workspace.graph.to_editor(&repo)?;
     let but_workspace::branch::move_branch::Outcome { rebase, ws_meta } =
-        but_workspace::branch::move_branch(&ws, editor, &mut meta, subject_branch, target_branch)?;
+        but_workspace::branch::move_branch(&workspace, editor, subject_branch, target_branch)?;
 
     rebase.materialize()?;
-    meta.set_workspace(&ws_meta)?;
-    ws.refresh_from_head(&repo, &meta)?;
+    if let Some((ws_meta, ref_name)) = ws_meta.zip(workspace.ref_name()) {
+        let mut md = meta.workspace(ref_name)?;
+        *md = ws_meta;
+        meta.set_workspace(&md)?;
+    }
+    workspace.refresh_from_head(&repo, &meta)?;
 
     Ok(())
 }
--- a/crates/but-api/src/branch.rs
+++ b/crates/but-api/src/branch.rs
@@ -1,7 +1,8 @@
 use but_api_macros::but_api;
-use but_core::{ui::TreeChanges, worktree::checkout::UncommitedWorktreeChanges};
+use but_core::{RefMetadata, ui::TreeChanges, worktree::checkout::UncommitedWorktreeChanges};
 use but_ctx::Context;
 use but_oplog::legacy::{OperationKind, SnapshotDetails, Trailer};
+use but_rebase::graph_rebase::GraphExt;
 use but_workspace::branch::{
     OnWorkspaceMergeConflict,
     apply::{WorkspaceMerge, WorkspaceReferenceNaming},
@@ -97,4 +98,26 @@ pub fn branch_diff(ctx: &Context, branch: String) -> anyhow::Result<TreeChanges>
     let (_guard, repo, ws, _) = ctx.workspace_and_db()?;
     let branch = repo.find_reference(&branch)?;
     but_workspace::ui::diff::changes_in_branch(&repo, &ws, branch.name())
+}
+
+/// Move a branch on top of another
+#[but_api]
+#[instrument(err(Debug))]
+pub fn move_branch(
+    ctx: &mut but_ctx::Context,
+    subject_branch: &gix::refs::FullNameRef,
+    target_branch: &gix::refs::FullNameRef,
+) -> anyhow::Result<()> {
+    let mut meta = ctx.meta()?;
+    let (_guard, repo, mut ws, _) = ctx.workspace_mut_and_db()?;
+    let editor = ws.graph.to_editor(&repo)?;
+
+    let but_workspace::branch::move_branch::Outcome { rebase, ws_meta } =
+        but_workspace::branch::move_branch(&ws, editor, &mut meta, subject_branch, target_branch)?;
+
+    rebase.materialize()?;
+    meta.set_workspace(&ws_meta)?;
+    ws.refresh_from_head(&repo, &meta)?;
+
+    Ok(())
 }
--- a/crates/but-rebase/src/graph_rebase/mutate.rs
+++ b/crates/but-rebase/src/graph_rebase/mutate.rs
@@ -242,6 +242,9 @@ impl Editor {
     /// `target` delimiter's child and parent can be the same node.
     /// This is the way to disconnect a single node.
     ///
+    /// All disconnected children will be reconnected to all the disconnected parents unless
+    /// the `skip_reconnect_step` is set to true.
+    ///
     /// Returns an error when:
     /// - `parents_to_disconnect` is `SelectorSet::None`
     /// - `parents_to_disconnect` contains any parent that is not a direct parent of `target.parent`
@@ -251,6 +254,7 @@ impl Editor {
         target: SegmentDelimiter<C, P>,
         children_to_disconnect: SelectorSet,
         parents_to_disconnect: SelectorSet,
+        skip_reconnect_step: bool,
     ) -> Result<()>
     where
         C: ToSelector,
@@ -363,7 +367,12 @@ impl Editor {
                 .as_ref()
                 .is_none_or(|ids| ids.contains(&edge_source));
             if should_disconnect {
-                self.reconnect_edges_to_parents(&disconnected_parent_edges, edge_id, edge_source);
+                // Remove the child edge.
+                self.graph.remove_edge(edge_id);
+                // Reconnect the child node to all the disconnected parents.
+                if !skip_reconnect_step {
+                    self.reconnect_edges_to_parents(&disconnected_parent_edges, edge_source);
+                }
             }
         }
 
@@ -374,11 +383,8 @@ impl Editor {
     fn reconnect_edges_to_parents(
         &mut self,
         disconnected_parent_edges: &[(Edge, petgraph::prelude::NodeIndex)],
-        child_edge_id: petgraph::prelude::EdgeIndex,
         child_node: petgraph::prelude::NodeIndex,
     ) {
-        // Remove the child edge.
-        self.graph.remove_edge(child_edge_id);
         // Reconnect the child node to all the disconnected parents.
         for (parent_edge_weight, edge_target) in disconnected_parent_edges {
             self.graph
--- a/crates/but-workspace/src/branch/move_branch.rs
+++ b/crates/but-workspace/src/branch/move_branch.rs
@@ -9,11 +9,12 @@ use but_rebase::graph_rebase::{
 ///
 /// Returned by [function::move_branch()].
 #[derive(Debug)]
-pub struct Outcome<WorkspaceMeta> {
+pub struct Outcome {
     /// A successful rebase result for continuing operations.
     pub rebase: SuccessfulRebase,
     /// The updated workspace metadata that accompanies the move operation.
-    pub ws_meta: WorkspaceMeta,
+    /// It should replace the actual workspace metadata to configure moved 'virtual' branches segments, if `Some()`.
+    pub ws_meta: Option<but_core::ref_metadata::Workspace>,
 }
 
 pub(super) mod function {
@@ -23,13 +24,11 @@ pub(super) mod function {
     use super::Outcome;
     use anyhow::Context;
     use anyhow::bail;
-    use but_core::RefMetadata;
-    use but_core::ref_metadata::Workspace;
     use but_graph::projection::WorkspaceKind;
     use but_rebase::graph_rebase::Editor;
     use gix::refs::FullNameRef;
 
-    /// Move a branch between stacks in the workspace.
+    /// Move a branch between stacks in the `workspace`.
     ///
     /// `subject_branch_name` is the full reference name of the branch to move.
     ///
@@ -38,13 +37,12 @@ pub(super) mod function {
     ///
     /// Returns:
     /// - Successful rebase
-    pub fn move_branch<M: RefMetadata>(
+    pub fn move_branch(
         workspace: &but_graph::projection::Workspace,
         mut editor: Editor,
-        meta: &mut M,
         subject_branch_name: &FullNameRef,
         target_branch_name: &FullNameRef,
-    ) -> anyhow::Result<Outcome<M::Handle<Workspace>>> {
+    ) -> anyhow::Result<Outcome> {
         let Some(source) = workspace.find_segment_and_stack_by_refname(subject_branch_name) else {
             bail!(
                 "Couldn't find branch to move in workspace with reference name: {subject_branch_name}"
@@ -62,17 +60,20 @@ pub(super) mod function {
             bail!("Couldn't find workspace head.")
         };
 
-        let workspace_ref = match &workspace.kind {
-            WorkspaceKind::Managed { ref_info }
-            | WorkspaceKind::ManagedMissingWorkspaceCommit { ref_info } => {
-                ref_info.ref_name.clone()
+        // We're currently stopping the move branch operations imperatively at this stage, in order to
+        // reduce the scope of this first iteration of moving the branches.
+        // TODO: Enable and test that we can move branches in any kind of workspace.
+        match &workspace.kind {
+            WorkspaceKind::Managed { .. } => {}
+            WorkspaceKind::ManagedMissingWorkspaceCommit { .. } => {
+                bail!("Moving branches currently need a workspace commit")
             }
             WorkspaceKind::AdHoc => {
                 bail!("Moving branches in non-managed workspaces is not supported");
             }
         };
 
-        let mut ws_meta = meta.workspace(workspace_ref.as_ref())?;
+        let mut ws_meta = workspace.metadata.clone();
 
         let (source_stack, subject_segment) = source;
         let (_, target_segment) = destination;
@@ -107,15 +108,14 @@ pub(super) mod function {
 
         // Update the workspace meta if any of the branches we're handling is empty.
         // This is needed in order to disambiguate the intended operation.
-        let ws_meta = if subject_segment.commits.is_empty() || target_segment.commits.is_empty() {
+        if let Some(ws_meta) = ws_meta.as_mut()
+            && (subject_segment.commits.is_empty() || target_segment.commits.is_empty())
+        {
             ws_meta.remove_segment(subject_branch_name);
             ws_meta.insert_new_segment_above_anchor_if_not_present(
                 subject_branch_name,
                 target_branch_name,
             );
-            ws_meta
-        } else {
-            ws_meta
         };
 
         Ok(Outcome {
@@ -178,29 +178,25 @@ fn get_disconnect_parameters(
         .base_segment_id
         .map(|segment_idx| &workspace.graph[segment_idx]);
 
-    let parents_to_disconnect = match (stack_base_segment, graph_base_segment) {
-        (Some(stack_base_segment), _) => {
-            // Base segment is part of the source stack.
-            let base_segment_ref_name = stack_base_segment
-                .ref_name()
-                .context("Base segment doesn't have a ref name.")?;
-            let reference_selector = editor.select_reference(base_segment_ref_name)?;
-            let selectors = SomeSelectors::new(vec![reference_selector])?;
-            SelectorSet::Some(selectors)
-        }
-        (None, Some(graph_base_segment)) => {
-            // Base segment is outside of workspace (probably target branch).
-            let ref_name = graph_base_segment
-                .ref_name()
-                .context("Graph base segment doesn't have a ref name.")?;
-            let reference_selector = editor.select_reference(ref_name)?;
-            let selectors = SomeSelectors::new(vec![reference_selector])?;
-            SelectorSet::Some(selectors)
-        }
-        (None, None) => {
-            // No parents could be determined. Detach all.
-            SelectorSet::All
-        }
+    let parents_to_disconnect = if let Some(stack_base_segment) = stack_base_segment {
+        // Base segment is part of the source stack.
+        let base_segment_ref_name = stack_base_segment
+            .ref_name()
+            .context("Base segment doesn't have a ref name.")?;
+        let reference_selector = editor.select_reference(base_segment_ref_name)?;
+        let selectors = SomeSelectors::new(vec![reference_selector])?;
+        SelectorSet::Some(selectors)
+    } else if let Some(graph_base_segment) = graph_base_segment {
+        // Base segment is outside of workspace (probably target branch).
+        let ref_name = graph_base_segment
+            .ref_name()
+            .context("Graph base segment doesn't have a ref name.")?;
+        let reference_selector = editor.select_reference(ref_name)?;
+        let selectors = SomeSelectors::new(vec![reference_selector])?;
+        SelectorSet::Some(selectors)
+    } else {
+        // No parents could be determined. Detach all.
+        SelectorSet::All
     };
 
     if index_of_segment == 0 {
--- a/crates/but-workspace/src/branch/move_branch.rs
+++ b/crates/but-workspace/src/branch/move_branch.rs
@@ -1,0 +1,239 @@
+use anyhow::Context;
+use but_graph::projection::{Stack, StackSegment};
+use but_rebase::graph_rebase::{
+    Editor, Selector, SuccessfulRebase,
+    mutate::{SegmentDelimiter, SelectorSet, SomeSelectors},
+};
+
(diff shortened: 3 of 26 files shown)


crates/but/src/command/branch/move_branch.rs
```diff
@@ -0,0 +1,98 @@
+use anyhow::{Context, bail};
+use but_rebase::graph_rebase::GraphExt;
+
+use crate::{CliId, IdMap, id::parser::parse_sources, utils::OutputChannel};
+
+/// Apply a branch to the workspace, and return the full ref name to it.
```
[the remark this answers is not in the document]
estib-vega: Yep

crates/but/src/args/branch.rs
```diff
@@ -119,6 +119,20 @@ pub enum Subcommands {
...
+    ///
+    /// Just specify the branch to move (or take out of a stack) and insert it into another stack,
+    /// on top of another branch.
+    Move {
+        /// The name or CLI ID of the branch to move
+        #[clap(value_name = "BRANCH")]
+        branch: String,
+        // The name or CLI ID of the branch to move the branch on top of
+        #[clap(value_name = "TARGET_BRANCH")]
+        target_branch: String,
```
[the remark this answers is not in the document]
estib-vega: Yep

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -0,0 +1,212 @@
+use anyhow::Context;
+use but_graph::projection::{Stack, StackSegment};
+use but_rebase::graph_rebase::{
+    Editor, Selector, SuccessfulRebase,
+    mutate::{SegmentDelimiter, SelectorSet, SomeSelectors},
+};
+
+/// Outcome of moving branches betweent stacks.
```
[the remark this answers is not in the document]
estib-vega: Yep

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -0,0 +1,212 @@
...
+            bail!("Couldn't find workspace head.")
+        };
+
+        let (source_stack, subject_segment) = source;
+        let (_, target_segment) = destination;
+        let target_segement_ref_name = target_segment
+            .ref_name()
+            .context("Target segment doesn't have a ref")?;
+        let target_selector = editor
+            .select_reference(target_segement_ref_name)
```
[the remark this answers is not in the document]
estib-vega: Yep

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -0,0 +1,212 @@
...
+                None
+            } else {
+                Some(index)
+            }
+        })
+        .context("BUG: Unable to find subject segment on source stack.")?;
+
+    let subject_segment_ref_name = subject_segment
+        .ref_name()
+        .context("Subject segement doesn't have a ref name.")?;
```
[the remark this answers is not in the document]
estib-vega: yep

crates/but-graph/src/api.rs
```diff
@@ -572,6 +572,15 @@ impl Graph {
         self.inner.node_indices()
     }
 
+    /// Look for a segment in the graph by it's index.
```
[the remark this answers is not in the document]
estib-vega: yep

crates/but/src/lib.rs
```diff
@@ -320,14 +320,62 @@ async fn match_subcommand(
...
-                        .emit_metrics(metrics_ctx)
-                } else {
+            let mut ctx = setup::init_ctx(
+                &args,
+                InitCtxOptions {
+                    background_sync: BackgroundSync::Enabled,
+                    ..Default::default()
+                },
+                out,
+            )?;
```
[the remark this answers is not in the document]
estib-vega: Yeeeeeap

crates/but/tests/but/command/branch/move_branch.rs
```diff
@@ -0,0 +1,394 @@
...
+    env.but("branch move A C")
+        .assert()
+        .success()
+        .stdout_eq(str![[r#"
+Moved branch 'A' on top of 'C'.
+
+"#]])
+        .stderr_eq(str![""]);
+
+    // Check that the operation suceeded.
```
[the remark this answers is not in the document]
estib-vega: yep

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    /// Move a branch between stacks in the workspace.
+    ///
+    /// `subject_branch_name` is the full reference name of the branch to move.
+    ///
+    /// `target_branch_name` is the full reference name of the branch to move the subject
+    /// branch on top of.
+    ///
+    /// Returns:
+    /// - Succesful rebase
```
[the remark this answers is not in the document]
estib-vega: yep

crates/but-workspace/tests/fixtures/scenario/ws-ref-ws-commit-single-stack-double-stack.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
+
+source "${BASH_SOURCE[0]%/*}/shared.sh"
+
+### General Description
+
+# A ws-ref points to a workspace commit, with two stacks inside:
+# - One with a single commit with a single commit
```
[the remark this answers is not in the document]
estib-vega: yep

crates/but-rebase/src/graph_rebase/mutate.rs
```diff
@@ -242,6 +242,9 @@ impl Editor {
     /// `target` delimiter's child and parent can be the same node.
     /// This is the way to disconnect a single node.
     ///
+    /// All disconnected children will be reconnect to all the disconnected parents unless
```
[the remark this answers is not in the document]
estib-vega: Yep

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -0,0 +1,212 @@
...
+    mutate::{SegmentDelimiter, SelectorSet, SomeSelectors},
+};
+
+/// Outcome of moving branches between stacks.
+///
+/// Returned by [function::move_branch()].
+pub struct Outcome {
+    /// A successful rebase result for continuing operations.
+    pub rebase: SuccessfulRebase,
+}
```
Byron, line 18: Is there a plan to add more fields so `Outcome` is needed in anticipation?
If not, I'd remove it and return `SuccessfulRebase` directly.

Otherwise, I'd document why this was chosen, or what's missing.

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -0,0 +1,212 @@
...
+    ///
+    /// Returns:
+    /// - Successful rebase
+    pub fn move_branch(
+        workspace: &but_graph::projection::Workspace,
+        mut editor: Editor,
+        subject_branch_name: &FullNameRef,
+        target_branch_name: &FullNameRef,
+    ) -> anyhow::Result<Outcome> {
+        let Some(source) = workspace.find_segment_and_stack_by_refname(subject_branch_name) else {
```
Byron, line 46: Seeing this, and remembering [the conversation](https://discord.com/channels/1060193121130000425/1479061638186401925) about future head-infos, it seems that in future we might just use SegmentIDs to identify any segment the user may pick, anonymous or not.
I also think this can be tackled once we really want to move (or manipulate) anonymous segments more for improved tool support.

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -0,0 +1,212 @@
...
+    workspace_head: gix::ObjectId,
+) -> anyhow::Result<(
+    SegmentDelimiter<Selector, Selector>,
+    SelectorSet,
+    SelectorSet,
+)> {
+    let index_of_segment = source_stack
+        .segments
+        .iter()
+        .enumerate()
```
Byron: I think this is `iter().position()`.

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -0,0 +1,212 @@
...
+        let selectors = SomeSelectors::new(vec![workspace_head_selector])?;
+        let children_to_disconnect = SelectorSet::Some(selectors);
+
+        return Ok((delimiter, children_to_disconnect, parents_to_disconnect));
+    }
+
+    // Segment on top of the subject segment in the stack.
+    let child_segment = source_stack.segments.get(index_of_segment - 1).context(
+        "BUG: Unable to find child segment of subject segment but expected it to exist.",
+    )?;
```
Byron, line 216: I am pretty sure I have this backward, and `child` just means it's the descendant segment, or the stack segment that depends on the segment that the user wants to move.

I guess I am writing this in the hopes that we will eventually align on language which I don't think I did in the but-graph. Parent/child seems like a good choice. The graph talks about bottom and top or above and below a lot to match how we display things in the UI and how `git log` is working. 
Child and parent is universal, so it seems preferable, and maybe worth fixing in `but-graph` as well.

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -0,0 +1,212 @@
...
+            .segments
+            .iter()
+            .find(|segment| segment.id == base_segment_id)
+    });
+
+    let graph_base_segment = subject_segment
+        .base_segment_id
+        .and_then(|segment_idx| workspace.graph.find_segment(segment_idx));
+
+    let parents_to_disconnect = match (stack_base_segment, graph_base_segment) {
```
Byron: This wants to be `stack_base_segment.or(graph_base_segment)` to avoid the duplication below.

crates/but-workspace/tests/workspace/branch/move_branch.rs
Byron, line 1: Loving the tests!

But I also think it's worth to at least document in a test that handling reordering of virtual branches (i.e. refs on top of the workspace base) (probably) doesn't work, and basically most interactions with empty segments.
Or maybe some of these already would work, I can't tell.

crates/but-api/src/branch.rs
```diff
@@ -70,3 +71,28 @@ pub fn branch_diff(ctx: &Context, branch: String) -> anyhow::Result<TreeChanges>
...
     but_workspace::ui::diff::changes_in_branch(&repo, &ws, reference.name())
 }
+
+/// Move a branch on top of another
+#[but_api]
+#[instrument(err(Debug))]
+pub fn move_branch(
+    ctx: &mut but_ctx::Context,
+    subject_branch: String,
+    target_branch: String,
```
Byron: This is still legacy as long as it uses Stringy types.
I know the function above does the same, but that just slipped past me :D.

The UI should learn to send unambiguous full name refs, and for that we should have macro based type conversions.

Byron (commented):
I am loving how working with the new workspace and rebase engine can yield new mutations quickly!

With that said, I think we should have some tests to show what happens with purely virtual branches, just because I think that with `meta` this probably won't yield expected results. It's fine, but maybe we can just limit the participating branches to those that aren't empty.

Lastly, I set this PR back to draft as I think the stringy types in `but-api` must be fixed, so we can avoid falling back to former patterns of an API that is dominated by its JSON nature.

Finally, let me state that I spent only a little time with this, and didn't review anything outside of the api and the workspace, nor did I look at how this mutation is actually done or its logic. I did skim the tests to see what it can do though, and found them very easy to follow 🩷.

crates/but-workspace/src/branch/move_branch.rs
estib-vega: Will fix 👍

crates/but-workspace/src/branch/move_branch.rs
estib-vega: Yeah, I like child-parent terminology.

We had a talk once in the office, in which we were debating which was the best terminology for this (in the scope of documenting the order of commits and branches for review APIs).

They can't be time-based terminology (older-newer) since one can amend at any point in the future.
They can't be spacial (top-bottom) since that's an UI implementation detail. Specially for review, one might want to see the commits in the order of application (parent to child) and not in the conventional git log order (child to parent).

Child-parent terminology seems more absolute, past space-time continuity.
Which is basically also the plot of Interstellar.

crates/but-workspace/tests/workspace/branch/move_branch.rs
estib-vega: Truuue.

I don't think they will work as expected yet.
Trying to figure out the way to get them to work (through the metadata).

crates/but-api/src/branch.rs
estib-vega: I'll add that in a separate PR.
Will tag you to align on this

Byron: Meantime I have added the ability to take `gix::refs::FullName` or `&gix::refs::FullNameRef`, so doing so should be no problem if the UI is sending full branch names as well.

crates/but-workspace/src/branch/move_branch.rs
estib-vega: They'r different types, so I won't let me do that :/

`Segment` and `StackSegment`

crates/but-api/src/branch.rs
estib-vega: Updated the function signature

crates/but-workspace/tests/workspace/branch/move_branch.rs
estib-vega: Now they work, and they're tested

crates/but-workspace/src/branch/move_branch.rs
estib-vega: Fixed

crates/but-workspace/src/branch/move_branch.rs
estib-vega: Now we return the updated workspace metadata as well, so keeping this struct

crates/but/src/command/legacy/branch/mod.rs
```diff
@@ -15,182 +13,184 @@ mod json;
...
+    let stacks = but_api::legacy::workspace::stacks(
+        ctx,
+        Some(but_workspace::legacy::StacksFilter::InWorkspace),
+    )?;
+
+    // Find which stack this branch belongs to
+    for stack_entry in &stacks {
+        if stack_entry.heads.iter().all(|b| b.name != *branch_name) {
+            // Not found in this stack,
+            continue;
```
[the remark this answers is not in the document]
estib-vega: False

crates/but/src/command/branch/move_branch.rs
```diff
@@ -0,0 +1,101 @@
+use anyhow::{Context, bail};
+use but_core::RefMetadata;
+use but_rebase::graph_rebase::GraphExt;
+
+use crate::{CliId, IdMap, id::parser::parse_sources, utils::OutputChannel};
+
+/// Move a branchon top of another
+pub fn move_branch(
```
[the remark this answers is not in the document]
estib-vega: Will fix

crates/but/src/command/branch/move_branch.rs
```diff
@@ -0,0 +1,101 @@
+use anyhow::{Context, bail};
+use but_core::RefMetadata;
+use but_rebase::graph_rebase::GraphExt;
```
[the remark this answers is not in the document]
estib-vega: false

crates/but/src/args/mod.rs
```diff
@@ -345,6 +345,17 @@ pub enum Subcommands {
...
 
+    /// Move branches on top of others to stack them.
+    #[clap(verbatim_doc_comment)]
+    Stack {
+        /// The branch being moved.
+        #[clap(value_name = "BRANCH")]
+        branch: String,
+        /// The branch to move the other on top off.
+        #[clap(value_name = "TARGET_BRANCH")]
+        target_branch: String,
```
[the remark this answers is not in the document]
estib-vega: Will fix

crates/but-api/src/branch.rs
```diff
@@ -1,7 +1,8 @@
 use but_api_macros::but_api;
-use but_core::{ui::TreeChanges, worktree::checkout::UncommitedWorktreeChanges};
+use but_core::{RefMetadata, ui::TreeChanges, worktree::checkout::UncommitedWorktreeChanges};
 use but_ctx::Context;
 use but_oplog::legacy::{OperationKind, SnapshotDetails, Trailer};
+use but_rebase::graph_rebase::GraphExt;
 use but_workspace::branch::{
```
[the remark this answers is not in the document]
estib-vega: False

crates/but/src/lib.rs
```diff
@@ -322,18 +322,111 @@ async fn match_subcommand(
...
+            branch,
+            target_branch,
+        } => {
+            let ctx = but_ctx::Context::discover(&args.current_dir)?;
+            command::branch::move_branch(ctx, &branch, &target_branch, out)
         }
+        Subcommands::Branch(branch::Platform { cmd }) => match cmd {
+            #[cfg(not(feature = "legacy"))]
+            None => todo!("implement list and call recursively"),
+            #[cfg(feature = "legacy")]
```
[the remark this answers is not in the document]
estib-vega: All, good. This was so before

crates/but-workspace/src/branch/move_branch.rs
Byron: Alright, I will keep that in mind and see if I can make improvements to parts of the API in `but-graph` when I get a chance.
I could probably tell Codex to do it, but I don't feel strong enough right now 😅.

Byron (commented):
Thanks a lot!

I have pushed a refactor commit which straightens out the use of `RefMetadata` and further limits the move-branch functions to only work if it's a traditional managed commits. For everything else one would definitely need tests.

However, it's probably fair to kick that can hard and say that once `meta` is backed by the DB officially, and `but-rebase` alters it, `move-branch` will naturally work in these cases. So no need to put time into it beyond what 95% of the users need right now.

I did not look at:

- tests
- integration into but-api or but-rebase
- basically anything except for `but-workspace`, and even there I only dealt with typing/superficial things.

Codex found two issues, one of which is related to a missing edit-mode check. The other one is about possibly missing metrics, both seem worth another look.

---

[P1] Dispatch legacy branch moves through the guarded path

Repro:
1. Build/run the default `but` binary (`legacy` is enabled by default in `crates/but/Cargo.toml`).
2. In a repo with applied GitButler stacks, run `but setup`.
3. Leave open-workspace mode, e.g. `git checkout main`.
4. Run `but branch move A C`.

Expected:
This should fail through the legacy guarded path, the same way other legacy branch mutations do, because `gitbutler_branch_actions::move_branch` enforces `ensure_open_workspace_mode(...)` and records a `MoveBranch` snapshot for undo.

Actual:
`but branch move` now dispatches directly to `command::branch::move_branch` from `crates/but/src/lib.rs`, so it bypasses the legacy wrapper. In edit/outside-workspace mode it can still rewrite refs/workspace metadata, and the move is no longer recorded in the oplog.

---

[P2] Re-wrap branch dispatch in metrics emission

Repro:
1. Compare the old and new `Subcommands::Branch(...)` dispatch in `crates/but/src/lib.rs`.
2. Notice that `match_subcommand()` still creates `metrics_ctx`.
3. Notice that the new `Subcommands::Stack { .. }` arm and the refactored `Subcommands::Branch(...)` arm now return raw `Result<()>` values instead of calling `.emit_metrics(metrics_ctx)`.

Expected:
`but branch ...` and `but stack ...` should emit the same duration/error telemetry as other CLI commands, including the new `CommandName::BranchMove` mapping added in this patch.

Actual:
No metrics emission happens for these paths anymore, so branch commands silently stop reporting telemetry. That means the newly added `BranchMove` metric will never fire.

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -9,11 +9,12 @@ use but_rebase::graph_rebase::{
 ///
 /// Returned by [function::move_branch()].
 #[derive(Debug)]
-pub struct Outcome<WorkspaceMeta> {
```
Byron: This is a strong smell, and I noticed that LLMs have a hard time with `RefMetadata` when I tried just now. Basically everything it comes up with is probably not what you want.

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -9,11 +9,12 @@ use but_rebase::graph_rebase::{
...
+pub struct Outcome {
     /// A successful rebase result for continuing operations.
     pub rebase: SuccessfulRebase,
     /// The updated workspace metadata that accompanies the move operation.
-    pub ws_meta: WorkspaceMeta,
+    /// It should replace the actual workspace metadata to configure moved 'virtual' branches segments, if `Some()`.
+    pub ws_meta: Option<but_core::ref_metadata::Workspace>,
```
Byron, line 17: It's still quite undefined what should happen without metadata, but it seems fair to just ignore it as there is nothing we can do for now (unless one day there is a case for changing this, but I doubt it).

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -38,13 +37,12 @@ pub(super) mod function {
     ///
     /// Returns:
     /// - Successful rebase
-    pub fn move_branch<M: RefMetadata>(
+    pub fn move_branch(
         workspace: &but_graph::projection::Workspace,
         mut editor: Editor,
-        meta: &mut M,
```
Byron: The key here is that `but-graph` already collects metadata, and thus it should be queried from the `Workspace`.
This makes handling the metadata update here quite straightforward, but forces the caller to set it if present. Fair enough.

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -62,17 +60,17 @@ pub(super) mod function {
...
         };
 
-        let workspace_ref = match &workspace.kind {
-            WorkspaceKind::Managed { ref_info }
-            | WorkspaceKind::ManagedMissingWorkspaceCommit { ref_info } => {
-                ref_info.ref_name.clone()
+        match &workspace.kind {
+            WorkspaceKind::Managed { .. } => {}
+            WorkspaceKind::ManagedMissingWorkspaceCommit { .. } => {
+                bail!("Moving branches currently need a workspace commit")
```
Byron, line 69: Let's be even more conservative here, and opt it in with a barrage of tests.
Maybe that won't be needed if the rebase engine takes care of this, so postponing complexity seems like the way to go.

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -0,0 +1,232 @@
...
+            }
+            WorkspaceKind::AdHoc => {
+                bail!("Moving branches in non-managed workspaces is not supported");
+            }
+        };
+
+        let mut ws_meta = workspace.metadata.clone();
+
+        let (source_stack, subject_segment) = source;
+        let (_, target_segment) = destination;
```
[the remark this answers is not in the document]
Byron: Yeah, this probably means it should be set to `None` when it's not changed.

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -178,29 +175,25 @@ fn get_disconnect_parameters(
         .base_segment_id
         .map(|segment_idx| &workspace.graph[segment_idx]);
 
-    let parents_to_disconnect = match (stack_base_segment, graph_base_segment) {
```
Byron: I replaced this as it wasn't a `match` to me. Please double-check, I was lazy and asked my bot to do it 😅.

Byron (commented):
And some notes on the ominous `refactor` commit. I made it because I need to test what I *feel* is right IRL to know it's right.
Also, I get that it might work better for you with comments, so here they are, and it should be fine to drop the commit and redo it based on the gist.

Byron (approved):
There are still a couple of legit-looking LLM comments to look into, but besides that I think it's good to go.
Again with the note that I didn't actually try using it.

crates/but-workspace/src/branch/move_branch.rs
estib-vega: I don't like this default assumption that LLMs came up with my changes.

This I did on my own. Would you mind explaining why this is a smell, so that I can avoid that next time?

crates/but-workspace/src/branch/move_branch.rs
estib-vega: Gotcha. I guess this is purely stylistic, though.

Happy to keep talking about this offline some other time.

crates/but-workspace/tests/workspace/branch/move_branch.rs
```diff
@@ -0,0 +1,490 @@
...
+    ws: &but_graph::projection::Workspace,
+    ws_meta: Option<but_core::ref_metadata::Workspace>,
+) -> anyhow::Result<()> {
+    if let Some((ws_meta, ref_name)) = ws_meta.zip(ws.ref_name()) {
+        let mut md = meta.workspace(ref_name)?;
+        *md = ws_meta;
+        meta.set_workspace(&md)?;
+    }
+    Ok(())
+}
```
estib-vega, line 490: @Byron Curious, why is this necessary?

crates/but-workspace/src/branch/move_branch.rs
Byron: > I don't like this default assumption that LLMs came up with my changes.

Yeah, I will avoid mentioning it. Judging from myself, all I get to do is reviews and refactors these days. And since I am a late-bloomer, I just assume everyone else uses it *more* :D.

> This I did on my own. Would you mind explaining why this is a smell, so that I can avoid that next time?

It makes no sense to use a generic type for a struct for which only one implementation exists. But it doesn't look that way when looking at this `Handle` type and I really hope that it can be simplified. Something along these lines should be possible once it's finally in the database, at which point there isn't even a need for `RefMetadata` as trait anymore.

So it's a deal, the trait will go away.

crates/but-workspace/src/branch/move_branch.rs
Byron: I do think `if … else` is less ambiguous here and I'd prefer it unless exhaustiveness is needed. Using `match` won't kill me though, it just confuses me.

crates/but-workspace/src/branch/move_branch.rs
estib-vega: Yeah, the motivation here was to maintain dynamism with the correct types. I assumed that if we're dealing with the implementation of a trait, then it's feasible to get different workspace handles as well. 

Didn't know there was only the one

crates/but-workspace/tests/workspace/branch/move_branch.rs
Byron: Yeah, right!

`Git + metadata = graph -> workspace`. Then `move` copies the workspace metadata, and alters it. For refreshes of the workspace it wants `meta` to re-read reference metadata. So `meta` has to get the new metadata which is associated with a reference. 
It is implicit knowledge that `ws.ref_name()` is the `gitbutler/workspace` typically, and that's where we want to set the changed metadata in the metadata store (vb.toml + db mirror right now).

```rust
 *md = ws_meta;
```

This is the interesting line as one can't set the data directly. This is me predicting the future, thinking it will be easier when it's an actual database.

But probably it's just helping with the setup now, and in future all this will go away and be a very dumb assignment.

crates/but-workspace/src/branch/move_branch.rs
estib-vega: Makes sense to me. If else is easier to read 👍

crates/but-workspace/tests/workspace/branch/move_branch.rs
estib-vega: Gotcha. Thanks!

estib-vega:
Addressed the two open issues found by Codex.

1. No metrics are emitted in the branch and stack commands
🤦 now we do.

2. No more oplog entry
Now we do. We snapshot and store if everything went well.

TODO: Do we need to check for open workspace even if we're currently checking for being in a managed workspace?
Not sure. I didn't copy over the `ensure_open_workspace` call since that's still in the legacy corner.

PR #12716, opened 2026-03-07
feat(cli): allow entering edit mode from CLI

foxfirecodes:
this PR was written by Claude Code with assistance and review from myself

## 🧢 Changes

this PR introduces a new `edit-mode` subcommand that lets you enter edit mode via the CLI. this is meant to mirror the current functionality in the GUI where you can right click on a commit and click "Edit commit" to enter edit mode.

## ☕️ Reasoning

while i do enjoy using the GUI, i would like to be able to allow my AI agents to use edit mode directly, and sometimes its just easier to use a CLI. seeing as the functionality was all there and simply missing a dedicated command to interact with it, i took the opportunity to implement this.

the reason why i named this `edit-mode` instead of `edit` is to avoid conflicts with the existing hidden `edit` subcommand

the structure of both the code & the output is meant to align with what already exists in the resolve mode command. i did not extract shared code between the resolve and edit-mode commands but if you want me to do that i can do so, most of the code is basically the same. that being said the actual amount of code involved is pretty small to begin with.

there is a small frontend change here as well to solve what seemed to me to be a race condition. if you exit edit mode (or resolve mode) from the CLI, sometimes the app would show errors like this:

errors

it seems that the app would sometimes receive the `worktree_changes` event before it had unmounted the `EditMode` UI due to the mode changing, resulting in the error message above. im open to suggestions on a better solution, but this seems like a reasonable approach for a simple race condition like this

### DIFF
--- a/apps/desktop/src/lib/mode/modeService.ts
+++ b/apps/desktop/src/lib/mode/modeService.ts
@@ -1,3 +1,4 @@
+import { parseError } from "$lib/error/parser";
 import { hasBackendExtra } from "$lib/state/backendQuery";
 import { invalidatesList, providesList, ReduxTag } from "$lib/state/tags";
 import { InjectionToken } from "@gitbutler/core/context";
@@ -140,8 +141,17 @@ function injectEndpoints(api: BackendApi) {
 						`project://${arg.projectId}/worktree_changes`,
 						async (_) => {
 							if (finished) return;
-							const changes = await invoke<TreeChange[]>("edit_changes_from_initial", arg);
-							lifecycleApi.updateCachedData(() => changes);
+							try {
+								const changes = await invoke<TreeChange[]>("edit_changes_from_initial", arg);
+								lifecycleApi.updateCachedData(() => changes);
+							} catch (error: unknown) {
+								// Edit mode may have been exited (e.g. via CLI) before this
+								// listener was unsubscribed. Silently ignore that specific race.
+								const { message } = parseError(error);
+								if (!message.includes("Expected to be in edit mode")) {
+									throw error;
+								}
+							}
 						},
 					);
 					// The `cacheEntryRemoved` promise resolves when the result is removed
--- a/crates/but/skill/references/reference.md
+++ b/crates/but/skill/references/reference.md
@@ -396,6 +396,50 @@ but resolve cancel
 
 **Important:** Never use `git add`, `git commit`, or other git write commands during conflict resolution. Only use `but resolve` commands and edit files directly.
 
+## Edit Mode
+
+Edit any commit in your stack by checking it out, making changes, and saving.
+
+### `but edit-mode <commit>`
+
+Enter edit mode for a commit.
+
+```bash
+but edit-mode <commit-id>
+```
+
+### `but edit-mode status`
+
+Show files changed during the edit session.
+
+```bash
+but edit-mode status
+```
+
+### `but edit-mode finish`
+
+Save changes and return to workspace mode.
+
+```bash
+but edit-mode finish
+```
+
+### `but edit-mode cancel`
+
+Cancel edit and return to workspace mode.
+
+```bash
+but edit-mode cancel
+but edit-mode cancel --force  # discard changes
+```
+
+**Workflow:**
+
+1. `but edit-mode <commit>` - Enter edit mode
+2. Make your changes to files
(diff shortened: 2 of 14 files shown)


Byron:

Could you squash commits to two maybe, one for the CLI changes, and one for the frontend fix?
Then it would probably also be good to get feedback from @krlvi on the CLI API side of things. It feels like we want to find a more succinct name for all the new interaction.

@codex review

foxfirecodes:
squashed and updated the frontend to use `parseError` to be more resilient! also cleaned up some logic in the status command. im pretty happy w where this is at, i do agree that there may be a more succinct name we could use for this interaction than edit-mode. i would use `but edit` if it wasnt already a thing

Byron:
Thanks a lot!

Then I'd think that @krlvi would be able to review this PR and/or give product-related directions.

foxfirecodes:
hey there! bumping for vis, would it be possible to get a review on this? thanks!

Byron:
> hey there! bumping for vis, would it be possible to get a review on this? thanks!

I am not sure this is a serious request as there are a lot of unresolved auto-review comments, and there are conflicts with the branch to merge into.

It would also be good to differentiate this PR from `but resolve`.

----

What `but` already has is a conflict-resolution workflow that uses `gitbutler-edit-mode` under the hood:

- `but resolve <commit-id>` enters edit mode for a conflicted commit, per the CLI docs in [crates/but/src/args/mod.rs](/Users/byron/dev/github.com/gitbutlerapp/gitbutler/crates/but/src/args/mod.rs#L795) and the dispatcher in [crates/but/src/lib.rs](/Users/byron/dev/github.com/gitbutlerapp/gitbutler/crates/but/src/lib.rs#L1241).
- `but resolve status`, `but resolve finish`, and `but resolve cancel` are the interaction points once in that mode, implemented in [crates/but/src/command/legacy/resolve.rs](/Users/byron/dev/github.com/gitbutlerapp/gitbutler/crates/but/src/command/legacy/resolve.rs#L24).
- That command path calls the edit-mode API wrappers in [crates/but-api/src/legacy/modes.rs](/Users/byron/dev/github.com/gitbutlerapp/gitbutler/crates/but-api/src/legacy/modes.rs#L57), which in turn call `gitbutler_edit_mode::commands::*`.

Two important caveats:

- It is scoped to conflicted-commit resolution, not a generic “enter edit mode for any commit” CLI.
- `but edit` is unrelated; it is just a hidden built-in text editor for files in [crates/but/src/args/mod.rs](/Users/byron/dev/github.com/gitbutlerapp/gitbutler/crates/but/src/args/mod.rs#L201) and [crates/but/src/lib.rs](/Users/byron/dev/github.com/gitbutlerapp/gitbutler/crates/but/src/lib.rs#L760).

So the accurate answer is: yes, `but` already interacts with edit mode, but only indirectly via `but resolve`, not via a dedicated generic edit-mode subcommand.

foxfirecodes:
> I am not sure this is a serious request as there are a lot of unresolved auto-review comments, and there are conflicts with the branch to merge into.

@Byron this is in fact a serious proposal, it felt like a missed opportunity for the GUI to have a way to enter edit mode *outside* of the conflict resolution flow but not have an equivalent in the CLI.

especially because at the time of creating this i had some ideas of integrating edit mode into AI workflows, where it would be nice for an AI to be able to use the CLI to:
- enter edit mode on a commit
- make changes
- exit edit mode
- check if the change introduced conflicts
- automatically do the whole conflict resolution flow for any conflicting commits

the unresolved conflicts & auto review comments are because i rebased my workspace to work on other contributions to this repo and never ended up cleaning up this PR bc it sounds unlikely that this is a feature yall would want to include in the CLI.

if not then feel free to close and no point in me addressing the new conflicts & auto review comments. otherwise id be more than happy to get this PR back up to date with the latest changes to see this ship!

in hindsight it probably would have been better to open a feature request issue first to see if this was something the team even wanted in the CLI, and that is the route ill take in the future! it was just easy enough to whip up a working prototype that i figured id go ahead and do so, worst case it gets shot down

Byron:
Alright!

Since it's easy now to whip up anything, let's get back to the issue stage, the preferred contribution workflow, so this can be fleshed out. Some time has past, a lot has changed, it's probably better that way.

Thanks again!

PR #12718, opened 2026-03-07
Update to latest Anthropic Models

vidluther:
## 🧢 Changes
This pull request updates the available Anthropic Claude AI models in the desktop app to reflect the latest active models, removes deprecated ones, and updates both their display names and internal enum values.

**Anthropic Claude model updates:**

* Updated the `AnthropicModelName` enum in `types.ts` to only include active models, removing deprecated ones and adding the latest versions: `Haiku 4.5`, `Sonnet 4.6`, `Opus 4.6`, and `Opus 4.5`.
* Updated the model selection options in `AiSettings.svelte` to match the new enum, including new display labels and removing old model options.
## ☕️ Reasoning
Older models are giving errors: 
image


Byron:

@vidluther What do you think about switching to a model-only specification that doesn't include the version at all? This should auto-select the latest version, which is usually what people want.

Thank you

vidluther:
> Thanks a lot for contributing!
> 
> @vidluther What do you think about switching to a model-only specification that doesn't include the version at all? This should auto-select the latest version, which is usually what people want.
> 
> Thank you

@Byron you mean Opus, Sonnet, Haiku - always the latest? as listed here? 

https://platform.claude.com/docs/en/about-claude/models/overview

I think that's a good idea. I was going to do something similar with OpenAI

https://developers.openai.com/api/docs/models

Currently we're offering these.. 
```
export enum OpenAIModelName {
	O3mini = "o3-mini",
	O1mini = "o1-mini",
	GPT5 = "gpt-5",
	GPT5Mini = "gpt-5-mini",
	GPT4_1 = "gpt-4.1",
	GPT4_1Mini = "gpt-4.1-mini",
	GPT4oMini = "gpt-4o-mini",
}
```

Byron:
> @Byron you mean Opus, Sonnet, Haiku - always the latest? as listed here?

Yes, that's what I had in mind.

Please feel free to do the same for OpenAI in a separate commit along with an updated PR title.

Please also be sure to add checkboxes for which of these model names you have tested yourself.
Thanks again.

PR #12730, opened 2026-03-09
Add flake

OliverJAsh:
Initially this will only be used by me and @davidpdrsn, but over time I hope more people will adopt Nix. We could also use this in CI.


OliverJAsh:
Superseded by https://github.com/gitbutlerapp/gitbutler/pull/12731

PR #12731, opened 2026-03-09
add flake.nix

davidpdrsn:
Adds a basic flake with a dev shell including all require packages, at least all the ones codex could find. I've been using that today and seems to be working.


OliverJAsh (approved):
Works for me! Thanks.

PR #12732, opened 2026-03-09
Respect CARGO_TARGET_DIR in tests

davidpdrsn:
## 🧢 Changes

Makes `DRIVER` in gitbutler-testsupport not blindly assume that the built binary ends up in `../../target`. It’ll first check if `CARGO_TARGET_DIR` is set and if so, use that.

## ☕️ Reasoning

I like setting `CARGO_TARGET_DIR` to share build artifacts between crates. I find this especially useful with git worktrees because you don’t need to rebuild everything from scratch in the worktree dir.

### DIFF
--- a/crates/gitbutler-testsupport/src/lib.rs
+++ b/crates/gitbutler-testsupport/src/lib.rs
@@ -317,7 +317,10 @@ fn set_storage_path_for_testing(git_dir: &Path) -> anyhow::Result<()> {
     Ok(())
 }
 
-use std::path::{Path, PathBuf};
+use std::{
+    ffi::OsStr,
+    path::{Path, PathBuf},
+};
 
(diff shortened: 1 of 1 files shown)


Byron:
Also, we want to get rid of this testing style. Maybe an LLM can just refactor this to use `but-testsupport`?

davidpdrsn:
Sure I can try and look into that! However, do you mind that I merge this as is just to unblock myself? Then I can have an agent cook on a refactor while I work on other things.

For some reason codex is extremely slow for me today so need to get access to some opus.

Byron:
> Sure I can try and look into that! However, do you mind that I merge this as is just to unblock myself?

Totally, that sounds like the most reasonable approach as well. Just wanted you to know that this test-style is legacy within the legacy :D.

You can merge this PR once the commits are signed.

PR #12735, opened 2026-03-09
Provide commit mappings in all endpoints

Caleb-T-Owens:
Read commit descriptions for more details

### DIFF
--- a/crates/but-api/src/commit.rs
+++ b/crates/but-api/src/commit.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::collections::{BTreeMap, HashSet};
 
 use bstr::{BString, ByteSlice};
 use but_api_macros::but_api;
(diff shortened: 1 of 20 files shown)


crates/but-api/src/commit.rs
```diff
@@ -19,24 +19,42 @@ pub struct CommitCreateResult {
...
+    pub replaced_commits: BTreeMap<gix::ObjectId, gix::ObjectId>,
+}
+
+/// Outcome after inserting a blank commit.
+pub struct CommitInsertBlankResult {
+    /// The ID of the newly inserted blank commit.
+    pub new_commit: gix::ObjectId,
+    /// Commits that were replaced by this operation. Maps `old_id → new_id`.
+    pub replaced_commits: BTreeMap<gix::ObjectId, gix::ObjectId>,
 }
```
krlvi, line 46: why not use one type if they are identical?

krlvi (approved):

crates/but-api/src/commit.rs
Caleb-T-Owens: The overhead of having two types instead of one is negligible, and it provides room for them to diverge easier in the future.

crates/but-api/src/json.rs
```diff
@@ -467,6 +473,13 @@ pub struct UICommitCreateResult {
         but_core::tree::create_tree::RejectionReason,
         but_serde::BStringForFrontend,
     )>,
+    /// Commits that have been replaced as a side-effect of the create/amend.
+    /// Maps `oldId → newId`.
+    #[cfg_attr(
+        feature = "export-schema",
+        schemars(with = "std::collections::BTreeMap<String, String>")
+    )]
+    pub replaced_commits: std::collections::BTreeMap<HexHash, HexHash>,
```
Byron, line 482: I just now realised that an unwanted pattern is creeping in. We shouldn't put return types from 'lower' levels here, they go into `json` modules of their respective level to be co-located.
Will submit a PR to fix that.

PR #12740, opened 2026-03-09
Reduce boxing when building router

davidpdrsn:
Reduces boxing when converting a command handler into an axum route by structing the generics a bit differently.

Also cleans up some boilerplate by going from `post(json_response(handler))` to `but_post(handler)`. The code had a comment about generating a `Handler` implementation directly but I think that'll require `#[but_api]` to generate that which I’m not sure we want. I think this is a good middleground.

---
- <kbd>&nbsp;2&nbsp;</kbd> #12740 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #12739

### DIFF
--- a/crates/but-server/src/lib.rs
+++ b/crates/but-server/src/lib.rs
@@ -1,4 +1,4 @@
-use std::{future::Future, net::SocketAddr, sync::Arc};
+use std::{convert::Infallible, future::Future, net::SocketAddr, sync::Arc};
 
 use axum::{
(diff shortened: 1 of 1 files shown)


Byron (approved):
Just a couple of helpers, and 150+ SLOC removed, nice!

I thought it was interesting that the handlers are `Copy`, and only in hindsight it's obvious as we have actual functions here 🤦‍♂️.

PR #12745, opened 2026-03-10
Make clear how the text-entry will appear in the Claude chat view

Byron:
I am sure this is an untypical case, as the user must:

- have setup claude in GitButler
- not have run `claude` ever in the project before

Then the previous message was quite 'poetic', but not helpful in figuring out how to start a chat.

Now it says exactly what's expected by the code.

### DIFF
--- a/apps/desktop/src/components/codegen/CodegenMessages.svelte
+++ b/apps/desktop/src/components/codegen/CodegenMessages.svelte
@@ -110,6 +110,8 @@
 	const claudeSettings = $derived($settingsService?.claude);
 
 	const claudeAvailable = $derived(claudeCodeService.checkAvailable(undefined));
+	// const canEnterChat = $derived(claudeAvailable.status === "available" && !!projectRegistered);
+	const canEnterChat = $derived(!!projectRegistered);
 
 	let clearContextModal = $state<Modal>();
 	let modelContextMenu = $state<ContextMenu>();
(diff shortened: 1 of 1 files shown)


Caleb-T-Owens:
Screenshot 2026-03-10 at 11 58 59

@Byron What case are you trying to cover here exactly?

This message primarily shows up when you've got claude set up, but you just haven't asked it any questions yet. Here in my editor, the expectation is that I just send a message in our  UI and claude will start doing work.

Byron:
Now that I have recorded the video, I think it's clear what it does.

https://github.com/user-attachments/assets/731fe83d-9bd7-47f9-beac-a46313e7e5a7

However, it's also clear that the poetic version has to return when a session is available. I saw it's related to `~/.claude.json` containing an entry for the project directory.

Let's see if an agent can do it.

Byron:
Ok, now it works.

https://github.com/user-attachments/assets/99d9b7cb-4033-42f7-a0b5-ac3de4fa628a

Caleb-T-Owens:
Hmm... is there some way with the Claude SDK that we can just allow the project via GitButler. 

If you're interested in this problem we really ought to just fix the flow properly ourselves.

Byron:
I think this is an immediate improvement to the problem so I don't see a reason to not merge it.
Improvements to make this state unnecessary could be a follow-up, if it's deemed valuable enough.

Byron:
Merging as it works (with minimal changes) and it's an improvement.

PR #12752, opened 2026-03-10
Lite POC

OliverJAsh:
[GB-1116: Lite POC](https://linear.app/gitbutler/issue/GB-1116/lite-poc)

Depends on https://github.com/gitbutlerapp/gitbutler/pull/12763


.vscode/extensions.json
```diff
@@ -3,6 +3,10 @@
 		"dbaeumer.vscode-eslint",
 		"svelte.svelte-vscode",
 		"rust-lang.rust-analyzer",
-		"esbenp.prettier-vscode"
+		"esbenp.prettier-vscode",
+		"oxc.oxc-vscode",
+		"vitest.explorer",
+		"ms-playwright.playwright",
+		"jnoortheen.nix-ide"
```
Caleb-T-Owens: I'm wondering if all of these make sense to be recommendations.

In the past we've kept this list to be the minimal set of configuration to work well with the project.

The nix, vitest, and playwright plugins are more preference specific than requirements to work on the project.

OliverJAsh: I don't have strong opinions on this. I've removed the change for now and we can come back to it later.

PR #12755, opened 2026-03-10
feat(cli): show diff in editor for `commit` and `reword`

davidpdrsn:
Makes `but commit` an `but reword` show the diff in the editor. I think its generally useful to refer to changes while writing the commit message. Vim will also autocomplete words from the current buffer so having the diff inline effectively seeds that.

Adds `--diff` and `--no-diff` flags. If no flags are passed it'll show the diff unless its large (so if you accidentally commit a log file you don't stall your system). If `--diff` is passed it'll always show the diff regardless of the size. If `--no-diff` is passed it'll never show the diff. The two flags are mutually exclusive.

Looks like this in my editor:

image

In builtin editor:

image

I wanna try this in a few other editors but should be good to review.

### DIFF
--- a/crates/but/src/command/legacy/mod.rs
+++ b/crates/but/src/command/legacy/mod.rs
@@ -1,6 +1,8 @@
 //! Commands in need of being ported to the non-legacy world.
 //! Doing so means that no legacy-APIs are used.
 
+use anyhow::Result;
+
 pub mod absorb;
 pub mod actions;
 pub mod ai;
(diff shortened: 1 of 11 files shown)


crates/but/src/command/legacy/mod.rs
```diff
@@ -27,3 +29,113 @@ pub mod status;
...
+                state,
+                ..
+            } => {
+                let a = blob_size(&repo, &previous_state.id);
+                let b = blob_size(&repo, &state.id);
+                a.max(b)
+            }
+        })
+        .fold(0, |a, b| a.saturating_add(b)))
+}
```
davidpdrsn, line 116: I'm not sure that this is the best way to do it, but it's what Opus came up with.

krlvi (approved):
Nice!

crates/but/src/command/legacy/mod.rs
```diff
@@ -27,3 +29,113 @@ pub mod status;
...
+///
+/// 900KB is very roughly 15,000 lines at ~60 bytes per line. Just to protect the user from
+/// stalling their system if they accidentally commit a big log file.
+pub(crate) const MAX_DIFF_BLOB_SIZE_FOR_EDITOR_IF_UNSPECIFIED: u64 = 900_000;
+
+/// Sum the blob sizes involved in the given tree changes using cheap object header lookups.
+/// For modifications/renames, uses the larger of the two sides as an upper bound.
+pub(crate) fn estimate_diff_blob_size(
+    changes: &[but_core::TreeChange],
+    ctx: &mut but_ctx::Context,
```
Byron: It's strange this gets to be `&mut` without Rust complaining that it's not mutably used here.

crates/but/src/command/legacy/mod.rs
```diff
@@ -27,3 +29,113 @@ pub mod status;
...
+/// when the user hasn't specified a preference. This uses object header lookups
+/// which are cheap compared to actually computing diffs.
+///
+/// 900KB is very roughly 15,000 lines at ~60 bytes per line. Just to protect the user from
+/// stalling their system if they accidentally commit a big log file.
+pub(crate) const MAX_DIFF_BLOB_SIZE_FOR_EDITOR_IF_UNSPECIFIED: u64 = 900_000;
+
+/// Sum the blob sizes involved in the given tree changes using cheap object header lookups.
+/// For modifications/renames, uses the larger of the two sides as an upper bound.
+pub(crate) fn estimate_diff_blob_size(
```
Byron, line 83: While is think this estimation is suitable, I wonder if it should rather be done on a per-tree-change basis. Right now, one bigger file added would hide all smaller diffs which otherwise are fine to show.

Byron (commented):
Exciting feature! *(I only looked at that one function I commented on)*

crates/but/src/command/legacy/mod.rs
davidpdrsn: Good catch, changed to a read-only ref.

Afaik rust only only complains about unused `mut` not unused `&mut`

image

I'm sure there are good reasons for that 😅

crates/but/src/command/legacy/mod.rs
davidpdrsn: Yeah thats a good point. I'll note that in Linear as potential future improvements.

PR #12768, opened 2026-03-10
napi: Apply branch

estib-vega:
Export the apply branch API in the sdk

- Enable the safe parsing of FullNameRefs for Napi bindings


packages/but-sdk/src/generated/index.d.ts
```diff
@@ -752,6 +765,15 @@ export type ForgeUser = {
   details: GitlabAccountIdentifier;
 };
 
+/** The full name of a Git reference. */
+export type FullRefName = {
+  /**
+   * The full name, like `refs/heads/main` or `refs/remotes/origin/foo`.
+   * Note that it might be degenerated if it can't be represented in Unicode.
+   */
+  full: string;
```
Byron, line 774: Would this be a good time to also add `full_bytes` to the mix? This way, Lite can be made to use bytes from the beginning when passing the name back, preventing a probably rare class of bugs where the degeneration kicks in.

estib-vega: Maybe, it just would need some adaptions both in the list branches api and in the existing client (updating the expected type there)

PR #12788, opened 2026-03-11
Replace `#[allow(...)]` with `#[expect(...)]`

davidpdrsn:
`expect` works like `allow` except it fails if the lint isn't actually required. So helps to catch things like unnecessary `#[allow(dead_code)]`.

Done with search-n-replace and having an agent fixing `cargo clippy`.


Byron (approved):
Thanks a lot!

It would be great to have a clippy lint to enforce `expect()` over allow as well.

PR #12790, opened 2026-03-11
Increase GitHub check runs page size to 100

mtsgrd:
Looks like for the gitbutler repo 30 is not enough. Ideally it would..
be configurable, but until then we need to set it to max.


mtsgrd:
@estib-vega I figured this out by logging the check results and not finding the e2e thing in there(!)

estib-vega:
> @estib-vega I figured this out by logging the check results and not finding the e2e thing in there(!)

What e2e thing?

PR #12793, opened 2026-03-11
but-rebase: fix duplicated tree hash in octopus merge error message

### LINKED ISSUE #12791: Octopus merge error shows the same tree hash twice instead of both sides
The error message in `octopus()` (`crates/but-rebase/src/merge.rs`) prints the same tree ID twice when a 2-parent merge commit conflicts, producing:

```
Encountered conflict when merging tree 8c869f1 and tree 8c869f1
```

This makes it look like the system is merging a tree with itself, which makes debugging difficult.

**Root cause**

In the `else` branch (when `successfully_merged.len() == 1`, i.e. the common 2-parent merge case):

```rust
format!(" and tree {tree_to_merge}")
```

This reuses `tree_to_merge` — the same variable already interpolated in the outer format string. The result is `tree_to_merge` printed twice. It should reference `successfully_merged[0]` (the "ours" tree) instead, so the message shows both distinct tree IDs.

**Context**

PR #12069 improved the analogous error message in `graph_rebase/cherry_pick.rs` (`format_base_merge_error`), but this older code path in `merge.rs` was not updated. The legacy `octopus()` path is still used by `upstream_integration_statuses`, `save_edit_and_return_to_workspace`, `integrate_upstream_commits`, and `create_virtual_branch_from_branch`.

This affects at least #12692, #9836, #10034, #9756, #10637, #10512 — all show the duplicated hash in their error output.

sdrshn-nmbr:
The `octopus()` error message in `merge.rs` prints `tree_to_merge` twice when a 2-parent merge conflicts. The `else` branch (for `successfully_merged.len() == 1`) formats `" and tree {tree_to_merge}"`, duplicating the hash already in the outer string. Changed to reference `successfully_merged[0]` so the message shows both the ours and theirs tree IDs.

Before:
```
Encountered conflict when merging tree 8c869f1 and tree 8c869f1
```

After:
```
Encountered conflict when merging tree 8c869f1 and tree a3b72c4
```

Fixes #12791

### DIFF
--- a/crates/but-rebase/src/merge.rs
+++ b/crates/but-rebase/src/merge.rs
@@ -78,7 +78,7 @@ pub fn octopus(
                             .join(", ")
                     )
                 } else {
-                    format!(" and tree {tree_to_merge}")
+                    format!(" and tree {}", successfully_merged[0])
                 }
             )
             .context(ConflictErrorContext {


Byron:
Thanks a lot!
Unfortunately I have to close this in favor of #12792 which is more complete.

PR #12794, opened 2026-03-12
gitbutler-edit-mode: use new rebase engine to exit edit mode

jonathantanmy2:
This is now ready for review. There is no new crate, and no ability to switch between the old and new rebase engines - I'll let @Caleb-T-Owens comment on whether this is OK (and I need his review anyway due to the code owner review thing).


jonathantanmy2:
Also, we discussed that maybe we should start a new crate with this code - should I make that change now? If yes, I'll do that in this PR.

crates/gitbutler-edit-mode/src/lib.rs
```diff
@@ -285,24 +283,21 @@ pub(crate) fn save_and_return_to_workspace(ctx: &Context, perm: &mut RepoExclusi
...
-    let mut rebase = but_rebase::Rebase::new(&gix_repo, Some(merge_base), None)?;
-    rebase.rebase_noops(false);
-    rebase.steps(steps)?;
-    let output = rebase.rebase()?;
-
-    stack.set_heads_from_rebase_output(ctx, output.references)?;
+    let workspace_commit = gix_repo
+        .find_reference(WORKSPACE_BRANCH_REF)?
+        .peel_to_commit()?;
+    let meta = ctx.legacy_meta()?;
```
Byron: This should just be `meta()` - we want to avoid bringing in anything with `legacy` in the name.

Byron (commented):
I also think we have no feature toggle for using the new rebase engine or not, and the current feature flags don't show anything that would make me think otherwise.

Personally, I'd also prefer avoiding new crates unless they are needed to prevent dependency cycles, or to capture a unique set of added dependencies that shouldn't 'pollute' an existing crate.

Caleb-T-Owens:
Hi! We've so far been switching on the frontend https://github.com/gitbutlerapp/gitbutler/blob/master/apps/desktop/src/lib/stacks/stackService.svelte.ts#L543-L549

crates/gitbutler-edit-mode/src/lib.rs
```diff
@@ -285,24 +283,21 @@ pub(crate) fn save_and_return_to_workspace(ctx: &Context, perm: &mut RepoExclusi
...
+        workspace_commit.id(),
+        Some(gix::refs::FullName::try_from(WORKSPACE_BRANCH_REF)?),
+        &meta,
+        but_graph::init::Options::limited(),
+    )?;
+    let mut editor = graph.to_editor(gix_repo)?;
+    let (target_selector, _commit) = editor.find_selectable_commit(gix_commit.id)?;
+    editor.replace(target_selector, Step::new_pick(new_commit_oid.to_gix()))?;
+    let outcome = editor.rebase()?;
+    outcome.materialize()?;
```
Caleb-T-Owens: `.materialize()` currently updates all the references and also updates the uncommitted changes.

We should be able to remove most of the code after the `.materialize()` call. The aim of the `.materialize()` call is to update all the references and check out the new `HEAD` by also cherry-picking the uncommitted changes onto that new `HEAD`.

One thing to consider is that `.materialize()` calls `safe_checkout_from_head` when it updates the worktree. In this case, that might not be the right fit, so it may be worth considering some other kind of `.materialize()` that correctly handles the currently stashed uncommitted changes.

The `safe_checkout_from_head` that `materialize` currently calls uses a `safe_checkout` funciton, which allows you to pass a `current_head` along with the `new_head`, so it might be possible to add an option or third argument that also specifies using some other uncommitted changes tree.

Byron: Indeed, `safe_checkout` definitely isn't made to deal with this situation.
My feeling is that edit-mode should leave the worktree clean because it knows what it wanted to accomplish, and from there the new checkout or adjustment can happen.

For now I am tempted to say "just make it work" (without affecting `safe_checkout` because edit-mode as it is needs an overhaul (it can't put HEAD out of the commit-graph).

Since I don't know the motivation behind this PR, I can't say more. A `but-edit`, however, would set `HEAD` to a commit, and if it's conflicted, recreate that conflict in the index. Then the user  commits that, and the new commit is correctly integrated into the workspace.

crates/gitbutler-edit-mode/src/lib.rs
jonathantanmy2: Ah thanks. Switched to that.

crates/gitbutler-edit-mode/src/lib.rs
jonathantanmy2: > We should be able to remove most of the code after the `.materialize()` call. The aim of the `.materialize()` call is to update all the references and check out the new `HEAD` by also cherry-picking the uncommitted changes onto that new `HEAD`.

My understanding is that when `materialize()` checks out the new `HEAD`, this is meant to handle the case where `HEAD` (or more precisely, the branch that it points to) has changed as a result of the rebase. The behavior when exiting edit mode is, I think, substantially different from that. When exiting edit mode, `HEAD` is `refs/heads/gitbutler/edit`, and that ref doesn't change at all (a new commit with the ref's commit's parents is created, but that new commit has no relation to `refs/heads/gitbutler/edit` at all). The rebase engine's entry point in my PR is not `refs/heads/gitbutler/edit`, but `refs/heads/gitbutler/workspace`, so the rebase engine doesn't touch `refs/heads/gitbutler/edit` either.

As for cherry-picking the uncommitted changes onto `HEAD`, I think this is substantially different too.

| Category | Normal rebase engine use | Exiting edit mode |
| --- | --- | --- |
| Source of uncommitted changes | Git index | Tree in object DB |
| Nature of destination | Git index | Amendment of commit |
| Identity of destination | Git index (sometimes described as "on top of `HEAD`") | Commit somewhere in commit graph |

For now, I'm using `materialize_without_checkout()` with a code comment about why we're not using `materialize()`.

> One thing to consider is that `.materialize()` calls `safe_checkout_from_head` when it updates the worktree. In this case, that might not be the right fit, so it may be worth considering some other kind of `.materialize()` that correctly handles the currently stashed uncommitted changes.
> 
> The `safe_checkout_from_head` that `materialize` currently calls uses a `safe_checkout` funciton, which allows you to pass a `current_head` along with the `new_head`, so it might be possible to add an option or third argument that also specifies using some other uncommitted changes tree.

Yeah, this is one of the differences as I outlined in the table above, but even if we have an extra argument for this, we still need to handle the other 2 differences. I thought about this a bit, but couldn't think of a "unified theory" for both cases. I do think that we need both `materialize()` (possibly renamed to `materialize_and_checkout`) and `materialize_without_checkout()` (ideally this would, by default, assert that all `HEAD`s of all workspaces are not changed, with possibly an option to disable this assertion).

jonathantanmy2: > Indeed, `safe_checkout` definitely isn't made to deal with this situation. My feeling is that edit-mode should leave the worktree clean because it knows what it wanted to accomplish, and from there the new checkout or adjustment can happen.

Yes, I think this is what is happening (in `save_and_return_to_workspace()`, a tree containing the contents of the working directory is created (including uncommitted files), a commit is created with that tree, and eventually a `repo.checkout_head(Some(CheckoutBuilder::new().force()))?;` call which seems to discard all uncommitted files (I didn't verify this though).

jonathantanmy2:
Also I'm not sure if this is the best place to ask, but in https://github.com/gitbutlerapp/gitbutler/blob/master/crates/but-rebase/src/graph_rebase/creation.rs#L58 , when we push to `head_selectors`, I thought that we should be pushing refs that are pointed to by `HEAD`, not entry point refs (as in the code)? (The entry point ref is usually but not always `HEAD` - like in the case of exiting edit mode, the entry point ref is `gitbutler/workspace` but `HEAD` is `gitbutler/edit`.)

jonathantanmy2:
Even if @Caleb-T-Owens has no comments, I still need code owner review for this.

Caleb-T-Owens (approved):
LGTM, thanks for working on this 👍

PR #12801, opened 2026-03-12
aliases of estib

estib-vega:
busted. no more hiding.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,4 +1,5 @@
-<stron@me.com> <35891811+estib-vega@users.noreply.github.com>
+estib <stron@me.com> Jose Vega <stron@me.com>
+estib <stron@me.com> Esteban Vega <35891811+estib-vega@users.noreply.github.com>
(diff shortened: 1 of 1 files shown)


.mailmap
```diff
@@ -1,4 +1,5 @@
-<stron@me.com> <35891811+estib-vega@users.noreply.github.com>
+estib <stron@me.com> Jose Vega <stron@me.com>
+estib <stron@me.com> Esteban Vega <35891811+estib-vega@users.noreply.github.com>
```
Byron, line 2: These can all be in one line I think. It's matching by emails addresses as well, so no need to specify the whole name, email address is enough.
```suggestion
estib <stron@me.com> <35891811+estib-vega@users.noreply.github.com>
```

But please double-check, armchair edit.

estib-vega: It didn't work for me :(
It tried it, but no luck

estib-vega: Validated using  `git shortlog -sne | grep estib` and ` git shortlog -sne | grep stron`

Byron: Yeah, too bad. Not that it matters much, it works - it was just me completely misremembering how mailmap works, which is funny because I implemented it.
Humans…

PR #12802, opened 2026-03-12
POC: Long running napi

estib-vega:
napi nap nap

This is a proof of concept of starting an long running task in a non-blocking way from JS, having the task run and propagate updated from Rust back to JS through a callback, and all while being cancellable.

This *should* work for things like rust-side-events and other nice things, like streaming AI updates and such

### DIFF
--- a/apps/lite/electron/src/model/longRunning.ts
+++ b/apps/lite/electron/src/model/longRunning.ts
@@ -1,0 +1,155 @@
+import {
+	longRunningCancelTsfn,
+	LongRunningEventKind,
+	longRunningStartTsfn,
+} from "@gitbutler/but-sdk";
+import type { LongRunningTaskSnapshot } from "#electron/ipc";
+
+const MAX_DURATION_MS = 600000;
+type LongRunningTaskListener = (event: LongRunningTaskSnapshot) => void;
+
+const activeTaskIds = new Set<number>();
+const tasks = new Map<number, LongRunningTaskSnapshot>();
+const listeners = new Set<LongRunningTaskListener>();
+
+/**
+	* Starts a non-blocking task in Rust and tracks task snapshots for renderer consumption.
+ */
+export function startLongRunningTask(durationMs: number): number {
+	if (!Number.isInteger(durationMs) || durationMs < 1 || durationMs > MAX_DURATION_MS) {
+		throw new Error("durationMs must be an integer between 1 and 600000.");
+	}
+
+	let taskId = 0;
+
+	taskId = longRunningStartTsfn(durationMs, (err, event) => {
+		const currentSnapshot = tasks.get(taskId);
+		if (!currentSnapshot) {
+			return;
+		}
+
+		if (err) {
+			activeTaskIds.delete(taskId);
+			const nextSnapshot: LongRunningTaskSnapshot = {
+				...currentSnapshot,
+				status: "error",
+				message: err.message ?? "unknown error",
+			};
+			setTaskSnapshot(nextSnapshot);
+			return;
+		}
+
+		if (!activeTaskIds.has(taskId)) {
+			return;
+		}
+
+		const snapshotWithStep: LongRunningTaskSnapshot = {
+			...currentSnapshot,
+			step: typeof event.step === "number" ? event.step : currentSnapshot.step,
+		};
+
+		if (event.kind === LongRunningEventKind.Progress) {
+			setTaskSnapshot({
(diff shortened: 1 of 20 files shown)


apps/lite/ui/src/hooks.ts
```diff
@@ -0,0 +1,69 @@
...
+							return event;
+						})
+					);
+				}
+			);
+		});
+	}, [queryClient]);
+
+	return tasksQuery;
+}
```
estib-vega, line 45: This is how a subscription would work in a query world.
We fetch the initial data on mount, and subscribe to its events as well.
That way, we always have the latest information as its pushed from the main thread (and in turn from Rust).

On unmount, we unsubscribe **but we don't stop the process**.
That option is still open if we wanted to, but not enabled by default

apps/lite/electron/src/model/longRunning.ts
```diff
@@ -0,0 +1,155 @@
...
+	longRunningStartTsfn,
+} from "@gitbutler/but-sdk";
+import type { LongRunningTaskSnapshot } from "#electron/ipc";
+
+const MAX_DURATION_MS = 600000;
+type LongRunningTaskListener = (event: LongRunningTaskSnapshot) => void;
+
+const activeTaskIds = new Set<number>();
+const tasks = new Map<number, LongRunningTaskSnapshot>();
+const listeners = new Set<LongRunningTaskListener>();
```
estib-vega, line 13: We keep track of the tasks, active or not, and subscribers.

estib-vega: This also means that the processes are persisted across UI reloads.
Only explicitly cancelling the operations will terminate them.

crates/but-api/src/lib.rs
```diff
@@ -38,3 +38,388 @@ pub mod platform;
...
+                napi::Status::GenericFailure,
+                "task registry is poisoned".to_string(),
+            )
+        })?;
+        tasks.insert(task_id, should_interrupt.clone());
+        drop(tasks);
+
+        let rx = long_running_non_blocking_thread(
+            Duration::from_millis(u64::from(duration_ms)),
+            gix::progress::Discard,
```
Byron, line 184: In this 'optimal' Rust version, there is effectively 3 channels:

* progress
* interrupt checks
* return values

To handle progress, you'd have to do something [like this](https://github.com/GitoxideLabs/gitoxide/blob/9d4956551147e088fdcbdb02d2e414200780bf5b/src/shared.rs#L80-L84) which is effectively:

* create a `progress::Tree` to hold the hierarchial progress. This looks like : https://asciinema.org/a/346619
* create a thread that polls the status of the progress and sends it to the frontend. This would basically just send the entire state over and the UI can render it.

Here one can also use other progress crates, for instance to only support a single progress value, but this progress was specifically built to support complex layered operations that have multiple progresses on their own, i.e. `gix clone` uses it and you know what I mean.
Oh, before I forget, this `Progress` trait can also be used to send progress messages, so it's pretty powerful and good to show the user that something is happening.

My recommendation is to use it, but… it's also a UX question and maybe nobody needs more than one. If you were to run a `clone` though, it quickly becomes hard to map this to a single progress that also has some temporal resolution to it.

CC @OliverJAsh .

apps/lite/electron/src/model/longRunning.ts
```diff
@@ -0,0 +1,155 @@
...
+			activeTaskIds.delete(taskId);
+			setTaskSnapshot({
+				...snapshotWithStep,
+				status: "done",
+				message: undefined,
+			});
+			return;
+		}
+
+		if (event.kind === LongRunningEventKind.Cancelled) {
```
Byron, line 70: If I read this correctly, here you'd have to set `should_interrupt` to `true`.
After all, the frontend has to have a mechanism to set that boolean so the thread which polls it knows it should stop.

Byron: Yikes, GitHub you got me. I thought I am looking at `lib.rs`, but I am not.
Anyway, I don't seem to be able to find the spot where this variable is set.

estib-vega: The variable `event` is one of the arguments passed into this callback. So this comes originally from the rust side --> napified --> ends up here.

> If I read this correctly, here you'd have to set should_interrupt to true.

That's correct. In this implementation, the way we stop it from the JS side is through the `cancelLongRunningTask` function bellow.

PR #12810, opened 2026-03-12
feat(cli): initial `but tui`

davidpdrsn:
This adds an initial, basic, TUI thats highly inspired by the excellent [jjui](https://github.com/idursun/jjui). The goal is to make an interactive `but status` that feels lightweight and snappy:

https://github.com/user-attachments/assets/6c872774-3d48-4f55-a824-bfd417d8033d

Its currently pretty bare bones but has the core architecture in place:
- Integration with `but status` which can now render in "one shot" or "interactive" mode.
- Event loop with state management and error handling.
- Setup for adding key binds and modes.

The only operation it supports is rubbing but that'll be expanded.

The command is currently hidden from `--help` and should remain so until its more mature.

Known shortcomings:
- No scrolling.
- No mouse support.
- No tests.
- Currently lives in the `but` crate but should probably be its own standalone crate that uses the modern `but-api`s, like but-lite does.
- Cursor position is lost when rubbing into a commit. This is because the commit id changes so after refreshing the status we cannot find the commit the cursor was previously on.

### TODO

- [x] Self review
- [x] Docs

### DIFF
--- a/crates/but/src/command/legacy/status/output.rs
+++ b/crates/but/src/command/legacy/status/output.rs
@@ -1,0 +1,286 @@
+use std::sync::Arc;
+
+use ratatui::text::Span;
+
+use crate::CliId;
+
+#[derive(Default)]
+pub(super) struct StatusOutput {
+    pub(super) lines: Vec<StatusOutputLine>,
+}
+
+impl StatusOutput {
+    fn push_line(
+        &mut self,
+        connector: Option<Vec<Span<'static>>>,
+        line: Vec<Span<'static>>,
+        data: StatusOutputLineData,
+    ) {
+        self.lines.push(StatusOutputLine {
+            connector,
+            line,
+            data,
+        });
+    }
+
+    pub(super) fn update_notice(&mut self, line: Vec<Span<'static>>) {
+        self.push_line(None, line, StatusOutputLineData::UpdateNotice);
+    }
+
+    pub(super) fn connector(&mut self, connector: Vec<Span<'static>>) {
+        self.push_line(
+            Some(connector),
+            <_>::default(),
+            StatusOutputLineData::Connector,
+        );
+    }
+
+    pub(super) fn staged_changes(
+        &mut self,
+        connector: Vec<Span<'static>>,
+        line: Vec<Span<'static>>,
+        id: CliId,
+    ) {
+        self.push_line(
+            Some(connector),
+            line,
+            StatusOutputLineData::StagedChanges {
+                cli_id: Arc::new(id),
+            },
+        );
+    }
+
+    pub(super) fn staged_file(
+        &mut self,
+        connector: Vec<Span<'static>>,
+        line: Vec<Span<'static>>,
+        id: CliId,
+    ) {
+        self.push_line(
+            Some(connector),
+            line,
+            StatusOutputLineData::StagedFile {
+                cli_id: Arc::new(id),
+            },
+        );
+    }
+
+    pub(super) fn unstaged_changes(
+        &mut self,
+        connector: Vec<Span<'static>>,
+        line: Vec<Span<'static>>,
+        id: CliId,
+    ) {
+        self.push_line(
+            Some(connector),
+            line,
+            StatusOutputLineData::UnstagedChanges {
+                cli_id: Arc::new(id),
+            },
+        );
+    }
+
+    pub(super) fn unstaged_file(
+        &mut self,
+        connector: Vec<Span<'static>>,
+        line: Vec<Span<'static>>,
+        id: CliId,
+    ) {
+        self.push_line(
+            Some(connector),
+            line,
+            StatusOutputLineData::UnstagedFile {
+                cli_id: Arc::new(id),
+            },
+        );
+    }
+
+    pub(super) fn branch(
+        &mut self,
+        connector: Vec<Span<'static>>,
+        line: Vec<Span<'static>>,
+        id: CliId,
+    ) {
+        self.push_line(
+            Some(connector),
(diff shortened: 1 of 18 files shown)


Byron:
Exciting!

> Cursor position is lost when rubbing into a commit. This is because the commit id changes so after refreshing the status we cannot find the commit the cursor was previously on.

This will probably be fixed once the workspace contains guaranteed ChangeIDs. The prerequisite for this is #12799.

crates/but/src/command/legacy/status/render_oneshot.rs
```diff
@@ -0,0 +1,120 @@
...
+    utils::WriteWithUtils,
+};
+
+pub(super) fn render_oneshot(
+    output: &StatusOutput,
+    out: &mut dyn WriteWithUtils,
+) -> anyhow::Result<()> {
+    use std::io::IsTerminal as _;
+
+    let use_color = std::env::var_os("CLICOLOR_FORCE").is_some() || std::io::stdout().is_terminal();
```
davidpdrsn: Note to self: This should probably come from `out`.

crates/but/src/id/mod.rs
```diff
@@ -1042,11 +1042,21 @@ impl PartialEq for CliId {
...
+                    id: l_id,
+                    path: l_path,
+                    ..
+                },
+                Self::CommittedFile {
+                    id: r_id,
+                    path: r_path,
+                    ..
+                },
+            ) => l_id == r_id && l_path == r_path,
```
davidpdrsn, line 1056: Hopefully this doesn't break anything but feels like a bug that we didn't compare the path of the files 🤔 The TUI uses this to restore cursor position after reloading.

crates/but/src/utils/output_channel.rs
```diff
@@ -130,6 +130,7 @@ pub trait WriteWithUtils: std::fmt::Write + 'static {
     fn truncate_if_unpaged(&self, text: &str, max_width: usize) -> String;
 
     /// Return `true` if the output is being passed through a pager.
+    #[expect(dead_code)]
     fn is_paged(&self) -> bool;
```
davidpdrsn: Note to self: Should we still be using this?

crates/but/src/command/legacy/status/render_oneshot.rs
```diff
@@ -0,0 +1,122 @@
...
+    command::legacy::status::{StatusOutput, StatusOutputLine},
+    utils::WriteWithUtils,
+};
+
+// TODO(david): rather than buffing all the output in StatusRenderMode::Oneshot we should just send
+// it directly to the screen
+pub(super) fn render_oneshot(
+    output: StatusOutput,
+    out: &mut dyn WriteWithUtils,
+) -> anyhow::Result<()> {
```
davidpdrsn, line 18: I'm pretty surprised that something like this isn't built into ratatui 😕

It is covered by unit tests but would still much appreciate some manual testing of this because I'm slightly scared of breaking things. This code is responsible for rendering the regular `but status` output.

crates/but/src/command/legacy/status/render_oneshot.rs
davidpdrsn: Fixed by translating into `colored` output instead of doing our own ansi stuff. Should be more robust this way.

crates/but/src/utils/output_channel.rs
davidpdrsn: And indeed we were! The method exists on `WriteWithUtils` but `OutputChannel` also defined its own method with the exact same name. So just removed that and rely on the trait.

crates/but/src/command/legacy/status/tui/mod.rs
```diff
@@ -0,0 +1,526 @@
...
+    let mut app = App::new(output, flags, debug);
+
+    let mut messages = Vec::new();
+
+    // second buffer so we can send messages from `App::handle_message`
+    let mut other_messages = Vec::new();
+
+    loop {
+        if app.should_render {
+            guard.terminal_mut().draw(|frame| app.render(frame))?;
```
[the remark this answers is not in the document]
davidpdrsn: Lol good catch 😂

crates/but/src/command/legacy/status/tui/cursor.rs
```diff
@@ -0,0 +1,130 @@
...
+            })
+            .find_map(|(idx, cli_id)| {
+                if &**cli_id == selected_cli_id {
+                    Some(idx)
+                } else {
+                    None
+                }
+            })?;
+        Some(Self(idx))
+    }
```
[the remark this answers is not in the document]
davidpdrsn: Yes will be doing that next.

crates/but/src/command/legacy/status/output.rs
```diff
@@ -0,0 +1,286 @@
...
+            | StatusOutputLineData::MergeBase
+            | StatusOutputLineData::File { .. } => true,
+            StatusOutputLineData::Connector
+            | StatusOutputLineData::UpdateNotice
+            | StatusOutputLineData::Warning
+            | StatusOutputLineData::Hint
+            | StatusOutputLineData::NoAssignmentsUnstaged
+            | StatusOutputLineData::UpstreamChanges
+            | StatusOutputLineData::EmptyCommitMessage => false,
+        }
```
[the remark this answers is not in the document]
davidpdrsn: Thats intentional. You'll be able to do things with them in the future such as editing the commit message.

krlvi (approved):
Honestly - looks good to me. Im excited to play with this

PR #12811, opened 2026-03-12
Reduce CI time by not building `but-api-macro-tests` accidentally

Byron:
Instead, run these tests in a separate job so they don't drag down the compiletime for everything.

### Tasks

* [x] compare before/after

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -276,7 +276,7 @@ jobs:
       - name: Dependencies for 'keyring'
         run: sudo ./scripts/install-minimal-debian-dependencies.sh
       - name: Check documentation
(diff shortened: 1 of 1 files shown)


Byron:
This PR does what it set out to do, and restores the original runtimes of all Rust tests.

Screenshot 2026-03-13 at 07 31 44

**However**, it also shows that but-api-macros-tests in its matrix takes 19m. It's uncached, so I'd merge this to see what happens once a cache is present. If it's still too slow then, there is ways to make it faster.

Recommending merging this as inevitable first step.

slarse (approved):
Excessive running time absolutely seems like a caching thing, let's give it a go.

PR #12818, opened 2026-03-13
tauri: scope menu shortcuts to focused window

nshcr:
This PR fixes an issue where, when multiple GitButler window instances are open, all instances respond to the same shortcut key event.

The root cause was that menu shortcut events were broadcast to all windows by default, and the front-end listeners were also attached globally. This PR scopes the menu shortcut emission to the focused window and updates the front-end listener target to the Window instance, preventing the broadcast behavior so that only the active window handles the shortcut.

Comparison video before and after the fix:

https://github.com/user-attachments/assets/15dc70ee-5c4f-4fc0-9d3a-bdf0ad9991b7


Byron (approved):
Thanks a lot for the fix!

To my mind it's safe to merge. I also validated that it still works in the single-window case.

PR #12831, opened 2026-03-13
fix(tui): refresh workspace before reloading

davidpdrsn:
I've been working on rewording commits from the TUI, but if you reworded the same commit twice, the second time it would fail, saying it couldn't find the commit. It seems as if there's some internal caches that needs to be refreshed when we'd reload the status.

This does fix the issues that I got with rewarding the same commit multiple times, but I'm not really sure if this is the proper fix.


crates/but/src/command/legacy/status/tui/mod.rs
```diff
@@ -195,17 +206,21 @@ impl App {
...
                 }
 
                 messages.extend([Message::EnterNormalMode, Message::Reload]);
             }
             Message::Reload => {
+                {
+                    let meta = ctx.meta()?;
+                    let (_guard, repo, mut ws, _) = ctx.workspace_mut_and_db()?;
+                    ws.refresh_from_head(&repo, &meta)?;
+                }
```
davidpdrsn, line 220: This is the actual interesting bit, the rest is just plumbing for error handling.

Byron: (*I only looked at this hunk specifically*)
Yes, that's the easy way let the cached workspace representation update itself. When `but-api` functions are used, the expectation is that they do this for you, or in other ways leave you with a workspace that is in sync with `repo`.
When calling plumbing, this won't be the case.
The reason you are actually seeing this, which I think is good as we can 'practice', is that probably the `Context` is long-lived, which has the benefit of hot caches and overall, can lead to better performance.

However, there is definitely one more caveat: the `gix::Repository` won't auto-update its snapshots for Git configuration yet, so that will go stale *if* any other mechanism of config updates is used than to go through `repo.config_snapshot_mut()`. I am working on rectifying this by switching everything over to `gix`, see #12747 .

Edit: Actually, this also isn't always true as there is code that can open configuration files directly. So I presume it's something to be aware of, and maybe might lead to a concept of recreating the `ctx.repo` every now and then, or maybe I could at least add a way to refresh the snapshot 'brutally'. Yeah, let me add that to `gix`.

https://github.com/GitoxideLabs/gitoxide/pull/2469

Byron (approved):

PR #12834, opened 2026-03-13
Add code owners for Lite

OliverJAsh:


krlvi:
No problem adding this but while iterating on this, perhaps it would be valuable to be able to merge without review? 

Also, if we were to merge this, there is no second code owner who could approve your PRs ;) 

I recommend skipping this for now, but your call :)

OliverJAsh:
I definitely don't want to hinder experimentation, but I'm imagining we can create branches for that, and we'd only merge them if/when we're confident with both the overall UX direction and code. But for any code that's being merged I'd like to maintain a high level of quality—it's the only way I know how to create high quality software (UX-wise).

At this early stage I don't think any one needs to review my code because I have the full (front-end) system inside my head. 🙈

OliverJAsh:
Long term I'd like to adopt this, as we did at Unsplash: https://www.raycast.com/blog/no-code-reviews-by-default

But I think we need to lay the groundwork first.

krlvi:
> Long term I'd like to adopt this, as we did at Unsplash: https://www.raycast.com/blog/no-code-reviews-by-default
> 
> But I think we need to lay the groundwork first.

therefore i was surprised by this :) The codeowner thing here would enforce review requirement for this path

CODEOWNERS
```diff
@@ -13,3 +13,4 @@
 /crates/but-update @krlvi @Byron @Caleb-T-Owens
 /crates/but-workspace @krlvi @Byron @Caleb-T-Owens
 /crates/gitbutler-edit-mode @krlvi @Byron @Caleb-T-Owens
+/apps/lite @OliverJAsh
```
krlvi, line 16: If you wish to merge, please add at least a second user (otherwise when you create a PR there won't be anybody who can approve your PR)

krlvi (changes requested):

CODEOWNERS
OliverJAsh: Ah, I thought I would still be able to merge (since I'm the code owner), but reading up on it it seems that's not how GitHub works. 🫠

Apparently it's possible to define a bypass rule, but I can't access the repo settings.

krlvi: > Apparently it's possible to define a bypass rule, but I can't access the repo settings. 

That is interesting! Id like to set this up actually

PR #12838, opened 2026-03-13
Use checked split

Caleb-T-Owens:


Byron:
By now I am relying on the Copilot generated summaries to quickly understand what a change is doing because our commit-messages aren't typically revealing any motivation or case that triggered the change (you are not alone in this either). Besides, GH hides them behind two clicks while Copilot is immediately available.

And while that works for me, detective work in the Git history of course still won't be helped by this, but maybe a good tool would also pull GH information automatically, for when it's invented. And why should that not be GitButler :)?

In any case, I think it's fair to say "screw commit messages" and solve it with tooling one day. (There is one tool that was specifically for this and whose name I forgot, but it was not intuitive to use, so I think there really is feature potential).

PR #12845, opened 2026-03-14
DRIVER-less tests

Byron:
This PR updates the virtual-branches tests to avoid using the DRIVER, which is highly undesirable as it is slow, and it's also not properly isolated, which just got worse somehow with `gitbutler.storePath`.

This is done in more modern tests, and we shouldn't have to suffer through these old tests more. They also have a habit of not being well isolated, and that just got worse somehow with `gitbutler.storePath`.

So this is an agent-mode attempt to get rid of it, no need for it to be pretty, just to tests the same things.

## Notes for the reviewer

It's all reviewed by me and no actual test expectation was changed, just the setup, so it's just an improvement. Finally, none of these crates should exist for too much longer, so whatever facilitates a transition is good :)?

## Tasks

* [x] review more carefully


Byron:
Closing in favor of #12747 which now contains this commit.

PR #12852, opened 2026-03-15
Fix delayed checks invalidation after push

mtsgrd:
The delayed forge invalidation was using stackId-based tags, but
listChecks provides tags by branch name and getPr by PR number.
Fix by invalidating checks by upstream branch name and PRs via
the list tag, all through forgeFactory.

Also remove no-op cross-slice Checks and PullRequests invalidations
from pushStack.invalidatesTags, since those tags are only provided
by the forge API slice, not the backend slice.


mtsgrd:
Blame: b9cb6611

PR #12869, opened 2026-03-16
edit-mode: don't needlessly write to index

jonathantanmy2:
The code currently writes to the index before calling `create_wd_tree`, but that is not necessary because:
 - `create_wd_tree` already handles worktree changes in any form, including unstaged changes, and
 - later, we do a `checkout_head` with force.

I've included a basic test that demonstrates this.

I noticed this when looking at how the gitbutler-edit-mode code uses libgit2, as part of potentially porting it to using gitoxide as much as it can.

### DIFF
--- a/crates/gitbutler-edit-mode/tests/edit_mode.rs
+++ b/crates/gitbutler-edit-mode/tests/edit_mode.rs
@@ -1,4 +1,5 @@
 use anyhow::{Result, anyhow};
+use bstr::ByteSlice as _;
 use but_core::{
     RefMetadata, RepositoryExt,
     ref_metadata::{StackId, WorkspaceCommitRelation, WorkspaceStack, WorkspaceStackBranch},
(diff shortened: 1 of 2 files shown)


crates/gitbutler-edit-mode/tests/edit_mode.rs
```diff
@@ -63,6 +64,34 @@ fn seed_metadata(repo: &gix::Repository) -> Result<()> {
...
 }
 
+#[test]
+fn basic_leaving_edit_mode() -> Result<()> {
+    let (mut ctx, _tempdir) = command_ctx("conficted_entries_get_written_when_leaving_edit_mode")?;
+    let repo = ctx.git2_repo.get()?;
+
+    let foobar = repo.head()?.peel_to_commit()?.parent(0)?.id();
+
+    let vb_state = VirtualBranchesHandle::new(ctx.project_data_dir());
```
Byron, line 74: Just as a note, I will remove the VBHandle usage as well as `git2` usage in follow-up PRs, as part of the bigger picture of limiting `git2` to where it really is needed (still) and migrating away from vb.toml.

Thus, if you can, it's helpful to avoid introducing any of those even today.

Byron (approved):
Thanks so much! Doing less without negative impact is a net win, and the test further cements this.

Byron:
Merging as it's a good base for the `gix` conversion work I am doing.

PR #12876, opened 2026-03-17
Tear off a branch out of a stack

estib-vega:
Port the ability to tear off a branch out of a stack using the graph editor.

### DIFF
--- a/crates/but-workspace/src/branch/move_branch.rs
+++ b/crates/but-workspace/src/branch/move_branch.rs
@@ -5,7 +5,7 @@ use but_rebase::graph_rebase::{
     mutate::{SegmentDelimiter, SelectorSet, SomeSelectors},
 };
 
-/// Outcome of moving branches between stacks.
+/// Outcome of moving branches between or out of stacks.
 ///
 /// Returned by [function::move_branch()].
 #[derive(Debug)]
@@ -19,6 +19,9 @@ pub struct Outcome {
 
 pub(super) mod function {
 
+    use but_core::ref_metadata::StackId;
+    use but_rebase::graph_rebase::mutate::SomeSelectors;
+
     use super::get_disconnect_parameters;
 
     use super::Outcome;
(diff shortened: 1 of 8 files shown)


crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -28,12 +31,124 @@ pub(super) mod function {
...
+                subject_segment,
+                workspace_head,
+            )?;
+
+        let skip_reconnect_step = source_stack.segments.len() == 1;
+        editor.disconnect_segment_from(
+            subject_delimiter.clone(),
+            children_to_disconnect,
+            parents_to_disconnect,
+            skip_reconnect_step,
```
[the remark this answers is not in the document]
estib-vega: Yeeeep

Byron:
I should get to it soon. Meanwhile, @codex review (Codex didn't listen apparently)

crates/but-workspace/tests/workspace/branch/tear_off_branch.rs
```diff
@@ -0,0 +1,244 @@
...
+
+#[test]
+fn tear_off_top_most_branch() -> anyhow::Result<()> {
+    let (_tmp, graph, repo, mut meta, _description) =
+        named_writable_scenario_with_description_and_graph(
+            "ws-ref-ws-commit-single-stack-double-stack",
+            |meta| {
+                add_stack_with_segments(meta, 1, "A", StackState::InWorkspace, &[]);
+                add_stack_with_segments(meta, 2, "C", StackState::InWorkspace, &["B"]);
+            },
```
Byron, line 18: It's great to see this proliferate.

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -28,12 +31,124 @@ pub(super) mod function {
...
+        let head_selector = editor
+            .select_commit(workspace_head)
+            .context("Failed to find the workspace head in the graph.")?;
+
+        let Some(lower_bound_ref) = workspace
+            .lower_bound_segment_id
+            .map(|segment_id| &workspace.graph[segment_id])
+            .and_then(|segment| segment.ref_name())
+        else {
+            bail!("Tearing off a branch requires a workspace common base");
```
Byron, line 98: As the workspace must be a 'typical' one at this point, the only way this could happen is if the following is true:

* there is only one stack
* there is no target set

As a WS commit exists, it should be possible to tear the branch off (in future).

I might also add that the presence of a workspace commit is unlikely in future if there is only a single stack as the new unapply will be able to remove it automatically, so we don't have anything that isn't needed anymore.

All this I only write for reference, as having no target branch isn't allowed right now.

<details><summary>First Version</summary>

While it's fine to reduce scope to not be forced to handle all the cases, I wonder if there are valid usecases where this would get in the way.
If so, the bail message could convey that.

And while writing this, the only case I can imagine is to tear off a branch segment in single-branch mode which would then have to create a workspace of sorts automatically. It's probably super rare that this is a thing, so… probably Ok to assume we don't handle the no-lower-bound case.

</details>

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -28,12 +31,124 @@ pub(super) mod function {
...
+    ///
+    /// `subject_branch_name` - The branch to take out of a stack.
+    ///
+    /// Returns:
+    /// - Successful rebase
+    pub fn tear_off_branch(
+        mut editor: Editor,
+        workspace: &but_graph::projection::Workspace,
+        subject_branch_name: &FullNameRef,
+        stack_id_override: Option<StackId>,
```
Byron, line 52: `stack_id_override` is missing docs.

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -28,12 +31,124 @@ pub(super) mod function {
...
+    /// `editor` is assumed to have been generated from the given `workspace`
+    /// and therefore aligned.
+    ///
+    /// `workspace` - Used for getting the surrounding context of the branch being torn off.
+    ///     In the future, we should not rely on the projection and do it fully on the graph.
+    ///
+    /// `subject_branch_name` - The branch to take out of a stack.
+    ///
+    /// Returns:
+    /// - Successful rebase
```
Byron: This doesn't seem true when looking at the type. What seems better is to link the outcome via something like

```markdown
Returns an [outcome](Outcome) for potential materialisation.
```

The idea is to make it easy to jump to the more exhaustive docs of the outcome type which carries a lot of 'knowhow' to discover, like updated metadata and of course a rebase operation that thus far wasn't materialised.

crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -28,12 +31,124 @@ pub(super) mod function {
     use but_rebase::graph_rebase::Editor;
     use gix::refs::FullNameRef;
 
+    /// Remove a branch out of a stack, creating a new stack out of it.
```
Byron: Maybe it would be good to make the 'dry-run' semantics clearer with something like this:

```suggestion
    /// Remove a branch out of a stack, creating a new stack out of it, in memory.
```

The above is the shortest way I could come up with, but what I really wanted to do is this:

```suggestion
    /// Remove a branch out of a stack, creating a new stack out of it, which is observable only in the returned [`Outcome`].
```

That way the `Returns` portion can also be removed, while important semantics are immediately evident.

crates/but-workspace/tests/workspace/branch/tear_off_branch.rs
```diff
@@ -0,0 +1,244 @@
...
+    insta::assert_snapshot!(graph_workspace(&ws), @"
+    📕🏘️:0:gitbutler/workspace[🌳] <> ✓refs/remotes/origin/main on 85efbe4
+    ├── ≡📙:4:C on 85efbe4 {2}
+    │   ├── 📙:4:C
+    │   │   └── ·09bc93e (🏘️)
+    │   └── 📙:5:B
+    │       └── ·c813d8d (🏘️)
+    └── ≡📙:3:A on 85efbe4 {1}
+        └── 📙:3:A
+            └── ·09d8e52 (🏘️)
```
Byron, line 40: It would be interesting to see at least one test that does this to 'virutal' stacks and segments, i.e. those that don't own a commit.

crates/but-workspace/tests/workspace/branch/tear_off_branch.rs
```diff
@@ -0,0 +1,244 @@
...
```
Byron, line 190: What happens if there is stacks `A` and `B`, and they sit on top of shared segment `C` which sits on top of `D`, and we tear out `C`?

This like a fork-shape, with `D->C` being the handle, and `A` and `B` being the spiky bits.
My feeling is that it *should* work already.

Byron (approved):
Thanks a lot!

I only looked at `but-workspace` and didn't run anything locally (not knowing how), and came up with these bigger topics, in order of importance:

* tests for virtual segment tear off and shared segment tear off
* making the docs even better

All of that could be done in a follow-up, hence the approval, but it's probably good to wait for feedback on `but-rebase`.

krlvi (approved):

crates/but-workspace/tests/workspace/branch/tear_off_branch.rs
estib-vega: That should indeed work. Is that representable in the metadata?

If so I can add a test for that as a follow-up

Byron: It does't need metadata at all, none of these tests do from what I can tell. All they do is to color the books nicely, I think. 

There is two kind of metadata as well, the one that is assigned to the workspace to show what the stacks and segments are supposed to be, and the metadata associated with References to contain information like reviews and such. And all is optional.

estib-vega: I see.

We do play around with metadata. Both this and move branch update the meta (if any is available).

> What happens if there is stacks A and B, and they sit on top of shared segment C which sits on top of D, and we tear out C?

It should work indeed. As long as we have a place to put a branch at after tear-off (lower bound) then it should be fine.

PR #12884, opened 2026-03-17
add-bytes-for-referring-to-reference

Caleb-T-Owens:
* Closes GB-1128

### DIFF
--- a/crates/but-api/src/commit.rs
+++ b/crates/but-api/src/commit.rs
@@ -306,26 +306,23 @@ pub mod ui {
         #[serde(with = "but_serde::fullname_lossy")]
         #[cfg_attr(feature = "export-schema", schemars(with = "String"))]
         Reference(gix::refs::FullName),
+        /// Relative to a reference, this time with teeth
+        #[cfg_attr(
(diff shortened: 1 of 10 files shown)


crates/but-api/src/commit.rs
```diff
@@ -660,16 +657,17 @@ pub fn commit_move_changes_between(
 pub fn commit_move_only(
     ctx: &mut but_ctx::Context,
     subject_commit_id: gix::ObjectId,
-    anchor_commit_id: gix::ObjectId,
+    relative_to: ui::RelativeTo,
```
Byron, line 660: `ui` types shouldn't be used in the Rust portion of the `but-api`, instead there should be an unconvoluted type (this `ui` type has redundancies that aren't needed.

The problem here is that parameter conversion is limited via the `but_api` macro, which certainly should be used here, and I will try to address that.

Byron (commented):
Thanks, lots of code deleted, who wouldn't love that!

I found an issue related to the capabilities of the `but_api` macro that I will try to resolve.

PR #12891, opened 2026-03-17
gitbutler-commit: update CommitExt trait documentation

DanielEScherzer:
## 🧢 Changes

gitbutler-commit: update CommitExt trait documentation

## ☕️ Reasoning

After #5722 it was used for both `git2::Commit` and `gix::Commit`, since #11788 it has only been used for `gix::Commit`.


Byron (approved):
Thanks a lot, good catch!

PR #12893, opened 2026-03-17
gitbutler-edit-mode: remove more git2 usage

jonathantanmy2:
2 commits total (remove more git2 usage, and a no longer necessary call)


Byron (approved):
Thanks a lot, a great improvement!

I tested it locally in the GUI as well (and maybe I should get used to using the CLI for that as faster alternative).

PR #12909, opened 2026-03-18
Fix alias expansion failing when persistent flags precede the alias

mango766:
## Problem

Running `but --json st` (or any command with root-level flags before an alias) fails with:

```
Error: "but st" is not a command. Type "but --help" to see all available commands.
```

This happens because `expand_aliases()` always inspects `args[1]` as the potential alias. When persistent flags appear first, `args[1]` is a flag like `--json`, and the function short-circuits without ever seeing the actual alias.

Reported in #12697.

## Root Cause

In `crates/but/src/alias.rs`, `expand_aliases()` hardcodes the alias position at index 1:

```rust
let potential_alias = match args[1].to_str() { ... };
if potential_alias.starts_with('-') || is_known_subcommand(potential_alias) {
    return Ok(args);  // <-- bails here when args[1] is "--json"
}
```

## Fix

Introduce `find_first_positional()` which scans past leading flags and their values to locate the first positional argument. The function knows which root-level flags consume a value (`-C`, `--current-dir`, `-f`, `--format`) and skips them accordingly. Boolean flags (`--json`, `-t`, `--status-after`) are stepped over one at a time.

The expanded result preserves all surrounding arguments in their original positions:
- `but --json st` → `but --json status`
- `but -C /path stf --verbose` → `but -C /path status --files --verbose`

## Testing

Added unit tests covering:
- Alias after boolean flag (`but --json st`)
- Alias after flag with value (`but -C /path stf`)
- Trailing args preserved after expansion
- All-flags-no-positional passthrough

Closes #12697

### DIFF
--- a/crates/but/src/alias.rs
+++ b/crates/but/src/alias.rs
@@ -32,9 +32,14 @@ const DEFAULT_ALIASES: &[(&str, &str)] = &[
 
 /// Expands command aliases before argument parsing.
 ///
-/// If the first argument after "but" is an alias defined in git config
-/// (under `but.alias.<name>`), it will be expanded to its definition.
-/// Additional arguments are preserved and appended after the expansion.
+/// Scans through arguments after "but", skipping any leading flags (and their
+/// values), to find the first positional argument. If that argument is an alias
+/// defined in git config (under `but.alias.<name>`) or a built-in default alias,
+/// it will be expanded to its definition. All surrounding arguments (flags before
+/// and arguments after the alias) are preserved.
+///
+/// This allows persistent flags to appear before aliases, e.g.:
+///   `but --json st` expands to `but --json status`
 ///
 /// # Arguments
 ///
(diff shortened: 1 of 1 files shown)


crates/but/src/alias.rs
```diff
@@ -83,14 +95,54 @@ pub fn expand_aliases(args: Vec<OsString>) -> Result<Vec<OsString>> {
...
+    expanded.extend(args[..alias_index].iter().cloned()); // "but" + any leading flags
     expanded.extend(alias_parts);
-    expanded.extend(args[2..].iter().cloned()); // Remaining args after the alias
+    expanded.extend(args[alias_index + 1..].iter().cloned()); // Remaining args after the alias
 
     Ok(expanded)
 }
 
+/// Known root-level flags that consume the next argument as a value.
+const FLAGS_WITH_VALUE: &[&str] = &["-C", "--current-dir", "-f", "--format"];
```
Byron, line 108: This will probably go stale at some point, but then again, it should be rare to change global flags.

Byron (approved):

I think this looks good and solves a real problem, so I'd definitely want to merge this.

@codex review

Byron:
@mango766 Could you take a look at the bot-comments and at add your own comments to them (while addressing them or not)?
When done, I can swoop in and sign your commit to make the merge, if you didn't do it already by then.
*Setting to draft while the review comments are addressed*.

slarse:
Hi!

This has been sitting for a long time and is a rather brittle solution. After we introduced external commands, a new opportunity opened up to fix this in a more robust fashion, which I'm working on over in #13981. As I believe it's an overall better solution to this problem, I'm closing this PR in favor of #13981.

We thank you for the contribution and hope to see you back in the future!

PR #12913, opened 2026-03-18
Expose `RejectedChange` in `but-sdk`

OliverJAsh:
This type would be useful for the front-end.

I used AI to generate this and I have no idea if it's correct. If it looks wrong, would someone else be able to pick this up? 🙏

Closes GB-1168

### DIFF
--- a/crates/but-api/src/commit.rs
+++ b/crates/but-api/src/commit.rs
@@ -54,6 +54,7 @@ pub mod json {
     use serde::Serialize;
 
     use crate::{commit::CommitMoveResult, json::HexHash};
+    use but_core::tree::create_tree::RejectionReason;
 
     use super::{
         CommitCreateResult, CommitInsertBlankResult, CommitRewordResult, MoveChangesResult,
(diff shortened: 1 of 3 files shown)


krlvi:
@OliverJAsh the svelte app might need to be adapted for the new type. Also it would be neat to document the fields so that it's nice to consume

OliverJAsh:
> the svelte app might need to be adapted for the new type.

I was aiming to just expose the type without changing it, but I'm not sure if I achieved that…

krlvi:
right maybe nothing needs doing 🤷‍♂️

OliverJAsh:
I'll close this and open a ticket instead because I have no idea what I'm doing in Rust code.

krlvi:
as i was triaging the linear issue and looking at this closer - this seems fine to do!

krlvi:
oh, branch is gone

OliverJAsh:
We can restore the branch?

OliverJAsh:
Ah, conflicts. I think it would be better for someone with Rust knowledge to pick this up.

krlvi:
> Ah, conflicts. I think it would be better for someone with Rust knowledge to pick this up.

re-implemented it here https://github.com/gitbutlerapp/gitbutler/pull/12998

PR #12920, opened 2026-03-18
Hide featured nightly from other builds list

Byron:
This seems more natural to me, particularly because I fell for it multiple times now that I tried to download the previous version, but got the current one.


PavelLaptev:
@Byron what's the difference between `nightlies` and `OtheNightlies`?

PavelLaptev (approved):

Byron:
OtherNightlies is nightlies, without the current version. It's basically `nightlies[1..]`.
Unfortunately I can't test it, so could you merge it if you are sure this doesn't break anything @PavelLaptev ?

PR #12941, opened 2026-03-19
feat(desktop): show a working indicator

foxfirecodes:
## 🧢 Changes

this PR is a small UX improvement that i found myself wishing for constantly: show a small visual indicator whenever *any* operation is happening, somewhere that is readily visible.

## ☕️ Reasoning

while a lot of operations come with a dedicated loading state (like showing a spinner on a button), a lot of operations dont have any kind of interaction/loading indicator. which has lead to a number of times where i thought i clicked an item in a dropdown (like squash all commits) only to sit there staring waiting for it to finish... to then realize i didnt actually click it lmao

so by putting a persistent and consistent visual indicator that the app is working on something, it helps address confusion for interactions where its difficult to provide a good loading state.

## 📌 Todos

the only things i'm not 100% are:
- the positioning of the indicator in the bottom right is not my favorite thing. maybe it belongs in the header? it can sorta clash with toasts rn (it shows above the very bottom right corner of the toast)
- the way we're wrapping `invokeTauri` in `trackOperation` feels a little gross to me since it requires you to use promise methods instead of try/catch w/ async/await buuuut at the same time wrapping the entire async block in `trackOperation` would nest the code a lot more. eh /shrug

## 🎮 Demo

(pay attention to the bottom right corner of the window)

https://github.com/user-attachments/assets/f0d080c3-64f4-4c1e-ab3d-929142099d5b

### DIFF
--- a/apps/desktop/src/components/WorkingIndicator.svelte
+++ b/apps/desktop/src/components/WorkingIndicator.svelte
@@ -1,0 +1,46 @@
+<script lang="ts">
+	import { ActiveOperationIndicator } from "$lib/activeOperations.svelte";
+	import { onDestroy } from "svelte";
+
+	const indicator = new ActiveOperationIndicator();
+	onDestroy(() => indicator.destroy());
+</script>
+
+{#if indicator.visible}
+	<div class="working-indicator" role="status" aria-label="Operation in progress">
+		<div class="working-indicator__dot"></div>
+	</div>
+{/if}
+
+<style lang="postcss">
+	.working-indicator {
+		z-index: var(--z-floating);
+		position: fixed;
+		bottom: 8px;
+		right: 8px;
+		display: flex;
+		align-items: center;
+		justify-content: center;
+		pointer-events: none;
+	}
+
+	.working-indicator__dot {
+		width: 8px;
+		height: 8px;
+		border-radius: 50%;
+		background-color: var(--clr-theme-pop-element);
+		animation: pulse 1s ease-in-out infinite;
+	}
+
+	@keyframes pulse {
+		0%,
+		100% {
+			opacity: 1;
+			transform: scale(1);
+		}
+		50% {
+			opacity: 0.4;
+			transform: scale(0.75);
+		}
+	}
+</style>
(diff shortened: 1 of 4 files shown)


Byron:

While I like the concept and also think it's useful, it's probably still something to flesh out design wise. Something I'd not want to have is 'popping' or layout shift. @PavelLaptev should probably take this over for that reason.

I also recommend to rebase onto `master`, there was some hickup with generated types apparently that were fixed now.

@codex review

PavelLaptev:
@foxfirecodes thanks for the contribution and for identifying a real pain point — the lack of feedback during operations is definitely something that needs to be addressed!

However, I think a global indicator is masking the underlying problem rather than solving it. When a user clicks something, they need confirmation that their specific action registered and is being processed. A generic dot in the corner doesn't provide that reassurance. It also risks becoming background noise that users learn to ignore.

The proper fix is contextual loading states: the button you clicked disables/shows a spinner, the list item you triggered gives visual feedback, etc. That's a bigger and harder effort, but it's the right direction.

I'd suggest closing this PR for now. If you're interested in tackling the root problem, picking a few of the most-used interactions (like the dropdown actions you mentioned where we can show a progress bar ) and adding proper per-action loading states would be a great place to start.

I've added GH desktop as an example:

https://github.com/user-attachments/assets/a15e6e86-4c2f-4ce6-86b3-9f928f8848d8

foxfirecodes:
@PavelLaptev heard! i do agree that adding proper loading states throughout the app is the ideal solution. that being said there's some cases where i'm not quite sure how to show a loading state... like for example, if i drag some files over to a commit to amend it, where do i show the loading indicator? there's no modal, button, or dropdown item to show a spinner/disabled state on. i guess maybe we add a loading spinner on the commit itself? 

im fine with closing this for now if that's what you'd prefer, but just some food for thought; i do think there are some trickier to tackle actions which dont have a clear home for a loading state

edit: I see the example you gave with the modal. that's probably reasonable since you likely shouldn't be doing anything else in the app while an amend/squash operation is going anyways. im good to close this, i'll tackle individual cases and open PRs from there

PR #12945, opened 2026-03-19
Fix CI with update to generated files

mtsgrd:


Byron:
Something is wrong here. Maybe the order isn't stable?
How can it be that every PR now fails with something around these files? Too strange.

PR #12948, opened 2026-03-20
Add failing gix blob-merge tests for Myers algorithm issue

mtsgrd:
gix's blob merge (gix_merge::blob::builtin_driver::text) with the Myers
diff algorithm produces a false conflict on inputs where git merge-file
resolves cleanly. The trigger is a 3-way merge where base→ours is a
large expansion (~70 new lines inserted above a section comment) and
base→theirs is a single-line deletion of that same comment. imara-diff's
Myers implementation produces different hunk boundaries than git's xdiff
for this input, causing the merge to see the deletion as overlapping
with the insertion — a false conflict. The Histogram algorithm in
imara-diff handles the same input correctly.

This directly impacts GitButler's commit amend operation: create_tree()
uses gix's tree merge for the cherry-pick step, and the false conflict
causes valid changes to be rejected with CherryPickMergeConflict.

Add a #[should_panic] test that documents the upstream bug and will
start failing (reminding us to remove the annotation) once gix is fixed,
plus a Histogram-based test that verifies the working alternative.

Upstream:
- https://github.com/GitoxideLabs/gitoxide/pull/2476
- https://github.com/GitoxideLabs/gitoxide/issues/2475

### DIFF
--- a/crates/but-core/tests/core/main.rs
+++ b/crates/but-core/tests/core/main.rs
@@ -5,6 +5,7 @@ mod commit;
 mod diff;
 mod extract_remote_name_and_short_name;
 mod json_samples;
+mod merge;
 mod ref_metadata;
 mod settings;
 mod snapshot;
--- a/crates/but-core/tests/core/merge.rs
+++ b/crates/but-core/tests/core/merge.rs
@@ -1,0 +1,108 @@
+/// Tests for gix blob-merge behaviour.
+///
+/// Some tests in this module are expected to **fail** until an upstream gix bug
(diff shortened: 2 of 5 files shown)


Byron:
That's quite a find!

I wouldn't even be surprised if it's a problem in the actual diff/merge algorithm in `gix` and that it's just coincidentally right with the histogram algorithm.
Something that is still notably different in `gix` is that it uses `imara-diff` 0.1 which doesn't have Git-style slider adjustments, which might also affect the maybe buggy text merge.

PR #12960, opened 2026-03-20
TUI not committing/amending all hunks

davidpdrsn:
I noticed that if you used the TUI to

- Create a new commit from the unassigned changes
- or amend an existing commit using the but-api based rub (shift+r)

then sometimes not all hunks would get committed and some would be left unassigned.

This PR contains an attempt at fixing it but doesn't feel like its the right fix. It feels too local and I'm wondering if there is something wrong in but-api or perhaps there is some shared APIs thats missing.

For context [here](https://github.com/gitbutlerapp/gitbutler/blob/master/crates/but/src/command/legacy/status/tui/mod.rs#L955-L972) is how the TUI is finding the unassigned stuff to commit.

### DIFF
--- a/crates/but/src/command/legacy/status/tui/mod.rs
+++ b/crates/but/src/command/legacy/status/tui/mod.rs
@@ -998,6 +998,7 @@ impl App {
         };
 
         // create commit
+        let changes_to_commit = but_workspace::flatten_diff_specs(changes_to_commit);
         let commit_create_result = but_api::commit::create::commit_create(
             ctx,
             insert_commit_relative_to,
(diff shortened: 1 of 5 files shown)


crates/but/src/command/legacy/status/tui/mod.rs
```diff
@@ -2519,6 +2520,42 @@ fn has_unassigned_changes(ctx: &mut Context) -> anyhow::Result<bool> {
...
+
+    grouped
+        .into_iter()
+        .map(|(path, hunk_headers)| DiffSpec {
+            previous_path: None,
+            path,
+            hunk_headers,
+        })
+        .collect()
+}
```
davidpdrsn, line 2557: Building the DiffSpec in this way seems to fix it but feels wrong to just do this locally here.

davidpdrsn:
This was fixed by https://github.com/gitbutlerapp/gitbutler/pull/13030

PR #12969, opened 2026-03-20
feat(tui): leave modes the way you came in

davidpdrsn:
Maybe this is just me but I find it quite convenient to leave a mode using the
same key bind that enters it. For example you can press `m` to enter move mode,
realize you wanted to do something else, and press `m` again to leave mode
mode. I find that nicer than reaching all the way up to escape.


Byron:
A mode toggle seems very intuitive!

> I find that nicer than reaching all the way up to escape.

Bonus points for making `Ctrl + [` an ESC equivalent. In vim mode I only use that, and by now I expect it to work everywhere else (and it does at least in all editors I tried).

davidpdrsn:
Yes that's a good idea!

PR #12988, opened 2026-03-23
Add piped `OutputChannel`

davidpdrsn:
_This PR is a reincarnation of https://github.com/gitbutlerapp/gitbutler/pull/12981_

# Why

I'm working on aligning the operations performed by the CLI and the TUI. So committing with the TUI should call the same handler function that `but commit` does.

All those CLI handler functions require a `&mut OutputChannel` for printing to stdout. If committing from the TUI I think instead we should instead show the output in a toast. This enables that by adding a variant of `OutputChannel` that writes to a set of anonymous pipes created with [`std::io::pipe`](https://doc.rust-lang.org/stable/std/io/fn.pipe.html). The TUI can create those pipes and read from them to render a ratatui toast after calling the handler.

This also enables reading streaming output from long running tasks.

# What

`OutputChannel` now holds both the stdout and stderr that it uses as trait objects:

```rust
pub struct OutputChannel {
    stdout: Box<dyn WritePlusIsTerminal>,
    stderr: Box<dyn WritePlusIsTerminal>,
}
```

`WritePlusIsTerminal` is just `std::io::Write + std::io::Terminal`. We need this because `std::io::Terminal` isn't implemented `PipeWriter`.

`ProgressChannel` and `InputOutputChannel` now also hold a `&mut OutputChannel` so they can use the same outputs. That lead to some double borrows where we mixed writing to the normal output and the progress. I fixed that by changing

```rust
let progress = out.progress_channel();
writeln!(out, "something for normal output"); // ❌ can't use `out` since `progress` holds a `&mut out`
writeln!(progress, "something for progress");
```

to instead

```rust
writeln!(out, "something for normal output");
writeln!(out.progress_channel(), "something for progress"); // borrows `out` only for this call
```

This solution is a bit boilerplatey but probably fine.

It might also be possible to do something like

```rust
impl OutputChannel {
    fn split(&mut self) -> (NormalOutput<'_>, ProgressChannel<'_>) { ... }
}
```

but it gets more complicated with we also want to get input, so I didn't explore this.

### DIFF
--- a/crates/but/src/utils/output_channel.rs
+++ b/crates/but/src/utils/output_channel.rs
@@ -1,4 +1,4 @@
-use std::io::{IsTerminal, Write};
+use std::io::Write;
 
 use but_secret::Sensitive;
 use crossterm::event::{self, Event, KeyCode, KeyEvent, KeyEventKind, KeyModifiers};
@@ -37,8 +37,11 @@ pub enum ConfirmOrEmpty {
 pub struct OutputChannel {
     /// How to print the output, one should match on it. Match on this if you prefer this style.
     format: OutputFormat,
-    /// The output to use if there is no pager.
-    stdout: std::io::Stdout,
+    /// The stdout output to use if there is no pager.
+    stdout: Box<dyn WritePlusIsTerminal>,
+    /// The stderr output to use.
+    stderr: Box<dyn WritePlusIsTerminal>,
+    stderr_is_terminal: bool,
     /// Possibly a pager we are using. If `Some`, the pager itself is used for output instead of `stdout`.
     pager: Option<Pager>,
     /// When `Some`, JSON values written via `write_value` are captured here instead of going to stdout.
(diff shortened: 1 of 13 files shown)


crates/but/src/utils/output_channel.rs
```diff
@@ -1,4 +1,4 @@
-use std::io::{IsTerminal, Write};
+use std::io::Write;
```
davidpdrsn, line 1: All the intersting changes are in this file. Everything else is just mechanical changes resulting from this.

Byron:
Let's talk about this in the meeting as I think it's a key decision that I think will stick possibly forever. TUI + CLI intertwined or TUI as a separate client that uses APIs via `but-api`. 
What's happening now I see as a shortcut, and it's one that won't make us create the APIs that are needed to make clients (because we will just take shortcuts).

Another thought is that if the TUI was separated, it would still have to deal with shell-complexities when running `but` commands, including inputs and colors, and to do that right I guess there isn't many ways except for doing it like vim/helix do it.

But that's exactly the thing to think through, together.

davidpdrsn:
> Let's talk about this in the meeting as I think it's a key decision that I think will stick possibly forever. TUI + CLI intertwined or TUI as a separate client that uses APIs via but-api.

Sure! I'll hold off on this until we've decided on a direction.

To gather my own thoughts I'll still write what I think here.

> What's happening now I see as a shortcut, and it's one that won't make us create the APIs that are needed to make clients (because we will just take shortcuts).

My goals are:

- Feature parity between the CLI and the TUI. I think everything that the CLI can do the TUI should be able to do as well and vice versa.
- Avoid having to fix the same bugs in multiple places. I have found bugs in the TUI that aren't in the CLI and so I'd like to get those fixes without having to duplicate the CLI code.
  - Another example is running pre/post commit hooks. The CLI does that, the TUI doesn't. We should have non-legacy APIs to do that but until we do whats our strategy?
- Avoid having to port code to the new APIs twice. The TUI needs to things that we don't have non-legacy APIs for such as rubbing and squashing commits, as does the CLI. So by aligning the implementations we only have to port to non-legacy APIs once, instead of first in the TUI then later in the CLI.

I think we all want the same things but its more a question of sequencing.

> Another thought is that if the TUI was separated, it would still have to deal with shell-complexities when running but commands, including inputs and colors, and to do that right I guess there isn't many ways except for doing it like vim/helix do it.

I think this is largely solve already by having `OutputChannel` as the abstraction that all input/output goes through. If all the CLI commands just did `println!()` directly it would be much more of a pain.

Specifically about input: I think if you perform some operation in the TUI and get kicked out to shell for input then thats a bug. For any input the TUI needs to have nice ratatui UIs. That might mean refactoring some CLI handlers to request input differently. I was planning to cross that bridge when I got there.

Byron:
Thanks for sharing your thoughts, it's definitely what I was missing.

The path I see now is aligned with yours, which I understand toughly be the following:

- dealing with output seems manageable already and the CLI won't notice much (so its code doesn't get worse due to that)
- dealing with input might need a little more invasiveness later, but personally I don't think the code will get worse then as there already is just one way or two to get inputs.
- while the TUI is moving fast it benefits from being this close as CLI handlers and internals can be called directly to get certain functionality

I add to this that *if* there is a need, what's implemented in the CLI right now can be ported over to the `but-api` crate so the GUI's can also use it. That's extra work, but it's then warranted - more than 2 clients need that function, it must have proven itself.

That makes sense to me now, maybe there are other aspects though that come up in the session so it's definitely beneficial to spell it out there as well.

davidpdrsn:
I'll close this for now. Might come back to it later if we need it but so far it hasn't come up.

PR #12989, opened 2026-03-23
Don't prompt for user input in TUI tests

davidpdrsn:
When running `cargo test -p but --lib -- command_mode_runs_successful_command_and_returns_to_normal_mode` the test would stop and wait for the user to press enter. This happened because `std::io::stdin().is_terminal()` returns true even when running through cargo's test harness. I didn't notice this because I use cargo-nextest in which it returns false.

I think a decent fix is just to check `cfg!(test)`. That wont work in integration tests but since we don't have any integration tests of the TUI it should be fine.

We don't have integration tests of the TUI because of how [ratatui's test setup works](https://ratatui.rs/recipes/testing/snapshots/).


slarse (approved):
Fixes the issue on my end, thanks!

PR #12991, opened 2026-03-23
Remove unrequried static constraint

Caleb-T-Owens:
As a fly-by refactor, I saw this requiring a static lifetime which was a bit puzzling because we probably don’t want the metadata to have a  static lifetime, so I just tried to take it out.

### DIFF
--- a/crates/but-ctx/src/legacy.rs
+++ b/crates/but-ctx/src/legacy.rs
@@ -119,10 +119,7 @@ impl Context {
     pub fn workspace_and_meta_from_head(
         &self,
         _exclusive_access: &RepoExclusive,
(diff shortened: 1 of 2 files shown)


Caleb-T-Owens:
It seems like it might be required, but I'm really not sure that it should be required, and it would be worth while investigating whether this does actually cause some form of memory leakage

Byron:
Right now it's required due to how `meta()` is used in code. But in future, there should be a major change

* remove the RefMetadata trait entirely because it's more complex than it needs to be - we only need one implementation and it can live in `but-db` even.
  - this is also when `meta()` goes away entirely as `db` can be used to access the data.

PR #12998, opened 2026-03-23
Replace rejected changes tuple with typed UIRejectedChange struct

krlvi:
- Replaces the `Vec<(RejectionReason, BStringForFrontend)>` tuple in `UICommitCreateResult` with a proper `UIRejectedChange` struct containing `reason` and `path` fields
- Adds `RejectionReason` as a generated SDK type (proper string union instead of `string`)
- Renames field from `pathsToRejectedChanges` to `rejectedChanges`
- Updates desktop and lite frontends to use the new object shape
- Lite app now imports `UIRejectedChange` and `RejectionReason` from `@gitbutler/but-sdk` instead of defining them manually (resolves TODO comments)

### DIFF
--- a/crates/but-api/src/commit/json.rs
+++ b/crates/but-api/src/commit/json.rs
@@ -37,6 +37,21 @@ impl From<MoveChangesResult> for UIMoveChangesResult {
     }
 }
 
+/// A change that was rejected during commit creation, with the reason for rejection.
+#[derive(Debug, Serialize)]
+#[cfg_attr(feature = "export-schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "camelCase")]
+pub struct UIRejectedChange {
+    /// The reason the change was rejected.
+    pub reason: but_core::tree::create_tree::RejectionReason,
+    /// The file path of the rejected change.
+    #[cfg_attr(feature = "export-schema", schemars(with = "String"))]
+    pub path: but_serde::BStringForFrontend,
+}
+
+#[cfg(feature = "export-schema")]
+but_schemars::register_sdk_type!(UIRejectedChange);
+
 /// UI type for creating a commit in the rebase graph.
 #[derive(Debug, Serialize)]
 #[cfg_attr(feature = "export-schema", derive(schemars::JsonSchema))]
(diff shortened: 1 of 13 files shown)


crates/but-api/src/commit/json.rs
```diff
@@ -37,6 +37,21 @@ impl From<MoveChangesResult> for UIMoveChangesResult {
     }
 }
 
+/// A change that was rejected during commit creation, with the reason for rejection.
+#[derive(Debug, Serialize)]
+#[cfg_attr(feature = "export-schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "camelCase")]
+pub struct UIRejectedChange {
```
Byron, line 44: I think it's time to commit to a naming scheme. To me what mattered was less the UI, and more the fact that we have to use JSON as transport, hence the `json` module naming.
Unless we think the UI is relevant for this, I'd drop the UI prefixes entirely. If it is important, I'd think we have to be careful about `but-api` becoming specific to GUI needs, which is something a good general-purpose API (probably?) should avoid.

crates/but-core/src/tree/mod.rs
```diff
@@ -11,6 +11,7 @@ pub mod create_tree {
 
     /// Provide a description of why a [`crate::DiffSpec`] was rejected for application to the tree of a commit.
     #[derive(Default, Debug, Copy, Clone, PartialEq, serde::Serialize)]
+    #[cfg_attr(feature = "export-schema", derive(schemars::JsonSchema))]
```
Byron, line 14: Let's accept this as convenience as it's already in many places of `but-core`.

crates/but-api/src/commit/json.rs
```diff
@@ -37,6 +37,21 @@ impl From<MoveChangesResult> for UIMoveChangesResult {
...
+    /// The reason the change was rejected.
+    pub reason: but_core::tree::create_tree::RejectionReason,
+    /// The file path of the rejected change.
+    #[cfg_attr(feature = "export-schema", schemars(with = "String"))]
+    pub path: but_serde::BStringForFrontend,
```
Byron, line 49: I thought that `BStringForFrontend` is something to try to avoid in favor of `path_bytes: BString` and `path: String` combinations.
That also won't pollute a generic JSON type with frontend terminology that I thought we can avoid assuming.

Byron (approved):
All I have is a nit about naming, but I also think it's worth figuring it out going forward. Maybe the `BStringForFrontend` is also something to think about in `but-api`, it's something I might unleash an agent on right away.

Byron:
@krlvi Interestingly, the PR to remove BStringForFrontend from `but-api` yielded exactly what you are doing here, but a bit less complete. I liked that it doesn't alter `but-core`.

Please take a look, maybe something salvageable: https://github.com/gitbutlerapp/gitbutler/pull/13040

crates/but-api/src/commit/json.rs
krlvi: maybe we should invent ourselves a struct type `PathJson` or something that we can use across the application. And then the frontend can have helper functions to use it nicely. wdyt?
cc @OliverJAsh

krlvi:
Im gonna merge this since this is on a hot path of edits in the new frontend... and i dont wanna resolve conflicts anymore 😅

crates/but-api/src/commit/json.rs
Byron: This is two-pronged: BStringForFrontend shouldn't be used and it could just be String here. In a second step, we can probably just have a `WorktreeEntry` type which may have bytes and degenerated Unicode.

I will fix this up in a follow-up and propose that WorktreeEntry type.

PR #12999, opened 2026-03-23
feat(but-core): add programmatic signing control to commit creation

slarse:
This PR introduces programmatic overrides to signing behavior for the core commit flow. This will later be used to allow the graph rebase engine to decide when to sign programmatically  to be able to implement sign-on-push where we never sign for "local" operations.

`but-rebase` changes coming up in a separate PR.

<details>
<summary>Old description kept for historical reasons</summary>
This is big, but I couldn't find any super reasonable way to cut it down any more than I've done with this stack. It turns out that overhauling how we sign commits touches a lot of things :)

This PR puts down the groundwork for enabling sign-on-push functionality. Importantly, it contributes but-workspace abstractions to list unsigned commits and forcibly sign commits, as well as the underlying graph rebase engine work to actually accomplish the signing.

Note that there is still full backwards compatibility with the current sign-all-the-flippin-time functionality, and this PR alone should result in zero changes to end users as the new functionality isn't used anywhere.

See the tests in `crates/but-workspace/tests/workspace/commit/commits_list_unsigned.rs` and `crates/but-workspace/tests/workspace/commit/commits_sign.rs` for details about the new behavior.

Integration into the `but` CLI is done in a separate PR. I feel that the UX there still needs some polish, but I really need to land this PR as I've now had to fix two sets of merge conflicts just today :s

---
- <kbd>&nbsp;2&nbsp;</kbd> #13046 
- <kbd>&nbsp;1&nbsp;</kbd> #12999 👈 

</details>

### DIFF
--- a/crates/but-core/src/commit.rs
+++ b/crates/but-core/src/commit.rs
@@ -2,7 +2,7 @@ use std::{
     borrow::Cow, collections::HashSet, io::Write, path::Path, path::PathBuf, process::Stdio,
 };
 
-use anyhow::{Context as _, anyhow, bail};
+use anyhow::{Context as _, bail};
 use bstr::{BStr, BString, ByteSlice};
 use but_error::Code;
 use gix::objs::WriteTo;
@@ -148,6 +148,33 @@ impl From<&Headers> for Vec<(BString, BString)> {
     }
 }
 
+/// Determines how to sign the commit.
+#[derive(Default, PartialEq, Copy, Clone, Debug)]
+pub enum SignCommit {
+    /// Unconditionally sign the commit. Note that this places responsibility on the caller to
+    /// ensure that commit signing is configured, or at least handling that it was not.
+    Yes,
+    /// Do not sign the commit.
+    No,
+    /// Sign the commit only if `gitbutler.signCommits=true`, *or* `gitbutler.signCommits` is unset
+    /// but `commit.gpgSign=true`. In other words, `gitbutler.signCommits` takes precedence over
+    /// `commit.gpgSign`, the latter only being checked if the former is not at all configured.
+    ///
+    /// If signing fails, `gitbutler.signCommits` is set to `false` locally, preventing further
+    /// signing when this variant is supplied. This step is however skipped if
+    /// `gitbutler.signCommits` has been explicitly configured in non-local Git Config.
+    ///
+    /// The need for `gitbutler.signCommits` stems from the fact that it can be difficult to
+    /// impossible to validate before hand that signing is properly configured. Signing may also
+    /// break after validation has been performed. If signing is enabled for *all* committing but
+    /// fails, GitButler basically can't do anything, so we flip `gitbutler.signCommits=false` as a
+    /// kill switch to disable signing for GitButler. `gitbutler.signCommits` taking precedence
+    /// over `commit.gpgSign` means we can honor Git's signing settings by default, but disable it
+    /// in the event that we fail to sign without affecting Git.
+    #[default]
+    IfSignCommitsEnabled,
+}
+
 /// Write `commit` into `repo`, removing any existing commit signature first, optionally creating a
 /// new one based on repository configuration, and optionally updating `update_ref` to the new ID.
 ///
@@ -156,7 +183,7 @@ pub fn create(
     repo: &gix::Repository,
     mut commit: gix::objs::Commit,
     update_ref: Option<&gix::refs::FullNameRef>,
-    sign_if_configured: bool,
+    sign_commit: SignCommit,
 ) -> anyhow::Result<gix::ObjectId> {
     if let Some(pos) = commit
         .extra_headers()
@@ -165,7 +192,10 @@ pub fn create(
         commit.extra_headers.remove(pos);
     }
 
-    if sign_if_configured && repo.git_settings()?.gitbutler_sign_commits.unwrap_or(false) {
+    if (sign_commit == SignCommit::IfSignCommitsEnabled
+        && repo.git_settings()?.gitbutler_sign_commits.unwrap_or(false))
+        || sign_commit == SignCommit::Yes
+    {
         let mut buf = Vec::new();
         commit.write_to(&mut buf)?;
         match sign_buffer(repo, &buf) {
@@ -175,26 +205,28 @@ pub fn create(
                     .push((gix::objs::commit::SIGNATURE_FIELD_NAME.into(), signature));
             }
             Err(err) => {
-                if repo
-                    .config_snapshot()
-                    .boolean_filter("gitbutler.signCommits", |md| {
-                        md.source != gix::config::Source::Local
-                    })
-                    .is_none()
-                {
-                    repo.set_git_settings(&GitConfigSettings {
-                        gitbutler_sign_commits: Some(false),
-                        ..GitConfigSettings::default()
-                    })?;
-                    return Err(
-                        anyhow!("Failed to sign commit: {err}").context(Code::CommitSigningFailed)
-                    );
-                } else {
-                    tracing::warn!(
-                        "Commit signing failed but remains enabled as gitbutler.signCommits is explicitly enabled globally"
-                    );
-                    return Err(err);
+                tracing::warn!("Commit signing failed with sign_commit={sign_commit:?}");
+                if sign_commit == SignCommit::IfSignCommitsEnabled {
+                    if repo
+                        .config_snapshot()
+                        .boolean_filter("gitbutler.signCommits", |md| {
+                            md.source != gix::config::Source::Local
+                        })
+                        .is_none()
+                    {
+                        repo.set_git_settings(&GitConfigSettings {
+                            gitbutler_sign_commits: Some(false),
+                            ..GitConfigSettings::default()
+                        })?;
+                    } else {
+                        tracing::warn!(
+                            "Commit signing failed but remains enabled as gitbutler.signCommits is explicitly enabled globally"
+                        );
+                    }
                 }
+                return Err(err
+                    .context("Failed to sign commit")
+                    .context(Code::CommitSigningFailed));
             }
         }
     }
--- a/crates/but-core/tests/core/commit.rs
+++ b/crates/but-core/tests/core/commit.rs
@@ -162,15 +162,21 @@ mod headers {
 mod create {
     use anyhow::Context as _;
     use bstr::ByteSlice;
-    use but_core::commit;
-    use but_testsupport::writable_scenario_with_ssh_key;
+    use but_core::commit::{self, SignCommit};
+    use but_error::Code;
+    use but_testsupport::{writable_scenario, writable_scenario_with_ssh_key};
     use gix::{objs::commit::SIGNATURE_FIELD_NAME, refs};
 
     #[test]
     fn signs_commits_when_enabled() -> anyhow::Result<()> {
         let (repo, _tmp) = writable_scenario_with_ssh_key("single-signed");
 
-        let oid = commit::create(&repo, commit_from_head(&repo, "signed again")?, None, true)?;
+        let oid = commit::create(
+            &repo,
+            commit_from_head(&repo, "signed again")?,
+            None,
+            SignCommit::IfSignCommitsEnabled,
+        )?;
         let commit = repo.find_commit(oid)?;
         let commit = commit.decode()?;
 
@@ -178,6 +184,82 @@ mod create {
             commit.extra_headers().find(SIGNATURE_FIELD_NAME).is_some(),
             "a new signature should be written, the new commit needs to be signed"
         );
+        Ok(())
+    }
+
+    #[test]
+    fn sign_commit_yes_forcefully_signs_commit() -> anyhow::Result<()> {
+        let (mut repo, _tmp) = writable_scenario_with_ssh_key("single-signed");
+        repo.config_snapshot_mut()
+            .set_raw_value(&"gitbutler.signCommits", "false")?;
+        repo.config_snapshot_mut()
+            .set_raw_value(&"commit.gpgSign", "false")?;
+
+        let oid = commit::create(
+            &repo,
+            commit_from_head(&repo, "should sign")?,
+            None,
+            SignCommit::Yes,
+        )?;
+        let commit = repo.find_commit(oid)?;
+        let commit = commit.decode()?;
+
+        assert!(
+            commit.extra_headers().pgp_signature().is_some(),
+            "despite the settings saying everything is disabled, it still signs the commit"
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn sign_commit_yes_without_signing_key_errors() -> anyhow::Result<()> {
+        let (mut repo, _tmp) = writable_scenario("single-unsigned");
+
+        let err = commit::create(
+            &repo,
+            commit_from_head(&repo, "should not sign")?,
+            None,
+            SignCommit::Yes,
+        )
+        .expect_err("commit creation should fail as there is no signing key configured");
+
+        let error_code: Option<&Code> = err.downcast_ref();
+        assert_eq!(error_code, Some(&Code::CommitSigningFailed));
+
+        repo.reload()?;
+        assert_eq!(
+            repo.config_snapshot().boolean("gitbutler.signCommits"),
+            None,
+            "we do not touch this setting in this mode"
+        );
+
+        Ok(())
+    }
+
+    #[test]
+    fn sign_commit_if_enabled_without_signing_key_errors_and_sets_config() -> anyhow::Result<()> {
+        let (mut repo, _tmp) = writable_scenario_with_ssh_key("single-signed");
+        repo.config_snapshot_mut()
+            .set_raw_value(&"user.signingKey", "BAD-key")?;
+
+        let err = commit::create(
+            &repo,
+            commit_from_head(&repo, "should not sign")?,
+            None,
+            SignCommit::IfSignCommitsEnabled,
+        )
+        .expect_err("commit creation should fail as there is no signing key configured");
+
+        let error_code: Option<&Code> = err.downcast_ref();
+        assert_eq!(error_code, Some(&Code::CommitSigningFailed));
+
+        repo.reload()?;
+        assert_eq!(
+            repo.config_snapshot().boolean("gitbutler.signCommits"),
+            Some(false),
+            "prevent future failures by disabling signing for gitbutler"
+        );
+
         Ok(())
     }
 
@@ -193,7 +275,12 @@ mod create {
             "the fixture starts with a signed commit"
         );
 
-        let oid = commit::create(&repo, commit_from_head(&repo, "unsigned")?, None, false)?;
+        let oid = commit::create(
+            &repo,
+            commit_from_head(&repo, "unsigned")?,
+            None,
+            SignCommit::No,
+        )?;
         let commit = repo.find_commit(oid)?;
         let commit = commit.decode()?;
 
@@ -214,7 +301,7 @@ mod create {
             &repo,
             commit_from_head(&repo, "updates ref")?,
             Some(update_ref.as_ref()),
-            false,
+            SignCommit::No,
         )?;
         let reference = repo.find_reference(update_ref.as_ref())?;
 
@@ -238,7 +325,7 @@ mod create {
             .extra_headers
             .push(("x-test-header".into(), "kept".into()));
 
-        let oid = commit::create(&repo, new_commit, None, false)?;
+        let oid = commit::create(&repo, new_commit, None, SignCommit::No)?;
         let commit = repo.find_commit(oid)?;
         let commit = commit.decode()?;
 
--- a/crates/but-core/tests/fixtures/scenario/single-unsigned.sh
+++ b/crates/but-core/tests/fixtures/scenario/single-unsigned.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+git init
+
+echo "base" >base
+git add .
+git commit -m "base"
--- a/crates/but-rebase/src/commit.rs
+++ b/crates/but-rebase/src/commit.rs
@@ -1,6 +1,6 @@
 use anyhow::{Context as _, bail};
 use bstr::BStr;
-use but_core::RepositoryExt;
+use but_core::{RepositoryExt, commit::SignCommit};
 use gix::config::Source;
 
 /// What to do with the committer (actor) and the commit time when [creating a new commit](create()).
@@ -102,7 +102,16 @@ pub fn create(
     if settings.gitbutler_gerrit_mode.unwrap_or(false) {
         but_gerrit::set_trailers(&mut commit);
     }
-    but_core::commit::create(repo, commit, None, sign_if_configured)
+    but_core::commit::create(
+        repo,
+        commit,
+        None,
+        if sign_if_configured {
+            SignCommit::IfSignCommitsEnabled
+        } else {
+            SignCommit::No
+        },
+    )
 }
 
 /// Update the committer of `commit` to be the current one.
--- a/crates/gitbutler-repo/src/repository_ext.rs
+++ b/crates/gitbutler-repo/src/repository_ext.rs
@@ -2,7 +2,10 @@ use std::str;
 
 use anyhow::{Context as _, Result, anyhow, bail};
 use bstr::{BStr, BString};
-use but_core::{RepositoryExt as RepositoryExtGix, commit::Headers};
+use but_core::{
+    RepositoryExt as RepositoryExtGix,
+    commit::{Headers, SignCommit},
+};
 use but_error::Code;
 use but_oxidize::{
     ObjectIdExt as _, OidExt, git2_signature_to_gix_signature, gix_to_git2_signature,
@@ -64,7 +67,7 @@ fn commit_gix(
     tree: gix::ObjectId,
     parents: &[gix::ObjectId],
     commit_headers: Option<Headers>,
-    sign_if_configured: bool,
+    sign_commit: SignCommit,
 ) -> Result<gix::ObjectId> {
     let mut commit = gix::objs::Commit {
         message: message.into(),
@@ -87,7 +90,7 @@ fn commit_gix(
         repo,
         commit,
         update_ref.as_ref().map(|name| name.as_ref()),
-        sign_if_configured,
+        sign_commit,
     )
 }
 
@@ -112,7 +115,7 @@ pub fn commit_with_signature_gix(
         tree,
         parents,
         commit_headers,
-        true,
+        SignCommit::IfSignCommitsEnabled,
     )
 }
 
@@ -137,7 +140,7 @@ pub fn commit_without_signature_gix(
         tree,
         parents,
         commit_headers,
-        false,
+        SignCommit::No,
     )
 }
 


crates/but-rebase/src/graph_rebase/commit.rs
```diff
@@ -74,9 +75,12 @@ impl<M: RefMetadata> Editor<'_, '_, M> {
         commit: but_core::CommitOwned,
         date_mode: DateMode,
     ) -> Result<gix::ObjectId> {
-        // TODO(GB-983): As part of moving to only signing at the materializing
-        // step, this should have sign_if_configured false here.
```
slarse, line 78: With moving to sign-on-push, I don't think it's relevant to pursue "sign on materialize" anymore.

crates/but-rebase/src/graph_rebase/cherry_pick.rs
```diff
@@ -35,6 +35,48 @@ pub enum CherryPickOutcome {
...
+/// Note that for all sign modes but [`Self::Never`], signing a parent commit only causes
+/// descendants to be signed if they are also picked with an applicable [`PickSignMode`].
+/// For example, in a commit chain A <- B, where A is the root, picking A with
+/// [`PickSignMode::Force`] and B with [`PickSignMode::Never`] will cause A to be signed, but not
+/// B. However, if B is picked with [`PickSignMode::IfChanged`] in the same scenario, then B will
+/// get signed as its parent A has changed from the signature being applied to it.
+///
+/// An intentional pattern to cause a cascade of signatures is to make all picks with
+/// [`PickSignMode::IfChanged`], and then re-pick the intended root(s) of the signature cascade
+/// with [`PickSignMode::Force`].
```
slarse: This is the approach by which the sign-on-push functionality is to be implemented.

crates/but-workspace/tests/workspace/commit/commits_sign.rs
```diff
@@ -0,0 +1,222 @@
...
+    assert_eq!(unsigned.unsigned_commits.len(), 3);
+
+    let editor = Editor::create_with_opts(
+        &mut ws,
+        &mut meta,
+        &repo,
+        &GraphEditorOptions {
+            sign_mode: PickSignMode::IfChanged,
+        },
+    )?;
```
slarse: These tests somewhat by happenstance also test the new "custom" `sign_mode` functionality of the editor, as it's in the scenario of signing commits that it makes sense to make use of it. It would probably be prudent to have more tests for these over in the editor tests themselves, but I just haven't had the time to create them and these tests cover the important parts.

Note that you could create an editor with `PickSignMode::Force`, and it would forcibly sign every commit in the workspace just like that, but it's not a very useful thing to do so I haven't spent any time testing it.

crates/but-core/src/commit.rs
```diff
@@ -175,26 +189,28 @@ pub fn create(
...
+                        })?;
+                        return Err(anyhow!("Failed to sign commit: {err}")
+                            .context(Code::CommitSigningFailed));
+                    } else {
+                        tracing::warn!(
+                            "Commit signing failed but remains enabled as gitbutler.signCommits is explicitly enabled globally"
+                        );
+                    }
                 }
+                return Err(err);
```
slarse: Note that we do not fiddle with settings in the new code - this really isn't the place to fiddle with settings. Settings-fiddling only retained for backwards compatibility with the legacy sign behavior.

crates/but-rebase/src/graph_rebase/cherry_pick.rs
```diff
@@ -56,11 +98,11 @@ pub fn cherry_pick(
     repo: &gix::Repository,
     target: gix::ObjectId,
     ontos: &[gix::ObjectId],
-    sign_if_configured: bool,
+    sign_mode: PickSignMode,
 ) -> Result<CherryPickOutcome> {
     let target = but_core::Commit::from_id(target.attach(repo))?;
 
-    if ontos == target.parents.as_slice() {
+    if ontos == target.parents.as_slice() && sign_mode != PickSignMode::Force {
```
slarse: This is the _only_ place where `PickSignMode::Force` differs from `PickSignMode::IfChanged`.

crates/but-workspace/src/commit/commits_list_unsigned.rs
```diff
@@ -0,0 +1,49 @@
...
+    use but_core::ref_metadata::StackId;
+    use but_graph::projection::Workspace;
+    use gix::{ObjectId, hashtable::HashSet};
+
+    /// The result of listing unsigned commits in the workspace.
+    #[derive(Debug)]
+    pub struct CommitsListUnsignedOutcome {
+        /// IDs of all unsigned commits in the workspace
+        pub unsigned_commits: HashSet<ObjectId>,
+    }
```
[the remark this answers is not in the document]
slarse: I feel like this is fine. This is an efficient HashSet implementation for already hashed things, which is something we have a lot of in GitButler.

crates/but-core/src/commit.rs
```diff
@@ -148,6 +148,17 @@ impl From<&Headers> for Vec<(BString, BString)> {
...
+/// Determines how to sign the commit.
+#[derive(PartialEq, Copy, Clone, Debug)]
+pub enum SignCommit {
+    /// Sign
+    Yes,
+    /// Do not sign
+    No,
+    /// **Deprecated:** Legacy behavior, signs the commit if gitbutler.signCommits=true
+    LegacyIfSignCommitsEnabled,
+}
```
slarse, line 176: This can be reduced to a boolean once the legacy behavior is gone. I don't think we should be concerned with "different modes" this deep down, at this point the decision to sign or not should already have been made.

crates/but-rebase/src/graph_rebase/commit.rs
Caleb-T-Owens: Agreed - this solves the same problem set

Byron (changes requested):
Thanks, I really wanted to approve it but ultimately couldn't. I looked at `but-core`, and `but-workspace` in detail, but also glimpsed at `but-rebase`. 
What I'd want to be merged immediately is the improvements to the signing logic via `SignCommits`, as the enum offers a place to spell out behaviour previously hidden behind `true/false`. From that point of view it would have made sense to split the PR into distinct independent commits as it's possible.

Here is patch with changes that I hope you can apply. I would have pushed it into the branch, but since it's a GitButler stack I think it would be quite unpleasant to deal with right now.

My suggestion is to turn this PR into something without `but-rebase` logic changes (but with the new enum), and with `commit_signed` so it can be merged. More details follow.

```patch
commit a290f4c1df51a1e3241f5283f72595610374c15c
Date:   Tue Mar 24 12:02:53 2026 +0800

    review

diff --git a/crates/but-core/src/commit.rs b/crates/but-core/src/commit.rs
index 8648c869ab..d024e685a6 100644
--- a/crates/but-core/src/commit.rs
+++ b/crates/but-core/src/commit.rs
@@ -149,20 +149,29 @@ impl From<&Headers> for Vec<(BString, BString)> {
 }
 
 /// Determines how to sign the commit.
-#[derive(PartialEq, Copy, Clone, Debug)]
+#[derive(Default, PartialEq, Copy, Clone, Debug)]
 pub enum SignCommit {
-    /// Sign
+    /// Sign the commit.
     Yes,
-    /// Do not sign
+    /// Do not sign the commit.
     No,
-    /// **Deprecated:** Legacy behavior, signs the commit if gitbutler.signCommits=true
-    LegacyIfSignCommitsEnabled,
+    /// Signs the commit if `gitbutler.signCommits=true` *or* if `commit.gpgSign=true`.
+    /// or do *not* sign if neither of these are set.
+    /// `gitbutler.signCommits` takes precedence, so if it is set, it overrules  `commit.gpgSign`.
+    /// The reason for this is that some clients cannot ensure that they can execute GPG/SSH in such
+    /// a way that it will work correctly or be in the `PATH` at all, not to talk about the various
+    /// ways a password for a key can be requested.
+    /// This is the safest option, as the other options represent a programmatic override of
+    /// user preferences.
+    #[default]
+    IfSignCommitsEnabled,
 }
 
 /// Write `commit` into `repo`, removing any existing commit signature first, optionally creating a
 /// new one based on repository configuration, and optionally updating `update_ref` to the new ID.
 ///
 /// Apply any desired message/header mutations, such as Gerrit trailers, before calling this helper.
+/// `sign_commit` defines under which [circumstances](SignCommit) a signature for `commit` should be created.
 pub fn create(
     repo: &gix::Repository,
     mut commit: gix::objs::Commit,
@@ -176,7 +185,7 @@ pub fn create(
         commit.extra_headers.remove(pos);
     }
 
-    if (sign_commit == SignCommit::LegacyIfSignCommitsEnabled
+    if (sign_commit == SignCommit::IfSignCommitsEnabled
         && repo.git_settings()?.gitbutler_sign_commits.unwrap_or(false))
         || sign_commit == SignCommit::Yes
     {
@@ -190,7 +199,7 @@ pub fn create(
             }
             Err(err) => {
                 tracing::warn!("Commit signing failed with sign_commit={sign_commit:?}");
-                if sign_commit == SignCommit::LegacyIfSignCommitsEnabled {
+                if sign_commit == SignCommit::IfSignCommitsEnabled {
                     if repo
                         .config_snapshot()
                         .boolean_filter("gitbutler.signCommits", |md| {
diff --git a/crates/but-core/tests/core/commit.rs b/crates/but-core/tests/core/commit.rs
index b366b60977..5faef42900 100644
--- a/crates/but-core/tests/core/commit.rs
+++ b/crates/but-core/tests/core/commit.rs
@@ -174,7 +174,7 @@ mod create {
             &repo,
             commit_from_head(&repo, "signed again")?,
             None,
-            SignCommit::LegacyIfSignCommitsEnabled,
+            SignCommit::IfSignCommitsEnabled,
         )?;
         let commit = repo.find_commit(oid)?;
         let commit = commit.decode()?;
@@ -186,6 +186,30 @@ mod create {
         Ok(())
     }
 
+    #[test]
+    fn forcefully_signs_commits() -> anyhow::Result<()> {
+        let (mut repo, _tmp) = writable_scenario_with_ssh_key("single-signed");
+        repo.config_snapshot_mut()
+            .set_raw_value(&"gitbutler.signCommits", "false")?;
+        repo.config_snapshot_mut()
+            .set_raw_value(&"commit.gpgSign", "false")?;
+
+        let oid = commit::create(
+            &repo,
+            commit_from_head(&repo, "should not sign")?,
+            None,
+            SignCommit::Yes,
+        )?;
+        let commit = repo.find_commit(oid)?;
+        let commit = commit.decode()?;
+
+        assert!(
+            commit.extra_headers().find(SIGNATURE_FIELD_NAME).is_some(),
+            "despite the settings saying everything is disabled, it still signs the commit"
+        );
+        Ok(())
+    }
+
     #[test]
     fn removes_existing_signature_when_not_signing() -> anyhow::Result<()> {
         let (repo, _tmp) = writable_scenario_with_ssh_key("single-signed");
diff --git a/crates/but-rebase/src/cherry_pick.rs b/crates/but-rebase/src/cherry_pick.rs
index 060a1f0b58..c38ca99e3d 100644
--- a/crates/but-rebase/src/cherry_pick.rs
+++ b/crates/but-rebase/src/cherry_pick.rs
@@ -192,7 +192,7 @@ pub(crate) mod function {
             repo,
             new_commit,
             DateMode::CommitterUpdateAuthorKeep,
-            SignCommit::LegacyIfSignCommitsEnabled,
+            SignCommit::IfSignCommitsEnabled,
         )?
         .attach(repo))
     }
@@ -257,7 +257,7 @@ pub(crate) mod function {
             repo,
             to_rebase.inner,
             DateMode::CommitterUpdateAuthorKeep,
-            SignCommit::LegacyIfSignCommitsEnabled,
+            SignCommit::IfSignCommitsEnabled,
         )?
         .attach(repo))
     }
diff --git a/crates/but-rebase/src/graph_rebase/cherry_pick.rs b/crates/but-rebase/src/graph_rebase/cherry_pick.rs
index 4f2c8d85cd..28aab79b2c 100644
--- a/crates/but-rebase/src/graph_rebase/cherry_pick.rs
+++ b/crates/but-rebase/src/graph_rebase/cherry_pick.rs
@@ -51,13 +51,11 @@ pub enum CherryPickOutcome {
 pub enum PickSignMode {
     /// The picked commit is never signed.
     Never,
-    /// **Deprecated:** This is the legacy behavior, where a commit is signed if
-    /// gitbutler.signCommits is enabled. It's here as a compatibility mode only and will be
-    /// removed in the future.
+    /// A commit is signed according to the logic of [`SignCommit::IfSignCommitsEnabled`].
     ///
     /// This mode is generally unwieldy to use as it causes us to sign commits _all the time_ when
     /// rebasing.
-    LegacyIfSignCommitsEnabled,
+    IfSignCommitsEnabled,
     /// Signs the commit only if something has changed. If the Pick is otherwise a noop, this has
     /// no effect, even if the picked commit is currently unsigned.
     IfChanged,
@@ -72,7 +70,7 @@ impl From<PickSignMode> for SignCommit {
         match value {
             PickSignMode::Never => SignCommit::No,
             PickSignMode::Force | PickSignMode::IfChanged => SignCommit::Yes,
-            PickSignMode::LegacyIfSignCommitsEnabled => SignCommit::LegacyIfSignCommitsEnabled,
+            PickSignMode::IfSignCommitsEnabled => SignCommit::IfSignCommitsEnabled,
         }
     }
 }
diff --git a/crates/but-rebase/src/graph_rebase/commit.rs b/crates/but-rebase/src/graph_rebase/commit.rs
index 0220086d90..ecc61ae517 100644
--- a/crates/but-rebase/src/graph_rebase/commit.rs
+++ b/crates/but-rebase/src/graph_rebase/commit.rs
@@ -79,7 +79,7 @@ impl<M: RefMetadata> Editor<'_, '_, M> {
             &self.repo,
             commit.inner,
             date_mode,
-            SignCommit::LegacyIfSignCommitsEnabled,
+            SignCommit::IfSignCommitsEnabled,
         )
     }
 
diff --git a/crates/but-rebase/src/graph_rebase/creation.rs b/crates/but-rebase/src/graph_rebase/creation.rs
index 354ed1696e..e498c34f12 100644
--- a/crates/but-rebase/src/graph_rebase/creation.rs
+++ b/crates/but-rebase/src/graph_rebase/creation.rs
@@ -20,7 +20,7 @@ pub struct GraphEditorOptions {
 impl Default for GraphEditorOptions {
     fn default() -> Self {
         Self {
-            sign_mode: PickSignMode::LegacyIfSignCommitsEnabled,
+            sign_mode: PickSignMode::IfSignCommitsEnabled,
         }
     }
 }
diff --git a/crates/but-rebase/src/graph_rebase/mod.rs b/crates/but-rebase/src/graph_rebase/mod.rs
index 359ca33e6c..672611bce9 100644
--- a/crates/but-rebase/src/graph_rebase/mod.rs
+++ b/crates/but-rebase/src/graph_rebase/mod.rs
@@ -58,7 +58,7 @@ impl Pick {
             preserved_parents: None,
             conflictable: true,
             parents_must_be_references: false,
-            sign_mode: PickSignMode::LegacyIfSignCommitsEnabled,
+            sign_mode: PickSignMode::IfSignCommitsEnabled,
             exclude_from_tracking: false,
         }
     }
@@ -395,7 +395,7 @@ mod test {
                 preserved_parents: None,
                 conflictable: true,
                 parents_must_be_references: false,
-                sign_mode: PickSignMode::LegacyIfSignCommitsEnabled,
+                sign_mode: PickSignMode::IfSignCommitsEnabled,
                 exclude_from_tracking: false
             }
         );
diff --git a/crates/but-rebase/src/lib.rs b/crates/but-rebase/src/lib.rs
index 917691d5ce..13e61ce297 100644
--- a/crates/but-rebase/src/lib.rs
+++ b/crates/but-rebase/src/lib.rs
@@ -295,7 +295,7 @@ fn rebase(
                                 repo,
                                 new_commit,
                                 DateMode::CommitterUpdateAuthorKeep,
-                                SignCommit::LegacyIfSignCommitsEnabled,
+                                SignCommit::IfSignCommitsEnabled,
                             )?);
                         }
                         None => {
@@ -334,7 +334,7 @@ fn rebase(
                     repo,
                     new_commit,
                     DateMode::CommitterUpdateAuthorKeep,
-                    SignCommit::LegacyIfSignCommitsEnabled,
+                    SignCommit::IfSignCommitsEnabled,
                 )?;
             }
             RebaseStep::Reference(reference) => {
@@ -379,7 +379,7 @@ fn reword_commit(
         repo,
         new_commit,
         DateMode::CommitterUpdateAuthorKeep,
-        SignCommit::LegacyIfSignCommitsEnabled,
+        SignCommit::IfSignCommitsEnabled,
     )?)
 }
 
@@ -395,7 +395,7 @@ pub fn replace_commit_tree(
         repo,
         new_commit,
         DateMode::CommitterUpdateAuthorKeep,
-        SignCommit::LegacyIfSignCommitsEnabled,
+        SignCommit::IfSignCommitsEnabled,
     )?)
 }
 
diff --git a/crates/but-rebase/src/merge.rs b/crates/but-rebase/src/merge.rs
index 26e98706f4..b987077308 100644
--- a/crates/but-rebase/src/merge.rs
+++ b/crates/but-rebase/src/merge.rs
@@ -98,7 +98,7 @@ pub fn octopus(
             repo,
             target_merge_commit,
             DateMode::CommitterUpdateAuthorKeep,
-            SignCommit::LegacyIfSignCommitsEnabled,
+            SignCommit::IfSignCommitsEnabled,
         )
     } else {
         crate::commit::update_committer(repo, &mut target_merge_commit)?;
diff --git a/crates/but-rebase/tests/rebase/graph_rebase/cherry_pick.rs b/crates/but-rebase/tests/rebase/graph_rebase/cherry_pick.rs
index fb7c4cdf77..b06643dda5 100644
--- a/crates/but-rebase/tests/rebase/graph_rebase/cherry_pick.rs
+++ b/crates/but-rebase/tests/rebase/graph_rebase/cherry_pick.rs
@@ -21,12 +21,7 @@ fn basic_cherry_pick_clean() -> Result<()> {
     let target = repo.rev_parse_single("single-clean-commit")?.detach();
     let onto = repo.rev_parse_single("single-target")?.detach();
 
-    let result = cherry_pick(
-        &repo,
-        target,
-        &[onto],
-        PickSignMode::LegacyIfSignCommitsEnabled,
-    )?;
+    let result = cherry_pick(&repo, target, &[onto], PickSignMode::IfSignCommitsEnabled)?;
 
     insta::assert_debug_snapshot!(result, @"
     Commit(
@@ -57,12 +52,7 @@ fn basic_cherry_pick_cp_conflicts() -> Result<()> {
     let target = repo.rev_parse_single("single-conflicting-commit")?.detach();
     let onto = repo.rev_parse_single("single-target")?.detach();
 
-    let result = cherry_pick(
-        &repo,
-        target,
-        &[onto],
-        PickSignMode::LegacyIfSignCommitsEnabled,
-    )?;
+    let result = cherry_pick(&repo, target, &[onto], PickSignMode::IfSignCommitsEnabled)?;
 
     insta::assert_debug_snapshot!(result, @"
     ConflictedCommit(
@@ -110,7 +100,7 @@ fn basic_cherry_pick_identity() -> Result<()> {
         &repo,
         target.detach(),
         &parents,
-        PickSignMode::LegacyIfSignCommitsEnabled,
+        PickSignMode::IfSignCommitsEnabled,
     )?;
 
     insta::assert_debug_snapshot!(result, @"
@@ -134,7 +124,7 @@ fn single_parent_to_multiple_parents_clean() -> Result<()> {
         &repo,
         target,
         &[onto, onto2],
-        PickSignMode::LegacyIfSignCommitsEnabled,
+        PickSignMode::IfSignCommitsEnabled,
     )?;
 
     insta::assert_debug_snapshot!(result, @"
@@ -173,7 +163,7 @@ fn single_parent_to_multiple_parents_cp_conflicts() -> Result<()> {
         &repo,
         target,
         &[onto, onto2],
-        PickSignMode::LegacyIfSignCommitsEnabled,
+        PickSignMode::IfSignCommitsEnabled,
     )?;
 
     insta::assert_debug_snapshot!(result, @"
@@ -228,7 +218,7 @@ fn single_parent_to_multiple_parents_parents_conflict() -> Result<()> {
         &repo,
         target,
         &[onto, onto2],
-        PickSignMode::LegacyIfSignCommitsEnabled,
+        PickSignMode::IfSignCommitsEnabled,
     )?;
 
     insta::assert_debug_snapshot!(result, @"
@@ -256,12 +246,7 @@ fn multiple_parents_to_single_parent_clean() -> Result<()> {
     let target = repo.rev_parse_single("merge-clean-commit")?.detach();
     let onto = repo.rev_parse_single("single-target")?.detach();
 
-    let result = cherry_pick(
-        &repo,
-        target,
-        &[onto],
-        PickSignMode::LegacyIfSignCommitsEnabled,
-    )?;
+    let result = cherry_pick(&repo, target, &[onto], PickSignMode::IfSignCommitsEnabled)?;
 
     insta::assert_debug_snapshot!(result, @"
     Commit(
@@ -293,12 +278,7 @@ fn multiple_parents_to_single_parent_cp_conflicts() -> Result<()> {
     let target = repo.rev_parse_single("merge-conflicting-commit")?.detach();
     let onto = repo.rev_parse_single("single-target")?.detach();
 
-    let result = cherry_pick(
-        &repo,
-        target,
-        &[onto],
-        PickSignMode::LegacyIfSignCommitsEnabled,
-    )?;
+    let result = cherry_pick(&repo, target, &[onto], PickSignMode::IfSignCommitsEnabled)?;
 
     insta::assert_debug_snapshot!(result, @"
     ConflictedCommit(
@@ -348,12 +328,7 @@ fn multiple_parents_to_single_parent_parents_conflict() -> Result<()> {
         .detach();
     let onto = repo.rev_parse_single("single-target")?.detach();
 
-    let result = cherry_pick(
-        &repo,
-        target,
-        &[onto],
-        PickSignMode::LegacyIfSignCommitsEnabled,
-    )?;
+    let result = cherry_pick(&repo, target, &[onto], PickSignMode::IfSignCommitsEnabled)?;
 
     insta::assert_debug_snapshot!(result, @"
     FailedToMergeBases {
@@ -385,7 +360,7 @@ fn multiple_parents_to_multiple_parents_clean() -> Result<()> {
         &repo,
         target,
         &[onto, onto2],
-        PickSignMode::LegacyIfSignCommitsEnabled,
+        PickSignMode::IfSignCommitsEnabled,
     )?;
 
     insta::assert_debug_snapshot!(result, @"
@@ -424,7 +399,7 @@ fn multiple_parents_to_multiple_parents_cp_conflicts() -> Result<()> {
         &repo,
         target,
         &[onto, onto2],
-        PickSignMode::LegacyIfSignCommitsEnabled,
+        PickSignMode::IfSignCommitsEnabled,
     )?;
 
     insta::assert_debug_snapshot!(result, @"
@@ -483,7 +458,7 @@ fn multiple_parents_to_multiple_parents_base_parents_conflict() -> Result<()> {
         &repo,
         target,
         &[onto, onto2],
-        PickSignMode::LegacyIfSignCommitsEnabled,
+        PickSignMode::IfSignCommitsEnabled,
     )?;
 
     insta::assert_debug_snapshot!(result, @"
@@ -515,7 +490,7 @@ fn multiple_parents_to_multiple_parents_target_parents_conflict() -> Result<()>
         &repo,
         target,
         &[onto, onto2],
-        PickSignMode::LegacyIfSignCommitsEnabled,
+        PickSignMode::IfSignCommitsEnabled,
     )?;
 
     insta::assert_debug_snapshot!(result, @"
@@ -547,7 +522,7 @@ fn multiple_parents_to_multiple_parents_identity() -> Result<()> {
         &repo,
         target.detach(),
         &parents,
-        PickSignMode::LegacyIfSignCommitsEnabled,
+        PickSignMode::IfSignCommitsEnabled,
     )?;
 
     insta::assert_debug_snapshot!(result, @"
@@ -570,7 +545,7 @@ fn no_parents_identity() -> Result<()> {
         &repo,
         target.detach(),
         &[],
-        PickSignMode::LegacyIfSignCommitsEnabled,
+        PickSignMode::IfSignCommitsEnabled,
     )?;
 
     insta::assert_debug_snapshot!(result, @"
@@ -589,7 +564,7 @@ fn single_parent_to_no_parents_clean() -> Result<()> {
 
     let target = repo.rev_parse_single("single-clean-commit")?.detach();
 
-    let result = cherry_pick(&repo, target, &[], PickSignMode::LegacyIfSignCommitsEnabled)?;
+    let result = cherry_pick(&repo, target, &[], PickSignMode::IfSignCommitsEnabled)?;
 
     insta::assert_debug_snapshot!(result, @"
     Commit(
@@ -619,12 +594,7 @@ fn no_parents_to_single_parent_clean() -> Result<()> {
     let target = repo.rev_parse_single("base")?.detach();
     let onto = repo.rev_parse_single("single-target")?.detach();
 
-    let result = cherry_pick(
-        &repo,
-        target,
-        &[onto],
-        PickSignMode::LegacyIfSignCommitsEnabled,
-    )?;
+    let result = cherry_pick(&repo, target, &[onto], PickSignMode::IfSignCommitsEnabled)?;
 
     insta::assert_debug_snapshot!(result, @"
     Commit(
@@ -655,12 +625,7 @@ fn no_parents_to_single_parent_cp_conflicts() -> Result<()> {
     let target = repo.rev_parse_single("base-conflicting")?.detach();
     let onto = repo.rev_parse_single("single-target")?.detach();
 
-    let result = cherry_pick(
-        &repo,
-        target,
-        &[onto],
-        PickSignMode::LegacyIfSignCommitsEnabled,
-    )?;
+    let result = cherry_pick(&repo, target, &[onto], PickSignMode::IfSignCommitsEnabled)?;
 
     insta::assert_debug_snapshot!(result, @"
     ConflictedCommit(
@@ -708,7 +673,7 @@ fn cherry_pick_back_to_original_parents_unconflicts() -> Result<()> {
         &repo,
         target.detach(),
         &[onto, onto2],
-        PickSignMode::LegacyIfSignCommitsEnabled,
+        PickSignMode::IfSignCommitsEnabled,
     )?;
 
     insta::assert_debug_snapshot!(result, @"
@@ -723,12 +688,7 @@ fn cherry_pick_back_to_original_parents_unconflicts() -> Result<()> {
 
     assert_eq!(&get_parents(&id.attach(&repo))?, &[onto, onto2]);
 
-    let result = cherry_pick(
-        &repo,
-        id,
-        &parents,
-        PickSignMode::LegacyIfSignCommitsEnabled,
-    )?;
+    let result = cherry_pick(&repo, id, &parents, PickSignMode::IfSignCommitsEnabled)?;
 
     insta::assert_debug_snapshot!(result, @"
     Commit(
@@ -777,7 +737,7 @@ fn cherry_pick_recursive_merge() -> Result<()> {
         &repo,
         target.detach(),
         &[onto, onto2, onto3],
-        PickSignMode::LegacyIfSignCommitsEnabled,
+        PickSignMode::IfSignCommitsEnabled,
     )?;
 
     insta::assert_debug_snapshot!(result, @"
diff --git a/crates/but-workspace/src/commit/commits_list_unsigned.rs b/crates/but-workspace/src/commit/commits_list_unsigned.rs
deleted file mode 100644
index 2e1d418f84..0000000000
--- a/crates/but-workspace/src/commit/commits_list_unsigned.rs
+++ /dev/null
@@ -1,49 +0,0 @@
-//! An action to list unsigned commit IDs in the workspace.
-
-pub(crate) mod function {
-    use anyhow::Result;
-    use but_core::ref_metadata::StackId;
-    use but_graph::projection::Workspace;
-    use gix::{ObjectId, hashtable::HashSet};
-
-    /// The result of listing unsigned commits in the workspace.
-    #[derive(Debug)]
-    pub struct CommitsListUnsignedOutcome {
-        /// IDs of all unsigned commits in the workspace
-        pub unsigned_commits: HashSet<ObjectId>,
-    }
-
-    /// List unsigned commits in the workspace.
-    ///
-    /// `ws` - the workspace to operate in.
-    ///
-    /// `repo` - associated repository so signature information can be extracted for commits.
-    ///
-    /// `stack_id` - if provided, only commits for this stack are fetched.
-    pub fn commits_list_unsigned(
-        ws: &Workspace,
-        repo: &gix::Repository,
-        stack_id: Option<StackId>,
-    ) -> Result<CommitsListUnsignedOutcome> {
-        let commits = ws
-            .stacks
-            .iter()
-            .filter(|stack| stack_id.is_none() || stack_id == stack.id)
-            .flat_map(|stack| stack.segments.iter())
-            .flat_map(|segment| segment.commits.iter());
-
-        let mut unsigned_commits = HashSet::default();
-        for commit in commits {
-            if commit
-                .attach(repo)?
-                .extra_headers()
-                .pgp_signature()
-                .is_none()
-            {
-                unsigned_commits.insert(commit.id);
-            }
-        }
-
-        Ok(CommitsListUnsignedOutcome { unsigned_commits })
-    }
-}
diff --git a/crates/but-workspace/src/commit/mod.rs b/crates/but-workspace/src/commit/mod.rs
index 98c1cca15d..5e8d2535b1 100644
--- a/crates/but-workspace/src/commit/mod.rs
+++ b/crates/but-workspace/src/commit/mod.rs
@@ -12,8 +12,6 @@ pub mod commit_amend;
 pub use commit_amend::function::{CommitAmendOutcome, commit_amend};
 pub mod commits_sign;
 pub use commits_sign::function::{CommitsSignOutcome, commits_sign};
-pub mod commits_list_unsigned;
-pub use commits_list_unsigned::function::{CommitsListUnsignedOutcome, commits_list_unsigned};
 pub mod insert_blank_commit;
 pub use insert_blank_commit::function::insert_blank_commit;
 pub mod move_changes;
diff --git a/crates/but-workspace/src/commit_engine/mod.rs b/crates/but-workspace/src/commit_engine/mod.rs
index 86d3313ec2..e76650f4f9 100644
--- a/crates/but-workspace/src/commit_engine/mod.rs
+++ b/crates/but-workspace/src/commit_engine/mod.rs
@@ -195,7 +195,7 @@ pub fn create_commit(
                     repo,
                     commit.inner,
                     DateMode::CommitterUpdateAuthorKeep,
-                    SignCommit::LegacyIfSignCommitsEnabled,
+                    SignCommit::IfSignCommitsEnabled,
                 )?)
             }
         }
@@ -237,6 +237,6 @@ fn create_possibly_signed_commit(
         repo,
         commit,
         DateMode::CommitterKeepAuthorKeep,
-        SignCommit::LegacyIfSignCommitsEnabled,
+        SignCommit::IfSignCommitsEnabled,
     )
 }
diff --git a/crates/but-workspace/tests/fixtures/scenario/ws-with-mixed-signed-unsigned.sh b/crates/but-workspace/tests/fixtures/scenario/ws-with-mixed-signed-unsigned.sh
index 4bcf2ff25b..d852124b9b 100644
--- a/crates/but-workspace/tests/fixtures/scenario/ws-with-mixed-signed-unsigned.sh
+++ b/crates/but-workspace/tests/fixtures/scenario/ws-with-mixed-signed-unsigned.sh
@@ -6,7 +6,7 @@
 # Stack unsigned-*: A two-branch stack with all commits unsigned.
 # Stack mixed: A two-branch stack with a mix of signed and unsigned commits.
 #
-# This is useful for testing commits_list_unsigned and commits_sign.
+# This is useful for testing commits_sign.
 
 set -eu -o pipefail
 
diff --git a/crates/but-workspace/tests/workspace/commit/commits_list_unsigned.rs b/crates/but-workspace/tests/workspace/commit/commits_list_unsigned.rs
deleted file mode 100644
index dd87eaa5c2..0000000000
--- a/crates/but-workspace/tests/workspace/commit/commits_list_unsigned.rs
+++ /dev/null
@@ -1,116 +0,0 @@
-use anyhow::Result;
-use but_workspace::commit::commits_list_unsigned;
-
-use crate::ref_info::with_workspace_commit::utils::{
-    StackState, add_stack_with_segments, named_writable_scenario_with_description_and_graph,
-};
-
-#[test]
-fn filter_by_stack_id() -> Result<()> {
-    let (_tmp, graph, repo, _meta, _description) =
-        named_writable_scenario_with_description_and_graph(
-            "ws-with-mixed-signed-unsigned",
-            |meta| {
-                add_stack_with_segments(
-                    meta,
-                    1,
-                    "unsigned-top",
-                    StackState::InWorkspace,
-                    &["unsigned-bottom"],
-                );
-                add_stack_with_segments(
-                    meta,
-                    2,
-                    "mixed-top",
-                    StackState::InWorkspace,
-                    &["mixed-bottom"],
-                );
-            },
-        )?;
-
-    let unsigned_stack_id = gitbutler_stack::StackId::from_number_for_testing(1);
-    let mixed_stack_id = gitbutler_stack::StackId::from_number_for_testing(2);
-
-    let ws = graph.into_workspace()?;
-    let outcome_unsigned_stack = commits_list_unsigned(&ws, &repo, Some(unsigned_stack_id))?;
-    let outcome_mixed_stack = commits_list_unsigned(&ws, &repo, Some(mixed_stack_id))?;
-
-    assert_eq!(
-        outcome_unsigned_stack.unsigned_commits.len(),
-        2,
-        "expected 2 unsigned commits for unsigned stack, got {}: {:?}",
-        outcome_unsigned_stack.unsigned_commits.len(),
-        outcome_unsigned_stack.unsigned_commits,
-    );
-
-    assert_eq!(
-        outcome_mixed_stack.unsigned_commits.len(),
-        1,
-        "expected 1 unsigned commits for mixed stack, got {}: {:?}",
-        outcome_mixed_stack.unsigned_commits.len(),
-        outcome_mixed_stack.unsigned_commits,
-    );
-
-    Ok(())
-}
-
-#[test]
-fn all_unsigned_commits_are_listed() -> Result<()> {
-    let (_tmp, graph, repo, _meta, _description) =
-        named_writable_scenario_with_description_and_graph(
-            "ws-with-mixed-signed-unsigned",
-            |meta| {
-                add_stack_with_segments(
-                    meta,
-                    1,
-                    "unsigned-top",
-                    StackState::InWorkspace,
-                    &["unsigned-bottom"],
-                );
-                add_stack_with_segments(
-                    meta,
-                    2,
-                    "mixed-top",
-                    StackState::InWorkspace,
-                    &["mixed-bottom"],
-                );
-            },
-        )?;
-
-    let ws = graph.into_workspace()?;
-    let outcome = commits_list_unsigned(&ws, &repo, None)?;
-
-    assert_eq!(
-        outcome.unsigned_commits.len(),
-        3,
-        "expected 3 unsigned commits (signed mixed excluded), got {}: {:?}",
-        outcome.unsigned_commits.len(),
-        outcome.unsigned_commits
-    );
-
-    let unsigned_bottom = repo.rev_parse_single("unsigned-bottom")?.detach();
-    let unsigned_top = repo.rev_parse_single("unsigned-top")?.detach();
-    let mixed_bottom = repo.rev_parse_single("mixed-bottom")?.detach();
-    assert!(
-        outcome.unsigned_commits.contains(&unsigned_bottom),
-        "unsigned-bottom commit should be unsigned"
-    );
-    assert!(
-        outcome.unsigned_commits.contains(&unsigned_top),
-        "unsigned-top commit should be unsigned"
-    );
-    assert!(
-        outcome.unsigned_commits.contains(&mixed_bottom),
-        "mixed bottom commit should be unsigned",
-    );
-
-    // just to be on the safe side, we'll also explicitly check that the signed top-commit of the
-    // mixed stack is _not_ in the result set.
-    let mixed_top = repo.rev_parse_single("mixed-top")?.detach();
-    assert!(
-        !outcome.unsigned_commits.contains(&mixed_top),
-        "signed mixed-top commit should not be in the unsigned set"
-    );
-
-    Ok(())
-}
diff --git a/crates/but-workspace/tests/workspace/commit/commits_sign.rs b/crates/but-workspace/tests/workspace/commit/commits_sign.rs
index 10ca56e576..77b0385701 100644
--- a/crates/but-workspace/tests/workspace/commit/commits_sign.rs
+++ b/crates/but-workspace/tests/workspace/commit/commits_sign.rs
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use but_rebase::graph_rebase::{Editor, GraphEditorOptions, cherry_pick::PickSignMode};
-use but_workspace::commit::{commits_list_unsigned, commits_sign};
+use but_workspace::commit::commits_sign;
 
 use crate::ref_info::with_workspace_commit::utils::{
     StackState, add_stack_with_segments,
@@ -44,9 +44,9 @@ fn sign_all_unsigned_commits_in_workspace() -> Result<()> {
         })?;
 
     let mut ws = graph.into_workspace()?;
-
-    let unsigned = commits_list_unsigned(&ws, &repo, None)?;
-    assert_eq!(unsigned.unsigned_commits.len(), 3);
+    let unsigned_bottom = repo.rev_parse_single("unsigned-bottom")?.detach();
+    let unsigned_top = repo.rev_parse_single("unsigned-top")?.detach();
+    let mixed_bottom = repo.rev_parse_single("mixed-bottom")?.detach();
 
     let editor = Editor::create_with_opts(
         &mut ws,
@@ -56,7 +56,7 @@ fn sign_all_unsigned_commits_in_workspace() -> Result<()> {
             sign_mode: PickSignMode::IfChanged,
         },
     )?;
-    let outcome = commits_sign(editor, unsigned.unsigned_commits)?;
+    let outcome = commits_sign(editor, [unsigned_bottom, unsigned_top, mixed_bottom])?;
     let materialize_outcome = outcome.rebase.materialize()?;
 
     let workspace_commit_id = repo.head_id()?.detach();
@@ -72,16 +72,8 @@ fn sign_all_unsigned_commits_in_workspace() -> Result<()> {
         .map(|new_id| commit_is_signed(&repo, *new_id) as i32)
         .sum();
 
-    let num_unsigned_commits_in_ws = commits_list_unsigned(&ws, &repo, None)?
-        .unsigned_commits
-        .len();
-
     assert!(!commit_is_signed(&repo, workspace_commit_id));
     assert_eq!(num_signed_commits, 4, "expected all 4 commits to be signed");
-    assert_eq!(
-        num_unsigned_commits_in_ws, 0,
-        "expected no unsigned commits to be left in the workspace"
-    );
 
     Ok(())
 }
@@ -116,9 +108,6 @@ fn sign_entire_stack_by_signing_only_oldest_unsigned_commit() -> Result<()> {
                 &["mixed-bottom"],
             );
         })?;
-    let unsigned_stack_id = gitbutler_stack::StackId::from_number_for_testing(1);
-    let mixed_stack_id = gitbutler_stack::StackId::from_number_for_testing(2);
-
     let mut ws = graph.into_workspace()?;
 
     let unsigned_bottom = repo.rev_parse_single("unsigned-bottom")?.detach();
@@ -146,25 +135,14 @@ fn sign_entire_stack_by_signing_only_oldest_unsigned_commit() -> Result<()> {
         .values()
         .map(|new_id| commit_is_signed(&repo, *new_id) as i32)
         .sum();
-    let num_unsigned_commits_in_unsigned_stack =
-        commits_list_unsigned(&ws, &repo, Some(unsigned_stack_id))?
-            .unsigned_commits
-            .len();
-    let num_unsigned_commits_in_mixed_stack =
-        commits_list_unsigned(&ws, &repo, Some(mixed_stack_id))?
-            .unsigned_commits
-            .len();
+    let mixed_bottom = repo.rev_parse_single("mixed-bottom")?.detach();
 
     assert_eq!(
         num_signed_commits, 2,
         "expected all 2 commits in the unsigned stack to be signed"
     );
-    assert_eq!(
-        num_unsigned_commits_in_unsigned_stack, 0,
-        "expected no unsigned commits in unsigned stack"
-    );
-    assert_eq!(
-        num_unsigned_commits_in_mixed_stack, 1,
+    assert!(
+        !commit_is_signed(&repo, mixed_bottom),
         "expected to still have unsigned commit in mixed stack as we were not signing it",
     );
 
diff --git a/crates/but-workspace/tests/workspace/commit/mod.rs b/crates/but-workspace/tests/workspace/commit/mod.rs
index 752fe2eaf0..505d172399 100644
--- a/crates/but-workspace/tests/workspace/commit/mod.rs
+++ b/crates/but-workspace/tests/workspace/commit/mod.rs
@@ -1,6 +1,5 @@
 mod commit_amend;
 mod commit_create;
-mod commits_list_unsigned;
 mod commits_sign;
 mod insert_blank_commit;
 mod move_changes;
diff --git a/crates/gitbutler-edit-mode/src/lib.rs b/crates/gitbutler-edit-mode/src/lib.rs
index 3d3cf2064a..4c242a072c 100644
--- a/crates/gitbutler-edit-mode/src/lib.rs
+++ b/crates/gitbutler-edit-mode/src/lib.rs
@@ -329,7 +329,7 @@ pub(crate) fn save_and_return_to_workspace(ctx: &Context, perm: &mut RepoExclusi
             ..commit.inner
         },
         but_rebase::commit::DateMode::CommitterUpdateAuthorKeep,
-        SignCommit::LegacyIfSignCommitsEnabled,
+        SignCommit::IfSignCommitsEnabled,
     )?;
 
     editor.replace(target_selector, Step::new_pick(new_commit_oid))?;
diff --git a/crates/gitbutler-repo/src/repository_ext.rs b/crates/gitbutler-repo/src/repository_ext.rs
index e6c8d5c540..1d069dd98b 100644
--- a/crates/gitbutler-repo/src/repository_ext.rs
+++ b/crates/gitbutler-repo/src/repository_ext.rs
@@ -115,7 +115,7 @@ pub fn commit_with_signature_gix(
         tree,
         parents,
         commit_headers,
-        SignCommit::LegacyIfSignCommitsEnabled,
+        SignCommit::IfSignCommitsEnabled,
     )
 }
```

</details>

## Takeaway

This PR effectively deprecates the special  `gitbutler.signCommits` flag, and I am torn. While on the one hand it shouldn't be needed because Git knows when to sign and we just pick that up, on the other hand the GUI doesn't always pickup all the configuration that's needed to do that correctly.

The flag was introduced to have a way to say: Do not sign in GitButler because it doesn't work, and *we* set that flag per repository to prevent GitButler from failing permanently when trying to create commits. So after a failed signing attempt, we set the flag, surface a descriptive error via `.context(Code::CommitSigningFailed));`, and the user can try again.

For that reason, I don't think we can remove that flag or even call it legacy without a better way to solve it.

### `commit_list_unsigned`

I don't know what the plan is for it or how it will be used.
My thinking here is that instead of adding a new special method, rather extract that bit of information directly in `RefInfo`. It already parses the commits and it's trivial to add while universally useful In future, we *might* want to extend this to show more about the signature as well, and this function here just doesn't fit.

*If* it is supposed to be used as input for `commits_sign` I think it's not suitable to get that list using the workspace projection. What has to be done is to get it through the graph so no commits are missed.
I feel so strongly about this as I removed that function entirely 😅.

Edit: After making a round, I realize this was probably the meant to given an initial selection of commits which should be signed, and the rebase engine will take care of signing the right affected commits according to its rules. Nonetheless, I think it's fine to bring that back once it's clear how it's used and we don't have to make a big deal out of a commit traversal, while seeing where this functionality fits in to make it not so special anymore. My intuition is that the rebase engine can help with that more.

### Thoughts on `PickSignMode::LegacyIfSignCommitsEnabled`

Interestingly, if a smarter signing logic is supposed to be implemented, I also think that `but-rebase` will need to play along. However, that logic would also have to respect Git signing settings *or* it's clearly something very special that deserves its own mode. For the same reasons as at the beginning of this text, we can't remove it the `IfSignCommitsEnabled` mode.
Also, it's great to spell out these modes, so that alone would be something to keep and merge right away.

### Auto-review: `but-rebase`

I asked my agent to check for more and it came up with these reasonable sounding points about `but-rebase`.  Posting here in case it's helpful, while thinking that @Caleb-T-Owens should thoroughly review this as well.

<details><summary>Agent review with one high-prio item</summary>

I’m converting the review notes to GitHub-ready markdown with stable blob links pinned to the current `HEAD`, so the line references won’t drift when you paste them into a PR or issue.

Pinned to commit `a290f4c1df51a1e3241f5283f72595610374c15c`.

1. High: [`commits_sign()` replaces the entire `Pick` node instead of mutating only `sign_mode`](https://github.com/gitbutlerapp/gitbutler/blob/a290f4c1df51a1e3241f5283f72595610374c15c/crates/but-workspace/src/commit/commits_sign.rs#L24-L36). That resets `preserved_parents`, `conflictable`, `parents_must_be_references`, and `exclude_from_tracking` back to the defaults from [`Pick::new_pick()` / `Step::new_pick_with_sign_mode()`](https://github.com/gitbutlerapp/gitbutler/blob/a290f4c1df51a1e3241f5283f72595610374c15c/crates/but-rebase/src/graph_rebase/mod.rs#L28-L63) and [`Step::new_pick_with_sign_mode()`](https://github.com/gitbutlerapp/gitbutler/blob/a290f4c1df51a1e3241f5283f72595610374c15c/crates/but-rebase/src/graph_rebase/mod.rs#L103-L114). That is behavior-changing, not cosmetic: partial traversals explicitly preserve the real parents in [`creation.rs`](https://github.com/gitbutlerapp/gitbutler/blob/a290f4c1df51a1e3241f5283f72595610374c15c/crates/but-rebase/src/graph_rebase/creation.rs#L145-L162), and rebase execution consumes `preserved_parents` in [`rebase.rs`](https://github.com/gitbutlerapp/gitbutler/blob/a290f4c1df51a1e3241f5283f72595610374c15c/crates/but-rebase/src/graph_rebase/rebase.rs#L53-L73). As written, signing one of those commits can silently rewrite it onto graph-derived parents instead of its original preserved parents.

2. Medium: [`PickSignMode::Force` is documented as signing even otherwise-noop picks](https://github.com/gitbutlerapp/gitbutler/blob/a290f4c1df51a1e3241f5283f72595610374c15c/crates/but-rebase/src/graph_rebase/cherry_pick.rs#L59-L65), and the first identity fast-path was updated accordingly in [`cherry_pick()`](https://github.com/gitbutlerapp/gitbutler/blob/a290f4c1df51a1e3241f5283f72595610374c15c/crates/but-rebase/src/graph_rebase/cherry_pick.rs#L95-L105). But the later `(MergeOutcome::NoCommit, MergeOutcome::NoCommit)` arm still returns `Identity` unconditionally [here](https://github.com/gitbutlerapp/gitbutler/blob/a290f4c1df51a1e3241f5283f72595610374c15c/crates/but-rebase/src/graph_rebase/cherry_pick.rs#L114-L119). That means a parentless commit selected through [`commits_sign()`](https://github.com/gitbutlerapp/gitbutler/blob/a290f4c1df51a1e3241f5283f72595610374c15c/crates/but-workspace/src/commit/commits_sign.rs#L24-L36) still cannot be force-signed.

3. Medium: `commits_sign()` only cascades signatures if the caller constructs the editor with [`PickSignMode::IfChanged`](https://github.com/gitbutlerapp/gitbutler/blob/a290f4c1df51a1e3241f5283f72595610374c15c/crates/but-rebase/src/graph_rebase/cherry_pick.rs#L59-L64). The default editor path still seeds picks with [`PickSignMode::IfSignCommitsEnabled`](https://github.com/gitbutlerapp/gitbutler/blob/a290f4c1df51a1e3241f5283f72595610374c15c/crates/but-rebase/src/graph_rebase/creation.rs#L20-L24), while [`commits_sign()` itself only force-signs the explicitly selected commits](https://github.com/gitbutlerapp/gitbutler/blob/a290f4c1df51a1e3241f5283f72595610374c15c/crates/but-workspace/src/commit/commits_sign.rs#L28-L31). So if a caller uses `Editor::create(...)` and repo signing is disabled, descendants rewritten because their parent changed will still be recreated unsigned. The current tests avoid that path by explicitly using [`Editor::create_with_opts(... PickSignMode::IfChanged ...)`](https://github.com/gitbutlerapp/gitbutler/blob/a290f4c1df51a1e3241f5283f72595610374c15c/crates/but-workspace/tests/workspace/commit/commits_sign.rs#L51-L59).

</details>

Wow, I clock in at 115 minutes for this review, and am sorry that I don't think it should be merged in its current form.

slarse:
@Byron Thanks so much for the detailed review!

I think this review suffered a fate worse than big PR - it suffered big PR hat still didn't have enough context to the changes. I thought I made a sensible split in the stack to make this make sense standalone but I think I clearly failed (more on that below).

> Thanks, I really wanted to approve it but ultimately couldn't. I looked at `but-core`, and `but-workspace` in detail, but also glimpsed at `but-rebase`. What I'd want to be merged immediately is the improvements to the signing logic via `SignCommits`, as the enum offers a place to spell out behaviour previously hidden behind `true/false`. From that point of view it would have made sense to split the PR into distinct independent commits as it's possible.

I'll try to make that split make sense.

> Here is patch with changes that I hope you can apply. I would have pushed it into the branch, but since it's a GitButler stack I think it would be quite unpleasant to deal with right now.

I will fight tooth and nail to deprecate the `sign if configured` thing. The problem I have with it is that it makes the sign decision too late, at the very "bottom" of the entire application. This makes the upper parts of the application lose control of when signing happens, which is a very inconvenient thing in an app that is heavily structured around re-doing commits.

The key takeaway I had in designing this was that we _can't_ have a boolean for signing - sometimes we want the rebase engine to sign, and sometimes we don't want it to sign. For example, with sign-on-push, we want to sign _only_ if we're just about to push. At all other times, doing all the same things, we don't want to sign. This is not something that we can figure out at the point of creating the commit, we need to resolve that state way further up the stack and make the sign decisions up there.

> This PR effectively deprecates the special `gitbutler.signCommits` flag, and I am torn. While on the one hand it shouldn't be needed because Git knows when to sign and we just pick that up, on the other hand the GUI doesn't always pickup all the configuration that's needed to do that correctly.
>
> For that reason, I don't think we can remove that flag or even call it legacy without a better way to solve it.

This is where I think I excluded important information and should've just had the entire big-ass prototype be the PR. This is not to be removed outright, but replaced. The replacement for the boolean `gitbutler.signCommits` is `gitbutler.signMode`, and it's introduced in the next PR in this stack. I made it a new option instead of co-opting the old one to make it easier to develop the new behavior side by side.

This PR does nothing with settings, it only preemptively names the legacy mode legacy, because I truly don't think "sign all the time if we've set this thing in the config" is useful. But I realize the context for that was completely lost in this PR, and only @Caleb-T-Owens had that from our discussion on how this should work.

The outright deprecation of `sign if configured` is made with the observation that the behavior is kind of useless. There is no need to sign unless we intend to publish the commits (i.e. push). If _that_ part is something you feel strongly about, we should discuss that. There is no technical need to get rid of this behavior, it can coexist happily with the new behavior.

> ### `commit_list_unsigned`
> 
> I don't know what the plan is for it or how it will be used. My thinking here is that instead of adding a new special method, rather extract that bit of information directly in `RefInfo`. It already parses the commits and it's trivial to add while universally useful In future, we _might_ want to extend this to show more about the signature as well, and this function here just doesn't fit.
> 
> _If_ it is supposed to be used as input for `commits_sign` I think it's not suitable to get that list using the workspace projection. What has to be done is to get it through the graph so no commits are missed. I feel so strongly about this as I removed that function entirely 😅.
> 
> Edit: After making a round, I realize this was probably the meant to given an initial selection of commits which should be signed, and the rebase engine will take care of signing the right affected commits according to its rules. Nonetheless, I think it's fine to bring that back once it's clear how it's used and we don't have to make a big deal out of a commit traversal, while seeing where this functionality fits in to make it not so special anymore. My intuition is that the rebase engine can help with that more.

It's just to get unsigned commits as input for `commits_sign`. Putting it in but-workspace was just so the exact same mechanism can be reused elsewhere. I think both `commits_sign` and `commits_list_unsigned` feel like a whole lot of ceremony for barely any logic so I'm all ears for other solutions.

> ### Thoughts on `PickSignMode::LegacyIfSignCommitsEnabled`
> 
> Interestingly, if a smarter signing logic is supposed to be implemented, I also think that `but-rebase` will need to play along. However, that logic would also have to respect Git signing settings

I don't really understand what you mean by this. Do you mean to say that it's an important property of the rebase engine that it, mid rebase, inspects signing settings in Git? I don't agree on that design decision, I think it makes more sense to inspect the settings _before_ starting the rebase, make the decisions about how to sign when building the editor, and then do the rebase with those settings. It _might_ blow up if someone changes the settings at the very same time as the rebase is ongoing, but that race condition exists regardless of where you put the "read" of the settings.

Is I noted above, it's very inconvenient to decide on signing at the point of creating the commit. It makes sign-on-push kind of functionality impossible to implement. The decision of whether a commit is eligible for signing must be made much earlier.

> Wow, I clock in at 115 minutes for this review, and am sorry that I don't think it should be merged in its current form.

I'm sorry for taking up so much of your time :sweat:. It was pretty tough figuring out where to put what to realize this functionality.

I'll put in some work to split this up further and then we can take it from there!

Byron:
Just as a tiny reply: instead of writing a huge reply here we had a call and discovered interesting and previously unknown aspects of this topic that will change its trajectory, and to the better we think.
Will leave it at that, and I return to this PR when the next review happens.

Big picture stuff I took away from our meeting:

* The `Sign Commits` setting in the GUI will become a menu to select options from. The option names are just made up.
    - "Like Git" - the same as "unset" now (even though we might always disable it, I don't know)
    - "Disabled" - never sign. Good for GUIs that can't run signing programs
    - "On Push" - the new mode that is implemented here
* Instead of trying to predict what a push *would* do, we do a dry-run of the push preparation to see what would happen. This will be interesting as signing + dry-run is counter-intuitive, after all there should be no side-effects which program calls are. But I am sure we can figure something out, like temporarily changing the signing settings to something that has no side-effects (in-memory changes, not observable).
* `RefInfo` should definitely get information about whether a commit is signed or not

slarse:
Had a chat with @Byron and we came to the following conclusions:

1. Split the internal changes with `SignCommit` and `PickSignMode` into its own separate thing to be merged first
2. There _is_ a use case for the legacy "sign all the time if configured" mode - if you're not fully on GitButler and still interoperate with Git, that mode emulates Git's behavior. In that mode, it also makes sense to disable signing if it fails. @Byron's doc comments in the patch above also illuminate the connection between Git's signing settings and gitbutler.signCommits - these comments need to be incorporated into the code base.
3. `commits_list_unsigned` is really only necessary to be able to show what's going to happen -> better to just do a dry-run that you don't materialize to check what _actually_ happens. Thus it can be scrapped.

Re-stacking this to be easier to review and incorporate the above changes.

slarse:
FYI still working on tweaking this.

crates/but-core/src/commit.rs
```diff
@@ -148,6 +148,39 @@ impl From<&Headers> for Vec<(BString, BString)> {
...
+    /// fails, GitButler basically can't do anything, so we flip `gitbutler.signCommits=false` as a
+    /// kill switch to disable signing for GitButler. `gitbutler.signCommits` taking precedence
+    /// over `commit.gpgSign` means we can honor Git's signing settings by default, but disable it
+    /// in the event that we fail to sign without affecting Git.
+    ///
+    /// This is really only necessary for the use case where GitButler acts in a Git-compatible
+    /// fashion to sign commits all the time. When signing more conservatively, for example just
+    /// before pushing, the need for this kind of kill-switching is greatly reduced, perhaps to the
+    /// point where we can remove it. But as long as we want to smoothly interoperate with Git, it
+    /// should remain.
```
slarse: Please help me double check that this makes sense. It reads right to me and I've double and triple checked it, but... It's a lot of text.

crates/but-rebase/src/graph_rebase/cherry_pick.rs
```diff
@@ -35,7 +35,52 @@ pub enum CherryPickOutcome {
...
+    /// parents.
+    ///
+    /// [`SignCommit`] controls the signing behavior of the pick, if it gets cherry-picked into the
+    /// rebase.
+    IfChanged(SignCommit),
+    /// Forces a cherry-pick on a commit. This is useful for example to sign or unsign commits.
+    ///
+    /// [`SignCommit`] controls the signing behavior of the forced pick.
+    Force(SignCommit),
+}
```
slarse: I really struggled with getting this down the way I want it. I wasn't happy with the 4 PickSignMode variants I initially had here (Never, IfChanged, IfChangedAndSignCommitsEnabled, Force), because to the rebase engine the choice was still binary: Either the pick is forced, or it's not and we only cherry-pick it if something has changed.

What I finally realized was bugging me was that this single `PickSignMode` variant was controlling two things: whether the commit gets cherry-picked at all (the rebase engine's direct concern), and whether the commit gets signed (the commit machinery's concern, the rebase engine doesn't really care).

It would possibly be even cleaner to completely separate these concerns, s.t. we don't nest the `SignCommit` enum inside the `PickDecisionMode` variants. Right now I chose not to do that because I do think these are tightly related in terms of functionality.

I'm still uncertain about this design but I think this is the _least bad_ design I've come up with, and it cleanly separates the "when to pick" and "how to commit" parts of the cherry-picking.

Byron:
> FYI still working on tweaking this.

Let's indicate this with draft mode. Then the "Request for review" will automatically be sent out when you are ready and take it out of draft.

Also, Copilot will stop pestering you :D.

crates/but-rebase/src/graph_rebase/cherry_pick.rs
```diff
@@ -72,9 +121,23 @@ pub fn cherry_pick(
...
+            // We should really only end up here in the case where we are attempting to force-sign
+            // a commit that has no parents. In that case, we want to just re-write said commit
+            // as-is, with potentially changed sign_commit setting.
+            let commit = crate::commit::create(
+                repo,
+                target.inner,
+                DateMode::CommitterUpdateAuthorKeep,
+                pick_decision_mode.into(),
+            )?;
+            Ok(CherryPickOutcome::Commit(commit))
```
slarse: Unsure about this implementation. What I want to do is just to re-write the commit that was picked with PickDecisionMode::Force, which should be the target here.

slarse (commented):
@Byron Okay, so I've reworked this completely to only encompass the parts that programmatically control signing behavior internally inside of GitButler. That's really the core concept that paves the way for sign-on-push: we need to be able to control signing independently of configuration.

I think that this might _still_ not be ready to merge, there are some parts I'm not happy with and not confident in (see comments in code). I think we should take our time, anything that goes into `but-core` is... well, core. And the graph engine as well.

Ironically, this is the same size as the PR was before as I went ahead and added almost 400 lines of test code to cement the behavior. This can easily be split further into one PR to introduce `SignCommit`, and one to introduce the `PickSignMode` thing, but I think that kind of split detracts from understanding the purpose of it all. Instead I split the commits.

crates/but-rebase/src/graph_rebase/cherry_pick.rs
```diff
@@ -35,7 +35,52 @@ pub enum CherryPickOutcome {
...
+/// `PickDecisionMode::IfChanged(SignCommit::Yes)`, and then re-pick the intended root(s) of the
+/// signature cascade with `PickDecisionMode::Force(SignCommit::Yes)`.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum PickDecisionMode {
+    /// Cherry-picks the commit only if it's necessary because of changes to the commit or its
+    /// parents.
+    ///
+    /// [`SignCommit`] controls the signing behavior of the pick, if it gets cherry-picked into the
+    /// rebase.
+    IfChanged(SignCommit),
```
slarse: Question: How do we feel about exposing parts of crate `X` in the API of crate `Y`? I was in the process of creating a `but_rebase::SignCommit` wrapper that maps 1-to-1 with `but_core::SignCommit` just to not mix-and-match crate components. But then I stopped myself.

It has the downside of a potential explosion of such logic-less adapters, but it helps keep crate boundaries clean. Thoughts?

Byron: That's a good call!
If it's 1:1 the same type, we have been sharing them, something I believe is practical and fine among friendly crates.

The `but-api` crate would be more interesting in that regard, but even there we have types from multiple crates (among which is `gix`) mixed in.

As I don't think we we ever had a problem with that, I don't think it's an effort worth introducing now either.

crates/but-core/tests/fixtures/scenario/single-unsigned.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+git init
+
+export "GIT_CONFIG_COUNT=1"
+export "GIT_CONFIG_KEY_0=init.defaultBranch"
+export "GIT_CONFIG_VALUE_0=main"
```
Byron: This isn't needed as the fixture is executed through `gix-testtools`, which isolates Git with this and more.

Byron (commented):
Great, I will take a closer look soon. From what I could tell at a glimpse, `but-core` should be fine.

slarse:
@Caleb-T-Owens The `but-rebase` stuff in this PR needs your guiding eye. If you look at the commits you'll find the ones that only touch the graph rebase engine and adds tests for it, that's what I need your help with.

Let me know if you want to have a call about it.

crates/but-core/src/commit.rs
Byron: My main modification here is to not claim we ever want to not be Git compatible.
As such, I think this kill-switch is valuable even if sign-on-push is enabled, while agreeing that it's less necessary then.

crates/but-core/tests/core/commit.rs
```diff
@@ -181,6 +187,48 @@ mod create {
...
+
+        assert!(
+            commit.extra_headers().pgp_signature().is_some(),
+            "despite the settings saying everything is disabled, it still signs the commit"
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn sign_commit_yes_without_signing_key_errors() -> anyhow::Result<()> {
```
Byron, line 215: This is great to have!

Byron (approved):
Thanks a lot, now the parts I looked at, everything *but* `but-rebase` look good to me!
If you ask me, they could be merged first.
Maybe restructure this PR to make that possible and the `but-rebase` review can then stand on its own.

PS: I pushed directly in here because it seemed the PR isn't stacked anymore. But maybe that's wrong. Sorry for that.

slarse:
> Thanks a lot, now the parts I looked at, everything _but_ `but-rebase` look good to me! If you ask me, they could be merged first. Maybe restructure this PR to make that possible and the `but-rebase` review can then stand on its own.

Sounds good, I'll split out everything `but-rebase` and then merge what you reviewed. It's very easy to do, I just thought it detracted from the "how is this even going to be used" perspective to to that beforehand.

> PS: I pushed directly in here because it seemed the PR isn't stacked anymore. But maybe that's wrong. Sorry for that.

This was a correct assessment, so that's fine!

PR #13000, opened 2026-03-23
feat(but): add sign-on-push functionality to but CLI

slarse:
Adds in sign-on-push functionality, along with a new Git config setting to enable it. This is all backwards compatible with gitbutler.signCommits and will be invisible to end users for the time being, one has manually set signMode=OnPush in the config to enable the new behavior.

The UX here is still a bit rough. The push command is overall a bit rough to work with and I think this needs some major refactor. Perhaps not in this PR but soon.

# TODO
* Improve UX
* ...

---
- <kbd>&nbsp;1&nbsp;</kbd> #13000 👈


slarse:
no longer applicable

PR #13001, opened 2026-03-23
Add `but branch move --unstack <branch>` flag

davidpdrsn:
The command was changed from `but branch tear-off <branch>` to `but branch move --unstack <branch>`.

---

The TUI has the ability to move branches around. They can be moved onto other branches (stacking) and moved onto the merge base (torn off).

However the CLI only supports moving branches onto other branches. It doesn't support tearing them off. This adds that via a new `but branch tear-off` command.

Its pretty straight forward since a function in `but_api` already exists for doing this.

In the future PR I'll change the TUI to call this function so they're aligned.

## Naming

Tbh I'm not a big fan of `but branch tear-off <branch>`. The "tear-off" bit specifically. Sounds very harsh to me for some reason. I think I'd call it `unstack` as in `but branch unstack <branch>`. I guess the only potential downside to that is that `but branch stack` so its not symmetric.

Jujutsu has a very similar command called `jj parallelize` but I'm not a fan of "parallelize" either because its hard to spell. More generally jj does most "moving things around" via `jj rebase` so `jj parallelize` is just a convenience for doing a few rebases manually.

### DIFF
--- a/crates/but/tests/but/command/branch/tear_off.rs
+++ b/crates/but/tests/but/command/branch/tear_off.rs
@@ -10,11 +10,11 @@ fn tear_off_by_name() -> anyhow::Result<()> {
     env.setup_metadata(&["A", "B"])?;
     assert_initial_status(&env);
 
-    env.but("branch tear-off C")
+    env.but("branch move --unstack C")
         .assert()
         .success()
         .stdout_eq(str![[r#"
-Tore off branch 'C'.
+Unstacked branch 'C'.
 
 "#]])
         .stderr_eq(str![""]);
@@ -56,11 +56,11 @@ fn tear_off_by_cli_id() -> anyhow::Result<()> {
     env.setup_metadata(&["A", "B"])?;
     assert_initial_status(&env);
 
-    env.but("branch tear-off h0")
+    env.but("branch move --unstack h0")
         .assert()
         .success()
         .stdout_eq(str![[r#"
-Tore off branch 'C'.
+Unstacked branch 'C'.
 
 "#]])
         .stderr_eq(str![""]);
@@ -102,11 +102,11 @@ fn tear_off_single_branch_stack_succeeds() -> anyhow::Result<()> {
     env.setup_metadata(&["A", "B"])?;
     assert_initial_status(&env);
 
-    env.but("branch tear-off A")
+    env.but("branch move --unstack A")
         .assert()
         .success()
         .stdout_eq(str![[r#"
-Tore off branch 'A'.
+Unstacked branch 'A'.
 
 "#]])
         .stderr_eq(str![""]);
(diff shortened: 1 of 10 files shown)


krlvi:
Why do you feel "unstack" is problematic? Somehow it feels the most descriptive to me, as in.. does the opposite of stacking (i have something that is already stacked and this unstacks it).
To randomly drop another alternative to tear-off: split?

krlvi (approved):

davidpdrsn:
> Why do you feel "unstack" is problematic? Somehow it feels the most descriptive to me, as in.. does the opposite of stacking (i have something that is already stacked and this unstacks it).

Sorry if I wan't clear. I do think "unstack" is better and that we should use that. It would just be even nicer if we also had a "stack" command so things were symmetric.

crates/but/tests/but/command/branch/tear_off.rs
```diff
@@ -0,0 +1,167 @@
+use snapbox::str;
+
+use crate::utils::Sandbox;
+
+#[test]
+fn tear_off_by_name() -> anyhow::Result<()> {
+    let env = Sandbox::open_or_init_scenario_with_target_and_default_settings(
+        "two-stacks-one-single-and-ready-to-mingle-one-double",
```
slarse, line 8: Although not introduced in this PR, hat off to that scenario name.

crates/but/tests/but/command/branch/tear_off.rs
```diff
@@ -0,0 +1,167 @@
...
+├╯
+┊
+┊╭┄h0 [C]
+┊●   3842fc0 add C
+┊│
+┊├┄i0 [B]
+┊●   d3e2ba3 add B
+├╯
+┊
+┴ 0dc3733 [origin/main] 2000-01-02 add M
```
slarse, line 132: Complete tangent to this PR and not something to fix here, just want to bring it up: I find the visual presentation of the branches to be a bit too cramped. I was thinking about it today and couldn't help but comment now that I saw it again. I can _barely_ see that `C` and `B` form a stack here. Just adding one space reads much nicer today.


```bash
╭┄zz [unstaged changes]
┊     no changes
┊
┊ ╭┄g0 [A]
┊ ●   9477ae7 add A
├-╯
┊
┊ ╭┄h0 [C]
┊ ●   3842fc0 add C
┊ │
┊ ├┄i0 [B]
┊ ●   d3e2ba3 add B
├-╯
┊
┴ 0dc3733 [origin/main] 2000-01-02 add M
```

I know that stuff doesn't really align in my POC here but I wanted to raise the topic and see if you feel similarly, or if it's just me.

slarse (approved):
Looks nice, I was looking for this very thing earlier this morning as I was unclogging my mess of a sign-on-push contribution :D

The only thing I'm thinking of - do we need this to be a standalone command? Can't `but branch move` do this? E.g. `but branch move <branch> zz` or `but branch move <branch> --unstack` or something like that.

It _is_ moving a branch after all. So with a dedicated command for this we'll have two commands that both move branches, just slightly differently. IMO there is no conceptual difference between tearing off a branch from a stack and moving the branch to its own stack.

I generally feel that command overloading is bad, but here I'm not so sure.

davidpdrsn:
> It is moving a branch after all. So with a dedicated command for this we'll have two commands that both move branches, just slightly differently. IMO there is no conceptual difference between tearing off a branch from a stack and moving the branch to its own stack.

I completely agree! This is also closer to how it works in the TUI, at least conceptually.

Think I'm gonna go with `but branch move <branch> --unstack`.

crates/but/tests/but/command/branch/tear_off.rs
davidpdrsn: Thats a good idea! I agree its cramped. I'll fix it!

PR #13004, opened 2026-03-24
 `createDiffSpec`: omit hunk headers if change is addition or deletion

OliverJAsh:
Fixes GB-1149

### DIFF
--- a/apps/lite/ui/src/domain/DiffSpec.ts
+++ b/apps/lite/ui/src/domain/DiffSpec.ts
@@ -4,5 +4,6 @@ export const createDiffSpec = (change: TreeChange, hunkHeaders: Array<HunkHeader
 	pathBytes: change.pathBytes,
 	previousPathBytes:
 		change.status.type === "Rename" ? change.status.subject.previousPathBytes : null,
-	hunkHeaders,
+	hunkHeaders:
+		change.status.type === "Addition" || change.status.type === "Deletion" ? [] : hunkHeaders,
 });
--- a/apps/lite/ui/src/api/rub.ts
+++ b/apps/lite/ui/src/api/rub.ts
@@ -74,12 +74,12 @@ export const rub = async ({ projectId, source, target }: RubParams): Promise<Rub
 						Match.exhaustive,
 					),
 				),
-				Match.tag("Commit", (source) =>
+				Match.tag("Commit", (sourceParent) =>
 					Match.value(target).pipe(
 						Match.tag("Changes", async (target): Promise<RubResult> => {
 							const response = await window.lite.commitUncommitChanges({
 								projectId,
-								commitId: source.commitId,
+								commitId: sourceParent.commitId,
 								assignTo: target.stackId,
 								changes,
 							});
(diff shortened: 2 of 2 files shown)


apps/lite/ui/src/domain/DiffSpec.ts
```diff
@@ -4,5 +4,6 @@ export const createDiffSpec = (change: TreeChange, hunkHeaders: Array<HunkHeader
 	pathBytes: change.pathBytes,
 	previousPathBytes:
 		change.status.type === "Rename" ? change.status.subject.previousPathBytes : null,
-	hunkHeaders,
+	hunkHeaders:
+		change.status.type === "Addition" || change.status.type === "Deletion" ? [] : hunkHeaders,
```
Byron, line 8: While I am not qualified to review this PR and didn't spend much time with it, I am surprised that no addition or deletion works.
I thought the problem was that it's the last hunk in the file.

If that's not the case I wonder if this works correctly at all?

OliverJAsh: > I thought the problem was that it's the last hunk in the file.

If it's an addition or deletion, there can only be one hunk, if I understand correctly.

If that doesn't make sense, let's jump on a call as it's possible I'm misunderstanding as well!

Byron: Right, I was misreading this!

This check is on the change level, so it says that if a change is a deletion or addition, there is only ever one hunk. For these cases, this will work correctly.

Now I wonder what will happen if there is a modification that fills an empty file, or empties a file? If that hunk is dragged it's also the last one in the file. It *should* work, but it's definitely worth testing for to gain clarity - that hunk-related code was always a bit wonky and tuned to the needs of the current GUI, so surprising behaviour might be present.

OliverJAsh: > Now I wonder what will happen if there is a modification that fills an empty file, or empties a file? If that hunk is dragged it's also the last one in the file. It _should_ work, but it's definitely worth testing for to gain clarity - that hunk-related code was always a bit wonky and tuned to the needs of the current GUI, so surprising behaviour might be present.

It seems to work, both on `master` and on this branch.

Test case:

```sh
git init
git commit --allow-empty -m "Init"

git checkout -b feature-foo

touch foo.txt
git add foo.txt
git commit -m "add foo.txt"

echo "foo line 1" >> foo.txt
git add foo.txt
git commit -m "update foo.txt (1)"

git checkout main

but setup

but apply feature-foo
```

https://github.com/user-attachments/assets/559d096a-6eab-4468-870a-62fff7a57b6d

Byron: I am happy to hear that!
If you have a test-suite already, I'd love it if these cases could go in there a well. Pretty sure this is tested in the backend as well, but with hunk-based operations I feel more tests are better.

Byron (approved):
LGTM!

PS: I didn't want to hold up the merge by *not* approving. If I want to hold up I'd go for "Request Changes".

apps/lite/ui/src/domain/DiffSpec.ts
OliverJAsh: I don't have tests yet, but it's on my list of things to think about.

PR #13008, opened 2026-03-24
feat(tui): inline branch reword

davidpdrsn:
The TUI supports rewording commits inline by pressing enter. This extends that functionality to also support branches.

https://github.com/user-attachments/assets/dc929bef-af89-4c24-bbd2-afc8b0085e81

### DIFF
--- a/crates/but/src/command/legacy/status/tui/cursor/tests.rs
+++ b/crates/but/src/command/legacy/status/tui/cursor/tests.rs
@@ -1329,7 +1329,7 @@ fn movement_methods_can_move_cursor_in_inline_reword_mode() {
     ];
 
     let mut cursor = Cursor(1);
-    let inline_reword = Mode::InlineReword(InlineRewordMode {
+    let inline_reword = Mode::InlineReword(InlineRewordMode::Commit {
         commit_id: commit_id("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"),
         textarea: Box::new(TextArea::default()),
     });
@@ -1401,7 +1401,7 @@ fn is_selectable_is_true_in_inline_reword_mode() {
         cli_id: unassigned("s0"),
     });
 
-    let inline_reword = Mode::InlineReword(InlineRewordMode {
+    let inline_reword = Mode::InlineReword(InlineRewordMode::Commit {
         commit_id: commit_id("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"),
         textarea: Box::new(TextArea::default()),
     });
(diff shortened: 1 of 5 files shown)


slarse (approved):
I very much skimmed the implementation, but it looks good. Feels good to use, too.

When dicking around with it, I of course tried to name the branch bad things. And, unexpectedly, HEAD stuck. Which is bad because that's not an allowed branch name. I don't think this is at all related to the TUI (can do the same thing in CLI), so I'll make a separate bug ticket for it.

I was unable to get out of that state using GitButler, it gets very confused about it all (and rightfully so). I'll open a separate bug ticket for it and address it ASAP.

davidpdrsn:
Oh good catch! That's indeed a backend bug

Byron:
`refs/heads/HEAD` is a valid ref-name though, and I'd think that this is what happened.
The backend does validate and normalizee branch names, ensuring they are valid refs.

slarse:
@Byron 

> `refs/heads/HEAD` is a valid ref-name though, and I'd think that this is what happened. The backend does validate and normalizee branch names, ensuring they are valid refs.

You appear to be correct.

```bash
$ git check-ref-format refs/heads/HEAD
$ echo $?
0
```

I think I'm missing something here, is there a distinction between `refs/heads/X` and branches? In my world view, every ref in there is a local branch.

But Git [explicitly prohibits naming a branch `refs/heads/HEAD`](https://github.com/git/git/blob/ce74208c2fa13943fffa58f168ac27a76d0eb789/refs.c#L772-L774), it's just a hard-coded not allowed refname for branches.

```bash
$ git branch HEAD
fatal: 'HEAD' is not a valid branch name
hint: See 'git help check-ref-format'
hint: Disable this message with "git config set advice.refSyntax false"
```

`git help check-ref-format` says nothing about `refs/heads/HEAD`, by the way.

So all I can surmise is that `refs/heads/HEAD` is a syntactically correct ref name that is disallowed as a special case of ref name rules? But I think we should disallow the use of it to maintain Git compatibility, as opposed to trying to make it work? It bricks GitButler entirely and makes Git angry.

Although Git does allow me to rename the ref afterward, GitButler just gets sad as it resolves both the workspace commit and the branch tip commit.

```bash
[test-repository]$ but status
╭┄zz [unstaged changes]
┊     no changes
┊
┊╭┄my [my-valid-branch] (no commits)
├╯
┊
┴ 60dd738 [origin/main] 2026-02-23 Update README in GUI

Hint: run `but help` for all commands
[test-repository]$ but reword my -m HEAD
Renamed branch 'my-valid-branch' to 'HEAD'
[test-repository]$ but status
╭┄zz [unstaged changes]
┊     no changes
┊
┊╭┄HE [HEAD] (no commits)
├╯
┊
┴ 60dd738 [origin/main] 2026-02-23 Update README in GUI

Hint: run `but help` for all commands
[test-repository]$ but reword HE -m my-valid-branch
Error: The reference "refs/heads/HEAD" should have content c78d24390465ce7ce6d2a3aa3581ebd510eb2bc1, actual content was 60dd73885f3371ab2c1f78ced86976b9bb4ace3a
```

slarse:
@davidpdrsn Sorry about derailing your merged PR :D

davidpdrsn:
Unacceptable!

PR #13020, opened 2026-03-24
Fix broken lint task

OliverJAsh:
Previously we had "No tasks were executed as part of this run" ([example](https://github.com/gitbutlerapp/gitbutler/actions/runs/23514254384/job/68443242098#step:4:25)).

I guess this means linting wasn't enforced in CI?

It's been like this for years… I'm confused why it hasn't been noticed before now?

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
 		"knip:prod:fix": "knip --strict --fix",
 		"tauri": "cross-env \"CARGO_TARGET_DIR=$PNPM_SCRIPT_SRC_DIR/target/tauri\" tauri",
 		"tauri-for-release": "tauri",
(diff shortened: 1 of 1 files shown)


slarse:
@OliverJAsh I looked into this, and it's a bug in turborepo. The `//#whatever` syntax is for explicitly executing a root level task, something that got borked in turborepo 2.8.14 and fixed in 2.8.17. So it's been broken for that duration.

See this PR: https://github.com/vercel/turborepo/pull/12242

We upgraded to 2.8.17 a few days ago and it's been running correctly since.

OliverJAsh:
Cheers @slarse!

PR #13024, opened 2026-03-25
gitbutler-edit-mode: remove usage of ctx

jonathantanmy2:
Many usages still remain due to needing git2 features (especially those that checkout an index), but I removed what I could.


krlvi (approved):

Byron (approved):
Great!

PR #13045, opened 2026-03-26
refactor: update ASCII art banner design

krlvi:
It's still not great, but better than the X we had before... 
This is the ascii art we show when doing `but setup`

(fyi @PavelLaptev - this is super low prio so dont worry about it but i wanted to ping you as an FYI)


PavelLaptev:
I'll take a look later today and merge it. Yes, definitely an improvement over the X logo. I'll follow up with another PR shortly after.

PR #13046, opened 2026-03-26
feat(but-rebase): add programmatic signing control for graph engine

slarse:
This PR puts down the groundwork for enabling sign-on-push functionality by introducing new facilities to tweak sign behavior for the graph rebase engine. There are no plans to port this to the old rebase engine.

The key design goal is to allow for programmatically driven signing, where we can decide at runtime if a certain rebase operation should cause signing (e.g. just before a push), while another one should not.  At the same time, we want to retain the old `gitbutler.signCommits` behavior for the sake of Git interoperability - you might do some work with GitButler and then decide to push with Git, and that should still work, respecting signing settings.

The two concepts introduced to the graph rebase engine in this commit is a `PickMode`, allowing the caller to tune when a commit is cherry-picked, and `SignCommit`, which allows for tuning signing behavior beyond what was previously possible. Setting `PickMode::Force` and `SignCommit::Yes` or `SignCommit::IfSignCommitsEnabled` allows one to sign a commit that is otherwise unchanged. For descendants to also be signed, they need to be picked with an appropriate `SignCommit` setting.

None of the new behavior is as of yet exposed in any of the GitButler UIs, that'll come later down the road with sign-on-push functionality.

# Open questions

* When a commit is rewritten _only_ to be signed, do we want to update the committer date? This is currently what happens, and I tend to think that's correct as the signature _does_ change the commit. But I wanted to raise it as a question.

---
- <kbd>&nbsp;2&nbsp;</kbd> #13046 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #12999


crates/but-rebase/tests/fixtures/generated-archives/unsigned-commits-with-signing-key-setup-but-signing-disabled.tar
slarse, line 1: We should reuse a single key for these tests, but I don't feel like fixing that _here_ so I'll do it separately instead. May or may not be merged before this PR.

crates/but-rebase/src/graph_rebase/cherry_pick.rs
```diff
@@ -35,7 +35,18 @@ pub enum CherryPickOutcome {
...
+/// Controls when a commit is cherry-picked during a rebase.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum PickMode {
+    /// Cherry-picks the commit only if it's necessary because of changes to the commit or its
+    /// parents.
+    IfChanged,
+    /// Forces a cherry-pick on a commit. This is for example useful in combination with
+    /// [`SignCommit`] to sign/unsign commits that are otherwise unchanged.
+    Force,
+}
```
slarse, line 47: There is already a PickMode enum in the old rebase engine. I chose not to reuse it to keep the rebase engines cleanly separated, and this is such a trivial enum anyway.

crates/but-rebase/src/graph_rebase/cherry_pick.rs
```diff
@@ -72,9 +87,22 @@ pub fn cherry_pick(
...
+            // point, it's safe to simply recreate that commit outright.
+            //
+            // Currently, the only known use case for this is to forcibly sign/unsign root commits.
+            let commit = crate::commit::create(
+                repo,
+                target.inner,
+                DateMode::CommitterUpdateAuthorKeep,
+                sign_commit,
+            )?;
+            Ok(CherryPickOutcome::Commit(commit))
```
slarse, line 101: Please have an extra careful look at this comment and logic. I believe this to be accurate but I'm not entirely certain. I drew the conclusion that it's safe to just commit outright, rather than going through `commit_from_unconflicted_tree()`, but I'm just not sure about this.

There is a test case in `signing_preferences.rs` that tests force-picking and signing the root commit, and it fails if this is removed.

crates/but-rebase/src/graph_rebase/creation.rs
```diff
@@ -1,22 +1,50 @@
...
+    pub default_pick_mode: PickMode,
+    /// Determines how resulting commits are signed.
+    pub default_sign_commit: SignCommit,
+}
+
+impl Default for GraphEditorOptions {
+    fn default() -> Self {
+        Self {
+            default_pick_mode: PickMode::IfChanged,
+            default_sign_commit: SignCommit::IfSignCommitsEnabled,
```
slarse: This set of options reflects the behavior we currently have in the rebase engine.

Byron:
> When a commit is rewritten only to be signed, do we want to update the committer date? This is currently what happens, and I tend to think that's correct as the signature does change the commit. But I wanted to raise it as a question.

That's a great question and I have no answer, except for the intuition that cases can be made for both. I'd also think it's safe to keep it as is and alter the committer date (and committer?) when signing, and see how that fares in the wild.

Caleb-T-Owens (approved):
All LGTM.

I am wondering if we really need the `GraphOpts` because the behaviour of that is a bit strange. IE: If I construct a graph with "force rebase" and then add a pick that doesn't specify "force rebase" down at one of the edges of the graph, or with `preserve_parents` set, it wouldn't be observed.

If we only had "force rebase" on the pick steps, it _would_ mean there is more work for the signing thing to change all the pick steps, but it avoids these initialising only settings.

slarse:
> All LGTM.
> 
> I am wondering if we really need the `GraphOpts` because the behaviour of that is a bit strange. IE: If I construct a graph with "force rebase" and then add a pick that doesn't specify "force rebase" down at one of the edges of the graph, or with `preserve_parents` set, it wouldn't be observed.
> 
> If we only had "force rebase" on the pick steps, it _would_ mean there is more work for the signing thing to change all the pick steps, but it avoids these initialising only settings.

I'm not _entirely_ sure about what you're onto so I think we should take a call to discuss this.

The short answer is that to facilitate sign-on-push, there is no reason to create a graph with `default_pick_mode=PickMode::Force`. So from that perspective, we can omit it from the options. I chose to expose it simply because I thought it made the force-signing implementation more clear (default-picking with IfChanged, then selectively picking with Force).

crates/but-rebase/src/graph_rebase/cherry_pick.rs
Caleb-T-Owens: This does seem reasonable to me

crates/but-rebase/tests/fixtures/generated-archives/unsigned-commits-with-signing-key-setup-but-signing-disabled.tar
slarse: Oh poop, I did fix this separately, these shouldn't be here anymore, there's a static key in the `shared.sh` fixture helper. Either I somehow did not get a conflict here, or I sleepwalked through the resolution.

Caleb-T-Owens (approved):
LGTM

PR #13053, opened 2026-03-26
gitbutler-edit-mode: change most of test to gix

jonathantanmy2:
Most tests are converted from git2 to gix, except one which requires checking out an index.

The conversion is mostly straightforward except for one part where I couldn't figure out a better way to write a symbolic ref to HEAD.

### DIFF
--- a/crates/gitbutler-edit-mode/tests/edit_mode.rs
+++ b/crates/gitbutler-edit-mode/tests/edit_mode.rs
@@ -6,7 +6,6 @@ use but_core::{
 };
 use but_ctx::Context;
 use but_meta::{VirtualBranchesTomlMetadata, virtual_branches_legacy_types::Target};
-use but_oxidize::OidExt as _;
 use but_testsupport::{gix_testtools, open_repo};
 use git2::build::CheckoutBuilder;
 use gitbutler_edit_mode::commands::{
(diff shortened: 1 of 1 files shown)


krlvi (approved):

crates/gitbutler-edit-mode/tests/edit_mode.rs
```diff
@@ -282,25 +288,44 @@ fn enter_edit_mode_checks_out_conflicted_commit() -> Result<()> {
...
-        repo.reference(INTEGRATION_BRANCH_REF, workspace_head.id(), true, "")?;
-        repo.set_head(INTEGRATION_BRANCH_REF)?;
+        let repo = ctx.repo.get()?;
+        let workspace_head = repo.head_commit()?;
+        let foobar = workspace_head.decode()?.parents().next().unwrap();
+        repo.reference(
+            INTEGRATION_BRANCH_REF,
+            workspace_head.id(),
+            gix::refs::transaction::PreviousValue::Any,
+            "",
```
Byron, line 298: I see what's happening here, and it's not possible to create `INTEGRATION_BRANCH_REF` at the same time as setting `HEAD` to it. So all this is a bit wordy, but I also don't think there is a better way with the current APIs.

Issue #12924, opened 2026-03-19
Edit mode improvements


jonathantanmy2:
Some improvements that @Byron and I discussed just now (hopefully I didn't miss anything). Not sure if everything is feasible.

- [ ] edit mode commits should always have the parents of the commit being edited
- use `safe_checkout` for everything
  - [ ] entering edit mode (this means that we no longer stash existing uncommitted changes into a tree and reference it with a ref, but that whatever happens with `safe_checkout` now will happen, including aborting if there's a conflict)
  - [ ] exiting edit mode (since we no longer stash uncommitted changes when entering edit mode, it should be just a commit and then checkout, so we can use `safe_checkout` here - we don't need the safety but there's no harm using it)
- [ ] avoid `ctx`
- [ ] avoid `create_wd_tree` (the code currently uses it to create the commit - have to figure out how to do this properly)
- [ ] better integration with Git (e.g. showing MERGING or REBASING on the command prompt when in edit mode) - I'm not sure if possible
- [ ] the ability to create commits even when in edit mode, and have something reasonable occur when exiting edit mode - I'm not sure if possible

And one more thing that I should note.

- [ ] stop writing and reading the stack ID to/from edit mode metadata (I'll do this once the new edit mode is proven to be stable)

jonathantanmy2:
@Caleb-T-Owens in case he has opinions about this.

Caleb-T-Owens:
> edit mode commits should always have the parents of the commit being edited

Is this always possible in the case where a commit has a conflicted parent?

> entering edit mode (this means that we no longer stash existing uncommitted changes into a tree and reference it with a ref, but that whatever happens with safe_checkout now will happen, including aborting if there's a conflict)

What are the UX implications of this?

jonathantanmy2:
> > edit mode commits should always have the parents of the commit being edited
> 
> Is this always possible in the case where a commit has a conflicted parent?

I don't know of anything GitButler-specific that would prevent this, but I think that having no parent is as arbitrary as (or even more arbitrary than) reusing the commit's parents (and in Git, we can put whatever parent(s) we want on any commit). I don't think that a parent being conflicted would make a difference.

> > entering edit mode (this means that we no longer stash existing uncommitted changes into a tree and reference it with a ref, but that whatever happens with safe_checkout now will happen, including aborting if there's a conflict)
> 
> What are the UX implications of this?

If a user typically does not enter edit mode with uncommitted changes, there are no UX implications.

[...]

PR #13056, opened 2026-03-26
gitbutler-edit-mode: allow commit graph modification

jonathantanmy2:
Currently, when exiting edit mode, we only use the contents of the index. No matter where HEAD is, we take the index, create a commit out of it, and replace the edited commit in the commit graph with the newly created commit - in other words, it is always a one-to-one replacement.

This is inflexible for the user for not much gain. This feature was designed that way probably because the old rebase engine did not support arbitrary rebasing of graphs, but now that edit mode has switched over to the new rebase engine, this limitation no longer applies. Change edit mode so that we rebase on top of what the user currently has at HEAD, first applying any uncommitted files in the index to HEAD (for backwards compatibility, and also to be similar to what `git rebase -i` does).

There is a potentially unexpected interaction if HEAD is checked out by the user to a relatively new existing commit: we could end up applying a commit on itself. I've included a test demonstrating this behavior. This is consistent with how Git treats modifying a commit (e.g. by rebasing) as creating a completely new commit without deleting the old one, so things currently work as expected. However, if we end up introducing commit evolution, this no longer holds (there is a semantic of a commit being modified from A to B meaning that A is obsoleted, but in this interaction, both A and B coexist in the commit graph).

When implementing this, I made the HEAD (at the time of exiting edit mode) a "preserve parent" node in the rebase graph. This works but is arguably a hack. I've created an issue [1] to keep track of potential improvements to the new rebase engine that will eliminate the need for this hack.

[1] https://github.com/gitbutlerapp/gitbutler/issues/13055

### DIFF
--- a/crates/but-rebase/src/graph_rebase/mod.rs
+++ b/crates/but-rebase/src/graph_rebase/mod.rs
@@ -31,7 +31,7 @@ pub struct Pick {
     ///
     /// If this is Some, the commit WILL NOT be picked onto the parents the
     /// graph implies but instead on to the parents listed here.
-    pub(crate) preserved_parents: Option<Vec<gix::ObjectId>>,
+    pub preserved_parents: Option<Vec<gix::ObjectId>>,
     /// If set to false, a rebase will fail if this commit results in a
     /// conflicted state.
     pub conflictable: bool,
--- a/crates/gitbutler-edit-mode/src/lib.rs
+++ b/crates/gitbutler-edit-mode/src/lib.rs
@@ -11,7 +11,7 @@ use but_ctx::{
     access::{RepoExclusive, RepoShared},
 };
 use but_oxidize::{ObjectIdExt as _, OidExt, gix_to_git2_index};
-use but_rebase::graph_rebase::{Editor, Step};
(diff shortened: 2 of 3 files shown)


jonathantanmy2:
Forgot to mention that if a user enters edit mode with a version before #12935 and exits with this version, the UX will not be good. (Entering with a version at or after #12935 is fine though.)

Also, @Byron I think you mentioned a workflow in which you create multiple commits during edit mode here: https://github.com/gitbutlerapp/gitbutler/issues/12924#issuecomment-4101371590 . If you have time, could you look at the test in this PR and see if this PR's functionality would meet your needs?

Caleb-T-Owens (approved):

Byron:
It's great to see that edit mode is robust in the face of arbitrary commits being added while it's active! And while I don't see my own workflow in there, it's great to know that will be recovered from now.

jonathantanmy2:
> And while I don't see my own workflow in there, it's great to know that will be recovered from now.

Ah, thanks. Rereading your workflow, maybe what you want to do can be done by entering edit mode on the newest commit, and then checking out and/or resetting before starting work. But yeah, I guess your workflow is not directly supported.

PR #13058, opened 2026-03-27
Fix ghost index entries after oplog snapshot restore

mtsgrd:
After undoing an operation via the oplog, users could see phantom staged
changes — files appearing in `git status` as staged but not existing
ondisk. Running `git reset` would clear them, but the experience
wasconfusing and made undo feel unreliable.

Missing `index.write()`: `restore_snapshot` called   `index.read_tree()`
to load the snapshot's index into memory, but   never called
`index.write()` to flush it to disk. The on-disk index   retained its
pre-restore state, which included files from the   operation that was
being undone.


mtsgrd:
This was a dead end, it's about the index saved with the snapshot, and is not a source of ghost files.

PR #13079, opened 2026-03-27
Add concurrent commit race condition tests

krlvi:


krlvi:
@Byron FYI - this test should be helpful in debugging this https://github.com/gitbutlerapp/gitbutler/issues/13028

PR #13086, opened 2026-03-28
Add pre-flight conflict detection for cross-stack commit and branch moves

mtsgrd:
## Summary

When moving a commit or branch between stacks, GitButler now checks whether the move would produce conflicts *before* making any changes.

The motivation for preventing these conflicts upfront is that they are particularly hard to resolve: if a moved commit lands in a conflicted state, the only way to successfully save the resolution in the commit editor is to undo the changes that caused the dependency conflict — but nothing in the UI tells you that. Catching the problem before the move is a much better experience than leaving the user to figure out why their edits won't save.

The check runs two speculative rebases without modifying the repository:

- **Source**: would the commits remaining in the source stack still rebase cleanly after the move?
- **Destination**: would the moved commit(s) apply cleanly at the insertion point?

If either fails, the move is rejected with a clear error before `create_snapshot` is called — so a rejected move leaves the repository completely unchanged.

**What gets caught:**
- Moving a commit that other commits in the source stack depend on
- Moving a branch whose removal would leave dependent commits unable to rebase cleanly

**What is unaffected:**
- Intra-stack moves (source == destination)
- Moves between stacks that touch different files
- Moves from stacks that already have pre-existing conflicts (not treated as new problems)

## Test matrix

Eight tests covering the cross-stack move outcomes:

| Source remaining | Destination | Result | Tests |
|---|---|---|---|
| Clean | Clean | Allow | `*_non_overlapping` |
| Conflict (dependency removed) | Clean | Reject | `*_with_dependency_rejected` |
| Clean | Conflict | Reject | `*_destination_conflict` |
| Pre-existing conflict | Clean | Allow | `*_preexisting_conflict` |

Each row is tested for both commit moves and branch moves (8 total).

## Key files

- `utils.rs` — `check_rebase_for_conflicts()`: shared speculative rebase helper
- `move_commits.rs`, `move_branch.rs` — `preflight_check()` for each move type
- `actions.rs` — wires preflight before `create_snapshot`

### DIFF
--- a/crates/gitbutler-branch-actions/src/move_branch.rs
+++ b/crates/gitbutler-branch-actions/src/move_branch.rs
@@ -3,13 +3,18 @@ use but_core::ref_metadata::StackId;
 use but_ctx::{Context, access::RepoExclusive};
 use but_rebase::{Rebase, RebaseStep};
 use but_workspace::legacy::stack_ext::StackExt;
+use gitbutler_oplog::{
+    OplogExt,
+    entry::{OperationKind, SnapshotDetails},
+};
 use gitbutler_reference::{LocalRefname, Refname};
-use gitbutler_stack::{StackBranch, VirtualBranchesHandle};
+use gitbutler_stack::StackBranch;
 use gitbutler_workspace::branch_trees::{WorkspaceState, update_uncommitted_changes};
 use gix::refs::transaction::PreviousValue;
 use serde::Serialize;
 
-use crate::BranchManagerExt;
+use crate::{BranchManagerExt, VirtualBranchesExt as _, move_commits::bail_on_new_conflicts};
+use anyhow::bail;
 
 #[derive(Debug, Clone, Serialize)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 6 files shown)


mtsgrd:
@Byron this is something we discussed in our last eng meeting. It would be good to prevent moves that effectively cause cross-stack conflicts, since resolving them would require you undo each change where there was a dependency. That information isn't available while you're resolving a conflict, and fi you don't get it right you'll just be presented with a rebase error.

I'm not in a hurry to land this pr since I'm still validating both what the code does, but also what we should want in the given scenarios. But, meanwhile, it would be great if you take a quick look and let me know if it's largely ok or if there are things that stand out like a sore thumb?

crates/gitbutler-branch-actions/src/move_branch.rs
```diff
@@ -21,6 +21,57 @@ pub struct MoveBranchResult {
...
+    target_stack_id: StackId,
+    target_branch_name: &str,
+    source_stack_id: StackId,
+    subject_branch_name: &str,
+) -> Result<()> {
+    if source_stack_id == target_stack_id {
+        return Ok(());
+    }
+    let repo = ctx.repo.get()?;
+    let vb_state = VirtualBranchesHandle::new(ctx.project_data_dir());
```
Byron: Let's not introduce more VBH code, my main mission right now is actually to remove it 😅.
This information should be readily enough available in `ctx.workspace_*()`.

Byron (commented):
I just took a quick look and think it should be Ok in principle, as the check won't be observable.
One concern *could* be that it seems to preview the operation which *might* double its cost, but that shouldn't ever be in the way of correctness or even UX.

With that said, the PR shouldn't be merged with the dreaded VBH still in there, but removing it shouldn't be a problem.

crates/gitbutler-branch-actions/src/move_branch.rs
mtsgrd: Should be sorted now!

Caleb-T-Owens (changes requested):
I'm wondering if we really need to implement the rebases twice.

Would it be possible to inspect the output of the rebases before we update any metadata or references. This way there isn't any possible divergence between the check and the real operation, and we can avoid signing commits twice.

---

I'm also curious about the actual data loss scenario. In `gix` terms, we _are_ doing a force ours merge, but, any time we see that a force has happened, we then write it out as a conflicted commit which has all the data. We shouldn't end up loosing data. Could you help me understand the scenario better?

mtsgrd:
Sorry, I needed to update the PR description 🤦 This PR is really about preventing moves produce conflicted commits that are difficult to resolve.

That data loss thing was exaggerated by Claude during the draft phase of this pr. See the updated description at the top, and this commit message for an up-to-date description: https://github.com/gitbutlerapp/gitbutler/pull/13086/changes/36bab11b06c339beb171861e760bd9ef263aef44

Caleb-T-Owens:
> Sorry, I needed to update the PR description 🤦 This PR is really about preventing moves produce conflicted commits that are difficult to resolve.
> 
> That data loss thing was exaggerated by Claude during the draft phase of this pr. See the updated description at the top, and this commit message for an up-to-date description: [36bab11](https://github.com/gitbutlerapp/gitbutler/commit/36bab11b06c339beb171861e760bd9ef263aef44)

Ahh, ok.

The conflict scenario does make sense that it's difficult to resolve. I would quite like it if we could modify the operations to avoid needing to do the rebase work twice though.

mtsgrd:
> I would quite like it if we could modify the operations to avoid needing to do the rebase work twice though.

It would be clutch if that's possible!

Caleb-T-Owens:
> It would be clutch if that's possible!

Some of the operations may need some reorganisation, but you should be able to similarly inspect the output of the rebase before we update any metadata or references

mtsgrd:
@Caleb-T-Owens would you mind taking another look at this? It now only does a single pass rather than the speculative rebases.

Caleb-T-Owens (approved):

PR #13090, opened 2026-03-29
test(but-core): demonstrate data loss in safe_checkout_from_head deterministically

krlvi:
Add a single-threaded unit test that reproduces the data-loss bug without
any concurrency.  The key insight is that safe_checkout_from_head reads
repo.head_tree_id_or_empty() *at call time*.  If a concurrent process has
already updated HEAD (H0 → H_A), the next caller computes the diff H_A → H_B
where H_B was derived from the stale H0.  That diff marks A's file as DELETED
and the checkout removes it from disk.

The test sets up two workspace commits derived from the same parent H0:
  H_A = H0 + auth/session.ts
  H_B = H0 + api/pagination.ts   (no session.ts — stale derivation)

1. safe_checkout_from_head(H_A)  →  HEAD = H_A, session.ts created ✓
2. safe_checkout_from_head(H_B)  →  reads HEAD = H_A, diff H_A→H_B marks
   session.ts as DELETED → physically removes it  ← DATA LOSS

The test does not propagate errors immediately on the second checkout;
instead it checks whether session.ts still exists on disk and only then
emits the detailed failure message.  This ensures the data-loss assertion
fires even if the checkout also returns an error.

The test is #[ignore] because it currently FAILS (demonstrating the bug).
Remove the ignore and flip the assertion once the fix lands.

### DIFF
--- a/crates/but-core/tests/core/worktree/checkout.rs
+++ b/crates/but-core/tests/core/worktree/checkout.rs
@@ -1,4 +1,6 @@
-use but_core::worktree::{checkout, checkout::UncommitedWorktreeChanges, safe_checkout};
+use but_core::worktree::{
+    checkout, checkout::UncommitedWorktreeChanges, safe_checkout, safe_checkout_from_head,
+};
 use but_testsupport::{
     git_status, read_only_in_memory_scenario, visualize_commit_graph_all,
     visualize_disk_tree_skip_dot_git, visualize_index, visualize_tree, writable_scenario,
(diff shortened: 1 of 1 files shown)


Byron (changes requested):
@krlvi I took a look and don't think that the PR should not be merged, for two reasons:

- it doesn't show a race, it shows conflicting instructions. Checkout A then checkout B, and it does as it's told. Safe checkout is safe because it won't nuke unstaged or untracked files, but it's still allowed to set the target tree to match the given commit.
- safe_checkout is a plumbing function which expects to have a repository lock on the caller side. Once these are inter-process safe, safe-checkout will not easily be interleaved with conflicting calls.

Something that `safe_checkout` *could* do on top is to immediately grab a lock on the `.git/index` to prevent other checkouts, for instance by Git itself, from happening. This is definitely a possible improvement and one that I think should be done.

For example, right now, it would be possible for one `but commit` to wait for a commit-message, and a Git commit to come in in the meantime that will happily be overwritten once the commit message clears.

Will see to following this up.

krlvi:
> @krlvi I took a look and don't think that the PR should not be merged, for two reasons:
> 
>     * it doesn't show a race, it shows conflicting instructions. Checkout A then checkout B, and it does as it's told. Safe checkout is safe because it won't nuke unstaged or untracked files, but it's still allowed to set the target tree to match the given commit.
> 
>     * safe_checkout is a plumbing function which expects to have a repository lock on the caller side. Once these are inter-process safe, safe-checkout will not easily be interleaved with conflicting calls.
> 
> 
> Something that `safe_checkout` _could_ do on top is to immediately grab a lock on the `.git/index` to prevent other checkouts, for instance by Git itself, from happening. This is definitely a possible improvement and one that I think should be done.
> 
> For example, right now, it would be possible for one `but commit` to wait for a commit-message, and a Git commit to come in in the meantime that will happily be overwritten once the commit message clears.
> 
> Will see to following this up.

I see, this makes sense. So do i understand correctly that we mainly wanna address this at a "higher level"/app logic, like this test targets https://github.com/gitbutlerapp/gitbutler/pull/13079 ?

Byron:
@krlvi Yes, the higher level would be the way.
Sorry for the late response, still catching up.

PR #13101, opened 2026-03-30
Add flags for perf testing the tui

davidpdrsn:
This adds the following flags that help with benchmarking the TUI:

- `--quit-after <frames>`: Quit the TUI after the given number of frames. This is the most useful one as it allows you to see how long time it takes to launch the TUI, render a few frames, and then quit. That might be slowed down by rendering a large diff for example.
- `--headless`: Render to ratatui's `TestBackend` instead of a real terminal. Without this I couldn't get [cargo-flamegraph](https://github.com/flamegraph-rs/flamegraph) to work.
- `--diff`: Immediately show the diff view, which is otherwise not shown by default.
- `--select-commit <commit>`: Select a specific commit when opening the TUI.
- `--skip-status-after`: Dont print the status when exiting the TUI. Just slightly easier when the TUI doesn't print to the primary terminal screen.

With this you can benchmark the diff rendering with

```
hyperfine --warmup 2 "./target/release/but tui --quit-after 2 --skip-status-after --headless --select-commit fddbe8b378b55de48102efbc2bb2e54bc5be6802 --diff"
```

Where `fddbe8b378b55de48102efbc2bb2e54bc5be6802` is some commit with a large diff.

All these flags are only available if the CLI is compiled with `--features tui-profiling` so they wont be available in production builds.

### DIFF
--- a/crates/but/src/args/mod.rs
+++ b/crates/but/src/args/mod.rs
@@ -1147,22 +1147,40 @@ pub enum Subcommands {
     #[cfg(feature = "legacy")]
     Tui {
         /// Show debug pane with selected-line metadata.
+        ///
+        /// Requires `tui-profiling` feature.
+        #[cfg(feature = "tui-profiling")]
         #[clap(long, default_value_t = false)]
         debug: bool,
         /// Quit after rendering this many frames.
-        #[clap(long, hide = true)]
+        ///
+        /// Requires `tui-profiling` feature.
+        #[cfg(feature = "tui-profiling")]
+        #[clap(long)]
         quit_after: Option<u64>,
         /// Run the TUI with an in-memory terminal and no terminal event polling.
-        #[clap(long, hide = true)]
+        ///
+        /// Requires `tui-profiling` feature.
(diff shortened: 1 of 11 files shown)


crates/but/src/args/mod.rs
```diff
@@ -1147,8 +1147,41 @@ pub enum Subcommands {
...
         /// Show debug pane with selected-line metadata.
+        ///
+        /// Requires `tui-profiling` feature.
+        #[cfg(feature = "tui-profiling")]
         #[clap(long, default_value_t = false)]
         debug: bool,
+        /// Quit after rendering this many frames.
+        ///
+        /// Requires `tui-profiling` feature.
+        #[cfg(feature = "tui-profiling")]
```
Byron, line 1158: While this is internal and expert-mode, I really liked the way the 'HEADERS' or Titles (how was it called) of clap help to organize related flags. That way they can never ever mistaken for anything else if someone less expert would be compiling this.

This is how these look like then.

Image

PR #13107, opened 2026-03-30
discard commit

estib-vega:
Ability to completely ‘delete’ a commit and its associated changes.

- Add tests
- Add Oplog entry
- Expose API in the SDK

### DIFF
--- a/crates/but-api/src/commit/json.rs
+++ b/crates/but-api/src/commit/json.rs
@@ -1,7 +1,7 @@
 use bstr::ByteSlice;
 use serde::{Deserialize, Serialize};
 
-use crate::json::HexHash;
+use crate::{commit::types::CommitDiscardResult, json::HexHash};
 
 use super::types::{
     CommitCreateResult, CommitInsertBlankResult, CommitMoveResult, CommitRewordResult,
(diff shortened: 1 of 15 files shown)


crates/but-rebase/src/graph_rebase/mutate.rs
```diff
@@ -180,6 +180,12 @@ impl ToCommitSelector for gix::ObjectId {
     }
 }
 
+impl ToCommitSelector for gix::Id<'_> {
```
Byron, line 183: Having this helps to avoid `.detach()` calls. In generally, `gix::Id<'_>` should be acting like `gix::ObjectId` and everything else is at least a UX bug.

crates/but-api/src/commit/json.rs
```diff
@@ -209,6 +209,42 @@ impl From<CommitMoveResult> for UICommitMoveResult {
         }
     }
 }
+/// UI type for discarding a commit.
+#[derive(Debug, Serialize)]
+#[cfg_attr(feature = "export-schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "camelCase")]
+pub struct UICommitDiscardResult {
```
Byron, line 216: Something I have on my nit-list is the `UI` prefix, it serves no purpose as for all I can tell, this type is a JSON version of an API type. The `but-api` doesn't differentiate between UI and non-UI anymore.

Changing it means changes to one or two GUIs as well, but maybe new types could start not having it and the existing ones are renamed in a separate effort.

Byron (approved):
Thanks, I think it's good to go.

I added some workspace information to the tests and made them less noisy, but one unresolved issue remains: How to create tests that validate the implementation in `but-workspace` more than `but-rebase`?

Trusting `but-rebase` would mean there only have to be two tests for `discard_commit`, one for conflict rejection, the other one for the actual discarding.
If you think `but-rebase` is under-tested in a certain realm then I'd rather add discard-inspired tests there.

crates/but-api/src/commit/json.rs
estib-vega: I agree with the sentiment. I just kept it in there for alignment reasons.

Even thinking in the fact that these types will be exposed as part of the SDK, the fact that they're prefixed UI seems weird.

crates/but-rebase/src/graph_rebase/mutate.rs
estib-vega: Nice, thanks

PR #13138, opened 2026-04-01
Share forge credentials across dev, nightly, and release builds

mtsgrd:
GitHub and GitLab tokens were stored in per-channel keychain namespaces
and the account list (forge_settings.json) lived in per-channel data
directories. This made it impossible to be signed into GitHub on more
than one of dev/nightly/release at the same time — each build required
its own separate authentication.

Move keychain entries to the global namespace and the forge settings
file to the shared config directory so all builds share a single set
of forge credentials.


mtsgrd:
Estib will implement differently.

PR #13145, opened 2026-04-01
graph: squash commits

estib-vega:
Add an API that squashes commits using the new rebase engine.

### DIFF
--- a/crates/but-workspace/src/commit/squash_commits.rs
+++ b/crates/but-workspace/src/commit/squash_commits.rs
@@ -1,0 +1,186 @@
+//! An action to squash one commit into another.
+
+pub(crate) mod function {
+    use anyhow::{Result, bail};
+    use but_core::RefMetadata;
+    use but_graph::{SegmentIndex, SegmentRelation, projection::Workspace};
+    use but_rebase::{
+        commit::DateMode,
+        graph_rebase::{
+            Editor, Selector, Step, SuccessfulRebase, ToCommitSelector,
+            mutate::{SegmentDelimiter, SelectorSet},
+        },
+    };
+
+    #[derive(Debug, Clone, Copy, Eq, PartialEq)]
+    enum ReorderDirection {
+        MoveSubjectAboveTarget,
+        MoveSubjectBelowTarget,
+    }
+
+    fn find_commit_segment_index(
+        workspace: &Workspace,
+        commit_id: gix::ObjectId,
+    ) -> Option<SegmentIndex> {
+        let (_, stack_segment, _) = workspace.find_commit_and_containers(commit_id)?;
+        let commit_offset = stack_segment
+            .commits
+            .iter()
+            .position(|c| c.id == commit_id)?;
+
+        let mut owning_segment = stack_segment.id;
+        for (segment_id, offset) in &stack_segment.commits_by_segment {
+            if *offset > commit_offset {
+                break;
+            }
+            owning_segment = *segment_id;
+        }
+
+        Some(owning_segment)
+    }
+
+    fn determine_reorder_direction(
+        workspace: &Workspace,
+        repo: &gix::Repository,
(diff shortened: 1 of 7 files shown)


crates/but-workspace/src/commit/squash_commits.rs
```diff
@@ -0,0 +1,142 @@
...
+        MoveSubjectAboveTarget,
+        MoveSubjectBelowTarget,
+    }
+
+    fn determine_reorder_direction(
+        repo: &gix::Repository,
+        subject: &but_core::CommitOwned,
+        target: &but_core::CommitOwned,
+    ) -> Result<ReorderDirection> {
+        let merge_base = match repo.merge_base(subject.id, target.id) {
```
Byron, line 68: Can you try to use `editor.workspace.graph.find_git_merge_base(a, b)` here?
It's faster and the graph is expected to have all knowledge it needs to make that computation.

crates/but-workspace/tests/workspace/commit/squash_commits.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    insta::assert_snapshot!(visualize_commit_graph_all(&repo)?, @"
+    * c9f444c (HEAD -> three) commit three
+    * 16fd221 (origin/two, two) commit two
+    * 8b426d0 (one) commit one
+    ");
+
+    let subject_id = repo.rev_parse_single("three")?.detach();
+    let target_id = repo.rev_parse_single("two")?.detach();
+    let subject_tree = repo.find_commit(subject_id)?.tree_id()?.detach();
```
Byron, line 25: The last time I tried I was able to remove the `detach()` calls.

Byron (approved):
I took a look at it merely from a graph + gix perspectives and left a few comments, with the big one to avoid `repo.merge_base()` if a workspace is present.

It's also good to see a couple of tests involving a workspace, showing that branches are simply left empty once their only commit is squashed, which seems desirable.

Also, when looking at how merge-base is used, maybe we want something like `graph.relation_between(a, b)` yielding `Disjoint, Ancestor, Descendant, Identity` or something, to make clear what's going on?

If you agree, maybe just implement this in `but-graph` - it's pre-approved :D.

crates/but-workspace/src/commit/squash_commits.rs
estib-vega: Will do, out of curiosity: Does the fact that the workspace is a projection not mean that the merge base provided here might not be the true merge base?

crates/but-workspace/tests/workspace/commit/squash_commits.rs
estib-vega: Will try that

crates/but-workspace/src/commit/squash_commits.rs
estib-vega: Oh, `graph.find_git_merge_base` seems to be for segments, not commits like in this case

estib-vega: I'm using it at the beginning, but if the commits are part of the same segment, then I'll still use the repo's merge base

Byron: > Oh, `graph.find_git_merge_base` seems to be for segments, not commits like in this case

Yes, that's fine as every fork/split is happening in its own segments. Thus the segment you get always has a commit. You can see how it's already used in `but-graph`.

Byron: > I'm using it at the beginning, but if the commits are part of the same segment, then I'll still use the repo's merge base

I think now I understand that you tried using it, but it didn't give the desired results. This is something to look into and would love to understand the problem - merge-base computation should just be the exact same, and for all I could tell, it is.

PR #13153, opened 2026-04-02
commit ordering

estib-vega:
Create but-workspace commit ordering function that sorts the commits by
their parentage, according to the workspace appearance.

### Cool, but why?
In order to enable graph mutations with multiple commits, I need to sort the commits from parent-most to child-most.
That way I can do one operation at a time in memory, parent to child. And then materialize it at the very end.
 
Ordering should reduce the probability of running into conflicts.

### Disclaimer
This heavily uses AI for the generation of this algorithm

### Comprehensive docs
https://github.com/gitbutlerapp/gitbutler/pull/13153/changes#diff-bb49765eeedd900a433bba2fc60faf0d82730392d9053af079c8c1d2407aa905

### DIFF
--- a/crates/but-graph/src/projection/workspace/api.rs
+++ b/crates/but-graph/src/projection/workspace/api.rs
@@ -319,6 +319,28 @@ impl Workspace {
         })
     }
 
+    /// Try to find the owning graph segment of `commit_id` in the workspace.
+    ///
+    /// This uses the stack segment's `commits_by_segment` offsets to map a projected
+    /// commit back to its source graph segment.
+    pub fn find_commit_segment_index(&self, commit_id: gix::ObjectId) -> Option<SegmentIndex> {
+        let (_, stack_segment, _) = self.find_commit_and_containers(commit_id)?;
+        let commit_offset = stack_segment
+            .commits
+            .iter()
+            .position(|c| c.id == commit_id)?;
+
+        let mut owning_segment = stack_segment.id;
+        for (segment_id, offset) in &stack_segment.commits_by_segment {
+            if *offset > commit_offset {
+                break;
+            }
+            owning_segment = *segment_id;
+        }
+
+        Some(owning_segment)
+    }
+
     /// Like [`Self::find_segment_and_stack_by_refname`], but fails with an error.
     pub fn try_find_segment_and_stack_by_refname(
         &self,
--- a/crates/but-graph/src/projection/workspace/api.rs
+++ b/crates/but-graph/src/projection/workspace/api.rs
@@ -324,11 +324,14 @@ impl Workspace {
     /// This uses the stack segment's `commits_by_segment` offsets to map a projected
     /// commit back to its source graph segment.
     pub fn find_commit_segment_index(&self, commit_id: gix::ObjectId) -> Option<SegmentIndex> {
-        let (_, stack_segment, _) = self.find_commit_and_containers(commit_id)?;
-        let commit_offset = stack_segment
-            .commits
-            .iter()
-            .position(|c| c.id == commit_id)?;
+        let (stack_segment, commit_offset) = self.stacks.iter().find_map(|stack| {
+            stack.segments.iter().find_map(|seg| {
+                seg.commits
+                    .iter()
+                    .enumerate()
+                    .find_map(|(offset, commit)| (commit.id == commit_id).then_some((seg, offset)))
+            })
+        })?;
 
         let mut owning_segment = stack_segment.id;
         for (segment_id, offset) in &stack_segment.commits_by_segment {
--- a/crates/but-rebase/docs/commit_parentage.md
+++ b/crates/but-rebase/docs/commit_parentage.md
@@ -1,0 +1,141 @@
+# Commit Parentage Ordering
+
+This document explains how commit selector ordering works in
+`Editor::order_commit_selectors_by_parentage` (implemented in
+`but-rebase/src/graph_rebase/ordering.rs`), and why the implementation is
+structured the way it is.
+
+## Goal
+
+Given a set of commit selectors, produce a deterministic order such that:
+
+- Parent commits come before child commits.
+- Unrelated commits still have a stable, deterministic order.
+- Duplicate selectors are deduplicated by commit id (first occurrence wins).
+
+This ordering is useful for operations that must apply commits in dependency-safe order.
+
+## Inputs and Output
+
+Method: `editor.order_commit_selectors_by_parentage(selectors) -> Result<Vec<Selector>>`
+
+- Input selectors can be any type implementing `ToCommitSelector`.
+- The output is a list of normalized `Selector` values.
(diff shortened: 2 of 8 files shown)


crates/but-workspace/src/ordering/commit_parentage.rs
```diff
@@ -0,0 +1,221 @@
...
+    {
+        SegmentRelation::Ancestor => return Ok(Relation::LeftIsAncestorOfRight),
+        SegmentRelation::Descendant => return Ok(Relation::RightIsAncestorOfLeft),
+        SegmentRelation::Disjoint | SegmentRelation::Diverged => return Ok(Relation::Unrelated),
+        SegmentRelation::Identity => {
+            // Commits can still be in parent/child relation inside one segment.
+        }
+    }
+
+    let merge_base = match editor.repo().merge_base(left.id, right.id) {
```
Caleb-T-Owens: Something I’ve been thinking about in some of my graph overlay work is how to reconcile the fact that a pick statement’s OIDs may not imply the same history as the commit graph. For example, with `editor.insert_below(target, pick, below)`, the `target` will still have the same OID, but the editor graph itself has changed relative to what git would be telling you.

In most cases, that divergence should be *fine*. However, if more drastic operations are being performed, such as reparenting commits, it could lead to unexpected issues.

In `https://github.com/gitbutlerapp/gitbutler/pull/13080`, I’ve actually removed public access to the `workspace` property from the editor, because it always reflects the version of the workspace at the time the editor was created and therefore *isn’t* representative of the current graph state.

I’ve also modified the functions that were using it so that they call `.rebase()` and retrieve the latest overlay graph from `SuccessfulRebase`. That means that when we call `into_editor` again, we can work with the most up-to-date graph, aligned with the editor’s current state.

---

I’d love @Byron’s thoughts on this, but I’ve been wondering more and more whether it would be helpful to have some graph-inspection operations that run specifically on the editor’s step graph, so we can find things like merge bases there as well.

It *would* mean an additional implementation of logic similar to what already exists in `but_graph`, but it would also let us operate without the same desync concerns.

---

Perhaps I’m missing some context here, though. Do let me know.

crates/but-workspace/src/ordering/commit_parentage.md
Byron, line 1: There is actually a way to use this file as documentation for the Rust module itself by inclusion, `tower` does that a lot. That's definitely something I'd use to not have a loose file lying around in the source tree.

crates/but-workspace/src/lib.rs
```diff
@@ -35,6 +35,9 @@ pub mod legacy;
 /// Types specifically for the user-interface.
 pub mod ui;
 
+/// Utilities for deterministic ordering operations.
+pub mod ordering;
```
Byron: I'd definitely prefer this module to be in `commit` as it's about `commit` ordering.

crates/but-workspace/src/workspace_graph.rs
```diff
@@ -0,0 +1,26 @@
+use but_graph::{SegmentIndex, projection::Workspace};
+
+/// Find the owning graph segment for `commit_id` in `workspace`.
+///
+/// This uses the stack segment's `commits_by_segment` offsets to map a projected
+/// commit back to its source graph segment.
+pub(crate) fn find_commit_segment_index(
+    workspace: &Workspace,
+    commit_id: gix::ObjectId,
+) -> Option<SegmentIndex> {
```
Byron: This should certainly be in `but-graph`, given that it provides a way to find the actual segment owning a commit by ObjectId in a workspace. This is quite under-used right now, so a simple way to do this is missing.

Byron (commented):
I only looked at it from an organizational perspective and think changes have to be made. 
My second feeling is that `but-rebase` should probably have more native support for at least parts of what needs to be accomplished here, but it's nothing more than that (a feeling) either.

crates/but-workspace/src/ordering/commit_parentage.rs
Byron: > I’d love @Byron’s thoughts on this, but I’ve been wondering more and more whether it would be helpful to have some graph-inspection operations that run specifically on the editor’s step graph, so we can find things like merge bases there as well.
>
> It would mean an additional implementation of logic similar to what already exists in but_graph, but it would also let us operate without the same desync concerns.

@Caleb-T-Owens That's an interesting problem. The 'safest' way seems to be to provide `workspace()` or `graph()` so that the editor can ensure the graph is representing the latest edits. While its more expensive, I do fear de-sync a bit, but ultimately would also trust that it's probably fine. The `but-graph` version of the merge-base implementation is generated based on the implementation in `gix`, which was hand-implemented based on Git :D. 
However, my preference right now is to do the super-simple and explicit thing that makes the data structure clear, i.e. `editor.workspace().graph.merge_base…()`. @estib-vega I'd also think that a method could be added that returns CommitIDs instead of SegmentIDs.

crates/but-workspace/src/lib.rs
estib-vega: Hmmm maybe. The motivation is that this will be used by branch & by commits, and any API that should be able to take multiple commits as input parameters.

estib-vega: If we later need branch ordering (e.g. for moving multiple branches at a time) we could add that method there. Even if we're ordering commits behind the scenes

crates/but-workspace/src/workspace_graph.rs
estib-vega: Yep, that's true

crates/but-workspace/src/lib.rs
Byron: Ah, I see, maybe it's more about branches then? It's just that when I see `but_workspace::ordering`, I think of the ordering of workspaces. Is there any place that is more specific but still fitting?

estib-vega: This could be maybe also part of but-graph? Maybe we let the graph sort things since it should have all the information there. Merge bases and topology information.

Byron: If you think that would work, I am all for it. My question would be if it's better in the step-graph though as it is related to mutation, and it's technically a different graph which may (or may not) be better suited for this kind of thing.

In any case, it seems like this can move elsewhere then 🎉.

estib-vega:
Update:
- The function no longer uses the `workspace` at all. 
- The ordering function lives in `but-graph` and is accessible from the editor.

Addressed other feedback as well regarding the docs and where the `find_commit_segment_index` (accessible through the workspace).

Byron (approved):
Thanks for addressing my remarks. Everything looks good to me now.

I particularly looked at:

* documentation inclusion similar to how it's done in `tower`
* added function on `WorkspaceProjection`
* usage of that function in `but-workspace`

I didn't look at `but-rebase` at all, but am glad that `ordering` found its home there.

PR #13155, opened 2026-04-02
Fix workspace merge failure when stacks have adjacent edits

mtsgrd:
When two stacks each modify adjacent but non-overlapping lines in the
same file, git2 treats the adjacent hunks as conflicting. This caused
`update_uncommitted_changes` to fail with `MergeConflict (-24)` on any
mutation that recomputed the workspace tree (squash, reorder, etc.)
— even though the stacks don't actually conflict.

The workspace commit was already built using gix
(`remerged_workspace_tree_v2`), which resolves adjacent hunks cleanly,
so it was possible to reach a valid workspace state that then broke on
every subsequent operation.

**Fix**

Replace the git2 octopus merge in `merge_workspace` with the same
gix-based approach used for workspace commit creation:
`merge_options_fail_fast` + `has_unresolved_conflicts`. The function
now returns `gix::ObjectId`; callers in `integrate.rs` that pass the
result into git2 APIs convert with `.to_git2()`.

**Test**

Add a regression test that calls `merge_workspace` directly with two
stacks whose edits are adjacent from both sides (Stack A owns lines 1–5
and 11–15, Stack B owns lines 6–10). Verified the test fails with
`MergeConflict (-24)` under git2 and passes under gix.

### DIFF
--- a/crates/but-worktrees/src/integrate.rs
+++ b/crates/but-worktrees/src/integrate.rs
@@ -244,12 +244,13 @@ fn worktree_integration_inner(
     let wd_tree = repo.create_wd_tree(0)?;
 
     let working_dir_conflicts = {
+        let gix_repo = ctx.clone_repo_for_merging()?;
         let before_heads = stacks
             .iter()
             .map(|s| s.head_oid(ctx))
             .collect::<Result<Vec<_>>>()?;
         let before = WorkspaceState::create_from_heads(ctx, perm, &before_heads)?;
-        let before = merge_workspace(&*ctx.git2_repo.get()?, before)?;
+        let before = merge_workspace(&gix_repo, &before)?.to_git2();
         let mut after_heads = stacks
             .iter()
             .filter(|s| s.id != stack.id)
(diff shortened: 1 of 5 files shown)


mtsgrd:
@Byron this is different from the others but it has a similar consequence. In this screenshot there are two commits that touch Segment.svelte, they do not overlap, but have adjacent edits on lines 24:25.

As a result, any operation that recomputes the workspace commit failed, and knowing why takes knowledge of this bug. It took me a few minutes to find the culprit, even though I knew what to look for.

image

```
command: squash_commits
params: {"projectId":"L1VzZXJzL21hdHRpYXMvYy9naXRidXRsZXIvLmdpdA","stackId":"5323a6a0-fbcf-41ad-987e-9344d96a9a99","sourceCommitIds":["b8460f75257ee8dc5e5b9cf54caccf409892bc75"],"targetCommitId":"1ffae114a3d46b6e547c62b92a2e527a18829819"})

merge conflicts exist; class=Merge (22); code=MergeConflict (-24)
```

krlvi (approved):

Byron:
It's good to see that `gix` is actually fixing something merge-related this time 😅.

PR #13156, opened 2026-04-02
fix: prevent naming a branch HEAD

slarse:
NOTE: I've pinged @Byron for review in Discord. If you're not @Byron, feel free to review anyway if that interests you, but just know that I'll be waiting for our Byromation to trigger before I proceed with anything here. Code owners cannot be removed as reviewers, so I have no way of indicating whom I'm expecting a review from ...

## 🧢 Changes
Git disallows the name HEAD for a branch.

Try running e.g. `git check-ref-format --branch HEAD` or `git branch HEAD` and you should see an error message along the lines of `fatal: 'HEAD' is not a valid branch name`

I can't find this properly documented in Git's refspec, but it's clearly hard coded into `refs.c#check_branch_ref`.

https://github.com/git/git/blob/cf2139f8e1680b076e115bc0b349e369b4b0ecc4/refs.c#L776-L778

## Note
This fix applies to the GUI as well as we use the same normalization function there, but I've only implemented high-level tests for the CLI. I felt like unit-testing the function and having one high-level path tested is sufficient.

## ☕️ Reasoning
We want to check the name _after_ normalization, so I think the most reasonable place to perform the check is in the function that does the normalization so as not to normalize and then forget this validation check.

## Open questions
* What about case insensitive file systems (like macOS is by default, and I believe Windows as well)? Git does _not_ prevent creating a branch named e.g. `head`, but if you _do_ it starts printing warnings that "HEAD is ambiguous". For strict Git-compatibility, `head` (or `hEAd` or any other capitalization) is an allowed branch name, but for case insensitive file systems they seem to cause the same issues that `HEAD` as a branch name does. I feel like to some extent, Git just doesn't play nicely with case insensitive file systems, but maybe GitButler should?
* Why do case insensitive file systems exist?
* After writing tests that strip trailing dashes all day, I realize that there's absolutely nothing in the refspec that prevents trailing dashes. _Leading_ dashes are prohibited in branch names specifically, but trailing ones are fine.

```bash
[gitbutler]$ git check-ref-format --branch hello----
hello----
[gitbutler]$ git check-ref-format --branch HEAD-
HEAD-
[gitbutler]$ git check-ref-format --branch -hello
fatal: '-hello' is not a valid branch name
```

I don't see a reason to deviate from Git here?

### DIFF
--- a/crates/but-core/tests/core/branch/normalize_short_name.rs
+++ b/crates/but-core/tests/core/branch/normalize_short_name.rs
@@ -42,6 +42,24 @@ fn clear_error_on_failure() {
 }
 
 #[test]
+/// HEAD is not a valid branch name, so when we normalize names we should disallow it.
+///
+/// To validate this claim, run `git check-ref-format --branch 'HEAD'` or `git branch HEAD`, they
+/// will reject with `fatal: 'HEAD' is not a valid branch name`.
+fn head_is_rejected() {
+    assert_eq!(
+        normalize_short_name("HEAD").unwrap_err().to_string(),
+        "Could not turn \"HEAD\" into a valid reference name",
+        "HEAD is rejected"
+    );
+    assert_eq!(
+        normalize_short_name("HEAD-").unwrap_err().to_string(),
+        "Could not turn \"HEAD-\" into a valid reference name",
+        "name that normalizes to HEAD is rejected"
+    );
+}
+
+#[test]
 fn complex_valid() -> anyhow::Result<()> {
     assert_eq!(normalize_short_name("feature/branch")?, "feature/branch");
     assert_eq!(
--- a/crates/but-core/src/branch/normalize.rs
+++ b/crates/but-core/src/branch/normalize.rs
@@ -37,8 +37,9 @@ pub fn normalize_short_name<'a>(name: impl Into<&'a BStr>) -> anyhow::Result<BSt
         true
     });
 
-    if sanitized.is_empty() {
+    if sanitized.is_empty() || sanitized == "HEAD" {
         bail!("Could not turn {name:?} into a valid reference name")
     }
+
     Ok(sanitized)
 }
--- a/crates/but/tests/but/command/branch/new.rs
+++ b/crates/but/tests/but/command/branch/new.rs
@@ -34,6 +34,38 @@ fn outputs_branch_name() -> anyhow::Result<()> {
 }
 
 #[test]
+fn rejects_head() -> anyhow::Result<()> {
+    let env = Sandbox::init_scenario_with_target_and_default_settings("one-stack")?;
+    env.setup_metadata(&["A"])?;
+
+    env.but("branch new HEAD")
+        .assert()
+        .failure()
+        .stderr_eq(str![[r#"
+Error: Could not turn "HEAD" into a valid reference name
+
+"#]]);
+
+    Ok(())
+}
+
+#[test]
+fn rejects_name_that_normalizes_to_head() -> anyhow::Result<()> {
+    let env = Sandbox::init_scenario_with_target_and_default_settings("one-stack")?;
+    env.setup_metadata(&["A"])?;
+
+    env.but("branch new HEAD-")
+        .assert()
+        .failure()
+        .stderr_eq(str![[r#"
+Error: Could not turn "HEAD-" into a valid reference name
+
+"#]]);
+
+    Ok(())
+}
+
+#[test]
 fn with_json_output() -> anyhow::Result<()> {
     let env = Sandbox::init_scenario_with_target_and_default_settings("one-stack")?;
     insta::assert_snapshot!(env.git_log()?, @r"
--- a/crates/but/tests/but/command/reword.rs
+++ b/crates/but/tests/but/command/reword.rs
@@ -101,6 +101,24 @@ Renamed branch 'branch-to-rename-123' to 'renamed-branch'
 }
 
 #[test]
+fn reword_branch_rejects_head() -> anyhow::Result<()> {
+    let env = Sandbox::init_scenario_with_target_and_default_settings("one-stack")?;
+    env.setup_metadata(&["A"])?;
+
+    env.but("branch new branch-to-rename").assert().success();
+
+    env.but("reword branch-to-rename -m HEAD")
+        .assert()
+        .failure()
+        .stderr_eq(str![[r#"
+Error: Could not turn "HEAD" into a valid reference name
+
+"#]]);
+
+    Ok(())
+}
+
+#[test]
 fn reword_commit_with_same_message_succeeds_as_noop() -> anyhow::Result<()> {
     let env = Sandbox::init_scenario_with_target_and_default_settings("one-stack")?;
     insta::assert_snapshot!(env.git_log()?, @r"


crates/but-core/tests/core/branch/normalize_short_name.rs
```diff
@@ -41,6 +41,24 @@ fn clear_error_on_failure() {
...
+fn head_is_rejected() {
+    assert_eq!(
+        normalize_short_name("HEAD").unwrap_err().to_string(),
+        "Could not turn \"HEAD\" into a valid reference name",
+        "HEAD is rejected"
+    );
+    assert_eq!(
+        normalize_short_name("HEAD-").unwrap_err().to_string(),
+        "Could not turn \"HEAD-\" into a valid reference name",
+        "name that normalizes to HEAD is rejected"
```
slarse, line 58: According to Git, this branch name is actually fine, but we strip trailing dashes s.t. it gets caught by the HEAD check.

```bash
$ git check-ref-format --branch HEAD-
HEAD-
$ git check-ref-format --branch HEAD
fatal: 'HEAD' is not a valid branch name
```

Byron (approved):
Thanks a lot for picking this up and for the detailed analysis. It definitely got me thinking: Should this be coming from `gix`?

First things first: I think it's fine to fix this here and test for it as well, let's get it done.

> * What about case insensitive file systems (like macOS is by default, and I believe Windows as well)? Git does _not_ prevent creating a branch named e.g. `head`, but if you _do_ it starts printing warnings that "HEAD is ambiguous". For strict Git-compatibility, `head` (or `hEAd` or any other capitalization) is an allowed branch name, but for case insensitive file systems they seem to cause the same issues that `HEAD` as a branch name does. I feel like to some extent, Git just doesn't play nicely with case insensitive file systems, but maybe GitButler should?

With `gix`, it already plays nicer, in some way, take a look. In the example HEAD is `fadc92b2188b2edfff755f20cdabf3d8e42c7ff4`. This is on macOS

##### case-insensitive match
```
gitbutler ( master) [$] via 
❯ git rev-parse HEAd
fadc92b2188b2edfff755f20cdabf3d8e42c7ff4
```

##### Ambiguity that Git allows (for better or worse)

Now it's not possible anymore to access `HEAd` with `git`.

```
gitbutler ( master) [$] via 
❯ git rev-parse @~100
82634fc70519353220e98788fb5832dd88d12fd6

gitbutler ( master) [$] via 
❯ git branch HEAd 82634fc70519353220e98788fb5832dd88d12fd6

gitbutler ( master) [$] via 
❯ git rev-parse HEAD
warning: refname 'HEAD' is ambiguous.
fadc92b2188b2edfff755f20cdabf3d8e42c7ff4

gitbutler ( master) [$] via 
❯ git rev-parse HEAd
warning: refname 'HEAd' is ambiguous.
fadc92b2188b2edfff755f20cdabf3d8e42c7ff4
```

##### `gix` can do it though

This is why this issue is less pronounced for GitButler *I think*.

```
gitbutler ( master) [$] via 
❯ gix rev parse HEAd
82634fc70519353220e98788fb5832dd88d12fd6

gitbutler ( master) [$] via 
❯ gix rev parse HEAD
fadc92b2188b2edfff755f20cdabf3d8e42c7ff4
```

> * Why do case insensitive file systems exist?

Git 3.0 is planned for September this year, and that comes with reftable as default DB for reference. It's case-sensitive so this problem mostly goes away, or at least, is fully remedied with the implementation we have now.

> * After writing tests that strip trailing dashes all day, I realize that there's absolutely nothing in the refspec that prevents trailing dashes. _Leading_ dashes are prohibited in branch names specifically, but trailing ones are fine.

That could be a separate discussion. I'd personally think this is opinionated, but I'd wait for user complaints before changing it. Right now, it (`foo---` -> `foo`) might well be what people actually expect or prefer.

### Verdict

> Should this be coming from `gix`?

I think that `gix` should be better here. Right now, if `refs/heads/HEAD` also exists, we get this:

```
❯ gix rev parse HEAD
fadc92b2188b2edfff755f20cdabf3d8e42c7ff4

gitbutler ( master) [$] via 
❯ gix rev parse heads/HEAD
82634fc70519353220e98788fb5832dd88d12fd6
```

So it resolves `HEAD` first, but doesn't complain about ambiguity. I sense Byron didn't want to deal with this in greater detail.

Also, I think `gitoxide` should probably prohibit creating `refs/heads/HEAD` in the first place, and right now it obviously doesn't.

For that it would need specific checks in `gix-ref`, and it would also only be able to implement the rejection like Git does, without dealing with case-insensitivity.

The error message GitButler provides for `HEAD` refs is probably good enough, on a similar level to what Git provides today.

slarse:
> Thanks a lot for picking this up and for the detailed analysis. It definitely got me thinking: Should this be coming from `gix`?

@Byron thanks for the detailed response! I made a comment on your fix in `gix`. Without knowing the internals of `gix` too well, I think there is a little bit of danger to that change, but let's discuss that over in `gix`.

> That could be a separate discussion. I'd personally think this is opinionated, but I'd wait for user complaints before changing it. Right now, it (`foo---` -> `foo`) might well be what people actually expect or prefer.

In a broader sense, I actually think the auto-normalization is itself unexpected. In the GUI I guess it's fine to "help" the user a bit to create valid branch names (even though I think it's very unnecessary), but in the CLI I think it's outright harmful to try to automatically normalize the name. Because we don't currently normalize when operating on branches (as far as I know). So that's a very subtle way to create buggy scripts.

As for digressing from Git's ref spec, the fear I have with that is that we create functionality that is incompatible with Git, for no good reason. For example, let's say someone recognizes that we only normalize on creation but not when later operating on branches (e.g. when applying a branch to the workspace). If a helpful dev decides that it's better to make that consistent and normalize _all_ branch name inputs from the user, we suddenly wouldn't be able to operate on a subset of valid Git branches and would also create ambiguity between Git branches that previously was not there.

Sure, this is all a bit hypothetical, but I think there is merit to it. Especially the unexpectedness of getting a different branch name out than I put in to the CLI.

Byron:
> @Byron thanks for the detailed response! I made a comment on your fix in gix. Without knowing the internals of gix too well, I think there is a little bit of danger to that change, but let's discuss that over in gix.

Thanks, appreciated.

> In a broader sense, I actually think the auto-normalization is itself unexpected. In the GUI I guess it's fine to "help" the user a bit to create valid branch names (even though I think it's very unnecessary), but in the CLI I think it's outright harmful to try to automatically normalize the name. Because we don't currently normalize when operating on branches (as far as I know). So that's a very subtle way to create buggy scripts.

Yes, I see the point. Thanks for broadening my horizon on this. It really looks like CLI should be much more conservative when normalizing branches, to the point where it might not want to normalize them at all.

> Sure, this is all a bit hypothetical, but I think there is merit to it. Especially the unexpectedness of getting a different branch name out than I put in to the CLI.

Yes, I also arrive at the conclusion that normalizing branch names in the CLI is actively harmful and completely unexpected. This works on the graphical user interface, but not in a command-oriented application.

slarse:
> Yes, I also arrive at the conclusion that normalizing branch names in the CLI is actively harmful and completely unexpected. This works on the graphical user interface, but not in a command-oriented application.

That makes enough of us in agreement about this (@OliverJAsh and I have also briefly discussed it) that we should remove normalization from the CLI. And _maybe_ TUI, but IMO if it's in the GUI it should probably be in the TUI, too. @davidpdrsn ping, what do you think about automatic branch name normalization in the TUI?

PR #13159, opened 2026-04-02
fix(but): print normalized name after creating branch

slarse:
`but branch new` always created the new branch with a normalized name, but the name reported to users was the original non-normalized input, making it a li'l bit confusing.

Also took the opportunity to notify the user that the branch name was normalized.

Before:

before

After:
after

## TODO
Tests probably need revising as I realized that trailing dashes are actually fine by Git, it's just GitButler that strips them away for some reason. I think this is wrong.


slarse:
Has been effectively replaced by #13929

PR #13163, opened 2026-04-02
Add hidden `but remote` subcommand for local and remote browser access

mtsgrd:
## Summary

Adds a `but remote` subcommand that serves the embedded GitButler UI and API over HTTP.
The subcommand is available in dev and nightly builds only, gated behind the `remote` Cargo
feature. Release builds exclude `but-server` and the embedded frontend entirely.

- **`but remote`** (default) — opens a Cloudflare quick tunnel; all requests require a valid GitButler token
- **`but remote --local`** — binds to localhost only, no tunnel, no auth required
- **`but remote --name <tunnel> --hostname <host>`** — uses a pre-configured Cloudflare named tunnel with a custom domain
- **`but remote --remote-origin <url>`** — uses an existing reverse proxy / static origin instead of spawning cloudflared
- **`--dangerously-allow-anyone`** — disables auth entirely (LAN/dev use)

```sh
$ cd ~/my-repo
$ but remote
Starting cloudflare tunnel on port 8080
Tunnel: https://example-words-here.trycloudflare.com
```

```sh
$ cd ~/my-repo
$ but remote --local
Local: http://localhost:8080
```

## Security

- **DNS rebinding**: `--local` mode validates the `Host` header is `localhost`, `127.0.0.1`, or `[::1]`
- **Token auth**: validated against `app.gitbutler.com`; session cookie is `SameSite=Strict`
- **CORS**: explicit header/method allowlist, `allow_credentials(true)`, origin-locked in tunnel mode
- **CSRF**: `SameSite=Strict` cookie + Origin header required on all mutating requests (POST/PUT/DELETE/PATCH); requests without an Origin header are rejected with 403
- **Auth allowlist**: only `/auth/login` and `/auth/callback` are public; all other routes (including `/auth/logout`) require a valid session
- **Logout**: `POST /auth/logout` — using POST prevents CSRF-driven session termination
- **Login redirect**: `GET /auth/login` redirects to `/` when a valid session already exists, avoiding a login-page flash when following direct links
- **CSP**: header with per-script SHA-256 hashes on all responses
- **Compression**: response compression via `tower-http` `CompressionLayer`
- Fails fast if remote access is configured but no local GitButler user is logged in

## Auth / Login page

- Login flow: redirects to `app.gitbutler.com`, user pastes token back
- Async submit button with loading spinner
- `--dangerously-allow-anyone` flag bypasses auth for LAN/dev use

## Other

- Project is pinned to CWD on startup — `set_project_active` rejects other project IDs
- Mobile safe area: `viewport-fit=cover` + `env(safe-area-inset-bottom)` + `100dvh`
- Friendly error with install hint when `cloudflared` is missing
- `BUTLER_PORT` / `BUTLER_HOST` env vars preserved as fallbacks
- Remote origin normalised via `url::Url` (scheme + host + optional port only, path/query/fragment stripped)

## Test plan

- [ ] `cargo build -p but` builds cleanly
- [ ] `but remote --local` starts server, prints `Local: http://localhost:8080`, frontend loads to current repo
- [ ] `but remote` starts cloudflared, prints `Tunnel: https://….trycloudflare.com`, login required
- [ ] Visiting the tunnel URL on another device prompts login
- [ ] `--dangerously-allow-anyone` skips login
- [ ] Running in a repo auto-opens that project; switching projects is blocked
- [ ] Starting without a logged-in user (and no `--dangerously-allow-anyone`) exits with a clear error

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -828,6 +828,7 @@ dependencies = [
  "but-rules",
  "but-secret",
  "but-serde",
+ "but-server",
  "but-settings",
  "but-testsupport",
  "but-tools",
@@ -1583,6 +1584,7 @@ version = "0.0.0"
 dependencies = [
  "anyhow",
  "axum",
+ "base64 0.22.1",
  "but-api",
  "but-claude",
  "but-ctx",
@@ -1592,17 +1594,25 @@ dependencies = [
  "but-path",
  "but-secret",
  "but-settings",
+ "clap",
+ "colored",
  "futures-util",
  "gitbutler-repo-actions",
+ "gitbutler-user",
  "gitbutler-watcher",
+ "mime_guess",
+ "reqwest 0.12.28",
+ "rust-embed",
  "serde",
  "serde_json",
+ "sha2",
  "tokio",
  "tower",
  "tower-http",
  "tracing",
  "tracing-forest",
  "tracing-subscriber",
+ "url",
  "uuid",
 ]
 
@@ -2227,9 +2237,12 @@ version = "0.4.37"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "eb7b51a7d9c967fc26773061ba86150f19c50c0d65c887cb1fbe295fd16619b7"
 dependencies = [
+ "brotli",
  "compression-core",
  "flate2",
  "memchr",
+ "zstd",
+ "zstd-safe",
 ]
 
 [[package]]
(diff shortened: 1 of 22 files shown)


krlvi:
What if we made just one command for both? So but tunnel —local ?

mtsgrd:
Yeah, except local has nothing to do with tunneling? We could do `but remote --local`.

krlvi:
maybe similar to the irc functionality, this one is a good candidate to add a compile time feature flag on, wdyt?

krlvi (approved):
This is an experimental thing which is not compiled for the release version of the app

PR #13166, opened 2026-04-02
Misc tui fixes

davidpdrsn:
These commits should have been pushed with https://github.com/gitbutlerapp/gitbutler/pull/13165 but forgot to run `but push` before I merged.


davidpdrsn:
Hmm appears to have some flaky tests.

PR #13183, opened 2026-04-04
API: graphy squash

estib-vega:
Use the new squash command that's backed by graph operations in the CLI.

- Add an API to the but-api crate
- Add a binding for the API in question
- Use the new API in the but CLI

### DIFF
--- a/crates/but-api/src/commit/json.rs
+++ b/crates/but-api/src/commit/json.rs
@@ -5,7 +5,7 @@ use crate::{commit::types::CommitDiscardResult, json::HexHash};
 
 use super::types::{
     CommitCreateResult, CommitInsertBlankResult, CommitMoveResult, CommitRewordResult,
-    MoveChangesResult,
+    CommitSquashResult, MoveChangesResult,
 };
 
 /// UI type for a move changes between commits result.
@@ -129,6 +129,43 @@ but_schemars::register_sdk_type!(UICommitRewordResult);
 impl From<CommitRewordResult> for UICommitRewordResult {
     fn from(value: CommitRewordResult) -> Self {
         let CommitRewordResult {
+            new_commit,
+            replaced_commits,
+        } = value;
+
+        Self {
+            new_commit: new_commit.into(),
+            replaced_commits: replaced_commits
+                .into_iter()
+                .map(|(old, new)| (old.into(), new.into()))
+                .collect(),
+        }
+    }
+}
+
+/// UI type for squashing commits.
+#[derive(Debug, Serialize)]
+#[cfg_attr(feature = "export-schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "camelCase")]
+pub struct UICommitSquashResult {
+    /// The new commit ID after squashing.
+    #[cfg_attr(feature = "export-schema", schemars(with = "String"))]
+    pub new_commit: HexHash,
+    /// Commits that have been replaced as a side-effect of the squash.
+    /// Maps `oldId -> newId`.
+    #[cfg_attr(
+        feature = "export-schema",
+        schemars(with = "std::collections::BTreeMap<String, String>")
+    )]
+    pub replaced_commits: std::collections::BTreeMap<HexHash, HexHash>,
+}
+
+#[cfg(feature = "export-schema")]
+but_schemars::register_sdk_type!(UICommitSquashResult);
+
+impl From<CommitSquashResult> for UICommitSquashResult {
+    fn from(value: CommitSquashResult) -> Self {
+        let CommitSquashResult {
             new_commit,
             replaced_commits,
         } = value;
--- a/crates/but-api/src/commit/squash.rs
+++ b/crates/but-api/src/commit/squash.rs
@@ -1,0 +1,108 @@
+use but_api_macros::but_api;
+use but_core::sync::RepoExclusive;
+use but_oplog::legacy::{OperationKind, SnapshotDetails};
+use but_rebase::graph_rebase::{Editor, LookupStep as _};
+use tracing::instrument;
+
+use super::types::CommitSquashResult;
+
+/// Squash `subject_commit_id` into `target_commit_id`.
+///
+/// This acquires exclusive worktree access from `ctx` before rewriting the
+/// commits.
+///
+/// For details, see [`commit_squash_only_with_perm()`].
+#[but_api(crate::commit::json::UICommitSquashResult)]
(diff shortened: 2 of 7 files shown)


crates/but/src/command/legacy/rub/squash.rs
```diff
@@ -274,9 +275,26 @@ fn squash_commits_internal(
...
+
+        let squash_result = but_api::commit::squash::commit_squash_only_with_perm(
+            ctx,
+            remapped_source_oid,
+            new_commit_oid,
+            perm,
+        )?;
+        rewritten_commits.extend(squash_result.replaced_commits.iter().map(|(k, v)| (*k, *v)));
+        new_commit_oid = squash_result.new_commit;
+    }
```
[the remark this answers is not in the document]
estib-vega: yeeeep

crates/but/src/command/legacy/rub/squash.rs
```diff
@@ -274,44 +279,78 @@ fn squash_commits_internal(
...
-        squash_commits_with_perm(ctx, target_stack, source_oids.clone(), target_oid, perm)?;
+    // Keep the complete multi-rewrite sequence atomic: if any step fails,
+    // restore the pre-squash snapshot so we don't leave partial rewrites.
+    let snapshot = ctx.create_snapshot(SnapshotDetails::new(OperationKind::SquashCommit), perm)?;
+    let squash_result: anyhow::Result<ObjectId> = (|| {
+        // Perform the squash by folding sources into the current target one by one.
+        // We process from bottom to top so each next source remains adjacent to the
+        // rewritten target commit.
+        let mut new_commit_oid = target_oid;
+        let mut rewritten_commits = BTreeMap::<ObjectId, ObjectId>::new();
```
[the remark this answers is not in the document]
estib-vega: This will be addressed later, when multi-selection is supported

crates/but-api/src/commit/json.rs
```diff
@@ -143,6 +143,43 @@ impl From<CommitRewordResult> for UICommitRewordResult {
     }
 }
 
+/// UI type for squashing commits.
+#[derive(Debug, Serialize)]
+#[cfg_attr(feature = "export-schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "camelCase")]
+pub struct UICommitSquashResult {
```
Byron, line 150: Can we just not prefix these types with `UI` anymore to not add to the work that has to be done to remove the prefix from the existing types?
While consistency is valuable, it shouldn't add to the 'wrong' kind of consistency.
Maybe you are in a good position to perform such a rename as well so this gets fixed for good.

crates/but-api/src/commit/squash.rs
```diff
@@ -0,0 +1,108 @@
...
+/// create an oplog entry. For lower-level implementation details, see
+/// [`but_workspace::commit::squash_commits()`].
+pub fn commit_squash_only_with_perm(
+    ctx: &mut but_ctx::Context,
+    subject_commit_id: gix::ObjectId,
+    target_commit_id: gix::ObjectId,
+    perm: &mut RepoExclusive,
+) -> anyhow::Result<CommitSquashResult> {
+    let mut meta = ctx.meta()?;
+    let (repo, mut ws, _, _cache) = ctx.workspace_mut_and_db_and_cache_with_perm(perm)?;
```
Byron, line 45: This encourages me to roll the `cache` into `db` as calling a method that produces a cache just to ignore it makes nose sense.
Once `meta` is rolled into `db` as well, then this will become…


```suggestion
    let (repo, mut ws, db) = ctx.workspace_mut_and_db_with_perm(perm)?;
```

…which seems more manageable.

crates/but-api/src/commit/types.rs
```diff
@@ -26,6 +26,14 @@ pub struct CommitRewordResult {
     pub replaced_commits: BTreeMap<gix::ObjectId, gix::ObjectId>,
 }
 
+/// Outcome of squashing one commit into another.
+pub struct CommitSquashResult {
+    /// The ID of the newly created squashed commit.
+    pub new_commit: gix::ObjectId,
+    /// Commits that were replaced by this operation. Maps `old_id -> new_id`.
+    pub replaced_commits: BTreeMap<gix::ObjectId, gix::ObjectId>,
```
Byron, line 34: I vaguely remember to have changed one of these to `Vec`, probably as they were a HashMap before which is unstable.

The same isn't true for `BTreeMap`, so if you plan a cleanup, then I think `UIMoveChangesResult` and `MoveChangesResult` can also be updated with `BTreeMap`.

crates/but-api/src/commit/json.rs
estib-vega: Yesss, let's actually do that for all of them. Fixing that

crates/but-api/src/commit/squash.rs
estib-vega: Will fix

crates/but-api/src/commit/types.rs
estib-vega: Can replace that 👍

crates/but-api/src/commit/json.rs
estib-vega: https://github.com/gitbutlerapp/gitbutler/pull/13192

crates/but-api/src/commit/squash.rs
estib-vega: But, yeah. It also makes sense to have the cache be part of the db. There's a bunch of `workspace_and_*` and I have little to no context on when to use which one

crates/but-api/src/commit/types.rs
estib-vega: All I found seem to be using BTreeMap already

crates/but-api/src/commit/squash.rs
Byron: Assuming the `*and_cache` variants will be gone, the other ones unfortunately can't go away. Then choosing which one to use is about whether or not a `mut` is needed for either `ws` or `db`. The other variable is whether to get a guard or use `perm` from an existing `guard`. In future, I suppose most of the time one will use the `*with_perm()` version.

crates/but-api/src/commit/types.rs
Byron: When I look for `replaced_commits: Vec` I still find these 2 mentioned above. Not a big deal, but probably good to investigate if BTreeMap makes sense there.

PR #13190, opened 2026-04-06
Move project routes to /p/[projectId]/ prefix

mtsgrd:
Move project routes to a `/p/[projectId]/` URL prefix so the router
never needs to distinguish a project ID from sub-route names like
`workspace` or `branches`. Combined with a `(pinned)` route group, this
also gives clean `/workspace`, `/branches`, etc. URLs when a project is
server-side pinned via `but remote`.

**Routing**
- Add `routes/p/[projectId]/` with layout and all sub-pages
- Add a `(pinned)` route group for clean URLs in pinned mode
- Extract the project layout into `ProjectPageLayout.svelte`, shared by
  both the `/p/` and `(pinned)` layout wrappers
- `routes.svelte.ts`: `prefix()` returns `/p/{id}` normally and `""`
  when pinned; `isUrl()` covers both route ID forms
- Remove the param matcher file — `/p/[projectId]` accepts any string,
  so ProjectHandle values (base64url-encoded paths starting with `L`)
  work without a custom matcher
- Root `+page.svelte` redirects to `/workspace` when pinned

**`but remote` fixes**
- Inject a `gitbutler-server-base` meta tag into served `index.html` so
  `webInvoke` uses the correct API base path at runtime; fixes 404s in
  `--local` mode
- Move `embedded-frontend` to the `but` `remote` Cargo feature (not
  `legacy`) so release builds exclude server assets; `build.rs` handles
  a missing frontend directory gracefully so `cargo check --workspace`
  works without a prior frontend build
- Fix `ProjectSetup.svelte` to use the `/p/` prefix when redirecting
  after project creation

---
- <kbd>&nbsp;2&nbsp;</kbd> #13190 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #13163


Byron:
Newly added projects don't have UUIDs anymore, the have `ProjectHandle`'s which are the encoded paths. The backend supports both for now, with the plan to migrate to `ProjectHandle` at some point to remove UUIDs from the system entirely.

mtsgrd:
Thanks @Byron, let me investigate.

mtsgrd:
Closing this because, it's premature.

PR #13201, opened 2026-04-07
add-kitty-terminal

### LINKED ISSUE #13067: Feature Request: Add Kitty to supported terminals
Hi! I use [Kitty](https://sw.kovidgoyal.net/kitty/) as my main terminal.

It would be great to add it to the list of supported terminals.

sebastianhuus:
## 🧢 Changes

- Adds Kitty to supported terminals on MacOS
- Adds Kitty to supported terminals on Linux.

## ☕️ Reasoning

The "Open in terminal" feature already implements a bunch of terminals, but we are not supporting the Kitty terminal. This is a oversight.

Windows is not supported by Kitty, so not in scope.

Related to https://github.com/gitbutlerapp/gitbutler/pull/11878

## 🎫 Affected issues

Fixes: #13067

## 📌 Todos
- [x] Test Kitty on MacOS
- Works
- [x] Test Kitty on Linux
- [x] Squash commits into single commit

For testing the changes, you can download Kitty as instructed on [their website](https://sw.kovidgoyal.net/kitty/binary/)
`curl -L https://sw.kovidgoyal.net/kitty/installer.sh | sh /dev/stdin`

Then select your terminal in the GB settings and try the Open in Terminal command

### DIFF
--- a/crates/but-api/src/legacy/open.rs
+++ b/crates/but-api/src/legacy/open.rs
@@ -216,8 +216,10 @@ pub fn open_in_terminal(terminal_id: String, path: String) -> Result<()> {
                 .status()
                 .context("Failed to run 'open -Ra' to check application availability")?;
             if !status.success() {
-                return Err(anyhow::anyhow!("'{app_name}' was not found.")
-                    .context(but_error::Code::DefaultTerminalNotFound));
+                return Err(anyhow::anyhow!(
+                    "'{app_name}' was not found - `open -Ra {app_name}` failed."
+                )
+                .context(but_error::Code::DefaultTerminalNotFound));
             }
             Ok(())
         }
--- a/crates/but-api/src/legacy/open.rs
+++ b/crates/but-api/src/legacy/open.rs
@@ -118,6 +118,7 @@ pub fn open_url(url: String) -> Result<()> {
 /// - `alacritty-mac` - Alacritty
 /// - `wezterm-mac` - WezTerm
 /// - `hyper` - Hyper
+/// - `kitty` - Kitty
 ///
 /// **Windows:**
 /// - `wt` - Windows Terminal
@@ -133,6 +134,7 @@ pub fn open_url(url: String) -> Result<()> {
 /// - `warp` - Warp
 /// - `hyper` - Hyper
 /// - `wezterm` - WezTerm
+/// - `kitty` - Kitty
 ///
 /// # Errors
 /// Returns an error if:
@@ -245,6 +247,7 @@ pub fn open_in_terminal(terminal_id: String, path: String) -> Result<()> {
                     .arg(&path);
                 run_terminal_command(cmd, "Alacritty", &path)?;
             }
+            "kitty" => open_with_path("kitty", Some("Kitty"))?,
             // WezTerm does not support `open -a WezTerm <path>`. Their docs state you have to use their CLI.
             // https://wezterm.org/config/launch.html#specifying-the-current-working-directory
             "wezterm-mac" => {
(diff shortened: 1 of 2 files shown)


sebastianhuus:
I find that the "Open" command will open a new window every time instead of hijacking a terminal. Based on the previous PR, I assume that hijacking is undesirable as per the comment below:

> Fix issue where Alacritty will hijack existing terminal sessions if present, by using -n flag to force a new window.

Therefore I'll assume that new window every time is good for this particular integration, as Kitty does not support tabs like some other emulators do :)

krlvi:
This seems reasonable! If/when you make this non-draft, id be happy to merge

sebastianhuus:
I found that Kitty has two primary methods of installation with different solutions for us. 

1) Installed via the website and curl:
- This installs the app to `~/.local/kitty.app`
2) Install via `sudo apt install kitty`:
- This installs the app to PATH and will work with all the other methods. 

I implemented both, so first we try running it from path and then we try the .local variant. Tested on Linux and worked for me.

Anyone wanna try it before I finalize and take this out of draft?

crates/but-api/src/legacy/open.rs
```diff
@@ -266,21 +269,32 @@ pub fn open_in_terminal(terminal_id: String, path: String) -> Result<()> {
...
+            // Kitty installs to ~/.local/kitty.app/bin/kitty by default on Linux.
+            if terminal_id == "kitty" {
+                let kitty_default = std::path::PathBuf::from(std::env::var("HOME").ok()?)
+                    .join(".local/kitty.app/bin/kitty");
+                if kitty_default.is_file() {
+                    return Some(kitty_default);
+                }
+            }
+            None
+        });
```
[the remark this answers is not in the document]
sebastianhuus: Previously we already did `let binary_found = which::which(binary).is_ok();` so not relevant imo. We just add `.or_else` to handle the Kitty case.

crates/but-api/src/legacy/open.rs
```diff
@@ -266,21 +269,32 @@ pub fn open_in_terminal(terminal_id: String, path: String) -> Result<()> {
...
-        let binary_found = which::which(binary).is_ok();
-        if !binary_found {
+        // Resolve the full path to the terminal binary.
+        // Most terminals are on PATH, but some (e.g. kitty) install to a user-local directory
+        // that is only added to PATH via shell config, which GUI apps don't source.
+        let binary_path = which::which(binary).ok().or_else(|| {
+            // Kitty installs to ~/.local/kitty.app/bin/kitty by default on Linux.
+            if terminal_id == "kitty" {
+                let home = std::path::PathBuf::from(std::env::var("HOME").ok()?);
+                let kitty_default = home.join(".local/kitty.app/bin/kitty");
```
krlvi: Why do we need full path here? If kitty is on the path all we need is the exec name, like with the other terminals, could i be missing something?

sebastianhuus: As per [Kitty docs ](https://sw.kovidgoyal.net/kitty/binary/)

> The binaries will be installed in the standard location for your OS, [...] and ~/.local/kitty.app on Linux.

When I ran this on my Linux machine, GB and my terminal could not find it. Quick google shows this installer does not modify path at all. 

However if you use `apt` then it will install to path. 

This is to support both cases.

krlvi: It seems like they leave the path editing to be up to the user. IMO, if the user has chosen to to have the terminal bin on the path, so be it. 
Something about adding path composition here makes me feel uncomfortable, from a security / potential exploit standpoint. Can we just have "kitty" here, similar to the other terminal options?

sebastianhuus: Ohh makes sense. Didn't see it that way, can indeed be exploited if someone distributes a malicious app with that location.

How about either appending all the "terminal not found" error messages or adding an extra one for Kitty to help users a bit? Currently the error messages do not mention needing it in path at all. I have a feeling we can get another issue for "Kitty not working even though installed", so just adding a hint that they need to add it to PATH can probably help a lot too. 

For reference, here is the current error toast with no mention of needing it accessible on path:
image

Here's an image of the selection area in Settings where we also have no indication that it should be in path. 
image

I think the best solution is to just update the Rust error message to mention path for all Linux "terminal not found" errors. It is a one line change inside of a Linux-scoped cfg block. 

```rust 
// crates/but-api/src/legacy/open.rs
return Err(anyhow::anyhow!("'{binary}' was not found. Make sure it is installed and available on your PATH.")

```

sebastianhuus:
I've made the changes as discussed. Lemme know if you disagree on the error message. On one hand, less verbose is good - but on the other I think it makes for a better experience as a user

sebastianhuus:
Fixed lint issue with the new message.

Byron (approved):
This is great, thanks a lot!

Image

I have modified the error message on macOS to be akin to the one on Linux, making it more descriptive to show what we tried to open the terminal or to imply what needs to work so that the terminal can be found.

sebastianhuus:
The PR was blocked by my initial commit not being signed. I set up GPG signing on my GitHub but struggled a bit with the signing. Managed to apply a sign to your commit @Byron which seems to have made it partially signed. Unsure if this'll keep blocking merge (but at least now I got signing set up for next time 😎)

Byron:
Great!

I didn't notice that last time, so my auto-merge didn't go through. When you push, it also gets auto-disabled.

The partial verification is interesting - after all I am just a co-author and it gets a bit tricky then. But it's fine to merge anyway.

Screenshot 2026-04-13 at 07 58 32

Byron:
Merging, assuming that the blackbox test failure is just a flake.

PR #13205, opened 2026-04-07
Fix shutdown deadlock in file-monitor/BlockingPool interaction

mtsgrd:
## Problem

When the Tokio runtime shuts down, `BlockingPool::shutdown()` blocks the main thread waiting for all `spawn_blocking` tasks to complete. The file monitor (`gitbutler_filemonitor::spawn`) runs as one such blocking task, looping indefinitely until `cmd_rx` is disconnected (i.e. until the `FileMonitorHandle` that holds `cmd_tx` is dropped).

Previously, `FileMonitorHandle` was owned by the `tokio::spawn` async event-dispatch task — not by `WatcherHandle`. The intended shutdown chain was:

```
WatcherHandle drops
  → CancellationToken fires
    → async task sees cancellation, exits loop
      → async task drops FileMonitorHandle
        → cmd_rx disconnects
          → spawn_blocking task exits
```

The flaw: Tokio's `Runtime::drop()` calls `BlockingPool::shutdown()` *before* it drops the scheduler (and thus before async tasks are cleaned up). The async task can't run to drop `FileMonitorHandle` because the scheduler hasn't been torn down yet — and the scheduler won't be torn down because the blocking pool is waiting. A classic circular dependency:

```
BlockingPool::shutdown()  [main thread, waiting]
  ← blocked on spawn_blocking file monitor task
    ← needs cmd_rx disconnect
      ← needs FileMonitorHandle drop
        ← needs async task to run
          ← needs scheduler teardown
            ← needs BlockingPool to finish first
```

## How the deadlock was found

A `gitbutler-tauri` process froze after a user navigated to a newly added project. The macOS `sample` tool showed the main thread permanently blocked in `BlockingPool::shutdown → parking_lot::Condvar::wait`. No blocking tasks appeared to be actively running, yet the pool never drained. Cross-referencing the tracing logs (`time.busy=14664s` on the `file monitor` span) confirmed the blocking task had been running for hours before the shutdown attempt.

## Fix

Move `FileMonitorHandle` from the async task into `WatcherHandle` directly. `WatcherHandle::drop` now drops it synchronously — no async scheduling required — immediately disconnecting `cmd_rx` and letting the blocking task exit. This breaks the circular dependency at the earliest possible point.

As a side effect, the `flush()` path is simplified: rather than sending a signal through an async `unbounded_channel` to the event-dispatch task (which then called `monitor.flush()`), `WatcherHandle::flush` now calls `file_monitor.flush()` directly, removing the need for the `flush_tx`/`flush_rx` pair entirely.

### DIFF
--- a/crates/gitbutler-watcher/src/lib.rs
+++ b/crates/gitbutler-watcher/src/lib.rs
@@ -8,16 +8,13 @@ use anyhow::Result;
 use but_ctx::ProjectHandleOrLegacyProjectId;
 use but_settings::AppSettingsWithDiskSync;
 pub use handler::Handler;
-use tokio::{
-    sync::mpsc::{UnboundedSender, unbounded_channel},
-    task,
-};
+use tokio::{sync::mpsc::unbounded_channel, task};
 use tokio_util::sync::CancellationToken;
 
 mod events;
 
 pub use events::Change;
-use gitbutler_filemonitor::InternalEvent;
+use gitbutler_filemonitor::{FileMonitorHandle, InternalEvent};
 
 mod handler;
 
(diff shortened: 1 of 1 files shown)


Byron:
Thanks for the ping!

Now that `tokio::select!()` only deals with the cancellation token, I think it's worth checking if this crate needs `tokio` at all. All this snuck in due to `tauri` back in the days, and I don't think it was ever overhauled for maximum simplicity. All of our `tauri` commands are considered blocking anyway, so no need to use `async fn` and tokio unless there is a real advantage.

It seems Opus would be able to perform such a simplification, and if it works, great, if it freezes, it would be able to sort out a fix :D.

PR #13214, opened 2026-04-07
Follow but-api conventions in `assign_hunk`

davidpdrsn:
I noticed that `but_api::diff::assign_hunk` didn't create a snapshot and didn't follow the convention of having `_only`, `_with_perm`, and `_only_with_perm` variants. This should fix that.

There seem to be some bugs happening in `but undo` (see https://discord.com/channels/1060193121130000425/1442962582444183593/1491124718584070346) so couldn't actually test the snapshotting.

---
- <kbd>&nbsp;2&nbsp;</kbd> #13215
- <kbd>&nbsp;1&nbsp;</kbd> #13214 👈


krlvi (approved):
We can totally do that for consistency, but this oplog entry will be noop (i.e. restoring it does nothing). The reason is that the oplog data format currently does not capture the the assignments. 
Still good to merge this since we wanna change this in the future and we wanna have this consistent, but maybe worth adding a comment in the code

PR #13215, opened 2026-04-07
Add non-legacy function for undoing a commit

davidpdrsn:
Adds `but_api::commit::undo::commit_undo` (and friends). I need this in the TUI for the non-legacy version of rub I'm cooking.

---
- <kbd>&nbsp;2&nbsp;</kbd> #13215 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #13214

### DIFF
--- a/crates/but-api/src/commit/undo.rs
+++ b/crates/but-api/src/commit/undo.rs
@@ -1,0 +1,123 @@
+use std::collections::BTreeMap;
+
+use anyhow::Context as _;
+use but_api_macros::but_api;
+use but_core::{DiffSpec, sync::RepoExclusive};
+use but_oplog::legacy::{OperationKind, SnapshotDetails, Trailer};
+use tracing::instrument;
+
+use crate::commit::types::{CommitUndoResult, MoveChangesResult};
+
+/// Undo `subject_commit_id` using the behavior described by [`commit_undo_only_with_perm()`].
+#[but_api(napi, crate::commit::json::CommitUndoResult)]
+#[instrument(err(Debug))]
+pub fn commit_undo(
+    ctx: &mut but_ctx::Context,
+    subject_commit_id: gix::ObjectId,
+) -> anyhow::Result<CommitUndoResult> {
+    let mut guard = ctx.exclusive_worktree_access();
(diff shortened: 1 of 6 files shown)


crates/but-api/src/commit/undo.rs
```diff
@@ -0,0 +1,116 @@
...
+
+        let mut parent_ids = commit.parent_ids();
+        let first_parent = parent_ids.next().map(|id| id.detach());
+
+        // TODO: do we want to handle this?
+        anyhow::ensure!(
+            parent_ids.next().is_none(),
+            "expected {} to have at most one parent",
+            subject_commit_id.to_hex()
+        );
```
davidpdrsn, line 75: 👀

krlvi (approved):
Nice

crates/but-api/src/commit/undo.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+/// Undo `subject_commit_id`, under caller-held exclusive repository access.
+///
+/// This will move the changes in the commit to be unassigned and discard the commit.
+pub fn commit_undo_only_with_perm(
+    ctx: &mut but_ctx::Context,
+    subject_commit_id: gix::ObjectId,
+    perm: &mut RepoExclusive,
+) -> anyhow::Result<CommitUndoResult> {
```
Byron, line 62: @davidpdrsn Are the 'pure' versions of this API call actually used? It *feels* like we are always calling the version with side-effects.

If so, I don't think we have to add anything we don't have use for.

davidpdrsn: No I dont think so. I only use `commit_undo` in the TUI. It just seemed to me this was the convention since all other API functions were done this way.

Byron: It started out as needed for one specific case, but I don't think we should duplicate it.
Sorry for being late, that feedback would have been useful earlier. In any case, simplification would be appreciated so there is less chance of this stuff being duplicated in future.
Maybe there can be a sweep that does this for everything (but I am out of tokens for a while 😂).

PR #13230, opened 2026-04-08
Fix broken CI on master: trigger lint-node on e2e/playwright changes

mtsgrd:
Add `e2e/**` to the `node` paths filter so that changes to files like
`e2e/playwright/src/setup.ts` trigger `lint-node` in CI. This fixes
broken CI on master where a formatting issue in that file went undetected
because the job wasn't triggered for changes outside `apps/**` and `packages/**`.

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -44,6 +44,7 @@ jobs:
               - *workflows
               - 'apps/**'
               - 'packages/**'
+              - 'e2e/**'
               - 'package.json'
               - 'pnpm-lock.yaml'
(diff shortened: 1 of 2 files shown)


krlvi (approved):

.github/workflows/push.yaml
```diff
@@ -44,6 +44,7 @@ jobs:
               - *workflows
               - 'apps/**'
               - 'packages/**'
+              - 'e2e/**'
```
mtsgrd, line 47: @krlvi fyi

PR #13239, opened 2026-04-09
fix: escape GIT_SSH before promoting to GIT_SSH_COMMAND

slarse:
## 🧢 Changes
Title pretty much sums it up.

## 🎫 Affected issues
Fixes: #13088

### DIFF
--- a/crates/gitbutler-git/src/repository.rs
+++ b/crates/gitbutler-git/src/repository.rs
@@ -310,17 +310,33 @@ where
         .map_err(Error::<E>::Exec)
 }
 
+/// Resolve a command to be set in GIT_SSH_COMMAND. We need to use GIT_SSH_COMMAND specifically as
+/// we add command line options.
+///
+/// It's important that the resolved command respects the requirements of GIT_SSH and
+/// GIT_SSH_COMMAND, see [the Git docs](https://git-scm.com/docs/git#Documentation/git.txt-GITSSH)
+/// for details.
 fn resolve_git_ssh_command<P>(harness_env: &HarnessEnv<P>, envs: &HashMap<String, String>) -> String
 where
     P: AsRef<Path>,
 {
-    let base_ssh_command = match envs
-        .get("GIT_SSH_COMMAND")
-        .cloned()
-        .or_else(|| envs.get("GIT_SSH").cloned())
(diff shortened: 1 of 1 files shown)


slarse:
Okay, so after toying around with just escaping GIT_SSH, I've found that this whole thing of taking the user's configuration and haphazardly modifying it is just a plain bad idea.

Escaping hassle aside, if the user has set any of GIT_SSH, GIT_SSH_COMMAND or `core.sshCommand`, it's not safe for us to just add options to it (which is the entire purpose of promoting GIT_SSH to GIT_SSH_COMMAND).

For example, the user in #13088 is using `plink`. Even if we escape the path properly, `plink` still doesn't accept the additional OpenSSH-specific options we append (`-o StrictHostKeyChecking=accept-new -o KbdInteractiveAuthentication=no`), so it errors out even when the executable is correctly resolved.

So my take is that we should only set `GIT_SSH_COMMAND` to our preferred default if the user has configured _none_ of `GIT_SSH`, `GIT_SSH_COMMAND` or `core.sshCommand`.

I'll leave this PR here for reference and start over on a new fix.

crates/gitbutler-git/src/repository.rs
```diff
@@ -346,6 +362,29 @@ where
...
+    git_ssh_command: Option<P>,
+    git_ssh: Option<P>,
+) -> Option<String> {
+    match (git_ssh_command, git_ssh) {
+        (Some(git_ssh_command), _) => Some(git_ssh_command.as_ref().to_string()),
+        (_, Some(git_ssh)) => {
+            // Escape the path with single quotes so it's safe to execute by a POSIX shell.
+            //
+            // Even on Windows, Git uses a POSIX(ish) shell to execute GIT_SSH_COMMAND, see https://github.com/git/git/blob/b15384c06f77bc2d34d0d3623a8a58218313a561/run-command.c#L277-L286
+            let escaped_git_ssh = git_ssh.as_ref().replace("'", r"'\''");
```
Byron, line 381: You might be able to use https://docs.rs/gix-quote/latest/gix_quote/fn.single.html .

slarse: Thanks! Good to know that's there.

PR #13251, opened 2026-04-10
fix(gitbutler-git): avoid tampering with existing SSH configuration

slarse:
Historically, we have unconditionally 1) promoted GIT_SSH to GIT_SSH_COMMAND, and 2) appended OpenSSH-specific options to the command.

The GIT_SSH->GIT_SSH_COMMAND promotion caused issues with escaping of GIT_SSH, as GIT_SSH is explicitly a _program_ (path or just a name to be looked up on PATH), while GIT_SSH_COMMAND is an arbitrary shell command. So, for example, unescaped whitespace in GIT_SSH is fine but causes GIT_SSH_COMMAND to break.

The appending of OpenSSH-specific options causes alternative clients (such as plink.exe on Windows) to break.

To resolve these issues, we take a more conservative approach and only apply our own preferred defaults for OpenSSH if we cannot find any configuration for the SSH behavior in the current environment.

Replaces https://github.com/gitbutlerapp/gitbutler/pull/13239

Fixes https://github.com/gitbutlerapp/gitbutler/issues/13088

### DIFF
--- a/crates/gitbutler-git/src/repository.rs
+++ b/crates/gitbutler-git/src/repository.rs
@@ -195,8 +195,7 @@ where
     envs.insert("SSH_ASKPASS".into(), askpath_path.display().to_string());
     envs.insert("SSH_ASKPASS_REQUIRE".into(), "force".into());
 
-    let git_ssh_command = resolve_git_ssh_command(&harness_env, &envs);
-    envs.insert("GIT_SSH_COMMAND".into(), git_ssh_command);
+    ensure_git_ssh_is_configured(&harness_env, &mut envs);
 
     let cwd = match harness_env {
         HarnessEnv::Repo(p) | HarnessEnv::Global(p) => p,
@@ -295,10 +294,7 @@ where
     E: GitExecutor,
 {
     let mut envs = envs.unwrap_or_default();
-    envs.insert(
-        "GIT_SSH_COMMAND".into(),
-        resolve_git_ssh_command(&harness_env, &envs),
-    );
+    ensure_git_ssh_is_configured(&harness_env, &mut envs);
 
     let cwd = match harness_env {
         HarnessEnv::Repo(p) | HarnessEnv::Global(p) => p,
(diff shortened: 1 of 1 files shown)


slarse:
@Byron Please weigh in on this approach. I think it's better to respect user settings than to attempt to "augment" them, but there may be historical reasons for how this has worked that I'm not aware of.

crates/gitbutler-git/src/repository.rs
```diff
@@ -1,5 +1,6 @@
 use std::{collections::HashMap, path::Path, time::Duration};
 
+use anyhow::Result;
```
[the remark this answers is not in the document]
slarse: that wasn't even supposed to be there :)

crates/gitbutler-git/src/repository.rs
```diff
@@ -310,24 +301,48 @@ where
...
+    // It's entirely possible to have non-unicode byte sequences in paths, for example. This should
+    // be rewritten to use `std::env::var_os()` and the executor should be tweaked to let `envs`
+    // contain raw binary strings. This is however exceedingly unlikely to be a problem in practice
+    // so we'll keep it like this for now.
+    if let Ok(git_ssh_command) = std::env::var("GIT_SSH_COMMAND") {
+        envs.insert("GIT_SSH_COMMAND".into(), git_ssh_command);
+        return envs;
+    }
+
+    if let Ok(Some(git_ssh_command)) = get_core_sshcommand(harness_env) {
```
Byron: Just as a note: This automatically respects `GIT_SSH_COMMAND` despite asking for the `core.sshCommand` as `gitoxide` automatically lets env-vars override their respective Git configuration. This is done so that by default, users of these abstractions can act more Git like (while offering an opt-out that we don't use).

This means that the extra check for `GIT_SSH_COMMAND` above isn't needed in the case where the harness is non-global.

crates/gitbutler-git/src/repository.rs
Byron: One could consider fixing this to rather return `None` if the string conversion fails. This could be a bug right now that yields strange errors if the SSH command is not unicode compatible.

Byron (approved):
Interestingly, I also lack the historical knowledge to tell why the SSH command was always overridden, but agree that it's certainly better to not do that.
In the worst case, we re-learn what caused this in the first place.

While I have some notes, I think it's fine to merge already.

PS: Ideally, one fine day we can modernise this and remove abstractions and async, it seems overly complex for what it does, and this has been on my list forever.

crates/gitbutler-git/src/repository.rs
slarse: Agreed, I did not spot that lossy string conversion. That's a very subtle bug in the making, I'll fix it.

crates/gitbutler-git/src/repository.rs
slarse: Ohh, okay, I didn't know that. I'll clean this up.

slarse (commented):
@Byron 

> PS: Ideally, one fine day we can modernise this and remove abstractions and async, it seems overly complex for what it does, and this has been on my list forever.

You mean in terms of the "indirect askpass auth" execution route? I'd be all for it, but I don't have any great ideas for simplification as long as we're shelling out to Git and need to play askpass server for a little bit. Do you have something off the top of your head?

As we now need this behavior for the TUI to play nicely as well, now would be a great time to rethink this. I'll give it a think as well.

Byron:
> You mean in terms of the "indirect askpass auth" execution route? I'd be all for it, but I don't have any great ideas for simplification as long as we're shelling out to Git and need to play askpass server for a little bit. Do you have something off the top of your head?

I was not thinking in factional terms, only in structural ones. My gripe is with the trait which to my mind complicates understanding and following what's really happening. Also I don't think `tokio` was ever needed for this. It just slipped in because of `tauri`. So there I think we can be just using threads.

> As we now need this behavior for the TUI to play nicely as well, now would be a great time to rethink this. I'll give it a think as well.

One day maybe this crate wants to become `but`-prefixed and then it should definitely receive the complexity reduction that seems so feasible here.

PR #13261, opened 2026-04-10
Lite: tidy up tagged unions

OliverJAsh:
If we change our minds about using Effect's `Data.TaggedEnum` we can use agents to easily migrate away later on.


OliverJAsh:
Redux serialization 😭

PR #13265, opened 2026-04-10
Add a dry-run flag to most mutations

Caleb-T-Owens:
This is quite a wide PR, but hopfully it’s pretty easy to follow.

The bulk is in `crates/but-api/src/workspace_state.rs` where we define some new helper functions that enable us to build a workspace response from an either in-memory workspace or a materialized one.

The rest is mostly just wiring up the new flags into the frontends and using the new functions.


krlvi (approved):
lgtm

crates/but-api/src/workspace_state.rs
```diff
@@ -0,0 +1,79 @@
+use std::collections::BTreeMap;
+
+use but_core::RefMetadata;
+use but_rebase::graph_rebase::SuccessfulRebase;
+
+use crate::types::WorkspaceState;
+
+/// Build a [`WorkspaceState`] from an already-prepared overlayed graph.
+///
+/// Use this when the caller already has a graph describing the workspace after the
```
Byron:
```suggestion
/// Use this when the caller already has a `graph` describing the workspace after the
```

crates/but-api/src/workspace_state.rs
```diff
@@ -0,0 +1,79 @@
+use std::collections::BTreeMap;
+
+use but_core::RefMetadata;
+use but_rebase::graph_rebase::SuccessfulRebase;
+
+use crate::types::WorkspaceState;
+
+/// Build a [`WorkspaceState`] from an already-prepared overlayed graph.
```
Byron:
```suggestion
/// Build a [`WorkspaceState`] from an already-prepared overlayed `graph`.
```

crates/but-api/src/workspace_state.rs
```diff
@@ -0,0 +1,79 @@
...
+///
+/// Use this when the caller already has a graph describing the workspace after the
+/// intended operation, regardless of whether that graph came from a preview, a
+/// materialized rebase, or another graph-producing workflow. The caller is
+/// responsible for supplying the matching `replaced_commits` map for that graph.
+///
+/// This is the most direct constructor in this module and is the right choice when
+/// there is no need to inspect or materialize a [`SuccessfulRebase`].
+pub(crate) fn from_overlayed_graph(
```
Byron: Since `WorkspaceState` is owned by this crate, I think all of these constructors should be in `impl WorkspaceState`.

crates/but-api/src/workspace_state.rs
```diff
@@ -0,0 +1,79 @@
...
+    repo: &gix::Repository,
+    cache: &mut but_db::CacheHandle,
+    replaced_commits: BTreeMap<gix::ObjectId, gix::ObjectId>,
+) -> anyhow::Result<WorkspaceState> {
+    let head_info = but_workspace::ref_info_from_graph(
+        graph,
+        repo,
+        but_workspace::ref_info::Options {
+            traversal: but_graph::init::Options::limited(),
+            expensive_commit_info: true,
```
Byron: Just as a note: `expensive_commit_info` is something I added in preparation for better controlling long-running tasks.
Now that we have a chance to actually get them, I see this going away in favor of explicit 'enrichment' calls that provide certain kinds of expensive information behind a long-running task that may send computation results incrementally even so it can run in the background. For now we only have a PoC for it which last time I checked also didn't support increments.

While this is future talk, I can also imagine steps in this direction to keep using this field, and each call is limited to 1s of compute time with everything being cached so the next call can continue where the last one left off.

crates/but-api/src/workspace_state.rs
```diff
@@ -0,0 +1,79 @@
...
+        },
+        cache,
+    )?
+    .pruned_to_entrypoint();
```
Byron: This is something I never liked as I think the consumer should be built to deal with the notion of a `HEAD` or entrypoint. This feature was made specifically for the Svelte GUI, as quick workaround and I'd love it to not spread unless this really, really should be the way to go 😅.
CC @OliverJAsh .

crates/but-api/src/workspace_state.rs
```diff
@@ -0,0 +1,79 @@
...
+    repo: &gix::Repository,
+    cache: &mut but_db::CacheHandle,
+    dry_run: bool,
+) -> anyhow::Result<WorkspaceState> {
+    if dry_run {
+        return from_rebase_preview(&rebase, cache, rebase.history.commit_mappings());
+    }
+
+    let graph = rebase.overlayed_graph()?;
+    let materialized = rebase.materialize()?;
```
Byron: This seems to be a shortcut, but I don't understand why it's there.
Assuming the `overlayed_graph` is a temporary solution to not having in-memory references in `repo` (or `rebase.repo`), I'd expect the `graph` to be re-traversed from the materialized state.

So I'd expect something like this:

```suggestion
    let materialized = rebase.materialize()?;
    let graph = materialize.updated_graph()?;
```

crates/but-api/src/workspace_state.rs
```diff
@@ -0,0 +1,79 @@
...
+/// intended operation, regardless of whether that graph came from a preview, a
+/// materialized rebase, or another graph-producing workflow. The caller is
+/// responsible for supplying the matching `replaced_commits` map for that graph.
+///
+/// This is the most direct constructor in this module and is the right choice when
+/// there is no need to inspect or materialize a [`SuccessfulRebase`].
+pub(crate) fn from_overlayed_graph(
+    graph: but_graph::Graph,
+    repo: &gix::Repository,
+    cache: &mut but_db::CacheHandle,
```
Byron: Just as a note: the `cache` is gone now thanks to derived ChangeIDs (instead of random ones).

crates/but-api/src/types.rs
```diff
@@ -1,14 +1,31 @@
...
+    pub fn new(
+        replaced_commits: BTreeMap<gix::ObjectId, gix::ObjectId>,
+        head_info: RefInfo,
+    ) -> Self {
+        Self {
+            replaced_commits,
+            head_info,
+        }
+    }
+}
```
Byron: This would be better in `workspace_state.rs` since the type now has a dedicated module.

crates/but-workspace/src/ref_info.rs
```diff
@@ -400,6 +400,26 @@ pub(crate) mod function {
...
+        graph: Graph,
+        repo: &gix::Repository,
+        opts: super::Options,
+        cache: &mut but_db::CacheHandle,
+    ) -> anyhow::Result<RefInfo> {
+        if graph.hard_limit_hit() {
+            tracing::warn!(hard_limit=?opts.traversal.hard_limit,
+                "Commit-graph traversal might be incorrect as it was stopped too early due to hard limit",
+            );
+        }
```
Byron: This should now be in `graph_to_ref_info` to avoid this duplication.

crates/but-api/src/branch.rs
```diff
@@ -112,32 +103,45 @@ pub fn apply_only(
...
+    existing_branch: &gix::refs::FullNameRef,
+    apply_options: but_workspace::branch::apply::Options,
+    perm: &mut RepoExclusive,
+) -> anyhow::Result<but_workspace::branch::apply::Outcome<'static>> {
+    let repo = ctx.clone_repo_for_merging_non_persisting()?;
+    let mut meta = ManuallyDrop::new(ctx.meta()?);
+    let (_repo, ws, _) = ctx.workspace_mut_and_db_with_perm(perm)?;
+
+    but_workspace::branch::apply(existing_branch, &ws, &repo, &mut *meta, apply_options)
+        .map(but_workspace::branch::apply::Outcome::into_owned)
```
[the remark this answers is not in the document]
Byron: I agree - while `meta()` isn't officially backed by the database, I think it would be good to have `ctx.meta_non_persisting()`. 
Looking into the future, once `meta` is in `db`, `apply` will have to manage its own transaction, which means that `apply_options` would need the `dry_run` setting as well. Adding the field would be my preference, but might also be more work so `ctx.meta_non_persisting()` seems like a reasonable trade-off as it will probably be me dealing with it later when vb.toml is removed for good.

crates/but-api/src/json.rs
```diff
@@ -131,6 +131,40 @@ mod hex_hash {
...
+    pub replaced_commits: std::collections::BTreeMap<HexHash, HexHash>,
+    /// The post-operation workspace view presented to the frontend.
+    pub head_info: but_workspace::ui::RefInfo,
+}
+
+#[cfg(feature = "export-schema")]
+but_schemars::register_sdk_type!(WorkspaceState);
+
+impl From<crate::types::WorkspaceState> for WorkspaceState {
+    fn from(value: crate::types::WorkspaceState) -> Self {
```
Byron: It's better to destructure `value` here to prevent fields from being missed next time one is added in the non-JSON type.

crates/but-api/src/json.rs
```diff
@@ -131,6 +131,40 @@ mod hex_hash {
...
+#[serde(rename_all = "camelCase")]
+pub struct WorkspaceState {
+    /// Commits that were replaced by the operation. Maps `oldId -> newId`.
+    #[cfg_attr(
+        feature = "export-schema",
+        schemars(with = "std::collections::BTreeMap<String, String>")
+    )]
+    pub replaced_commits: std::collections::BTreeMap<HexHash, HexHash>,
+    /// The post-operation workspace view presented to the frontend.
+    pub head_info: but_workspace::ui::RefInfo,
```
Byron, line 146: It's great to finally have this built-in 🎉.

crates/but-api/src/json.rs
```diff
@@ -131,6 +131,40 @@ mod hex_hash {
...
+        Self {
+            replaced_commits: value
+                .replaced_commits
+                .into_iter()
+                .map(|(old, new)| (old.into(), new.into()))
+                .collect(),
+            head_info: value
+                .head_info
+                .try_into()
+                .expect("workspace state should convert to JSON"),
```
Byron: We can't make that assertion here, so the conversion must be fallible. `but_api` supports this with `try_from = `.

Byron (changes requested):
Thanks for inviting a review! It's exciting to see this happen, also vividly remembering the PoC that was recently demonstrated.

However, I think the PR needs work, with the biggest issue being the oplog entries that are actually done. Currently, this means quite a lot of work and actual objects being created, even though no work needs to be done at all for something that isn't affecting the mutation that follows it.
My recommendation is to add a helper to `but-oplog` that takes `dry_run` and is doing what all mutations need anyway, maybe to the point where `.ok()` can also be omitted or is built-in.

Also, given it's importance, I think instead of forcing ourselves to see `false` in many call sites, we should leverage `boolean-enums` (already a dependency) to turn this into `DryRun::No` (or even create a dedicated enum for it).

And there is more in the comments, so I am stopping the review here without having looked much at the actual `dry_run` implementation/port, which I'd think overall should be fine.

PS: the commit copies the PR message (or the other way around), making it talk about a PR which may be confusing.

crates/but-api/src/workspace_state.rs
Caleb-T-Owens: I'm not sure these make sense to backtick here.

crates/but-api/src/workspace_state.rs
Caleb-T-Owens: This has been done currently to match the `head_info` command. Removing it's usage should be another PR

crates/but-api/src/commit/uncommit.rs
```diff
@@ -45,13 +49,14 @@ pub fn commit_uncommit_changes_only_with_perm(
     commit_id: gix::ObjectId,
     changes: Vec<but_core::DiffSpec>,
     assign_to: Option<but_core::ref_metadata::StackId>,
+    dry_run: DryRun,
     perm: &mut but_ctx::access::RepoExclusive,
 ) -> anyhow::Result<MoveChangesResult> {
     let context_lines = ctx.settings.context_lines;
     let mut meta = ctx.meta()?;
-    let (repo, mut ws, mut db, _cache) = ctx.workspace_mut_and_db_mut_and_cache_with_perm(perm)?;
+    let (repo, mut ws, mut db) = ctx.workspace_mut_and_db_mut_with_perm(perm)?;
```
Byron, line 61: Here I think we have to take a closer look: the way I see it, hunk assignments are only dependent on the database `db`. This means it has dry-run built-in by starting a transaction that then is either discarded or committed.

Byron (approved):
Thanks a lot - looks good now, except for the one thing about how dry-run is implemented for `db`-only calls.

While `dry_run: DryRun` isn't documented now, I had that done. Also I changed `UmaterialisedSnapshot` back to what it was, and changed it to return an `Option` as it is what most (if not all) callers want.

Please note that I assume that the hardcoded `DryRun::No` will eventually also be able to support dry-run, but that this is done when needed.

(intermediate review, I stopped at the `db` portion in `uncommit.rs` that should use a `db.transaction()`)

crates/but-api/src/commit/uncommit.rs
```diff
@@ -68,62 +78,101 @@ pub fn commit_uncommit_changes_only_with_perm(
...
-
-        let to_assign: Vec<_> = after_assignments
-            .into_iter()
-            .filter(|assignment| assignment.id.is_some_and(|id| !before_ids.contains(&id)))
-            .map(|assignment| HunkAssignmentRequest {
-                hunk_header: assignment.hunk_header,
-                path_bytes: assignment.path_bytes,
-                target: Some(HunkAssignmentTarget::Stack { stack_id }),
-            })
-            .collect();
```
Byron, line 95: Here would be the question if there can be a dry_run-friendly materialization, in the service of letting code just run through the normal path, without having to deal with whether or not it's actually materialized.

To my mind, this *could* work as long as the code only uses data from the materialization itself, which either is in memory or it is not.

In future, with references also being in memory via `gix::Repository`, the caller of the materialisation can't tell anymore.

crates/but-api/src/commit/undo.rs
```diff
@@ -78,46 +100,42 @@ pub fn commit_undo_only_with_perm(
...
+            .with_context(|| format!("failed to discard {}", subject_commit_id.to_hex()))?
     };
 
-    let mut discard_result =
-        crate::commit::discard_commit::commit_discard_only_with_perm(ctx, new_commit, perm)
-            .with_context(|| format!("failed to discard {}", subject_commit_id.to_hex()))?;
-
-    replaced_commits.append(&mut discard_result.replaced_commits);
+    let workspace = if dry_run.into() {
+        WorkspaceState::from_rebase_preview(&final_rebase, final_rebase.history.commit_mappings())?
```
Byron, line 127: It seems that `final_rebase` is really the only data needed here.
I wonder if the `WorkspaceState` could rather do something like

`WorkspaceState::from_rebase(dry_run)` and it matrialises on the fly?

Ideally we find a way of doing try-run that leaves the code that didn't support dry-run before as unchanged as possible. 
That would lead to future mutations to naturally be implemented with `dry_run` support as it's no extra effort.

crates/but-core/src/lib.rs
```diff
@@ -450,3 +451,7 @@ pub struct WorktreeChanges {
     /// The conflicting index entries, along with their relative path `(rela_path, [Entries(base, ours, theirs)])`.
     pub index_conflicts: Vec<(BString, Box<[Option<ConflictIndexEntry>; 3]>)>,
 }
+
+// Represents whether an operation should be materialized on disk or remain in
+// memory.
+gen_boolean_enum!(pub serde DryRun);
```
Byron, line 457: Let's see if I can try to contribute something that will actually give the type docs.

crates/but-workspace/src/ref_info.rs
```diff
@@ -448,11 +443,21 @@ pub(crate) mod function {
...
     }
 
-    fn graph_to_ref_info(
-        graph: Graph,
+    /// Gather information about graph and the workspace that might be associated with it,
+    /// based on data in `repo` and `meta`. Use `options` to further configure the call.
+    ///
+    /// For details, see [`ref_info()`].
+    pub fn graph_to_ref_info(
+        workspace: &but_graph::projection::Workspace,
```
Byron, line 451: This is great!

Byron (commented):
Here is one more patch that I'd love to see in there, it's structural stuff around `WorkspaceState`.

The important bit here is that we don't rename types, but use them from their location. Then a type that is returned should also be available so should be explicitly exported.

```patch
commit 230d0f5f76d23f538d5f65ffd0ffcaecf42e2195
Date:   Tue Apr 14 17:34:53 2026 +0800

    review2 as patch

diff --git a/crates/but-api/src/branch.rs b/crates/but-api/src/branch.rs
index f5fff2a885..839e96bec7 100644
--- a/crates/but-api/src/branch.rs
+++ b/crates/but-api/src/branch.rs
@@ -1,4 +1,4 @@
-use crate::workspace_state::WorkspaceState;
+use crate::WorkspaceState;
 use but_api_macros::but_api;
 use but_core::{
     DryRun, sync::RepoExclusive, ui::TreeChanges, worktree::checkout::UncommitedWorktreeChanges,
diff --git a/crates/but-api/src/commit/amend.rs b/crates/but-api/src/commit/amend.rs
index 1467182b58..37564d52ef 100644
--- a/crates/but-api/src/commit/amend.rs
+++ b/crates/but-api/src/commit/amend.rs
@@ -1,4 +1,4 @@
-use crate::workspace_state::WorkspaceState;
+use crate::WorkspaceState;
 use but_api_macros::but_api;
 use but_core::{DiffSpec, DryRun, sync::RepoExclusive};
 use but_oplog::legacy::{OperationKind, SnapshotDetails};
diff --git a/crates/but-api/src/commit/create.rs b/crates/but-api/src/commit/create.rs
index 117879823f..cad24d35c9 100644
--- a/crates/but-api/src/commit/create.rs
+++ b/crates/but-api/src/commit/create.rs
@@ -1,4 +1,4 @@
-use crate::workspace_state::WorkspaceState;
+use crate::WorkspaceState;
 use but_api_macros::but_api;
 use but_core::{DiffSpec, DryRun, sync::RepoExclusive};
 use but_oplog::legacy::{OperationKind, SnapshotDetails};
diff --git a/crates/but-api/src/commit/discard_commit.rs b/crates/but-api/src/commit/discard_commit.rs
index ec8d18c7a4..307644b5c2 100644
--- a/crates/but-api/src/commit/discard_commit.rs
+++ b/crates/but-api/src/commit/discard_commit.rs
@@ -1,4 +1,4 @@
-use crate::workspace_state::WorkspaceState;
+use crate::WorkspaceState;
 use but_api_macros::but_api;
 use but_core::{DryRun, sync::RepoExclusive};
 use but_oplog::legacy::{OperationKind, SnapshotDetails, Trailer};
diff --git a/crates/but-api/src/commit/insert_blank.rs b/crates/but-api/src/commit/insert_blank.rs
index 4b09219e05..7952208254 100644
--- a/crates/but-api/src/commit/insert_blank.rs
+++ b/crates/but-api/src/commit/insert_blank.rs
@@ -1,4 +1,4 @@
-use crate::workspace_state::WorkspaceState;
+use crate::WorkspaceState;
 use but_api_macros::but_api;
 use but_core::{DryRun, sync::RepoExclusive};
 use but_oplog::legacy::{OperationKind, SnapshotDetails};
diff --git a/crates/but-api/src/commit/move_changes.rs b/crates/but-api/src/commit/move_changes.rs
index b394c21fdf..2b6dcbe87b 100644
--- a/crates/but-api/src/commit/move_changes.rs
+++ b/crates/but-api/src/commit/move_changes.rs
@@ -1,4 +1,4 @@
-use crate::workspace_state::WorkspaceState;
+use crate::WorkspaceState;
 use but_api_macros::but_api;
 use but_core::{DryRun, sync::RepoExclusive};
 use but_oplog::legacy::{OperationKind, SnapshotDetails};
diff --git a/crates/but-api/src/commit/move_commit.rs b/crates/but-api/src/commit/move_commit.rs
index c645d6c9d5..48812829e7 100644
--- a/crates/but-api/src/commit/move_commit.rs
+++ b/crates/but-api/src/commit/move_commit.rs
@@ -7,7 +7,7 @@ use but_rebase::graph_rebase::{
 };
 use tracing::instrument;
 
-use crate::workspace_state::WorkspaceState;
+use crate::WorkspaceState;
 
 use super::types::CommitMoveResult;
 
diff --git a/crates/but-api/src/commit/reword.rs b/crates/but-api/src/commit/reword.rs
index 3cace98ed4..8f07b4d71c 100644
--- a/crates/but-api/src/commit/reword.rs
+++ b/crates/but-api/src/commit/reword.rs
@@ -1,4 +1,4 @@
-use crate::workspace_state::WorkspaceState;
+use crate::WorkspaceState;
 use bstr::{BString, ByteSlice};
 use but_api_macros::but_api;
 use but_core::{DryRun, sync::RepoExclusive};
diff --git a/crates/but-api/src/commit/squash.rs b/crates/but-api/src/commit/squash.rs
index 220a491b88..da220cfe4b 100644
--- a/crates/but-api/src/commit/squash.rs
+++ b/crates/but-api/src/commit/squash.rs
@@ -1,4 +1,4 @@
-use crate::workspace_state::WorkspaceState;
+use crate::WorkspaceState;
 use but_api_macros::but_api;
 use but_core::{DryRun, sync::RepoExclusive};
 use but_oplog::legacy::{OperationKind, SnapshotDetails};
diff --git a/crates/but-api/src/commit/types.rs b/crates/but-api/src/commit/types.rs
index d178acfe65..0f11a85fc1 100644
--- a/crates/but-api/src/commit/types.rs
+++ b/crates/but-api/src/commit/types.rs
@@ -1,4 +1,4 @@
-use crate::workspace_state::WorkspaceState;
+use crate::WorkspaceState;
 use but_core::{DiffSpec, tree::create_tree::RejectionReason};
 
 /// Outcome after creating a commit.
diff --git a/crates/but-api/src/commit/uncommit.rs b/crates/but-api/src/commit/uncommit.rs
index c5ce820b6c..d9ddd1958a 100644
--- a/crates/but-api/src/commit/uncommit.rs
+++ b/crates/but-api/src/commit/uncommit.rs
@@ -1,6 +1,6 @@
 use std::collections::HashSet;
 
-use crate::workspace_state::WorkspaceState;
+use crate::WorkspaceState;
 use but_api_macros::but_api;
 use but_core::DryRun;
 use but_hunk_assignment::{HunkAssignmentRequest, HunkAssignmentTarget};
diff --git a/crates/but-api/src/commit/undo.rs b/crates/but-api/src/commit/undo.rs
index 4f03a9f0c3..ff445b27d6 100644
--- a/crates/but-api/src/commit/undo.rs
+++ b/crates/but-api/src/commit/undo.rs
@@ -1,4 +1,4 @@
-use crate::workspace_state::WorkspaceState;
+use crate::WorkspaceState;
 use anyhow::Context as _;
 use but_api_macros::but_api;
 use but_core::{DiffSpec, DryRun, sync::RepoExclusive};
diff --git a/crates/but-api/src/json.rs b/crates/but-api/src/json.rs
index dd8454e7e8..a6444b5d7c 100644
--- a/crates/but-api/src/json.rs
+++ b/crates/but-api/src/json.rs
@@ -6,7 +6,6 @@
 //! If a JSON type only mirrors one API submodule, define it next to that API in
 //! a local `json` module instead of adding it here. See `crate::branch::json`,
 //! `crate::commit::json`, and `crate::diff::json` for the intended pattern.
-use crate::workspace_state::WorkspaceState as InternalWorkspaceState;
 pub use error::{Error, ToJsonError, UnmarkedError};
 use gix::refs::Target;
 use schemars::{self, JsonSchema};
@@ -150,14 +149,14 @@ pub struct WorkspaceState {
 #[cfg(feature = "export-schema")]
 but_schemars::register_sdk_type!(WorkspaceState);
 
-impl TryFrom<InternalWorkspaceState> for WorkspaceState {
+impl TryFrom<crate::WorkspaceState> for WorkspaceState {
     type Error = anyhow::Error;
 
     fn try_from(
-        InternalWorkspaceState {
+        crate::WorkspaceState {
             replaced_commits,
             head_info,
-        }: InternalWorkspaceState,
+        }: crate::WorkspaceState,
     ) -> Result<Self, Self::Error> {
         Ok(Self {
             replaced_commits: replaced_commits
diff --git a/crates/but-api/src/lib.rs b/crates/but-api/src/lib.rs
index 69b3772e29..be9274fb78 100644
--- a/crates/but-api/src/lib.rs
+++ b/crates/but-api/src/lib.rs
@@ -9,6 +9,8 @@
 #![cfg_attr(feature = "napi", deny(unsafe_code))]
 #![deny(missing_docs)]
 
+use std::collections::BTreeMap;
+
 #[cfg(feature = "legacy")]
 pub mod legacy;
 
@@ -40,4 +42,16 @@ pub mod panic_capture;
 #[cfg(feature = "export-schema")]
 pub mod watcher;
 
+/// Represents the workspace after any mutation was performed.
+///
+/// This describes the post-operation workspace view that mutations should
+/// return regardless of whether they executed for real or as a dry-run.
+#[derive(Debug, Clone)]
+pub struct WorkspaceState {
+    /// Commits that were replaced by the operation. Maps `old_id -> new_id`.
+    /// The workspace presented for the frontend. See [`RefInfo`] for more
+    /// detail.
+    pub head_info: but_workspace::RefInfo,
+}
 mod workspace_state;
diff --git a/crates/but-api/src/workspace_state.rs b/crates/but-api/src/workspace_state.rs
index ba16f9e2a6..7317476e6f 100644
--- a/crates/but-api/src/workspace_state.rs
+++ b/crates/but-api/src/workspace_state.rs
@@ -1,21 +1,10 @@
+use super::WorkspaceState;
 use std::collections::BTreeMap;
 
 use but_core::{DryRun, RefMetadata};
 use but_rebase::graph_rebase::SuccessfulRebase;
 
 use but_workspace::RefInfo;
-/// Represents the workspace for the frontend
-///
-/// This describes the post-operation workspace view that mutations should
-/// return regardless of whether they executed for real or as a dry-run.
-#[derive(Debug, Clone)]
-pub struct WorkspaceState {
-    /// Commits that were replaced by the operation. Maps `old_id -> new_id`.
-    pub replaced_commits: BTreeMap<gix::ObjectId, gix::ObjectId>,
-    /// The workspace presented for the frontend. See [`RefInfo`] for more
-    /// detail.
-    pub head_info: RefInfo,
-}
 
 impl WorkspaceState {
     /// Create a new workspace state from operation outputs.
```

crates/but-api/src/commit/undo.rs
Caleb-T-Owens: The `from_successful_rebase` helper _is_ doing a materialize.

In this function, we're doing a materialize_without_checkout. We could have a `from_successful_rebase_without_checkout` but it's really not that common.
```rs
    let workspace = if dry_run.into() {
        WorkspaceState::from_rebase_preview(&final_rebase, final_rebase.history.commit_mappings())?
    } else {
        let materialized = final_rebase.materialize_without_checkout()?;
        WorkspaceState::from_workspace(
            materialized.workspace,
            &repo,
            materialized.history.commit_mappings(),
        )?
    };
```

PR #13269, opened 2026-04-10
fix(build): Fix build on FreeBSD.

### LINKED ISSUE #13268: [ feature request ] Fix build on FreeBSD
I have a PR to make `but` build on FreeBSD.

yonas:
## 🧢 Changes

- Cargo.toml
- crates/but-claude/src/claude_settings.rs
- crates/but-installer/src/download.rs
- crates/but-installer/src/lib.rs
- crates/but-installer/src/shell.rs

## 🎫 Affected issues

Fixes: #13268

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -229,10 +229,9 @@ serde_json = "1.0.148"
 thiserror = "2.0.17"
 tokio = { version = "1.50.0", default-features = false }
 keyring = { version = "3.6.3", features = [
-    "apple-native",
-    "windows-native",
-    "linux-native-sync-persistent",
     "crypto-rust",
+    "sync-secret-service",
+
 ] }
 anyhow = "1.0.100"
 parking_lot = "0.12.4"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2160,7 +2160,7 @@ dependencies = [
  "bitflags 2.11.0",
  "block",
  "cocoa-foundation",
- "core-foundation 0.10.1",
+ "core-foundation",
  "core-graphics",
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -229,10 +229,9 @@ serde_json = "1.0.148"
 thiserror = "2.0.17"
 tokio = { version = "1.50.0", default-features = false }
 keyring = { version = "3.6.3", features = [
-    "apple-native",
-    "windows-native",
-    "linux-native-sync-persistent",
     "crypto-rust",
+    "sync-secret-service",
+
```
Caleb-T-Owens, line 234: Could you please explain this change in feature flags. Why are we removing the other os' native keyring options?

yonas: Yeah, this change made the compilation work but isn't portable. Let me think about it.

Byron (changes requested):
I set it back to draft and make the change-request from the comments official, as we definitely can't merge it in its current form and have to think about merging it at all without CI support.
It's too easy to regress in future otherwise, so I'd make CI support a requirement for a merge as well.

slarse:
I'm going to be a bit of a bummer here, but unless we have an internal champion for FreeBSD I don't think it's a good idea to add support for it. Just because it compiles doesn't mean it behaves well, as evidenced by all the quirks I've been working out across the Linux and Windows distributions over the last couple of months. And there are still more quirks to work out.

I'd urge some due consideration and team alignment before proceeding with this.

Byron:
Thanks, I think you right!

Let's close this and require serious community pressure to take this on.

PR #13276, opened 2026-04-12
Fix/floating commit box title lost

devanandn54:
Fixes #11968

The FloatingCommitBox was always displaying the static 
`floatingBoxHeader` ("Create commit") instead of the actual 
typed commit title. Changed to `title || floatingBoxHeader` 
so the typed title shows in the floating window header.
Screenshot 2026-04-11 at 9 25 38 PM
Screenshot 2026-04-11 at 9 25 45 PM

### DIFF
--- a/apps/desktop/src/components/commit/CommitMessageEditor.svelte
+++ b/apps/desktop/src/components/commit/CommitMessageEditor.svelte
@@ -245,7 +245,7 @@
 
 {#if useFloatingBox.current}
 	<FloatingCommitBox
-		title={floatingBoxHeader}
+		title={title || floatingBoxHeader}
 		onExitFloatingModeClick={() => {
 			useFloatingBox.set(false);
 		}}


Byron:

Apparently, with the latest [Nightly](https://gitbutler.com/nightly) build the issue as described in the PR doesn't reproduce anymore.

https://github.com/user-attachments/assets/c2ae5c9f-f1d7-41a0-8931-4b82a3fc0cd1

Could you double-check or show in a video how the issue this is fixing can be reproduced?

Thanks again.

devanandn54:
> Thanks a lot for contributing.
> 
> Apparently, with the latest [Nightly](https://gitbutler.com/nightly) build the issue as described in the PR doesn't reproduce anymore.
> 
>  Screen.Recording.2026-04-12.at.21.18.36.mov 
> Could you double-check or show in a video how the issue this is fixing can be reproduced?
> 
> Thanks again.

Hi Byron, thanks for sharing the video — I now understand the actual issue better. The bug shown is that the title field gets overwritten with AI-generated content when switching between floating and docked modes. My fix only addressed the window header label. I'll investigate the correct fix for the title content corruption and update this PR.

PR #13289, opened 2026-04-13
feat(but-installer): add skill install prompt

slarse:
## 🧢 Changes
Adds a prompt to install the but skill files at the end of the interactive installer.

If you accept the prompt, you're dropped right into `but skill install --global`.

accept_skill

If you reject the prompt, you get instructions for how to "do it again".

reject_skill

We never fail the install if something goes wrong installing the skill, at most we print some warnings.

## Thoughts
Having gone through adding this as a simple shim to the installed `but` binary, just invoking `<installed_bin> skill install --global`, it struck me that maybe we should take this approach for the shell config as well. Or, really, anything that requires interactivity. This would allow us to more easily provide a seamless UX across the installer and `but`, while still keeping the installer to the bare minimum.

I'll ponder it for a while. There is the obvious downside that any bugs we have in the interactive parts of the installer would not be possible to patch without making a new release (and would not be patched in older releases), so perhaps it's just not that great of an idea.

---
- <kbd>&nbsp;2&nbsp;</kbd> #13289 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #13285

## TODO
`but skill install` currently cannot run interactively when `curl | sh`-installing as it requires stdin to be a terminal, and in this common case stdin is a pipe. We need to also allow `/dev/tty` as a valid input, but still require that stdout is a direct terminal output.

### DIFF
--- a/crates/but-installer/src/lib.rs
+++ b/crates/but-installer/src/lib.rs
@@ -11,6 +11,7 @@ mod http;
 mod install;
 mod release;
 mod shell;
+mod skill;
 pub mod ui;
 
 #[cfg(target_os = "linux")]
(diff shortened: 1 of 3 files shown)


krlvi (approved):
great idea!

PR #13294, opened 2026-04-14
Fix credential test polluting dev mode user data

mtsgrd:
## Summary

- Remove the `gitbutler_user::set_user()` call from credential tests that was writing test fixture data (Byron's profile) to the real dev app data directory (`~/Library/Application Support/com.gitbutler.app.dev/user.json`)
- Every `cargo test` run overwrote the logged-in developer's user data, causing the desktop app to show the wrong profile picture
- The `help()` function under test only uses `project.preferred_key` — it never reads user data — so the call was unnecessary

## Test plan

- [x] `cargo test -p gitbutler-repo credentials` — all 4 tests pass
- [x] Verified `help()` never calls `get_user()` — the `set_user()` was dead code with a harmful side effect

### DIFF
--- a/crates/gitbutler-repo/tests/repo/credentials.rs
+++ b/crates/gitbutler-repo/tests/repo/credentials.rs
@@ -1,29 +1,20 @@
-use std::{path::PathBuf, str};
+use std::path::PathBuf;
 
 use crate::support::{setup_blackhole_store, test_repository};
 use but_ctx::{Context, RepoOpenMode};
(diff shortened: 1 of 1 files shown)


Byron:
No wonder that *I* never noticed!!
Strange that nobody else did until now as well!

This makes me wonder if this was isolated at some point, probably with environment variables, and that got refactored as away? Probably it's that, as `gitbutler_user::set_user()` is affected by `E2E` variables.

PR #13302, opened 2026-04-14
Route auth API calls through the backend to avoid CORS issues

mtsgrd:
## Summary

When running in `but remote` mode, browser security blocks cross-origin requests from the frontend to `app.gitbutler.com`. Rather than adding CORS headers to the staging/production servers, we route all authenticated API calls through the Rust backend. This keeps the API client in `gitbutler-user::api` reusable for a future `but auth` CLI command without any web framework dependency.

### What's routed through the backend

| Endpoint | Command | Used by |
|---|---|---|
| `POST /api/login/token.json` | `get_login_token` | Login flow |
| `GET /api/login/whoami` | `login_with_token` | Token validation |
| `GET /api/user.json` | `get_user_profile` | Settings page (refresh profile) |
| `PUT /api/user.json` | `update_user_profile` | Settings page (update name/avatar) |

### Key changes

- **`gitbutler-user::api`** — new auth API client with sync public API (async HTTP runs on a dedicated thread with a short-lived Tokio runtime, following the `but-forge` pattern)
- **`default_api_url()`** — compile-time `AppChannel` determines the API URL, with a `GITBUTLER_API_URL` runtime env var override for local development
- **`but_api` commands** — `get_login_token`, `login_with_token`, `get_user_profile`, `update_user_profile` exposed via the `#[but_api]` macro so both Tauri and but-server use the same code path
- **`but-server` routes** — all four commands routed through `but_post` for consistent error envelopes
- **Frontend** — `userService.ts` now calls `backend.invoke(...)` instead of `httpClient.get/put(...)` for all user API calls; avatar uploads are base64-encoded in the browser and re-encoded as multipart form data in Rust
- **`auth.rs`** simplified — delegates token validation to `gitbutler_user::api::validate_token_owner`, which returns `Ok(false)` for 401/403 (not the owner) and `Err` for everything else
- **CSP fix** — removed invalid `ws://[::1]:{port}` from `connect-src` (brackets are not valid in CSP source expressions; `ws://localhost` covers both IPv4 and IPv6)
- **HTTP timeouts** — 10s connect / 30s request timeout on all API calls to prevent hangs in the auth middleware path
- **Typed errors** — `ApiHttpError` used consistently across all fetch functions for status-code matching

### Security notes

- Access tokens now travel backend-to-API only, never as cross-origin browser requests
- `api_url` is compile-time by default — the runtime override (`GITBUTLER_API_URL`) is opt-in and not derived from untrusted input, so there is no SSRF risk
- In remote mode, the new commands sit behind the existing auth middleware; the initial login flow uses the separate public `/auth/login` + `/auth/callback` routes
- In local mode, the auth middleware is a no-op (localhost only, matching the existing trust model)


crates/gitbutler-user/src/api.rs
```diff
@@ -0,0 +1,97 @@
...
+/// environment variable.
+///
+/// - `"release"` / `"nightly"` → `https://app.gitbutler.com`
+/// - anything else (including unset) → `https://app.staging.gitbutler.com`
+pub fn default_api_url() -> &'static str {
+    match option_env!("CHANNEL") {
+        Some("release" | "nightly") => "https://app.gitbutler.com",
+        _ => "https://app.staging.gitbutler.com",
+    }
+}
```
estib-vega: Commenting for documentation and future reference:

The API endpoint will be controlled by the CHANNEL envvar now. 
Before, for development purposes, we'd update the .env.* files but now we can only control it through the build target.

**Request** @mtsgrd could we add a way to pass an ENV variable that overrides the API endpoint? That way we can still pass e.g. `http://localhost:3000` for development without having to manually change the code

estib-vega (approved):
It's a great idea, thanks! Just added a small doc comment + request

crates/gitbutler-user/src/api.rs
mtsgrd: Done!

estib-vega: Thanks!

crates/gitbutler-user/src/api.rs
```diff
@@ -0,0 +1,103 @@
...
+        Some("release" | "nightly") => "https://app.gitbutler.com",
+        _ => "https://app.staging.gitbutler.com",
+    }
+    .to_string()
+}
+
+/// Response from `POST /api/login/token.json`.
+#[derive(Debug, Deserialize, Serialize)]
+pub struct LoginToken {
+    /// Polling token (server-side only; never sent to the browser).
```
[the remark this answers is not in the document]
mtsgrd: Fixed — updated the doc comment to reflect actual exposure.

crates/but-server/src/auth.rs
```diff
@@ -23,107 +23,67 @@ const TOKEN_CACHE_TTL: Duration = Duration::from_secs(300);
...
-            .await?;
-
-        if !resp.status().is_success() {
-            return Ok(false);
-        }
-
-        let user: ApiUser = resp.json().await?;
+        let api_url = gitbutler_user::api::default_api_url();
+        let is_owner =
+            gitbutler_user::api::validate_token_owner(&api_url, token, self.owner_id).await?;
```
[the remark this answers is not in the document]
mtsgrd: Fixed — `validate_token_owner` now returns `Ok(false)` for client errors (401/403) via a typed `ApiHttpError`, reserving `Err` for network/server failures.

apps/desktop/src/lib/user/userService.ts
```diff
@@ -141,8 +133,7 @@ export class UserService {
 	private async getLoginUrl(): Promise<string> {
 		this.forgetUserCredentials();
 		try {
```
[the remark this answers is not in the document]
mtsgrd: Fixed — added the missing `await`.

crates/gitbutler-user/src/api.rs
Byron: Can we use the `AppChannel` type here? I think so.
Ideally, it's parameterized up to the highest level so we avoid globals (and compile-time conditions) as much as possible.

crates/but-api/src/legacy/users.rs
```diff
@@ -98,3 +99,17 @@ pub fn set_user(user: User) -> Result<()> {
...
+#[instrument(err(Debug))]
+pub async fn get_login_token() -> Result<LoginToken> {
+    let api_url = gitbutler_user::api::default_api_url();
+    gitbutler_user::api::fetch_login_token(&api_url).await
+}
+
+#[but_api]
+#[instrument(err(Debug))]
+pub async fn login_with_token(token: String) -> Result<serde_json::Value> {
+    let api_url = gitbutler_user::api::default_api_url();
```
krlvi: Maybe we can avoid leaking this out of the `gitbutler-user` crate?

crates/gitbutler-user/src/api.rs
```diff
@@ -0,0 +1,126 @@
...
+        .context("Failed to parse login token response")
+}
+
+/// Validate an access token against the GitButler API and return the user info.
+///
+/// Calls `GET /api/login/whoami` with the given token. On success the full
+/// user object is returned as a [`serde_json::Value`] so callers can
+/// deserialize into whatever type they need (the frontend `User` has
+/// different fields than the Rust `User`).
+pub async fn fetch_user_by_token(api_url: &str, token: &str) -> Result<serde_json::Value> {
```
krlvi: Maybe make only this function public - the rest can be internal to the crate

krlvi: we have previously discussed async - we dont want our crates to have their APIs async. Main reason is that it has "viral" effect downstream forcing every consumer (eg cli tui) to be async.
Take a look at `but-forge` for how we are handling network there

krlvi (changes requested):

crates/but-api/src/legacy/users.rs
mtsgrd: Fixed — `fetch_user_by_token` now returns `serde_json::Value` only within the crate. The `but_api` command wraps it. The public API no longer leaks `serde_json::Value` in its function signatures beyond what's needed for the JSON response.

crates/gitbutler-user/src/api.rs
mtsgrd: Addressed — `fetch_login_token` and `fetch_user_by_token` are now public (they're the main entry points), but they no longer take `api_url` — they call `default_api_url()` internally. `validate_token_owner` is also public as a convenience for the auth middleware.

mtsgrd: Fixed — the public API is now fully synchronous. Async HTTP is executed on a dedicated thread with a short-lived Tokio runtime, following the `but-forge` pattern (`std::thread::spawn` + `Runtime::new().block_on()`). The `but_api` commands are no longer async.

crates/gitbutler-user/src/api.rs
mtsgrd: Done — now using `AppChannel::new()` from `but-path` instead of raw `option_env\!("CHANNEL")`.

mtsgrd:
Thanks guys! I just pushed cad9237bb4ea98fb41e4cd766071a8ea3525a4fd addressing the feedback. *edit: I'll squash and push before any merge.

crates/gitbutler-user/src/api.rs
```diff
@@ -0,0 +1,165 @@
...
+        let client = http_client();
+        let resp = client
+            .post(&url)
+            .send()
+            .await
+            .context("Failed to reach GitButler API for login token")?;
+        if !resp.status().is_success() {
+            let status = resp.status();
+            let body = resp.text().await.unwrap_or_default();
+            anyhow::bail!("Login token request failed ({status}): {body}");
```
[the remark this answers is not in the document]
mtsgrd: Fixed — `fetch_login_token` now returns `ApiHttpError` on non-2xx, matching `fetch_user_by_token`.

krlvi (approved):

PR #13321, opened 2026-04-15
But CLI: Use modern apply for top-level but apply

krlvi:
## Summary
- route top-level `but apply` through `command::branch::apply`
- make the modern apply command available in legacy builds while preserving shell output
- remove the unused legacy apply implementation and module export

## Verification
- cargo test -p but apply
- cargo fmt --check --all
- cargo clippy -p but --all-targets


estib-vega:
image

slarse:
Nice!

PR #13322, opened 2026-04-15
Remove function modules

Caleb-T-Owens:
These modules were a pattern that was starting to spread and seems to add very liittle value to the project.

There is no functional change. Review recommended to use "Hide whitespace" to ignore indentation changes in the diff.

Screenshot 2026-04-15 at 13 51 39

### DIFF
--- a/crates/but-rebase/src/cherry_pick.rs
+++ b/crates/but-rebase/src/cherry_pick.rs
@@ -22,321 +22,312 @@ pub enum EmptyCommit {
     UsePrevious,
 }
 
-pub(crate) mod function {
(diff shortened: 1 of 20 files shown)


krlvi (approved):

Byron (commented):
I think that's a great looking change, as it reduces the verticality of the coat, as there is one indentation level less.

I also agree that these function modules should only be used when they actually have a use, also in preparation from actually putting this into its own module file, which would also have removed the indentation level. However, here it was probably mostly done as a pattern and not out of necessity.

So I think it's good to remove it and to re-add it when needed.

PR #13324, opened 2026-04-15
feat(but): add CLI/TUI theming

slarse:
This adds a global theme for the CLI/TUI to use.

The theme can be overridden by placing a JSON file at `~/.config/gitbutler/but-theme.json` (on UNIX-likes). For example, this would make remote branches white + underlined and additions red + bold.

```json
{
  "remote_branch": { "fg": "White", "add_modifier": "UNDERLINED" },
  "addition": { "fg": "Red", "add_modifier": "BOLD" }
}
```

I don't think we should _advertise_ the above as a feature for quite some time, but it's _very_ useful to just play around with different colors without having to recompile the app.

Example usage for status (using ratatui) and diff (using plain writes) is included.

# Questions
* ~~Can we use ratatui's `Style` across the board or is there good reason to keep `colored` around for the plain one-shot CLI output?~~
* ~~Should we wipe the slate clean and remove _all_ other styling as part of this PR, and then methodically go through each command and add styling again?~~

### DIFF
--- a/crates/but/src/command/legacy/status/render_oneshot.rs
+++ b/crates/but/src/command/legacy/status/render_oneshot.rs
@@ -1,11 +1,9 @@
-use colored::{ColoredString, Colorize};
-use ratatui::style::{Color, Modifier, Style};
-
 use crate::{
     command::legacy::status::{
         FileLineContent, StatusOutputLine,
         output::{BranchLineContent, CommitLineContent, StatusOutputContent},
     },
+    theme::Paint,
     utils::WriteWithUtils,
 };
 
@@ -21,8 +19,6 @@ pub(super) fn render_oneshot(
         content,
         data: _,
     } = line;
-
-    let should_colorize = colored::control::SHOULD_COLORIZE.should_colorize();
 
     let mut spans = Vec::new();
     if let Some(mut connector) = connector {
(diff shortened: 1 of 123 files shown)


Byron:
This PR looks like it's going to be a nightmare to get merged, as it's already conflicting with master, despite its short lifetime.

With that said, it feels like ultimately being able to configure colors is the only way to get it right for everyone as we can never get it right for everyone. People must be able to customize it. That's just the way of the world.

> Can we use ratatui's Style across the board or is there good reason to keep colored around for the plain one-shot CLI output?

Even though ordinarily I wouldn't want to be adding a Ratatoui dependency to the CLI, I think there's no denying that it's required for the TUI and thus we already have this dependency and might as well use it for colors to simplify theming.

> Should we wipe the slate clean and remove all other styling as part of this PR, and then methodically go through each command and add styling again?

Doing so seems like it's in the way to get the foundation in this PR merged quickly. So I'd probably just follow this one up with a couple of PRs later to finish the styling changes in smaller increments.

With that said, from a feature perspective, it would of course be nice to have it all done with this PR.

slarse:
> This PR looks like it's going to be a nightmare to get merged, as it's already conflicting with master, despite its short lifetime.

I actually think it'll be fine. The current conflicts are just snapshots, and any non-snapshot conflicts are most likely going to be trivial to resolve.

> Doing so seems like it's in the way to get the foundation in this PR merged quickly. So I'd probably just follow this one up with a couple of PRs later to finish the styling changes in smaller increments.

That is fair point. @estib-vega and I were just pretty keen to bring out the big guns as, clearly, the styling is currently all over the place :)

davidpdrsn:
I really like this!

I agree that we shouldn't advertise the json file yet. I think also if we were to do something user facing then we should have a format specifically designed for that (and maybe use toml instead of json but we can cross that bridge when we get there)

> Can we use ratatui's Style across the board or is there good reason to keep colored around for the plain one-shot CLI output?

Like I said in discord I think keeping colored is fine for the cases where its more convenient to use. But if all coloring goes through the theme (which I think should) then maybe that abstraction could just use ratatui inside for consistency?

> Should we wipe the slate clean and remove all other styling as part of this PR, and then methodically go through each command and add styling again?

I think this PR should focus on getting the theming machinery in place and then we can go through and update the colors in follow up PRs. That'll probably also make it easier to actually get it merged since I update the TUI snapshots all the time 😆

crates/but/src/command/legacy/status/render_oneshot.rs
```diff
@@ -22,8 +20,6 @@ pub(super) fn render_oneshot(
         data: _,
     } = line;
 
-    let should_colorize = colored::control::SHOULD_COLORIZE.should_colorize();
```
slarse, line 25: This is baked into `Paint::paint()`

crates/but/src/command/legacy/status/render_oneshot.rs
```diff
@@ -68,121 +64,11 @@ pub(super) fn render_oneshot(
     }
 
     for span in spans {
-        let style = span.style;
-        let rendered = render_span_with_colored(&span.content, style, should_colorize);
+        let rendered = span.style.paint(&span.content);
         write!(out, "{rendered}")?;
-        if should_colorize && style_has_effect(style) {
-            write!(out, "\x1b[0m")?;
-        }
```
slarse, line 76: This is redundant, `ColoredString::fmt()` always appends a reset, see the integration tests added in 0b8d11cda044342bb911c7884021fc46c848bddd

crates/but/src/theme.rs
```diff
@@ -0,0 +1,339 @@
...
+    /// Apply this style to `text`, producing a [`String`] with ANSI escape codes baked in.
+    ///
+    /// Each paint application produces an independently styled [`String`] in the sense that it is
+    /// terminated with a reset code.
+    ///
+    /// Note that at this time, any implementation _must_ respect the control mechanisms via
+    /// [`colored::control::SHOULD_COLORIZE`] for enabling/disabling styled output. If we require
+    /// multiple implementations of this trait in the future, we'll probably want to roll our own
+    /// control mechanism.
+    fn paint<S: AsRef<str>>(&self, text: S) -> String;
```
slarse, line 91: Important: I changed this to return a `String` instead of a `ColoredString`, so we confine our styling to ratatui's `Style` struct. One way to style with the theme, and then two ways to render!

crates/but/src/theme.rs
```diff
@@ -0,0 +1,339 @@
...
+        Color::Gray | Color::White => styled.on_white(),
+        Color::DarkGray => styled.on_bright_black(),
+        Color::LightRed => styled.on_bright_red(),
+        Color::LightGreen => styled.on_bright_green(),
+        Color::LightYellow => styled.on_bright_yellow(),
+        Color::LightBlue => styled.on_bright_blue(),
+        Color::LightMagenta => styled.on_bright_magenta(),
+        Color::LightCyan => styled.on_bright_cyan(),
+        Color::Rgb(r, g, b) => styled.on_truecolor(r, g, b),
+        Color::Indexed(_) | Color::Reset => styled,
```
slarse, line 158: Note: This is questionable and could lead to unexpected behavior.

`Color::Indexed()` is just an 8-bit color, it can be mapped to `styled.on_truecolor()`. Won't do that in this PR but I think it should probably be done.

`Color::Reset` is _arguably_ not applicable, as every `ColoredString` is self-contained by terminating with a reset. However, you could argue that the presence of a `Color::Reset` means this string should _start_ with a reset. If we ever mix-n-match `ColoredString` with some other way of controlling the ANSI escape codes that are actually printed to the screen, it's possible we'd end up in a situation where the TUI gracefully resets as a "start action" of some string output, while `but` does it only at the end of the same string output. Probably a massive edge case but it could happen.

crates/but/tests/theme.rs
```diff
@@ -0,0 +1,42 @@
...
+    let result = but::theme::Paint::paint(&style, "hello");
+
+    let bold_green = "\x1b[1;32m";
+    let reset = "\x1b[0m";
+    let expected = format!("{bold_green}hello{reset}");
+
+    assert_eq!(result, expected);
+
+    colored::control::unset_override();
+}
```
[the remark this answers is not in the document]
slarse: Overkill, there are just two tests in there.

crates/but/src/lib.rs
```diff
@@ -135,6 +136,21 @@ pub async fn handle_args(args: impl Iterator<Item = OsString>) -> Result<()> {
 
     // If no subcommand is provided, but we have source and target, default to rub
     let mut out = OutputChannel::new_with_optional_pager(output_format, use_pager);
+    // Initialize the global color theme, loading from a user config file if present.
+    {
+        let theme = dirs::config_dir()
+            .map(|dir| dir.join("gitbutler").join("but-theme.json"))
+            .filter(|p| p.exists())
+            .and_then(|p| theme::load(&p).ok())
+            .unwrap_or_default();
```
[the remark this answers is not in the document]
slarse: This is fine for now.

crates/but/src/lib.rs
```diff
@@ -135,6 +136,21 @@ pub async fn handle_args(args: impl Iterator<Item = OsString>) -> Result<()> {
...
+        tracing::warn!(
+            "{:?}",
+            dirs::config_dir().map(|dir| dir.join("gitbutler").join("but-theme.json"))
+        );
```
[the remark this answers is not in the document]
slarse: That warning should not be there!

estib-vega (approved):

PR #13325, opened 2026-04-15
Add specific compile step, and require compile step to complete

Caleb-T-Owens:
`dev` only required `package`, but the but-sdk compile was considered a `build`.

This adds a specific `compile` step for tasks that compiles rust, and fix the dependencies.

I’ve also updated the dependencies to use the `^` notation, which AFAIK means that we actually wait for the dependency to complete before continuing (which seems like sane default behaviour?) so this probably makes sense.


krlvi:
do we still need this PR?

Caleb-T-Owens:
We can close for now. I want to test and see if something like this is still needed, but I'll re-create it once it starts anoying me again

PR #13337, opened 2026-04-16
but-graph: detect force-push need when remote tip is owned by an upper stack segment

mtsgrd:
What we are fixing
------------------
When a stack contains two dependent branches (`top` on `bottom`) and a
previous combined push left `origin/bottom` pointing at the top
commit (`T`), splitting the stack locally so that `bottom` only
contains `B` again left GitButler reporting `bottom` as
`NothingToPush`. In reality the lower branch must be force-pushed to
remove `T` from `origin/bottom`, otherwise the remote keeps a commit
that locally now belongs to `top`.

How it was discovered
---------------------
The bug surfaced as a user report: with two stacked branches of one
commit each where the lower one had previously been pushed in its
combined form, GitButler thought both branches were already pushed.
Reproducing it against a small repo showed that
`but_workspace::ui::PushStatus` resolved to `NothingToPush` for the
lower branch even though `origin/bottom` was strictly ahead of the
local tip and divergent from where a fast-forward push would land.

Tracing the projection showed that
`Workspace::mark_remote_reachability` only sets stack-segment flags
for the local owners of those commits — it never adds them to the
matching local segment's `commits_on_remote`. Because
`PushStatus::derive_from_commits` keys force-push detection off
`commits_on_remote.is_empty()`, the lower branch silently degraded to
`NothingToPush`.

How the fix works
-----------------
The fix extracts `commits_on_remote` population into a new
`add_commits_on_remote` method that iterates stacks by segment index.
It performs two walks from each remote tracking branch:

  1. An all-parents walk that collects commits from fully-remote
     segments (the existing behavior, now isolated from the flag-
     setting logic). It stops at segments containing non-remote
     commits or belonging to another remote branch.

  2. A first-parent walk that detects non-integrated commits from
     upper stack segments still on the remote — the "branch split"
     case. The first-parent restriction avoids attributing merge-
     parent commits from sibling branches.

The set of upper-segment commit IDs (`above_commit_ids`) is
maintained cumulatively across the segment loop so each commit ID is
inserted at most once.

How it is tested
----------------
Two tests cover the new behavior:

  * `but-graph` snapshot test
    `stacked_bottom_remote_still_points_at_now_split_top` asserts the
    workspace projection: `bottom`'s `commits_on_remote` contains `T`
    (rendered as `🟣` with `⇣1`) while `top` is unaffected.

  * `but-workspace` API-level test of the same name asserts that
    `bottom`'s `push_status` is `UnpushedCommitsRequiringForce` with
    exactly the `T` commit in `commits_on_remote`, while `top`
    remains `CompletelyUnpushed`.

Both tests load a new fixture
`stacked-bottom-remote-still-points-at-now-split-top` added to the
respective scenarios scripts.

### DIFF
--- a/crates/but-graph/src/projection/workspace/init.rs
+++ b/crates/but-graph/src/projection/workspace/init.rs
@@ -9,7 +9,7 @@ use but_core::ref_metadata::{
     self, StackId,
     StackKind::{Applied, AppliedAndUnapplied},
 };
-use gix::refs::Category;
+use gix::{ObjectId, refs::Category};
 use itertools::Itertools;
 use petgraph::{
     Direction,
@@ -386,7 +386,7 @@ impl Graph {
         target_ref: Option<&TargetRef>,
         target_commit: Option<&TargetCommit>,
         additional: impl IntoIterator<Item = SegmentIndex>,
-    ) -> Option<(gix::ObjectId, SegmentIndex)> {
+    ) -> Option<(ObjectId, SegmentIndex)> {
         // It's important to not start from the tip, but instead find paths to the merge-base from each stack individually.
         // Otherwise, we may end up with a short path to a segment that isn't actually reachable by all stacks.
         let (tips, actual_tip) = match tip {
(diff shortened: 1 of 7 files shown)


mtsgrd:
@Caleb-T-Owens is this something you'd be interested in looking at?

Caleb-T-Owens:
@mtsgrd, yes. I'd like to take a look tomorrow. My knowledge of the WS projection & graph is not super deep so it might be beneficial to hop on a call and have you walk me through the change.

mtsgrd:
@Caleb-T-Owens I've revised the commit and I think it's more in line what we discussed earlier.

Byron:
Thanks a lot for the fix! The tests are very valuable as they show an obviously real case that we need to cover.

Note that I did remove a small portion of the algorithm as it wasn't covered by tests, while thinking that it *should* not be required. This might be wrong, but I'd rather run into it than adding complexity without a 'trigger'.

I think there are two aspects to this:

* First is the question if remote commits that that are also in the workspace should be listed as remote commits and should be included in the `commits_on_remote` vec even if they are already included in the workspace through other means. I think the answer to this is positive as this is needed to offer ways to deal with diverged remote tracking branches.
    - Here I see a future where we don't have to inspect these arrays at all, but use the graph to know what's going on, independently of the workspace projection.
* Second is if the function `PushStatus::derive_from_commits` and how it's implemented. It's one of those hacks where I tried to just keep something that seemed like it's working similar to how it was implemented. However, it's quite abstract even though now we have a graph to traverse, which can make it much clearer what it actually figures out: The relationship between a local and remote tracking branch. When we are improving the methods to decide what to do (i.e. fast forward, merge, rebase, reset), these methods will also need to use the graph to make the right decisions. So why not start here? 

While this PR should be merged on the first point alone, let me try to see if the second point is already in reach.

Byron (approved):

Byron:
Let's merge this and put the second point into a follow-up.

PR #13339, opened 2026-04-16
Fetch/Push - only Git executable support

krlvi:


crates/gitbutler-repo-actions/src/repository.rs
```diff
@@ -143,198 +142,109 @@ impl RepoActionsExt for Context {
...
-            // with the refspec force-format which seems to override them, leading to incorrect results
-            // in conjunction with `force_push_protection`.
-            let prefix = if with_force && !use_git_executable {
-                "+"
-            } else {
-                Default::default()
-            };
-            format!("{prefix}{}:refs/heads/{}", head, branch.branch())
-        });
+        let refspec = refspec.unwrap_or_else(|| format!("{}:refs/heads/{}", head, branch.branch()));
```
Caleb-T-Owens, line 145: Is changing how this refspec is created safe?

Caleb-T-Owens (approved):
Left one comment on a refspec change, but other than that it seems fine

crates/gitbutler-repo-actions/src/repository.rs
krlvi: It seems like previously we were adding a prefix only if the git executable was **not** used. 
Now we directly omit a prefix and construct the same thing

PR #13342, opened 2026-04-16
ci: add minimal workflow for building gitbutler-lite

slarse:
## 🧢 Changes
Minimal workflow for building distributions of gitbutler-lite across macOS (aarch64 only), Linux (x86-64 only) and Windows (x86-64 only). The artifacts are uploaded to the workflow run's artifact store only, they're not distributed elsewhere at this time.

I haven't tried the Windows build yet but... it builds at least.

## Important
This is _very_ minimal and not production ready. It's just to get a starting point. Things that aren't fixed/checked yet include:

* CSP
* Preventing navigation to the Internet
* Signing of executables
  - On macOS, you may get `"gitbutler-lite" is damaged`, in which case you can run `xattr -c <path/to/application.app>` to explicitly mark it as safe.
  - DO NOT DO THIS unless you are a dev on this project, it won't be necessary once we start signing the app
* Reloading the app doesn't work (you get a blank screen as assets refuse to load)
* [Bitwarden's Desktop client](https://github.com/bitwarden/clients) has a very similar setup with Napi and electron-builder that we can draw inspiration from

### DIFF
--- a/apps/lite/package.json
+++ b/apps/lite/package.json
@@ -1,5 +1,7 @@
 {
 	"name": "@gitbutler/lite",
+	"productName": "gitbutler-lite",
+	"homepage": "https://gitbutler.com",
 	"private": true,
 	"version": "0.0.1",
 	"type": "module",
@@ -32,6 +34,31 @@
 		],
 		"mac": {
 			"target": "dmg"
+		},
+		"linux": {
+			"target": [
+				"deb",
+				"appimage"
+			],
+			"maintainer": "GitButler <gitbutler@gitbutler.com>"
+		},
+		"win": {
+			"target": [
+				"portable"
+			]
+		},
+		"deb": {
+			"packageName": "gitbutler-lite",
+			"artifactName": "${productName}-${version}-${arch}.${ext}",
+			"depends": [
+				"libnotify4",
+				"libxtst6",
+				"libnss3",
+				"libxss1"
+			]
+		},
+		"appImage": {
+			"artifactName": "${productName}-${version}-${arch}.${ext}"
 		}
 	},
 	"dependencies": {
@@ -46,7 +73,6 @@
 		"@tanstack/react-query-devtools": "^5.91.3",
 		"@tanstack/react-router": "^1.167.4",
 		"effect": "^3.20.0",
-		"electron": "^41.1.0",
 		"electron-devtools-installer": "^4.0.0",
 		"react": "^19.2.4",
 		"react-dom": "^19.2.4",
@@ -61,6 +87,7 @@
 		"@typescript/native-preview": "7.0.0-dev.20260311.1",
 		"@vitejs/plugin-react": "^5.2.0",
 		"babel-plugin-react-compiler": "^1.0.0",
+		"electron": "^41.1.0",
 		"concurrently": "^9.2.1",
 		"cross-env": "^10.1.0",
 		"electron-builder": "^26.8.1",
--- a/apps/lite/ui/vite.config.ts
+++ b/apps/lite/ui/vite.config.ts
@@ -15,6 +15,7 @@ export default defineConfig({
 			},
 		}),
 	],
+	base: "./",
 	build: {
 		outDir: "../dist/ui",
 		emptyOutDir: true,
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -20,7 +20,7 @@
 			],
 			"project": ["electron/src/**!", "ui/src/**!"],
 			"includeEntryExports": true,
-			"ignoreDependencies": ["@tanstack/eslint-plugin-query"],
+			"ignoreDependencies": ["@tanstack/eslint-plugin-query", "electron"],
 		},
 	},
 }
--- a/.github/workflows/lite.yml
+++ b/.github/workflows/lite.yml
@@ -1,0 +1,58 @@
+name: "Lite"
+on:
+  schedule:
+    - cron: "0 2 * * *"
+  workflow_dispatch: {}
+
+env:
+  RUST_BACKTRACE: full
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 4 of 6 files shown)


krlvi (approved):
Ship it :shipit:  (when ready)

slarse:
After some fiddling around, I've now discovered the brand new version of the `GLIBC` mismatch for Lite, which is unfortunately _way_ less clear than in Tauri.

```bash
$ ./@gitbutlerlite
JavaScript error occurred in the main process
Uncaught Exception:
Error: Cannot find native binding. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
    at file:///home/slarse/unpack/opt/gitbutler-lite/resources/app.asar/node_modules/@gitbutler/but-sdk/src/generated/index.js:567:11
    at ModuleJob.run (node:internal/modules/esm/module_job:430:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:661:26)
    at async node:electron/js2c/browser_init:2:145250
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:116:5)
```

What it doesn't show you is the reason for that crash. As the only kind of error that exists in this world is a linker error, we find the culprit with some linker debugging.

```bash
$ LD_DEBUG=libs @gitbutlerlite
     92213:	@gitbutlerlite: error: symbol lookup error: undefined symbol: nspr_use_zone_allocator (fatal)
     92213:	/lib/x86_64-linux-gnu/libc.so.6: error: version lookup error: version `GLIBC_2.38' not found (required by /home/slarse/unpack/opt/gitbutler-lite/resources/app.asar.unpacked/node_modules/@gitbutler/but-sdk/src/generated/gitbutler-sdk.linux-x64-gnu.node) (fatal)
A JavaScript error occurred in the main process
[ ... same error as above ... ]
```

So that's cool. But I think that has Linux mostly sorted for a basic build, moving on to macOS.

And yes, I'll fiddle with settings s.t. the executable isn't named `@gitbutlerlite` :D

slarse:
> Ship it :shipit: (when ready)

I will make sure to have _something_ in today. I'm taking it pretty slow and deliberate because I really want to understand exactly how we build this, and also be able to explain it to @estib-vega so we've got a bus factor of >=2 :)

estib-vega:
@slarse The but-sdk, as currently configured, is only building the NAPI bindings for macOS, and that might the source of the error you're experiencing

slarse:
> @slarse The but-sdk, as currently configured, is only building the NAPI bindings for macOS, and that might the source of the error you're experiencing

I haven't really looked at the `but-sdk` config, but it appears to just build for whatever platform it's running on, so I'm not facing any issues there :)

The note I made above is a common linker issue on Linux: you build with a particular version of `glibc` and run on a system with an _older_ version. That typically yields an error message like this from the linker:

```
     92213:	/lib/x86_64-linux-gnu/libc.so.6: error: version lookup error: version `GLIBC_2.38' not found (required by /home/slarse/unpack/opt/gitbutler-lite/resources/app.asar.unpacked/node_modules/@gitbutler/but-sdk/src/generated/gitbutler-sdk.linux-x64-gnu.node) (fatal)
 ```
 
 But _unfortunately_, that's not surfaced all the way up. Instead, the napi layer just prints "Cannot find native binding", even though that's not actually the problem :s

**Comment by estib-vega** [comment]
If you get a nice 'damaged' error in macOS, I found this solution for it:

Screenshot 2026-04-17 at 15 12 18


`xattr -c <path/to/application.app>`

**Comment by slarse** [comment]
> If you get a nice 'damaged' error in macOS, I found this solution for it:
> `xattr -c <path/to/application.app>`

Context: I've not added signing/notarizing to the mac distro yet. I'll wait with that a little bit until I've addressed some major flaws. So the above might happen for you and you need the workaround (curiously it did not happen for me).

**Review by estib-vega — APPROVED** [review]

File: apps/lite/ui/vite.config.ts [hunk 1]
```diff
@@ -15,6 +15,7 @@ export default defineConfig({
 			},
 		}),
 	],
+	base: "./",
```
slarse, line 18: Without this, the bundles couldn't find the assets. There may be a different solution to this but this is what I found.

apps/lite/package.json
```diff
@@ -32,6 +34,31 @@
...
+				"libnotify4",
+				"libxtst6",
+				"libnss3",
+				"libxss1"
+			]
+		},
+		"appImage": {
+			"artifactName": "${productName}-${version}-${arch}.${ext}"
 		}
 	},
```
slarse, line 63: Very minimal targets for just producing _something_. Nothing is signed and some executables have very funky names. For example, if you install the deb package, you get the wonderful `@gitbutlerlite` executable on your path!

apps/lite/package.json
```diff
@@ -46,7 +73,6 @@
 		"@tanstack/react-query-devtools": "^5.91.3",
 		"@tanstack/react-router": "^1.167.4",
 		"effect": "^3.20.0",
-		"electron": "^41.1.0",
```
slarse, line 49: This is significant: `electron-builder` prohibits Electron from being a direct dependency. It must be a dev dependency. I hope this doesn't cause too much development headache but the `electron-builder` just straight up refuses packaging otherwise.

knip.jsonc
```diff
@@ -20,7 +20,7 @@
 			],
 			"project": ["electron/src/**!", "ui/src/**!"],
 			"includeEntryExports": true,
-			"ignoreDependencies": ["@tanstack/eslint-plugin-query"],
+			"ignoreDependencies": ["@tanstack/eslint-plugin-query", "electron"],
```
slarse, line 23: As `electron` was moved to dev dependency, the linter started complaining about undeclared deps.

PR #13348, opened 2026-04-16
Speed up playwright and blackbox workflow

mtsgrd:
Parallelize independent build steps (Rust, SvelteKit, Playwright
install, tauri-driver) so they run concurrently instead of sequentially.

Skip node setup, SvelteKit, and test steps on master pushes since those
runs only exist to warm the Rust cache.

Add paths-ignore to both workflows so PRs that only touch docs, scripts,
config files, or unrelated crates (CLI, TUI, debugging tools) skip
the e2e jobs entirely.

Install only chromium and webkit (no Firefox) and run apt-get update
before Playwright install to avoid stale package index 404s.

Upload artifacts only on failure instead of on every non-cancelled run.

### DIFF
--- a/.github/workflows/test-e2e-blackbox.yml
+++ b/.github/workflows/test-e2e-blackbox.yml
@@ -5,7 +5,20 @@ on:
       - master
   pull_request:
     paths-ignore:
+      - "*.md"
+      - ".cargo/**"
+      - ".claude/**"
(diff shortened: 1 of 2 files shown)


Byron (approved):
This seems to have the desired effect and CI is still green.

PR #13367, opened 2026-04-17
Make `but status` 5x faster in some scenarios

davidpdrsn:
I noticed that `but status` would occasionally be slow especially if you hadn't pulled in a while.

With the openclaw repo (which moves insanely quickly):

```
hyperfine --warmup 2 "./target/release/but -C /Users/davidpdrsn/code/openclaw status"
Benchmark 1: ./target/release/but -C /Users/davidpdrsn/code/openclaw status
  Time (mean ± σ):     455.4 ms ±   1.2 ms    [User: 443.3 ms, System: 162.3 ms]
  Range (min … max):   453.5 ms … 457.3 ms    10 runs
```

In the same repo `git status` takes ~30 ms.

I ran a flamegraph and found most of the time was spent in [`gitbutler_branch_actions::base::target_to_base_branch`](https://github.com/gitbutlerapp/gitbutler/blob/b5f62abf1dc5257570755cc8cd755eb5ea58b653/crates/gitbutler-branch-actions/src/base.rs#L344) which was called twice:

1. Once as part of [`but_api::legacy::forge::list_reviews`](https://github.com/gitbutlerapp/gitbutler/blob/b5f62abf1dc5257570755cc8cd755eb5ea58b653/crates/but-api/src/legacy/forge.rs#L170) to get the remote url of the target branch.
2. And again via [`but_api::legacy::virtual_branches::get_base_branch_data`](https://github.com/gitbutlerapp/gitbutler/blob/b5f62abf1dc5257570755cc8cd755eb5ea58b653/crates/but/src/command/legacy/status/mod.rs#L321) to get various upstream data such as how far behind you are.

## `list_reviews`

For 1 (getting the remote url) we're calling `target_to_base_branch` just to get a single field. We don't even look at all the commits it traverses. The first commit addresses that by making a dedicated helper just to get the remote url. This gave a 2x speed up.

## `get_base_branch_data`

For 2 the real culprit was [`first_parent_commit_ids_until`](https://github.com/gitbutlerapp/gitbutler/blob/b5f62abf1dc5257570755cc8cd755eb5ea58b653/crates/gitbutler-repo/src/lib.rs#L8) which is called three times from `target_to_base_branch`:

- For finding `upstream_commits` which is used for the `(upstream) ⏫ 2 new commits` line
- `diverged_ahead` and `diverged_behind` both of which aren't used by `but status`

After a bunch of back and forth with codex about what a "hidden walk" is I think I've come up with a faster version of `first_parent_commit_ids_until` that behaves the same. The main issue was `.with_hidden(Some(stop_before))` which appears to do perform a ton of extra work including decompressing commits. The flamegraph looked like this:

image

That was fixed in the second commit and gave another 3x speed up.

## The result

After both of these changes `but status` is down to 90 ms in openclaw:

```
hyperfine --warmup 2 "./target/release/but -C /Users/davidpdrsn/code/openclaw status"
Benchmark 1: ./target/release/but -C /Users/davidpdrsn/code/openclaw status
  Time (mean ± σ):      89.3 ms ±   2.0 ms    [User: 71.8 ms, System: 174.0 ms]
  Range (min … max):    87.1 ms …  98.4 ms    32 runs
```

This also impacts the TUI since it runs `but status` before starting and again after each operation.

### DIFF
--- a/crates/gitbutler-branch-actions/src/base.rs
+++ b/crates/gitbutler-branch-actions/src/base.rs
@@ -100,6 +100,12 @@ pub fn get_base_branch_data(ctx: &Context) -> Result<BaseBranch> {
     Ok(base)
 }
 
+#[instrument(skip(ctx), err(Debug))]
+pub fn get_base_branch_remote_url(ctx: &Context) -> Result<String> {
+    let target = default_target(&ctx.project_data_dir())?;
+    remote_url_of_target_to_base_branch(ctx, &target)
+}
+
 /// Restore the default target metadata if it is missing in the currently configured storage
 /// location while an existing `gitbutler/workspace` ref already proves the repository was
 /// initialized before.
@@ -395,22 +401,7 @@ pub(crate) fn target_to_base_branch(ctx: &Context, target: &Target) -> Result<Ba
         None => target.remote_url.clone(),
     };
 
-    // Fallback to the remote URL of the branch if the target remote URL is empty
-    let remote_url = if target.remote_url.is_empty() {
-        let remote = repo.find_remote(target.branch.remote()).context(format!(
-            "failed to find remote for branch {}",
-            target.branch.fullname()
-        ))?;
-        remote
-            .url(gix::remote::Direction::Fetch)
-            .map(|url| url.to_bstring().to_string())
-            .context(format!(
-                "failed to get remote url for {}",
-                target.branch.fullname()
-            ))?
-    } else {
-        target.remote_url.clone()
-    };
+    let remote_url = remote_url_of_target_to_base_branch(ctx, target)?;
 
     let branch_name = target.branch.fullname();
     let remote_name = target.branch.remote().to_string();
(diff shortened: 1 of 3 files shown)


crates/gitbutler-branch-actions/src/base.rs
```diff
@@ -445,6 +436,29 @@ pub(crate) fn target_to_base_branch(ctx: &Context, target: &Target) -> Result<Ba
     Ok(base)
 }
 
+fn remote_url_of_target_to_base_branch(ctx: &Context, target: &Target) -> Result<String> {
+    let repo = &*ctx.repo.get()?;
+
+    // Fallback to the remote URL of the branch if the target remote URL is empty
```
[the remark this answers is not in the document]
davidpdrsn: `OnDemand::get` does no locking, just some runtime borrow checks so its very cheap.

crates/gitbutler-repo/src/lib.rs
```diff
@@ -4,21 +4,44 @@ mod commands;
 mod traversal {
     use anyhow::Result;
 
-    /// Return first-parent ancestors from `from` until `stop_before`, excluding `stop_before`.
+    /// Return the prefix of `from`'s first-parent history that is not in `stop_before`'s ancestry.
```
Byron: Really just a nit, but I don't understand `prefix` in relation to commits. For me this would mean a partial hash, i.e. prefix, but that's clearly not meant there.

crates/gitbutler-repo/src/lib.rs
```diff
@@ -4,21 +4,44 @@ mod commands;
...
+    ///
+    /// The returned commits are ordered from `from` backwards along the first-parent chain, excluding
+    /// the first commit that is reachable from `stop_before` by ancestry.
+    ///
+    /// This matches the semantics of a first-parent walk with `stop_before` hidden, but avoids the
+    /// up-front hidden-side graph painting that makes `with_hidden(stop_before)` expensive in large
```
Byron, line 14: I'm really sorry to not have documented this behavior enough because it's very well known to me. I was also known that this will bite eventually, but I considered it a useful nudge to use the graph `but-graph` more when the need arises.

As a matter of fact, this function is only used by old code or old style code because all the needs we have should be served by that one traversal that is made by `but-graph`.

Byron (commented):
Thanks so much for looking into this.

I'm also happy that `first_parent_commit_ids_until` is now fast again, which makes me think that maybe `with_hidden()` could also leverage a similar system to be faster by default.

Just to restate, the plan here is to use the `but-graph` for everything, which is one traversal that we reuse, and that already is as fast as it can be. Until this happens, these utilities should certainly serve us well 🙏

Byron (approved):

krlvi:
This is really awesome!

crates/gitbutler-repo/src/lib.rs
davidpdrsn: Thats fair. I've reworded it a bit to not say "prefix".

PR #13371, opened 2026-04-18
Add noShadow app setting for macOS window shadows

schacon:


krlvi:
@schacon we are likely better off putting this config in `UiSettings` in `crates/but-settings/src/app_settings.rs` similar to `use_native_title_bar`.
The function `create` in `crates/gitbutler-tauri/src/window.rs` already has the settings loaded there, checking `use_native_title_bar` so `no_shadow` can be just like it. 
You can then set it via `~/Library/Application\ Support/gitbutler/settings.json` 

There's a plan to migrate the app settings to a dedicated gitconfig file at some point in the future but haven't been able to prioritize it, and it's nicer to keep it consistent for now. Plus we avoid introducing libgi2

schacon:
@krlvi is this better?

krlvi:
> @krlvi is this better?

perfect. i pushed a CI fix

krlvi:
btw @schacon i had to rewrite your commits since yours were missing signatures

PR #13375, opened 2026-04-18
Make snapshotting faster

davidpdrsn:
Baseline
```
$ hyperfine --warmup 2 "./target/release/but commit empty dp-branch-123"
Benchmark 1: ./target/release/but commit empty dp-branch-123
  Time (mean ± σ):     167.5 ms ±   2.4 ms    [User: 126.7 ms, System: 81.8 ms]
  Range (min … max):   163.4 ms … 173.4 ms    17 runs
```

After
```
$ hyperfine --warmup 2 "./target/release/but commit empty dp-branch-123"
Benchmark 1: ./target/release/but commit empty dp-branch-123
  Time (mean ± σ):      93.3 ms ±   4.8 ms    [User: 48.5 ms, System: 83.9 ms]
  Range (min … max):    86.7 ms … 104.8 ms    33 runs
```

Not a huge speedup (~2x) but still noticeable in the TUI.

Was essentially the same issue as https://github.com/gitbutlerapp/gitbutler/pull/13367 where `.with_hidden(...)` is slow. First commit contains the interesting changes. The second commit just cleans up the organization a bit.


Byron (commented):
Thanks a lot!

I think it's fair to go with the cached-graph-with-merge-base-calls workaround until we get to use `but-graph` exclusively to avoid any extra traversal, cached or not.

PR #13378, opened 2026-04-18
feat: extract askpass to dedicated crate

krlvi:
This will help us get rid of `gitbutler-branch-actions` sooner. This is just a move + docs cleanup

### DIFF
--- a/crates/but-askpass/Cargo.toml
+++ b/crates/but-askpass/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "but-askpass"
+version = "0.0.0"
+edition.workspace = true
+authors.workspace = true
+publish = false
+rust-version.workspace = true
+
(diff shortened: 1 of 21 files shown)


crates/but-askpass/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[lib]
+doctest = false
+
+[dependencies]
+but-core.workspace = true
+
+serde.workspace = true
+tokio = { workspace = true, features = ["sync"] }
+log = "^0.4"
```
Byron, line 17: I'd probably quickly replace this with `tracing`.

Byron (approved):
That's great! I wonder how much more we know we want to keep from `gitbutler-branch-actions`, but also wouldn't be surprised if that's mostly it.

PR #13395, opened 2026-04-20
refactor: remove vendored openssl from git2

krlvi:
This should no longer be needed at this point


krlvi:
nevermind - this only removes vendoring

Byron:
Yeah, it would have been so nice if OpenSSL could just be dropped. My feeling is that libgit2 would allow that, but `git2` just doesn't expose it. In any case, I think it's just going to be a couple of months until `libgit2` will be gone completely.

PR #13401, opened 2026-04-20
refactor(but): add symbols to theme

slarse:
## 🧢 Changes
An abstraction to provide symbols styled with the current theme that can either be turned into raw ANSI-escaped strings or a Ratatui Spans.

### DIFF
--- a/crates/but/src/command/update.rs
+++ b/crates/but/src/command/update.rs
@@ -5,7 +5,7 @@ use but_settings::AppSettings;
 use but_update::{AppName, CheckUpdateStatus, check_status};
 use colored::Colorize;
 
-use crate::{args::update, utils::OutputChannel};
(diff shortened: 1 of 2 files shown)


crates/but/src/command/update.rs
```diff
@@ -46,11 +46,12 @@ fn check_for_updates(out: &mut OutputChannel, app_settings: &AppSettings) -> Res
 }
 
 fn print_human_output(writer: &mut dyn std::fmt::Write, status: &CheckUpdateStatus) -> Result<()> {
+    let t = theme::get();
     if status.up_to_date {
         writeln!(
             writer,
             "{} You're running the latest version ({})",
-            "✓".green().bold(),
+            t.sym().success,
```
slarse, line 54: This is just here to show a very minimal example of how to use it.

estib-vega (approved):

PR #13402, opened 2026-04-20
fix: remove unused dev profile option

slarse:
## 🧢 Changes
`cargo` is constantly warning me that there is no package with this name. Seems safe to remove.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -340,7 +340,6 @@ incremental = false
 [profile.dev.package]
 foldhash = { opt-level = 3 }
 gix-imara-diff = { opt-level = 3 }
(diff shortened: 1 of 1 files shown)


Byron (approved):
Definitely a gonner!
Strange I never noticed it, thanks for the fix!

There might also be a way to deny this, or turn this warning into an error automatically.

slarse:
> Definitely a gonner! Strange I never noticed it, thanks for the fix!
> 
> There might also be a way to deny this, or turn this warning into an error automatically.

I couldn't find a way. It's coming from Cargo itself, not from `rustc`, so adding e.g. `-D warnings` to rustflags doesn't work.

Opus also failed to find a solution.

PR #13404, opened 2026-04-20
id: use cached stacks instead of generating RefInfo

jonathantanmy2:
Generate CLI IDs from the stacks in the workspace struct instead of generating a RefInfo struct. This improves performance for commands that do not need the extra information in RefInfo, which is all commands except for `status`.

Currently, stacks information is copied, but I'm looking into reducing the number of copies (probably by using Rc). This will be done in another PR.

Note that there are some algorithmic changes that had to be done as a result of migrating to the stacks in the workspace struct:
 - The order of stacks are reversed in RefInfo but not the workspace struct, so IdMap is changed to also reverse them.
 - Remote commits that match local commits are not in RefInfo but are in the workspace struct, so `status` is changed to also filter them out.

### DIFF
--- a/crates/but/src/command/legacy/status/mod.rs
+++ b/crates/but/src/command/legacy/status/mod.rs
@@ -3,7 +3,7 @@
     reason = "VirtualBranchesHandle should be replaced with ctx.workspace_* helpers"
 )]
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 
 use anyhow::Context as _;
 use assignment::FileAssignment;
@@ -12,7 +12,11 @@ use but_api::diff::ComputeLineStats;
 use but_core::{RepositoryExt, TreeStatus, ui};
 use but_ctx::Context;
 use but_forge::ForgeReview;
-use but_workspace::{ref_info::LocalCommitRelation, ui::PushStatus};
+use but_graph::SegmentIndex;
+use but_workspace::{
+    ref_info::{Commit, LocalCommit, LocalCommitRelation, Segment},
+    ui::PushStatus,
+};
 use gitbutler_branch_actions::upstream_integration::BranchStatus as UpstreamBranchStatus;
 use gitbutler_operating_modes::OperatingMode;
 use gitbutler_stack::StackId;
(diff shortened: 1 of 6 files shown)


krlvi (approved):

crates/but/src/command/legacy/status/mod.rs
```diff
@@ -245,21 +252,44 @@ async fn build_status_context<'a>(
...
+        let head_info = but_workspace::head_info(
             &*ctx.repo.get()?,
             &meta,
             but_workspace::ref_info::Options {
                 expensive_commit_info: true,
                 ..Default::default()
             },
-        )?
+        )?;
+        std::mem::drop(guard);
```
Byron: While I understand that the code wasn't structured correctly before, something I would like to avoid is to reacquire the guard multiple times and instead would like to encourage refactoring it so that the guard is acquired once.

crates/but/src/command/legacy/status/mod.rs
```diff
@@ -245,21 +252,44 @@ async fn build_status_context<'a>(
...
+        let mut segments_by_id = HashMap::<SegmentIndex, Segment>::new();
+        let mut local_commits_by_id = HashMap::<gix::ObjectId, LocalCommit>::new();
+        let mut remote_commits_by_id = HashMap::<gix::ObjectId, Commit>::new();
+        for stack in head_info.stacks {
+            for mut segment in stack.segments {
+                for local_commit in std::mem::take(&mut segment.commits) {
```
Byron: Here we effectively consume parts of the contents of the structure, which I fear will confuse downstream code that might read it and expect something there. While this apparently works here, I'd like to avoid dangerous precedents. So a copy would be preferred, I don't think we have a reason to avoid allocations or clones just yet.

Byron (commented):
Thanks so much!
There is, however, one bit that I'd love to see improved, please see the comments below.

> Currently, stacks information is copied, but I'm looking into reducing the number of copies (probably by using Rc). This will be done in another PR.

I don't think it matters, unless you have benchmarks that prove otherwise. What I'd like to avoid is to add Rc's to fields in structs.
My preference here is to keep it simple and stupid for as long as possible, unless we really have no other option than to make intrusive changes to our structs to improve performance. My guess is that we are in a very good position if this is the last remaining avenue to have get performance gains.

crates/but/src/command/legacy/status/mod.rs
jonathantanmy2: Done (used `ctx.workspace_and_db_with_perm` instead).

crates/but/src/command/legacy/status/mod.rs
jonathantanmy2: Done.

jonathantanmy2:
> I don't think it matters, unless you have benchmarks that prove otherwise. What I'd like to avoid is to add Rc's to fields in structs.
> My preference here is to keep it simple and stupid for as long as possible, unless we really have no other option than to make intrusive changes to our structs to improve performance. My guess is that we are in a very good position if this is the last remaining avenue to have get performance gains.

Makes sense. OK I won't follow-up with my `Rc` idea then.

PR #13405, opened 2026-04-20
Reject cross-stack file deletions during amend

mtsgrd:
## Summary

- Add `reject_cross_stack_file_deletions` guard in `but-workspace` that prevents amending a whole-file deletion into a stack when that file is already committed in another stack.
- The guard is called from both amend paths:
  - **New path** (`but rub`, frontend/Tauri) → `but_api::commit_amend()` → `but_workspace::commit::commit_amend()` → `create_commit()` → `create_tree()`
  - **Legacy path** (`but amend`) → `amend_diff_specs()` → `create_commit_and_update_refs_with_project()` → `create_commit()` → `create_tree()`
  
  Both converge on the same `create_commit()` → `create_tree()`, which is where the deletion is applied. The guard is placed before this shared code in each entry point.
- Hunk-level amends are unaffected — their content naturally won't match the target commit's tree, so they already fail safely.
- Include a test that constructs a workspace with two stacks (one modifying a file, one deleting it), calls the guard directly, and asserts it rejects with an error mentioning the conflicting file and the owning stack.

## Context

When a file is committed in stack A and you delete it on disk, then amend that deletion into stack B, the amend succeeds silently. But the resulting workspace merge sees a modify/delete conflict and skips stack B entirely — losing all of B's other changes from the workspace tree.

`create_tree()` has a cherry-pick merge that catches modify/delete conflicts when file content differs between the workspace tree and the target commit tree. But when the target stack never modified the file (content is identical), the cherry-pick cleanly applies the deletion — silently creating the bad state. The guard catches this case by checking whether the deleted path appears in any other stack's diff.

## Test plan

- [x] `cargo test -p but-workspace reject_file_deletion` passes
- [x] Guard rejects cross-stack file deletions in both amend paths

### DIFF
--- a/crates/but-api/src/commit/amend.rs
+++ b/crates/but-api/src/commit/amend.rs
@@ -42,6 +42,7 @@ pub(crate) fn commit_amend_only_impl(
 ) -> anyhow::Result<CommitCreateResult> {
     let mut meta = ctx.meta()?;
     let (repo, mut ws, _) = ctx.workspace_mut_and_db_with_perm(perm)?;
+    but_workspace::commit::reject_cross_stack_file_deletions(&ws, &repo, commit_id, &changes)?;
     let editor = Editor::create(&mut ws, &mut meta, &repo)?;
 
     let but_workspace::commit::CommitAmendOutcome {
--- a/crates/but-workspace/src/commit/commit_amend.rs
+++ b/crates/but-workspace/src/commit/commit_amend.rs
@@ -1,6 +1,9 @@
 //! An action to amend an existing commit with selected changes.
 
+use std::collections::BTreeSet;
+
 use anyhow::{Result, bail};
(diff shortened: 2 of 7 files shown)


mtsgrd:
@Caleb-T-Owens the solution in this PR is not efficient, but the problem it highlights is real. Could you help me figure out what we want to do here?

Caleb-T-Owens:
Can you help me understand what is going wrong here?

I'm trying to understand why this isn't caught as a conflict in the workspace commit

Caleb-T-Owens:
Your test case also doesn't seem to align with your description of the problem?

mtsgrd:
Ugh, struggling to reproduce. Sorry for distracting you @Caleb-T-Owens, I will re-open this PR once I figure out what happened yesterday.

PR #13413, opened 2026-04-20
Reduce usage of VirtualBranchesHandle in but cli

krlvi:


crates/but/src/command/legacy/workspace_target.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use anyhow::{Context as _, Result};
+use but_core::sync::RepoShared;
+use but_ctx::Context;
+
+/// Resolve the effective target commit OID from workspace projection data.
+fn target_oid_from_workspace(
+    workspace: &but_graph::projection::Workspace,
+) -> Result<gix::ObjectId> {
+    workspace
```
Byron, line 16: Since this is a very common request, it might be good to have this method on the workspace type itself.

crates/but/src/command/legacy/branch/list.rs
```diff
@@ -47,12 +42,17 @@ pub fn list(
         Some(but_workspace::legacy::StacksFilter::InWorkspace),
     )?;
 
-    // Filter out empty branches unless --empty is requested
-    let target_oid_for_filter: Option<gix::ObjectId> = if !show_empty {
-        get_target_oid(ctx).ok()
+    // Resolve the target once for all target-based filtering and calculations we may perform.
+    let target_oid: Option<gix::ObjectId> = if !show_empty || ahead || check_merge {
+        let guard = ctx.shared_worktree_access();
```
Byron, line 47: Getting a temporary `guard` is to be avoided., and we should try to refactor the code so we have a guard where we need it and one that covers the entire section that it needs to cover.
It's definitely something to watch out for and to try to clean up.

Byron (commented):
Thanks for your help, it's much appreciated!

PR #13428, opened 2026-04-21
Check generated SDK types in CI

mtsgrd:
When d23c1dbf migrated type generation from ts-rs to but-ts, the CI job
that verified generated types were up-to-date was removed and never
replaced. This meant Rust type changes could land without regenerating
the TypeScript definitions, causing silent drift.

- Make but-ts idempotent: strip any previous generated block (from the
  marker line onward) before appending, so running `build:types` multiple
  times produces the same output without needing to truncate first.
- Add `scripts/generate-sdk-types.sh` that formats and fails if `git diff`
  is non-empty.
- Add a generated-types freshness check to the existing
  `but-sdk-build-check` CI job, which already runs `pnpm build:sdk`.
- Register `but_error::Code` with schemars / `register_sdk_type!()` behind
  an `export-schema` feature so `but-ts` emits it. It was previously only
  present as a transient `$defs` dependency that silently disappeared.
- Regenerate types to pick up `Code`, the `PushResult` doc link fix, and
  new `WatcherGitRemoteActivity` variants.


scripts/generate-sdk-types.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+git_root="$(git rev-parse --show-toplevel 2>/dev/null)"
+cd "$git_root"
+
+cargo run -p but-ts -- --output packages/but-sdk/src/generated/index.d.ts
```
estib-vega: I think you probably want this to be idempotent.

Generating the index.d.ts file is a two step operation:
1. `pnpm --filter @gitbutler/but-sdk build:napi` that creates or overrides the file if it exists and adds the function definitions
2. `pnpm --filter @gitbutler/but-sdk build:types` that appends the type definitions from but-ts

Only calling the second step (if an index.d.ts is already present) will re-append the types to that file. 

There's a top-level command `pnpm build:sdk` that can be call from the repository root, and that will execute both operations in order

estib-vega: `pnpm build:sdk` is idempotent, since the `build:napi` step will override the file if it exists, and that's probably what you want.

estib-vega: > Make but-ts idempotent: strip any previous generated block (from the
marker line onward) before appending, so running build:types multiple
times produces the same output without needing to truncate first.

Ah, just read this. Still the rest of the comment holds

crates/but-ts/src/main.rs
```diff
@@ -98,7 +97,13 @@ fn main() -> anyhow::Result<()> {
...
     if output_path.exists() {
         let existing = std::fs::read_to_string(output_path)?;
-        final_output = format!("{}\n{}", existing, ts_output);
+        // Strip any previous but-ts output so the command is idempotent.
+        let prefix = if let Some(idx) = existing.find(MARKER) {
+            existing[..idx].trim_end().to_string() + "\n\n"
+        } else {
+            existing
+        };
+        final_output = format!("{prefix}{ts_output}");
```
estib-vega, line 106: Nice

.github/workflows/push.yaml
estib-vega, line 139: I think `but-sdk-build-check` was already doing the check

estib-vega: Ohhh it's not doing a freshness check, it's only checking that the types are not broken.
 Maybe we can combine your new script with this? Or have your script also check the type consistency and just replace this one

mtsgrd: Just realized this, can you take another look? It should be done the way it used to be now.

estib-vega: Yeap, perfect 👌

crates/but-error/Cargo.toml
```diff
@@ -9,8 +9,13 @@ rust-version.workspace = true
 [lib]
 doctest = true
 
+[features]
+export-schema = ["dep:schemars", "dep:but-schemars"]
+
 [dependencies]
 anyhow.workspace = true
+schemars = { workspace = true, optional = true }
+but-schemars = { path = "../but-schemars", optional = true }
```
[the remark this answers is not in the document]
estib-vega: @mtsgrd This is a good suggestion

krlvi (approved):

PR #13438, opened 2026-04-21
feat(but): apply theme to tui

slarse:
## 🧢 Changes
Apply theme to TUI

There are still a few details that need to be tweaked to make a light theme work for the TUI, I'll follow up with that.

### DIFF
--- a/crates/but/src/command/legacy/status/tui/tests/utils.rs
+++ b/crates/but/src/command/legacy/status/tui/tests/utils.rs
@@ -17,6 +17,7 @@ use crate::{
         build_status_output,
         tui::{App, EventPolling, Message, render_loop_once},
     },
+    theme,
     tui::TerminalGuard,
     utils::OutputChannel,
 };
@@ -217,7 +218,10 @@ impl TestTuiInputThenRenderResult<'_> {
         let backend = self.0.terminal.backend();
         let buffer = backend.buffer();
         let area = *buffer.area();
-        let selected_bg = *super::super::CURSOR_BG;
+        let selected_bg = theme::get()
+            .selection_highlight
+            .bg
+            .expect("background must be set on selection_highlight");
 
         (area.y..area.y.saturating_add(area.height))
             .find(|&y| {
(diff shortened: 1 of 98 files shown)


crates/but/src/command/legacy/status/tui/tests/utils.rs
```diff
@@ -506,8 +510,9 @@ fn is_blue_bold_hex_cell(cell: &ratatui::buffer::Cell) -> bool {
         && is_single_ascii_hex(cell.symbol())
 }
 
-fn is_blue_hex_cell(cell: &ratatui::buffer::Cell) -> bool {
-    matches!(cell.fg, Color::Blue) && is_single_ascii_hex(cell.symbol())
+fn is_commit_id_hex_cell(cell: &ratatui::buffer::Cell) -> bool {
+    let commit_id_fg = crate::theme::get().commit_id.fg;
+    commit_id_fg.is_some_and(|color| cell.fg == color) && is_single_ascii_hex(cell.symbol())
```
slarse, line 515: @davidpdrsn Had a test case failing due to unstable commit IDs, and it took me a while to figure out that it was because of this function failing to detect them. This function needed to change as commit IDs are now `Cyan` by default, rather than `Blue`.

 I think this strategy is also utilized by the to the `is_blue_bold_hex_cell()` to (probably?) identify CLI IDs. I think that one should similarly use the theme's `cli_id` style, but as it's not necessary to alter that right now for tests to align I'll leave it for the moment. But let's connect on this when you're back.

slarse (commented):
@davidpdrsn FYI this PR applies the theme to the vast majority of the TUI. I still need to apply a few more tweaks to be able to add a light theme, but it's close now.

Sorry I managed to time these changes with you being unavailable.

crates/but/src/command/legacy/status/tui/tests/utils.rs
davidpdrsn: Yes I'm not surprised this broke, its quite brittle currently. But yeah lets talk about it!

slarse: Crashed and burned on Windows as well :(

PR #13445, opened 2026-04-21
Disable rename detection when merging stack trees

mtsgrd:
## Problem

When rebuilding the workspace commit, `merged_tree_from_commits` merges stack tip trees pairwise using a 3-way merge. gix's rename detection matches deleted files to newly added files based on content similarity — and this produces **false-positive renames across independent stacks**, silently corrupting the workspace tree.

### Example scenario

Two stacks share a base with `file-a.txt` and `file-b.txt`:

- **Stack 1** modifies `file-b.txt`
- **Stack 2** deletes both files and adds `file-combined.txt` (content similar to `file-b.txt`)

When merging these stack trees, gix sees `file-b.txt → file-combined.txt` as a rename (high content similarity). This has two effects:

1. The real **delete-vs-modify conflict** on `file-b.txt` is hidden — gix treats it as a rename+content merge instead
2. `file-a.txt`'s clean deletion is **silently dropped** from the result tree

**Without the fix**, the workspace tree contains:
```
├── file-a.txt          ← should have been deleted by stack 2
├── file-combined.txt
└── other-file.txt
```

**With the fix**, the delete-vs-modify conflict on `file-b.txt` is correctly surfaced as `FailedToMergeBases`, preventing the silent data loss.

## Fix

Introduce a `TreeMergeMode` enum (`WithRenames` / `WithoutRenames`) threaded through `cherry_pick` and `merged_tree_from_commits`:

- **`TreeMergeMode::WithoutRenames`** — disables rename detection via `merge_options_no_rewrites_fail_fast`. Used for workspace commits whose parents are independent stack heads.
- **`TreeMergeMode::WithRenames`** — keeps rename detection enabled, maintaining parity with git for regular merge commits.

### Why is this safe?

`merged_tree_from_commits` is only called when merging **multiple parents** of a commit. For single-parent (linear) commits, the merge loop never executes and the options are never used.

The actual cherry-pick 3-way merge (base → onto → target) uses `merge_options_force_ours`, which is a completely separate code path unaffected by this change. Rename detection remains fully active there.

Since stacks are independent by definition, a rename detected *between* two stack trees is always a false positive. Disabling it here is correct.

### Other changes

- Rename `tree_from_merging_commits` → `merged_tree_from_commits` for clarity
- Clarify `conflictable` field docs to specify it's checked during rebase, not merge

## Test plan

- Added a regression test (`workspace_merge_surfaces_delete_vs_modify_conflict`) with a fixture that reproduces the exact scenario
- The test verifies the conflict is surfaced rather than silently producing a wrong tree
- All 102 existing tests continue to pass

### DIFF
--- a/crates/but-rebase/src/graph_rebase/cherry_pick.rs
+++ b/crates/but-rebase/src/graph_rebase/cherry_pick.rs
@@ -35,6 +35,22 @@ pub enum CherryPickOutcome {
     },
 }
 
+/// Controls how parent trees are merged during cherry-pick.
+///
+/// When cherry-picking a commit with multiple parents, both the old parents
+/// (base) and the new parents (ontos) are merged pairwise. This enum controls
+/// the merge options used for that pairwise merging.
+#[derive(Debug, Clone, Copy, PartialEq, Default)]
+pub enum TreeMergeMode {
+    /// Merge with rename detection enabled (default).
+    #[default]
+    WithRenames,
+    /// Merge with rename detection disabled.
+    /// Useful when parents are independent and renames detected across them
+    /// would be false positives.
+    WithoutRenames,
+}
+
 /// Controls when a commit is cherry-picked during a rebase.
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum PickMode {
@@ -65,12 +81,14 @@ pub enum PickMode {
 /// X's "base" sub-tree as the base.
 ///
 /// `pick_mode` - controls how to determine if a commit should be cherry-picked.
+/// `tree_merge_mode` - controls how parent trees are merged for merge commits.
 /// `sign_commit` - controls how the resulting commit is signed.
 pub fn cherry_pick(
     repo: &gix::Repository,
     target: gix::ObjectId,
     ontos: &[gix::ObjectId],
     pick_mode: PickMode,
+    tree_merge_mode: TreeMergeMode,
     sign_commit: SignCommit,
 ) -> Result<CherryPickOutcome> {
     let target = but_core::Commit::from_id(target.attach(repo))?;
@@ -80,11 +98,11 @@ pub fn cherry_pick(
         return Ok(CherryPickOutcome::Identity(target.id.detach()));
     }
 
-    let base_t = find_base_tree(&target)?;
+    let base_t = find_base_tree(&target, tree_merge_mode)?;
     // We always want the "theirs-ist" side of the target if it's conflicted.
     let target_t = find_real_tree(&target, TreeKind::Theirs)?;
     // We want to cherry-pick onto the merge result.
-    let onto_t = tree_from_merging_commits(repo, ontos, TreeKind::AutoResolution)?;
+    let onto_t = merged_tree_from_commits(repo, ontos, tree_merge_mode, TreeKind::AutoResolution)?;
 
     match (&base_t, &onto_t) {
         (MergeOutcome::NoCommit, MergeOutcome::NoCommit) if pick_mode == PickMode::Force => {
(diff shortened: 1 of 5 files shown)


mtsgrd:
@Byron would you be able to help me validate this?

crates/but-rebase/src/graph_rebase/cherry_pick.rs
```diff
@@ -231,7 +231,10 @@ fn tree_from_merging_commits(
 
         let commit = but_core::Commit::from_id(commit.attach(repo))?;
         let tree = find_real_tree(&commit, preference)?;
-        let (options, conflicts) = repo.merge_options_fail_fast()?;
+        // Rename detection must be disabled here: stack tips are independent, so
+        // renames detected across them are always false positives. With detection
+        // enabled, gix can silently swallow clean file deletions.
+        let (options, conflicts) = repo.merge_options_no_rewrites_fail_fast()?;
```
Byron: This makes sense to me!

I might also add that assuming the rename is not a false positive, then `merge` can get smart with how it treats these files and it's definitely not what we want for a megamerge.
The worst here is that the rename detection is not as elaborate as the one in Git, merely 'good enough', but even if it was great it's probably still not what we want here.

The question might still be if a cherry-pick would benefit from this, and also what Git does on cherry-pick. Does it run rename-detection to get better picks, maybe? If they also turn it off, it's a clear answer though.

mtsgrd: Git does seem to use rename detection for cherry-pick, so we now only disable rename detection for workspace commits with `TreeMergeMode::Workspace` passed through cherry_pick. Regular merge commits within stacks use `TreeMergeMode::Standard` with rename detection enabled. Let me know what you think?

Byron: That sounds better, let's do what Git does, and workaround the workspace-commit cherry-pick issue like this for now.

If you have the time, you might want to setup a testcase where Git cherry-picks such a merge-commit (can it even do that?), and if it's different/better, it's a bug in `gix`.

mtsgrd: I'd say it's working as intended?

> I tested what Git does in this scenario. Git's cherry-pick -m 1 works correctly here because it operates fundamentally differently from our approach: it extracts a diff relative to the selected parent and applies it as a 3-way merge. In that diff, file-a.txt is untouched, so it's never affected by rename detection.

> Our merged_tree_from_commits works differently — it merges the full trees of the parent commits pairwise (not diffs). This is specific to how we construct workspace commits from multiple stack heads. In that full-tree merge, gix's rename detection matches file-b → file-combined and silently absorbs the deletion of file-a.txt.

> So this isn't a gix bug — rename detection is working as designed for tree merges. The issue is that rename detection doesn't make sense for our workspace merges, where stacks are independent and a "rename" across stacks is coincidental. Disabling it only for TreeMergeMode::Workspace keeps standard cherry-pick/rebase behavior unchanged.

mtsgrd:
@Byron if you're happy with these changes can we go ahead and merge?

crates/but-rebase/tests/rebase/graph_rebase/snapshots/rebase__graph_rebase__cherry_pick__workspace_merge_surfaces_delete_vs_modify_conflict.snap
```diff
@@ -0,0 +1,13 @@
...
+---
+* 8ef051f (stack-2-after) stack-2-after: combine files
+| *   978e614 (HEAD -> workspace-before) GitButler Workspace Commit
+| |\  
+| | * b8f64ac (stack-2-before) stack-2-before: unrelated change
+| |/  
+|/|   
+| * f02613a (stack-1) stack-1: modify file-b
+|/  
+* 57993f6 (main, base) base
```
Caleb-T-Owens: Can we use inline snapshots rather than separate snap files?

crates/but-rebase/src/graph_rebase/cherry_pick.rs
```diff
@@ -35,6 +35,22 @@ pub enum CherryPickOutcome {
...
+/// the merge options used for that pairwise merging.
+#[derive(Debug, Clone, Copy, PartialEq, Default)]
+pub enum TreeMergeMode {
+    /// Standard merge with rename detection enabled.
+    #[default]
+    Standard,
+    /// Workspace merge where parents are independent stack heads.
+    /// Rename detection is disabled because renames across independent
+    /// stacks are always false positives.
+    Workspace,
```
Caleb-T-Owens: Can we change these names to be something along the lines of `WithRenames` and `WithoutRenames`.

I really want to avoid encoding ideas of a special "workspace" commit into the rebase engine.

crates/but-rebase/tests/rebase/graph_rebase/snapshots/rebase__graph_rebase__cherry_pick__workspace_merge_surfaces_delete_vs_modify_conflict.snap
mtsgrd: done!

crates/but-rebase/src/graph_rebase/cherry_pick.rs
mtsgrd: done!

Byron (approved):
LGTM!*

*The PR mostly affects `but-rebase` so my approval can only be for the changes to the way merges are computed, i.e. to skip rename tracking for what looks like workspace commits.

PR #13484, opened 2026-04-22
Drop but-testing in favor of using but

krlvi:
Reducing some maintenance burden & legacy code usage


Byron (approved):
Good bye `but-testing graph`, hello `but-debug graph`, one day :), as a clean-room implementation of the same thing.

PR #13488, opened 2026-04-23
feat: add but-debug CLI for GitButler internals

krlvi:


Byron (approved):
Thanks for helping me getting `but-debug graph` back 🙏.

Approving knowing this will soon follow the file structure of `but`.

PR #13496, opened 2026-04-23
ci: remove duplicated `but` test run and unify `but-server` tests

slarse:
## 🧢 Changes
Removes some test redundancy and unnecessary test separation.

This speeds up the test execution by ~30%, cutting down running times from ~13-14 to ~9-10 minutes.

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -282,11 +282,9 @@ jobs:
           save-if: ${{ github.ref == 'refs/heads/master' }}
       - name: Dependencies for 'keyring'
         run: sudo ./scripts/install-minimal-debian-dependencies.sh
-      - run: cargo test --workspace --exclude gitbutler-tauri --exclude but-server --exclude but-api-macros-tests
+      - run: cargo test --workspace --exclude gitbutler-tauri --exclude but-api-macros-tests
       # It's intentional to use 'name equals run-script' so it's easy to re-run locally on failure.
-      - run: cargo test -p but
       - run: make test-modern-but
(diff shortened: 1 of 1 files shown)


.github/workflows/push.yaml
```diff
@@ -282,11 +282,9 @@ jobs:
           save-if: ${{ github.ref == 'refs/heads/master' }}
       - name: Dependencies for 'keyring'
         run: sudo ./scripts/install-minimal-debian-dependencies.sh
-      - run: cargo test --workspace --exclude gitbutler-tauri --exclude but-server --exclude but-api-macros-tests
+      - run: cargo test --workspace --exclude gitbutler-tauri --exclude but-api-macros-tests
       # It's intentional to use 'name equals run-script' so it's easy to re-run locally on failure.
-      - run: cargo test -p but
       - run: make test-modern-but
```
slarse, line 287: As the `but` crate is not excluded on line 285, line 287 effectively means we're testing `but` twice. So let's cut that, which saves us a couple of minutes of CI time.

This _used_ to run with `--no-default-features`, but that's now encompassed in `make test-modern-but`. So I think that's where the duplication was added.

.github/workflows/push.yaml
```diff
@@ -282,11 +282,9 @@ jobs:
...
-      - run: cargo test -p but-server
```
slarse, line 289: I so no good reason to separate out these tests to their own `run` step. This will just lead to slower workflow execution. I've removed the exclusion on line 285 so they run there instead.

krlvi (approved):

PR #13502, opened 2026-04-23
Cache forge user profiles for offline support

mtsgrd:
## Summary

Cache forge user profiles so the app can display user info (avatar, name, email) without requiring an active network connection.

- Cached profiles are stored in a top-level `cachedProfiles` map in `forge_settings.json`, keyed by `access_token_key` — separate from account credentials so the cache can evolve independently
- On successful API calls, the cache is updated (writes are skipped when data is unchanged to avoid unnecessary disk I/O)
- On network errors, the cached profile is returned so the app works offline
- On auth errors (401/403), the cache is cleared so stale data isn't shown for revoked tokens
- Fixes a pre-existing bug where `GitHubAccount::username()` / `GitLabAccount::username()` incorrectly returned the `host` field for Enterprise/SelfHosted variants

## Frontend changes (uncommitted)

Adds graceful offline handling to the PR card UI:
- `ReduxResult` gets an `errorOverrides` prop for custom error rendering by error type
- `PullRequestCard` shows "{forgeName} not available" on network errors instead of a generic error
- `ghQuery.ts` maps network errors to `code: "NetworkError"`

### DIFF
--- a/crates/but-forge-storage/src/settings.rs
+++ b/crates/but-forge-storage/src/settings.rs
@@ -1,4 +1,18 @@
+use std::collections::HashMap;
+
 use serde::{Deserialize, Deserializer, Serialize, de::DeserializeOwned};
+
+/// Cached user profile data fetched from the forge API.
+///
+/// Stored separately from account credentials so the cache can evolve
+/// independently and be easily swapped for a different storage backend.
+#[derive(Clone, Debug, Serialize, Deserialize, Default, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct CachedProfile {
+    pub avatar_url: Option<String>,
+    pub name: Option<String>,
+    pub email: Option<String>,
+}
 
 #[derive(Clone, Debug, Serialize, Deserialize, Default)]
 #[serde(rename_all = "camelCase")]
@@ -8,6 +22,9 @@ pub struct ForgeSettings {
     /// GitLab-specific settings.
     #[serde(default)]
     pub gitlab: GitLabSettings,
+    /// Cached user profiles, keyed by account `access_token_key`.
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub cached_profiles: HashMap<String, CachedProfile>,
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, Default)]
@@ -18,7 +35,7 @@ pub struct GitHubSettings {
     pub known_accounts: Vec<GitHubAccount>,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(tag = "type")]
 pub enum GitHubAccount {
     OAuth {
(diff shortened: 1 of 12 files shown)


mtsgrd:
@estib-vega is this a direction we'd want to go?

crates/but-forge-storage/src/settings.rs
```diff
@@ -65,41 +86,65 @@ impl GitHubAccount {
             GitHubAccount::Enterprise { host, .. } => host,
         }
     }
+
```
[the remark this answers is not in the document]
estib-vega: This is a good suggestion. Related to my comment in https://github.com/gitbutlerapp/gitbutler/pull/13541#issuecomment-4325442887

estib-vega (commented):
Adding one comment about a fly-by fix that would be great to have.

estib-vega:
### Some thoughts for posterity
I really like the intended result of separating (unable to fetch the data for a profile) from being authenticated or not.

I'm not sure whether storing the cache within the known accounts is the best approach.
I'd like to take a bit of time thinking about this issue:

### Issues with the forge settings file
The forge_settings are persisted to disk as a JSON file, which in it self is not great since that opens the door to people manually editing it. This shouldn't be the case, since this is not intended for users to change.

### Drawbacks of this caching approach
I don't love the fact that this change caches the profile information within the account entries. Ideally the cache would be an independent (also, non file-based) storage system that could be easily changed. Storing this information in the file will make it hard to update if needed.

### In an ideal world
- Would be nice to store known_accounts in a global DB (out of this scope)
- Would be nice to store the cache in a DB (separately from the known accounts, so that we can change what we store without changing the known accounts)

I don't think we have a global DB, yet, though.

crates/but-forge-storage/src/settings.rs
estib-vega, line 28: I don't want to add another global JSON file with information we persist, but I would prefer if we wouldn't store the cache within the known account entries.

@mtsgrd What do you think about instead of adding the cache in the Account itself, we add a top-level map with the cached information directly in the settings, so that we:
1. Unify the cache in one place that can be easily swapped if needed.
2. Can be modified to store different information in the future without having to do type-migration jumps later

mtsgrd:
@estib-vega All of that makes sense. Since you frame it as thoughts for posterity, are you in favour of merging these changes _now_ with the intent of a better solution in the future?

crates/but-forge-storage/src/settings.rs
estib-vega: > @estib-vega All of that makes sense. Since you frame it as thoughts for posterity, are you in favour of merging these changes now with the intent of a better solution in the future?
>

Before merging I think maybe we could do this change?

estib-vega: So to end up with something like this

```rs
pub struct ForgeSettings {
    /// GitHub-specific settings.
    pub github: GitHubSettings,
    /// GitLab-specific settings.
    #[serde(default)]
    pub gitlab: GitLabSettings,
    /// Cache
    #[serde(default)]
    pub cache: Vec<CacheEntry> // <-- this is where the cache would live instead
}
```

estib-vega: Or a map
```rs
pub struct ForgeSettings {
    /// GitHub-specific settings.
    pub github: GitHubSettings,
    /// GitLab-specific settings.
    #[serde(default)]
    pub gitlab: GitLabSettings,
    /// Cache
    #[serde(default)]
    pub cache: HashMap<String, CacheEntry> // <-- this is where the cache would live instead
}
```

estib-vega: Thanks!

estib-vega (approved):

crates/but-forge-storage/src/settings.rs
mtsgrd: No thank _you_! :)

PR #13504, opened 2026-04-23
Allow selectors to be put in hash sets

Caleb-T-Owens:


Byron (approved):
Would be so nice if "your" crates wouldn't need approval if you make an edit.

PR #13505, opened 2026-04-23
fix(but): prevent `but discard` from failing on most multi-hunk files

slarse:
## 🧢 Changes
Fixes `but discard` failing on (most) multi-hunk files by flattening DiffSpecs before discarding.

The gist of it is that every hunk was (well, _is still_) naively added into a DiffSpec of its own. So a file with 2 hunks gets 2 DiffSpecs. We then applied these diffspecs in order, one at a time, independently. We resolve the hunks from the worktree again between each application so a preceeding DiffSpec with a hunk that displaces lines causes any succeeding DiffSpec to not line up anymore, triggering the bug.

This is trivial to reproduce: prepend a line to any moderately sized file, and then append a line to the same file s.t. the edits end up in different hunks. Discarding the first hunk displaces the rest of the lines of the file, so discarding the second hunk fails.

Here's a reproduction of the error: https://github.com/gitbutlerapp/gitbutler/actions/runs/24852478421/job/72756448821?pr=13505#step:5:2080

The fix I'm proposing here is to flatten the DiffSpecs on the level of the `but-workspace` crate so this kind of mistake cannot be reproduced in other places.

There's still an issue with DiffSpecs requiring certain ordering to be successful (see comments on commits and in code), so I've also updated the flattening to preserve order. It's still the responsibility of the caller to pass in specs in the correct order, though, which is not optimal as it's easy to get it wrong.

## Potential TODO
`but discard` is still passing in one `DiffSpec` per hunk, which is not very efficient. It would be better to just pass empty DiffSpecs for the file(s) to be restored, if the target is a file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1757,6 +1757,7 @@ dependencies = [
  "gitbutler-repo",
  "gitbutler-stack",
  "gix",
+ "indexmap 2.13.0",
  "insta",
  "itertools",
  "md5",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -264,6 +264,7 @@ rust-embed = { version = "8", features = ["axum"] }
 mime_guess = "2"
 chrono = { version = "0.4.42", default-features = false, features = ["std"] }
 hex = "0.4.3"
+indexmap = "2"
 md5 = "0.8.0"
 sha2 = "0.10"
 base64 = "0.22"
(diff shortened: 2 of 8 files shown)


Caleb-T-Owens (approved):
Lovely stuff

estib-vega (approved):

PR #13508, opened 2026-04-24
Replace test use of `gitbutler-cli` with `but`

krlvi:


slarse (approved):
I think the title should reference `gitbutler-cli` rather than `gitbutler-git`, but otherwise lgtm!

krlvi:
> I think the title should reference `gitbutler-cli` rather than `gitbutler-git`, but otherwise lgtm!

thanks! thats was a slip

Byron:
While using `but` instead was probably easiest and fastest, it might just be as easy to use `but-testsupport` and write a couple of bash scripts along with a vb.toml file that is then placed in the right spot. I've done the same previously, and it worked well (as in "clankers can do it").

krlvi:
> While using `but` instead was probably easiest and fastest, it might just be as easy to use `but-testsupport` and write a couple of bash scripts along with a vb.toml file that is then placed in the right spot. I've done the same previously, and it worked well (as in "clankers can do it").

Sounds good and we should probably do that. I mainly wanted to reduce the legacy surface area as much as possible now. 

I think in general using `but` in test setup may be annoying and undesirable because as we iterate on the `but` CLI we may change subcommands or even get rid of some (eg but setup) - and doing so would make unrelated e2e tests fail...

PR #13519, opened 2026-04-24
feat: migrate garbage collection to RefMetadata

krlvi:


crates/but-meta/src/legacy/mod.rs
```diff
@@ -511,6 +511,44 @@ impl VirtualBranchesTomlMetadata {
...
+            {
+                to_remove.push(stack.id);
+            }
+        }
+
+        if !to_remove.is_empty() {
+            for stack_id in to_remove {
+                self.data_mut().branches.remove(&stack_id);
+            }
+            self.snapshot.set_changed_to_necessitate_write();
```
Caleb-T-Owens:
```suggestion
            {
                self.data_mut().branches.remove(&stack_id);
                self.snapshot.set_changed_to_necessitate_write();
            }   
```

Caleb-T-Owens (approved):
Left a small simplification your LLM can make.

Part of me is wondering as a future point: What is the point of keeping branch metadata for branches that are not in your workspace at all.

crates/but-meta/src/legacy/mod.rs
Caleb-T-Owens: you can get rid of the extra `to_remove` vec by just removing the branches directly from the data structure

krlvi:
@Byron this was implemented on `VirtualBranchesTomlMetadata` - it's kinda not clear to me if this is a compatibility only type - if these are not meant to be used the way I used it here... it's probably a signal that our APIs are leaky and we should make some parts more private.

Byron:
@krlvi If all it does is to remove stacks and segments that aren't in the repository anymore, then no, I definitely would not have put it onto this compatibility type that is going to go away.

The workflow should have been like this in principle.

```rust
let meta = ctx.meta();
let ws = meta.workspace("refs/heads/gitbutler/workspace")?;
ws.gc(&repo)?; // <- This is `but_core::ref_metadata::Workspace`, dumb data that we can alter as needed.
meta.set_workspace(ws)?;
```

Please note that all this I write without having looked at the code. But maybe this can lead to a follow-up to clean this up to not alter `VirtualBranchesTomlMetadata` unnecessarily.

PR #13522, opened 2026-04-24
chore: remove dummy `gitbutler-git-setsid` binary

slarse:
The functionality of the setsid binary was removed quite a while back, but we had to keep a dummy binary around as the then-version of the macOS `but-installer` would verify the existence of said binary. So, old versions of the CLI would fail to upgrade to new versions.

Now it's been sufficiently long that I think we can get rid of this relic. For the very few people that may still be on a super-duper old version of the installer, they can repair the install by downloading the standalone `but-installer` (e.g. using the `install.sh` shell script).

### DIFF
--- a/crates/gitbutler-git/Cargo.toml
+++ b/crates/gitbutler-git/Cargo.toml
@@ -14,11 +14,6 @@ name = "gitbutler-git-askpass"
 path = "src/bin/askpass.rs"
 test = false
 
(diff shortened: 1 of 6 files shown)


slarse:
@Byron I know you were wondering when we were getting rid of this dummy binary - I'd say just about now :)

PR #13529, opened 2026-04-24
Reduce usage of VirtualBranchesHandle

krlvi:
This removes the get_default_target and set_default_target


crates/but-core/src/lib.rs
```diff
@@ -117,14 +117,17 @@ pub use ext::ObjectStorageExt;
 mod repo_ext;
 pub use repo_ext::{RepositoryExt, update_head_reference};
 
+/// The legacy single-workspace reference used by GitButler.
+pub const WORKSPACE_REF_NAME: &str = "refs/heads/gitbutler/workspace";
```
Byron, line 121: Just as a note: the system isn't hardcoded to this and it's entirely OK to have multiple workspace branches in the future. So while we have this now, eventually this will go away in favor of a workspace enumeration (which `ctx.meta()` supports).

crates/but-workspace/src/legacy/mod.rs
```diff
@@ -39,7 +42,11 @@ pub mod stack_ext;
 /// Returns the last-seen fork-point that the workspace has with the target branch with which it wants to integrate.
 // TODO: at some point this should be optional, integration branch doesn't have to be defined.
 pub fn common_merge_base_with_target_branch(gb_dir: &Path) -> anyhow::Result<gix::ObjectId> {
-    Ok(VirtualBranchesHandle::new(gb_dir).get_default_target()?.sha)
+    let meta = VirtualBranchesTomlMetadata::from_path(gb_dir.join("virtual_branches.toml"))?;
```
Byron: Everything below I wrote before realising that this function is unused :).

----

[`Workspace::merge_base_with_target_branch(commit_id)` ](https://github.com/gitbutlerapp/gitbutler/blob/8612afd674164943ba41d69c2975e4fdad740b28/crates/but-graph/src/projection/workspace/api.rs#L162) probably wants to be used whereever this is called.

Meantime, it would be better to pass `ctx` in here so you don't have to rely on the file (it's going away), su `ctx.legacy_meta()` instead.

crates/gitbutler-branch-actions/src/base.rs
```diff
@@ -466,8 +472,13 @@ fn remote_url_of_target_to_base_branch(ctx: &Context, target: &Target) -> Result
     Ok(remote_url)
 }
 
-fn default_target(base_path: &Path) -> Result<Target> {
-    VirtualBranchesHandle::new(base_path).get_default_target()
+fn default_target(ctx: &Context) -> Result<Target> {
+    ctx.legacy_meta()?
+        .data()
+        .default_target
```
Byron, line 478: All these `data().detault_target` related functions probably want to be helpers on `VirtualBranchesTomlRefMetadata` in `but-meta`.

Byron (commented):
This is great, thanks so much for your help!

I made some changes in a separate commit, please do take a look, there is a bunch of comments.

Big things to think about, also for me:

* `Workspace::metadata` adds ambiguity as Code now has to decide which data source to use. Is it `Workspace::target_commit_id` or `Workspace::metadata::target_commit_id`? There should be no ambiguity, and I think `metadata` might want to be private.

### Detailed review

Looked at those in detail.

* [x] but-core
* [x] but-graph
* [x] but-workspace
* [x] look more closely at places where workspace projection is used (workspace_and_db_with_perm)

Everything else I just glossed over, which should be fine for `gitbutler-` prefixed crates in particular. Nothing stuck out.

krlvi:
I agree about 

> There should be no ambiguity, and I think metadata might want to be private

crates/but-graph/Cargo.toml
```diff
@@ -22,6 +27,7 @@ tracing.workspace = true
 itertools.workspace = true
 
 [dev-dependencies]
+but-graph = { path = "." , features = ["legacy"] }
```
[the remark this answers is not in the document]
Byron: Not true

crates/gitbutler-branch-actions/src/stack.rs
```diff
@@ -155,8 +156,43 @@ pub fn push_stack(
         .context("Requires an open workspace mode")?;
     let virtual_branches = ctx.virtual_branches();
     let stack = virtual_branches.get_stack(stack_id)?;
-    let default_target = virtual_branches.get_default_target()?;
-    let push_env = PushStackEnv::new(ctx, &stack, default_target, skip_force_push_protection)?;
+    let (target_ref_name, target_base_oid, target_push_remote_name, target_branch_name) = {
+        let (repo, workspace, _db) = ctx.workspace_and_db_with_perm(guard.read_permission())?;
```
Byron: It's a bit hard for me to understand how this can expand into so many lines even though it's supposed to be related to `push_stack()`, but I suppose it's Ok and … lets just fully remove this crate soon.

crates/gitbutler-edit-mode/src/lib.rs
```diff
@@ -273,7 +273,17 @@ pub(crate) fn save_and_return_to_workspace(ctx: &Context, perm: &mut RepoExclusi
     let git2_repo = &*ctx.git2_repo.get()?;
     let repo = &*ctx.repo.get()?;
 
-    let old_workspace = WorkspaceState::create(ctx, perm.read_permission())?;
+    let old_workspace_projection = workspace_from_workspace_ref(ctx)?;
```
Byron, line 283: I think this function is more of a leftover from older times and should rather have been removed when there was the chance. And while it's okay to have it here, I now officially deprecated it to prevent further spread, even in a gitbutler-prefixed crate.

Byron (approved):
I see that a common need has been to get the target ref name and the stored target base commit id at once, which previously was the `Target`. My guess is that it's fine to have verbose code for this now as I don't think we will need to know these value quite as much. And if we do in `but-` crates, there will be helpers or abstractions.

A note regarding `target_base_commit_id()`: I hope that code won't have to use it directly as this is already included in the `ws.lower_bound` of the workspace, basically the forkpoint of all stacks.

`gitbutler-oplog` seems like there are future plans, so I left it as is for now. It would definitely be big to get rid of the VBH there, probably just by reading the metadata directly. To store and safe the data, one can use `ctx.legacy_meta()`, with the benefit if it automatically working once the vb.toml file is gone.

`last_pushed_base` is interesting, I seem to have no recollection of it and don't think it's used anywhere else.

My recommendation for the next fixup target is `gitbutler-oplog`. There will probably be some work on or with it to improve `but undo` as well.

PS: I also just skimmed over the tests, if you can call it that even.

##### Notes to self

One big problem with these old crates is that they calculate merge-bases themselevs. If they have to live longer, they should be tuned to use the `ws.graph` instead. The lower bound is already in `ws.lower_bound`, and new merge-bases can be computed with `Graph::find_git_merge_base()`. But changing the code to use it is probably harder than rewriting it, or adjusting our current code to use it more, if needed at all given `but-rebase`.

PR #13532, opened 2026-04-25
Fix CI publish failure introduced by embedded web frontend change

mtsgrd:
The previous commit built the web frontend inside
`tauri-before-build-command.sh` using `pnpm --filter`, which skips
turborepo's workspace dependency builds and fails to resolve
`@gitbutler/core`. Move the web frontend build into the `build-sveltekit`
CI job instead, where turborepo handles dependency ordering, and pass the
output as a separate artifact to `build-tauri`.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -63,6 +63,14 @@ jobs:
         run: pnpm build:desktop -- --mode ${{ env.vitemode }}
         env:
           SENTRY_RELEASE: ${{ env.version }}
+      - name: Build embedded web frontend
+        if: env.channel != 'release'
+        run: pnpm build:desktop
+        env:
+          SVELTEKIT_OUT_DIR: embedded-frontend
+          VITE_BUILD_TARGET: web
+          VITE_BUTLER_API_BASE_URL: /api
+          VITE_EMBEDDED_BUILD: "1"
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         name: Upload pnpm-store contents
         with:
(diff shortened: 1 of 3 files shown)


krlvi (approved):

Byron:
It looks like `master` is broken right now and with it probably every PR. Codex identified this PR as the fix, so I hope we can merge it before Monday.

---

Confirmed: `master` publish CI is genuinely broken.

Root cause: the release script now runs `pnpm --filter @gitbutler/desktop build` inside `crates/gitbutler-tauri/tauri-before-build-command.sh` for Windows/Linux nightly builds. That bypasses Turbo’s `package` dependency step, so fresh CI runners do not build `packages/core/dist`. Vite then fails resolving `@gitbutler/core/context`.

Failing jobs in run `24945373895`:
- `build-tauri (ubuntu-22.04)`
- `build-tauri (ubuntu-22.04-arm)`
- `build-tauri (windows-latest)`

All show the same class of error: Rollup cannot resolve `@gitbutler/core/context` during `@gitbutler/desktop` build.

There is already a focused fix open: PR #13532, `fix-ci-publish`: https://github.com/gitbutlerapp/gitbutler/pull/13532

That PR moves the embedded web frontend build into `build-sveltekit`, uploads it as an artifact, downloads it in `build-tauri`, and makes the release script skip that frontend build in CI. A manual `Publish` run on that fix branch succeeded across the Tauri build jobs: https://github.com/gitbutlerapp/gitbutler/actions/runs/24930536924

I did not change local files. Recommended action: mark PR #13532 ready and merge/cherry-pick it to `master`; the latest force-push only appears to adjust `turbo.json` formatting/pass-through config, while the prior semantic fix already passed `Publish`.

PR #13537, opened 2026-04-26
Fix "Worktree changes would be overwritten by checkout" on partial commits

mtsgrd:
## The bug

Committing a subset of lines from a single hunk fails with:

> Worktree changes would be overwritten by checkout

### Repro 1: adjacent additions

Start with a committed file:

```
line1
line2
line3
```

Add two adjacent lines in the worktree:

```diff
 line1
+added-a
+added-b
 line2
 line3
```

Commit only the first addition (`added-a`). The three-way merge to
transfer the remaining worktree change becomes:

```
  base (original):    line1 / line2 / line3
  ours (committed):   line1 / added-a / line2 / line3
  theirs (worktree):  line1 / added-a / added-b / line2 / line3
```

Both `ours` and `theirs` insert after `line1` — conflict.

### Repro 2: deletion + insertion

Start with:

```
line1
old-line
line3
```

Replace `old-line` with `new-line` in the worktree:

```diff
 line1
-old-line
+new-line
 line3
```

Commit only the deletion. The three-way merge becomes:

```
  base (original):    line1 / old-line / line3
  ours (committed):   line1 / line3
  theirs (worktree):  line1 / new-line / line3
```

Both sides modify `old-line` — conflict.

## The fix

`commit_create` no longer performs a checkout at all. The worktree
already contains the right content — committed changes came *from* the
worktree, and creating a commit cannot cause descendant conflicts, so
the workspace tree can't diverge from what's on disk.

Instead we use `materialize_without_checkout` to update refs, then
`sync_index_to_tree` to update `.git/index` to match the new workspace
tree. This is the same pattern used by the legacy commit engine and by
edit mode. The logic lives in a new
`WorkspaceState::from_successful_rebase_without_checkout` helper.

### Why not amend too?

Even though amend also takes changes from the worktree, amending a
mid-stack commit can cause conflicts in descendant commits. Those
conflicts change the workspace tree in ways the worktree doesn't
reflect, so amend still needs a full checkout.

### Why not SkipConflicting?

An earlier iteration added a `SkipConflicting` checkout mode that
dropped conflicting files from the checkout path list. While it worked,
skipping individual conflicts is a code smell — if we're comfortable
saying "don't update these files", we should be comfortable saying that
for *all* files, which is what `materialize_without_checkout` achieves.

### Side-effect: index module cleanup

`sync_index_to_tree` (renamed from `apply_lhs_to_rhs`) was moved from
the legacy module into `commit_engine::index`, making it available
without the `legacy` feature flag. The legacy caller now imports from
the new location.

## Test plan

- `partial_commit_with_adjacent_lines_conflicts_on_checkout` — repro 1
- `partial_commit_with_deletion_plus_insertion_conflicts_on_checkout` — repro 2
- Both demonstrate that `safe_checkout` conflicts on partial commits,
  which is why `commit_create` uses `materialize_without_checkout` instead
- All 13 checkout tests, 40 legacy commit engine tests, and 247
  but-workspace tests pass


mtsgrd:
@Caleb-T-Owens is this something you could help me out with?

Caleb-T-Owens (approved):

PR #13541, opened 2026-04-27
Fix hosted github/gitlab username accessor

mtsgrd:


mtsgrd:
@estib-vega just want to make sure this wasn't intentionally returning host when accessing username?

estib-vega:
Very nice catch to a dumb mistype on my side. Thanks!

I see I made this twice as well. The user for github also returns the host instead of the user for GH Enterprise

estib-vega (approved):

PR #13543, opened 2026-04-27
feat(lite): add in-app auto-updater

slarse:
## 🧢 Changes
Adds an auto-updater to Lite. This is _literally_ auto-updating. On startup, the updater checks for a new version and downloads it if it finds one. When downloaded, the user is forced to restart the app to install the update.

To opt out of this behavior, one must start the app with `LITE_NO_AUTOUPDATE=1` set.

For an end-user release, we probably want update behavior configurable. I did it this way now as I wanted to add the bare minimum of UI necessary to not introduce a bunch of slop into the product.

### DIFF
--- a/apps/lite/electron/src/updater.ts
+++ b/apps/lite/electron/src/updater.ts
@@ -1,0 +1,41 @@
+import { app, BrowserWindow, ipcMain } from "electron";
+import electronUpdater, { type AppUpdater, type UpdateDownloadedEvent } from "electron-updater";
+import { liteIpcChannels } from "./ipc.js";
+import { env } from "node:process";
+
+let updaterWindow: BrowserWindow | null = null;
+let updaterRegistered = false;
+
+const getAutoUpdater = (): AppUpdater => {
+	const { autoUpdater } = electronUpdater;
+	return autoUpdater;
+};
+
+const sendUpdateDownloaded = (event: UpdateDownloadedEvent): void => {
+	if (!updaterWindow || updaterWindow.isDestroyed()) return;
+	updaterWindow.webContents.send(liteIpcChannels.updaterUpdateDownloaded, event);
+};
+
+export const registerUpdater = (mainWindow: BrowserWindow): void => {
+	updaterWindow = mainWindow;
+	if (updaterRegistered) return;
+	updaterRegistered = true;
+
+	const autoUpdater = getAutoUpdater();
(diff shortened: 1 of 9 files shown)


krlvi (approved):

apps/lite/electron/src/updater.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export const registerUpdater = (mainWindow: BrowserWindow): void => {
+	updaterWindow = mainWindow;
+	if (updaterRegistered) return;
+	updaterRegistered = true;
+
+	const autoUpdater = getAutoUpdater();
+	autoUpdater.autoDownload = true;
+	autoUpdater.autoInstallOnAppQuit = true;
```
slarse, line 26: This is supposed to make the install proceed regardless of how the app is exited, but it seemed to not function on macOS. Therefore, for now, the update dialog cannot be dismissed.

apps/lite/electron/src/updater.ts
```diff
@@ -0,0 +1,41 @@
+import { app, BrowserWindow, ipcMain } from "electron";
+import electronUpdater, { type AppUpdater, type UpdateDownloadedEvent } from "electron-updater";
+import { liteIpcChannels } from "./ipc.js";
+import { env } from "node:process";
+
+let updaterWindow: BrowserWindow | null = null;
+let updaterRegistered = false;
```
slarse, line 7: This seems like a really dirty way of doing things, but it also works, so...

apps/lite/electron/src/updater.ts
```diff
@@ -0,0 +1,41 @@
...
+	if (updaterRegistered) return;
+	updaterRegistered = true;
+
+	const autoUpdater = getAutoUpdater();
+	autoUpdater.autoDownload = true;
+	autoUpdater.autoInstallOnAppQuit = true;
+	autoUpdater.on("update-downloaded", (event) => sendUpdateDownloaded(event));
+	autoUpdater.on("error", (error) => {
+		// oxlint-disable-next-line no-console
+		console.error("Update error", error);
```
slarse, line 30: Forgive me for my sins, but I did not want to have to bother with injecting a toast in here. IMO a semi-silent failure of the auto-updater is the right way to go about it, at least for the moment.

slarse:
@OliverJAsh @samhh FYI, here's the updater!

PR #13545, opened 2026-04-27
Fix new commit appearing in all branches when added to bottom of stack

mtsgrd:
`InsertSide::Above` was redirecting all incoming edges to the newly
inserted node. When the target is a shared commit (e.g. the merge base),
this pulled Reference edges from other stacks, causing their refs to
point to the new commit and making it appear in every branch.

Now, when the target has both Reference and non-Reference (Pick) incoming
edges, only the Pick edges are redirected. Reference edges are preserved
so other stacks' refs stay where they are. When all incoming edges are
References (e.g. inserting above a branch tip), they are still redirected
so the new commit becomes part of that stack.


mtsgrd:
Closing because we want a front end fix for this instead.

PR #13546, opened 2026-04-27
Use merge-base override to prevent consumed changes reappearing as uncommitted

mtsgrd:
## Problem

When committing or amending worktree changes, committed hunks could reappear as uncommitted modifications — duplicating lines on disk. This happened when:
- Amending changes into an earlier commit in a stack
- Creating a partial commit (selecting specific hunks)

## Cause

`safe_checkout` preserves uncommitted worktree state across rebases using a 3-way merge:

```
base   = old HEAD tree
ours   = new HEAD tree (after rebase)
theirs = snapshot of worktree before the operation
```

Consumed changes appear in both **ours** (baked in by the rebase) and **theirs** (captured in the snapshot), so the merge treats them as independent additions and duplicates them.

## Fix

Build a **committed tree** (old HEAD + consumed DiffSpecs) and use it as the merge base:

```
base   = old HEAD tree + consumed changes
ours   = new HEAD tree (after rebase)
theirs = snapshot of worktree before the operation
```

Consumed changes now appear in both **base** and **theirs**, producing no diff on the theirs side — so they're not cherry-picked onto the destination. Only genuinely remaining uncommitted changes survive.

The committed tree is built additive-only (stripping `previous_path` for renames) to match the snapshot's additive-only semantics. Without this, a rename's source-side deletion would appear only in the base, causing the merge to treat the source file in the snapshot as a new addition and restore it on disk.

This works at the tree level for any granularity (full files, full hunks, sub-hunk selections) without hunk-level arithmetic.

Both `commit_create` and `commit_amend` now share this mechanism, replacing the previous `materialize_without_checkout` workaround for commit creation.

### DIFF
--- a/crates/but-core/src/snapshot/resolve_tree.rs
+++ b/crates/but-core/src/snapshot/resolve_tree.rs
@@ -33,6 +33,13 @@ pub struct Options {
     ///
     /// If `None`, perform the merge just like Git, which will include conflict markers!
     pub worktree_cherry_pick: Option<gix::merge::tree::Options>,
+    /// If set, use this tree as the merge base instead of the `HEAD` entry from the snapshot.
+    ///
+    /// This is used when an operation consumed some worktree changes (e.g. amend/commit).
+    /// By setting the merge base to `HEAD^{tree}` + consumed changes, the consumed hunks
+    /// are already present in the base and won't appear as diffs in the snapshot side of
+    /// the merge — eliminating duplication without any hunk-level subtraction.
+    pub merge_base_override: Option<gix::ObjectId>,
 }
 
 pub(super) mod function {
@@ -60,6 +67,7 @@ pub(super) mod function {
         target_worktree_tree_id: gix::ObjectId,
         Options {
             worktree_cherry_pick: worktree_cherry_pick_options,
+            merge_base_override,
         }: Options,
     ) -> anyhow::Result<Outcome<'_>> {
         let repo = snapshot_tree.repo;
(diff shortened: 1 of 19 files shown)


mtsgrd:
@Caleb-T-Owens I've reworked this PR to solve both the commit and amend issue by altering the merge base slightly. Would you mind having another look?

Byron:
Can we wait with this one until https://github.com/gitbutlerapp/gitbutler/pull/13536 lands?
They feel related. I plan to work on it next.

crates/but-core/src/worktree/checkout/utils.rs
```diff
@@ -18,56 +18,14 @@ use crate::{
     worktree::checkout::{Outcome, UncommitedWorktreeChanges},
 };
 
-/// Preserve any uncommitted worktree changes that would be affected by a checkout from one tree to another.
```
Byron, line 21: I liked these docs, they helped me 😭.

crates/but-core/src/snapshot/resolve_tree.rs
Byron, line 82: I think I understand what this does and I agree with it wholeheartedly.
When I do a full review, I'd just change names and improve docs, as part of making it 'me-safe', i.e. harder to misunderstand.

Byron (commented):
I think I will have a go at it soon, as I think the `but-core` part is a needed addition independently of anything else.

This is just my first look and some thoughts, all of which I'd fixup myself then, so no action needed.

crates/but-core/src/snapshot/resolve_tree.rs
mtsgrd: Awesome, thank you.

krlvi (approved):

PR #13554, opened 2026-04-28
improve but-graph performance

Byron:
* [x] avoid HashMap with u32 as key, BTreeMap may be better?
    - As SegmentIds are incremental, using `vec![0; max(graph[segment_id])]` would be optimal and is totally feasible


Byron:
Hi @estib-vega , just wanted to hand this over to you (pushed to GB so you can push into it). You seem to have better local reproducers.

estib-vega:
Using this other PR instead: https://github.com/gitbutlerapp/gitbutler/pull/13566

PR #13568, opened 2026-04-28
Fix unapply introducing uncommitted changes with integrated commits

mtsgrd:
## Summary

When you unapply a stack, files that have nothing to do with that stack can appear as uncommitted changes in your workspace. For example:

1. You have two stacks: **Stack A** and **Stack B**
2. Stack A was branched from an older version of `master` that contained a file (`ghost.txt`) which has since been deleted upstream
3. You unapply Stack B (which never touched `ghost.txt`)
4. **Bug:** `ghost.txt` suddenly appears as an untracked file in your workspace

This happens because the old unapply code used an ancestor commit (`workspace_base`) as the merge target. That ancestor still contained files that were deleted in the current target, so the 3-way merge re-introduced them.

## Fix

Replace both the merge base and "theirs" in the unapply 3-way merge:

- **base**: workspace commit tree (HEAD) instead of the unapplied stack's head tree
- **theirs**: remerged tree (target + remaining stacks) instead of the old `workspace_base` ancestor

This way, files with no uncommitted edits satisfy `base == ours`, so `theirs` (the correct new state) wins cleanly. Files with uncommitted edits have `base ≠ ours` and are preserved via `FileFavor::Ours`.

## Test plan

- [x] Added regression test that sets up the exact ghost-file scenario and asserts no spurious uncommitted changes after unapply
- [x] Verified test **fails** against the old code (`ghost.txt: Status(WT_NEW)`)
- [x] Verified test **passes** with the fix

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3940,7 +3940,6 @@ dependencies = [
  "git2",
  "git2-hooks",
  "gitbutler-branch",
- "gitbutler-cherry-pick",
  "gitbutler-commit",
  "gitbutler-git",
  "gitbutler-operating-modes",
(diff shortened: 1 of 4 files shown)


Byron (commented):
Since this is in `gitbutler-branch-actions` and a test was added for validation, I think this needs no further review or delay.
Ship it.

PR #13582, opened 2026-04-30
Improve Playwright e2e test infrastructure

mtsgrd:
## Summary

- **Remove redundant helpers**: Delete `expectVisible` and `sleep` from `util.ts` — `expectVisible` duplicated the config-level `expect.timeout`, and `sleep` calls were either unnecessary (Playwright assertions already auto-retry) or replaced with `page.waitForTimeout()`
- **Bump assertion timeout to 15s**: Raise `expect.timeout` from 10s to 15s in `playwright.config.ts` so no test needs inline timeout overrides
- **Flatten CI artifacts & capture console logs**: Rewrite the test fixture to write console logs (`.log`) and videos (`.webm`) as flat files named after the test, instead of nested per-test folders with `trace.zip`

## Test plan

- [ ] CI e2e tests pass
- [ ] On failure, the `playwright-report` artifact contains flat `.log` and `.webm` files named after the failing tests

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -173,9 +173,5 @@ jobs:
         if: failure()
         with:
           name: playwright-report
-          path: |
-            ./e2e/playwright-report/
-            ./e2e/test-results/**/trace.zip
-            ./e2e/test-results/**/video.webm
-            ./e2e/test-results/**/error-context.md
+          path: ./e2e/playwright/test-results-flat/
           retention-days: 30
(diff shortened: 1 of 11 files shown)


Caleb-T-Owens (approved):
Curious about the output format change, but rest looks good

mtsgrd:
If you're talking about the flat folder, I hope it will be great. Problem is, this pr hasn't had a failure with the playwright tests so I haven't seen it materialized yet. 😏

PR #13600, opened 2026-05-01
Fix move-branch failing when base segment has no ref name

mtsgrd:
## Problem

Moving a branch onto another stack fails with:

```
"Graph base segment doesn't have a ref name"
```

This happens when `origin/master` advances past the fork point where branches diverge — the old fork point becomes an unnamed segment in the graph, and the disconnect logic in `get_disconnect_parameters` assumed every base segment would carry a ref.

## Changes

- Extract a `select_segment` helper that tries the ref name first, then falls back to `segment.tip()`.
- Introduce a `SegmentLike` trait to unify `StackSegment` and `but_graph::Segment` behind a common interface.
- Add a crate-level test (`move_branch_when_base_segment_has_no_ref_name`) that directly exercises the fallback path with an unnamed base segment.
- Add an E2E test that reproduces the full scenario: apply two branches, advance `origin/master`, fetch, then move one branch onto the other.

### DIFF
--- a/crates/but-workspace/src/branch/move_branch.rs
+++ b/crates/but-workspace/src/branch/move_branch.rs
@@ -375,20 +375,10 @@ fn get_disconnect_parameters<'ws, 'meta, M: RefMetadata>(
 
     let parents_to_disconnect = if let Some(stack_base_segment) = stack_base_segment {
         // Base segment is part of the source stack.
-        let base_segment_ref_name = stack_base_segment
-            .ref_name()
-            .context("Base segment doesn't have a ref name.")?;
-        let reference_selector = editor.select_reference(base_segment_ref_name)?;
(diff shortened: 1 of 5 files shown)


estib-vega:
This is a great fix! I naively assumed even if the origin/master would advance, there would still be a named segment there, which of course doesn't make sense. 

I agree with the copilot fixes. Ideally we don't depend on the e2e tests for this: Would be able to add tests at the but-workspace level too?

I'm not sure sure if there are already fixtures that accommodate testing this scenario, but it shouldn't be too hard to add.

Thanks for catching this!

mtsgrd:
@estib-vega added a rust test, could you have a look?

estib-vega (approved):
Thanks for the fix!

Test looks good to me

PR #13615, opened 2026-05-04
fix: Start ssh agent for nix

estib-vega:
SSH agent is started inside nix, if needed

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -52,6 +52,15 @@
           # We use different versions of Playwright in different packages... consider also
           # voidus/nix-playwright-browsers.
           export PLAYWRIGHT_BROWSERS_PATH=${pkgs.playwright-driver.browsers}
+
+          # --- FIX: forward SSH agent ---
+          if [ -z "$SSH_AUTH_SOCK" ]; then
+            echo "Starting ssh-agent inside nix shell..."
+            eval "$(ssh-agent -s)"
+            ssh-add ~/.ssh/id_ed25519 2>/dev/null
+          else
+            echo "Using existing SSH agent: $SSH_AUTH_SOCK"
+          fi
         '';
       };
     });


estib-vega:
This is still a POC, need to test this

flake.nix
```diff
@@ -52,6 +52,15 @@
...
           # voidus/nix-playwright-browsers.
           export PLAYWRIGHT_BROWSERS_PATH=${pkgs.playwright-driver.browsers}
+
+          # --- FIX: forward SSH agent ---
+          if [ -z "$SSH_AUTH_SOCK" ]; then
+            echo "Starting ssh-agent inside nix shell..."
+            eval "$(ssh-agent -s)"
+            ssh-add ~/.ssh/id_ed25519 2>/dev/null
+          else
+            echo "Using existing SSH agent: $SSH_AUTH_SOCK"
```
OliverJAsh, line 62: I don't think this will change anything. For example testing on my machine I'm seeing this line echo'd, meaning it's a no-op.

estib-vega: Yes, also realized this probably doesn't fix the underlying issue.
@slarse came up with the hypothesis of it probably being a permissions issue. We can sync tomorrow

OliverJAsh: What do you see when you apply this patch and try to commit for the first time after starting the process? I see this:

> `[GB-1236] SSH commit signing attempt: SSH_AUTH_SOCK=None`

Presumably you see something different?

If your `SSH_AUTH_SOCK` is present then it's a question of how.

```diff
Commit ID: 68bf2b07ad46b3f7447c692e4a9ce75e190d59dd
Change ID: yxvuotrpwxxryyyzorvzsyxzrmupxupu
Author   : Oliver Joseph Ash <oliverjash@gmail.com> (2026-05-04 18:25:31)
Committer: Oliver Joseph Ash <oliverjash@gmail.com> (2026-05-04 18:46:33)

    (no description set)

diff --git a/crates/but-core/src/commit/mod.rs b/crates/but-core/src/commit/mod.rs
index 10c73dbba3..753d138ebb 100644
--- a/crates/but-core/src/commit/mod.rs
+++ b/crates/but-core/src/commit/mod.rs
@@ -323,6 +323,11 @@
     let is_ssh = sign_format.is_some_and(|value| value.as_ref() == "ssh");
 
     if is_ssh {
+        println!(
+            "[GB-1236] SSH commit signing attempt: SSH_AUTH_SOCK={:?}",
+            std::env::var_os("SSH_AUTH_SOCK")
+        );
+
         let mut signature_storage = tempfile::NamedTempFile::new()?;
         signature_storage.write_all(buffer)?;
         let buffer_file_to_sign_path = signature_storage.into_temp_path();
```

slarse: @OliverJAsh Gimme a ping when you're on and let's dive into this, I have some ideas.

estib-vega: Yep, I see it

```
[GB-1236] SSH commit signing attempt: SSH_AUTH_SOCK=Some("/private/tmp/com.apple.launchd.fGXjRGjBoy/Listeners")
```

PR #13627, opened 2026-05-05
Prune integrated commits at or below the target from all stacks

mtsgrd:
## Summary

Rewrites integrated-commit pruning so that **all stacks** are pruned equally, but only removes integrated commits at or below the workspace's `target_commit_id`.

### What changed

- **Removed the `is_metadata_tracked` skip** — metadata-tracked stacks are no longer exempt from pruning
- **Rewrote `truncate_at_fork_point`** — determines "at or below target" via graph reachability (walking outward from the target segment to collect a `HashSet` of commit IDs), not index comparisons. When a segment is fully emptied by pruning, it is removed entirely.
- **Fixed target-in-segment assumption** — the old code assumed the target commit was always the first commit in its segment; the new code finds the target's actual position and only includes commits from that point downward
- **Metadata-tracked empty branches are preserved** — a branch just added at the fork point with no commits yet is kept, but branches emptied by pruning are removed
- **Removed `is_metadata_tracked` and `find_commit_in_stack` helpers** — no longer needed
- **Added regression test** — `integrated_commits_above_target_are_kept` verifies both that integrated commits above the target survive, and that they are pruned once the target advances

### Why

Two problems with the old behavior:

1. **Inconsistent** — auto-discovered stacks got pruned but metadata-tracked stacks did not, leaving fully-integrated commits visible in the workspace
2. **Too aggressive** — pruning based solely on the `Integrated` flag could remove branches whose work was merged upstream before the user ran integrate, preventing `integrate_upstream` from detecting and processing them

### How it works

The pruning boundary is the workspace's `target_commit_id` — the SHA stored in metadata that stays fixed until the user integrates upstream.

To determine which commits are "at or below" the target, we:

1. Walk the graph outward (`Direction::Outgoing`) from the target segment, collecting all reachable commit IDs into a `HashSet`
2. Slice the target segment itself from the target commit's actual position downward — commits above the target in the same segment are not included

Then for each stack, the contiguous integrated tail whose commits are in the set is removed:

```
commit C        ← kept (above target, even if integrated)
commit B        ← kept (above target, even if integrated)
─── target ───
commit A        ← pruned (integrated + in HashSet)
base            ← pruned (integrated + in HashSet)
```

When a branch is merged upstream but the user hasn't integrated yet, the target hasn't moved — so the branch and its commits stay visible. Once the user integrates and the target advances, they are pruned automatically.

Fully-integrated stacks are cleaned up entirely. Empty branches that existed before pruning (e.g. a newly added branch at the fork point) are preserved if they appear in workspace metadata.

### Fallback

When `target_commit` metadata is not set (e.g. in tests or older workspaces), the target segment is resolved from `target_ref`. This keeps pruning working in fixtures that don't set `target_commit_id`.

### DIFF
--- a/crates/but-graph/src/projection/workspace/init.rs
+++ b/crates/but-graph/src/projection/workspace/init.rs
@@ -755,25 +755,47 @@ impl WorkspaceState {
         }
     }
 
-    /// Truncate each stack to only show commits above its fork point with the
-    /// target, then remove empty branches and stacks.
-    ///
-    /// The fork point is the first commit on the first-parent chain (walked via
-    /// the graph's `CommitFlags::Integrated`) that is reachable from the target.
-    ///
-    /// Stacks explicitly tracked in workspace metadata are never pruned —
-    /// their integrated branches and commits must remain visible so that
-    /// `integrate_upstream` can detect and remove them.
+    /// Remove integrated commits and empty branches at the bottom of each
+    /// stack, but only those at or below the workspace's target commit.
+    /// Integrated commits above the target are kept until the user advances
+    /// the target via upstream integration.
     fn prune_integrated_segments(&mut self, graph: &Graph) {
         if self.extra_target.is_some() || self.target_ref.is_none() {
             return;
         }
+        let (target_segment_index, target_commit_id) = if let Some(tc) = self.target_commit.as_ref()
+        {
+            (tc.segment_index, Some(tc.commit_id))
+        } else if let Some(tr) = self.target_ref.as_ref() {
+            (tr.segment_index, None)
+        } else {
+            return;
+        };
+
+        // Collect all commit IDs that are the target itself or ancestors
+        // of it by walking the graph toward its ancestors
+        // (Direction::Outgoing). In the target segment itself, only
+        // include commits from the target position downward — commits
+        // above it in the same segment are not reachable from the target.
+        let mut is_target_or_below = HashSet::new();
+        graph.visit_all_segments_excluding_start_until(
+            target_segment_index,
+            Direction::Outgoing,
+            |s| {
+                is_target_or_below.extend(s.commits.iter().map(|c| c.id));
+                false
+            },
+        );
+        // For the target segment, include commits from the target onward.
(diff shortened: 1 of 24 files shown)


mtsgrd:
@Caleb-T-Owens your pr #13632 makes the tests here green. 🙏

crates/but-graph/src/projection/workspace/init.rs
```diff
@@ -967,52 +968,83 @@ impl WorkspaceState {
...
-        for &(graph_sidx, ofs) in &seg.commits_by_segment {
-            for (i, commit) in graph[graph_sidx].commits.iter().enumerate() {
+        for &(graph_sidx, ofs) in seg.commits_by_segment.iter().rev() {
+            for (i, commit) in graph[graph_sidx].commits.iter().enumerate().rev() {
+                let offset = ofs + i;
+                let above_target = seg_idx < target_seg_idx
+                    || (seg_idx == target_seg_idx && offset < target_offset);
+                if above_target {
+                    break 'outer;
+                }
```
Caleb-T-Owens: We absolutely can **not** do comparisons on indices to determine this stuff.

We provide no guarantees in the order that the but graph is created. Even the order of edges can't be relied on.

I would encourage you to take an approach where you get a HashSet of all the commits beneath your `target_commit`, then loop over the segments, dropping commits or full segments if they're fully included in the HashSet

Caleb-T-Owens: If you just look at commits, empty refs will be ambiguous, but your traversal of the graph (`visit_all_segments_including_start_until`) will also let you  gather the relevant reference names too

mtsgrd: Thanks, I've pushed a new version!

mtsgrd:
@Caleb-T-Owens I've currently got your [branch](https://github.com/gitbutlerapp/gitbutler/tree/target-sha-manually-moved) included here in an ad-hoc way, and the result is decent imho. I've spent a bit of time looking at the changed snapshots, and they have all been accurate so far.

Caleb-T-Owens (approved):

PR #13630, opened 2026-05-05
git config: Treat empty strings as removal

estib-vega:
When passing empty strings to the git config, remove the entry
altogether.

---

This fixes some observed issues in which we’d store empty strings as config values for e.g. `gpg.program` leading to errors even outside of gitbutler.


estib-vega:
cc @slarse this should fix the issue you observed yesterday

krlvi (approved):

PR #13640, opened 2026-05-05
remove but-tools legacy implementations

krlvi:
This removes the:
- auto commit from the GUI and CLI (but absorb --new) (unfortunately due to being built on the legacy code paths)
- but mcp --internal


crates/but/skill/references/reference.md
```diff
@@ -219,7 +219,6 @@ but absorb <branch-id>        # Absorb all changes staged to this branch
 but absorb                    # Absorb ALL uncommitted changes (use with caution)
 but absorb --dry-run          # Preview without making changes
 but absorb <file-id> --dry-run  # Preview specific file absorption
-but absorb --new/-n           # Create new commits instead of amending existing ones
```
estib-vega, line 222: wop wop wop, this one we want to keep

krlvi: yeah but... it uses the legacy code path :/

estib-vega: Ahh, dang. Then I'll port it later to the non-legacy

apps/desktop/src/components/commit/AutoCommitModalContent.svelte
estib-vega, line 1: @krlvi Let's not delete this yet please. Even if we get rid of the auto-commit API for now, I don't want to lose this for when it's ported

krlvi: okay i will bring it back and keep it as orphan component

estib-vega: Thanks!

PR #13644, opened 2026-05-05
fix(lite): add `lite://` protocol to serve local content

slarse:
When opening a project in the packaged app, React Router sets the path to `/project/<id>/workspace`. As we used to serve via the `file://` protocol, this made the effective `location.href` `file:///project/<id>/workspace`. When reloading the page, Electron just requests whatever is set in the location. As that file does not exist, we fail the reload.

This commit fixes that by introducing a custom `lite://` protocol where we simply serve the index file for any route-looking path, and let the router handle the rest.

This also plugs a security hole inherent in serving local files via the `file://` protocol, see https://www.electronjs.org/docs/latest/tutorial/security#18-avoid-usage-of-the-file-protocol-and-prefer-usage-of-custom-protocols

### DIFF
--- a/apps/lite/electron/src/main.ts
+++ b/apps/lite/electron/src/main.ts
@@ -58,15 +58,73 @@ import {
 	updateBranchName,
 	BranchListingFilter,
 } from "@gitbutler/but-sdk";
-import { app, BrowserWindow, ipcMain, Menu, type MenuItemConstructorOptions } from "electron";
+import {
+	app,
+	BrowserWindow,
+	ipcMain,
+	Menu,
+	net,
+	protocol,
+	type MenuItemConstructorOptions,
+} from "electron";
 import { REACT_DEVELOPER_TOOLS, installExtension } from "electron-devtools-installer";
 import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
-import { fileURLToPath } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 
 const currentFilePath = fileURLToPath(import.meta.url);
 const currentDirPath = path.dirname(currentFilePath);
+
+const liteProtocolScheme = "lite";
+const liteProtocolHost = "app";
+const contentRootURL = pathToFileURL(path.join(currentDirPath, "../ui"));
+
+// Custom scheme to serve files. This is necessary for two reasons:
+//
+// 1. Security, as serving via file:// opens up a wider attack surface than is desirable (see https://www.electronjs.org/docs/latest/tutorial/security#18-avoid-usage-of-the-file-protocol-and-prefer-usage-of-custom-protocols)
+// 2. The ability to reload the page when we've set a route that does not correspond to a file we can actually serve
+protocol.registerSchemesAsPrivileged([
+	{
+		scheme: liteProtocolScheme,
+		privileges: {
+			standard: true,
+			secure: true,
+			supportFetchAPI: true,
+		},
+	},
+]);
+
+const registerLiteProtocolHandler = () => {
+	// Handler based on the examples in https://www.electronjs.org/docs/latest/api/protocol#protocolhandlescheme-handler
+	protocol.handle(liteProtocolScheme, async (req) => {
+		const { host, pathname } = new URL(req.url);
+
+		// Our bundle is served with a primary index.html and a flat assets directory, so there's
+		// no need for relative directory traversal to serve our content at this time. We can
+		// therefore trivially prevent path traversal by simply disallowing any ..
+		//
+		// Don't name files with any intermediate .. and we don't need to make this check account for that :)
+		//
+		// In addition, we only have the single host to serve from for now.
+		if (pathname.includes("..") || host !== liteProtocolHost)
+			return new Response("Not found", {
+				status: 404,
+				headers: { "content-type": "text/html" },
+			});
+
+		// We default to serving the index file unless the pathname indicates it's an asset. This is
+		// important to be compatible with React Router's "soft navigation" where it changes the
+		// location to track where you are in the app, but it's still an SPA with only an index file
+		// to actually serve from the backend. For example, if the user navigates somewhere and then
+		// reloads the page, we should still serve up the index file, and React Router will handle the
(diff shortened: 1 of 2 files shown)


apps/lite/electron/src/main.ts
```diff
@@ -295,7 +353,9 @@ const createMainWindow = async (): Promise<void> => {
 		return;
 	}
 
-	await mainWindow.loadFile(path.join(currentDirPath, "../ui/index.html"));
+	registerLiteProtocolHandler();
+	const rootUrl = `${liteProtocolScheme}://${liteProtocolHost}/`;
+	await mainWindow.loadURL(rootUrl);
```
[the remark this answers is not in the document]
slarse: Probably valid, will fix.

apps/lite/electron/src/main.ts
```diff
@@ -58,16 +58,74 @@ import {
...
+		// therefore trivially prevent path traversal by simply disallowing any ..
+		//
+		// Don't name files with any intermediate .. and we don't need to make this check account for that :)
+		//
+		// In addition, we only have the single host to serve from for now.
+		if (pathname === ".." || host !== liteProtocolHost)
+			return new Response("Not found", {
+				status: 404,
+				headers: { "content-type": "text/html" },
+			});
```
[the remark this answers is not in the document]
slarse: Hah, this is a good catch, I removed the wrong part of this check when I simplified it to not account for paths with intermediate `..`.

apps/lite/electron/src/main.ts
```diff
@@ -58,16 +58,74 @@ import {
...
+		// If the pathname does not have a file extension, we're dealing with the root or a route within
+		// the app. We then simply serve the index page and React Router reads the path and navigates
+		// to the correct place.
+		if (path.parse(pathname).ext === "") {
+			const pathToServe = path.join(contentRoot, "/index.html");
+			return net.fetch(pathToFileURL(pathToServe).toString());
+		}
+
+		const pathToServe = path.join(contentRoot, pathname);
+		return net.fetch(pathToFileURL(pathToServe).toString());
```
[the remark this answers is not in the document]
slarse: This is not how `path.join()` works, it just joins the arguments together and normalizes. Leading or trailing slashes have no effect.

apps/lite/electron/src/main.ts
```diff
@@ -58,16 +58,74 @@ import {
...
+		if (pathname.includes("..") || host !== liteProtocolHost)
+			return new Response("Not found", {
+				status: 404,
+				headers: { "content-type": "text/html" },
+			});
+
+		// If the pathname does not have a file extension, we're dealing with the root or a route within
+		// the app. We then simply serve the index page and React Router reads the path and navigates
+		// to the correct place.
+		if (path.parse(pathname).ext === "") {
```
slarse: This is a pretty fat assumption. If we create a route with a file-like name (i.e. has a `.` anywhere in the final part of the pathname), reloading that page will cause us to serve the index page on reload.

We could reverse this check and look specifically for `/`, `/index.html` and `/assets`. That might be preferable because it'll be more obvious if we immediately fail to serve an actual file, than if we fail to reload a specific route.

Any opinions?

apps/lite/ui/vite.config.ts
```diff
@@ -15,7 +15,7 @@ export default defineConfig({
 			},
 		}),
 	],
-	base: "./",
+	base: "/",
```
slarse, line 18: This change is necessary as, otherwise, when you reload e.g. on the route `/project/<id>/workspace`, assets are fetched relative to that (e.g. `/project/<id>/workspace/assets/blabla.css`).

I introduced this relative behavior in the first place as it was necessary when serving via `file://`, since the "ui root" does not correspond to the actual root of the app. But now with the `lite://` protocol we want to have fetches be absolute, and then we direct them to the correct subdirectory as necessary.

apps/lite/electron/src/main.ts
```diff
@@ -58,16 +58,74 @@ import {
...
+		const { host, pathname } = new URL(req.url);
+
+		// Our bundle is served with a primary index.html and a flat assets directory, so there's
+		// no need for relative directory traversal to serve our content at this time. We can
+		// therefore trivially prevent path traversal by simply disallowing any ..
+		//
+		// Don't name files with any intermediate .. and we don't need to make this check account for that :)
+		//
+		// In addition, we only have the single host to serve from for now.
+		if (pathname.includes("..") || host !== liteProtocolHost)
```
slarse, line 110: Note that this broadly disallows files or routes with `..` in them, only a small subset of which actually represent path traversals. I think this is fine for now, but I could see how we potentially want to add commit range paths in the future, and they typically contain `..`.

I don't want to code too much for the future here - we'll add support when we have the need.

OliverJAsh (approved):

apps/lite/electron/src/main.ts
```diff
@@ -58,16 +58,74 @@ import {
...
+		// If the pathname does not have a file extension, we're dealing with the root or a route within
+		// the app. We then simply serve the index page and React Router reads the path and navigates
+		// to the correct place.
+		if (path.parse(pathname).ext === "") {
+			const pathToServe = path.join(contentRoot, "/index.html");
+			return net.fetch(pathToFileURL(pathToServe).toString());
+		}
+
+		const pathToServe = path.join(contentRoot, pathname);
+		return net.fetch(pathToFileURL(pathToServe).toString());
```
OliverJAsh: As discussed:

```
const contentRootURL = pathToFileURL(contentRoot)
const pathToServe = new URL(pathname, contentRootURL)
return net.fetch(pathToServe.toString());
```

slarse: So, this turned out to not be what we want, and it's because the "base" of a URL only constitutes the protocol and the host, and so the URL is unexpectedly incorrect. For example:

```typescript
new URL("c", "file://a/b").href
```

Our desired output is `file://a/b/c`, but the actual output is `file://a/c`.

So we do want the properties of path.join and will just have to bother with decoding and encoding the pathname.

PR #13647, opened 2026-05-06
Ignore updates for unversioned builds

robertmonka:
## 🧢 Changes

- Treats missing, empty, or `0.0.0` compile-time `VERSION` values as unversioned development builds.
- Skips remote update checks when the build does not have a real release/nightly version.
- Ignores cached update results for unversioned builds instead of comparing them against a `0.0.0` fallback.
- Changes the default update status version from `0.0.0` to `dev`.
- Updates `but update check` human output to explain that update checks are only available for versioned release and nightly builds.
- Adds focused coverage for version-aware cached update handling and current-version request behavior.

## ☕️ Reasoning

Local and development builds can be compiled without a real `VERSION`. Previously those builds fell back to `0.0.0`, which made update checks compare a dev binary against published release data and could surface a misleading available update.

This makes unversioned builds opt out of release-channel update comparisons entirely, while keeping normal update behavior for packaged release and nightly builds that do include a real compile-time version.

## 🧪 Testing

- Added/updated Rust tests for cached update handling with and without a real build version.
- Not run locally while updating this PR description.

### DIFF
--- a/crates/but-update/src/cache.rs
+++ b/crates/but-update/src/cache.rs
@@ -51,6 +51,7 @@ impl AvailableUpdate {
 /// This function checks the cache for a previously performed update check and returns
 /// update information if:
 /// - A cached update check exists
+/// - The current build has a real compile-time `VERSION`
 /// - The cached status indicates an update is available (`up_to_date == false`)
 /// - The update is not currently suppressed
 /// - The available version differs from the current version (not a no-op)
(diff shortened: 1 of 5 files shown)


Byron:
I'm very sorry, but I have to close all of your PRs on the basis of them being fully automated coming from a profile that doesn't show much related history.

Thanks a lot for your understanding.

PR #13648, opened 2026-05-06
Launch WSL editors via CLI

robertmonka:
## 🧢 Changes

- Launches VS Code-like editor URLs through the editor CLI when running inside WSL.
- Supports `vscode`, `vscode-insiders`, `vscodium`, `cursor`, `windsurf`, and `trae` URL schemes.
- Converts `editor://file/...` URLs into CLI arguments, including `--goto` for line/column targets.
- Preserves `windowId=_blank` behavior by passing `--new-window`.
- Percent-decodes file paths before launching the editor command.
- Falls back to the existing URL opener when not running in WSL, when the editor scheme is unsupported, or when the editor command is not available.
- Keeps the AppImage environment cleanup shared between GUI URL launching and CLI editor launching.
- Adds Linux-only unit tests for URL-to-editor-command conversion.

## ☕️ Reasoning

WSL does not reliably handle desktop editor URL schemes through the normal GUI opener path. For editors that also expose a command-line launcher, invoking the CLI directly is more predictable and lets GitButler open the requested file, line, column, and project window without relying on Windows-side URL association behavior.

The fallback path is intentionally unchanged for non-WSL Linux environments and unsupported editors, so existing URL opening behavior remains in place outside this specific case.

## 🧪 Testing

- Added Linux-only unit coverage for Cursor, VS Code new-window URLs, percent-decoded paths, and unsupported editor fallback.
- Not run locally while updating this PR description.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -983,6 +983,7 @@ dependencies = [
  "tokio",
  "tracing",
  "url",
+ "urlencoding",
  "uuid",
  "which",
 ]
--- a/crates/but-api/Cargo.toml
+++ b/crates/but-api/Cargo.toml
@@ -138,6 +138,7 @@ serde-error = "0.1.3"
 gix = { workspace = true, features = ["worktree-mutation"] }
 open.workspace = true
 url = { version = "2.5", optional = true }
+urlencoding.workspace = true
 uuid.workspace = true
 which = "8.0.2"
 but-schemars.workspace = true
--- a/crates/but-api/src/legacy/open.rs
+++ b/crates/but-api/src/legacy/open.rs
@@ -6,6 +6,47 @@ use but_api_macros::but_api;
 use tracing::instrument;
 use url::Url;
 
+fn clean_env_vars<'a, 'b>(
+    var_names: &'a [&'b str],
+) -> impl Iterator<Item = (&'b str, String)> + 'a {
+    var_names
+        .iter()
+        .filter_map(|name| env::var(name).map(|value| (*name, value)).ok())
+        .map(|(name, value)| {
+            (
+                name,
+                value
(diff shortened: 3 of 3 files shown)


Byron:
I'm very sorry, but I have to close all of your PRs on the basis of them being fully automated coming from a profile that doesn't show much related history.

Thanks a lot for your understanding.

robertmonka:
Hi Byron, thanks for the explanation. I understand the concern.

These PRs were created with AI-agent assistance, and I should have made that clearer. For this one specifically, I verified the behavior in practice in a WSL setup, understand the functional intent of the change, and I’m responsible for following up on any issues or requested adjustments.

To reduce review burden, would you be open to reconsidering just this focused WSL bugfix first?

The intent here is narrow: make editor launching from WSL go through the CLI path so configured editors can open correctly from that environment.

I’m happy to resubmit it with a shorter, more conventional PR description, add or adjust tests if needed, and avoid opening more PRs until we agree on the expected contribution process.

Byron:
Thanks for checking back.

I looked at the code and find it unusable to the point where I wouldn't want to spend any time on it. But that's no problem, everything can be regenerated, right? Next time without refactorings, minimal changes and doc-strings.

To me, what matters is to have someone test a feature because we don't really use Windows or have even have access to a machine with WSL installed. So on top of the generated code that passes CI and is minimal, I'd need to be able to trust that the PR is tested and is working on an actual Windows WSL installation. For that checklists could be used or short report can be written of the things that were tried, how they didn't work before and how they failed and how they work now.

Also, it helps to disclose prompts, to gain trust and quickly evaluate if it makes sense to proceed with the review or not (after all, I can do the same but might have a 'better' prompt).

PR #13653, opened 2026-05-06
Fix backend forge detection for GitHub Enterprise users

mtsgrd:
The `forge_provider` backend command was missing from the Tauri ACL
permissions list (fixed in an earlier commit), which meant the query
silently failed for all desktop users since 0.19.4. Regular github.com
users were unaffected because the frontend had redundant domain-matching
logic that masked the failure. GitHub Enterprise Server users (e.g.
github.ourinternaldomain.net) had no such fallback and lost their forge
integration (PR button, reviews, etc.).

ACLs are going away soon (tauri-apps/tauri#15266), so rather than
papering over backend failures with frontend fallbacks, this commit:

- Removes the redundant frontend domain-matching in forgeFactory that
  was masking the broken backend query
- Simplifies forge type resolution to: manual override → backend
  detection → default
- Adds Rust tests confirming GHES subdomain detection works for both
  SSH and HTTPS URL forms

### DIFF
--- a/crates/but-forge/src/lib.rs
+++ b/crates/but-forge/src/lib.rs
@@ -148,7 +148,8 @@ pub fn get_all_forge_accounts() -> anyhow::Result<Vec<ForgeUser>> {
 #[cfg(test)]
 mod tests {
     use super::{
-        ForgeName, ForgeUser, match_host_to_accounts_custom_host, normalize_host_for_comparison,
+        ForgeName, ForgeUser, derive_forge_repo_info, match_host_to_accounts_custom_host,
+        normalize_host_for_comparison,
     };
 
     #[test]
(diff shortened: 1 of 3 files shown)


crates/but-forge/src/lib.rs
```diff
@@ -280,4 +281,23 @@ mod tests {
...
+
+    #[test]
+    fn derive_forge_detects_github_enterprise_subdomain_https() {
+        let info =
+            derive_forge_repo_info("https://github.ourinternaldomain.net/our-org/my-repo.git")
+                .expect("should detect GHES from github.* subdomain");
+        assert_eq!(info.forge, ForgeName::GitHub);
+        assert_eq!(info.owner, "our-org");
+        assert_eq!(info.repo, "my-repo");
+    }
```
estib-vega, line 302: The determination of the forge is now happening mainly in the backend, only to fallback to the frontend system if it can't catch it. So any fixes should mainly go to the rust side, if anything.

I see that we're adding just extra tests (which is nice) but does the functionality need to be fixed in that case as well?

estib-vega:
> The forge_provider backend command was missing from the Tauri ACL
permissions list (fixed in an earlier commit), which meant the query
silently failed for all desktop users since 0.19.4
>

Intersting what's the Tauri ACL?

PR #13655, opened 2026-05-06
Remove explicit ACL permissions file, fix misconfigured entries

mtsgrd:
## Summary

Removes the hand-maintained `permissions/default.toml` ACL file and its references, now that the Tauri bump in #13654 makes it unnecessary.

## Why

Every new `#[tauri::command]` had to be manually added to `permissions/default.toml`. Forgetting an entry caused the command to **silently fail** at runtime with no compile-time error. This has bitten us multiple times:
- #13076 — missing ACL for new custom commands
- #13384 — missing `forge_provider` permission
- #13653 — another missing permission fix

With Tauri 2.11.0, custom commands are allowed for local origins by default, so we no longer need the explicit manifest.

## Changes

- Deleted `crates/gitbutler-tauri/permissions/default.toml`
- Removed `"allow-custom-commands"` from `capabilities/main.json`
- Simplified `build.rs` to use `tauri_build::build()` instead of `AppManifest`

## Bugs found & fixed

- **Missing**: `commit_squash` was registered in `generate_handler!` but absent from the ACL
- **Stale**: `forge_branch_chat` and `forge_provider` were in the ACL but their commands had been removed

---
- <kbd>&nbsp;1&nbsp;</kbd> #13655 👈


slarse (commented):
It's great if we can get rid of the ACL.

> With Tauri 2.11.0, custom commands are allowed for local origins by default, so we no longer need the explicit manifest.

This comment has me a bit concerned that this is missing the point. The problem the ACL solved wasn't to allow local origins, it was to deny non-local ones. Counter-intuitive as that may seem. Can you validate that, with this patch, we still deny non-local origins from accessing the IPC bridge?

To be able to validate, you need to first remove the `on_navigate()` navigation block in `crates/gitbutler-tauri/src/window.rs`. After that you can navigate to a different origin by just setting `location.href = "whatever"` in the console, and execute an IPC command with e.g. `__TAURI_INTERNALS__.invoke("list_projects");`. It should be denied.

krlvi:
do we still need this PR?

PR #13660, opened 2026-05-06
remove unused legacy implementations in but-workspace

krlvi:


crates/but-workspace/src/legacy/mod.rs
```diff
@@ -25,9 +25,7 @@ pub use stacks::{
 };
 pub use tree_manipulation::{
     MoveChangesResult,
-    remove_changes_from_commit_in_stack::remove_changes_from_commit_in_stack,
     split_branch::{split_branch, split_into_dependent_branch},
```
estib-vega, line 28: Other APIs we should remove (even from the the FE) are the split branch ones. I don't expect them to be used too much and need to be ported to the new rebase engine as well

estib-vega (approved):

PR #13665, opened 2026-05-06
Detect and reject individual no-effect changes after cherry-pick

mtsgrd:
## Problem

When committing a mix of changes to a branch, some changes could be silently dropped — the commit succeeded, but the caller received no rejection feedback for the omitted changes.

**Example:** deleting a file that only exists in another stack's workspace tree. The deletion has no effect on this branch's tree, but was not reported as rejected.

## Root cause

The post-cherry-pick check only compared entire trees:

```
if tree_with_changes == target_tree { reject all }
```

When at least one change was effective the trees differed, so individual no-ops slipped through undetected.

## Fix

After the cherry-pick, check each change individually via an `entries_match` helper that compares both mode and object_id at a given path between the target and result trees:

```
for each change:
  if entries_match(target, result, change.path)
     && entries_match(target, result, change.previous_path):
    reject as NoEffectiveChanges
```

Comparing mode (not just object_id) catches mode-only changes like executable bit toggles. Checking `previous_path` catches renames where only the source removal was effective.

## Test

A new `two-stacks-with-extra-file` fixture verifies that committing a file deletion from another stack alongside a valid modification:
- rejects the deletion as `NoEffectiveChanges`
- still produces a commit for the modification

### DIFF
--- a/crates/but-core/src/tree/mod.rs
+++ b/crates/but-core/src/tree/mod.rs
@@ -153,6 +153,50 @@ pub fn create_tree(
                 }
                 tree_with_changes = merge_result.tree.write()?.detach();
             }
+            // After the cherry-pick, some changes may have had no effect on the target tree
+            // (e.g. deleting a file that only exists in another stack's tree, not in this
+            // branch). Check each change individually by comparing its path entry between
+            // the target tree and the result tree — if they match, the change was a no-op.
+            // We compare both object_id and mode to catch mode-only changes (e.g. executable bit).
+            // For renames, we also check that the previous_path was actually removed.
+            if needs_cherry_pick {
+                let target = target_tree.attach(repo).object()?.peel_to_tree()?;
+                let result = tree_with_changes.attach(repo).object()?.peel_to_tree()?;
+                fn entries_match(
+                    target: &gix::Tree<'_>,
+                    result: &gix::Tree<'_>,
+                    path: &[u8],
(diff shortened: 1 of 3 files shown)


crates/but-core/src/tree/mod.rs
```diff
@@ -153,6 +153,50 @@ pub fn create_tree(
                 }
                 tree_with_changes = merge_result.tree.write()?.detach();
             }
+            // After the cherry-pick, some changes may have had no effect on the target tree
+            // (e.g. deleting a file that only exists in another stack's tree, not in this
+            // branch). Check each change individually by comparing its path entry between
+            // the target tree and the result tree — if they match, the change was a no-op.
+            // We compare both object_id and mode to catch mode-only changes (e.g. executable bit).
+            // For renames, we also check that the previous_path was actually removed.
+            if needs_cherry_pick {
```
Byron, line 162: I think it makes sense to put this into the previous `if needs_cherry_pick` branch, instead of opening a new one.

Byron (approved):
Thanks a lot!

I took a close look and think this is sound, despite this only being an armchair review.
Please see my one comment that might warrant a change, but it's more of a nit.

As a general note, these lower-level internal functions tend to not be well documented on as a function (along their parameters), and I think it's always worth fixing this while one is at it.
These days, I treat the documentation as a big part of specifying intent, with the code being replaceable.

PR #13668, opened 2026-05-06
Remove target.sha update from reconcilliation step

mtsgrd:


Byron (approved):
Now that (soon) no old code runs anymore, this is the right move!
When this is fully true, all of this reconciliation can be removed as well.

To me the latter seems safest once the vb.toml data-structure has been fully migrated to `but_core::ref_metadata::Workspace` though.

Caleb-T-Owens (approved):

PR #13682, opened 2026-05-07
lite: add various Electron hardening fixes

slarse:
## 🧢 Changes
Adds a slew of security fixes from the [Electron security checklist](https://www.electronjs.org/docs/latest/tutorial/security#17-validate-the-sender-of-all-ipc-messages).

I've taken the stance of disabling pretty much everything we don't use. When we get to a point where we do want to use any of the disabled features, we'll intentionally enable them as we go.

Every commit in this PR is a standalone fix that could be merged independently of all of the others, but as everything is related to app startup I decided to keep it all in this one PR. See the individual commits for details.

### DIFF
--- a/apps/lite/electron/src/main.ts
+++ b/apps/lite/electron/src/main.ts
@@ -508,4 +508,10 @@ app.on("web-contents-created", (_, contents) => {
 		console.error(`Blocked opening new window for ${url}`);
 		return { action: "deny" };
 	});
+
+	contents.on("will-attach-webview", (event, webPreferences, _) => {
+		// oxlint-disable-next-line no-console
+		console.error(`Blocked attaching webview ${JSON.stringify(webPreferences)}`);
+		event.preventDefault();
+	});
 });
--- a/apps/lite/electron/src/main.ts
+++ b/apps/lite/electron/src/main.ts
@@ -456,3 +456,11 @@ app.on("window-all-closed", () => {
 	WatcherManager.getInstance().destroy();
 	if (process.platform !== "darwin") app.quit();
 });
+
+app.on("web-contents-created", (_, contents) => {
+	contents.on("will-navigate", (event, navigationUrl) => {
+		// oxlint-disable-next-line no-console
+		console.error(`Blocked navigation to ${navigationUrl}`);
+		event.preventDefault();
+	});
+});
--- a/apps/lite/electron/src/main.ts
+++ b/apps/lite/electron/src/main.ts
@@ -79,6 +79,30 @@ import { fileURLToPath, pathToFileURL } from "node:url";
 const currentFilePath = fileURLToPath(import.meta.url);
 const currentDirPath = path.dirname(currentFilePath);
 
+// Permissions in this array are allowed by default for trusted origins, without prompting the user for input.
+const trustedOriginDefaultPermissions: Array<
+	| "clipboard-read"
+	| "clipboard-sanitized-write"
+	| "display-capture"
+	| "fullscreen"
+	| "geolocation"
+	| "idle-detection"
+	| "media"
+	| "mediaKeySystem"
+	| "midi"
+	| "midiSysex"
+	| "notifications"
+	| "pointerLock"
+	| "keyboardLock"
+	| "openExternal"
+	| "speaker-selection"
+	| "storage-access"
+	| "top-level-storage-access"
+	| "window-management"
+	| "unknown"
+	| "fileSystem"
+> = ["clipboard-sanitized-write"] as const;
+
 const liteProtocolScheme = "lite";
 const liteProtocolHost = "app";
 const contentRootURL = pathToFileURL(path.join(currentDirPath, "../ui"));
@@ -478,6 +502,16 @@ void app.whenReady().then(async () => {
 			});
 		});
 	}
+
+	session.defaultSession.setPermissionRequestHandler((webContents, permission, callback) => {
+		const url = newUrlOrNull(webContents.getURL());
+		if (isTrustedLocalOrigin(url) && trustedOriginDefaultPermissions.includes(permission))
+			return callback(true);
+
+		// oxlint-disable-next-line no-console
+		console.error(`Blocked permission request for ${permission} from ${url?.href ?? "<unknown>"}`);
+		return callback(false);
+	});
 
 	registerIpcHandlers();
 	await createMainWindow();
--- a/apps/lite/electron/src/main.ts
+++ b/apps/lite/electron/src/main.ts
@@ -463,4 +463,10 @@ app.on("web-contents-created", (_, contents) => {
 		console.error(`Blocked navigation to ${navigationUrl}`);
 		event.preventDefault();
 	});
+
+	contents.setWindowOpenHandler(({ url }) => {
+		// oxlint-disable-next-line no-console
+		console.error(`Blocked opening new window for ${url}`);
+		return { action: "deny" };
+	});
 });
--- a/apps/lite/electron/src/main.ts
+++ b/apps/lite/electron/src/main.ts
@@ -177,72 +177,110 @@ const buildNativeMenuTemplate = (
 		};
 	});
 
+// Returns true if the `url` is from an origin we trust to perform privileged actions such as executing IPC commands.
+const isTrustedLocalOrigin = (url: URL | null) =>
+	url !== null &&
+	(app.isPackaged
+		? url.protocol === `${liteProtocolScheme}:` && url.host === liteProtocolHost
+		: url.protocol === "http:" && url.host === "127.0.0.1:5173");
+
+const newUrlOrNull = (url: string): URL | null => {
+	try {
+		return new URL(url);
+	} catch {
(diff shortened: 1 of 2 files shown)


apps/lite/electron/src/main.ts
```diff
@@ -382,3 +484,23 @@ app.on("window-all-closed", () => {
 	WatcherManager.getInstance().destroy();
 	if (process.platform !== "darwin") app.quit();
 });
+
+app.on("web-contents-created", (_, contents) => {
+	contents.on("will-navigate", (event, navigationUrl) => {
+		// oxlint-disable-next-line no-console
+		console.error(`Blocked navigation to ${navigationUrl}`);
+		event.preventDefault();
+	});
```
estib-vega, line 538: Probably you've tested this already, but does this work with TanStack router? Or do we need to filter out the trusted URLs here as well?

apps/lite/electron/src/main.ts
```diff
@@ -359,14 +391,84 @@ const createMainWindow = async (): Promise<void> => {
...
+
+	session.defaultSession.setPermissionRequestHandler((webContents, _, callback) => {
+		const url = new URL(webContents.getURL());
+		if (!isTrustedLocalOrigin(url)) {
+			// oxlint-disable-next-line no-console
+			console.error(`Blocked permission request for ${url.href}`);
+			return callback(false);
+		}
+
+		return callback(true);
```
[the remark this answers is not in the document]
estib-vega: This is also a good idea, but I don't know which permissions we'd need.

For now, maybe we can deny all permissions and see how far we get without them? Adding them as we need.

wdyt?

apps/lite/electron/src/main.ts
slarse: This works as is, the router doesn't perform a real navigation when it changes the location so this never triggers at present.

estib-vega: Gotcha 👍

apps/lite/electron/src/main.ts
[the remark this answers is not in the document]
slarse: I was considering doing just this. The current approach doesn't "undermine hardening" because allowing permissions only for local origins is a heckofalot better than allowing permissions for _any_ origin. I didn't go all the way because I thought it might be a hassle.

But maybe it isn't. Let's try it.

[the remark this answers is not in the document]
slarse: @estib-vega see revised commit [6d54f43](https://github.com/gitbutlerapp/gitbutler/pull/13682/commits/6d54f430d276a68fc79ab95eb19f5876524b5e9f)

estib-vega (approved):

apps/lite/electron/src/main.ts
```diff
@@ -76,6 +77,30 @@ import { fileURLToPath, pathToFileURL } from "node:url";
...
+	| "pointerLock"
+	| "keyboardLock"
+	| "openExternal"
+	| "speaker-selection"
+	| "storage-access"
+	| "top-level-storage-access"
+	| "window-management"
+	| "unknown"
+	| "fileSystem"
+> = [] as const;
```
estib-vega: I'm guessing `fileSystem` is Electron specific?
Clipboard for sure we'll need.

I guess we can trim this is needed 👍

slarse: I think this is easy to misread: That's just the type. I needed to add that for TypeScript not to complain about the `trustedOriginDefaultPermissions.includes(permission)` check. There is unfortunately no defined type in Electron that I could find, this union is defined with literals in function argument list.

If you look at the actual array, it's 100% empty :D

Note that this is also programmatic access, so e.g. not having `clipboard-read` doesn't mean that you as a user can't copy and paste in the app.

estib-vega: Aaaaaaaaahh duuuh haha.

Thanks.

> Note that this is also programmatic access, so e.g. not having clipboard-read doesn't mean that you as a user can't copy and paste in the app.
>
Yep yep, was just saying that we'll need `clipboard-write` for some features like copying the branch name or the PR link and such.

slarse: > Yep yep, was just saying that we'll need clipboard-write for some features like copying the branch name or the PR link and such.

Gotcha, I'll just add that in immediately.

slarse: I'm guessing `clipboard-sanitized-write` is the one? I don't think there is a `clipboard-write`.

estib-vega: Sounds *write* ;)

apps/lite/electron/src/main.ts
```diff
@@ -389,3 +513,23 @@ app.on("window-all-closed", () => {
 	WatcherManager.getInstance().destroy();
 	if (process.platform !== "darwin") app.quit();
 });
+
+app.on("web-contents-created", (_, contents) => {
+	contents.on("will-navigate", (event, navigationUrl) => {
+		// oxlint-disable-next-line no-console
+		console.error(`Blocked navigation to ${navigationUrl}`);
+		event.preventDefault();
+	});
```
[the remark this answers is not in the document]
slarse: That's not how this works.

apps/lite/electron/src/main.ts
```diff
@@ -366,14 +421,83 @@ const createMainWindow = async (): Promise<void> => {
...
 	}
+
+	session.defaultSession.setPermissionRequestHandler((webContents, permission, callback) => {
+		const url = new URL(webContents.getURL());
+		if (isTrustedLocalOrigin(url) && trustedOriginDefaultPermissions.includes(permission))
+			return callback(true);
+
+		// oxlint-disable-next-line no-console
+		console.error(`Blocked permission request for ${permission} from ${url.href}`);
+		return callback(false);
```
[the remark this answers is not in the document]
slarse: Fair point, will fix.

apps/lite/electron/src/main.ts
```diff
@@ -176,72 +201,101 @@ const buildNativeMenuTemplate = (
...
+		const senderValidatingListener: typeof listener = (event, ...args) => {
+			// Validate that the frame is from a trusted origin. This is crucial to prevent unauthorized
+			// access to the IPC bridge if we ever render non-local content.
+			//
+			// See https://www.electronjs.org/docs/latest/tutorial/security#17-validate-the-sender-of-all-ipc-messages
+			const isSenderFrameTrusted =
+				event.senderFrame !== null && isTrustedLocalOrigin(new URL(event.senderFrame.url));
+			if (isSenderFrameTrusted)
+				// eslint-disable-next-line @typescript-eslint/no-unsafe-argument @typescript-eslint/no-unsafe-return
+				return listener(event, ...args);
```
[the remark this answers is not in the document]
slarse: Fair point, will fix.

apps/lite/electron/src/main.ts
```diff
@@ -368,13 +369,71 @@ const createMainWindow = async (): Promise<void> => {
...
-		const dockIcon = getMacDockIcon();
-		if (dockIcon !== undefined && app.dock) app.dock.setIcon(dockIcon);
+	if (app.isPackaged) {
+		registerLiteProtocolHandler();
+
+		// Basic non-Strict CSP based on https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html#basic-non-strict-csp-policy
+		const productionCsp =
+			"default-src 'none';" +
+			"script-src 'self' 'wasm-unsafe-eval';" +
+			"style-src 'self' 'unsafe-inline';" +
```
slarse: Explain why `unsafe-inline`

apps/lite/electron/src/main.ts
```diff
@@ -366,14 +430,90 @@ const createMainWindow = async (): Promise<void> => {
...
+
+		// Basic non-Strict CSP based on https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html#basic-non-strict-csp-policy
+		const productionCsp =
+			"default-src 'none';" +
+			"script-src 'self' 'wasm-unsafe-eval';" +
+			// Hash is for inline style in index.html
+			"style-src 'self' 'sha256-XBXaUBQCe+0UGd1QCfoPFCc7UsLKd8xrn9oXNYqjFog=';" +
+			// react-resizable-panels has inline styles in elements. `style-src-attr 'unsafe-inline'` is slightly more narrow
+			// than just `style-src 'unsafe-inline'`, but we should still try to get rid of this.
+			"style-src-attr 'unsafe-inline';" +
```
slarse, line 446: @OliverJAsh @samhh @estib-vega So I did some extra digging here and it turns out the `unsafe-inline` was necessary for two reasons:

1. Inline style in index.html (fixed with hash, could also be fixed by moving the style to a CSS file)
2. The library `react-resizable-panels` uses inline styles in elements (i.e. in style attrs). So I've allowed `style-src-attr 'unsafe-inline'`, which is marginally less bad than `style 'unsafe-inline'` as it's more specific.

PR #13689, opened 2026-05-07
Compute behind count from octopus merge-base and update upstream display

mtsgrd:
### Behind count now reflects the farthest-behind stack

`behind` and `upstream_commits` on `BaseBranch` are now computed using the
octopus merge-base across all workspace stack tips and the target tip, instead
of counting from the stored `target.sha`. This means the behind count shows
the worst case across all stacks in the workspace.

### Updated CLI display

**Before:**
```
┊● 801d97d (upstream) ⏫ 10 new commits
├╯ 801d97d [origin/main] 2000-01-02 add upstream-10
```

**After:**
```
┊● 801d97d (merge base) ⏫ 10 commits
├╯ 0dc3733 (common base) 2000-01-02 add M
```

- Top line: target tip labeled `(merge base)` with commit count between the two.
- Bottom line: octopus merge-base labeled `(common base)` — the deepest shared
  ancestor across all stacks. Previously showed `[origin/main]`, which was
  misleading since this commit isn't where the target ref points.
- With `--upstream`, the individual commits are listed between the two lines.

### Desktop app button

- Shows `N behind` with a single stack, `≤N behind` with multiple stacks.

### Robustness improvements (from copilot review)

- Use `try_find_reference` for the workspace ref with fallback to `target.sha`.
- Avoid extra allocation and object lookup in `merge_base_octopus` call.
- Simplify `lowest_merge_base` from `Option<ObjectId>` to plain `ObjectId`.

## Test plan

- [x] `cargo test --workspace` — all tests pass
- [x] New test: `behind_reflects_farthest_behind_stack` asserts the behind count
  matches the farthest-behind stack (3 behind, not 1)
- [ ] Manual: verify `but status` and `but status --upstream` display
- [ ] Manual: verify desktop button shows `N behind` / `≤N behind`

### DIFF
--- a/crates/but/tests/but/command/snapshots/status/two-worktrees/status-with-worktrees-verbose.stdout.term.svg
+++ b/crates/but/tests/but/command/snapshots/status/two-worktrees/status-with-worktrees-verbose.stdout.term.svg
@@ -1,4 +1,4 @@
-<svg width="740px" height="398px" xmlns="http://www.w3.org/2000/svg">
+<svg width="740px" height="416px" xmlns="http://www.w3.org/2000/svg">
   <style>
     .fg { fill: #AAAAAA }
     .bg { fill: #000000 }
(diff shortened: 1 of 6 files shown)


Caleb-T-Owens (approved):
Lgtm.

Kind of confusing that the target ref commit is called `target_commit_id` given that terms is now synonyms with the target sha, but it's fine I guess

mtsgrd:
> Kind of confusing that the target ref commit is called target_commit_id given that terms is now synonyms with the target sha, but it's fine I guess

@Caleb-T-Owens I think you'll find about 150 mentions of this term in the code base, but I agree, I'd prefer `target_sha`.. maybe we could bulk replace tem?

PR #13695, opened 2026-05-07
cli+tui: make repeated but undos work

davidpdrsn:
Previously, if you ran `but undo` and then `but undo` again, you'd undo the undo. Thats not intended and instead, what we want to do is undo the operation before the previous undo, basically walking backwards in history. This makes `but undo` do that.

## How it works

I find the undo target by walking from newest to oldest through the oplog and counting how many undos we see. For each undo we see, we skip one non-undo entry. When we're at 0, we know the target is the next entry.

Example when all undos are at the top

```
                                    (n=0)
[UNDO]   <- skip one real operation (n=1)
[UNDO]   <- skip one real operation (n=2)
[REWORD] <- skipped                 (n=1)
[REWORD] <- skipped                 (n=0)
[REWORD] <- target                  take this!
```

And with undos mixed with other operations

```
                                         (n=0)
[UNDO]   <- skip one real operation      (n=1)
[REWORD] <- skipped                      (n=0)
[UNDO]   <- skip one more real operation (n=1)
[REWORD] <- skipped                      (n=0)
[REWORD] <- target                       take this!
```

## Undoing `but oplog restore`

I want `but oplog restore` to be treated differently from `but undo`, i.e., if you run `but oplog restore` and then `but undo`, then we should undo the restore. The restore shouldn't be counted as an undo.

This required changing `OperationKind`, which previously used `RestoreFromSnapshot` for both undos and restores. I've added a new `RestoreFromSnapshotViaUndo` variant that is used for undos so we can tell them apart from explicit restores. That means old undo entries will be categorized as restores, but I don't think there is anything we can do about that.

## `but redo`

We also want to add `but redo` to walk forwards in the oplog. I'm pretty confident we can use the same general strategy to support that as well, but if someone disagrees, I'd like to know 😊 since this introduces data changes. I want to make sure we don't need to make further changes when we do `but redo`. I might wait to merge this until I have a working `but redo` implementation as well.

## Iterating the oplog

Before, we had `ctx.list_snapshots()`, which returned a `Vec<Snapshot>` and required passing a limit on how many oplog entries to retrieve. I've changed that to instead be `ctx.snapshots_iter()`, which returns a lazy iterator. We need this because we don't know specifically how far to walk back to find the undo target, so we cannot provide a limit upfront.

One small behavior change here is that when iterating from a specific oplog commit, that commit isn't included, whereas before it would be. I think this behavior is more appropriate for pagination since clients can just pass the last commit and get exactly the next page to render. There was even a TODO about that in the Svelte app.

## GUI changes

The new `OperationKind` did require some changes in Svelte land, which I'm really not familiar with, so I would like some extra eyes on that.

### DIFF
--- a/crates/but-api/src/legacy/oplog.rs
+++ b/crates/but-api/src/legacy/oplog.rs
@@ -26,13 +26,14 @@ use tracing::instrument;
 
 /// List snapshots in the oplog.
 ///
-/// - `project_id`: The ID of the project to list snapshots for.
 /// - `limit`: Maximum number of snapshots to return.
-/// - `sha`: Optional SHA to filter snapshots starting from a specific commit.
+/// - `sha`: Optional SHA to filter snapshots starting after a specific commit.
 /// - `exclude_kind`: Optional list of operation kinds to exclude from the results.
 /// - `include_kind`: Optional list of operation kinds to include (if set, only these kinds are returned).
 ///
 /// Returns a vector of `Snapshot` entries.
+///
+/// Prefer using [`snapshots_iter`] if possible.
 ///
 /// # Errors
 /// Returns an error if the project cannot be found or if there is an issue accessing the oplog.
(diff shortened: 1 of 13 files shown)


crates/gitbutler-oplog/src/oplog.rs
```diff
@@ -718,9 +659,18 @@ fn commit_snapshot(
     Ok(snapshot_commit_id)
 }
 
+#[derive(Debug, Copy, Clone)]
+pub enum RestoreKind {
+    /// Restore via `but undo`
+    RestoreFromSnapshotViaUndo,
+    /// Restore via `but oplog restore`
+    ExplicitRestoreFromSnapshot,
+}
```
davidpdrsn: Note to self: Why did we need this?

crates/but-api/src/legacy/oplog.rs
```diff
@@ -102,9 +173,13 @@ pub fn create_snapshot(
...
+pub fn restore_snapshot(
+    ctx: &mut but_ctx::Context,
+    restore_kind: RestoreKind,
+    sha: gix::ObjectId,
+) -> Result<()> {
     let mut guard = ctx.exclusive_worktree_access();
-    ctx.restore_snapshot(sha, guard.write_permission())?;
+    ctx.restore_snapshot(sha, restore_kind, guard.write_permission())?;
     Ok(())
 }
```
davidpdrsn: Does changing this impact the GUI?

crates/but-api/src/legacy/oplog.rs
```diff
@@ -45,11 +46,79 @@ pub fn list_snapshots(
...
+        }
+
+        if real_operations_to_skip == 0 {
+            return Ok(Some(snapshot));
+        }
+        real_operations_to_skip -= 1;
+    }
+
+    Ok(None)
+}
```
slarse, line 120: For undos alone, this looks correct to me.

I think this can also be extended for redos, under the constraint that a redo can only be performed directly following an undo or another redo. That seems like a reasonable constraint given that I don't really know what the semantics would be if you do a sequence like `[COMMIT] -> [UNDO] -> [COMMIT] -> [REDO]`.

Given the constraint that redos are always direct descendants of other redos or undos, a redo will simply find the "target undo" by the same semantics as this algorithm, but with "redo as the undo" and "undo as the normal ops".

We then resolve the target of the undo and restore to its direct parent.

Undo would has to account for redos with a separate counter to adjust the final target. They can't affect the termination of the search as we then would only land on the "target redo", and as redo is defined by resolving the "target undo" we need undo to be dominant. Or the other way around, but one needs to be dominant. I think an easy way to account for redos when resolving an undo is to always skip over them and increment a separate counter. Once the termination status has been reached, we use that separate counter to adjust the final target.

This is based on a very short amount of thinking but it seems very doable with this structure.

crates/but/tests/but/command/undo.rs
```diff
@@ -173,3 +191,301 @@ fn can_undo_rewording_branch() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[test]
```
slarse, line 195: I think it would be good to have a test where we undo until there's nothing more to undo. Example:

```
but commit ...

but undo # <--- success
but undo # <--- error: no more operations to undo
```

crates/gitbutler-oplog/src/oplog.rs
```diff
@@ -1062,3 +1021,83 @@ fn find_oplog_child(
...
+    next_commit_id: Option<gix::ObjectId>,
+    skip_initial_commit: bool,
+    exclude_kind: Vec<OperationKind>,
+    include_kind: Option<Vec<OperationKind>>,
+}
+
+impl Iterator for SnapshotIter {
+    type Item = Result<Snapshot>;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
slarse, line 1036: I did not read this carefully, let me know if you don't feel confident with it and want me to take closer look.

slarse (approved):

crates/but-api/src/legacy/oplog.rs
davidpdrsn: Yep this was also the general direction that I had in mind.

crates/gitbutler-oplog/src/oplog.rs
davidpdrsn: This particular change was done by a clanker but I think its safe. We have pretty good test coverage of this and it was mostly moving code not really writing new stuff. Did review it myself as well.

crates/but/tests/but/command/undo.rs
davidpdrsn: Done!

mtsgrd:
lgtm

PR #13697, opened 2026-05-07
Fix PR/MR base branches going stale after pushing stacks

mtsgrd:
## Problem

When a stack's structure changes and gets pushed, existing PRs keep pointing at their **old base branch**, showing incorrect diffs on the forge.

```
Before: main ← A (PR#1, base: main) ← B (PR#2, base: A)
After:  main ← X ← A ← B    ← PR#1 still targets main (wrong)
```

## Fix

After pushing, walk every stack bottom-to-top, compute each review's expected target, and update any that drifted. The GitHub/GitLab update APIs already had `base`/`target_branch` fields — they were just always set to `None`.

**Where it runs:**
- **`but push`** — best-effort post-push correction (forge errors are logged, not fatal)
- **`but review`** — fixes drifted targets when publishing/updating reviews

**Key changes:**
- **`compute_review_target_updates()`** — pure function that maps stack branches to expected PR targets
- **`update_review_description_tables()`** — now handles both footer updates and target branch corrections in a single API call per review, with per-review error resilience
- **`ForgeReviewDescriptionUpdate.target_branch`** — new optional field; when set, the review's base is updated alongside its footer

## Tests

6 unit tests on `compute_review_target_updates` covering empty stacks, no-review branches, single PRs, full chains, skipped branches, and gaps in the middle.

### DIFF
--- a/crates/but-api/src/legacy/forge.rs
+++ b/crates/but-api/src/legacy/forge.rs
@@ -365,12 +365,12 @@ pub async fn set_review_draftiness(
     .await
 }
 
-/// Update the stacked review descriptions to have the correct footers.
+/// Update stacked reviews: description footers and, optionally, target branches.
 #[but_api(napi)]
 #[instrument(err(Debug))]
 pub async fn update_review_footers(
     ctx: ThreadSafeContext,
-    reviews: Vec<but_forge::ForgeReviewDescriptionUpdate>,
+    reviews: Vec<but_forge::ForgeReviewUpdate>,
 ) -> Result<()> {
     let (storage, forge_repo_info, preferred_forge_user) = {
         let ctx = ctx.into_thread_local();
@@ -384,7 +384,7 @@ pub async fn update_review_footers(
         )
     };
 
-    but_forge::update_review_description_tables(
+    but_forge::sync_reviews(
         &preferred_forge_user,
         &forge_repo_info.context("No forge could be determined for this repository branch")?,
         &reviews,
--- a/crates/but-forge/src/review.rs
+++ b/crates/but-forge/src/review.rs
@@ -1087,38 +1087,59 @@ pub async fn create_forge_review(
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[cfg_attr(feature = "export-schema", derive(schemars::JsonSchema))]
 #[serde(rename_all = "camelCase")]
-pub struct ForgeReviewDescriptionUpdate {
+pub struct ForgeReviewUpdate {
     /// The unique identifier number for this review within its repository. This can be a PR or MR number.
     pub number: i64,
     /// The current body/description of the review, which may be None if no description is set.
     pub body: Option<String>,
     /// The platform-specific symbol for this review type (e.g., "#" for GitHub pull requests and "!" for MRs).
     pub unit_symbol: String,
+    /// If set, update the base/target branch of this review to the given value.
+    pub target_branch: Option<String>,
 }
 
 #[cfg(feature = "export-schema")]
-but_schemars::register_sdk_type!(ForgeReviewDescriptionUpdate);
+but_schemars::register_sdk_type!(ForgeReviewUpdate);
 
-impl From<ForgeReview> for ForgeReviewDescriptionUpdate {
+impl From<ForgeReview> for ForgeReviewUpdate {
     fn from(review: ForgeReview) -> Self {
-        ForgeReviewDescriptionUpdate {
+        ForgeReviewUpdate {
             number: review.number,
             body: review.body,
             unit_symbol: review.unit_symbol,
+            target_branch: Some(review.target_branch),
         }
     }
 }
 
-/// Update the review description tables for a set of reviews
-pub async fn update_review_description_tables(
+impl From<ForgeReviewTargetUpdate> for ForgeReviewUpdate {
+    fn from(update: ForgeReviewTargetUpdate) -> Self {
+        ForgeReviewUpdate {
+            number: update.number,
+            body: None,
+            unit_symbol: String::new(),
+            target_branch: Some(update.target_branch),
+        }
+    }
+}
+
+/// Update reviews: description footers and, optionally, target/base branches.
+///
+/// Per-review failures are collected rather than aborting the batch.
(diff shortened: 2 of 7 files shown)


slarse:
@mtsgrd This PRs exact title and description were also applied over on https://github.com/gitbutlerapp/gitbutler/pull/13695 (I'm restoring it so check the history if you want to verify).

I also saw that the buildkite PR conspicuously got rebranded as a speedup PR: https://github.com/gitbutlerapp/gitbutler/pull/13699

Were those manual blunders or might there be some bug in this PR? It's about updating PRs after all.

krlvi:
this appears to be adding new API surface - we should probably take a deeper look

mtsgrd:
@slarse thanks for cleaning up my mistakes. it's unfortunately ai hallucinations, 99% of the time when I say update my pr description it gets it right, but apparently not _every_ time. I need to make sure I'm more specific in my requests going forward.

crates/but-api/src/legacy/forge.rs
```diff
@@ -393,6 +393,35 @@ pub async fn update_review_footers(
...
+    };
+
+    but_forge::update_review_target_branches(
+        &preferred_forge_user,
+        &forge_repo_info.context("No forge could be determined for this repository branch")?,
+        &updates,
+        &storage,
+    )
+    .await
+}
```
estib-vega: A quick overview of the code makes sense, the only change I'd like to request is that this should be combined with the function to update the footers: `update_review_footers`

This is already called in `but pr`. Ideally this is a single API that fixes the base and the review footers, no need of doing it in two separate API calls

estib-vega: `update_review_description_tables` already calls the update API, so might as well do all in there. 
Just need to rename it, of course

mtsgrd: Thanks, I've pushed a new version if you wouldn't mind taking a look?

crates/but-forge/src/review.rs
```diff
@@ -1105,11 +1107,14 @@ impl From<ForgeReview> for ForgeReviewDescriptionUpdate {
             number: review.number,
             body: review.body,
             unit_symbol: review.unit_symbol,
+            target_branch: None,
         }
```
estib-vega, line 1122: `ForgeReview` has a `target_branch` field.

We could rename `ForgeReviewDescriptionUpdate` to `ForgeReviewUpdate` so that we don't need to manually set the `target_branch` bellow

crates/but-forge/src/review.rs
```diff
@@ -1105,11 +1107,14 @@ impl From<ForgeReview> for ForgeReviewDescriptionUpdate {
...
     }
 }
 
-/// Update the review description tables for a set of reviews
+/// Update reviews: description footers and, optionally, target/base branches.
+///
+/// Per-review failures are collected rather than aborting the batch.
 pub async fn update_review_description_tables(
```
estib-vega: Can we rename this too? `update_review` or `sync_reviews`. I don't know what would be the best name, but we ideally convey from the name that we're fixing the reviews to have both targets and (if needed) descriptions updated

crates/but/src/command/legacy/forge/review.rs
```diff
@@ -723,12 +722,21 @@ async fn publish_reviews_for_branch_and_dependents(
...
+        .into_iter()
+        .map(|(review, expected_target)| {
+            let target_branch = if review.target_branch != expected_target {
+                Some(expected_target)
+            } else {
+                None
+            };
+            let mut update: but_forge::ForgeReviewDescriptionUpdate = review.into();
+            update.target_branch = target_branch;
+            update
```
estib-vega: Read the comment at the beginning. Ideally the `From` implementation takes cage of this already

crates/but-forge/src/review.rs
estib-vega: (nice to have): Also, would be cool if there were a `From<ForgeReviewTargetUpdate>` implementation as well, so that you don't have to manually map the type from `compute_review_target_updates`

crates/but-api/src/legacy/forge.rs
estib-vega: Left a couple of comments more.
Thanks for unifying the APIs!

mtsgrd:
Amazing, thanks @estib-vega. I think I've addressed all of your comments if you wouldn't mind another look?

Here's what I understand about the new flow:
```
                  ┌─────────────┐
                  │  but push   │
                  └──────┬──────┘
                         │
                         ▼
              push branches to remote
                         │
                         ▼
        update_review_targets_for_stacks()
                         │
                         ▼
           compute_review_target_updates()
           (walks stacks, computes expected
            target for each review)
                         │
                         ▼
          ForgeReviewTargetUpdate ──Into──▶ ForgeReviewUpdate
                                            • body: None
                                            • unit_symbol: ""
                                            • target_branch: Some(...)
                                                     │
                                                     │
                  ┌─────────────┐                    │
                  │ but review  │                    │
                  └──────┬──────┘                    │
                         │                           │
                         ▼                           │
            publish / fetch reviews                  │
            (has full ForgeReview)                   │
                         │                           │
                         ▼                           │
          ForgeReview ──Into──▶ ForgeReviewUpdate    │
                                • body: Some(...)    │
                                • unit_symbol: "#"   │
                                • target_branch:     │
                                  Some if drifted,   │
                                  None if matches    │
                                         │           │
                                         ▼           ▼
                                ┌────────────────────────┐
                                │ update_review_footers  │  (but-api)
                                └───────────┬────────────┘
                                            │
                                            ▼
                                  ┌───────────────────┐
                                  │   sync_reviews    │  (but-forge)
                                  └───────────┬───────┘
                                              │
                              has_footer_content?
                              (any unit_symbol non-empty?)
                                    /        \
                                  yes         no
                                  /             \
                           recompute         skip body
                           PR footer
                                  \             /
                                   \           /
                                    ▼         ▼
                              ┌──────────────────────┐
                              │  GitHub/GitLab API   │
                              │  per review:         │
                              │  • body (if present) │
                              │  • base/target_branch│
                              │    (if present)      │
                              └──────────────────────┘
```

PR #13706, opened 2026-05-08
Parallelize but-api-macros CI tests with a matrix strategy

mtsgrd:
The but-api-macros tests were running 4 sequential cargo test invocations
(default, legacy, tauri, napi features), each triggering a recompile.
Use a GitHub Actions matrix to run them in parallel on separate runners.


Caleb-T-Owens (approved):
LGTM, it would be nice then to get rid of the corresponding Makefile entry.

PR #13709, opened 2026-05-08
Add CLI E2E tests for live GitHub and GitLab forge workflows

mtsgrd:
First CLI-level end-to-end test suite that exercises `but` against real
forge APIs on a twice-daily schedule via GitHub Actions.

Test coverage (e2e/cli/):
  - Create single PR / MR
  - Create draft PR / MR
  - Toggle draft ↔ ready state
  - Create stacked PRs / MRs with correct base branch targets
  - Force-push after amend keeps PR / MR open

Infrastructure:
  - `but config forge auth --provider <github|gitlab> --token <PAT>`
    Non-interactive auth for CI — stores token via but-api without
    prompts. Also available via BUT_FORGE_TOKEN env var.
  - `but-secret`: activate git-credentials backend when
    E2E_TEST_APP_DATA_DIR is set (CI containers lack a system keychain).
  - Test helper creates isolated app-data dir + git credential-store
    per run, clones a sandbox repo, and cleans up branches/PRs after.

CI (.github/workflows/e2e-cli-continuous.yml):
  schedule: 0 6,18 * * *  (twice daily)
  build CLI → run vitest → upload JUnit XML to Buildkite Test Engine

### DIFF
--- a/crates/but/src/args/config.rs
+++ b/crates/but/src/args/config.rs
@@ -404,6 +404,13 @@ pub enum MetricsStatus {
     Disable,
 }
 
+/// Forge providers for non-interactive authentication.
+#[derive(Debug, Clone, Copy, clap::ValueEnum)]
+pub enum ForgeProvider {
+    Github,
+    Gitlab,
+}
+
 impl MetricsStatus {
     pub fn enabled(self) -> bool {
         matches!(self, MetricsStatus::Enable)
@@ -435,7 +442,24 @@ pub enum ForgeSubcommand {
     /// GitLab
     ///  - Personal Access Token (PAT)
     ///  - Self-Hosted
-    Auth,
+    ///
+    /// ## Non-interactive usage (CI)
+    ///
+    /// ```text
+    /// but config forge auth --provider github --token ghp_...
+    /// but config forge auth --provider gitlab --token glpat-...
+    /// ```
+    Auth {
+        /// Personal Access Token to store. When provided, authentication is non-interactive.
+        #[clap(long, env = "BUT_FORGE_TOKEN", hide_env_values = true)]
+        token: Option<String>,
+        /// Forge provider to authenticate with (required when --token is used).
+        #[clap(long, value_enum, requires = "token")]
+        provider: Option<ForgeProvider>,
+        /// Host for GitHub Enterprise or self-hosted GitLab instances.
+        #[clap(long)]
+        host: Option<String>,
+    },
 
     /// List authenticated forge accounts known to GitButler.
     ///
--- a/.github/workflows/e2e-cli-continuous.yml
+++ b/.github/workflows/e2e-cli-continuous.yml
@@ -1,0 +1,79 @@
+name: CLI E2E (Continuous)
+
+on:
+  schedule:
+    # Twice daily: 6am and 6pm UTC.
+    - cron: "0 6,18 * * *"
+  workflow_dispatch:
+
+jobs:
+  cli-e2e:
+    name: cli-e2e-forge
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: read
+    container:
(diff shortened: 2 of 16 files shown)


mtsgrd:
@krlvi in order to get live test runs twice daily using the `but` command, I'd need the following if you can help out?

#### 1. Create sandbox repos

- **GitHub**: Create a repo like `gitbutler-test/cli-e2e-sandbox` with a single initial commit on `main`. No branch protection rules.
- **GitLab**: Create a repo like `gitbutler-test/cli-e2e-sandbox` with the same setup.

Both should be empty repos that tests can freely push to and open PRs/MRs against.

#### 2. Create bot accounts / tokens

- **GitHub**: Fine-grained PAT scoped to the sandbox repo with `Contents: Read & Write` + `Pull Requests: Read & Write`
- **GitLab**: Project access token with `api` scope

#### 3. Add secrets and variables in GitHub Actions

Go to the repo's **Settings > Secrets and variables > Actions**:

**Secrets:**

| Name | Value |
|------|-------|
| `CLI_E2E_GITHUB_TOKEN` | The GitHub PAT from step 2 |
| `CLI_E2E_GITLAB_TOKEN` | The GitLab token from step 2 |
| `BUILDKITE_SUITE_TOKEN_CLI_E2E` | The Buildkite suite token from step 3 |

**Variables:**

| Name | Value |
|------|-------|
| `CLI_E2E_GITHUB_REPO` | e.g. `gitbutler-test/cli-e2e-sandbox` |
| `CLI_E2E_GITLAB_REPO` | e.g. `gitbutler-test/cli-e2e-sandbox` |

krlvi:
im not sure about the e2e setup here with typescript. Id like to defer to @davidpdrsn @slarse @Byron + others involved in the CLI - How do we want the CLI e2e setup to look like?

Byron:
Thanks for reading me in.

I took a look and tried to be open-minded about the E2E CLI. At first I thought if it's just about managing the sandbox repo using the GH API, TS probably isn't too bad. Then I thought that a lot of that `but-github` could probably do as well, and if not it certainly is good to have even for the clients one day.
Then I saw code that sets up a test-repository, and thought if this would not already be covered by `but-testsupport`?

Whether or not existing plumbing crates should see changes to support E2E with forges, I think the implementation language should be familiar to those doing the `but` CLI, which then would force Rust.

Knowing myself, I tend to not care much about TS because I find it too 'fluffy', which would reduce my interest in interacting with it and making it better. Also, I'd always feel I couldn't review TS properly even if an agent does it, and I don't even want to get into TS anymore either.

While admitting bias, it seems objectively better to do this in Rust even if the code will certainly look more complex due to typing. But… it's worth it?

crates/but/src/args/config.rs
```diff
@@ -435,7 +442,24 @@ pub enum ForgeSubcommand {
...
+    /// ## Non-interactive usage (CI)
+    ///
+    /// ```text
+    /// but config forge auth --provider github --token ghp_...
+    /// but config forge auth --provider gitlab --token glpat-...
+    /// ```
+    Auth {
+        /// Personal Access Token to store. When provided, authentication is non-interactive.
+        #[clap(long, env = "BUT_FORGE_TOKEN", hide_env_values = true)]
+        token: Option<String>,
```
slarse, line 455: Taking a token as a CLI argument is not good practice, this is a prime avenue for unwitting devs to accidentally leak secrets. We don't need it for CI so I think it shouldn't be there.

I'd suggest only reading from the environment variable, or potentially taking it via stdin, to be used something like:

```bash
$ but config forge auth --provider github --token-stdin < token.txt
```

slarse (commented):
Minor comment on passing token via command line argument.

Other than that, I agree with @Byron. There should be a strong reason for writing tests in a different language, and I don't see any here. This seems to largely recreate `but-testsupport` with some additional and missing features.

That being said, this is a great initiative and I'd love to assist in getting this kind of testing in. I just don't think this is the way to do it.

PR #13723, opened 2026-05-08
Remove superflous anonymous lifetime

Caleb-T-Owens:


Byron (approved):
LGTM!

While at it, I updated the `rust-version` to what we actually use which may drive clippy to show more hints.

PR #13727, opened 2026-05-09
Update the `rust-version` for all crates to the one we actually use

Byron:
This is something we'd have to do from time to time for the latest lints. Unfortunately, nothing new comes up.
Motivated by #13723

### Tasks

* [x] update rust-version
* [x] update `gix-testtools`


krlvi (approved):
nice

PR #13738, opened 2026-05-10
Launch WSL editor URLs via CLI

robertmonka:
## Summary

  Fix opening VS Code-compatible editor URLs when GitButler is running inside WSL.

  Before this change, editor URLs such as `vscode://file/...` or `cursor://file/...` did not open the
editor from WSL. They were passed to the normal Linux URL opener and ended up opening the browser
instead.

  For supported editors, GitButler now launches the matching editor CLI directly and keeps the existing
opener as a fallback.

  ## Changes

  - Detect WSL on Linux before using the generic URL opener.
  - Launch supported editor URLs through their CLI:
    - `vscode` -> `code`
    - `vscode-insiders` -> `code-insiders`
    - `vscodium` -> `codium`
    - `cursor` -> `cursor`
    - `windsurf` -> `windsurf`
    - `trae` -> `trae`
  - Preserve `windowId=_blank` with `--new-window`.
  - Preserve line/column targets with `--goto`.
  - No new dependencies.

  ## Validation

  - `cargo fmt --check --all`
  - `cargo test -p but-api --features legacy open::tests --lib`

  Manual WSL check:
  - Built and installed GitButler locally for WSL and Windows.
  - Before: editor URLs opened the browser instead of the editor.
  - After: editor URLs from a WSL-backed project open through the editor CLI.
  - Unsupported schemes still fall back to the existing opener path.

  AI assistance was used while preparing the patch. I reviewed the diff and tested it on a Windows/WSL
setup.

### DIFF
--- a/crates/but-api/src/legacy/open.rs
+++ b/crates/but-api/src/legacy/open.rs
@@ -25,6 +25,11 @@ pub(crate) fn open_that(target_url: &Url) -> anyhow::Result<()> {
         bail!("Invalid path scheme: {}", target_url.scheme());
     }
 
+    #[cfg(target_os = "linux")]
+    if open_wsl_editor_url(target_url) {
+        return Ok(());
+    }
+
     fn clean_env_vars<'a, 'b>(
         var_names: &'a [&'b str],
     ) -> impl Iterator<Item = (&'b str, String)> + 'a {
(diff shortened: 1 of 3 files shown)


crates/but-api/src/legacy/open.rs
```diff
@@ -94,6 +99,150 @@ pub(crate) fn open_that(target_url: &Url) -> anyhow::Result<()> {
     Ok(())
 }
 
+#[cfg(target_os = "linux")]
+fn open_wsl_editor_url(target_url: &Url) -> bool {
```
Byron: While this module isn't documented at all, let's add documentation to this function just to explain why it's there and what it does.

Byron:
Thanks a lot for contributing, it's much appreciated.

I gave it a little makeover and hope you can test it in the next nightly, just to be sure.

Byron (approved):

robertmonka:
Opened follow-up PR #13811 for the WSL editor-launch regression.

The #13738 code path was merged, but the WSL detection was still too fragile for the GUI process. It relied on `WSL_DISTRO_NAME` / `WSL_INTEROP`, which are not guaranteed to be present in GitButler's GUI environment. When both are missing, GitButler skips the WSL editor CLI path and falls back to the generic URL opener, which can open the browser again for `vscode://file/...` URLs.

The follow-up switches detection to `/proc/sys/kernel/osrelease`, which reflects the running WSL kernel independently of process environment.

PR #13759, opened 2026-05-12
Persist change id in lower level commit function

Caleb-T-Owens:
By persisting the determined change id in the commit.rs `create` function, we’re able to handle it well in the `new_commit` function, which means that operations like reword will persist their change ids as well


Byron (approved):
It's great to see this come full circle!

While I think that Copilot should be considered, I think the approach is sound as `fn change_id()` takes what's already there in the header or generates a change id from an object id.

Please note that this was an armchair review, and there may be details I am missing in my naiveté.

PR #13761, opened 2026-05-12
Rebuild workspace commit when upstream target is already up-to-date

mtsgrd:
When `default_target.sha` in TOML already matches origin/master (e.g.
after `bootstrap_default_target_if_missing` writes the remote tip
directly), `integrate_upstream` returned `UpToDate` and skipped
rebuilding the workspace commit. This left the workspace commit
parented on an older base while the UI showed commits behind.

Now we always rebuild the workspace commit in the `UpToDate` path so
the worktree reflects the current target even when the metadata was
already advanced.


mtsgrd:
@Caleb-T-Owens could you take a quick look?

PR #13791, opened 2026-05-13
Factor out usages of `BaseBranch` for obtaining push URLs

Caleb-T-Owens:
The goal here was not to change any behaviour or the relationship between `remote_url` and `push_remote_url`.

My primary goal was to simply remove some reliance on the `BaseBranch` datastructure and get the ws_meta passed into `target_to_base_branch`

### DIFF
--- a/crates/but-core/src/ref_metadata.rs
+++ b/crates/but-core/src/ref_metadata.rs
@@ -1,3 +1,4 @@
+use anyhow::{Context as _, Result, bail};
 use gix::refs::FullNameRef;
 use uuid::Uuid;
 
@@ -34,6 +35,12 @@ pub struct Workspace {
     /// Note that even though this is per workspace, the implementation can fill in global information at will.
     pub target_ref: Option<gix::refs::FullName>,
 
+    /// The remote that we used to push to. Now primarily used to determine what
+    /// forge we're using.
+    ///
+    /// You probably want to use [`Self::push_remote_url`] instead in most
+    /// cases. Ideally we should wean this property out of the codebase.
+    pub remote_url: String,
     /// The commit id of a commit that was reachable by [`Self::target_ref`] and that should be included in the workspace.
     /// This is useful to make workspaces appear stable in relationship to the target reference, which may be updated each
     /// time a `git fetch` is performed.
@@ -53,6 +60,7 @@ impl std::fmt::Debug for Workspace {
             ref_info,
             stacks,
             target_ref,
+            remote_url,
             push_remote,
             target_commit_id,
         } = self;
@@ -63,6 +71,7 @@ impl std::fmt::Debug for Workspace {
                 "target_ref",
                 &MaybeDebug(&target_ref.as_ref().map(|rn| rn.as_bstr())),
             )
+            .field("remote_url", remote_url)
             .field("target_commit_id", &MaybeDebug(target_commit_id))
             .field("push_remote", &MaybeDebug(push_remote))
             .finish()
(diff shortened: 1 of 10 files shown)


krlvi (approved):

crates/gitbutler-branch-actions/src/base.rs
```diff
@@ -347,18 +342,25 @@ fn set_exclude_decoration(ctx: &Context) -> Result<()> {
...
-        .id;
+    // This function is presuming a workspace ref name.
+    let ws_meta = meta.workspace(WORKSPACE_REF_NAME.try_into()?)?;
+    let target_ref = ws_meta
+        .target_ref
+        .as_ref()
+        .context("target_to_base_branch require a defined target_ref")?;
+    let target_sha = ws_meta
+        .target_commit_id
+        .context("target_to_base_branch require a defined target_commit_id")?;
```
estib-vega, line 356: I guess this is intended to use the same metadata for any checkout, event if we're not in the `gitbutler/workspace` ref, right?

estib-vega (approved):
Makes sense to me. Did not do any particular testing, but I agree with the intent

crates/gitbutler-branch-actions/src/base.rs
Caleb-T-Owens: There is the thought that this particular metadata probably wants to be global. Aside from that however, choosing this reference preserves the previous behaviour of this function, which feels reasonable.

crates/but-core/src/ref_metadata.rs
```diff
@@ -158,6 +167,51 @@ impl Workspace {
...
+                .ok()
+                .and_then(|remote| {
+                    remote
+                        .url(gix::remote::Direction::Push)
+                        .or_else(|| remote.url(gix::remote::Direction::Fetch))
+                        .map(|url| url.to_bstring().to_string())
+                })
+                .unwrap_or_else(|| self.remote_url.clone()),
+            None => self.remote_url.clone(),
+        };
```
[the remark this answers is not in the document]
Caleb-T-Owens: wrong?

crates/gitbutler-branch-actions/src/base.rs
```diff
@@ -83,19 +83,12 @@ impl BaseBranch {
 
 #[instrument(skip(ctx), err(Debug))]
 pub fn get_base_branch_data(ctx: &Context) -> Result<BaseBranch> {
-    let target = default_target(ctx)?;
     let repo = ctx.repo.get()?;
-    let base = target_to_base_branch(&repo, &ctx.legacy_project, &target)?;
+    let meta = ctx.meta()?;
+    let base = target_to_base_branch(&repo, &ctx.legacy_project, &meta)?;
     Ok(base)
```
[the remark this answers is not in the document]
Caleb-T-Owens: fixed

crates/but-workspace/tests/workspace/branch/apply_unapply_commit_uncommit.rs
```diff
@@ -861,6 +861,7 @@ fn apply_multiple_segments_of_stack_in_order_merge_if_needed() -> anyhow::Result
                 },
             ],
             target_ref: "refs/remotes/origin/main",
+            remote_url: "should not be needed and when it is extract it from `repo`",
```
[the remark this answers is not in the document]
Caleb-T-Owens: not relevant

crates/but-core/src/ref_metadata.rs
```diff
@@ -34,6 +35,12 @@ pub struct Workspace {
     /// Note that even though this is per workspace, the implementation can fill in global information at will.
     pub target_ref: Option<gix::refs::FullName>,
 
+    /// The remote that we used to push to. Now primarily used to determine what
+    /// forge we're using.
+    ///
+    /// You probably want to use [`Self::push_remote_url`] instead in most
+    /// cases. Ideally we should wean this property out of the codebase.
+    pub remote_url: String,
```
Byron, line 43: This is a danger-zone as `Workspace` is supposed to not contain redundant information. The `target_ref` would be enough to obtain the remote, and from there one can obtain everything else.

There should also not be any pieces of data that could contradict each other, or that could contradict Git.

If UX requires overrides for certain things, we have to call them that, but then there is a strong argument for placing these in `.git/config` to make them discoverable. All the metadata here will eventually move to the DB after all.

And… it's merged. Great 😁.

Byron: I should have written @Caleb-T-Owens above.
By now I removed this field.

PR #13792, opened 2026-05-13
Use standard upstream commits metric

Caleb-T-Owens:
The reason why we were seeing the “Branches are all up to date” error was because the metric that was being used by `get_base_branch_data` and `integrate_upstream` where different.

This PR both corrects the list of upstream commits to correctly account for merge commits, and updates both suspect functions to use the same metric

### DIFF
--- a/crates/but-api/src/legacy/virtual_branches.rs
+++ b/crates/but-api/src/legacy/virtual_branches.rs
@@ -270,8 +270,9 @@ pub fn switch_back_to_workspace_with_perm(
     ctx: &mut but_ctx::Context,
     perm: &mut RepoExclusive,
 ) -> Result<BaseBranch> {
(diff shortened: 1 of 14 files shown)


Byron (commented):
I see what's needed here: more traversal utilities to do `a ^b` kind of traversals, fast.

With that said, this PR *would* need a lot of work and while this implementation might yield desired results, I recommend not bothering with using the graph like this and instead use a `repo` based ancestor traversal with `hide(b)`. It will be faster, and I have less code to look at when bringing in the latest improvements from [this PR](https://github.com/gitbutlerapp/gitbutler/pull/13786).

krlvi (approved):

Caleb-T-Owens:
Fixed the daft AI nits. I'm going to follow up with some tests and other improvements, but in the interest of landing an improved release, I'm OK with landing this now.

PR #13793, opened 2026-05-13
feat(but): add `--checkout-to` override for teardown

slarse:
This is a combined bugfix and feature addition. Or rather, the feature addition obviates the need for a bugfix. There are many ways to fail to resolve the correct local branch automatically, so instead of fixing that I'm just adding in a `--checkout-to` option that allows the user to explicitly say where to checkout to when tearing down.

`but teardown` would fail if there were no active stacks, as it used active stacks to determine the branch to checkout to. With `--checkout-to` you can now decide yourself where to checkout to.

In the event the user encounters the "no active stacks" situation, a message suggesting the use of `--checkout-to` is printed.

### DIFF
--- a/crates/but/src/args/mod.rs
+++ b/crates/but/src/args/mod.rs
@@ -299,6 +299,7 @@ pub enum Subcommands {
     /// This command:
     /// - Creates an oplog snapshot of the current state
     /// - Finds the first active branch and checks it out
+    ///     - Alternatively, use `--checkout-to <branch>` to override this default
     /// - Cherry-picks any dangling commits from gitbutler/workspace
     /// - Provides instructions on how to return to GitButler mode
     ///
@@ -313,9 +314,17 @@ pub enum Subcommands {
     /// but teardown
     /// ```
     ///
+    /// ```text
+    /// but teardown --checkout-to my-feature-branch
+    /// ```
+    ///
     #[cfg(feature = "legacy")]
     #[clap(verbatim_doc_comment)]
-    Teardown,
+    Teardown {
+        /// Explicit override for which local branch to checkout to.
+        #[clap(long, short = 'c', value_name = "LOCAL_BRANCH")]
+        checkout_to: Option<String>,
+    },
 
     /// Updates all applied branches to be up to date with the target branch.
     ///
--- a/crates/but/src/command/legacy/teardown.rs
+++ b/crates/but/src/command/legacy/teardown.rs
@@ -59,12 +59,10 @@ pub(crate) fn teardown(
         let ref_name: gix::refs::PartialName = checkout_to
             .clone()
             .try_into()
-            .context(format!("Invalid ref name: {checkout_to}"))?;
+            .with_context(|| format!("Invalid ref name: {checkout_to}"))?;
         let resolved_ref = repo.find_reference(ref_name.as_ref())?;
         if !matches!(resolved_ref.name().category(), Some(Category::LocalBranch)) {
-            anyhow::bail!(format!(
-                "Invalid ref for checkout: '{checkout_to}' is not a local branch"
-            ))
+            anyhow::bail!("Invalid ref for checkout: '{checkout_to}' is not a local branch")
         }
         Some(resolved_ref.name().shorten().to_string())
     } else {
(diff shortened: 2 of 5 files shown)


crates/but/src/args/mod.rs
```diff
@@ -313,9 +314,17 @@ pub enum Subcommands {
...
+    /// but teardown --checkout-to my-feature-branch
+    /// ```
+    ///
     #[cfg(feature = "legacy")]
     #[clap(verbatim_doc_comment)]
-    Teardown,
+    Teardown {
+        /// Explicit override for which local branch to checkout to.
+        #[clap(long, short = 'c', value_name = "LOCAL_BRANCH")]
+        checkout_to: Option<String>,
```
slarse, line 326: `String` is not entirely correct here as refs do not have to be UTF8, but... we use it everywhere else in the CLI so I don't want to make it inconsistent here alone.

We'll have to do some overhaul on replacing ref-inputs with BString at some point, but now is not that time.

crates/but/src/command/legacy/teardown.rs
```diff
@@ -97,19 +117,27 @@ pub(crate) fn teardown(ctx: &mut Context, out: &mut OutputChannel) -> anyhow::Re
...
     };
 
-    // Sort by order to ensure we get the leftmost (lowest order) stack first
-    stacks.sort_by_key(|s| s.order.unwrap_or(usize::MAX));
-
-    let target_stack = stacks
-        .first()
-        .ok_or_else(|| anyhow::anyhow!("No active branches found"))?;
+    let target_branch_name = if let Some(checkout_to) = checkout_to {
+        checkout_to
```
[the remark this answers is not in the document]
slarse: Good point.

crates/but/src/command/legacy/teardown.rs
```diff
@@ -48,6 +53,24 @@ pub(crate) fn teardown(ctx: &mut Context, out: &mut OutputChannel) -> anyhow::Re
         anyhow::bail!("Not on gitbutler/workspace branch");
     }
 
+    // Note: Validate checkout_to before snapshot creation to prevent unnecessary snapshot
+    let checkout_to = if let Some(checkout_to) = &checkout_to {
+        let repo = ctx.repo.get()?;
+        let ref_name: gix::refs::PartialName = checkout_to
+            .clone()
+            .try_into()
+            .context(format!("Invalid ref name: {checkout_to}"))?;
```
davidpdrsn: drive-by-nit: you can use `with_context` to only allocate a string if something actually goes wrong

```suggestion
            .with_context(|| format!("Invalid ref name: {checkout_to}"))?;
```

crates/but/src/command/legacy/teardown.rs
```diff
@@ -48,6 +53,24 @@ pub(crate) fn teardown(ctx: &mut Context, out: &mut OutputChannel) -> anyhow::Re
...
+        let resolved_ref = repo.find_reference(ref_name.as_ref())?;
+        if !matches!(resolved_ref.name().category(), Some(Category::LocalBranch)) {
+            anyhow::bail!(format!(
+                "Invalid ref for checkout: '{checkout_to}' is not a local branch"
+            ))
```
davidpdrsn: nit: I might be wrong but I'm pretty sure `bail!` accepts format strings directly

```suggestion
            anyhow::bail!(
                "Invalid ref for checkout: '{checkout_to}' is not a local branch"
            )
```

estib-vega (approved):

estib-vega:
Thanks for fixing this

crates/but/src/command/legacy/teardown.rs
slarse: It appears it does, fixed.

crates/but/src/command/legacy/teardown.rs
slarse: I'm split on this kind of optimization. The stickler in me thinks "yes, goooood, gooood". But the pragmatic in me considers how this slightly reduces readability and is a tiny piss in a large ocean where a single `but status` opens every file with unassigned changes 5 times over :s

Stickler always wins, though. Fixed!

davidpdrsn: Maybe I have Stockholm syndrome but I don't think it meaningfully hurts readability.

PR #13797, opened 2026-05-13
add oplog things to sdk

davidpdrsn:
This adds the oplog related functions that the TUI uses to the SDK:

- `get_undo_target_snapshot`: Returns the snapshot that and `undo` should restore to. Just returns the snapshot, doesn't actually do the restore, so that we can show a confirmation modal.
- `get_redo_target_snapshot`: Same but for `redo`.
- `restore_snapshot_with_kind`: Actually does the restore. The `kind` is required to say if it was a undo/redo/manual restore.
- `peel_restore_snapshot`: Takes a restore snapshot and finds the actual snapshot that would be restored to. Otherwise the confirmation for a redo cannot show what snapshot we'll actually restore. It would just say "Redo 4220dc7 Undo" but with this it can say "Redo 4220dc7 Inserted blank commit"


estib-vega (approved):
Niceee

PR #13803, opened 2026-05-13
gitbutler-edit-mode: write evolution metadata

jonathantanmy2:
This is the beginning of commit evolution metadata in GitButler. In this commit, a new experimental feature is introduced; by default, it is off in production but on in tests. When it is on, evolution metadata is written whenever the user exits edit mode.

This commit also introduces a dependency on git-meta-lib. Because
GitButler depends on a specific Git commit of gix (instead of declaring
a version number to be obtained from crates.io), this commit adds
a "[patch]" section to Cargo.toml to override the gix pulled in by
git-meta-lib to be the same as the one that GitButler uses (so that
we don't include two versions of gix in the final build, bloating the
binary).

As of this commit, commit metadata is only written, not used. The usage of commit metadata will be introduced in a subsequent commit (well, after a non-trivial amount of work).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -217,7 +217,7 @@ bstr = "1.12.1"
 zip = { version = "4", default-features = false, features = ["bzip2"] }
 # Add the `tracing` or `tracing-detail` features to see more of gitoxide in the logs. Useful to see which programs it invokes.
 # When adjusting this, update `gix-testtools` as well!
-gix = { version = "0.83.0", git = "https://github.com/GitoxideLabs/gitoxide", rev = "575113dfb10b3ba12eb57f57a81b241e773968bd", default-features = false, features = [
+gix = { version = "0.83.0", default-features = false, features = [
     "sha1", "sha256"
 ] }
 gix-testtools = { version = "0.19.0", git = "https://github.com/GitoxideLabs/gitoxide", rev = "575113dfb10b3ba12eb57f57a81b241e773968bd" }
@@ -311,6 +311,12 @@ self_cell = "1.2.2"
 unicode-segmentation = "1.13.2"
 urlencoding = "2.1"
 inventory = "0.3"
+git-meta-lib = { version = "0.1.10" }
+
+[patch.crates-io]
+# Make all dependencies use the same gix as our crates, so that we don't end up
+# with two versions of gix compiled in.
+gix = { git = "https://github.com/GitoxideLabs/gitoxide", rev = "575113dfb10b3ba12eb57f57a81b241e773968bd" }
 
 [workspace.lints.clippy]
 all = { level = "deny", priority = -1 }
--- a/crates/gitbutler-edit-mode/Cargo.toml
+++ b/crates/gitbutler-edit-mode/Cargo.toml
@@ -33,6 +33,7 @@ serde.workspace = true
 tracing.workspace = true
 schemars = { workspace = true, optional = true }
 but-schemars = { workspace = true, optional = true }
+git-meta-lib.workspace = true
 
 [dev-dependencies]
 but-testsupport.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -177,7 +177,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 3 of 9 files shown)


Byron:
> This commit also introduces a dependency on git-meta-lib. Because GitButler depends on a specific Git commit of gix [..]

Now that this turns out to be an issue, I will try harder to keep it on the released version.
The next release is due on this 22., and then I will switch it over and try to keep it there.

crates/gitbutler-edit-mode/Cargo.toml
```diff
@@ -33,6 +33,7 @@ serde.workspace = true
 tracing.workspace = true
 schemars = { workspace = true, optional = true }
 but-schemars = { workspace = true, optional = true }
+git-meta-lib.workspace = true
```
Byron, line 36: I can't find evidence of this dependency being optional, and I also don't see the feature toggle just yet.

Byron (commented):
For safety I have put this PR into DRAFT mode so there are no accidental merges while the `git-meta` feature isn't optional.
Maybe my brief review is too early, but then again, the PR wasn't draft before.

crates/gitbutler-edit-mode/Cargo.toml
jonathantanmy2: Using a Cargo feature does mean that we can compile GitButler without including git-meta-lib, reducing binary size (and compile time), but I'm not convinced that it's a good tradeoff. Assuming that we turn on the feature by default in CI, that means that we'll be releasing a different binary compared to what we test in CI, and it also makes it hard for users to test the feature if they want (they have to compile GitButler themselves). But if binary size is important and/or this is how we typically build experimental features, let me know and I'll change it.

Also, @Caleb-T-Owens might have opinions on this.

Byron: Ah, sorry. I really thought you meant cargo features and somehow didn't implement it yet.

By the looks of it, `git-meta` is going to happen anyway, so might as well bring it in now. The duplicate dependencies can be fixed later, even though it will always be an issue once these implementations drift, once  a month. `git-meta` I can't update.

On the other hand, generally, yes, I find it very important to not let experimental features be added to the dependency tree, unless they're explicitly requested. I'd compile this one in. It's just a matter of tidiness to me, and efficiency as well, compile times and binary sizes have an impact downstream.

And no, I don't see a problem with releasing something that has features deactivated, as long as that path is at least compiled by CI. In `gitoxide` there is a ton of cargo features and CI just tests all relevant combinations on them. For those who are inclined, there even is a tool that runs all permutations.

There already is a precedent for this as well, which is the IRC feature, disabled in production builds and not compiled in.

jonathantanmy2: If the gix dependency version is ever a problem, we can specify a range for it just like we do for rusqlite (see https://github.com/git-meta/git-meta/pull/62 ).

And just checking...so for this case, would you rather use a Cargo feature or (like in this PR currently) have something that's always compiled in?

Byron: As this experiment would be observable at least with larger binaries in the product, I'd love to use cargo features to kick the duplicate `gitoxide` problem a bit further into the future.

jonathantanmy2: One thing I just realized is that the current `master` already pulls in `git-meta-lib` (in the `but-agentlog` crate) meaning that there is already 2 versions of `gix` compiled in. Even if we make this PR a Cargo feature, we still can't avoid the duplicate `gitoxide` problem.

So I didn't make this PR a Cargo feature, but since binary size is a concern, I've instead updated this PR to use `[patch]` to override `git-meta-lib`'s dependency on `gix`. You can see from the `Cargo.lock` that there is now only one `gix` being compiled in.

Cargo.toml
```diff
@@ -311,6 +311,10 @@ self_cell = "1.2.2"
 unicode-segmentation = "1.13.2"
 urlencoding = "2.1"
 inventory = "0.3"
+git-meta-lib = { version = "0.1.10" }
+
+[patch.crates-io]
+gix = { git = "https://github.com/GitoxideLabs/gitoxide", rev = "575113dfb10b3ba12eb57f57a81b241e773968bd" }
```
Byron, line 319: This is a great idea! Solves the problem.

Could you leave some comments to explain why this is patched, for future us?

crates/gitbutler-edit-mode/Cargo.toml
Byron: Oh wow, this passed by my eagle eye 😅.

But if anything, then this conversation led to a deduplication of our dependencies, which is great all by itself. I also never thought of using patches for this, but it makes perfect sense and I will keep this in mind if a similar problem arises in the future.

Thanks!

Cargo.toml
jonathantanmy2: Ah, good point. Done.

jonathantanmy2:
I've updated the commit message and PR description to match the new behavior (only one copy of `gix` will be compiled in) and have turned off draft (i.e. this is ready for review). I'll still need code owner review.

krlvi (approved):

PR #13811, opened 2026-05-14
Extend WSL detection to check kernel version strings

robertmonka:
## Summary

Follow-up to #13738.

That PR added a WSL-specific path that opens editor URLs through the editor CLI instead of falling back to the generic URL opener. The code path was merged correctly, but WSL detection still relied on `WSL_DISTRO_NAME` / `WSL_INTEROP`.

Those environment variables are not guaranteed to be present in the GitButler GUI process. When both are missing, GitButler does not detect WSL and falls back to the generic URL opener, which can open the browser again for `vscode://file/...` URLs.

This changes WSL detection to use `/proc/sys/kernel/osrelease` as the single source of truth and check for the `microsoft` marker in the running kernel release.

## Notes

On the affected WSL2 machine:

```text
/proc/sys/kernel/osrelease = 6.6.114.1-microsoft-standard-WSL2
```

`/proc/sys/kernel/osrelease` comes from the running kernel, so it does not depend on which environment variables the GUI process inherited.

## Verification

Built and installed locally with:

```sh
GITBUTLER_SKIP_SKILL_UPDATE=1 ./compile.sh bin
```

### DIFF
--- a/crates/but-api/src/legacy/open.rs
+++ b/crates/but-api/src/legacy/open.rs
@@ -123,8 +123,13 @@ fn open_editor_url_as_command_invocation_on_wsl(target_url: &Url) -> bool {
 }
 
 fn is_wsl() -> bool {
-    cfg!(target_os = "linux") && env::var_os("WSL_DISTRO_NAME").is_some()
-        || env::var_os("WSL_INTEROP").is_some()
+    cfg!(target_os = "linux")
+        && std::fs::read_to_string("/proc/sys/kernel/osrelease")
+            .map(|release| {
+                let release = release.to_ascii_lowercase();
+                release.contains("microsoft")
+            })
+            .unwrap_or(false)
 }
 
 /// Builds the editor CLI command needed to open a WSL editor URL.


Byron (commented):
Thanks a lot, this looks good to me.

And indeed, I should have mentioned that environment variables are not the same when a GUI app is launched, and I didn't pick up on you probably launching it from the terminal.

Before merging though, could you do me a favor and create a bundled app that actually can be launched without a terminal?

    pnpm tauri build --debug  --config crates/gitbutler-tauri/tauri.conf.nightly-local.json

(*the `--debug` is of great importance*)

If that app is double-clicked like a GUI app, it should come up like one as well with the same environment - do not open it from a terminal. But it's special as you can type `env` and it will make a backend call that can be inspected in the network tab of the dev tools.

Image

Are there any markers that you think relate to WSL? If so, let's check for these instead or only check for the file in `/proc` if that marker is set.

The reason for this is that I'd like to avoid prodding `/proc` in general, so restricting it to a subset would be good.

robertmonka:
Thanks for the review and for asking me to validate this from the GUI environment.

I went back and re-tested the issue from scratch. The browser fallback I was seeing turned out not to be caused by GitButler's WSL detection. In my local WSL setup, the VS Code WSL server was corrupted/mis-extracted, so `code` itself was failing:

```text
wslCode.sh: 60: ~/.vscode-server/bin/<commit>/bin/remote-cli/code: not found
```

After removing that broken VS Code server install and letting VS Code recreate `~/.vscode-server/bin/<commit>`, `code` works again and GitButler opens VS Code correctly without any GitButler changes.

So I don't think this PR is the right fix for the behavior I was trying to address. I'll close/drop this and avoid adding WSL detection changes based on a bad local repro.

If we still want to harden WSL detection separately, I can open a smaller follow-up that only gates the `/proc` check behind a known WSL/WSLg marker, but that would be a separate cleanup rather than a fix for this issue.

PR #13822, opened 2026-05-15
but-rebase: Immutable extra refs

estib-vega:
Add the ability to include refs to traverse in the editor graph,
without mutating them. This is required to be able to pick commits in
refs that we don’t want to change.

---

### Why do we need this?

In the case of e.g. integration upstream changes from the remote into the local branch, we want to be able to pick, or merge in commits from the remote, without modifying it. This ensures that we are able to do that safely.

### DIFF
--- a/crates/but-rebase/src/graph_rebase/rebase.rs
+++ b/crates/but-rebase/src/graph_rebase/rebase.rs
@@ -29,13 +29,17 @@ impl<'ws, 'graph, M: RefMetadata> Editor<'ws, 'graph, M> {
         // multiple.
 
         let mut ref_edits = vec![];
-        let steps_to_pick = order_steps_picking(
-            &self.graph,
-            &self
-                .graph
-                .externals(Direction::Incoming)
-                .collect::<Vec<_>>(),
-        );
+        let rebase_heads = self
+            .graph
+            .externals(Direction::Incoming)
+            .filter(|idx| {
+                !matches!(
+                    &self.graph[*idx],
+                    Step::Reference { refname } if self.immutable_references.contains(refname)
+                )
+            })
+            .collect::<Vec<_>>();
+        let steps_to_pick = order_steps_picking(&self.graph, &rebase_heads);
 
         // A 1 to 1 mapping between the incoming graph and the output graph
         let mut graph_mapping: HashMap<StepGraphIndex, StepGraphIndex> = HashMap::new();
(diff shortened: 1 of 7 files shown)


crates/but-rebase/src/graph_rebase/rebase.rs
```diff
@@ -139,7 +144,7 @@ impl<'ws, 'graph, M: RefMetadata> Editor<'ws, 'graph, M> {
                             gix::refs::TargetRef::Object(id) => {
                                 if id == to_reference {
                                     unchanged_references.push(refname.clone());
-                                } else {
+                                } else if !is_immutable {
```
Caleb-T-Owens, line 147: Should we error if we have tried to update an immutable reference?

If we had `A <- B <- C (refs/heads/foo - immutable) (refs/heads/bar - mutable)`.

If we modify B, we end up with `A <- B <- C (refs/heads/foo - immutable) A <- B' <- C'  (refs/heads/bar - mutable)` which seems quite undesirable and probably something we should loudly complain about at the very least in test builds

Caleb-T-Owens (approved):
I love the concept. This is a helpful thing to have.

crates/but-rebase/src/graph_rebase/rebase.rs
estib-vega: I think that's actually the intent of this change. 
I want to be able to modify nodes and other refs with the assurance that all the commits reachable from the immutable ref stay as is, despite that creating divergence with other overlapping refs.

This is exactly what I need fo interactive upstream branch integration.

There's a test for that as well.

crates/but-rebase/tests/rebase/graph_rebase/materialize.rs
```diff
@@ -195,3 +200,105 @@ fn both_methods_update_references_identically() -> Result<()> {
...
+    |\  
+    * | c869f24 (stack-1) Commit D
+    * | 07a9b49 Commit C
+    |/  
+    | * a0f2ac5 (origin/main, main) Commit X
+    | * f555940 Commit A
+    |/  
+    * d664be0 (stack-2) Commit B
+    * fafd9d0 init
+    ");
```
estib-vega, line 255: Here: Even if I remove a commit reachable from a mutable ref, the immutable ref preserves it

estib-vega:
Merging this for now, but happy to keep the conversation about failure conditions going

PR #13823, opened 2026-05-15
Extend new stack dropzone to accept commits

mtsgrd:
Allow dragging commits onto the new stack dropzone to create a new
stack and move the commits there. Uses commitMove with withStackBusy
for proper UI state management. Also adds Stacks/StackDetails cache
invalidation to the commitMove endpoint.


cbjeukendrup:
Oh. I already had an open PR for the same thing... https://github.com/gitbutlerapp/gitbutler/pull/13191

PR #13825, opened 2026-05-15
Immutable parent refs

Caleb-T-Owens:
When working with the editor, we want to be really careful to only edit references that are displayed in the GitButler workspace.

If we were to modify references the user can’t see, this is almost certainly unexpected “bad gitizen” behaviour.

This PR refines the newly introduced “immutable” reference behaviour to consider any reference in the revset `(immutable entrypoints) ^(mutable entrypoints)` also as an immutable reference.

### DIFF
--- a/crates/but-rebase/src/graph_rebase/creation.rs
+++ b/crates/but-rebase/src/graph_rebase/creation.rs
@@ -62,14 +62,8 @@ impl<'ws, 'meta, M: RefMetadata> Editor<'ws, 'meta, M> {
         // common base
         let entrypoint = workspace.graph.lookup_entrypoint()?;
 
-        let mut entrypoints = vec![entrypoint.segment_index];
-
-        let immutable_references = options
-            .extra_refs
-            .iter()
-            .filter(|extra_ref| extra_ref.mutability == ExtraRefMutability::Immutable)
-            .map(|extra_ref| extra_ref.ref_name.to_owned())
-            .collect::<HashSet<_>>();
+        let mut mutable_entrypoints = vec![entrypoint.segment_index];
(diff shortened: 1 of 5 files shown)


estib-vega (approved):
Tested against the needs for integration of branch upstream commits, and it works as expected

crates/but-rebase/src/graph_rebase/creation.rs
```diff
@@ -81,18 +75,46 @@ impl<'ws, 'meta, M: RefMetadata> Editor<'ws, 'meta, M> {
...
+                immutable_entrypoints.push(segment.id);
+            }
         }
 
         let mut segments_to_add = vec![];
         let mut seen_segments = HashSet::new();
 
-        for entrypoint in entrypoints {
+        for entrypoint in mutable_entrypoints {
+            workspace.graph.visit_all_segments_including_start_until(
```
Byron, line 89: @Caleb-T-Owens There should be no need to re-traverse the graph as each commit is already classified, commits have `flags`. Commits marked as `Integrated` are reachable from the target ref.
They may additionally be marked as InWorkspace if they are reachable by the workspace tip. Those who are Integrated (in-workspace or not) or only remote (I think there is a utility method for this) would probably be the ones that shouldn't be changed as we don't want to rewrite integrated history or history we don't own (i.e. remote history).

If you think you must roll your own, then please use `find_segments_reachable_from_a_not_b` as what I see here is the naive and slow version - there is no need for that anymore.

My recommendation is to use an agent as an oracle as they would find APIs of interest, even if it doesn't implement anything.

Let me CC @estib-vega here as I think he might also work more with the Graph already or in future.

PR #13855, opened 2026-05-17
Fix flaky CI test in many_commits_in_all_branch_types

mtsgrd:
The test was skipped on CI with an `is_ci` guard added as a workaround.
This removes the skip and fixes the root cause.

## Root cause

The fixture used a `tick()` function that set `GIT_COMMITTER_DATE` to a
fixed timestamp in January 2023. On git 2.54, `git maintenance` runs
automatically in the background after commits (triggered via `--auto`).
It packs loose objects and creates cruft packs, assigning each object an
mtime derived from its referencing commit's committer date. With a 2023
timestamp, all objects appeared over two years old — exceeding
`gc.cruftPackExpire=2.weeks.ago` — and were pruned immediately,
corrupting the repository mid-fixture.

## Fix

- Remove `tick()` so commits use the current wall-clock time, preventing
  objects from appearing stale to the cruft pack expiry logic.
- Reduce commit counts (feature: 100→10, non-virtual-feature: 50→10,
  a-branch-1: 10→3) so the total loose object count stays below the
  `maintenance.loose-objects` auto threshold of 100. This prevents
  background packing from triggering at all during fixture creation,
  eliminating the race entirely.
- Remove the `is_ci` skip guard so the test actually runs on CI.

## Verification

Stress tested with `reproduce-flakiness.sh` using git 2.54.0 on
ubuntu:24.04 (matching CI):

| Fixture | Runs | Failures |
|---------|------|----------|
| Original (tick + 100/50 commits) | 300 | ~4% |
| Fixed (no tick + 10/10 commits) | 500 | 0 |

Written by claude

### DIFF
--- a/crates/gitbutler-branch-actions/tests/branch-actions/virtual_branches/list_details.rs
+++ b/crates/gitbutler-branch-actions/tests/branch-actions/virtual_branches/list_details.rs
@@ -1,4 +1,3 @@
-use but_testsupport::gix_testtools::is_ci;
 use gitbutler_branch_actions::BranchListingDetails;
 
 use crate::virtual_branches::list;
@@ -51,10 +50,6 @@ fn one_vbranch_in_workspace_single_commit() -> anyhow::Result<()> {
 
 #[test]
 fn many_commits_in_all_branch_types() -> anyhow::Result<()> {
-    if is_ci::cached() {
-        // This test is flaky on CI and it's code that is due to removal.
-        return Ok(());
-    }
     let ctx = project_ctx("complex-repo")?;
     let list = branch_details(&ctx, ["feature", "non-virtual-feature"])?;
     assert_eq!(list.len(), 2);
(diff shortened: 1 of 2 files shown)


Byron:
I was just disabling this one on CI over here: https://github.com/gitbutlerapp/gitbutler/pull/13821 .

If this merges first, please be aware and rebase to remove that early-bail once it's properly fixed here.

Regarding the analysis here, I am not sure it's correct. What happens is that it executes the script, which writes everything into the directory which is then the location that can be *read* from. If any of the tests that use the same fixture would accidentally write to it, other readers might be affected. You could try to provide writable fixtures for proper isolation. Copying hard-links will be fine, it's just a full copy then if it crosses filesystems.

mtsgrd:
Thanks @Byron. I eventually got to what I think is a valid conclusion. Could you have another look?

Byron:
Thanks so much for digging into this!

I now see that the large amount of commits created in that fixture really did trigger GC :D, and then it all unfolded as described in the commit:

> Root cause: the fixture used a tick() function that set GIT_COMMITTER_DATE
to a fixed timestamp in January 2023. On git 2.54, git maintenance runs
automatically in the background after commits (triggered by git commit via
--auto). It packs loose objects and creates cruft packs, assigning each
object an mtime derived from its referencing commit's committer date. With
a 2023 timestamp, all objects appeared over two years old — exceeding
gc.cruftPackExpire=2.weeks.ago — and were pruned immediately, corrupting
the repository mid-fixture.

On the 22nd, there will be a new release which automatically prevents this, in future.
In any case, I think it makes sense to not have 100 iteration for that fixture as well, so I don't think this will ever bug us again.

PR #13858, opened 2026-05-18
follow-up on but-graph updates

Byron:
Follow-up on  https://github.com/gitbutlerapp/gitbutler/pull/13821.

TBD

### Tasks

* [ ] The traversal could auto-pick target refs if unset to fill in a target most of the time, including one read from a global ref like `GITBUTLER_TARGET`.
    * [ ] If so, have an example `but` version that works in SBM automatically as long as there is a remote. If not, it 
* [ ] auto-set the target ref in single-branch mode according to the new discovery rules (remote/HEAD, initial branch name, never the same as the current branch, communicated through tips_of_interest)
* [ ] https://github.com/gitbutlerapp/gitbutler/pull/13626
    - [ ] use it's test-case (presuming it is what I think it should be) as case for adding the `highest-base` to points of interest.
    - [ ]  Show that we pick up virtual branches at the highest and lowest base
    - [ ] Add an option to create reference so it can add it at the highest or lowest base respectively.
* [ ] https://github.com/gitbutlerapp/gitbutler/pull/13627
* [ ] https://github.com/gitbutlerapp/gitbutler/pull/13546
* [ ] https://github.com/gitbutlerapp/gitbutler/pull/13620
    - [ ] deduplicate and offer a way to attach more data `Graph`. This way, the per-commit enrichment can be shared in `ctx`
* [ ] TODO: about #13552


Byron:
I am now pretty certain that this review doesn't need to happen.

PR #13872, opened 2026-05-18
ci: add checksum validation to cargo-nextest

slarse:
## 🧢 Changes
Adds in checksum validation for the cargo-nextest archive. This is necessary to ensure that we're installing what we think we're installing.


krlvi (approved):

slarse:
Stuck in CI limbo, will try to close and reopen.

PR #13880, opened 2026-05-19
upgrade reqwest to v0.13 which comes with platform TLS certs

### LINKED ISSUE #13407: Gitbutler no longer recognizes security sert for selfhosted gitlab instance Unknown Issuer
### Version

0.19.4-current

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

As of GitButler 0.19.4 configuring intergration of self hosted gitlab instance no longer works. After update this is the visual. There is an error loading the user information. The cert from the instance is trusted, the PAT has API Permisions and if I downgrade back to 0.19.3 it works again.

Below are error photos, with host and username redacted for security reasons. Happy to talk to someone directly about this issue to help get it fixed, ive tried everything.

UI Error
Image

Console Error
Image

Logs

Image

1. Configure Gitlab integration on 0.19.3
2. Update to 0.19.4+

_No response_

```shell

```

Byron:
Various fixes as part of daily catchup.

### Tasks

* [x] That way, `rustls` will be able to load personal trust chains which
      in turn will validate https certificates to internal forges. Fixes #13407

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -345,12 +345,11 @@ dependencies = [
 
 [[package]]
 name = "async-openai"
-version = "0.33.1"
+version = "0.38.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc48c3deb4ad9a2ee8c8e364c79eb0f74e69e17ed7e883d55988b90ea44fe986"
+checksum = "2bb666d696156a8da537278ecb3195806d885edf25a573ce23ceec4dee2b0eee"
 dependencies = [
  "async-openai-macros",
- "backoff",
  "base64 0.22.1",
  "bytes",
  "derive_builder",
@@ -358,8 +357,7 @@ dependencies = [
  "futures",
  "getrandom 0.3.4",
  "rand 0.9.4",
- "reqwest 0.12.28",
- "reqwest-eventsource",
+ "reqwest",
  "secrecy",
  "serde",
  "serde_json",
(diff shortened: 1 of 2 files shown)


Byron:
CC @slarse due to the change to how we handle the verification of certificates when creation connections in the backend.

Byron:
@slarse Sorry for all the noise, but I think you should be the one to review this innocent looking, but possibly wide-ranging PR properly. I was just driving by and didn't think about the repercussions of these feature options, nor did I ask an agent to help with the research. Please, could you compensate for my sloppiness 🙏🥹?

slarse (commented):
@Byron Had a look at this!

Switching over to the native cert roots is probably a good thing overall. My main concern here is that the next version of reqwest (0.13) completely changes how to configure this and also uses a different verifier backend. The feature flag enabled here doesn't exist in 0.13.

From what I can gather, the [default in 0.13 is to use `rustls`](https://github.com/seanmonstar/reqwest/discussions/2906) together with [rustls-platform-verifier](https://github.com/rustls/rustls-platform-verifier), which also uses OS roots and native cert validation (where available). It seems like an overall significantly better option than the [rustls-native-certs](https://github.com/rustls/rustls-native-certs) backend that backs the feature flag enabled in this PR.

Would it be an option to upgrade `reqwest` to 0.13.3 instead?

> Note: I'm unsure of the availability of OS roots on various Linux distributions. I think it should be good for the vast majority of major distros, but it's possible that the lib may fail to discover them so we may need a [fallback on webpki](https://github.com/rustls/rustls-platform-verifier#user-content-fn-3-b385347fff23b65533a7570160a8e18c). Not entirely sure how to accomplish that via `reqwest` but I suppose it should be possible.

Byron:
Thank you!

> Would it be an option to upgrade reqwest to 0.13.3 instead?

I didn't attempt this in this PR and also never investigated why we need this specific version. Right now the OpenAI crate already pulls in v0.13, so it would definitely be better to unify it.

Let's try to upgrade.

Byron:
Alright, that should be it!

Thanks again!

slarse (approved):
Nice! I'll be sure to try out the nightly on a few Linux distros to see that we don't run into any issues.

PR #13885, opened 2026-05-19
Support Git-style external subcommands (but-<command>)

schacon:
<details>
<summary>Original description by @schacon </summary>
Allow unknown subcommands to invoke PATH helpers named but-<command>
(e.g. `but forecast` runs `but-forecast`) while preserving legacy
behavior for filesystem paths. Add argument/docs updates, an external
dispatcher (command/external.rs), wiring into the command dispatch and
metrics, tests for Unix PATH delegation, and update help tests and docs.
Map spawning NotFound errors to the friendly “not a command”
message.
</details>

I (@slarse) have completely clobbered this PR with my own stuff now. I pushed to this PR to retain the discussion.

Anyhow, this PR adds in support for Git-style external subcommands by allowing `but <COMMAND> [ARGS]` to delegate to any program on PATH called `but-<COMMAND>`, under the following conditions:

* `<COMMAND>` is not a built-in command
* `<COMMAND>` only consists of characters in the set `[a-zA-Z_-]`. This is _overly_ restrictive but I figured we can loosen it up later, I just didn't feel confident you couldn't pull any shenanigans with this.

An important design goal here is that this external command execution shouldn't make error messages worse. If we fail to match `<COMMAND>` against anything on PATH (we currently just try to execute), we reparse the command alone with external command support disabled, which will still print the nice Clap error message about the subcommand not being recognized, including any potential near matches. See tests for details.

> Note: Given that I could solve the "deterioration" of error messages when failing to match against an external command, I find no good reason to put this behind an app setting, feature toggle or environment variable. It'll just become available to everyone, and most will never notice it's there.

> Windows? Nope, not in this PR. I will try to add that separately but I need to setup a Windows dev environment first.

### DIFF
--- a/crates/but/skill/references/reference.md
+++ b/crates/but/skill/references/reference.md
@@ -580,6 +580,14 @@ Available on most commands:
 - `-C, --current-dir <PATH>` - Run as if started in different directory
 - `-h, --help` - Show help for command
 
+## External commands (PATH helpers)
+
+> Important: Not available for Windows yet
+
+Similar to Git, if `<command>` is not a built-in `but` command and `but-<command>` exists on `PATH`, `but` runs that executable instead (for example `but forecast …` invokes `but-forecast …`).
+
+Restriction: `<command>` must consist of characters in the set `[a-zA-Z_-]`
+
 ## Getting More Help
 
 ```bash
--- a/crates/but/src/args/mod.rs
+++ b/crates/but/src/args/mod.rs
@@ -6,6 +6,8 @@
 ///
 /// Nearly all documentation for the CLI is defined here using `clap` attributes,
 /// which are then used to generate help messages and online documentation.
+#[cfg(unix)]
+use std::ffi::OsString;
 use std::path::PathBuf;
 
 #[derive(Debug, clap::Parser)]
@@ -50,7 +52,10 @@ pub struct Args {
     /// {"result": ..., "status_error": ...} if the status query fails (in which case "status" is absent).
     #[clap(long = "status-after", global = true)]
     pub status_after: bool,
-    /// Subcommand to run.
+    /// Subcommand to run (`but <COMMAND>`).
+    ///
+    /// On UNIX, if `<COMMAND>` is not built in and `but-<COMMAND>` exists on the PATH, that program
+    /// is executed with the remaining arguments (`but <COMMAND> [ARGS]`).
     #[clap(subcommand)]
     pub cmd: Option<Subcommands>,
 }
@@ -1189,6 +1194,11 @@ pub enum Subcommands {
     #[clap(hide = true)]
     #[clap(verbatim_doc_comment)]
     EvalHook,
+
+    #[cfg(unix)]
+    #[strum(disabled)]
+    #[clap(hide = true, external_subcommand)]
+    External(Vec<OsString>),
 }
 
 pub mod alias;
--- a/crates/but/src/command/external.rs
+++ b/crates/but/src/command/external.rs
@@ -1,0 +1,73 @@
+//! Run `but-{name}` executables discovered on [`std::env::var_os`]("PATH"), akin to Git's `git-send-email`-style helpers.
+
+use std::{
+    ffi::{OsStr, OsString},
+    io::ErrorKind,
+    os::unix::ffi::OsStrExt,
+    path::Path,
+    process::{Command, Stdio},
+};
+
+use anyhow::Context;
+
+use crate::{CliError, CliResult, bad_input};
+
+pub(crate) fn dispatch(current_dir: &Path, extra: &[std::ffi::OsString]) -> CliResult<()> {
+    let subcommand_name = match extra {
+        [head, ..] => head,
+        _ => {
+            return Err(
+                anyhow::anyhow!("BUG: external subcommand parsed without any arguments").into(),
+            );
+        }
+    };
+
+    if !is_allowed_command_name(subcommand_name) {
+        return Err(bad_input("Subcommand contains illegal characters")
+            .arg_value(subcommand_name.to_string_lossy())
+            .hint("Are you trying to write an extension command 'but-<command>'? Make sure that '<command>' only contains characters in the set [a-zA-Z_-]")
+            .into());
+    }
+
+    let mut prefixed = OsString::from("but-");
+    prefixed.push(subcommand_name);
+
+    let status = match Command::new(&prefixed)
+        .args(&extra[1..])
+        .current_dir(current_dir)
+        .stdin(Stdio::inherit())
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
+        .status()
+    {
+        Ok(status) => status,
+        Err(e) if e.kind() == ErrorKind::NotFound => {
+            return Err(CliError::ExternalCommandNotFound(
+                subcommand_name.to_owned(),
+            ));
+        }
+        Err(e) => {
+            return Err(e)
+                .with_context(|| format!("could not invoke `{prefixed:?}` from PATH"))
+                .map_err(CliError::from);
+        }
+    };
+
+    if status.success() {
+        Ok(())
+    } else {
+        std::process::exit(status.code().unwrap_or(1));
+    }
+}
+
+/// A command name must consist of characters [a-zA-Z_-].
+///
+/// This is overly restrictive, but I just want to prevent people from doing anything funny here. We
+/// can loosen this as we need to in the future.
+fn is_allowed_command_name(command_name: &OsStr) -> bool {
+    command_name
+        .as_bytes()
+        .iter()
+        .copied()
+        .all(|c| c.is_ascii_uppercase() || c.is_ascii_lowercase() || c == b'-' || c == b'_')
+}
--- a/crates/but/src/error.rs
+++ b/crates/but/src/error.rs
@@ -2,6 +2,9 @@
 
 use std::fmt::Display;
 
+#[cfg(unix)]
+use std::ffi::OsString;
+
 use crate::theme::{self, Paint};
 
 /// Signifies that a command could not complete its intended action due to the user providing input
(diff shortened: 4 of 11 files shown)


crates/but/src/command/external.rs
```diff
@@ -0,0 +1,77 @@
...
+        Err(e) if e.kind() == ErrorKind::NotFound => {
+            anyhow::bail!(
+                "\"but {}\" is not a command. Type \"but --help\" to see all available commands.",
+                head.to_string_lossy()
+            );
+        }
+        Err(e) => {
+            return Err(e).with_context(|| format!("could not invoke `{prefixed}` from PATH"));
+        }
+    };
```
Byron, line 54: I wondered if there was anything else to implementing this, and if this somehow could be abused.
But overall, I think there is "absolutely nothing to worry about"™️.

crates/but/src/command/external.rs
```diff
@@ -0,0 +1,77 @@
...
+    );
+
+    if extra.len() == 1 {
+        let candidate = Path::new(&extra[0]);
+        let resolved = resolve_against_current_dir(current_dir, candidate);
+        if resolved.exists() {
+            return crate::command::gui::open(&resolved)
+                .with_context(|| format!("failed to open path `{}`", resolved.display()));
+        }
+    }
```
Byron: This could be problematic as attacker-controlled repos with submodules maybe can now bait users to open their repos.
It's far-fetched though as this `gui::open` just launches the GUI on a given directory that also must be a repo to get anywhere, and the question is if this is abusable. Maybe it is if the GUI is vulnerable to certain injection attacks.

Git doesn't have that problem.

It feels like to mitigate this, one would have to search the PATH by hand, and if a program exists one would always have to execute it.

Byron (commented):
I took a quick look and tried really hard to abuse this, also in comparison to how Git implements it. Codex helped with that, and I filtered it down to just one possible issue.

It would be great to also summon @slarse to the devsec party.

<details><Summary>research by codex</Summary>

Summary for GitHub:

`git foo` dispatches through Git’s top-level wrapper. Git parses global options, prepends its trusted exec path (`GIT_EXEC_PATH` / compiled gitexecdir) to `PATH`, tries builtins first, then invokes an external `git-foo` via argv-based process execution. It does not run a shell for normal external commands. Shell execution is limited to explicit `!` aliases. Git also prevents aliases from overriding normal commands, except deprecated commands, and repository safety checks such as `safe.directory` limit reading config/hooks from untrusted ownership contexts. External `git-foo` helpers are still trusted code from the effective `PATH`.

The Rust `but-{name}` dispatcher has the important no-shell property too: `Command::new(...).args(...)` passes arguments directly. The larger difference is that it relies on the ambient `PATH` without first prepending or prioritizing a trusted helper directory. That makes execution sensitive to PATH order, relative PATH entries, current-directory entries, or writable directories ahead of the intended install location. It also constructs the helper name with `to_string_lossy()`, which can alter non-UTF-8 input, and uses a permissive command-name check.

Suggested hardening:

- Try builtins before external helpers.
- Resolve helpers from a trusted install/libexec directory before ambient `PATH`.
- Resolve the executable to an absolute path before setting `current_dir`.
- Build `but-<name>` as an `OsString`, avoiding lossy Unicode conversion.
- Restrict helper names to a simple ASCII allowlist such as `[A-Za-z0-9-]+`.
- Treat `but-*` helpers as trusted plugin code and document that clearly.
- Consider sanitizing environment/file descriptor inheritance if helpers may run in sensitive contexts.

</details>

crates/but/src/command/external.rs
slarse: There's an easy solution to this: Kill it. We decided on a removal of this feature in the last engineering meeting as it's a low-frequency command that really doesn't deserve top-level placement. Doing `but gui <path>` instead is perfectly serviceable.

crates/but/src/command/external.rs
slarse: I don't really see anything either. This seems fine.

crates/but/tests/but/command/external.rs
slarse, line 1: We need a lot more testing for the negative cases

crates/but/tests/but/command/mod.rs
```diff
@@ -16,6 +16,8 @@ mod config;
 mod diff;
 #[cfg(feature = "legacy")]
 mod discard;
+#[cfg(all(unix, feature = "legacy"))]
+mod external;
```
slarse: What's legacy about this? And should we also support Windows?

slarse (commented):
I don't really have any major reservations here. While of course it opens up for some amount of misuse, the utility of the feature outweighs any such concerns IMO. Also, if a `but-malicious` is on the path, whether it's accidentally executed through `but` or by just running `but-malicious` outright doesn't seem like a great distinction.

If we want to be _extra_ prudent about it, we could have a command allow-list in our app settings, or potentially just an "allow external commands" setting (e.g. `but config external enable`). That would limit any potential security concerns to the probably very small subset of people that would ever use this feature. I think that kind of toggle might be worthwhile.

Other than that I think we should get this in ASAP. @schacon @Byron wdyt?

crates/but/tests/but/command/mod.rs
Byron: Indeed, no need.

slarse:
I'm taking over this PR :)

crates/but/src/command/external.rs
```diff
@@ -0,0 +1,75 @@
+//! Run `but-{name}` executables discovered on [`std::env::var_os`]("PATH"), akin to Git's `git-send-email`-style helpers.
+
+use std::{
+    ffi::{OsStr, OsString},
+    io::ErrorKind,
+    os::unix::ffi::OsStrExt,
+    path::Path,
+    process::{Command, Stdio},
+};
```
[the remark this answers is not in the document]
slarse: Already fixing this.

crates/but/src/error.rs
```diff
@@ -123,6 +128,9 @@ impl Display for CliError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Self::BadInput(value) => value.fmt(f),
+            Self::ExternalCommandNotFound(_) => {
+                writeln!(f, "TODO what to do here?")
+            }
             Self::Internal(value) => value.fmt(f),
         }
     }
```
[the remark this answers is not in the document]
slarse: Oops!

crates/but/skill/references/reference.md
```diff
@@ -580,6 +580,12 @@ Available on most commands:
 - `-C, --current-dir <PATH>` - Run as if started in different directory
 - `-h, --help` - Show help for command
 
+## External commands (PATH helpers)
+
+Similar to Git, if `<command>` is not a built-in `but` command and `but-<command>` exists on `PATH`, `but` runs that executable instead (for example `but forecast …` invokes `but-forecast …`).
+
+Restriction: `<command>` must consist of characters in the set `[a-ZA-Z_-]`
```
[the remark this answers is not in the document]
slarse: Oops!

crates/but/src/args/mod.rs
```diff
@@ -50,7 +50,10 @@ pub struct Args {
     /// {"result": ..., "status_error": ...} if the status query fails (in which case "status" is absent).
     #[clap(long = "status-after", global = true)]
     pub status_after: bool,
-    /// Subcommand to run.
+    /// Subcommand to run (`but <COMMAND>`).
+    ///
+    /// If `<COMMAND>` is not built in and `but-<COMMAND>` exists on the PATH, that program is
+    /// executed with the remaining arguments (`but <COMMAND> [ARGS]`).
```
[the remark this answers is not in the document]
slarse: Yep

crates/but/src/command/external.rs
```diff
@@ -0,0 +1,73 @@
...
+        .stdin(Stdio::inherit())
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
+        .status()
+    {
+        Ok(status) => status,
+        Err(e) if e.kind() == ErrorKind::NotFound => {
+            return Err(CliError::ExternalCommandNotFound(
+                subcommand_name.to_owned(),
+            ));
```
slarse, line 47: A new CliError variant. This is necessary to signal to the calling code that there was no external command matching the name, so we can act on that. Current action is to re-parse without external command support to generate a nice error message.

crates/but/src/error.rs
```diff
@@ -75,16 +78,18 @@ impl Display for BadInput {
...
-            (None, Some(value)) => writeln!(f, "Bad input '{}'", t.attention.paint(value))?,
+            (None, Some(value)) => {
+                writeln!(f, "Bad input '{}'", t.attention.paint(value))?;
+                writeln!(f)?;
+            }
             _ => (),
         }
 
-        write!(f, "{}", self.message)?;
+        writeln!(f, "{}", self.message)?;
```
slarse, line 88: Needed to change this to align with how `clap` formats its errors, they contain a trailing newline.

crates/but/src/lib.rs
```diff
@@ -219,16 +224,40 @@ pub async fn handle_args(args: impl Iterator<Item = OsString>) -> Result<()> {
...
+                .allow_external_subcommands(false)
+                .try_get_matches_from(argv);
+
+            match reparse_result {
+                Err(parse_err) => print_and_exit_non_zero(parse_err),
+                _ => {
+                    // If for some reason we succeeded to parse now, we'll just the original error
+                    print_and_exit_non_zero(CliError::ExternalCommandNotFound(command_name))
+                }
+            }
```
slarse: This is where the new error variant is utilized. I think the enum approach over "codes" is paying off here.

And yeah, the reparsing is a bit of a hack but I honestly don't even hate it :D

davidpdrsn (approved):
LGTM!

crates/but/src/lib.rs
slarse: Note: There may be an issue with colored output here, investigating ...

slarse: Fixed.

PR #13887, opened 2026-05-19
commit_create: no premature `to_selector()`

jonathantanmy2:
A subsequent PR will add a new enum variant to `Step`, which also requires modifications to `parent_commit_id_for_new_commit()` to handle the new enum variant. However, upon closer observation, that function does not need to take a `Step` in the first place. Change it and its caller to take a `RelativeToRef` instead.

### DIFF
--- a/crates/but-workspace/src/commit/commit_create.rs
+++ b/crates/but-workspace/src/commit/commit_create.rs
@@ -3,7 +3,8 @@
 use anyhow::Result;
 use but_core::{DiffSpec, RefMetadata};
 use but_rebase::graph_rebase::{
-    Editor, LookupStep, Pick, Selector, Step, SuccessfulRebase, ToSelector, mutate::InsertSide,
+    Editor, Selector, Step, SuccessfulRebase, ToSelector,
+    mutate::{InsertSide, RelativeToRef},
 };
 
 use crate::commit_engine::{Destination, create_commit};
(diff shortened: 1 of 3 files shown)


crates/but-workspace/src/commit/commit_create.rs
```diff
@@ -48,14 +49,12 @@ pub struct CommitCreateOutcome<'ws, 'meta, M: RefMetadata> {
 pub fn commit_create<'ws, 'meta, M: RefMetadata>(
     mut editor: Editor<'ws, 'meta, M>,
     changes: Vec<DiffSpec>,
-    relative_to: impl ToSelector,
+    relative_to: RelativeToRef<'_>,
```
Caleb-T-Owens, line 52: This was taking an `impl ToSelector` as the input as it allows us to chain multiple operations together. IE:

```
rebase, new_commit_selector = create_commit(editor, relative_to: "a2g123c", side: Above, message: "foobar");
editor = rebase.into_editor();
rebase, second_commit_selector = create_commit(editor, relative_to: new_commit_selector, side: Above, message: "foobar");
```

We wouldn't just want to return a commit id from `create_commit`, because the `.materialize()` might do a second rewrite where the commits get signed, so for an end consumer, we really only want to lookup the real id after we materialize.

---

Do we loose this ability with this change?

jonathantanmy2: Ah, thanks for the context. My PR does make it more cumbersome for the user if they want to chain operations. I'll drop this PR.

PR #13891, opened 2026-05-19
Rename find_segments_reachable_from_a_not_b to find_commits

Caleb-T-Owens:
It finds commits not segments.

Also adds a new function that actually finds segments.


krlvi (approved):
makes sense

PR #13893, opened 2026-05-19
Hydrate LFS pointers after upstream integration

Yeusepe:
## Summary

- hydrate deferred Git LFS pointers after upstream integration completes
- hydrate a workspace path immediately if the Unity conflict flow writes an LFS pointer into the worktree
- add pointer detection and local-object hydration fallback for modified pointer files that `git lfs checkout` leaves untouched

## Root Cause

GitButler intentionally sets `GIT_LFS_SKIP_SMUDGE=1` during upstream integration to avoid expensive implicit LFS hydration. In projects that track Unity scenes/assets with LFS, that can leave raw pointer files in the worktree. Unity then opens those tiny pointer files as invalid or empty scenes.

## Validation

- `cargo check -p but-api -p gitbutler-tauri`
- `cargo test -p but-core lfs::tests::detects_lfs_pointer_content`


Byron:
The motivation behind this PR, is certainly admirable, nonetheless I have to close it based on the contributing guidelines that require an issue to be created first for feedback.

And even if those guidelines did not exist, a PR with twenty three thousand edit lines is completely unreviewable, maybe that saves you time, next time, as well. I certainly wouldn't mind PRs with just below a hundred lines changed that fix a genuine issue, even without raising an issue about this first.

With that said, I believe that by the end of this year worktree filters will be supported.
Thanks!

PR #13894, opened 2026-05-19
EXPERIMENT: Start of super experimental graph views of the workspace

Caleb-T-Owens:
This was some stuff I bashed together as an experiment. It shouldn’t be seen as some decision having been made.

I want to play with renderdag, the library that JJ and sapling use to render their graphs. It might be a good starting point for producing a list of rows that could be consumed by both the CLI and the web frontend

### DIFF
--- a/crates/but-workspace/tests/workspace/experiment.rs
+++ b/crates/but-workspace/tests/workspace/experiment.rs
@@ -1,0 +1,90 @@
+use anyhow::Result;
+use but_graph::init::Options;
+use but_meta::virtual_branches_legacy_types::Target;
+use but_testsupport::visualize_commit_graph_all;
+use but_workspace::experiment::{RowData, render_stacks};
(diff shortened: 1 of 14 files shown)


crates/but-workspace/tests/workspace/experiment.rs
```diff
@@ -0,0 +1,79 @@
...
+    │ ◎  refs/heads/D
+    │ │
+    │ ●  D
+    ├─╯
+    ◎  refs/heads/A
+    │
+    ●  A
+    │
+    ~
+    ");
```
Byron: 🤩🩷👍

The only reason I wouldn't instantly want to use it in all places for debug rendering (with more info in it, of course) is that I'd love to condense it vertically by cutting out empty segments like Git.

In any case, having more ways to visualize graphs is definitely good to have.

PR #13900, opened 2026-05-20
feat(but): add explicit error type to signify bad input from user

slarse:
This PR introduces a `BadInput` struct that's used as an explicit return value in `but` commands. This is the `but` equivalent of a 4xx HTTP response. They are distinct from errors as it's squarely on the user providing bad input in some shape or form, so we want to treat them distinctly from errors.

These can only reasonably be produced at the very top layer of any `but` command, where sufficient information about what exactly is user-provided input is still retained.

The downside is that it does get rather verbose, but it's hard to get around that if we want to provide sufficient information to the user about what they did wrong.

This PR applies `BadInput` to `but teardown` and `but branch delete`. The rest needs to be incrementally migrated and tested - a big bang migration seems like a bad idea here.

Note that `but branch delete` exited zero even on failure due to bad user input - that has also been fixed here and I suspect that the same happens elsewhere.

**See the tests for output examples.**


crates/but/tests/but/command/teardown.rs
```diff
@@ -499,10 +496,9 @@ fn teardown_checkout_to_handles_missing_branch() -> anyhow::Result<()> {
         .failure()
         .stderr_eq(str![
             r#"
-Error: Failed to teardown GitButler project.
+error: bad input for '--checkout-to'
```
slarse: Formatting decision: I went lowercase here to make it align with `clap` errors, but there are other errors that start with uppercase. For example, `Failed to determine checkout target branch. Specify a target branch with --checkout-to <branch>.` makes sense to start uppercase as it's a multi-sentence error.

Any opinions?

crates/but/src/utils/metrics.rs
```diff
@@ -403,16 +403,16 @@ fn posthog_client(app_settings: AppSettings) -> Option<impl Future<Output = post
...
         let Some(OneshotMetricsContext { start, command }) = ctx else {
-            return self.map(|_| ());
+            return self;
         };
 
         let props = Props::from_result(start, &self);
         let Some(v) = command.to_possible_value() else {
             tracing::warn!("BUG: didn't get string value for {command:?}");
-            return self.map(|_| ());
+            return self;
```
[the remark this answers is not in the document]
slarse: @davidpdrsn You and I discussed this briefly, and I recall that we concluded that bad user input should _not_ be emitted as error metrics. It might still be useful to have some other means of seeing if users frequently add bad input.

@mtsgrd Do you have any opinions here?

crates/but/src/lib.rs
```diff
@@ -233,20 +233,28 @@ pub async fn handle_args(args: impl Iterator<Item = OsString>) -> Result<()> {
...
+                    Ok(None)
                 }
             }
         }
         Some(cmd) => match_subcommand(cmd, args, app_settings, out).await,
+    }?;
+
+    if let Some(bad_input) = maybe_bad_input {
+        use std::io::Write;
+        writeln!(std::io::stderr(), "{bad_input}")?;
```
Byron: We bailed before, and I think it should still bail - I don't like the side-channel.
Staying with the analogy: 4XX is still not 200 - 200 is Ok, everything else is Err.
So I think we still have to look at error-handling for this.

Here is the standard version of this which seems to serve the purpose:

#### Text

```
❯ cargo run -p but -- -C ~/dev-gb/deno teardown --checkout-to foo
    Finished [`dev` profile [unoptimized + debuginfo]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 0.26s
     Running `target/debug/but -C /Users/byron/dev-gb/deno teardown --checkout-to foo`
Error: Failed to teardown GitButler project.

Caused by:
    error: bad input for '--checkout-to'

    The reference 'foo' did not exist
```

#### Pretty

Image

```patch
diff --git a/crates/but/src/command/legacy/teardown.rs b/crates/but/src/command/legacy/teardown.rs
index fb99df5620..73a72f1127 100644
--- a/crates/but/src/command/legacy/teardown.rs
+++ b/crates/but/src/command/legacy/teardown.rs
@@ -1,4 +1,4 @@
-use anyhow::Context as _;
+use anyhow::{Context as _, bail};
 use but_core::WORKSPACE_REF_NAME;
 use but_ctx::Context;
 use but_workspace::legacy::StacksFilter;
@@ -56,28 +56,24 @@ pub(crate) fn teardown(
     // Note: Validate checkout_to before snapshot creation to prevent unnecessary snapshot
     let checkout_to = if let Some(checkout_to) = &checkout_to {
         let repo = ctx.repo.get()?;
-        let ref_name: gix::refs::PartialName = if let Ok(ref_name) = checkout_to.clone().try_into()
-        {
-            ref_name
-        } else {
-            return BadInput::new(format!("Invalid ref name: {checkout_to}",))
-                .arg("--checkout-to")
-                .into_result();
-        };
-        let resolved_ref = match repo.try_find_reference(ref_name.as_ref())? {
-            Some(resolved_ref) => resolved_ref,
-            None => {
-                return BadInput::new(format!("The reference '{checkout_to}' did not exist"))
+        let ref_name: gix::refs::PartialName =
+            checkout_to.clone().try_into().with_context(|| {
+                BadInput::new(format!("Invalid ref name: {checkout_to}",)).arg("--checkout-to")
+            })?;
+        let resolved_ref = repo
+            .try_find_reference(ref_name.as_ref())?
+            .with_context(|| {
+                BadInput::new(format!("The reference '{checkout_to}' did not exist"))
                     .arg("--checkout-to")
-                    .into_result();
-            }
-        };
+            })?;
         if !matches!(resolved_ref.name().category(), Some(Category::LocalBranch)) {
-            return BadInput::new(format!(
-                "Invalid ref for checkout: '{checkout_to}' is not a local branch"
-            ))
-            .arg("--checkout-to")
-            .into_result();
+            bail!(
+                "{}",
+                BadInput::new(format!(
+                    "Invalid ref for checkout: '{checkout_to}' is not a local branch"
+                ))
+                .arg("--checkout-to")
+            );
         }
         Some(resolved_ref.name().shorten().to_string())
     } else {
```

</details>

And this additional patch keeps it close to what was originally intended here, to control the error output.

```patch
diff --git a/crates/but/src/lib.rs b/crates/but/src/lib.rs
index 072041423f..a7ab7b4580 100644
--- a/crates/but/src/lib.rs
+++ b/crates/but/src/lib.rs
@@ -183,7 +183,7 @@ pub async fn handle_args(args: impl Iterator<Item = OsString>) -> Result<()> {
 
     let app_settings = AppSettings::load_from_default_path_creating_without_customization()?;
 
-    let maybe_bad_input = match args.cmd.take() {
+    let err = match args.cmd.take() {
         None if args.path.is_some() => {
             // If one argument is provided without a subcommand, check if this is a valid path.
             let maybe_path = args
@@ -238,15 +238,20 @@ pub async fn handle_args(args: impl Iterator<Item = OsString>) -> Result<()> {
             }
         }
         Some(cmd) => match_subcommand(cmd, args, app_settings, out).await,
-    }?;
+    }.err();
 
-    if let Some(bad_input) = maybe_bad_input {
-        use std::io::Write;
-        writeln!(std::io::stderr(), "{bad_input}")?;
-        std::process::exit(1);
+    match err {
+        Some(err) => {
+            if let Some(bad_input) = err.downcast_ref::<BadInput>() {
+                use std::io::Write;
+                writeln!(std::io::stderr(), "{bad_input}")?;
+                std::process::exit(1);
+            } else {
+                Err(err)
+            }
+        }
+        None => Ok(()),
     }
-
-    Ok(())
 }
 
 async fn match_subcommand(
diff --git a/crates/but/src/utils/bad_input.rs b/crates/but/src/utils/bad_input.rs
index c7be689ebd..7ecbca70b3 100644
--- a/crates/but/src/utils/bad_input.rs
+++ b/crates/but/src/utils/bad_input.rs
@@ -8,6 +8,7 @@ use crate::theme::{self, Paint};
 /// that is somehow invalid.
 ///
 /// Think of this as the `but` equivalent of a 4xx HTTP response.
+#[derive(Debug)]
 pub struct BadInput {
     /// A message to print verbatim to the user
     pub message: String,
```

Image

Byron (commented):
Listen to the Rust compiler's whispers, and of course, to Byron yelling across the planet: Don't do it, it's going to lead to terrible eye strain. Instead, please look at an equivalent version of this that is the way the Rust gods intended 😁 (Edit for ambiguity: I am a mere disciple, I of course talk of the one and only David Tolnay) .

✌️

crates/but/src/lib.rs
slarse: @Byron This is going to sound like a complete cop-out, but I don't have particularly strong opinions about this, this is all based on a discussion with @davidpdrsn who preferred separating bad user inputs from errors entirely.

I don't disagree with the eyestrain argument, but for the idiomatic-ness of it all I'll have to defer to @davidpdrsn.

@davidpdrsn would you like to weigh in on the discussion?

crates/but/src/utils/metrics.rs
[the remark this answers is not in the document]
Byron: Me, me 🙋‍♂️ :D.

I didn't think of metrics but I think with this patch applied as a whole (which is combined from [the detailed comment](https://github.com/gitbutlerapp/gitbutler/pull/13900#pullrequestreview-4326834714), and an extra cehck) will work for not emitting metrics as well.

<details><summary>Patch with no metrics on BadInput</summary>
<p>

```patch
commit 58854baed08a5665626e98229cf8971e8558350a
Date:   Wed May 20 17:10:25 2026 +0800

    use anyhow as intended

diff --git a/crates/but/src/command/legacy/teardown.rs b/crates/but/src/command/legacy/teardown.rs
index fb99df5620..73a72f1127 100644
--- a/crates/but/src/command/legacy/teardown.rs
+++ b/crates/but/src/command/legacy/teardown.rs
@@ -1,4 +1,4 @@
-use anyhow::Context as _;
+use anyhow::{Context as _, bail};
 use but_core::WORKSPACE_REF_NAME;
 use but_ctx::Context;
 use but_workspace::legacy::StacksFilter;
@@ -56,28 +56,24 @@ pub(crate) fn teardown(
     // Note: Validate checkout_to before snapshot creation to prevent unnecessary snapshot
     let checkout_to = if let Some(checkout_to) = &checkout_to {
         let repo = ctx.repo.get()?;
-        let ref_name: gix::refs::PartialName = if let Ok(ref_name) = checkout_to.clone().try_into()
-        {
-            ref_name
-        } else {
-            return BadInput::new(format!("Invalid ref name: {checkout_to}",))
-                .arg("--checkout-to")
-                .into_result();
-        };
-        let resolved_ref = match repo.try_find_reference(ref_name.as_ref())? {
-            Some(resolved_ref) => resolved_ref,
-            None => {
-                return BadInput::new(format!("The reference '{checkout_to}' did not exist"))
+        let ref_name: gix::refs::PartialName =
+            checkout_to.clone().try_into().with_context(|| {
+                BadInput::new(format!("Invalid ref name: {checkout_to}",)).arg("--checkout-to")
+            })?;
+        let resolved_ref = repo
+            .try_find_reference(ref_name.as_ref())?
+            .with_context(|| {
+                BadInput::new(format!("The reference '{checkout_to}' did not exist"))
                     .arg("--checkout-to")
-                    .into_result();
-            }
-        };
+            })?;
         if !matches!(resolved_ref.name().category(), Some(Category::LocalBranch)) {
-            return BadInput::new(format!(
-                "Invalid ref for checkout: '{checkout_to}' is not a local branch"
-            ))
-            .arg("--checkout-to")
-            .into_result();
+            bail!(
+                "{}",
+                BadInput::new(format!(
+                    "Invalid ref for checkout: '{checkout_to}' is not a local branch"
+                ))
+                .arg("--checkout-to")
+            );
         }
         Some(resolved_ref.name().shorten().to_string())
     } else {
diff --git a/crates/but/src/lib.rs b/crates/but/src/lib.rs
index 072041423f..a7ab7b4580 100644
--- a/crates/but/src/lib.rs
+++ b/crates/but/src/lib.rs
@@ -183,7 +183,7 @@ pub async fn handle_args(args: impl Iterator<Item = OsString>) -> Result<()> {
 
     let app_settings = AppSettings::load_from_default_path_creating_without_customization()?;
 
-    let maybe_bad_input = match args.cmd.take() {
+    let err = match args.cmd.take() {
         None if args.path.is_some() => {
             // If one argument is provided without a subcommand, check if this is a valid path.
             let maybe_path = args
@@ -238,15 +238,20 @@ pub async fn handle_args(args: impl Iterator<Item = OsString>) -> Result<()> {
             }
         }
         Some(cmd) => match_subcommand(cmd, args, app_settings, out).await,
-    }?;
+    }.err();
 
-    if let Some(bad_input) = maybe_bad_input {
-        use std::io::Write;
-        writeln!(std::io::stderr(), "{bad_input}")?;
-        std::process::exit(1);
+    match err {
+        Some(err) => {
+            if let Some(bad_input) = err.downcast_ref::<BadInput>() {
+                use std::io::Write;
+                writeln!(std::io::stderr(), "{bad_input}")?;
+                std::process::exit(1);
+            } else {
+                Err(err)
+            }
+        }
+        None => Ok(()),
     }
-
-    Ok(())
 }
 
 async fn match_subcommand(
diff --git a/crates/but/src/utils/bad_input.rs b/crates/but/src/utils/bad_input.rs
index c7be689ebd..7ecbca70b3 100644
--- a/crates/but/src/utils/bad_input.rs
+++ b/crates/but/src/utils/bad_input.rs
@@ -8,6 +8,7 @@ use crate::theme::{self, Paint};
 /// that is somehow invalid.
 ///
 /// Think of this as the `but` equivalent of a 4xx HTTP response.
+#[derive(Debug)]
 pub struct BadInput {
     /// A message to print verbatim to the user
     pub message: String,
diff --git a/crates/but/src/utils/metrics.rs b/crates/but/src/utils/metrics.rs
index 154a962b0f..9e93afaa00 100644
--- a/crates/but/src/utils/metrics.rs
+++ b/crates/but/src/utils/metrics.rs
@@ -9,7 +9,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     args::{Subcommands, config, metrics::CommandName},
-    utils::{ResultMetricsExt, binary_path},
+    utils::{ResultMetricsExt, bad_input::BadInput, binary_path},
 };
 
 pub(super) mod types {
@@ -408,6 +408,9 @@ impl<T> ResultMetricsExt<T> for anyhow::Result<T> {
         let Some(OneshotMetricsContext { start, command }) = ctx else {
             return self;
         };
+        if self.as_ref().err().is_some_and(|err| err.is::<BadInput>()) {
+            return self;
+        }
 
         let props = Props::from_result(start, &self);
         let Some(v) = command.to_possible_value() else {
```

</p>
</details>

crates/but/src/lib.rs
slarse: My impression is however that letting `BadInput` be errors is sleeker than having it be an Ok() return value, but I still see _some_ value in a complete separation as it makes it more explicit. I don't love the idea of arbitrarily hiding errors inside `anyhow` results as that feels like sidestepping the type system, but perhaps it's the pragmatic middleground the Rust community has agreed upon.

davidpdrsn: I don't buy the eye strain argument 😜 I really don't think its that bad. This wouldn't impact library functions, just the CLI shell. Ideally all CLI commands should follow a similar pattern:

- Resolve user input such as resolving short codes and checking if branches exist. Here we might return `BadInput`
- Do the actual operation. Here we don't return `BadInput` and just good old `anyhow::Result<_>`

If it were to spread into libraries like `but-api` and `but-workspace` I'd agree that its a bad idea but thats not the intention afaik.

The goal I had in mind when discussing this with Simon was that we could customize whats printed to the terminal depending on whether the error was the users fault or our fault. For things like `but branch delete some-branch-that-doesnt-exist` I don't think anyhow's default error formatting is appropriate and I'd rather have some we explicitly control.

In general I'd rather have more explicit types than having to remember to do `err.downcast_ref::<BadInput>()`. I think anyhow is awesome as long as you don't care about the actual error and just wanna propagate it or add more context. But once you care about the actual error (as we do here) I think its the wrong abstraction.

Another approach would be a new error enum that has one variant for each case we care about:

```rust
#[derive(Debug)]
enum CliError {
    BadInput(String),
    Internal(anyhow::Error),
}

impl From<anyhow::Error> for CliError {
    fn from(err: anyhow::Error) -> Self {
        Self::Internal(err)
    }
}
```

which would still use explicit types and work with `?`:

```rust
fn cli_handler() -> Result<(), CliError> {
    some_library_function()?; // still works with `?`

    ...
}

fn another_cli_handler(branch: &str) -> Result<(), CliError> {
    // can flag bad input by returning explicit error
    if !branch_exists(branch) {
        return Err(CliError::BadInput(format!("{branch} not not found")));
    }

    ...
}

fn some_library_function() -> anyhow::Result<()> { ... }
```

Byron: Thanks for the call!
Let's go with the custom error type , I think this example really hits the nail on the head.

````
fn cli_handler() -> Result<(), CliError> {
    some_library_function()?; // still works with `?`

    ...
}

fn another_cli_handler(branch: &str) -> Result<(), CliError> {
    // can flag bad input by returning explicit error
    if !branch_exists(branch) {
        return Err(CliError::BadInput(format!("{branch} not not found")));
    }

    ...
}

fn some_library_function() -> anyhow::Result<()> { ... }
````

While the example completely drops the underlying error, that's also the intent here so it's a good fit.

File: crates/but/tests/but/command/teardown.rs [hunk 4]
```diff
@@ -566,10 +559,9 @@ fn teardown_checkout_to_disallows_non_local_branch_ref() -> anyhow::Result<()> {
         .assert()
         .failure()
         .stderr_eq(str![[r#"
-Error: Failed to teardown GitButler project.
+Error: Bad input for '--checkout-to'
```
slarse, line 562: This kind of formatting clashes with clap's input validation, as it would output something like `error: blablabla '--checkout-to'`. But most of our errors are formatted like this and I don't want to run through and change everything right now.

crates/but/src/error.rs
```diff
@@ -0,0 +1,102 @@
...
+    pub fn context<C>(self, context: C) -> Self
+    where
+        C: Display + Send + Sync + 'static,
+    {
+        match self {
+            Self::BadInput(value) => Self::BadInput(value),
+            Self::Internal(value) => Self::Internal(value.context(context)),
+        }
+    }
+}
```
slarse, line 83: This is a bit awkward, we just drop the added context if it's a "bad input" result. Which is what we want to do semantically, but it's very obtuse from the call site. I couldn't find a better way to allow to add context at the top level that doesn't get overly verbose.

slarse (commented):
So, here are the enum variants. It's not really less verbose and required rewriting a lot more, but it does do the job.

crates/but/src/utils/metrics.rs
```diff
@@ -403,16 +403,19 @@ fn posthog_client(app_settings: AppSettings) -> Option<impl Future<Output = post
...
```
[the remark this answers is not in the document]
slarse: Intentional for now, want to discuss metrics with Mattias

crates/but/src/utils/metrics.rs
```diff
@@ -426,6 +429,6 @@ impl<T> ResultMetricsExt for anyhow::Result<T> {
             .group()
             .kill_on_drop(false)
             .spawn()?;
-        self.map(|_| ())
+        self
```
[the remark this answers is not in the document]
slarse: Fair but preexisting, will fix separately

crates/but/src/error.rs
Byron: In the previous `anyhow` version, the `context` would be kept, which could lead to `BadInput` to not be the tail of the chain. In the end, it didn't matter as it was pulled out of there anyway.

So I guess this implementation is just 'honest', and in future could even decide to keep the context to present it in some way if we think something gets lost.

crates/but/src/error.rs
```diff
@@ -0,0 +1,102 @@
...
+
+impl<E> From<E> for CliError
+where
+    E: Into<anyhow::Error>,
+{
+    fn from(value: E) -> Self {
+        Self::Internal(value.into())
+    }
+}
```
slarse, line 70: Kind of shady blanket conversion but without this every single `?` in top level commands became an absolute bother.

davidpdrsn: I don't think this is shady 😆 Its a pretty standard practice in Rust when wrapping errors I'd say.

davidpdrsn (approved):
LGTM!

crates/but/src/error.rs
slarse: Ah. I had come to some form of understanding that very broad conversions aren't good practice, and this conversion is pretty darn broad!

PR #13904, opened 2026-05-20
Use `head_info` rather than `stacks` or `stack_details` when possible

Caleb-T-Owens:
We still need some usage of `stacks` for the branch listing page AFAIK, but using `head_info` for the main projection is a big help since it avoids the issues with some strange inconsistency issues between `stack_details` and `stacks`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3957,6 +3957,7 @@ dependencies = [
  "gitbutler-workspace",
  "gix",
  "glob",
+ "insta",
  "itertools",
  "md5",
  "pretty_assertions",
--- a/apps/desktop/src/components/branch/BranchCard.svelte
+++ b/apps/desktop/src/components/branch/BranchCard.svelte
@@ -207,6 +207,7 @@
 				onclick={args.disableClick ? undefined : args.onclick}
 				disableClick={args.disableClick}
 				menu={args.menu}
+				buttons={args.buttons}
 				conflicts={args.isConflicted}
 				{showPrCreation}
 				changedFiles={args.changedFiles}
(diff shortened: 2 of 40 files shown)


krlvi (approved):
awesome!

slarse:
@Caleb-T-Owens Smoke tests failed with

```
Error: database is locked

Caused by:
    Error code 5: database is locked
```

That's _not_ related to the code in this PR (see [this comment](https://github.com/gitbutlerapp/gitbutler/blob/fa82a9f1d6fd064930309ba74e0e492f327bd3e4/.github/workflows/push.yaml#L74-L82)) and probably a spurious error, but it's still worrying. My best guess is that it's the background sync interfering with the next command being executed. `but` is not great for scripting at the moment.

Byron:
I think there is a way to turn backtraces on with `anyhow`, so it can then show the origin of the error. Maybe RUST_BACKTRACE=full` would do the trick? Maybe there is more to it as well, might need some research.

Then once it's clear where this is happening, it can be fixed.
What might be happening here is that something doesn't obtain a repository RW-lock, so the database acts as sync mechanism, and slow-running opts take too long and trigger a busy timeout. Our  own repository lock has no timeout.
But… that's just a guess.

Caleb-T-Owens:
I've come into the realisation that my mapping layer to the old datastructures introduces a potential regression.

If you have a branch segment that doesn't have a reference, it fails hard.

Anon segments really need to be a first class supported entity, so I'm switching the frontend over to using the new datastructures which will allow us to properly support the scenario.

krlvi (approved):

Caleb-T-Owens:
The entity adapter problem is real, and needs addressing, but doing so is no small feat of refactoring.

As such, this PR is going to be landed acknowledging that this _is_ an active problem. We're not planning a new release soon so there is time for follow up refactors.

PR #13905, opened 2026-05-20
Head info handles gerrit metadata

Caleb-T-Owens:
This brings the old gerrit-mode logic that previously was only present on the super legacy `stacks_details` function, and brings in into the `RefInfo` construction.

I realise that this is the first time bringing the database into the `but-workspace` crate, but in this instance, it felt reasonable to provide a specific `GerritMetadataHandle` directly rather than adding more abstraction layers over it.


krlvi (approved):
thanks!

crates/but-workspace/src/legacy/stacks.rs
```diff
@@ -304,11 +305,15 @@ pub fn stack_details_v3(
...
             .find(|stack| stack.id == Some(stack_id))
     }
-    let mut ref_info_options = ref_info::Options {
-        // TODO(perf): make this so it can be enabled for a specific stack-id.
-        expensive_commit_info: true,
-        traversal: but_graph::init::Options::limited(),
-    };
+    fn new_ref_info_options() -> ref_info::Options<'static> {
+        ref_info::Options {
+            // TODO(perf): make this so it can be enabled for a specific stack-id.
```
Byron: I think this TODO can be removed by now, as the whole "expensive operations" problem will benefit for a more general solution.

crates/but-workspace/src/ref_info.rs
```diff
@@ -222,15 +227,48 @@ impl WorkspaceExt for but_graph::Workspace {
...
+#[derive(Default, Debug)]
+pub struct Options<'db> {
     /// Control how to traverse the commit-graph as the basis for the workspace conversion.
     pub traversal: but_graph::init::Options,
     /// Perform expensive computations on a per-commit basis.
     ///
     /// Note that less expensive checks are still performed.
     pub expensive_commit_info: bool,
+    /// Configure whether Gerrit metadata should augment the standard graph-derived result.
+    pub gerrit_mode: GerritMode<'db>,
```
Byron, line 271: I think this can work and even serve as a good pattern for optional data sources.

The alternative would be to pass the whole `db` as part of the standard constructor, and make this a boolean, which seems more invasive than it has to be.

However, if I were to predict the future, this is exactly what's going to happen as `db.cache` will be used provide 'cheap' data from the cache right away, and that data is varied enough to not want to split it up into individual handles.

If you buy into that future, pass in `db` to the constructor, if not, this is perfect to show how scoped this data is right now.

And if I were to follow my own rules, then "don't try to predict the future" 😁.

Byron (approved):
I think this is great, particularly the added docs.

Overall, having the `db` in involved in the RefInfo construction is inevitable.
Once `meta` is part of `db`, what *could* happen is that `db` is passed directly and it can get the handles it needs internally for convenience. Technically it's better to only pass what's actually used, so maybe that's the route we take instead, and it's a route now established by this PR. That's a good thing!

crates/but-workspace/src/ref_info.rs
Caleb-T-Owens: This is the same decision tree I was going through :joy:

I'm happy to start with this because it's just a boilerplate change that a clanker can make for us if/when this future comes into reality

crates/but-workspace/src/legacy/stacks.rs
Caleb-T-Owens: Cool! I'll snip it out.

crates/but-workspace/src/ref_info.rs
```diff
@@ -518,9 +556,90 @@ pub fn graph_to_ref_info(
...
+    ) -> anyhow::Result<()> {
+        for segment in self
+            .stacks
+            .iter_mut()
+            .flat_map(|stack| stack.segments.iter_mut())
+        {
+            for commit in &mut segment.commits {
+                let Some(meta) = metadata.get(&commit.change_id().to_string())? else {
+                    continue;
+                };
```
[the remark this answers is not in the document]
Caleb-T-Owens: This is nonsensical. Sqlite is fast with repeated queries, so unless this can be proven to really hurt performance, we shouldn't worry about it.

Especially since gerrit workflows won't have loads and loads of these commits applied at once

crates/but-workspace/tests/workspace/ref_info/with_workspace_commit/mod.rs
```diff
@@ -24,6 +24,101 @@ pub fn ref_info(
     but_workspace::ref_info(existing_ref, meta, opts)
 }
 
+#[test]
+fn gerrit_mode_uses_metadata_for_commit_review_urls_and_push_status() -> anyhow::Result<()> {
+    let (repo, mut meta) = read_only_in_memory_scenario("remote-advanced-ff")?;
+    add_stack(&mut meta, 1, "A", StackState::InWorkspace);
+    let base_info = head_info(&repo, &meta, standard_options())?;
+    let local_commit = base_info.stacks[0].segments[0].commits[0].clone();
+    let change_id = local_commit.change_id().to_string();
```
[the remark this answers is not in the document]
Caleb-T-Owens: This is in a test, it doesn't really matter

crates/but-workspace/tests/workspace/ref_info/mod.rs
```diff
@@ -857,10 +857,11 @@ mod utils {
         Ok((tmp, repo, meta))
     }
 
-    pub fn standard_options() -> but_workspace::ref_info::Options {
+    pub fn standard_options() -> but_workspace::ref_info::Options<'static> {
         ref_info::Options {
             expensive_commit_info: true,
             traversal: Default::default(),
+            ..Default::default()
         }
```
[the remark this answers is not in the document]
Caleb-T-Owens: same, this is a test context.

PR #13927, opened 2026-05-22
fix(but): short term fixes for `but commit empty`

slarse:
Makes some short term fixes to `but commit empty` to prevent it from putting the workspace in a bad state by adding a commit above a stack head. See the individual commits for details.

The long term plan is to move `but commit empty` as a flag `but commit --empty`, as most of the arguments to `but commit` make sense for `but commit empty`, and all arguments for `but commit empty` make sense for `but commit`.

### Better error messages side quest
I'm continuing down the path of improving user feedback on errors. Here, I've added a `hint` to the `BadInput` struct that gets displayed at the bottom of the error, with the intention to guide the user to correct usage.

error_with_hint

### A note on testing of `git commit empty`
The tests of `git commit empty` largely only assert that a commit is added _somewhere_, using `git log` to count commits. This is very weak, as the placement of the commits is important. For example, the variations that add commits above stack heads aren't even detected with the current testing methodology.

I've strengthened a few of the tests by adding before/after status checks, and I think that we largely should go down that route for testing. It's verbose, but maintaining the snapshots is relatively easy using `SNAPSHOTS=overwrite` and the status output is a fairly good indication that the workspace is in a good state. Much more so than using Git commands.

### DIFF
--- a/crates/but/src/command/legacy/commit.rs
+++ b/crates/but/src/command/legacy/commit.rs
@@ -86,9 +86,9 @@ pub(crate) fn insert_blank_commit(
                 guard.write_permission(),
             )?;
             let success_message = match insert_side {
-                InsertSide::Above => format!("Created blank commit at the top of stack '{name}'"),
+                InsertSide::Above => format!("Created blank commit above branch '{name}'"),
                 InsertSide::Below => {
-                    format!("Created blank commit at the bottom of stack '{name}'")
+                    format!("Created blank commit at the tip of branch '{name}'")
                 }
             };
             (outcome, success_message)
--- a/crates/but/src/command/legacy/commit.rs
+++ b/crates/but/src/command/legacy/commit.rs
@@ -13,7 +13,7 @@ use gitbutler_repo::hooks;
 
 use super::{ShowDiffInEditor, estimate_diff_blob_size};
 use crate::{
-    CliId, IdMap,
+    BadInput, CliId, IdMap,
     command::legacy::status::assignment::{CLIHunkAssignment, FileAssignment},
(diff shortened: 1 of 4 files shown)


crates/but/src/command/legacy/commit.rs
```diff
@@ -78,6 +78,25 @@ pub(crate) fn insert_blank_commit(
...
+                    if let Some(stack_head_ref) = stack.ref_name()
+                        && stack_head_ref == &reference.name
+                    {
+                        return BadInput::new("Cannot insert empty commit above stack head")
+                            .arg("--after")
+                            .hint("Use '--before' to insert at the tip of the stack")
+                            .into_cli_result();
+                    }
+                }
+            }
```
slarse, line 98: Is there a better way to determine that a ref points to a stack head? This feels a bit heavy-handed.

I was also considering checking if the branch references a commit that is a direct ancestor of the workspace commit. I'm not sure which approach is more sound.

Caleb-T-Owens: You are right. A reference is a stack head if the commit it points to is a parent of the workspace commit.

Computing the `head_info` isn't as heavy handed as it seems on the surface however.

We cache the creation of the internal `Workspace` projection which inserting the blank commit would do anyway, so it's not _that_ bad.

Caleb-T-Owens (approved):

crates/but/src/command/legacy/commit.rs
slarse: Right, that makes sense. Thanks for the sanity check.

PR #13929, opened 2026-05-22
breaking(but): make name validation in `but branch new` strict

slarse:
This solves several problems in `but branch new`:

1. The command would previously normalize branch names automatically, which is unexpected for a CLI. If you type `but branch new 'my branch'` and get a success back, you expect there to be a branch called 'my branch'. But the actual name of the new branch was normalized to 'my-branch' (although the output would lie and say it created `my branch`). The new behavior is to reject any non-normalized name and print hint with the normalized name instead.
2. Branch creation was idempotent for branches already applied in the workspace, which was very confusing as the CLI claimed to have successfully created a new branch while in fact it had done nothing. Such branch creation is now clearly rejected, which also falls in line with how Git handles such a situation.
3. Various obtuse error messages have been improved here. For example, when creating a new branch with a name that already existed and anchoring it to anywhere, you'd get an error like this::

```
Error: The reference "refs/heads/<name>" should have content b54c2567115498ffcb87b31a4f1f95f0bead2546, actual content was 6078e6bb61d3b1ac4f35f57f05e2cce3382359ff
```

The new error just says that the branch already exists.

> The rest of this PR body is more of a writeup than anything else, but I've done a great deal of thinking about how to provide good user feedback over the past few days and I think this is a good context to share it in.

## Better error messages side quest
Continuing down the path of better error messages to help users out, I've introduced the ability to specify argument name and argument value separately. This allows us to have a standardized way of saying "hey, you done goofed on this argument". I'm not yet convinced exactly when and how to use all this, but it's a start.

For example, when the input is malformed (i.e. the equivalent of a 400 Bad Request HTTP response), I think it's important to clearly connect the dots between the input argument and the badness and try to help the user (or agent) rectify the issue, like this:

invalid_branch_name

> **Note:** Whether its helpful to name the positional argument in the output is something I'm still having an internal debate about. It can certainly be helpful if there are multiple positionals, but for just one it's rather questionable.
>
> The formatting here can probably also be improved, but it's a start, and the point is mostly to standardize what the errors look like at this point.

However, when the branch name is valid but the state of the repository doesn't allow it to be created (i.e. the equivalent of a 409 Conflict HTTP response), I think we can be much more concise with the amount of information in the output, like so:

existing

## Look before you leap
This error handling is LBYL-based, rather than "better to ask for forgiveness than permission" using the new `PreconditionFailed` code. To a large extent I think we'll need to do LBYL-checks for _really_ solid error messages. Here are a few reasons.

### Not enough information from the underlying implementation
We simply don't have enough error information from the internals to directly correlate the error to the user's input. Even if the response code is `PreconditionFailed`, we don't know _what_ precondition. How can we tell if it's the name that's bad, or if it's the anchor?

### `PreconditionFailed` does not imply user error
It implies _caller error_. A trivial example in this piece of code is that we create a canned branch name if the user doesn't pass a branch name. If _that_ fails a precondition, it's still a system error, not a user error. While it's unlikely that we'd create an invalid branch name, it is logically possible and only meant to illustrate the fallacy of assuming that a failing precondition is the user's fault.

What's worse is that a `PreconditionFailed` code only directly implies caller error to the _direct_ caller, as the function that emits the code can only know about its direct inputs.  In fact, as long as it's possible that we generate or manipulate input that can cause a `PreconditionFailed`, the code itself doesn't really add much information to the application as a whole. A caller can never know if the code was emitted from the directly called function, or from a function further down where the inputs aren't necessarily directly controlled by the original caller.

To summarize, it is not _sound_ to assume that `PreconditionFailed` implies user error, or even caller error more than one stack frame away.

### `PreconditionFailed` is a moving target
Right now, `PreconditionFailed` can only be emitted when the branch name is bad. If you build the code around that assumption, adding another `PreconditionFailed` cause in the internals breaks that assumption without there being any mechanism to find out from the call site.

### There are downsides to LBYL, too
This is not to say that LBYL error handling is _perfect_.

* It adds overhead - because the checks should absolutely be performed again by the internals.
* It may be too strict and disallow use that should be allowed
* It adds more code to maintain

But to provide that terrific user feedback, I firmly believe at this point that we need to do input validation immediately, despite these drawbacks. We should also not attempt to do _perfect_ input validation - if we can provide great user feedback for most invalid inputs, and produce obtuse errors for some, that's still a win.

## Further notes
* This replaces https://github.com/gitbutlerapp/gitbutler/pull/13159
* The same name validation should be applied to renaming branches with `but reword`, and will thereby replace https://github.com/gitbutlerapp/gitbutler/pull/13160
* There's still a quirk I don't like: `but branch new refs/heads/my-branch` creates the branch `my-branch`. This is inconsistent with Git, which would literally create the branch `refs/heads/my-branch` (i.e. with full name `refs/heads/refs/heads/my-branch`).

### DIFF
--- a/crates/but/src/command/legacy/branch/mod.rs
+++ b/crates/but/src/command/legacy/branch/mod.rs
@@ -1,4 +1,5 @@
 use anyhow::bail;
+use bstr::BStr;
 use but_api::json::HexHash;
 use but_core::{ref_metadata::StackId, sync::RepoExclusive};
 
@@ -52,15 +53,19 @@ pub fn new(
     out: &mut OutputChannel,
     branch_name: Option<String>,
     anchor: Option<String>,
-) -> Result<(), anyhow::Error> {
+) -> crate::CliResult<()> {
     let t = theme::get();
 
     let mut guard = ctx.exclusive_worktree_access();
     let id_map = IdMap::new_from_context(ctx, None, guard.read_permission())?;
     // Get branch name or use canned name
-    let branch_name = branch_name
-        .map(Ok)
-        .unwrap_or_else(|| but_api::legacy::workspace::canned_branch_name(ctx))?;
+    let branch_name = if let Some(branch_name) = branch_name {
+        let repo = ctx.repo.get()?;
+        check_can_create_branch_with_user_provided_name(&repo, &branch_name)?;
+        branch_name
+    } else {
+        but_api::legacy::workspace::canned_branch_name(ctx)?
+    };
 
     // Store anchor string for JSON output
     let anchor_for_json = anchor.clone();
@@ -71,12 +76,13 @@ pub fn new(
         // Resolve the anchor string to a CliId
         let anchor_ids = id_map.parse_using_context(&anchor_str, ctx)?;
         if anchor_ids.is_empty() {
-            return Err(anyhow::anyhow!("Could not find anchor: {anchor_str}"));
+            return BadInput::new(format!("Could not find anchor: {anchor_str}")).into_cli_result();
         }
         if anchor_ids.len() > 1 {
-            return Err(anyhow::anyhow!(
+            return BadInput::new(format!(
                 "Ambiguous anchor '{anchor_str}', matches multiple items"
-            ));
+            ))
+            .into_cli_result();
         }
         let anchor_id = &anchor_ids[0];
 
@@ -96,10 +102,11 @@ pub fn new(
                 },
             ),
             _ => {
-                return Err(anyhow::anyhow!(
+                return BadInput::new(format!(
                     "Invalid anchor type: {}, expected commit or branch",
                     anchor_id.kind_for_humans()
-                ));
+                ))
+                .into_cli_result();
             }
         }
     } else {
@@ -154,6 +161,51 @@ pub fn new(
         };
         out.write_value(value)?;
     }
+    Ok(())
+}
+
+/// Validate that `user_provided_branch_name` is a valid branch name that does not already exist.
+///
+/// Unlike the GUI, we don't normalize branch names for users in the CLI, as this could lead to
+/// unexpected behavior in scripts. This function rejects names that are possible to normalize.
+fn check_can_create_branch_with_user_provided_name(
+    repo: &gix::Repository,
+    user_provided_branch_name: &str,
+) -> Result<(), CliError> {
+    let normalized =
+        but_core::branch::normalize_short_name(user_provided_branch_name).map_err(|err| {
+            CliError::from(
+                BadInput::new(format!("Invalid branch name: {err}"))
+                    .arg_name("<BRANCH_NAME>")
+                    .arg_value(user_provided_branch_name),
+            )
+        })?;
+
+    let user_name_bstr: &BStr = user_provided_branch_name.into();
+    if normalized != user_name_bstr {
+        return BadInput::new("Invalid branch name")
+            .arg_name("<BRANCH_NAME>")
+            .arg_value(user_provided_branch_name)
+            .hint(format!("Try '{normalized}' instead"))
+            .into_cli_result();
+    }
+
+    let branch_ref_name = if user_provided_branch_name.starts_with("refs/heads") {
+        user_provided_branch_name.to_string()
+    } else {
+        format!("refs/heads/{user_provided_branch_name}")
+    };
+
+    if repo
+        .try_find_reference(&branch_ref_name.to_owned())?
+        .is_some()
+    {
+        return BadInput::new(format!(
+            "A branch named '{user_provided_branch_name}' already exists"
+        ))
+        .into_cli_result();
+    }
+
     Ok(())
 }
 
--- a/crates/but/src/command/legacy/branch/mod.rs
+++ b/crates/but/src/command/legacy/branch/mod.rs
@@ -4,7 +4,7 @@ use but_api::json::HexHash;
 use but_core::{ref_metadata::StackId, sync::RepoExclusive};
 
 use crate::{
-    BadInput, CliError, CliId, CliResult, IdMap,
+    CliError, CliId, CliResult, IdMap, bad_input,
     theme::{self, Paint},
     utils::{Confirm, ConfirmDefault, OutputChannel, shorten_object_id},
 };
@@ -45,7 +45,7 @@ pub fn delete(
         }
     }
 
-    BadInput::new(format!("Branch '{branch_name}' not found in any stack")).into_cli_result()
+    Err(bad_input(format!("Branch '{branch_name}' not found in any stack")).into())
 }
 
 pub fn new(
@@ -76,13 +76,13 @@ pub fn new(
         // Resolve the anchor string to a CliId
         let anchor_ids = id_map.parse_using_context(&anchor_str, ctx)?;
         if anchor_ids.is_empty() {
-            return BadInput::new(format!("Could not find anchor: {anchor_str}")).into_cli_result();
+            return Err(bad_input(format!("Could not find anchor: {anchor_str}")).into());
         }
         if anchor_ids.len() > 1 {
-            return BadInput::new(format!(
+            return Err(bad_input(format!(
                 "Ambiguous anchor '{anchor_str}', matches multiple items"
             ))
-            .into_cli_result();
+            .into());
         }
         let anchor_id = &anchor_ids[0];
 
@@ -102,11 +102,11 @@ pub fn new(
                 },
             ),
             _ => {
-                return BadInput::new(format!(
+                return Err(bad_input(format!(
                     "Invalid anchor type: {}, expected commit or branch",
                     anchor_id.kind_for_humans()
                 ))
-                .into_cli_result();
+                .into());
             }
         }
     } else {
@@ -175,7 +175,7 @@ fn check_can_create_branch_with_user_provided_name(
     let normalized =
         but_core::branch::normalize_short_name(user_provided_branch_name).map_err(|err| {
             CliError::from(
-                BadInput::new(format!("Invalid branch name: {err}"))
+                bad_input(format!("Invalid branch name: {err}"))
                     .arg_name("<BRANCH_NAME>")
                     .arg_value(user_provided_branch_name),
             )
@@ -183,11 +183,11 @@ fn check_can_create_branch_with_user_provided_name(
 
     let user_name_bstr: &BStr = user_provided_branch_name.into();
     if normalized != user_name_bstr {
-        return BadInput::new("Invalid branch name")
+        return Err(bad_input("Invalid branch name")
             .arg_name("<BRANCH_NAME>")
             .arg_value(user_provided_branch_name)
             .hint(format!("Try '{normalized}' instead"))
-            .into_cli_result();
+            .into());
     }
 
     let branch_ref_name = if user_provided_branch_name.starts_with("refs/heads") {
@@ -200,10 +200,10 @@ fn check_can_create_branch_with_user_provided_name(
         .try_find_reference(&branch_ref_name.to_owned())?
         .is_some()
     {
-        return BadInput::new(format!(
+        return Err(bad_input(format!(
             "A branch named '{user_provided_branch_name}' already exists"
         ))
-        .into_cli_result();
+        .into());
     }
 
     Ok(())
--- a/crates/but/src/command/legacy/commit.rs
+++ b/crates/but/src/command/legacy/commit.rs
@@ -90,7 +90,7 @@ pub(crate) fn insert_blank_commit(
                         && stack_head_ref == &reference.name
                     {
                         return BadInput::new("Cannot insert empty commit above stack head")
-                            .arg("--after")
+                            .arg_name("--after")
                             .hint("Use '--before' to insert at the tip of the stack")
                             .into_cli_result();
                     }
--- a/crates/but/src/command/legacy/commit.rs
+++ b/crates/but/src/command/legacy/commit.rs
@@ -13,7 +13,7 @@ use gitbutler_repo::hooks;
 
 use super::{ShowDiffInEditor, estimate_diff_blob_size};
 use crate::{
-    BadInput, CliId, IdMap,
+    CliId, IdMap, bad_input,
     command::legacy::status::assignment::{CLIHunkAssignment, FileAssignment},
     theme::{self, Paint},
     tui,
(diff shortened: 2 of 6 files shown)


crates/but/src/command/legacy/branch/mod.rs
```diff
@@ -157,6 +163,53 @@ pub fn new(
...
+            )
+        })?;
+
+    let user_name_bstr: &BStr = user_provided_branch_name.into();
+    if normalized != user_name_bstr {
+        return BadInput::new("Invalid branch name")
+            .arg_name("<BRANCH_NAME>")
+            .arg_value(user_provided_branch_name)
+            .hint(format!("Try '{normalized}' instead"))
+            .into_cli_result();
```
Byron: I am still choking on this a little, but wouldn't if it was something like this:

```suggestion
        return Err(BadInput::new("Invalid branch name")
            .arg_name("<BRANCH_NAME>")
            .arg_value(user_provided_branch_name)
            .hint(format!("Try '{normalized}' instead"))
            .into());
```

I am pretty sure that this wouldn't format nicely, so maybe it's just something to get used to.
What might be nicer is a macro akin to `bail!()`, but more like this (I don't know if this is possible, but…)

```suggestion
        bad_input!("Invalid branch name", <BRANCH_NAME>  = user_provided_branch_name, "Try '{normalized}' instead");
```

Byron: And if you want to beef up the CLI as much as it seems like, I think it's worth investing into ergonomics.

crates/but/src/command/legacy/branch/mod.rs
```diff
@@ -157,6 +163,53 @@ pub fn new(
     Ok(())
 }
 
+/// Validate that `user_provided_branch_name` is a valid branch name that does not already exist.
+///
+/// Unlike the GUI, we don't normalize branch names for users in the CLI, as this could lead to
+/// unexpected behavior in scripts. This function rejects names that are possible to normalize.
+fn check_can_create_branch_with_user_provided_name(
+    ctx: &but_ctx::Context,
```
Byron: Avoid passing `ctx` down to utilities, it's an antipattern. The god object should stay on top, with specifics passed down. Here `repo` is enough, and it's much clearer that way as well.

Byron (commented):
That's a great initiative and I think it will make the CLI much better. I wasn't aware of LBYL, everything seems to have cool acronyms :D.

PS: This message is ignorant of the PR content itself, what's below is just some things that stood out to me.

crates/but/src/command/legacy/branch/mod.rs
slarse: > And if you want to beef up the CLI as much as it seems like, I think it's worth investing into ergonomics.

I think a nice macro should absolutely be doable. I'll give that a go.

crates/but/src/command/legacy/branch/mod.rs
slarse: Good spot. This is technically possible to detect statically (i.e. passing in `Context` but only using a single top-level field on it), I wonder if we should invest a little bit of time in doing that. It seems like the instructions in `crates/AGENTS.md` aren't sufficient for AI code review to catch this kind of thing. Or maybe there needs to be an explicit instruction to review code based on the architecture requirements in that file. I'm not sure.

Policing will only go so far. This wasn't even me not _knowing_ about this being an anti-pattern, it was just that I failed to spot it after the code had mutated a lot of times.


Did you just spot it or did an agent spot it?

Byron: > Did you just spot it or did an agent spot it?

Loving it that you aren't saying "your agent" :D. Reminds me of this scene in Fightclub.

No, this one stung itself into my fleshy eyes, and I agree, agent.md could probably be clearer about this.
I wouldn't fret too much about it anymore, because:

- lot's of code all around still does it and
- it's often practical and not a problem if `perm` is also passed and
- we can easily debug deadlocks now if code does abuse the power of the `ctx` for guard acquisition

Making agents.md better is probably a good first step, but I don't expect agents to follow it - they prefer copying from the environment for all I can tell.
So one will probably have to go in and fix it up everywhere, all at once, over a weekend or so. But for now I'd not fret it beyond keeping an eye out for that.

slarse: I see, I'll stay vigilant for now.

crates/but/src/command/legacy/commit.rs
```diff
@@ -33,15 +33,14 @@ pub(crate) fn insert_blank_commit(
     let cli_ids = id_map.parse_using_context(target, ctx)?;
 
     if cli_ids.is_empty() {
-        return BadInput::new(format!("Target '{target}' not found")).into_cli_result();
+        bail_bad_input!("Target '{target}' not found");
```
slarse: In simple cases like these where `arg` and `hint` are not used, it just seems strictly better. We already have `bail!()` all over, so `bail_bad_input!()` should be intuitive if restricted to the first format string.

crates/but/src/command/legacy/teardown.rs
```diff
@@ -51,31 +51,24 @@ pub(crate) fn teardown(
...
+            .clone()
+            .try_into()
+            .map_err(|_| bad_input!("Invalid ref name"; arg checkout_to = checkout_to))?;
         let resolved_ref = match repo.try_find_reference(ref_name.as_ref())? {
             Some(resolved_ref) => resolved_ref,
             None => {
-                return BadInput::new(format!("The reference '{checkout_to}' did not exist"))
-                    .arg_name("--checkout-to")
-                    .into_cli_result();
+                bail_bad_input!("Reference does not exist"; arg checkout_to = checkout_to);
```
slarse: Example of argument name and value. Clearly less verbose than the prior variant, but you do need to learn some new syntax.

crates/but/src/command/legacy/commit.rs
```diff
@@ -89,10 +88,11 @@ pub(crate) fn insert_blank_commit(
...
                     {
-                        return BadInput::new("Cannot insert empty commit above stack head")
-                            .arg_name("--after")
-                            .hint("Use '--before' to insert at the tip of the stack")
-                            .into_cli_result();
+                        bail_bad_input!(
+                            "Cannot insert empty commit above stack head";
+                            arg after;
+                            hint "Use '--before' to insert at the tip of the stack",
+                        );
```
slarse: Example of very plain usage: Message and hint with no formatters and argument name without value.

Here, verbosity is a bit of a tossup and there's still the new syntax for `arg` and `hint`.

crates/but/src/command/legacy/branch/mod.rs
```diff
@@ -172,22 +168,13 @@ fn check_can_create_branch_with_user_provided_name(
...
+    )?;
 
     let user_name_bstr: &BStr = user_provided_branch_name.into();
     if normalized != user_name_bstr {
-        return BadInput::new("Invalid branch name")
-            .arg_name("<BRANCH_NAME>")
-            .arg_value(user_provided_branch_name)
-            .hint(format!("Try '{normalized}' instead"))
-            .into_cli_result();
+        bail_bad_input!("Invalid branch name"; arg _ = user_provided_branch_name; hint "Try '{normalized}' instead");
```
slarse: Example of argument value but no name (seems clearer for a single positional).

crates/but/src/command/legacy/branch/mod.rs
slarse: I spent a good few hours just toying around with macros and getting something down that feels ergonomic to me. I'm very curious what you think about it, because while I do think it serves the purpose of providing better DX, it _also_ falls into the trap of requiring the dev to "learn something new", which you clearly wanted to avoid at the onset.

I personally prefer the new macros, but they're also custom made for _me_, so in the context of other devs using this I think it may be a bit of a tossup compared to the previous solution. There's probably a middle ground where the macros are less powerful than this but still provide most of the functionality.

I've added a few comments for you here to show examples (https://github.com/gitbutlerapp/gitbutler/pull/13929/changes/BASE..6a80812a203185673708c7755ef0308f0dfc5e59), and then there's a pretty comprehensive doc comment for the `bad_input` macro.

krlvi (approved):

crates/but/src/command/legacy/teardown.rs
Byron: I also have seen people using the `x => b` syntax for assignments, or even `x = b`, in case that's easier to match on. I say this because I have never seen (or remember) an `arg` prefix.

crates/but/src/command/legacy/commit.rs
Byron: Here there isn't much of a perceived advantage even though it still seems to be less noisy.

`arg after;` threw me off because I expected `arg "after"`.

I wonder if there can be one syntax for argument assignments.

```Rust
   bail_bad_input!(
                            "Cannot insert empty commit above stack head";
                            arg = "after",
                            hint = "Use '--before' to insert at the tip of the stack",
                        );
```

But now that I have written it, I don't like it quite as much as the plain version. What I found unusual is the `;` as I expected `,`. This is done in `tracing::info!` for instance. What it also does is to support single character prefixes like `%after` or `?after`.

In the end, it's just a few keywords to learn, I'd just hope it's possible to make it less alien, maybe `,` instead of `,` would already help.
And I say all this knowing basically nothing about what's convenient or truly possible.

slarse: I'll just consolidate the discussion here as it seems like a good place!

I agree that there is some alien stuff here (and magic), which is why I wanted to continue the discussion. Let's tackle the sources of alienness one at time. Each one of them is due to a perceived syntactical benefit on my part but whether or not that's worth the "magic" factor is not clear cut.

We can cut this down all the way to 0 magic and 0 exotic syntax, but it's tradeoffs all the way down!

### The `arg` magic
The `arg` part has a fair bit of magic that needs to be justified.

First, there's the use of an identifier `arg_name` identifier (e.g. `arg checkout_to`) that is later automatically formatted into `"--checkout-to"`. This could just as well be a literal instead of an identifier, but I found it confusing that a literal would change shape in any way and would then propose `arg "--checkout-to"`, i.e. no "smart formatting".

The second part of magic in `arg` is the fact that there's an optional assignment to the value that was passed, i.e. `arg <arg_name> = <arg_value>`. I just found this convenient.

My only justification for this is that I thought it was nice and concise. I'm not opposed to splitting these into separat `arg_name` and `arg_value` arguments and just make them literals.

### That darned `;`
The semicolon is needed to separate the format argument list of the message part (which is comma-separated) from the rest of the arguments.

```rust
bail_bad_input!(
    "This {} {} message", "is", "a"; // <--- needed here
    arg after; // <--- needed here
    hint "Some {} hint", "nice" // <--- would be needed here if there was something after, but there is not
)
```

Some alternatives to get rid of the semicolon entirely:

1. Don't allow format arg list for the message, only allow inline interpolation (e.g. `"This {arg1} {arg2} message"`). This means that you can't call methods or otherwise change the argument inline, but that may be worthwhile sacrifice.
2. Use some other separator for the format arg list - this moves the exotic syntax to the probably less common case!
3. Don't incorporate `format!()` capabilities at all - if you want to format, you explicitly use `format!()`.

My personal preference here would probably be to go with number 1, if we feel that not having the semicolon separator is important.

### `arg` and `hint` prefixes
The `arg` and `hint` prefixes are there to disambiguate between the two arguments. The matching complexity here stems from the fact that both the arg and hint parts are optional, which I think is an important property, but it does mean that they need distinct terminal tokens at the very start to disambiguate.

There are _endless_ ways to change this. Here are some alternatives:

**Label syntax:**

```rust
bail_bad_input!(
    "Some {} message", "nice";
    arg: after = some_expr,
    hint: "some {} hint", "nice"
);
```

The one downside I see with this is that it might make you think the labeled arguments are unordered, which is not the case.

**Enclosing in parentheses:**

```rust
bail_bad_input!(
    "Some {} message", "nice";
    arg(after = some_expr),
    hint("some {} hint", "nice")
);
```

**Clever symbol prefixing (even more magic!):**

```rust
bail_bad_input!(
    "Some {} message", "nice";
    => after = some_expr;
    ? "some {} hint", "nice"
);
```

**Plain name/value assignments (implies removing `arg` magic):**

```rust
bail_bad_input!(
    "Some {} message", "nice";
    arg_name = "--after";
    arg_value = some_expr;
    hint = "some {} hint", "nice"
);
```

Are you feeling drawn toward any of these alternatives?

davidpdrsn: Sorry for being late to this discussion but I wanna say that I strongly dislike macros like this.

I think hiding boilerplate behind simple macros like `anyhow::bail!()` is totally but once the macro starts having custom syntax for passing arguments then I don't think its worth it.

I think

```rust
bail_bad_input!(
    "Some {} message", "nice";
    arg_name = "--after";
    arg_value = some_expr;
    hint = "some {} hint", "nice"
);
```

should instead just be

```rust
return Err(bad_input(format!("Some {var} message"))
    .arg_name("--after")
    .arg_value(&after_value)
    .hint(format!("some {var} hint"))
    .into());
```

sure its more typing but I think there are clear benefits:

- It works with rust-analyzer. I don't need to read the docs to learn the macros custom syntax.
- It supports rustfmt.
- We can create abstractions around like it functions to commonly used "bad input errors". You cannot abstract over macros (except with more macros)

Writing good macros is also really hard and the code often becomes hard to read:

- Can you specify `arg_name`, `arg_value`, and `hint` in any order? This often leads to whats called a "TT muncher" which makes the code hard to follow.
- If you cannot pass the args in any order then the error messages are likely poor. They often just say "didn't expect `arg_name` here". Without saying why.
- You can pull tricks like just using the `arg_name` tokens in `arg_name = "..."` as method names on builders like `bad_input_builder.$arg_name_ident($arg_name_value)` but then your macro just replicates a builder
- How to support format strings for the values is also tricky for some of the same reasons. Supporting `bail_bad_input!("no vars")`, `bail_bad_input!("with some {}", "one")`, and `bail_bad_input!("with some {}", "a", "b", "c",)` (note the trailing comma) is totally doable but not trivial.

slarse: @davidpdrsn You're not late, I was waiting for you to come back from holiday :)

This is all the kind of stuff that I meant with this comment https://github.com/gitbutlerapp/gitbutler/pull/13929#discussion_r3297659980

I think this really concludes that this isn't a good idea. It would be perfect for a solo project for me, but @Byron desires some adjustments that aren't syntactically possible to fulfill, and every concern that @davidpdrsn has is one that I share.

The _one_ case where I think a macro would make sense is some `bad_input!()` macro that only does the equivalent of `BadInput::new(format!(...))`. But I feel like even that isn't really worthwhile as customizing the message is probably not even that interesting - it's mostly in the arg_name, arg_value and hint where you want dynamic content!

I'll write this off as a failed experiment and another validation that creating a DSL, however tempting, is never the right solution.

Byron: Thanks everyone!
Indeed, `bail_bad_input!` is a different beast to `bail` and that I had in mind. Ultimately, being explicit about `Err(BadInput...)` already resolves the convolution of the brain when it tries to parse it for what it is, and it has all these advantages that @davidpdrsn already pointed out.
KISS!

slarse: I've now removed `into_cli_result()` and added a tiny convenience wrapper `bad_input()` that just delegates to `BadInput::new()`, which I feel solves most of the problems the rather convoluted macro set out to solve.

Even though there are quite a few `bad_input(format!())` calls in this code sample, my plan is for most of them to go away and move details to `arg_name()`, `arg_value()` etc. So I still think there is little to no use case for a `bad_input!()` macro that wraps `fomrat!()` for the message.

Thanks for the inputs, I hope to have more constructive discussions like this again!

Issue #13932, opened 2026-05-23
Linux .deb AppStream metadata still reports 0.18.0, causing Flathub/flatpak to show stale version


n-a-m-e:
### Version

0.19.13

### Operating System

Linux

### Distribution Method

deb (Linux)

### Describe the issue

Hi! I noticed that the GitButler Flatpak on Flathub is receiving fresh builds, but Flathub and `flatpak list` still show version `0.18.0`.

I dug into the current Linux `.deb` release artifact and it looks like the Debian package version is correct, but the bundled AppStream metadata is stale.

From the `.deb` control metadata:

```text
Package: git-butler
Version: 0.19.13
```

But inside the payload:

```text
./usr/share/metainfo/com.gitbutler.gitbutler.metainfo.xml
```

the `<releases>` block still starts with:

```xml
<releases>
  <release version="0.18.0" date="2025-11-20"/>
  <release version="0.13.8" date="2024-10-30"/>
  ...
</releases>
```

This appears to explain why Flathub/Flatpak still display `0.18.0`, even though the actual app binary is newer.

I also looked at `scripts/release.sh`. It updates the temporary Tauri config version using the passed `--version`:

```bash
jq --arg version "$VERSION" \
   --argjson externalBin "$EXTERNAL_BIN" \
  '.version = $version | .bundle.externalBin = $externalBin' "$CONFIG_PATH" >"$TMP_DIR/tauri.conf.json"
```

but I could not see a corresponding step that updates the AppStream metainfo release entry before building the Linux `.deb`.

One possible fix would be to copy `crates/gitbutler-tauri/com.gitbutler.gitbutler.metainfo.xml` into the temporary build directory, insert the current release version into the copied file, and point the temporary Tauri config at that generated metainfo file for the `.deb`. That avoids mutating the source file during release builds.

I have attached 

[gitbutler-clean-metainfo-temp.patch](https://github.com/user-attachments/files/28166349/gitbutler-clean-metainfo-temp.patch)

a small patch that does this if useful, but I wanted to check whether you would prefer this to be handled in the release script, generated from release notes, or fixed some other way.

Impact:

* `.deb` package version is correct
* app menu reports a newer version
* Flathub/`flatpak list` still report `0.18.0`
* Flathub users may think the app is not updating even though new Flatpak commits are being published

### Relevant log output (Optional)

```shell

```

Byron:
Thanks a lot for reporting, and for looking into it.

[...]

PR #13936, opened 2026-05-24
Add missing version

ghost:
Please don't forget to update the metainfo file before releases.

## 🧢 Changes

Metadata related changes.

## ☕️ Reasoning

The Linux version of the application shows the wrong version.

### DIFF
--- a/crates/gitbutler-tauri/com.gitbutler.gitbutler.metainfo.xml
+++ b/crates/gitbutler-tauri/com.gitbutler.gitbutler.metainfo.xml
@@ -37,6 +37,9 @@
   </screenshots>
 
   <releases>
+    <release version="0.19.13" date="2026-05-19">
+      <url>https://github.com/gitbutlerapp/gitbutler/releases/tag/release%2F0.19.13</url>
+    </release>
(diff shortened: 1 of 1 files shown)


Byron:
Thanks for sharing!

Can you help me understand what problem this solves? From what I can tell, the releases on macOS at least know their version and it matches the app-reported version as well.

Screenshot 2026-05-25 at 05 31 58
Screenshot 2026-05-25 at 05 32 16

Thanks!

Byron:
Ah, now I see, [this comment](https://github.com/gitbutlerapp/gitbutler/issues/13932#issuecomment-4527961118) is the answer.

> Not just Flatpak, every Linux application software/store that relies on the metadata files.

We already substitute values into these files, and it looks like there should be one more manipulation as part of the release process.

@slarse is probably still the most versed in this matter.

slarse:
Edit: moved comment to https://github.com/gitbutlerapp/gitbutler/issues/13932#issuecomment-4532000227

slarse:
@yakushabb thanks for providing this concrete fix. I've replaced this PR with #13942, you're credited on the backfill there.

PR #13949, opened 2026-05-26
docs(security): retire rewards program

slarse:
With very few exceptions, the security reports we get hold very low quality and are clearly generated. We spend significantly more time vetting reports than addressing issues.

The utility of this kind of rewards program seems to have come to an end. It is at this point much more efficient for us to just generate reports ourselves, using our internal domain knowledge to point agents in the right direction.


krlvi (approved):
👍

PR #13954, opened 2026-05-26
Make "pointer" cursor default for buttons and treelists

PavelLaptev:


apps/lite/ui/src/global.css
```diff
@@ -1,5 +1,9 @@
 @import "@gitbutler/design-core/core";
 
+[role="treeitem"] {
+	cursor: pointer;
```
OliverJAsh, line 4: This feels a bit too generic: I imagine not all `treeitem`s will want pointer cursors. This is really specific to the commit graph and list of files. Ideally we'd apply this style to those treeitems only.

Edit: actually, first we should talk about `cursor: pointer`. Linear for example doesn't use it.

PavelLaptev: @OliverJAsh, yeah, I wasn't sure if we’d have non-interactive `treeitems`. I was assuming all tree views would be clickable/interactive.

There are `cursor: pointer` for buttons and links, those live at the reset level in the `design-core` package right now.

Good callout on Linear. We actually wrestled with the pointer vs. default question when we first kicked off the Tauri app. The pointer definitely feels very "webby," but we ended up looking at VSCode as a benchmark, and they use pointers for all interactive elements. cc @krlvi

Down to revisit this though! Let’s chat about it during the next Lite call.

PR #13961, opened 2026-05-27
edit-mode: label "New base" with parent message

jonathantanmy2:
Commit 2bf9813087 (gitbutler-edit-mode: edit edited commit directly, 2026-03-19) indirectly changed how conflicts were represented in the working tree when entering edit mode. Prior to that commit, the checked-out commit is the parent of the edited commit, and the "New base" conflict marker takes its description from the checked-out commit (i.e. the parent of the edited commit). That commit changed it such that the checked-out commit is the edited commit itself; the "New base" conflict marker still takes its description from the checked-out commit (now, the edited commit itself).

However, this results in both the "Current commit" and "New base" conflict markers taking their descriptions from the same commit, resulting in end-user confusion. The "New base" should actually take its description from the parent of the edited commit; it should not take it from what is checked out (the fact that the checked-out commit prior to commit 2bf9813087 was the parent of the edited commit is a coincidence). Restore this behavior.

### DIFF
--- a/crates/gitbutler-edit-mode/src/lib.rs
+++ b/crates/gitbutler-edit-mode/src/lib.rs
@@ -34,6 +34,8 @@ fn commit_title_to_merge_conflict_label(commit: &gix::Commit<'_>) -> String {
         but_core::commit::strip_conflict_markers(commit.message_bstr()).as_ref(),
     )
     .title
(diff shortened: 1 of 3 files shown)


estib-vega (approved):
Thanks Jonathan!

PR #13964, opened 2026-05-27
fix(cli): support deleting non-empty branches

davidpdrsn:
`but branch delete <branch>` will now first delete the commits then delete the branch. Previously deleting a non-empty branch would fail.

### DIFF
--- a/crates/but/tests/but/command/branch/delete.rs
+++ b/crates/but/tests/but/command/branch/delete.rs
@@ -17,3 +17,150 @@ Error: Branch 'no-such-branch' not found in any stack
 
     Ok(())
 }
+
+#[test]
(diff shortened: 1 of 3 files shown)


crates/but/tests/but/command/branch/delete.rs
```diff
@@ -17,3 +17,36 @@ Error: Branch 'no-such-branch' not found in any stack
 
     Ok(())
 }
+
+#[test]
+fn can_delete_branch_with_commits() -> anyhow::Result<()> {
```
slarse, line 22: Could we add a test for what happens if we delete a branch in the middle of a stack?

davidpdrsn: Added tests for deleting the bottom, middle, and top branches in a stack of three.

PR #13966, opened 2026-05-27
feat(cli): remove confirmation prompt from `oplog restore` and `branch delete`

davidpdrsn:
`but undo` supports both of these so don't think we need a confirmation.


slarse (approved):
Didn't read the code, but I approve of removing any confirm prompt where the action can be reversed by `but undo`. The confirm prompts are really quite annoying to me, we should have faith in our snapshotting (and fix it if it doesn't hold up).

PR #13968, opened 2026-05-27
Introduce CLI argument atoms

davidpdrsn:
This PR is an experiment aimed at reducing duplication in the CLI and making argument handling more consistent.

A lot of CLI commands currently accept plain `String` arguments for things like branch names, commit IDs, and CLI short IDs, and then each command resolves those strings in its own handler. That leads to duplicated logic, slightly different behavior between commands, and inconsistent bad-input error messages.

This introduces a small abstraction for typed CLI inputs: **argument atoms**.

Instead of commands taking raw strings like:

```rust
struct SomeCommand {
    branch: String,
}
```

they can now take purpose-specific argument types:

```rust
struct SomeCommand {
    branch: BranchArg,
}
```

The argument atoms are still parsed from strings by clap, but they make the command’s intent explicit and centralize resolution/error handling.

The current atoms are:

- `BranchArg`
- `CliIdArg`
- `CommitArg`

### `BranchArg`

Represents branch-ish user input and provides branch-specific resolution helpers:

- `resolve_local_branch_name()`  
  Converts to a local ref name like `refs/heads/foo`.
- `resolve_segment(ctx)`  
  Resolves to the corresponding workspace segment.
- `resolve_for_creation(repo)`  
  Validates that the provided name can be used for a new branch.
- `resolve_branch(ctx)`  
  Resolves to an existing branch listing.
- `try_resolve_stack(ctx)`  
  Attempts to resolve to an in-workspace stack, returning `None` if not found.

### `CliIdArg`

Represents a CLI ID that may point at multiple kinds of things, such as commits, branches, files, stacks, etc.

Resolution is now intentionally scoped to the workspace:

- `resolve_in_workspace(ctx, id_map, purpose)`
- `try_resolve(ctx, id_map, purpose)`
- `resolve_commit_in_workspace(ctx, id_map)`
- `try_resolve_commit(ctx, id_map)`
- `resolve_branch_in_workspace(ctx, id_map)`
- `try_resolve_branch(ctx, id_map)`

These methods return either the requested specific type or a `ResolvedCliIdArg`:

```rust
enum ResolvedCliIdArg {
    Commit(gix::ObjectId),
    Branch(BranchArg),
}
```

They also take a `Purpose` where needed, so error messages can say whether the user was trying to resolve an anchor, branch, or target.

### `CommitArg`

Represents commit-ish input and resolves directly against the repository object database:

- `resolve(repo)`

This handles invalid prefixes, ambiguous prefixes, and missing commits through consistent bad-input errors.

This PR applies the pattern to parts of the legacy CLI, including `but branch` and `but reword`, so handlers now more clearly follow the desired structure:

1. Resolve inputs, returning bad-input errors where appropriate.
2. Perform the command logic.
3. Render the requested output format.


slarse (approved):
@davidpdrsn This looks stellar! I fully agree with the direction this takes. The _many_ places in which we resolve CLI IDs have especially been chafing at me.

PR #13998, opened 2026-05-29
fix problem with `commit already belongs to another branch`

mtsgrd:
## 🧢 Changes

`but branch new <name>` — and the modern `create_virtual_branch`, which takes the same path — failed to create a branch whenever the target branch (`origin/main`) had advanced past the applied workspace:

```
Error: Branch '<name>' cannot be created: the target commit (<sha>) already
belongs to another branch in the workspace. Each commit can only belong to
one branch at a time.
```

Both go through `create_reference(.., None, ..)`. This PR:

- **Anchors** a no-anchor ("independent") branch at `merge_base(target_tip, workspace_commit)` — the fork point, always *inside* the workspace — instead of `resolved_target_commit_id()` (the target tip, which can sit *outside* it).
- **Errors clearly** when there is no resolvable workspace commit (a headless / unmanaged workspace), rather than silently falling back to the out-of-workspace target tip.
- Adds a deterministic test scenario (`stack-below-advanced-target`) and a regression test.

### When it happens

A stack is applied below a target that has since moved ahead:

```
* 1021d74 (HEAD -> gitbutler/workspace) GitButler Workspace Commit
* 49d4b34 (A) A1
| * bce0c5e (origin/main, main) M2
|/
* 3183e43 M1
```

`A` forks from `M1`; `origin/main` advanced to `M2` (a sibling of `A1`). So the target tip `M2` sits **outside** the workspace — note `⇣1` (target is one commit ahead) and `on 3183e43` (the workspace base is `M1`):

```
📕🏘️:0:gitbutler/workspace[🌳] <> ✓refs/remotes/origin/main⇣1 on 3183e43
└── ≡📙:3:A on 3183e43 {0}
    └── 📙:3:A
        └── ·49d4b34 (🏘️)
```

### Before — the bug

The no-anchor path anchored the new ref at `resolved_target_commit_id()` = `M2`. Because `M2` isn't in the workspace, re-projection can't surface the new ref as its own segment, so creation bailed:

```
Error: Branch 'new-branch' cannot be created: the target commit (bce0c5e…)
already belongs to another branch in the workspace.
```

### After — the fix

Anchoring at `merge_base(M2, ws-commit)` = `M1` (inside the workspace), `new-branch` emerges cleanly as its own stack, based at `M1`:

```
📕🏘️:0:gitbutler/workspace[🌳] <> ✓refs/remotes/origin/main⇣1 on 3183e43
├── ≡📙:3:A on 3183e43 {0}
│   └── 📙:3:A
│       └── ·49d4b34 (🏘️)
└── ≡📙:5:new-branch on 3183e43 {3e5}
    └── 📙:5:new-branch
```

### Verification

End-to-end via `repro.sh` (`but apply foo` below an advanced `origin/main`, then `but branch new bar`), run from an **empty** directory:

| `but` binary | `but branch new bar` |
| --- | --- |
| nightly `0.5.2047` | ❌ errors |
| this branch | ✅ creates the branch |

- Library regression test `no_anchor_branch_with_target_tip_outside_workspace` pins both graphs above.

### DIFF
--- a/crates/but-workspace/tests/workspace/branch/create_reference.rs
+++ b/crates/but-workspace/tests/workspace/branch/create_reference.rs
@@ -1226,6 +1226,79 @@ mod with_workspace {
         );
         Ok(())
     }
+
+    /// Regression: `but branch new <name>` and `create_virtual_branch` both go through
+    /// `create_reference(.., None, ..)`. When the target (`origin/main`) is advanced past the
+    /// workspace, its tip sits OUTSIDE the workspace. The no-anchor path used to anchor the new
+    /// ref at that tip and bail with "the target commit ... already belongs to another branch".
+    /// It must instead anchor at `merge_base(target_tip, ws-commit)` (here M1), inside the
+    /// workspace, so the branch emerges cleanly as its own stack.
+    #[test]
+    fn no_anchor_branch_with_target_tip_outside_workspace() -> anyhow::Result<()> {
+        let (_tmp, repo, mut meta) = named_writable_scenario("stack-below-advanced-target")?;
+        insta::assert_snapshot!(visualize_commit_graph_all(&repo)?, @"
+        * 1021d74 (HEAD -> gitbutler/workspace) GitButler Workspace Commit
+        * 49d4b34 (A) A1
+        | * bce0c5e (origin/main, main) M2
+        |/  
+        * 3183e43 M1
+        ");
+
+        // `A` is applied (in the workspace), based at M1.
+        add_stack_with_segments(&mut meta, 0, "A", StackState::InWorkspace, &[]);
+
+        let graph = but_graph::Graph::from_head(&repo, &meta, Options::limited())?;
+        let ws = graph.into_workspace()?;
+        insta::assert_snapshot!(graph_workspace(&ws), @"
+        📕🏘️:0:gitbutler/workspace[🌳] <> ✓refs/remotes/origin/main⇣1 on 3183e43
+        └── ≡📙:3:A on 3183e43 {0}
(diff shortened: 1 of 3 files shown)


crates/but-workspace/tests/workspace/branch/create_reference.rs
```diff
@@ -1226,6 +1226,79 @@ mod with_workspace {
...
+            .expect("the scenario sets a default target");
+        assert!(
+            ws.find_owner_indexes_by_commit_id(target_id).is_none(),
+            "the target tip must be outside the workspace for this repro"
+        );
+
+        // Creating the no-anchor branch now succeeds (it used to bail).
+        let new_name = rc("refs/heads/new-branch");
+        let new_ref = new_name.as_ref();
+        let updated_ws = but_workspace::branch::create_reference(
```
OliverJAsh, line 1272: I don't know much about how our testing works, but I'm curious if it's possible to write a test that uses the original reduced test case script I shared, where it fails on `but branch new`.

The nice thing about this is we can easily run the script to verify the behaviour, without getting into low-level Rust details like we do here. More generally this would allow us to run tests against any other reduced test cases we receive from user bug reports.

I guess what I'm suggesting is a higher-level form of testing, not as a replacement necessarily but in addition to what we have here.

It's very possible I'm being incoherent here—there's still a lot for me to wrap my head around.

Caleb-T-Owens (approved):
Merging because it seems like an improvement in behaviour.

Having read the algorithm at play here, I'm a little puzzled about some of the behaviour going on and would like to revisit this in more depth at some point.

PR #14004, opened 2026-05-30
Expose undo/redo hotkeys globally

OliverJAsh:
I don't see a reason why they should only be exposed in the outline panel.


samhh:
My thinking was that we might have actions to undo in other panels in the future that aren't inputs, for example adding and then undoing/removing a comment on a diff.

We can punt this in any case.

OliverJAsh:
Good point, we'll need to think about that. For now though I think regular undo should still work e.g. when files panel is focused.

PR #14024, opened 2026-06-01
Fix discarding conflicted commits

Caleb-T-Owens:
There is no reason why we shouldn’t be able to?


estib-vega (approved):

estib-vega:
To me it makes sense to remove. There's a test that still guards this, but we can remove that as well

PR #14025, opened 2026-06-01
feat(cli): beginnings of cli revamp (commit2/move2/squash2)

davidpdrsn:
This adds documentation for the revamped CLI commands that @slarse and I are working on.

It also adds a POC `but commit2` command with a test. The new commands are hidden and requires building but with `--features but-2`. We'll do that in nightly but not in stable releases. This required a few CI tweaks.

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -300,6 +300,8 @@ jobs:
           rm cargo-nextest.tar.gz
       - name: Run workspace tests
         run: cargo nextest run --workspace --exclude gitbutler-tauri --exclude but-api-macros-tests --profile ci
+      - name: Run experimental cli command tests
+        run: make test-but-2
       - name: Run doctests
         run: cargo test --doc --workspace --exclude gitbutler-tauri --exclude but-api-macros-tests
       - name: Upload results to Buildkite Test Engine
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,16 @@ check-modern-but:
 test-modern-but:
 	cargo test -p but --no-default-features
 
+# Lint: `but` with new experimental commands.
+.PHONY: clippy-but-2
+clippy-but-2:
+	cargo clippy -p but --features but-2 -- -D warnings
+
+# CI executes this to test a version of `but` with new experimental commands.
+.PHONY: test-but-2
+test-but-2:
+	cargo test -p but --features but-2
+
 # Run all tests in the entire workspace and show all failures in the end.
 .PHONY: nextest
 nextest:
--- a/crates/but/Cargo.toml
+++ b/crates/but/Cargo.toml
@@ -41,6 +41,8 @@ legacy = [
 ]
 # Feature that enables profiling related features of the TUI.
 tui-profiling = []
+# Enables new experimental CLI commands that'll eventually replace old commands
+but-2 = []
 
 [dependencies]
 but-core.workspace = true
--- a/crates/but/but-cli-revamp/branch-new.md
+++ b/crates/but/but-cli-revamp/branch-new.md
@@ -1,0 +1,170 @@
+```
+but branch new <name>?
+but branch new -A/--above <commit or branch>
+but branch new -B/--below <commit or branch>
+but branch new --format human/json/shell
+```
+
+Example baseline
+```
+┊╭┄op [top]
+┊●   20e235d six
+┊●   e28a167 five
+┊│
+┊├┄mi [middle]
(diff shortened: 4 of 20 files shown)


cli-revamp/commit-alt.md
```diff
@@ -0,0 +1,335 @@
...
+┊●   f3e56a9 some commit (no changes)
+┊●   NEW COMMIT GOES HERE
+┊│
+┊├┄i0 [middle]
+┊●   b440dea add bar-1.txt
+┊│
+┊├┄fo [bottom]
+┊●   9cd3abf add foo.txt
+├╯
+```
```
slarse: This feels inconsistent with `--above` as this actually places the commit "inside" the branch. It's rather at the bottom of the branch, than below it. I'm also a bit concerned about the possibility of non-linear segments causing ambiguity as to where to place the commit.

Another concern that I have is the utility of it. To me, the most useful thing to do is to insert at the head of a branch, and it's unclear how to do that here. While you can do `--above <head_commit>`, that's not super nice because `<head_commit>` is a moving target. If I want to insert two commits at the head of a branch, I need to do `but commit --below <head_commit>` and `but commit --below <new_head_commit>`. It doesn't seem smooth.

I think we should take this to a broader audience to get some more input.

> Note: When considering `but move` and `but commit` in conjunction, I think the argument that `--below` should treat the branch as a bucket becomes pretty strong, but I'm still hesitant about the details.

cli-revamp/commit.md
```diff
@@ -0,0 +1,350 @@
...
+but commit --format json/shell
+```
+
+These examples all assume we remove staging/assigments. The rational that if
+deciding what to commit is easy and commits are easy to mutate, then you don’t
+need staging
+
+# Questions
+
+- Are commit messages required?
```
slarse: IMO no, but I think there should be a flag to leave the message empty, as suggested below with `--no-edit`.

In most cases, no commit message is a mistake on the user's part.

cli-revamp/commit.md
```diff
@@ -0,0 +1,350 @@
...
+- If you save and quit the commit message without writing a message what happens?
```
slarse: Abort the commit. It would be very unexpected to me that it would proceed with the commit with an empty message.

cli-revamp/commit.md
```diff
@@ -0,0 +1,350 @@
...
+- Can you commit without a message and without opening the editor, like `git commit --no-edit`?
```
slarse: Yes.

cli-revamp/move.md
```diff
@@ -0,0 +1,258 @@
...
+┊●   0db9c35 three
+┊│
+┊├┄fo [bottom]
+┊●   9cd3abf two
+┊●   58f9cc1 one
+├╯
+```
+
+## Moving commits
+but move 0db9c35 --target top
```
slarse: I think maybe `--into` (or similar synonym) is more descriptive and contrasts more clearly to `--above` and `--below`.

```suggestion
but move 0db9c35 --into top
```

This also speaks for the fact that `--below <branch>` should be at the bottom of/below the branch, as otherwise `--into <branch>` and `--below <branch>` will have the same meaning, which is strange.

I'm also wondering, could it be optional s.t. we assume that the last argument is the `--into` target? I.e. these two being equivalent.

```suggestion
but move 0db9c35 top
but move 0db9c35 --into top
```

I think that'd be nicer overall.

cli-revamp/move.md
```diff
@@ -0,0 +1,258 @@
...
+
+Can you do `but move f3e56a9 --target zz` to create a uncommit?
+
+Can you move all the commits on a branch, without moving the branch itself?
+
+Can you specify commit ranges?
+
+Can you do `but move f3e56a9 --after top --branch bar` to create a new branch?
+
+Can you mix sources like `but move f3e56a9 middle`?
```
slarse: Spontaneous answer is no. I think it opens up for a lot of nonsense input possibilities and utility seems questionable. Just allowing multiple sources (which I think is implied) is complex enough without them being of different types.

crates/but/but-cli-2-0/patterns.md
```diff
@@ -0,0 +1 @@
+- Avoid `--x` and `--no-x` flags
```
slarse: What do we prefer?

* `-x` and `-X`?
* `-x=[true|false]`?
* ?

krlvi (approved):

PR #14026, opened 2026-06-01
Use CodeView

samhh:
https://diffs.com/docs#codeview

This performs substantially better with large diffs. At this point our biggest performance bottleneck is our unvirtualised files list, as per the speedup when it’s toggled closed.

If we’re using Pierre then this is very likely how we’ll want to consume the library. I'd rather adopt it earlier than later as it may have implications for how we approach implementation, UX tradeoffs, etc.

This change drops the minimal custom styling we had for file headers (I haven't bothered removing the CSS). We can reintroduce that later when the design is more settled.

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/DetailsPanel.tsx
+++ b/apps/lite/ui/src/routes/project/$id/workspace/DetailsPanel.tsx
@@ -30,7 +30,7 @@ import { OperationSourceC } from "#ui/routes/project/$id/workspace/OperationSour
 import { useAppDispatch, useAppSelector } from "#ui/store.ts";
 import { classes } from "#ui/components/classes.ts";
 import { Tooltip } from "@base-ui/react";
-import { DiffHunk, HunkHeader, TreeChange, UnifiedPatch } from "@gitbutler/but-sdk";
+import { HunkHeader, TreeChange, UnifiedPatch } from "@gitbutler/but-sdk";
 import { PatchDiff, Virtualizer } from "@pierre/diffs/react";
 import { useSuspenseQueries } from "@tanstack/react-query";
 import { useParams } from "@tanstack/react-router";
@@ -64,23 +64,6 @@ const patchHeaderForChange = (change: TreeChange, lineEnding: string): string =>
 		Match.exhaustive,
 	);
 
-const hunkKey = (hunk: HunkHeader): string =>
-	`${hunk.oldStart}:${hunk.oldLines}:${hunk.newStart}:${hunk.newLines}`;
-
-const Hunk: FC<{
-	change: TreeChange;
-	hunk: DiffHunk;
-}> = ({ change, hunk }) => (
-	<PatchDiff
-		patch={`${patchHeaderForChange(change, lineEndingForDiff(hunk.diff))}${hunk.diff}`}
-		options={{
-			diffStyle: "unified",
-			themeType: "system",
-			disableFileHeader: true,
-		}}
-	/>
-);
-
 const FileDiff: FC<{
 	change: TreeChange;
 	diff: UnifiedPatch | null;
@@ -99,8 +82,15 @@ const FileDiff: FC<{
 			return (
 				<ul>
 					{hunks.map((hunk) => (
-						<li key={hunkKey(hunk)}>
-							<Hunk change={change} hunk={hunk} />
+						<li key={`${hunk.oldStart}:${hunk.oldLines}:${hunk.newStart}:${hunk.newLines}`}>
+							<PatchDiff
+								patch={`${patchHeaderForChange(change, lineEndingForDiff(hunk.diff))}${hunk.diff}`}
+								options={{
+									diffStyle: "unified",
+									themeType: "system",
+									disableFileHeader: true,
+								}}
+							/>
 						</li>
 					))}
 				</ul>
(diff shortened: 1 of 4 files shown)


apps/lite/ui/src/routes/project/$id/workspace/DetailsPanel.tsx
```diff
@@ -55,160 +40,103 @@ const patchHeaderForChange = (change: TreeChange, lineEnding: string): string =>
...
+				].join(lineEnding) + lineEnding,
 		),
+
 		Match.when(
 			{ type: "Rename" },
-			({ subject }) => `--- ${subject.previousPath}${lineEnding}+++ ${change.path}${lineEnding}`,
+			({ subject }) =>
+				[
+					`diff --git a/${subject.previousPath} b/${change.path}`,
+					"similarity index 99%",
```
samhh, line 78: This should be 100% for renames with no changes (a "pure rename"). Likewise we're not currently passing every type of patch through to Pierre. I'd like to solve that on another day.

samhh:
This PR isn't 100% cooked but if approved I'd rather merge now to avoid conflicts and deal with what's missing tomorrow. 🙂

OliverJAsh:
How will we support drag and drop for files and hunks in the diff view? I think that is a pretty important factor (along with keyboard selection) as to how we render diffs.

(Previously this was already working via `OperandSourceC`.)

> rather merge now to avoid conflicts

Skill issue 😉 You need to master the conflicts! Mergiraf to the rescue.

samhh:
lol true. Not much time to code today and I wanted to squeeze this in before heading offline.

For drag and drop I imagine the escape hatch we'll use is `onPostRender`, I had a POC for this pre-`CodeView`. We'll need to hook some stuff up inside the library's shadow DOM. There are other, simpler APIs we can use for styling.

For context that prior experiment was aimed at using file diffs instead of hunk diffs, to avoid what I believe they mention here: [https://pierre.computer/writing/on-rendering-diffs#:\~:text=There%20was%20still%20a%20lot%20of%20O(n×m)%20complexity%2C%20high%20memory%20usage%2C%20and%20virtualization%20blanking.](<https://pierre.computer/writing/on-rendering-diffs#:~:text=There%20was%20still%20a%20lot%20of%20O(n%C3%97m)%20complexity%2C%20high%20memory%20usage%2C%20and%20virtualization%20blanking.>)

OliverJAsh:
I also wonder about selection (up/down) for hunks and files. That's a really important feature that might be fundamentally at odds with this library. 😬 I hope not.

apps/lite/ui/src/routes/project/$id/workspace/DetailsPanel.tsx
```diff
@@ -64,77 +57,50 @@ const patchHeaderForChange = (change: TreeChange, lineEnding: string): string =>
...
-const FileDiff: FC<{
-	change: TreeChange;
-	diff: UnifiedPatch | null;
-}> = ({ change, diff }) =>
-	Match.value(diff).pipe(
-		Match.when(null, () => <div>No diff available for this file.</div>),
-		Match.when({ type: "Binary" }, () => <div>Binary file (diff not available).</div>),
-		Match.when({ type: "TooLarge" }, ({ subject }) => (
-			<div>Diff too large ({subject.sizeInBytes} bytes).</div>
-		)),
```
OliverJAsh, line 143: Do we not need to handle these anymore? 🤔 Especially curious about `TooLarge`.

OliverJAsh:
Another thing that might be related: we still need to reset scroll when changing outline selection. (We don't need to do it when changing file selection because as discussed we're planning to have one scroller for all files.)

samhh:
I can mentally picture how we'd achieve selections even if it's not first-class supported by the library; we own the state, we can manipulate styling, the library natively supports line selection and adjusting scroll position.

Resetting scroll appears to be unaffected by this PR. Scroll position is retained for the same scope of work on modification e.g. changes in worktree, as per the ID + version combo.

OliverJAsh:
I think we also need to consider accessibility for selections, which probably means we need some control of the DOM. For outline and files we use the treeview pattern.

samhh:
Performance is an accessibility concern as well.

apps/lite/ui/src/routes/project/$id/workspace/DetailsPanel.tsx
samhh: We do, but I'm feeling we can put this off until later. If the library doesn't have anything for e.g. blobs then we can inject our own entries at the top or bottom as a workaround.

samhh: I've added support for file headers in the diff for binaries. The other two scenarios are curious and need revisiting in any case.

amadeus:
Hey just randomly came across this (dev behind CodeView/diffs). I tried to figure out how to render the diffs in gitbutler (both stable and nightly, but couldn't figure out how). Regardless though, please let me know if there are any feature requests/issues you guys have with implementation so I can make sure i take a closer look!

samhh:
> Hey just randomly came across this (dev behind CodeView/diffs). I tried to figure out how to render the diffs in gitbutler (both stable and nightly, but couldn't figure out how). Regardless though, please let me know if there are any feature requests/issues you guys have with implementation so I can make sure i take a closer look!

Hey! 👋 

We're using diffs in a new workspace (`apps/lite`) that will become a future version of GitButler's GUI. Something very roughly like this should get you up and running if you were curious:

```console
$ pnpm install
$ pnpm -F @gitbutler/but-sdk build
$ pnpm dev:lite
```

Though for now you'll also need to add a repo/project outside of this app. I can't remember the fastest way to do that. cc @OliverJAsh 

We've raised a couple of tickets on your side already, I'm sure you'll come to recognise our avatars 😛 thanks for popping by!

PR #14027, opened 2026-06-01
Lite: use default cursors for buttons

OliverJAsh:
As discussed. @PavelLaptev to remove from design-core.


PavelLaptev:
@OliverJAsh looks good. I’ll also include a commit here with the updated `design-core` and let’s merge it.

PR #12169, opened 2026-02-03
WIP: Auto resolving pick bases

Caleb-T-Owens:

PR #14032, opened 2026-06-02
but-rebase: cherry pick writes out gb merge conflicts

estib-vega:
`cherry_pick` writes out a GitButler conflicted-trees commit under the circumstance of it being a ‘conventional merge commit':
- Having exactly two parents
- Being empty
  
This allows us to use the existing representation of conflicted commits.

---

### Notes for the reviewer

I encountered this while testing remote-branch integration strategies in which merging the remote into the local branch would result in a conflict.
As it was, this would result in an error instead of a gb-managed conflict:
```
Failed to merge bases while cherry picking commit ...
Encountered a conflict while merging the commit's new bases: local, remote.
```

### Root cause

Given a commit X with initial parents Y,Z and a destination parents A,B (we want to rebase X on top of A,B).
What `cherry_pick` does is to calculate the merge tree of Y+Z = N and the merge tree of A+B = M, so that we can use N as the 'base' M as the 'ours' side and X as the 'theirs' side for merging.

The issue was that if the destination parents (A + B) resulted on a conflict, we would raise an error.
What we'd want to happen is to write the auto-resolution tree into that commit instead.
The only **caveat** is that we can only do that for the specific case of a 'classic' merge commit: 
The tree of X must be the empty tree and the destination parents must be exactly 2.

### Things to consider
This makes the PR a bit more *magical* in that we handle a very specific scenario in a special way.
I think this is ok since the merge-commit scenario itself is special and should be handled that way.

The *more-correct* way of solving this would be to extend the way we represent conflicts to support N-sides of conflicts.

### Effects of this PR

Integrating the latest target into a stack, that results in a conflict doesn't fail, but it writes a conflicted commit that we can later resolve.

### DIFF
--- a/crates/but-rebase/src/graph_rebase/cherry_pick.rs
+++ b/crates/but-rebase/src/graph_rebase/cherry_pick.rs
@@ -80,9 +80,28 @@ pub enum PickMode {
 /// Except in the case where X is conflicted. In that case we then make use of
 /// X's "base" sub-tree as the base.
 ///
-/// `pick_mode` - controls how to determine if a commit should be cherry-picked.
-/// `tree_merge_mode` - controls how parent trees are merged for merge commits.
-/// `sign_commit` - controls how the resulting commit is signed.
+/// Special case: when a synthetic merge template has no original parents, an
+/// empty tree, and exactly two new parents whose trees conflict with each
+/// other, we materialize that conflict as a GitButler conflicted merge commit
+/// instead of returning [`CherryPickOutcome::FailedToMergeBases`].
+///
+/// `repo`
+/// The repository that owns `target`, `ontos`, and any newly written commit or tree objects.
+///
+/// `target`
+/// The commit to cherry-pick, possibly including conflicted-tree metadata or a synthetic merge template.
+///
+/// `ontos`
+/// The parent commits that `target` should be replayed onto in the result graph.
+///
+/// `pick_mode`
+/// Controls whether the cherry-pick may short-circuit to identity or must recreate the commit.
+///
+/// `tree_merge_mode`
+/// Controls how parent trees are merged while constructing merge-commit bases and ontos.
+///
+/// `sign_commit`
+/// Controls whether the newly written commit should be signed.
 pub fn cherry_pick(
     repo: &gix::Repository,
     target: gix::ObjectId,
@@ -123,11 +142,6 @@ pub fn cherry_pick(
             // by the ontos & parents comparison or the PickMode::Force case.
             Ok(CherryPickOutcome::Identity(target.id.detach()))
         }
-        // TODO(cto): We can handle the specific case where (the base is
-        // _not_ conflicted & the ontos _is_ conflicted & the target == base
-        // & ontos.len() == 2) by writing out the ontos conflict as a
-        // conflicted merge commit, without expanding our conflict
-        // representation.
         (MergeOutcome::Conflict { commits: bases }, MergeOutcome::Conflict { commits: ontos }) => {
             Ok(CherryPickOutcome::FailedToMergeBases {
                 base_merge_failed: true,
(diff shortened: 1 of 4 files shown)


crates/but-rebase/src/graph_rebase/cherry_pick.rs
```diff
@@ -142,12 +156,26 @@ pub fn cherry_pick(
...
-        }),
+        (_, MergeOutcome::Conflict { commits }) => {
+            if let Some(outcome) = maybe_materialize_conflicted_onto_merge(
+                repo,
+                &target,
+                ontos,
+                &base_t,
+                target_t.detach(),
+                tree_merge_mode,
+                sign_commit,
```
estib-vega, line 167: If the onto merge is conflicted, and the target is empty, we write the GB-managed conflicted commit.

crates/but-workspace/tests/workspace/upstream_integration.rs
```diff
@@ -321,3 +323,104 @@ fn diamond_partially_content_integrated_merge() -> Result<()> {
...
+
+    insta::assert_snapshot!(visualize_commit_graph_all(&repo)?, @r"
+    * 379fa91 (HEAD -> gitbutler/workspace) GitButler Workspace Commit
+    *   9b4efdf (A) [conflict] Merge refs/remotes/origin/A into merge
+    |\  
+    | * f03fc2c (origin/A, new-origin) remote change in A 1
+    * | 61c4a24 local change in A 1
+    |/  
+    * 2b73dee (origin/main, main) init-integration
+    ");
```
estib-vega, line 378: This change allows the upstream integration to write out conflicted commits that we can resolve

crates/but-rebase/src/graph_rebase/cherry_pick.rs
```diff
@@ -190,6 +218,98 @@ pub fn cherry_pick(
...
+    {
+        return Ok(None);
+    }
+
+    let Some(base_commit_id) = merge_base(repo, ontos[0], ontos[1])? else {
+        return Ok(None);
+    };
+    let base_tree_id = but_core::Commit::from_id(base_commit_id.attach(repo))?
+        .tree_id_or_auto_resolution()?
+        .detach();
```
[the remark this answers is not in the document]
estib-vega: Done

Caleb-T-Owens:
Nice PR!

In the case where we're cherrypicking a merge commit that doesn't have any contents, this works well.

I'm wondering how it handles the case where the merge commit contains some additional changes.

A scenario could be:

```
* X: foo.txt: x
|\
* | B: foo.txt: b
| * A: foo.txt: a
|/
| * C: foo.txt: c
|/
* Base: foo.txt: A
```

We rebase $X$ from it's original parents $B$ and $A$, onto $A$ and $C$.

With your algorithm, are we going to end up with $X'$ containing `foo.txt: c`?

I tried something similar here https://github.com/gitbutlerapp/gitbutler/pull/12169/changes but ran into issues when the number of bases changed.

estib-vega:
> I'm wondering how it handles the case where the merge commit contains some additional changes.
>

@Caleb-T-Owens It wouldn't haha, it very specifically checks whether the onto is exactly 2 parents and the target commits is empty

crates/but-rebase/src/graph_rebase/cherry_pick.rs
```diff
@@ -190,6 +218,93 @@ pub fn cherry_pick(
...
+    let empty_tree = gix::ObjectId::empty_tree(repo.object_hash());
+    if !matches!(base_t, MergeOutcome::NoCommit)
+        || target.is_conflicted()
+        || !target.parents.is_empty()
+        || target_tree_id != empty_tree
+        || ontos.len() != 2
+        || tree_merge_mode != TreeMergeMode::WithRenames
+    {
+        return Ok(None);
+    }
```
estib-vega, line 267: This is the conditions that would make it **not** write the GB conflicted commit

Caleb-T-Owens (approved):
Makes sense

PR #14033, opened 2026-06-02
but gui: open project in a new window

estib-vega:
Add the ability to open a project in a new window from the but CLI.

Usage:
```
but gui -n [path/to/my/project]
but gui --new-window [path/to/my/project]
```

### DIFF
--- a/crates/but-path/src/lib.rs
+++ b/crates/but-path/src/lib.rs
@@ -195,7 +195,11 @@ impl AppChannel {
     ///
     /// Note: On Linux, we don't have a good way of distinguishing between channels in the installed
     /// binaries, so we always just resolve `gitbutler-tauri`.
-    pub fn open(&self, possibly_project_dir: &std::path::Path) -> anyhow::Result<()> {
+    pub fn open(
+        &self,
+        possibly_project_dir: &std::path::Path,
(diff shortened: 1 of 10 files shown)


crates/but-path/src/lib.rs
```diff
@@ -210,10 +214,11 @@ impl AppChannel {
             .as_millis();
 
         let url = format!(
-            "{}://open?path={}&t={}",
+            "{}://open?path={}&t={}&new_window={}",
```
slarse, line 217: Snake case in query param is kind of odd.

slarse (approved):

crates/but-path/src/lib.rs
estib-vega: Is that not the default 😮 ?

slarse: Maybe we do that? Because of the prevalence of TS/JS in web, query parameters are (in my experience) most often camelCased. Like keys in JSON objects tend to be camelCase (same reason).

It really doesn't matter, though, I regret wasting braincells even making that comment.

Ship it!

AndrewMast:
Thanks, I appreciate y'all getting this done so fast!

PR #14034, opened 2026-06-02
fix id_for_tree_diff: Don't deref the in-memory

estib-vega:
Changing that argument to &repo.objects keeps the full object proxy in
use. That means `gix::diff::tree()` can resolve both persisted objects
and dry-run overlay objects, so workspace preview generation can diff
conflicted commits without requiring preview-only trees to be
materialized into the real repository.

--- 

This fixes an issue in which merge-conflict trees written in dry-run are found even if they don't reference an existing tree already

### DIFF
--- a/crates/but-workspace/src/changeset.rs
+++ b/crates/but-workspace/src/changeset.rs
@@ -7,7 +7,6 @@
 use std::{
     borrow::Borrow,
     collections::{HashMap, HashSet, VecDeque, hash_map::Entry},
-    ops::Deref,
     time::Duration,
 };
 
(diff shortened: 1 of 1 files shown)


Caleb-T-Owens (approved):

estib-vega:
I found this because of this new functionality: https://github.com/gitbutlerapp/gitbutler/pull/14032

Now that we can write the conflicted trees of a merge, we need to be able to find those in the in-memory repo

crates/but-workspace/src/changeset.rs
```diff
@@ -723,7 +722,7 @@ fn id_for_tree_diff(
         ),
         gix::objs::TreeRefIter::from_bytes(&rhs_tree.data, repo.object_hash()),
         &mut state,
-        repo.objects.deref(),
+        &repo.objects,
```
Byron, line 725: Very subtle!
And something I will try to remember - previously I wasn't really aware of differences between `&` and `.deref()`, even though it makes sense that the latter can have a custom implementation which isn't necessarily triggered (i.e. auto-deref) by `&` if `&T` already implements what's needed.

PR #14037, opened 2026-06-02
feat(GB-1121): add sign-on-push functionality

slarse:
Picking up sign-on-push again!

This is just a starting point and by no means done, I just want a basis for conversation.

Will add some comments on things to discuss tomorrow.

### DIFF
--- a/crates/but-workspace/src/commit/sign_commits.rs
+++ b/crates/but-workspace/src/commit/sign_commits.rs
@@ -1,0 +1,69 @@
+//! An action to forcibly sign all commits on a branch.
+
+use std::borrow::Borrow;
+
+use anyhow::Result;
+use but_core::{RefMetadata, commit::SignCommit};
+use but_graph::CommitFlags;
+use but_rebase::graph_rebase::{
+    Editor, GraphEditorOptions, Pick, Step, SuccessfulRebase, cherry_pick::PickMode,
+};
+use gix::prelude::ObjectIdExt;
+
+/// The result signing all commits on a branch.
+#[derive(Debug)]
+pub struct SignCommitsOutcome<'ws, 'meta, M: RefMetadata> {
+    /// A successful rebase result for continuing operations. This will be always provided
+    /// regardless of whether any commits were signed.
+    pub rebase: SuccessfulRebase<'ws, 'meta, M>,
+}
+
+/// Sign all commits on the branch.
+///
+/// If all commits are already signed, this is a NOOP.
+pub fn sign_commits<'ws, 'meta, M: RefMetadata>(
+    ref_name: impl Borrow<gix::refs::FullNameRef>,
+    repo: &gix::Repository,
+    workspace: &'ws mut but_graph::Workspace,
+    meta: &'meta mut M,
+) -> Result<SignCommitsOutcome<'ws, 'meta, M>> {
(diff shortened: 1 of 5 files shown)


crates/but/src/args/push.rs
```diff
@@ -11,6 +11,9 @@ pub struct Command {
     /// Bypass pre-push hooks
     #[clap(long = "no-hooks", visible_alias = "no-verify")]
     pub no_hooks: bool,
+    /// Sign commits before pushing
+    #[clap(long, default_value_t = false, hide = true)]
+    pub sign_commits: bool,
```
slarse, line 16: I don't think this is what we wanna do, rather it should be setting in Git-config. But potentially we could start here for experimentation purposes while I dogfood the feature.

crates/but/src/command/legacy/push.rs
```diff
@@ -57,9 +61,9 @@ pub async fn handle(
         return handle_dry_run(ctx, &args.branch_id, out);
     }
 
-    let guard = ctx.shared_worktree_access();
-    let perm = guard.read_permission();
-    let id_map = IdMap::new_from_context(ctx, None, perm)?;
+    let mut guard = ctx.exclusive_worktree_access();
+    let perm = guard.write_permission();
+    let id_map = IdMap::new_from_context(ctx, None, perm.read_permission())?;
```
slarse, line 70: We need exclusive access to be able to create a mutable workspace to sign commits.

crates/but-workspace/src/commit/sign_commits.rs
```diff
@@ -0,0 +1,69 @@
...
+            let mut pick = Pick::new_pick(graph_commit.id);
+            pick.sign_commit = SignCommit::Yes;
+            pick.pick_mode = PickMode::Force;
+            editor.replace(selector, Step::Pick(pick))?;
+        }
+    }
+
+    let rebase = editor.rebase()?;
+    Ok(SignCommitsOutcome { rebase })
+}
```
slarse, line 69: This represents what I want to do, but I'm not sure where to put this or how to split this. I can't pass _in_ an editor both because I also need the graph to resolve the segment, but also because the editor must be instantiated in a particular way (`SignCommit::Yes`) for this to work as expected. So passing in an editor seems like the wrong thing to do; if it's not instantiated correctly the result won't be the expected one.

So while on the one hand I want to split this into "find" and "execute", I don't think it's the right way to go about it and this should probably be just a single, isolated operation.

crates/but/src/command/legacy/push.rs
```diff
@@ -663,6 +669,16 @@ fn push_single_branch_impl(
...
 
+    if args.sign_commits {
+        let (repo, mut ws, _db) = ctx.workspace_mut_and_db_with_perm(perm)?;
+        let mut meta = ctx.meta()?;
+        let ref_name: gix::refs::FullName = format!("refs/heads/{branch_name}").try_into()?;
+        let outcome =
+            but_workspace::commit::sign_commits::sign_commits(ref_name, &repo, &mut ws, &mut meta)?;
+
+        outcome.rebase.materialize()?;
+    }
```
slarse, line 644: Probably need to fetch here to minimize the risk of signing already integrated commits.

PR #14046, opened 2026-06-03
Add undo/redo to native menu bar

OliverJAsh:
Closes GB-1531


samhh:
I still feel like this isn't very relevant to dogfooding and it'd be easier to put it off until we're revisiting commands. Don't mind if you'd rather merge though!

PR #14051, opened 2026-06-03
Replace legacy CLI stack APIs with head info

krlvi:
image


Byron:
This is a great step in the right direction @krlvi.

I just wanted to mention that we should use the `ctx.ws` (`but_graph::Workspace`) types where possible. And for code that is only interested in what a workspace generally looks like, this is the case.
That way, only the 'minimal' and most stable data-type is used, avoiding `head_info()`  which can easily trigger long-running operations which aren't cached and generally are still to be figured out beyond the PoC that we have done some time ago.

PR #14059, opened 2026-06-03
Clarify comment regarding `remove_branch` restrictions

OliverJAsh:
As far as I can see the "not the top branch" part of this comment is incorrect. @estib-vega Can you confirm? The comment originates from 232cb1fe7b9a264f9cc5be9f3afbc11591a3a4d2.

### DIFF
--- a/crates/but-api/src/legacy/stack.rs
+++ b/crates/but-api/src/legacy/stack.rs
@@ -243,7 +243,7 @@ pub fn remove_branch_only(
 /// This acquires exclusive worktree access from `ctx` before creating the
 /// removal snapshot and detaching the branch.
 ///
-/// This can only be called on a branch that's inside of a stack of multiple branches and is not the top branch,
+/// This can only be called on a branch that's inside of a stack of multiple branches,
 /// or on a branch that's empty.
 #[but_api(napi)]
 #[instrument(err(Debug))]
--- a/packages/but-sdk/src/generated/index.d.ts
+++ b/packages/but-sdk/src/generated/index.d.ts
@@ -272,7 +272,7 @@ export declare function pushStack(projectId: string, stackId: string, withForce:
  * This acquires exclusive worktree access from `ctx` before creating the
  * removal snapshot and detaching the branch.
  *
- * This can only be called on a branch that's inside of a stack of multiple branches and is not the top branch,
+ * This can only be called on a branch that's inside of a stack of multiple branches,
  * or on a branch that's empty.
  */
 export declare function removeBranch(projectId: string, stackId: string, branchName: string): Promise<void>


estib-vega:
Not super sure about whether it will raise an error, but this probably refers to the previous inability of handling anonymous segments.

I think that's no longer as big as a problem now aside from downstream issues (how do you push and so).

OliverJAsh:
> Not super sure about whether it will raise an error,

I interpreted the comment to mean you can't delete the top, but it looks like you can without errors, in both Lite and Tauri:

https://github.com/user-attachments/assets/76fdc509-2a66-4b7d-bea6-e341610aa711

estib-vega:
Interesting, I wonder how the git graph looks like. The code of that API only removes the reference, but according to the application the commits are somehow absorbed into the parent branch after deleting the child branch.

OliverJAsh:
It looks like it's moving the `feature-parent` reference as well as removing `feature-child-1`. Is this a bug?

Before:

```
┊╭┄ch [feature-child-1]
┊●   0d59452 update child-1.txt (2)
┊●   bb21ef1 update child-1.txt (1)
┊●   2fac80a add child-1.txt
┊│
┊├┄pa [feature-parent]
┊●   85fdf93 update parent.txt (2)
┊●   2675226 update parent.txt (1)
┊●   078d4c9 add parent.txt
├╯
```

After:

```
┊╭┄at [feature-parent]
┊●   0d59452 update child-1.txt (2)
┊●   bb21ef1 update child-1.txt (1)
┊●   2fac80a add child-1.txt
┊●   85fdf93 update parent.txt (2)
┊●   2675226 update parent.txt (1)
┊●   078d4c9 add parent.txt
├╯
```

estib-vega:
**If that's how it looks in git as well**, it's seems like a *feature* bug to me hahah.

For sure it's unintended behavior: The intention is only to remove the reference, in which case I would expect for the commits in child to be left as 'dangling commits' in git. Or at least directly under the `gitbutler/workspace` reference.

That's why the comment also talks about the branch wither not being the top-most or is empty. If it's empty, there would be no dangling commits left.

---

In any case, I think I'd leave the comment there.
In the tauri application we had the FE guard against that, we could also guard against that in the BE, but ideally this is just not an issue in the coming iterations of the graph-friendly-workspace

OliverJAsh:
I could add a guard to Lite, but it looks like Tauri doesn't have a guard for it 🤔

estib-vega:
Dang, you are right, it seems we removed that condition

OliverJAsh:
For my own future reference: the conversation continued [here in Discord](<https://discord.com/channels/1060193121130000425/1512017636270276700/1512017642863591494>), https://github.com/gitbutlerapp/gitbutler/pull/14070, and https://github.com/gitbutlerapp/gitbutler/pull/14072.

PR #14061, opened 2026-06-03
feat(cli)!: remove `--json` in favor of `--format json`

davidpdrsn:
Removes the global `--json` flag in favor of using `--format json`. That way, there is only one way to specify the format.

@slarse and I think not having `--format` as a global flag is good since it might not make sense for all commands. Instead, the commands that support format should handle it specifically. This PR doesn't change that; `--format` is still global, but we'll look into it later. But to move in that direction, I added

```rust
#[derive(clap::Args)]
pub struct OutputFormatArg {
    #[clap(...)]
    pub format: OutputFormat,
}
```

which can be `#[clap(flatten)]`'ed into other arg structs later. Currently, it's just used in the top-level `Args` type.

### DIFF
--- a/apps/web/src/routes/cli/scripts.json
+++ b/apps/web/src/routes/cli/scripts.json
@@ -230,11 +230,11 @@
 		"script": [
 			{
 				"type": "input",
-				"lines": ["# Everything takes --json"]
+				"lines": ["# Everything takes --format json"]
 			},
 			{
 				"type": "input",
-				"lines": ["$ but --json branch new feature/ai-integration | jq ."]
+				"lines": ["$ but --format json branch new feature/ai-integration | jq ."]
 			},
 			{
 				"type": "output",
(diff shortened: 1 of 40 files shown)


krlvi (approved):
This is fine. The agents are not using the `--json` flag but it is still useful for scripting. 
Just dont forget to update the docs http://github.com/gitbutlerapp/gitbutler-docs

slarse:
> This is fine. The agents are not using the `--json` flag but it is still useful for scripting. Just dont forget to update the docs [http://github.com/gitbutlerapp/gitbutler-docs](https://github.com/gitbutlerapp/gitbutler-docs)

We shouldn't update the docs before release, should we? Updating docs should be automated but it should never happen before release unless we have an nightly docs subdomain or similar. Which we don't?

davidpdrsn:
I've opened https://github.com/gitbutlerapp/gitbutler-docs/pull/198 which we can merge when making a new release.

PR #14062, opened 2026-06-03
Lite: simplify navigation index

OliverJAsh:


OliverJAsh:
The AI comments are correct, but it's a trade-off I'm willing to make. We can come back to it.

PR #14069, opened 2026-06-04
Enable `eslint-plugin-react-you-might-not-need-an-effect`

OliverJAsh:
https://github.com/nickjvandyke/eslint-plugin-react-you-might-not-need-an-effect


samhh (commented):
I was hoping that Oxlint's built react-hooks rules would become robust, like the React ones, and then we could justify dropping the JS plugins that slow down the linter. Then again Lite is currently pretty small and still fast enough to lint so it's up to you. 🙂

OliverJAsh:
Let's try it, I'm open to removing it if it becomes a problem or doesn't bring value. It already produced an error for the selection `useEffect` we had previously…

PR #14070, opened 2026-06-04
fix: Re-add the deletion guard in the top-most branch

estib-vega:
If a branch is at the top of a multi-branch stack, and it has commits in
it, its reference can’t be deleted.


estib-vega:
Waiting on confirmation about whether we want to reintroduce this guard or not

PR #14075, opened 2026-06-04
but status: Display worktree conflicts

estib-vega:
Display the incomming conflicts that would be brought by integrating
upstream into the project.

### DIFF
--- a/crates/but/src/command/legacy/status/json.rs
+++ b/crates/but/src/command/legacy/status/json.rs
@@ -12,6 +12,7 @@
 use std::collections::HashMap;
 
 use anyhow::Context as _;
+use bstr::ByteSlice;
 use but_graph::SegmentIndex;
 use but_workspace::ref_info::LocalCommit;
 use chrono::{DateTime, Utc};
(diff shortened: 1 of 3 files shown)


estib-vega:
Waiting on whether we're ready to merge this change. Big changes are coming, so might as well wait and rebase

slarse:
@estib-vega I'm so sorry, this completely flew under my radar.

I think @davidpdrsn, you and I should have a li'l chat about how this should be represented before we pull it in.

estib-vega:
> @estib-vega I'm so sorry, this completely flew under my radar.
> 
> I think @davidpdrsn, you and I should have a li'l chat about how this should be represented before we pull it in.

No worries! Yup I just put this in in response to what Scott mentioned, but I think he wanted slightly different (maybe in addition to this):
An after-the-fact warning about the fact that we wrote conflicts into his uncommitted changes.

I think the bigger solution we actually are going to go for is to be able to **display uncommitted conflicts** in the actual uncommitted file list.

PR #14081, opened 2026-06-04
Lite: remount `Diff` when outline selection changes to reset scroll

OliverJAsh:


samhh (approved):
My only thought is that it'd possibly be nice to retain scroll position per at least unchanged outline selection e.g. select A, scroll a bit, select B, select A again (unchanged), be back where you started. But this PR is an improvement over what we have right now regardless.

PR #14083, opened 2026-06-04
Some plain git mode gui follow-ups

mtsgrd:
A few things that were amended but not pushed when I merged last pr, and a few copilot comments addressed.

### DIFF
--- a/crates/but-workspace/src/legacy/ui.rs
+++ b/crates/but-workspace/src/legacy/ui.rs
@@ -77,9 +77,14 @@ but_schemars::register_sdk_type!(StackEntry);
 /// **Temporary type to help transitioning to the optional version of stack-entry** and ultimately, to [`crate::RefInfo`].
 /// WARNING: for use by parts in the code that can rely on having a non-optional `stack_id`. The goal is to have none of these.
 #[derive(Debug, Clone, Serialize)]
+#[cfg_attr(feature = "export-schema", derive(schemars::JsonSchema))]
 #[serde(rename_all = "camelCase")]
 pub struct StackEntryNoOpt {
     /// The ID of the stack.
+    #[cfg_attr(
+        feature = "export-schema",
+        schemars(schema_with = "but_schemars::stack_id")
+    )]
     pub id: StackId,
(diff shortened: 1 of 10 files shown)


packages/but-sdk/src/generated/index.d.ts
```diff
@@ -1898,6 +1898,27 @@ export type StackEntry = {
   isCheckedOut: boolean;
 };
 
+/**
+ * **Temporary type to help transitioning to the optional version of stack-entry** and ultimately, to [`crate::RefInfo`].
+ * WARNING: for use by parts in the code that can rely on having a non-optional `stack_id`. The goal is to have none of these.
+ */
+export type StackEntryNoOpt = {
+  /** The ID of the stack. */
+  id: string;
```
mtsgrd, line 1907: @Byron I don't fully understand this comment, so just pinging you in case we don't want to be exporting this type to .ts?

crates/but-workspace/src/legacy/ui.rs
```diff
@@ -77,22 +77,33 @@ but_schemars::register_sdk_type!(StackEntry);
...
+        schemars(schema_with = "but_schemars::object_id")
+    )]
     pub tip: gix::ObjectId,
     /// The zero-based index for sorting stacks.
     pub order: Option<usize>,
     /// If `true`, then any head in this stack is checked directly so `HEAD` points to it, and this is only ever `true` for a single stack.
     pub is_checked_out: bool,
 }
+#[cfg(feature = "export-schema")]
+but_schemars::register_sdk_type!(StackEntryNoOpt);
```
estib-vega, line 106: This is the only part that needs code-owners approval. Exporting a type to the SDK is fine 👍

estib-vega (approved):

packages/but-sdk/src/generated/index.d.ts
Byron: While it's used it can be exported as well, that should be fine. But once `head_info()` is used everywhere it shouldn't be needed anymore.

PR #14092, opened 2026-06-05
Fix HMR page reloads in Lite

samhh:
See commit.


slarse (approved):
Curious, I didn't notice any issues with HMR when I played around with this. Is a page reload somewhat rare, perhaps?

samhh:
There are many failure modes for HMR to the extent that there are linters to encourage HMR-friendly code, for example only exporting components in component modules. Some libraries also offer bundler integration to specifically support refresh compat.

PR #14099, opened 2026-06-05
Allow left/right to navigate between panes

krlvi:
wdyt?

### DIFF
--- a/apps/lite/ui/src/selection-scopes.ts
+++ b/apps/lite/ui/src/selection-scopes.ts
@@ -4,6 +4,7 @@ import { keyboardTransferOperationMode } from "#ui/outline/mode.ts";
 import { fileOperand, operandIdentityKey, type FileOperand, type Operand } from "#ui/operands.ts";
 import {
 	projectActions,
+	selectProjectFilesVisible,
 	selectProjectOutlineModeState,
 	selectProjectSelectionFiles,
 	selectProjectSelectionOutline,
@@ -114,6 +115,7 @@ export const useNavigationIndexHotkeys = <T>({
 	getKey: (item: T) => string;
 }) => {
 	const dispatch = useAppDispatch();
+	const filesVisible = useAppSelector((state) => selectProjectFilesVisible(state, projectId));
 
 	const selectAndFocus = (newItem: T) => {
 		select(newItem);
@@ -180,6 +182,28 @@ export const useNavigationIndexHotkeys = <T>({
 	};
 
 	useHotkeys([
+		{
+			hotkey: "ArrowLeft",
+			callback: () => {
+				focusAdjacentSelectionScope(filesVisible, -1);
+			},
+			options: {
+				conflictBehavior: "allow",
+				target: ref,
+				meta: { group: "Selection scopes", name: "Focus previous selection scope" },
+			},
+		},
+		{
+			hotkey: "ArrowRight",
+			callback: () => {
+				focusAdjacentSelectionScope(filesVisible, 1);
+			},
+			options: {
+				conflictBehavior: "allow",
+				target: ref,
+				meta: { group: "Selection scopes", name: "Focus next selection scope" },
+			},
+		},
 		{
 			hotkey: "ArrowUp",
 			callback: selectPreviousItem,


OliverJAsh:
I’m imagining that when we add the ability to view files as a tree rather than list, we’ll use left/right shortcuts for expanding/collapsing folders and jumping selection to the parent, [as per the AGP tree view pattern](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/#keyboardinteraction). Similarly for the outline tree where we’ll eventually have the ability to expand/collapse stacks, upstream changes, and maybe branches. I’m keen to make sure we follow well established accessibility patterns.

I wonder if there’s another shortcut we could use. It would be nice to have a shortcut that works even if you’re focused inside a textarea e.g. commit message.

Perhaps we could also have shortcuts for “focus outline”, “focus files” and “focus diff” in addition to the direction shortcuts. Lazygit for example has cmd+n where n is the panel number.

krlvi:
per the discussion we had on this over video the other day - I really don't know. 
The part that bothers me is the inconsistency with HJKL - it is a common convention that those are equivalent to arrow keys. In my usage of Lite, i am often very confused with left / right not working consistently with HJKL. 

I think in the typical workflow it is much more common to want to navigate between commits, and then go into a commit take a look around, up and down and go back to check another commit. And doing that with the arrow cluster is convenient.  What do you think we look into another shortkey patter to use for expanding /collapsing folders (which I think is objectively the less frequent operation in relation the main navigation)

OliverJAsh:
I agree there's an inconsistency there, but I think it may have been a mistake to use `l`/`h` for moving focus between "panels" in the first place, so I would frame it the other way around: we should fix the inconsistency by changing the `l`/`h` shortcuts to something else.

I'm open minded but I feel pretty strongly that left/right arrow keys should be used as described in the tree view pattern docs I linked to.

If we give this some more thought I think we can find something that's just as convenient.

I wonder if these shortcuts could be framed in terms of "ascend/descend the hierarchy" rather than "move selection left/right". Lazygit for example has `enter` for descend (e.g. commit -> file -> hunk), and `escape` to ascend. We could do that if we change reword from `enter` to something else like `r`.

Note left/right isn't just for expanding/collapsing but also jumping to the parent, e.g. pressing left on a commit should jump to parent branch. See VS Code's file explorer as an example. We currently have `shift+up`/`shift+k` for that, but those will have to be removed when we add multi-selection.

krlvi:
The difference between vscode file viewer and our app though is that we are deliberately choosing not to expand content vertically in-line with the list. E.g. we "expand" the commits's content/files to the right. 

Because of this i think the file tree comparison is incorrect. What we are doing is more similar to the MacOS finder with multiple columns represent the hierarchy in the data

image

And in this view the behavior of the arrow cluster the way you would expect from it. 

Because my monkey brain has been trained by patterns in existing applications, using the Lite app sometimes can feel _highly frustrating_, because that pattern and expectation of symmetry in ←↑↓→

File trees are not exactly primary to the app data hierarchy because:
- they appear only in the second pane (not in the left-most lane like in a file browser)
- even in the second pane, files may be showed flattened (as a plain list) and files can even be rendered in-line (when space is limited or the user prefers it)


> if we change reword from enter to something else like r

I think this would be a good idea to do for the defaults of the app no matter what. It's all too easy to trigger this by accident and it's a friction point

OliverJAsh:
It’s not just the file tree, it’s also the outline in the left panel. This is also a ‘tree view’, e.g., stacks, branches and upstream changes being branches in the tree.

I think what we have is more of a mix of Finder’s tree view and the one you showed.

krlvi:
I dont want to necessarily prescribe a solution, as im not really an UX expert. And perhaps this was a bit of a reaction out of usage-induced frustration. At the end of the day we wanna make a too that is dead obvious to use. 

I still press 'enter' on commits and it starting a reword is something i run into very consistently and it is very disruptive - hence I try arrow keys and they don't feel intuitive either. Imo, this is the type of stuff we will only get right through usage and iteration, and this is perhaps the main thing to get to

OliverJAsh:
I just thought, we’ll probably have expand/collapse in the diff as well.

> I still press 'enter' on commits and it starting a reword is something i run into very consistently and it is very disruptive

I’m curious, when you press enter, what are you expecting to happen?

krlvi:
> I just thought, we’ll probably have expand/collapse in the diff as well.
> 
> > I still press 'enter' on commits and it starting a reword is something i run into very consistently and it is very disruptive
> 
> I’m curious, when you press enter, what are you expecting to happen?

probably to "open" it / navigate into it. Not sure if arrow cluster or enter is the way to go but right now it feels very frustrating for me to use

PR #14102, opened 2026-06-07
Refresh workspace on a metadata sentinel file

mtsgrd:
## What

Changes from separate processes (e.g. cli) now show up in the desktop app automatically, with no manual refresh.

## How

- **Signal** — at the single metadata-write choke point, GitButler writers touch a sentinel file (`.git/gitbutler/REFRESH`), so *every* mutation (GUI or CLI) is announced.
- **React** — the desktop file watcher watches it and emits one coalesced `WorkspaceActivity` event, invalidating the stack/branch caches so fresh data is re-read from disk — no forge round-trip.
- **No echo** — the sentinel carries the writing process's id, so the watcher ignores the GUI's own writes (already applied in-process) while still reacting to the `but` CLI or a second window.

## Decoupled from the storage backend

Virtual-branch state is migrating from `virtual_branches.toml` into `but.sqlite` (WAL mode), whose main file doesn't update reliably on write. A writer-touched sentinel keeps the refresh signal independent of where the data actually lives.

## Also

- Renames `Change::GitRemoteActivity` → `WorkspaceActivity` (it now covers metadata writes, not just remote-ref updates).


mtsgrd:
@slarse I opted for a per process token for echo suppression over CLI writes only, allowing e.g. nightly to trigger refresh in official release as well. Wdyt?

crates/but-project-handle/src/storage_path.rs
```diff
@@ -110,8 +110,40 @@ fn validate_in_git_storage_path(
...
+/// touch. Errors are logged and swallowed, never failing the mutation.
+pub fn touch_refresh_sentinel(metadata_file: &Path) {
+    let Some(dir) = metadata_file.parent() else {
+        return;
+    };
+    let Some(filename) = Path::new(REFRESH_SENTINEL_PATH).file_name() else {
+        return;
+    };
+    let sentinel = dir.join(filename);
+    if let Err(err) = std::fs::write(&sentinel, process_sentinel_token()) {
```
slarse, line 144: I'm a bit uncertain about one thing here. `std::fs::write()` is not atomic, meaning that multiple writers may garble each others' outputs so the token doesn't match either. If those concurrent writes are different GitButler instances, then I don't see it as an issue: the resulting token doesn't match either instance and so they both refresh (correctly).

The garbled write case that might be problematic is if it's a single GitButler instance that concurrently performs multiple operations, concurrently writing this file within that single instance, garbling the output s.t. it doesn't match its own token, causing an unwarranted refresh. Atomic writes would prevent that.

If we write to this file with relatively low frequency (which I think is the case) or intra-process concurrent writes aren't possible (which I'm not sure if they are), this problem is probably purely academic and can be ignored. But if we start writing with high frequency and concurrently within a single process, it may very well cause "self-refreshing".

crates/but-project-handle/src/storage_path.rs
```diff
@@ -110,8 +110,40 @@ fn validate_in_git_storage_path(
...
+/// compare right after the watcher fires on a fresh write.
+pub fn process_sentinel_token() -> String {
+    std::process::id().to_string()
+}
+
+/// Best-effort touch of the refresh sentinel in `metadata_file`'s directory —
+/// how out-of-process writes (notably the `but` CLI) reach the desktop watcher.
+/// Contents are this process's [`process_sentinel_token`] so it can skip its own
+/// touch. Errors are logged and swallowed, never failing the mutation.
+pub fn touch_refresh_sentinel(metadata_file: &Path) {
```
slarse: `touch` is very specific in UNIX-land, I think I'd call this something else. `write_refresh_sentinel()`, perhaps.

crates/gitbutler-watcher/src/handler.rs
```diff
@@ -187,9 +188,21 @@ impl Handler {
...
+                REFRESH_SENTINEL_PATH => {
+                    // Skip the echo of our own write (its pid is already handled
+                    // in-process); a different pid, or a read error, means refresh.
+                    let sentinel = ctx.repo.get()?.path().join(file_name);
+                    let wrote_by_us = std::fs::read_to_string(sentinel)
+                        .map(|content| content.trim() == process_sentinel_token())
+                        .unwrap_or(false);
+                    if !wrote_by_us {
+                        saw_workspace_activity = true;
+                    }
```
slarse, line 205: This has an obvious race condition in that in between the event firing and us reading the file, we might've written to the file again. If that write was performed by the "current process", it'll just read its own token and potentially miss a refresh.

This does not seem like a big issue to me for an initial stab at this, because the worst case is that He app works just like it did before (not refreshing on "external writes"). But if we decide that missing refreshes is not acceptable, this approach is not the way to go.

My intuition is however that this will work just fine in practice.

slarse (approved):
I think this looks like a good first attempt! I've noted some theoretical problems with the approach, but I think we should still try this out as my intuition is that this will most likely work well in practice.

crates/but-project-handle/src/storage_path.rs
```diff
@@ -110,8 +110,40 @@ fn validate_in_git_storage_path(
...
+
+/// Identity written into the refresh sentinel so a process can skip the touch
+/// it caused (its own writes are already handled in-process) while still
+/// reacting to others — the `but` CLI, or a second GUI window.
+///
+/// A pid suffices: a recycled one can't be misread as ours, since we only
+/// compare right after the watcher fires on a fresh write.
+pub fn process_sentinel_token() -> String {
+    std::process::id().to_string()
+}
```
slarse: This probably holds for the GUI assuming that we have a single process for all mutations (seems reasonable). If we want to use this same thing for the TUI there may be issues with the command mode (used for e.g. pushing and pulling atm), as that executes `but` commands as sub processes (i.e. different PIDs).

krlvi (approved):

crates/but-project-handle/src/storage_path.rs
mtsgrd: thanks, done!

crates/but-project-handle/src/storage_path.rs
mtsgrd: A'ight, good to know. @davidpdrsn do you have anything you want to chime in with here? It would be great if we can use the same signal in the tui.

mtsgrd:
Thanks @slarse! I agree with your theoretical points, and also agree it's most likely fine. Any race or overwrite is self-healing, so it'd have to get concurrent very often before it becomes an issue. I'll go ahead and land and we can fix forward if necessary!

PR #14105, opened 2026-06-07
Fix phantom uncommitted changes from stale workspace merge-base

mtsgrd:
## What used to go wrong

The workspace tree is rebuilt by merging the applied stacks. Both rebuild paths — `remerged_workspace_tree_v2` (workspace-HEAD updates) and `merge_workspace` (integrate-upstream) — **seeded the merge with the target tree** and merged each stack onto it.

That breaks the invariant that the workspace tree is a faithful merge of the **stack heads** (the workspace commit's parents). It bites whenever a stack's base has drifted from the persisted target (the `[default_target]` commit in `virtual_branches.toml`, ≈ your local `master`), in two ways — each surfacing as phantom uncommitted changes:

```
o──o──T   (persisted target)
│
└─ stack  (forked below T)
```

1. **Stack below the target loses upstream commits.** With T's tree as the merge base, commits the stack predates look like deletions the stack made → stripped from the workspace tree. (This was the refactor-disappearing case.)
2. **Target ahead of all stacks injects target-only content.** If T has files no stack carries, seeding from T puts them in the workspace tree even though no parent has them.

Either way HEAD's tree stops matching the worktree, so `git status` reports the difference as edits nobody made. Trigger: an op that rebuilds the workspace HEAD — most often an amend on an *unrelated* stack.

## How it's fixed

Octopus-merge the **stack heads only** — the target is never a merge input (only the empty-workspace fallback). Every head is merged against a **single shared base: the octopus merge-base (lowest common ancestor) of all heads**, which keeps the result independent of stack order and turns genuine disagreements into conflicts rather than order-dependent picks.

`WorkspaceState` now holds commit ids (`heads` + `target`) instead of pre-resolved trees; its trivial constructor is `from_heads_and_target`.

> **Why a single base is enough (no recursive/virtual base):** a virtual base would only matter if two stack heads had *multiple* merge-bases (a criss-cross). Cross-stack merges don't exist, so every common ancestor of two heads lives in the target history — and when stacks share a fork point that point dominates everything below, giving exactly one merge-base. The only way to get two is the doubly-anomalous case of stacks forked at *different* target commits **and** a criss-cross in the target's own history straddling those forks.

## Tests

`diverged-stacks` fixture — one change per commit, no reverts; `data.txt` has three slots:

```
A {AAA,BBB,CCC}  base = merge-base of the two stacks
├─ B {XXX,BBB,CCC}  target            (slot1 AAA→XXX)
│  └─ stack_c {XXX,YYY,CCC}  on target (slot2 BBB→YYY)
└─ stack_d {AAA,BBB,ZZZ}  below target (slot3 CCC→ZZZ)
```

| test | guards |
|---|---|
| `merge_workspace_merges_stack_heads_only` | (1) merge is `{XXX,YYY,ZZZ}`; `stack_d` must not resurrect the phantom `AAA` the target introduced |
| `update_workspace_commit_merges_stack_heads_only` | (1) same, end-to-end through the workspace-HEAD path |
| `merge_workspace_excludes_target_only_content` | (2) a target-only file (`only_upstream.txt`) must not leak in |

Each fails on the pre-fix target-seeded behavior (verified by reverting).

### DIFF
--- a/crates/but-workspace/src/legacy/head.rs
+++ b/crates/but-workspace/src/legacy/head.rs
@@ -52,36 +52,54 @@ pub fn remerged_workspace_tree_v2(
     let target_base_oid = ctx.persisted_default_target()?.sha;
     let mut stacks: Vec<Stack> = vb_state.list_stacks_in_workspace()?;
 
-    let target_commit = repo.find_commit(target_base_oid)?;
-    let mut workspace_tree_id = repo
-        .find_real_tree(&target_commit, Default::default())?
-        .detach();
-
-    let (merge_options_fail_fast, conflict_kind) = repo.merge_options_fail_fast()?;
-    let merge_tree_id = target_commit.tree_id()?.detach();
-    for stack in stacks.iter_mut() {
-        let stack_head = repo.find_commit(stack.head_oid(ctx)?)?;
-        let branch_tree_id = repo
-            .find_real_tree(&stack_head, Default::default())?
+    let heads = stacks
+        .iter()
+        .map(|s| s.head_oid(ctx))
+        .collect::<Result<Vec<_>>>()?;
+    let workspace_tree_id = if heads.is_empty() {
+        repo.find_real_tree(&repo.find_commit(target_base_oid)?, Default::default())?
+            .detach()
+    } else if heads.len() == 1 {
+        let commit = repo.find_commit(*heads.first().expect("Heads is length 1"))?;
+        let first_head = repo.find_real_tree(&commit, Default::default())?;
+        first_head.detach()
+    } else {
+        let base_tree_id = repo
+            .find_real_tree(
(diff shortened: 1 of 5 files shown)


jonathantanmy2:
> That base is wrong for any stack **forked below** the target:
> 
> ```
> v1 ── v2 ── v3      v3 = target;  x = a line upstream edits: 1 → 2 → 3
> │
> └ stack_a           forked at v1; adds a file, never touches x (so x = 1)
> ```

For what it's worth, I found it difficult to understand this graph (and everything below it until "Symptom"). In particular, `x` is never defined anywhere.

Having said that, I looked at the code and tests and it looks like a straightforward and good fix.

jonathantanmy2 (approved):

mtsgrd:
You're right, I could have done better with the description. :(

PR #14110, opened 2026-06-08
Lite: add fullscreen details

OliverJAsh:
Closes GB-1580


apps/lite/ui/src/components/iconNames.ts
OliverJAsh, line 1: cc @PavelLaptev You may want to change these icons.

PR #14114, opened 2026-06-08
fix(but): reject apply when branch conflicts with workspace stacks

### LINKED ISSUE #14043: CLI apply reports success but leaves outdated branch unapplied
### Version

`but 0.5.2053`

### Operating System

macOS 26.5, Apple Silicon

### Distribution Method

Nightly / CLI

### Describe the issue

When applying an older local branch with `but apply`, the CLI can report success even though the branch remains unapplied and is not added to the workspace commit.

The command exits with status 0 and prints a success message:

```shell
Applied branch 'feature/outdated-branch' to workspace
```

However, checking afterward shows that the branch is still listed under unapplied branches and is absent from the applied workspace state.

### How to reproduce

I do not have a minimal public repository yet, but this reproduced in a repository with several applied virtual branches and one older, non-updated local branch.

1. Start with a GitButler workspace that has multiple applied branches.
2. Have an older local branch, e.g. `feature/outdated-branch`, that is listed under `Unapplied Branches` by `but branch list`.
3. Run:

```shell
but apply feature/outdated-branch
```

4. Observe the command output:

```shell
Applied branch 'feature/outdated-branch' to workspace
```

5. Run:

```shell
but branch list
but status -fv
git log --oneline --decorate --graph --max-count=12 gitbutler/workspace
```

### Actual behavior

`but apply` exits successfully and prints that the branch was applied, but:

- `but branch list` still shows `feature/outdated-branch` under `Unapplied Branches`
- `but status -fv` does not show the branch in the workspace
- `gitbutler/workspace` does not include the branch

### Expected behavior

Either:

- the branch is actually applied to the workspace, or
- the command exits non-zero and reports why it could not apply the branch, such as conflicts, branch needing update, or another workspace state issue

### Notes

This may be related to existing apply/conflict handling issues, but the confusing part here is specifically the CLI success message and zero exit status when the branch remains unapplied.

AasheeshLikePanner:
## 🧢 Changes

Added a guard in the CLI's `apply` command handler (`crates/but/src/command/branch/apply.rs`) that checks `Outcome::conflicting_stack_ids` after calling `apply_with_perm()`. When the vector is non-empty, the branch was not actually applied due to merge conflicts — the CLI now returns a non-zero exit with a clear error message instead of printing "Applied branch" and exiting 0.

Added a regression test (`crates/but/tests/but/command/branch/apply.rs`) that creates a workspace with one branch, then tries to apply a second branch that modifies the same file differently. It asserts the command fails with the expected error.

## ☕️ Reasoning

The workspace-level `Outcome` struct populates `applied_branches` *before* conflict detection runs. In the default `AbortAndReportConflictingStacks` mode, a conflicting merge returns early with `applied_branches` still containing the requested branch — so the CLI thought everything succeeded. The real signal is `conflicting_stack_ids`: when non-empty, the apply was aborted and the branch was left unapplied.

## 🎫 Affected issues

Fixes: #14043


Byron (commented):
Thanks a lot for attempting a fix!

The PR could be merged as is, but I think it's worth it to improve the `apply()` API to make such cases easier to detect.

There are actually two issues here:
* this case isn't correctly documented, and it's unclear that the `apply()` operation can succeed even though there was a conflict because the existing stack conflicted.
* https://github.com/gitbutlerapp/gitbutler/blob/fd8bb0ee4743fdf5d4b1693be39f5ae69ece7f66/crates/but-api/src/branch.rs#L450 is actually incorrect, the original behavior of the clients is to remove the conflicting stacks. If that was changed to [MaterializeAndReportConflictingStacks](https://github.com/gitbutlerapp/gitbutler/blob/fd8bb0ee4743fdf5d4b1693be39f5ae69ece7f66/crates/but-workspace/src/branch/mod.rs#L466), it would work as expected, with the conflicting stack removed. This is what's happening in the [GUI-codepath](https://github.com/gitbutlerapp/gitbutler/blob/fd8bb0ee4743fdf5d4b1693be39f5ae69ece7f66/crates/gitbutler-branch-actions/src/branch_manager/branch_creation.rs#L117-L118), and the CLI should probably be the same.

I will make the respective changes on top of yours and then get the PR merged.

Byron (approved):
@krlvi I think the question here is: Should the CLI also unapply conflicting branches automatically, just like the UI? This is more of a follow-up to this PR as it changes behavior. Also CC @slarse.

It's worth mentioning that this behavior, fail on conflict, was something I seem to have 'decided' in a time when the API function was probably unused.

slarse:
> @krlvi I think the question here is: Should the CLI also unapply conflicting branches automatically, just like the UI? This is more of a follow-up to this PR as it changes behavior. Also CC @slarse.
> 
> It's worth mentioning that this behavior, fail on conflict, was something I seem to have 'decided' in a time when the API function was probably unused.

I'm a bit late to this party, but overall we're going with rejecting instead of performing extra steps the user didn't ask for. This is in part because you don't see the workspace state at all times (like you do with the GUI), and to prevent subtle bugs in scripts. Better to fail hard and require the user to make a choice than to do "something".

For example, the new commit command similarly rejects committing if there are rejected specs, unlike the "old" commit command which just commits what it can and leaves rejected specs untouched.

Byron:
That's good to know @slarse!

Then I recommend to spell out the enum variant explicitly here so it doesn't change if defaults change.

https://github.com/gitbutlerapp/gitbutler/blob/fd8bb0ee4743fdf5d4b1693be39f5ae69ece7f66/crates/but-api/src/branch.rs#L450

Further, it might be worth documenting that this is intended, or maybe pass parameters/options to the API call to let the CLI take control over this. Right now it stands on shaky ground 😅.

PR #14128, opened 2026-06-09
poc: time out on waiting for interprocess lock

slarse:
## Background
Currently, we just wait forever when trying to acquire the interprocess file lock. This is very bad for `but` as every single command tries to acquire the lock when setting up the context.

For example, if you committing with `but commit` and then pause midway through writing the message, _any_ `but` command you execute in the same repository will just hang forever waiting for the lock that will never be released. I've done this to myself on a number of occasions, being equally confused each time.

The fact that every command requires an exclusive lock is a problem in and of itself that we should look over, but the fact that there is no timeout on acquiring the lock will remain as a problem regardless of that.

## 🧢 Changes
This is just a POC for a timeout on acquiring the interprocess lock. We of course don't want to have a `panic!()` as a core mechanism, it was just to get the pieces in place.

## Open questions
We don't really know how to proceed with this.

* Is there a different way we could achieve this?
* Do we want a blanket timeout (for GUI and CLI alike), or are there different needs for different clients?
* The code just ignores failures to acquire the lock stemming from errors, this feels a bit off? I can understand that we don't do it if the file system doesn't support locking, but if it does then discarding errors an ignoring the fact that we couldn't get a lock feels very strange.
* Database transactions also effectively take interprocess locks, so we may end up in the same situation there. It may perhaps be that we need to reconsider how we do locking.

### DIFF
--- a/crates/but-core/src/sync.rs
+++ b/crates/but-core/src/sync.rs
@@ -2,10 +2,12 @@ use std::{
     collections::BTreeMap,
     path::{Path, PathBuf},
     sync::Arc,
+    time::{Duration, Instant},
 };
 
 use anyhow::{Context as _, bail};
 use parking_lot::{ArcRwLockReadGuard, ArcRwLockWriteGuard, RawRwLock};
+
 /// The scope of a lock. It can be either on the entire project or on specific operations.
 #[derive(Debug, Clone, Copy, Default)]
 pub enum LockScope {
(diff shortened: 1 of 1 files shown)


Byron (commented):
This is very interesting! Let me share what I know.

You have probably seen the `BUT_WS_LOCK_DEBUG` bypass I put in to be able to debug hangs caused by the same application, which at least shows how terrible "hang forever" is semantically 😅.

What's good about this file-lock is that it can't be stale. What's bad is that it might not work on all filesystems (and that we don't handle its absence properly).

I don't think using the DB's inter-process locking mechanism as proxy makes anything better either (without elaborating in the interest of time).

All I can think of is that this can only be solved on application level. The intuition is like yours: the CLI should't take exclusive locks all the time, and maybe with SBM this requirement can go away as well. There shouldn't be a real need for it, unless the operation itself needs it, of course (but right now, probably even read-only ops will acquire a write-lock during startup).

As an immediate pain-killer for you I can imagine is to update `BUT_WS_LOCK_DEBUG` to support values, and you could implement a value that sets a timeout (instead of panicking immediately).

That way, the PR can be merged without opening a new rabbit hole.

slarse:
> As an immediate pain-killer for you I can imagine is to update `BUT_WS_LOCK_DEBUG` to support values, and you could implement a value that sets a timeout (instead of panicking immediately).

I'm more of a mind to solve this for users who may not be aware this problem exists. I'm now aware and know how to work around it. So I don't think extending this env variable is really all that useful for the problem at large.

> All I can think of is that this can only be solved on application level. The intuition is like yours: the CLI should't take exclusive locks all the time, and maybe with SBM this requirement can go away as well. There shouldn't be a real need for it, unless the operation itself needs it, of course (but right now, probably even read-only ops will acquire a write-lock during startup).

Yeah, on setting up the context I believe every command acquires a write lock. I'll see if I can solve that but it's more of an amplifier for the issue than the core issue itself. As long as we _can_ acquire locks that block other commands, I don't think it's acceptable to just hang and we need to figure something else out.

Byron:
> As long as we can acquire locks that block other commands, I don't think it's acceptable to just hang and we need to figure something else out.

Maybe this approach works: goal is that any interaction doesn't take longer than Xms. Lock acquisition should thus take longer than Y * Xms. If it does, fail loudly as something is probably wrong. That message would then have to say something like "Either another `but` mutation takes a long time, or there is a bug you might want to report at URL".

For the first part, long running conflicting tasks, failing isn't nice, but it's better than hanging forever. Just printing a message to show what's going on is probably more desirable. Then, if it takes even longer, one might print another message saying that it's probably deadlocked somehow. This I'd think should never happen, but would help finding issues like the one with the editor.

PR #14134, opened 2026-06-09
fix(but): ensure cache dir is created

slarse:
For some reason, we never attempted to create the cache dir when creating a new `AppCacheHandle`, but we do it for `CacheHandle`. As a result, `but` would never create the cache directory and therefore never store the update check result.

I see no good reason not to create the cache directory for `AppCacheHandle` as well - checking if the directory is there is very cheap so it's a one-time cost.

### DIFF
--- a/crates/but-db/src/cache/handle.rs
+++ b/crates/but-db/src/cache/handle.rs
@@ -10,6 +10,9 @@ impl AppCacheHandle {
     /// with an infallible constructor that falls back to an in-memory database.
     pub fn new_in_directory(dir: Option<impl AsRef<Path>>) -> Self {
         let db_path = dir.map_or(":memory:".into(), |d| d.as_ref().join("app-cache.sqlite"));
+        if let Some(cache_dir) = db_path.parent() {
+            create_dir_all_log_failure(cache_dir);
+        }
         Self::new_at_path(db_path)
     }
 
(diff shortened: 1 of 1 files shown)


crates/but-db/src/cache/handle.rs
```diff
@@ -44,17 +47,8 @@ impl CacheHandle {
     /// Infallible constructor that opens the project-local cache from `dir`,
     /// with an infallible fallback to an in-memory database.
     pub fn new_in_directory(dir: impl AsRef<Path>) -> Self {
-        let dir = dir.as_ref();
-        if !dir.exists()
-            && let Err(err) = std::fs::create_dir_all(dir)
```
slarse, line 49: Note: `std::fs::create_dir_all()` is idempotent, so there is no need to check if the directory exists before creation.

krlvi (approved):

crates/but-db/src/cache/handle.rs
```diff
@@ -10,6 +10,9 @@ impl AppCacheHandle {
     /// with an infallible constructor that falls back to an in-memory database.
     pub fn new_in_directory(dir: Option<impl AsRef<Path>>) -> Self {
         let db_path = dir.map_or(":memory:".into(), |d| d.as_ref().join("app-cache.sqlite"));
+        if let Some(cache_dir) = db_path.parent() {
+            create_dir_all_log_failure(cache_dir);
+        }
```
Byron, line 15: Nice one!
Interesting that this could stand for so long without being noticed.
Probably this is a macOS thing where this directory is just expected to exist because some other part of GitButler (still) creates it?

In any case, it's good to see that here and the creation unified.

slarse: I think it passed us by as the directory is created by the Tauri app, so most every developer would already have had it. I happened to notice (a few months back, really...) that it didn't exist for the dev channel, where I never used the Tauri app.

PR #14141, opened 2026-06-09
Set env var which allows to read the target stack where the commit will be applied

asam139:
## 🧢 Changes

Set `GITBUTLER_TARGET_STACK` env var which allows to read the target stack where the commit will be applied

## ☕️ Reasoning

This env var allows to be read during hooks, specially commit-msg so the commit can contain the nº of ticket which is a rule in companies

## 🎫 Affected issues

Fixes: [11302](https://github.com/gitbutlerapp/gitbutler/issues/11302)

## 📌 Todos

*  Wait feedback to see if it would be possible to optimize it

|   	|   	|
|---	|---	|
|  Screenshot 2026-06-04 at 14 32 47   |   [commit-msg.txt](https://github.com/user-attachments/files/28761633/commit-msg.txt)	|

### DIFF
--- a/crates/but-api/src/commit/create.rs
+++ b/crates/but-api/src/commit/create.rs
@@ -1,6 +1,7 @@
 use crate::WorkspaceState;
 use but_api_macros::but_api;
 use but_core::{DiffSpec, DryRun, sync::RepoExclusive};
+use but_graph::Workspace;
 use but_oplog::legacy::{OperationKind, SnapshotDetails};
 use but_rebase::graph_rebase::{
     Editor, LookupStep as _,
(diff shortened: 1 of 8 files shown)


Byron:

Unfortunately, I don't think this PR is fit to be reviewed and I believe more analysis is needed to understand how this is best implemented, particularly how to maintain Git compatibility without being incredibly invasive to GitButler.
In the end, I would expect it to be nothing more than setting additional environment variables to support hooks that may know about GitButler calling them instead of just plain Git.

asam139:
@jonathantanmy2 any update, could take a look? I would like to know if solution makes sense or we should discard it at all in favor of other (we really miss this feature)

PR #14143, opened 2026-06-09
upstream integration fixes

estib-vega:
- We should be able to update empty workspaces.
- We should reparent the workspace commit correctly onto the latest tip of the target ref after updating the workspace

---
Basically I didn't account for the scenario of leaving the workspace commit orphan. Now we have the new ref adopt it correctly

### More tests
- More tests in the rust side and on the app e2e

### DIFF
--- a/crates/but-workspace/src/upstream_integration.rs
+++ b/crates/but-workspace/src/upstream_integration.rs
@@ -292,6 +292,11 @@ pub fn integrate_upstream<'ws, 'meta, M: RefMetadata>(
         for selector in &fully_integrated_workspace_parents {
             editor.remove_edges(workspace_commit_selector, *selector)?;
         }
+        if !fully_integrated_workspace_parents.is_empty()
+            && editor.direct_parents(workspace_commit_selector)?.is_empty()
+        {
+            editor.add_edge(workspace_commit_selector, target_ref_selector, 0)?;
+        }
     }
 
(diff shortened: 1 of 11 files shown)


crates/but-workspace/src/upstream_integration.rs
```diff
@@ -292,6 +292,11 @@ pub fn integrate_upstream<'ws, 'meta, M: RefMetadata>(
         for selector in &fully_integrated_workspace_parents {
             editor.remove_edges(workspace_commit_selector, *selector)?;
         }
+        if !fully_integrated_workspace_parents.is_empty()
+            && editor.direct_parents(workspace_commit_selector)?.is_empty()
+        {
+            editor.add_edge(workspace_commit_selector, target_ref_selector, 0)?;
+        }
```
estib-vega, line 299: This is the actual fix. If the workspace loses all its parents, reparent onto the target ref. Duh

Caleb-T-Owens (approved):
lol, nice one

crates/but-workspace/src/upstream_integration.rs
krlvi: i wonder if there are more conditions to account for

estib-vega: I'll make sure there are no other conditions. Will add more tests if I find them

PR #14144, opened 2026-06-09
Fix crash on conflicted binary files in edit mode

### LINKED ISSUE #13412: Conflicted binary file: `TypeError null is not an object (evaluating 's.split')`
### Version

0.19.8

### Operating System

Linux

### Distribution Method

rpm (Linux)

### Describe the issue

I have a conflict on a binary file.

When attempting to resole the conflict / edit the commit, I am unable to "Save changes and exit". I get the following error:

```
null is not an object (evaluating 's.split')
```

Console shows the error here in `conflictEntryPresence.ts` line 34:

```
function looksConflicted(file: string): boolean {
	const lines = file.split("\n");
	for (const line of lines) {
		if (line.startsWith("<<<<<<<")) {
			return true;
		}
	}
	return false;
}
```

_No response_

_No response_

```shell

```

ameyypawar:
Fixes #13412

### Problem

With a conflicted **binary** file present, "Save changes and exit" in edit mode crashes:

```
TypeError: null is not an object (evaluating 's.split')
```

The backend returns `FileInfo.content` as `None` for binary files, since they have no text representation, but the frontend's hand-written `FileInfo` declared `content: string`. That `null` reached `looksConflicted`, which calls `file.split("\n")`, leaving no way to save and exit edit mode.

### Fix

The frontend type was hand-written and had drifted from the backend, so rather than widen it by hand, `FileInfo` is now generated:

- `gitbutler_repo::FileInfo` derives `JsonSchema` behind a new `export-schema` feature and is registered with `but_schemars::register_sdk_type!`, wired through `but-api` so `but-ts` emits it.
- `apps/desktop/src/lib/files/file.ts` is deleted, and `fileService.ts` imports `FileInfo` from `@gitbutler/but-sdk`.
- Generating it also corrects drift the hand-written type had: the wire field is `fileName` and is required, not `name?`, and `size`/`mimeType` are nullable rather than optional.
- `getConflictState` now takes `string | null`, so the null case is driven by the type instead of being guarded against a type that claimed it could not happen.

A conflicted binary file carries no text to scan for conflict markers, so `getConflictState` reports `"unknown"` and saving is not blocked.

### Test

`apps/desktop/src/lib/files/conflictCheck.test.ts` drives `null` content through `hasUnresolvedConflictsOnDisk` with both sides present. It fails against the previous implementation with the same `null`/`split` error and passes with this change.

### Notes for review

- `FileInfo` carries a `// TODO: turn this whole struct into an enum`. Registering it publishes the current shape through the SDK, so if that redesign is near, landing the enum first may be preferable — happy to hold this or split it.
- Not blocking the save for binary files is a deliberate choice; the alternative is to treat them as conflicted and route through the confirm modal. This path previously threw, so there is no prior behaviour to preserve either way.


Byron (changes requested):
Thanks a lot for taking a look.

When looking at the tests, it seems like the typing is off on this one. After all, the backend should use `Option<String>` there and it would pass that on directly.


Here seems to be the culprit:

https://github.com/gitbutlerapp/gitbutler/blob/bf51832eef11e184dbb870d3cb6a8bbcbf0bf14d/apps/desktop/src/lib/files/file.ts#L1-L7

Maybe this is a type we should also generate automatically, and then adjust the code?
CC @estib-vega and @mtsgrd .

PR #14148, opened 2026-06-09
Lite: use checked commits for transfer mode source

OliverJAsh:
Towards GB-1334


OliverJAsh:
Closing for now to work on other more important things.

PR #14149, opened 2026-06-09
Fix stacked-PR footer "part X" numbering to match the position badges

matas-zanevicius:
## What

In a stacked PR, the footer renders a header — "This is **part X of N in a stack**" —
above a list of the stack's PRs, each tagged with a `<kbd>` position badge. The badges
number from the base of the stack (`1`) up to the top (`N`). But the header's `X` was
computed as `stack_length - stack_index`, counting from the *top* instead — so the header
and a PR's own badge disagreed for every PR except the exact middle.

### Before (5-PR stack, viewing the base PR `#29`)

```
This is **part 5 of 5 in a stack** made with GitButler:
- <kbd> 5 </kbd> #34
- <kbd> 4 </kbd> #33
- <kbd> 3 </kbd> #32
- <kbd> 2 </kbd> #30
- <kbd> 1 </kbd> #29 👈
```

`#29` is badge **1**, yet the header says **part 5 of 5**.

### After

```
This is **part 1 of 5 in a stack** made with GitButler:
- <kbd> 5 </kbd> #34
- <kbd> 4 </kbd> #33
- <kbd> 3 </kbd> #32
- <kbd> 2 </kbd> #30
- <kbd> 1 </kbd> #29 👈
```

## Root cause & fix

One line in the Rust generator (`crates/but-forge/src/review.rs`, `generate_footer`):
compute the part number as the PR's 1-based position from the base (`stack_index + 1`)
instead of from the top, so it equals the badge.

## Why the desktop generator is *not* changed

It consumes the stack in the opposite order and already produced the correct output; it's
only exported here so it can be unit-tested.

## Tests

- **Rust:** updated `test_generate_footer_large_stack`; added
  `test_footer_part_number_matches_position_badge`, which asserts for every PR in a stack
  that the header's part number equals that PR's `<kbd>` badge — the invariant that was
  broken (fails on `master`).
- **Desktop:** added `prFooter.test.ts` pinning the generator's contract for the desktop's
  `[top, …, base]` input order — top-first listing, base = badge 1, header matching the badge.

### DIFF
--- a/apps/desktop/src/lib/forge/shared/prFooter.ts
+++ b/apps/desktop/src/lib/forge/shared/prFooter.ts
@@ -155,14 +155,15 @@ function clearFooter(body: string | undefined) {
 function generateFooter(forPrNumber: number, allPrNumbers: number[], symbol: string) {
 	const stackLength = allPrNumbers.length;
 	const stackIndex = allPrNumbers.findIndex((number) => number === forPrNumber);
-	const nth = stackLength - stackIndex;
+	const nth = stackIndex + 1;
 	let footer = "";
 	footer += STACKING_FOOTER_BOUNDARY_TOP + "\n";
 	footer += "---\n";
 	footer += `This is **part ${nth} of ${stackLength} in a stack** made with GitButler:\n`;
-	allPrNumbers.forEach((prNumber, i) => {
-		const current = i === stackIndex;
-		footer += `- <kbd>&nbsp;${stackLength - i}&nbsp;</kbd> ${symbol}${prNumber} ${current ? "👈 " : ""}\n`;
+	[...allPrNumbers].reverse().forEach((prNumber, i) => {
+		const position = stackLength - i;
+		const current = prNumber === forPrNumber;
+		footer += `- <kbd>&nbsp;${position}&nbsp;</kbd> ${symbol}${prNumber}${current ? " 👈 " : ""}\n`;
 	});
 	footer += STACKING_FOOTER_BOUNDARY_BOTTOM;
 	return footer;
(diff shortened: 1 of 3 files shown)


apps/desktop/src/lib/forge/shared/prFooter.ts
```diff
@@ -155,14 +155,15 @@ function clearFooter(body: string | undefined) {
...
 	footer += STACKING_FOOTER_BOUNDARY_TOP + "\n";
 	footer += "---\n";
 	footer += `This is **part ${nth} of ${stackLength} in a stack** made with GitButler:\n`;
-	allPrNumbers.forEach((prNumber, i) => {
-		const current = i === stackIndex;
-		footer += `- <kbd>&nbsp;${stackLength - i}&nbsp;</kbd> ${symbol}${prNumber} ${current ? "👈 " : ""}\n`;
+	[...allPrNumbers].reverse().forEach((prNumber, i) => {
+		const position = stackLength - i;
+		const current = prNumber === forPrNumber;
+		footer += `- <kbd>&nbsp;${position}&nbsp;</kbd> ${symbol}${prNumber}${current ? " 👈 " : ""}\n`;
```
estib-vega: This seems to have the wrong assumption, in the FE the order that we receive the PRs in are child to parent: [top, ..., base]

While in rust is in the push order: [base, ..., top]. 

Could you add a unit test for this, ensuring that you get the right order of the footers given the expected segment order as described above?

estib-vega (approved):
Thank you!

apps/desktop/src/lib/forge/shared/prFooter.ts
matas-zanevicius: @estib-vega Thanks for pointing out my misunderstanding — you're right. I've updated the code accordingly and added the unit test you requested.

PR #14153, opened 2026-06-10
fix(but): improve error message when branch collides by path prefix

slarse:
If a branch exists at `<path>`, and you create a new branch `<path>/name`, then that collides with the existing branch at `<path>` due to the need to now create a directory at `<path>`. The error message generated in that situation used to be very obtuse, and this commit improves upon it to be more clear.

There is an underlying issue in `gix` where if you do `repo.try_find_reference("<path>/name")` in that same situation, an error is returned with a `NotADirectory` kind. This is explicitly asserted inside gix so need to sync with @Byron if this should be changed in `gix` or if we just work around it here.

We want this nice error message regardless so the code here will probably change minimally regardless.

> **Edit:** I realised this is also a problem for the GUI so we probably need a lower-level fix than this.

### DIFF
--- a/crates/but/src/args/atoms/branch_arg.rs
+++ b/crates/but/src/args/atoms/branch_arg.rs
@@ -1,5 +1,7 @@
+use std::error::Error;
+
 use bstr::{BStr, ByteSlice};
-use gix::refs::{Category, FullName};
+use gix::refs::{Category, FullName, FullNameRef};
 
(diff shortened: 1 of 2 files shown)


Byron:
Thanks for the ping! Indeed, this issue should be fixed in `gitoxide` and I'm working on it.
When it's available I will make sure the fix here is removed (but not the test).

PR #14157, opened 2026-06-10
Clean up edit-mode refs when leaving edit mode

mtsgrd:
### Problem

Entering edit mode creates three artifacts that neither exit path cleaned up, so `gitbutler/edit` lingered after every edit session:

- `refs/heads/gitbutler/edit`
- `refs/gitbutler/edit-uncommitted-changes` (stash ref)
- `edit_mode_metadata.toml`

### Fix

A new `cleanup_edit_mode` helper deletes all three, called at the end of both `save_and_return_to_workspace` and `abort_and_return_to_workspace` — **after** `HEAD` is back on the workspace branch, so removing the edit ref can't leave `HEAD` dangling.

Deletion tolerates already-absent targets, and a forced abort still bails before cleanup when there are unsaved changes.


krlvi (approved):
seems reasonable

PR #14162, opened 2026-06-10
lite: askpass + git prompt

estib-vega:
- Initialize the askpass binary for the lite application.
- Setup a listener for the `git_prompt` events.
- Bundle the askpass binary with the Electron app.

### Visuals

Screenshot 2026-06-10 at 14 52 15

  
  Fixes: GB-1245, GB-1246

### DIFF
--- a/apps/lite/scripts/prepare-askpass.mjs
+++ b/apps/lite/scripts/prepare-askpass.mjs
@@ -1,0 +1,46 @@
+import { copyFileSync, chmodSync, mkdirSync } from "node:fs";
+import path from "node:path";
+import { spawnSync } from "node:child_process";
+import { fileURLToPath } from "node:url";
+
+const scriptPath = fileURLToPath(import.meta.url);
+const liteRoot = path.resolve(path.dirname(scriptPath), "..");
+const repoRoot = path.resolve(liteRoot, "../..");
+const release =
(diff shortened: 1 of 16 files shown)


apps/lite/scripts/prepare-askpass.mjs
estib-vega, line 1: @slarse I just wanted to give out a heads up about bundling the askpass with lite, as well as the little changes I did in order to be able to hook-it-up to lite

apps/lite/ui/src/AskpassPromptDialog.tsx
estib-vega, line 1: @OliverJAsh This is the actual component for the dialog. It just opens when there's a new prompt. Feel free to change it as you see fit :)

crates/gitbutler-git/src/repository.rs
```diff
@@ -297,6 +296,21 @@ where
     }
 }
 
+fn askpass_executable_path(current_exe: &Path) -> PathBuf {
+    askpass_executable_path_from_override(std::env::var_os("GITBUTLER_ASKPASS_BIN"), current_exe)
+}
```
estib-vega, line 301: Allow for lite (or other clients) to override the binary path, so that we can point it to the bundled askpass

PR #14163, opened 2026-06-10
fix: Can't move a branch onto itself

estib-vega:
Fail early if we attempt to move a branch onto itself

Fixes GB-1288


crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -166,6 +166,10 @@ pub(super) mod function {
         subject_branch_name: &FullNameRef,
         target_branch_name: &FullNameRef,
     ) -> anyhow::Result<Outcome<'ws, 'meta, M>> {
+        if subject_branch_name == target_branch_name {
+            bail!("Cannot move branch {subject_branch_name} onto itself");
+        }
+
```
estib-vega, line 172: Little safe-guard

krlvi (approved):

PR #14167, opened 2026-06-10
insert segment: Insert as first parents

estib-vega:
By default, insert the segments as first parents when inserting below. If
insertng above, inserts the target as the first parent.

Fixes GB-1525

Implementation notes:
- Added `ParentReparentingOrder` so advanced callers can choose whether reparented parents are prepended or appended.
- `insert_segment` keeps the default behavior simple by using `Prepend`.
- Added parent-order regression coverage for the GB-1525 merge-commit reorder case and explicit append-policy coverage.

### DIFF
--- a/crates/but-rebase/src/graph_rebase/mutate.rs
+++ b/crates/but-rebase/src/graph_rebase/mutate.rs
@@ -30,6 +30,16 @@ pub enum InsertSide {
 #[cfg(feature = "export-schema")]
 but_schemars::register_sdk_type!(InsertSide);
 
+/// Controls where reparented insertion-location parents are ordered relative to
+/// existing parents on the segment.
+#[derive(Debug, Clone)]
+pub enum ParentReparentingOrder {
+    /// Put reparented insertion-location parents before existing segment parents.
+    Prepend,
+    /// Put reparented insertion-location parents after existing segment parents.
+    Append,
+}
+
 /// Defines the start and end of a segment by pointing to it's parent-most and child-most nodes.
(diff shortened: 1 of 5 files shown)


crates/but-rebase/src/graph_rebase/mutate.rs
```diff
@@ -498,6 +508,85 @@ impl<M: RefMetadata> Editor<'_, '_, M> {
         }
     }
 
+    fn add_edges_to_parents(
+        &mut self,
+        child_node: petgraph::prelude::NodeIndex,
+        new_parent_nodes: impl IntoIterator<Item = petgraph::prelude::NodeIndex>,
+        parent_reparenting_order: ParentReparentingOrder,
+    ) {
```
Caleb-T-Owens, line 516: It seems like this function both deduplicates the list of new parents, and the list of old parents, as well as removing duplicate existing parents if they now show up in the new parents.

`git` doesn't allow you to create duplicate parents through the CLI, but it does support it in it's data model, and the stance from the git discord server was that it's probably fine to make such a monstrosity.

I'm wondering if the de-duplication is really needed, or if it might end up causing more confusion in the future

Caleb-T-Owens (approved):
LGTM.

crates/but-rebase/src/graph_rebase/mutate.rs
estib-vega: Interesting, I didn't know that the data model could handle duplicated parents.

I can take it out

PR #14171, opened 2026-06-10
Lite: use file paths instead of file operands

OliverJAsh:
Unsure if we should use the path string or bytes?

### DIFF
--- a/apps/lite/ui/src/projects/state.ts
+++ b/apps/lite/ui/src/projects/state.ts
@@ -1,16 +1,15 @@
-import { createSlice, type PayloadAction } from "@reduxjs/toolkit";
-import type { RootState } from "#ui/store.ts";
-import { type AbsorptionTarget, type RefInfo, type RelativeTo } from "@gitbutler/but-sdk";
 import {
 	type BranchOperand,
 	type CommitOperand,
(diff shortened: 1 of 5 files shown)


Byron:
> Unsure if we should use the path string or bytes?

The `path_bytes` are the identifiers, as the string versions of paths are lossy in nature.

OliverJAsh:
> The `path_bytes` are the identifiers, as the string versions of paths are lossy in nature.

Unfortunately sometimes we only get a path string back from the SDK, e.g. for conflicted files. Branches have a similar issue: [GB-1161: Refs/branches should use bytes instead of strings](<https://linear.app/gitbutler/issue/GB-1161/refsbranches-should-use-bytes-instead-of-strings>)

It would be great to fix those issues and make the SDK consistent in how it identifies branches and paths. 🙏

Byron:
That's fair (and the issue is pretty old/long-known as well).

I wonder if replacing `FileOperand` with a generic `string` is helpful in eventually migrating to something else. If it was `PathDisplay` or something like it it would be easier later to later change it so that `PathDisplay` is never used as `Path` for instance (which probably would be `bytes` and is used for non-lossy identification).

PR #14180, opened 2026-06-11
Fix Zed IDE compatibility (Open in Editor)

sergeykuzmich:
## 🧢 Changes

Add a set of VSCode-compatible IDEs (similar to those in Rust). Also, depending on this condition, the query parameter `?windowId=_blank` is removed because Zed treats it as part of the project path, which prevents proper functionality.

## ☕️ Reasoning

GitButler didn't open project properly in Zed IDE.

## 🎫 Affected issues

Fixes: #7910

### DIFF
--- a/apps/desktop/src/lib/backend/url.test.ts
+++ b/apps/desktop/src/lib/backend/url.test.ts
@@ -1,0 +1,44 @@
+import { getEditorUri } from "$lib/backend/url";
+import { describe, expect, test } from "vitest";
+
+describe("getEditorUri", () => {
+	test.each(["vscode", "vscode-insiders", "vscodium", "cursor", "windsurf", "trae"])(
+		"%s keeps query parameters",
+		(schemeId) => {
+			expect(
+				getEditorUri({
+					schemeId,
+					path: ["/home/example/project"],
+					searchParams: { windowId: "_blank" },
+				}),
+			).toBe(`${schemeId}://file/home/example/project?windowId=_blank`);
+		},
+	);
+
+	test("zed drops query parameters since it treats them as part of the file path", () => {
+		expect(
+			getEditorUri({
+				schemeId: "zed",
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks a lot, this looks great to me!

It's notable that we also have this in the backend now, which is probably what we should switch to in future.

https://github.com/gitbutlerapp/gitbutler/blob/bf51832eef11e184dbb870d3cb6a8bbcbf0bf14d/crates/but-api/src/open/mod.rs#L639-L641

The latter doesn't seem to deal with URL parameters similarly, but maybe it's not necessary either.

Byron:
I validated that Zed now works with the fix applied.

@slarse Would you be interested in bringing the backend implementation for `open_in_editor` full circle and call that instead? This way some code in the frontend can be deleted, and the backend can handle the menu-item invocation fully (I think).

slarse:
> I validated that Zed now works with the fix applied.
> 
> @slarse Would you be interested in bringing the backend implementation for `open_in_editor` full circle and call that instead? This way some code in the frontend can be deleted, and the backend can handle the menu-item invocation fully (I think).

For some context, the `open_in_editor` API is under development and will probably change quite a lot over the coming weeks. It's made for Lite and `but` (well, soon at least), and we don't really have plans to utilize it in the Tauri app. It's still got a whole heap of pitfalls so I wouldn't use it in anger just yet.

As it's also going to change a whole lot over the coming weeks (or days, even), I don't really want to put it into the Tauri app now, or possibly ever given how fast Lite is coming along.

Byron:
> As it's also going to change a whole lot over the coming weeks (or days, even), I don't really want to put it into the Tauri app now, or possibly ever given how fast Lite is coming along.

Thanks, and I see now what the state of the Svelte GUI is. Good to know.

PR #14189, opened 2026-06-11
revive `but worktree` subcommands

krlvi:
@estib-vega @Caleb-T-Owens - this may not be the final form of this, but with this change the worktree stuff seems to be alive once again.


krlvi:
replaced by https://github.com/gitbutlerapp/gitbutler/pull/14213

PR #14197, opened 2026-06-12
but-rebase: write 3+ sides of conflicts (sometimes)

jonathantanmy2:
The start of the effort to support a jj-like conflict tree algebra (with many sides and bases) instead of the current system that only supports at most 2 sides and 1 base.

---

`merge_commit_changes_to_tree()` currently only writes 2 sides + 1 base of a conflict, silently dropping other sides and bases. Teach it not to drop them, and to use all sides and bases of conflicted commits when merging.

This is written in a backwards-compatible way: due to the merging process stopping at the first conflict (as documented in the module-level documentation), the first 2 sides and the first base that are written are the same before and after this commit. The only difference is that additional sides and bases are preserved. So code that only understands 2 sides + 1 base will still work.

This only affects a limited number of cases. I plan to take a look at `merged_tree_from_commits()` next and see if it can be made to share code with this commit.

There was an optimization: when the target commit is conflicted and nothing is to be merged into it, the target commit tree is reused as-is (instead of performing a merge). This optimization is probably not worth the code, since the situation in which it is triggered is so unlikely, so I have not reimplemented it.

### DIFF
--- a/crates/but-core/src/commit/mod.rs
+++ b/crates/but-core/src/commit/mod.rs
@@ -621,6 +621,36 @@ impl<'repo> Commit<'repo> {
         )))
     }
 
+    /// If the commit is not conflicted, returns an empty vec.
+    pub fn base_tree_ids(&self) -> anyhow::Result<Vec<gix::ObjectId>> {
+        if !self.is_conflicted() {
+            return Ok(Vec::new());
+        }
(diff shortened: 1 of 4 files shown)


Caleb-T-Owens (approved):
:shipit:

PR #14200, opened 2026-06-12
Fix custom diff CSS in Lite builds

samhh:
I'm unable to build on the macOS 27 developer beta for some reason, please test this works in my stead. 😅

### DIFF
--- a/apps/lite/electron/src/main.ts
+++ b/apps/lite/electron/src/main.ts
@@ -574,8 +574,7 @@ void app.whenReady().then(async () => {
 		const productionCsp =
 			"default-src 'none';" +
 			"script-src 'self' 'wasm-unsafe-eval';" +
-			// Hash is for inline style in index.html
-			"style-src 'self' 'sha256-XBXaUBQCe+0UGd1QCfoPFCc7UsLKd8xrn9oXNYqjFog=';" +
+			"style-src 'self' 'unsafe-inline';" +
 			"font-src 'self';" +
 			"connect-src 'self';" +
 			"object-src 'none';" +


apps/lite/electron/src/main.ts
```diff
@@ -574,8 +574,7 @@ void app.whenReady().then(async () => {
 		const productionCsp =
 			"default-src 'none';" +
 			"script-src 'self' 'wasm-unsafe-eval';" +
-			// Hash is for inline style in index.html
-			"style-src 'self' 'sha256-XBXaUBQCe+0UGd1QCfoPFCc7UsLKd8xrn9oXNYqjFog=';" +
+			"style-src 'self' 'unsafe-inline';" +
```
slarse, line 577: Could we add a comment saying why we need unsafe-inline? I recall it was some specific library?

I'll give nonce-based CSP a go once I get some time over, and it'll be helpful to know what lib is inserting the inline styles.

I'm confident this will do the trick, don't see a need to test it before merge.

slarse (commented):
> I'm unable to build on the macOS 27 developer beta for some reason, please test this works in my stead. 😅

If I recall correctly, the thing that builds the icons in electron-builder is hard coded to depend on Xcode 26. Might be the issue. Let's see if we can work around it, it's important that you can build the bundle locally.

apps/lite/electron/src/main.ts
samhh: It's the Pierre diff library, same as the WASM one above it.

slarse: Right.

Just keep in mind that while we have `unsafe-inline` here, rendering user-submitted content (e.g. PR stuff) is pretty dangerous. If someone manages to sneak in an inline style, it'll get rendered, which could escalate to other badness (but usually is sufficient in its own right for e.g. phishing and scams like that). I'll put it on my TODO list to get this sorted with nonces, Vite supports it. It was just a heckin' pain to get it working last time I tried it out.

samhh: We can revisit this much closer to release (see also WASM), I'd be keen to avoid or at least defer the DX cost. FWIW you have to do something unusual to wire client content into the markup sans sanitisation, the APIs guide you away from it.

PR #14216, opened 2026-06-13
Return new branch reference in `updateBranchName`

OliverJAsh:
The client needs the new branch name after normalization in order to preserve the user's selection.

I took the liberty of using AI to write the Rust code. It seems simple enough, but feel free to jump on the PR because I have no idea what I'm doing… 🙈


estib-vega (approved):
Makes sense 👍 

Thanks :)

PR #14235, opened 2026-06-15
Don't require metadata when updating workspace

Caleb-T-Owens:
This changes the mechanism for detecting whether individual references are integrated from using the ws_meta to one that does some inspection of the graph.

We want to delete a reference if all the commits that it exclusivly references in the stack have been either historically or content integrated.

We will also delete references if they have been deemed historically integrated and they have no child commits.


Byron:
Yes, `ws_meta` is only needed for building the graph, which reconciles it with reality. From that point on, it's not needed anymore unless it should be changed.

estib-vega (approved):

PR #14236, opened 2026-06-15
but-graph: keep the target in sbm

estib-vega:
### Actual changes
- Preserve the target ref and target sha information, if available in the ProjectMeta when in single-branch mode.
- Prune all integrated commits when in single branch mode.

---
### Important test changes

`but-graph` snapshot updates:
- `just_init_with_branches`: `main` is the workspace base. Snapshot now asserts target context is shown and the integrated base commit is pruned, leaving an empty `main` container.
- `proper_remote_ahead`: checked-out `main` is outside the managed workspace. Snapshot now keeps `origin/main` target context, shows only target-side ahead commits, and prunes integrated local/base commits.
- `integrated_tips_stop_early_if_remote_is_not_configured`: integrated branch `A` downgrades to single-branch view. Snapshots now assert the branch container remains but integrated commits are pruned.
- `integrated_tips_do_not_stop_early`: same downgraded-entrypoint behavior with remote target discovery. Updated `A`, `main`, and detached snapshots to keep target context and prune integrated history.
- `partitions_with_long_and_short_connections_to_each_other`: outside-workspace `main` projection now preserves target context and prunes integrated local/base commits, including the limited traversal variant.
- `partitions_with_long_and_short_connections_to_each_other_part_2`: same rule for the second partition case: keep `main` container, show target-side commits, prune integrated base commits.

`but-workspace` ref-info snapshot for an unrelated branch below an empty workspace: it now returns preserved target metadata with an empty commit list instead of no target metadata plus an integrated commit.

---

fixes GB-1626

### DIFF
--- a/crates/but-graph/src/projection/workspace/init.rs
+++ b/crates/but-graph/src/projection/workspace/init.rs
@@ -200,16 +200,22 @@ impl Graph {
             }
         };
 
-        let mut target_ref = self
+        let configured_target_ref = self
             .project_meta
             .target_ref
             .as_ref()
-            .and_then(|target_ref| TargetRef::from_ref_name_without_commits_ahead(target_ref, self))
-            .or_else(|| self.integrated_tip_target_ref());
-        let mut target_commit = self
+            .and_then(|target_ref| {
+                TargetRef::from_ref_name_without_commits_ahead(target_ref, self)
+            });
+        let configured_target_commit = self
             .project_meta
             .target_commit_id
-            .and_then(|target_commit_id| TargetCommit::from_commit(target_commit_id, self))
+            .and_then(|target_commit_id| TargetCommit::from_commit(target_commit_id, self));
+        let mut target_ref = configured_target_ref
+            .clone()
+            .or_else(|| self.integrated_tip_target_ref());
+        let mut target_commit = configured_target_commit
+            .clone()
             .or_else(|| self.integrated_tip_target_commit(target_ref.as_ref()));
         let integrated_tip_segments =
             self.integrated_tip_segments_excluding_target_ref_tip(target_ref.as_ref());
@@ -290,8 +296,8 @@ impl Graph {
             // so we are outside the workspace limits which is above us. Turn the data back into entrypoint-only.
             ws_tip_segment_id = ep_sidx;
             kind = WorkspaceKind::AdHoc;
-            target_ref = None;
-            target_commit = None;
+            target_ref = configured_target_ref;
+            target_commit = configured_target_commit;
             metadata = None;
             lower_bound = None;
             lower_bound_segment_id = None;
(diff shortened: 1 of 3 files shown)


crates/but-graph/src/projection/workspace/init.rs
```diff
@@ -290,8 +296,8 @@ impl Graph {
             // so we are outside the workspace limits which is above us. Turn the data back into entrypoint-only.
             ws_tip_segment_id = ep_sidx;
             kind = WorkspaceKind::AdHoc;
-            target_ref = None;
-            target_commit = None;
+            target_ref = configured_target_ref;
+            target_commit = configured_target_commit;
```
estib-vega, line 300: Keep the target ref and commit info even in SBM

crates/but-graph/src/projection/workspace/init.rs
```diff
@@ -1117,12 +1123,15 @@ impl WorkspaceState {
...
         for stack in &mut self.stacks {
             // Upstream advanced: floor the stack at its fork point but keep a fully-integrated
-            // tip so it survives for `integrate_upstream`. Up to date: prune everything at or
-            // below the target.
-            prune_integrated_stack_segments(stack, &prune_segments, upstream_advanced_past_target);
-            remove_empty_branches(stack, metadata);
+            // tip in managed workspaces so it survives for `integrate_upstream`. Single-branch
+            // mode keeps the branch shell, but prunes integrated target/base commits from it.
+            prune_integrated_stack_segments(stack, &prune_segments, keep_if_fully_integrated);
+            remove_empty_branches(stack, metadata, keep_empty_segment_id);
```
estib-vega, line 1134: Prune all integrated commits when in SBM

crates/but-graph/tests/graph/init/with_workspace.rs
```diff
@@ -1257,13 +1257,13 @@ fn just_init_with_branches() -> anyhow::Result<()> {
         └── →:0: (main[🌳] →:2:)
     ");
 
-    // There is no workspace as `main` is the base of the workspace, so it's shown directly,
-    // outside the workspace.
+    // There is no workspace as `main` is the base of the workspace, so it's shown directly
+    // as a downgraded single-branch view. The target context is preserved, and the fully
+    // integrated base commit is pruned while keeping the branch container.
```
Byron, line 1262: I think this is the way - if test expectations change, we must update (or add) comments or second arguments to insta that explain why this is the desired state. That way we are forced to truly vet and understand if the change is indeed desired, in all cases.

With that said, maybe you can also take a look at the changes https://github.com/gitbutlerapp/gitbutler/pull/14186 which seems like something went wrong and it's now unclear if there still is tech debt or not.

Caleb-T-Owens (approved):

PR #14238, opened 2026-06-15
feat(but2): add support for `commit2 --above/--below <commit or branch>`

slarse:
## 🧢 Changes
Adds support for `--above/--below` position for `commit2`.

Also attempts to fix reference insertion in the `RecordingMetadata` implementation. That commit is clanked and I don't have high confidence in it - but the entire implementation is that way so right now we just want to move forward with the code that matters. It may not be entirely correct but at least we know about it and this unblocks us.

### DIFF
--- a/crates/but-transaction/src/lib.rs
+++ b/crates/but-transaction/src/lib.rs
@@ -318,6 +318,32 @@ where
             .expect("rebase is always Some(_)")
             .overlayed_graph()?;
         let workspace = graph.into_workspace()?;
+        let anchor_segment_oldest_commit_id = match &anchor {
+            Some(but_workspace::branch::create_reference::Anchor::AtSegment {
+                ref_name, ..
+            }) => {
+                let (_, segment) =
+                    workspace.try_find_segment_and_stack_by_refname(ref_name.as_ref())?;
+                Some(
+                    segment
+                        .commits
+                        .last()
+                        .map(|commit| commit.id)
+                        .or_else(|| {
+                            workspace
+                                .tip_commit_by_segment_id(segment.id)
+                                .map(|commit| commit.id)
+                        })
+                        .ok_or_else(|| {
+                            anyhow::anyhow!(
+                                "Cannot position reference below unborn segment '{}'",
+                                ref_name.shorten()
+                            )
+                        })?,
+                )
+            }
+            _ => None,
+        };
         let mut meta = RecordingMetadata {
             workspace_name: workspace.ref_name().map(ToOwned::to_owned),
             workspace: workspace.metadata.clone(),
@@ -326,7 +352,7 @@ where
 
         but_workspace::branch::create_reference(
             ref_name,
-            anchor,
+            anchor.clone(),
             self.repo(),
             &workspace,
             &mut meta,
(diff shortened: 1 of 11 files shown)


crates/but-transaction/src/lib.rs
slarse, line 1: I have _low_ confidence in these generated changes. But from my testing - they work.

crates/but/src/args/atoms/branch_arg.rs
```diff
@@ -149,12 +149,10 @@ impl BranchArg {
     }
 
     /// Resolve the argument to a branch that exists in the repository.
-    pub fn resolve_branch(&self, ctx: &but_ctx::Context) -> CliResult<ResolvedBranchRef> {
-        let repo = ctx.repo.get()?;
-
+    pub fn resolve_branch(&self, repo: &gix::Repository) -> CliResult<ResolvedBranchRef> {
```
slarse, line 152: Keep replacing `ctx` with `repo`, it makes the functions more flexible.

crates/but/tests/but/command/commit2.rs
```diff
@@ -514,5 +514,461 @@ Hint: run `but diff` to see uncommitted changes and `but stage <file>` to stage
...
+┊
+┴ 0dc3733 (common base) 2000-01-02 add M
+
+Hint: run `but help` for all commands
+
+"#]]);
+}
+
+#[test]
+fn commit_below_branch_with_multiple_commits_treats_branch_as_bucket() {
```
slarse, line 804: This is probably the most interesting test case as it's the diciest situation.

crates/but/src/command/legacy/commit2.rs
```diff
@@ -214,65 +244,116 @@ fn run(
...
+    /// Targets above the [`CliIdArg`], which must denote either a commit or a branch.
+    Above(CliIdArg),
+    /// Targets below the [`CliIdArg`], which must denote either a commit or a branch. For commits,
+    /// this is directly below. For branches, this is below the segment.
+    Below(CliIdArg),
+    /// The default target, makes a sensible choice about where to put the commit, creating a branch
+    /// if necessary. This should be used if there is no input from the user about where to put the
+    /// commit.
+    Default,
+}
```
slarse, line 262: This is my attempt to consolidate what we can target (ish). The `Branch(BranchArg)` is only a target if the branch exists (which we don't know at the point of parsing), and `UnstackedCannedBranch` is a target only in the sense that it targets the newest base. `Default` is arguably a "smart target".

crates/but/src/command/legacy/commit2.rs
```diff
@@ -319,39 +409,103 @@ impl CommitToExistingBranchOperation {
...
+    },
+    Reference {
+        name: FullName,
+        position: CommitRelativeToTargetPosition,
+    },
+    Branch {
+        name: FullName,
+        position: CommitRelativeToTargetPosition,
+    },
+}
```
davidpdrsn, line 478: I think some documentation on these would be good. I was initially a little confused about the difference between `Reference` and `Branch`.

It seems to me that `Branch` means "branch as a bucket" and is used with `commit2 --above existing-branch`, while `Segment` is only used with `commit2 -b existing-branch` or when there is exactly one stack.

Also, does it make sense to have `position` for `Reference`? It seems to always be `Below` in the code, and maybe `Above` doesn't even make sense given that we decided not to encourage anonymous segments for now?

davidpdrsn (approved):

crates/but/src/command/legacy/commit2.rs
slarse: > I think some documentation on these would be good. I was initially a little confused about the difference between Reference and Branch.

Good point, I'll add some docs. It's precisely like you say: `Commit` and `Reference` both effectively resolve to a commit, while branch resolves to a "branch as a bucket".

> Also, does it make sense to have position for Reference?

I think it makes fine enough sense, but you're right that there's no code path that currently can actually resolve to that (only `-b` resolves to `Reference` and that is always below). Good enough reason to get rid of it.

crates/but-transaction/src/lib.rs
```diff
@@ -356,11 +382,71 @@ where
...
+                    position: but_workspace::branch::create_reference::Position::Below,
+                    ..
+                }) => {
+                    let anchor_oldest_commit = anchor_segment_oldest_commit_id
+                        .expect("AtSegment anchor always has oldest commit resolved");
+                    editor.insert(
+                        editor.select_commit(anchor_oldest_commit)?,
+                        reference,
+                        InsertSide::Below,
+                    )?;
```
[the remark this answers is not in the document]
slarse: That's the intention, so nice of you to clarify.

crates/but/src/command/legacy/commit2.rs
```diff
@@ -214,65 +244,115 @@ fn run(
...
+                if stacks.next().is_some() {
+                    return Err(anyhow::anyhow!("Found more than one stack, badness!").into());
+                }
+
+                let ref_info = stack
+                    .segments
+                    .first()
+                    .and_then(|segment| segment.ref_info.as_ref())
+                    .context("Head stack as no ref")?;
+
```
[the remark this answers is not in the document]
slarse: whatever

PR #14247, opened 2026-06-16
docs(but): document `commit2`

slarse:
## 🧢 Changes
Documents the `commit2` command.

In doing this, I found that `clap` does actually format Markdown into something reasonable, supporting most of the common Markdown syntax (not sure of the exact subset, but I haven't encountered anything unsupported yet). The reason we have so much raw Markdown in our docs is that we put `#[clap(verbatim_doc_comment)]` on all the commands. I don't really get why we do that, it makes the docs less readable IMO.

As these docs are added on the `Platform` struct, they are formatted properly even though the `commit2` subcommand also has the verbatim thing.

The output looks something like this (the text here is outdated but the formatting is like this):
commit2_help

### DIFF
--- a/crates/but/src/args/commit2.rs
+++ b/crates/but/src/args/commit2.rs
@@ -1,21 +1,80 @@
+#![deny(missing_docs)]
+//! Arguments for `commit2`.
 use crate::args::atoms::{BranchArg, CliIdArg};
 
+/// Create a commit.
+///
+/// By default, all uncommitted changes are included in the commit. This can be controlled with
+/// change flags such as `--empty` and `--interactive`, or by providing `CHANGES` as positional
+/// arguments.
+///
+/// If there are no branches applied, a new branch is created for the commit. If there is only one
+/// branch applied, the commit is placed at the tip of that branch. Otherwise, the targeting flags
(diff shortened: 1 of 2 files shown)


crates/but/src/args/commit2.rs
```diff
@@ -1,17 +1,56 @@
...
+/// By default, all uncommitted changes are included in the commit. Content controlling flags such
+/// as `--empty` can be used to alter this behavior.
+///
+/// If there is only one branch applied, the commit is placed at the tip of that branch. Otherwise,
+/// the targeting flags `--above`, `--below` and `--branch` can be used to control where the commit
+/// is placed. Only one of the targeting flags can be provided at a time.
+///
+/// The commit is expected to have a commit message unless `--no-message` is provided. If neither of
+/// `--no-message` or `--message` is provided, the user's preferred editor is opened to input a
+/// message.
```
slarse, line 18: I think it makes much more sense to place this overall description on the `Platform` struct rather than on the `Subcommands` variant. It's pretty strange to refer to arguments in a doc comment when you can't even see said arguments!

davidpdrsn (approved):
Great!

PR #14266, opened 2026-06-17
fix: integration steps picked the wrong order

estib-vega:
Ensuere that we always pick the next available edge when building<br>the editor graph for integration or remote changes into the remote

fixes [GB-1586](https://linear.app/gitbutler/issue/GB-1586/error-when-resolving-branch-divergence-an-edge-with-desired-order-o)

### DIFF
--- a/crates/but-workspace/src/graph_manipulation.rs
+++ b/crates/but-workspace/src/graph_manipulation.rs
@@ -297,11 +297,16 @@ pub(crate) fn selected_edges_from_set<M: RefMetadata>(
 /// selectors.
 ///
 /// `children` are the previously captured child edges that should point back to
-/// `delimiter.child` with their original order.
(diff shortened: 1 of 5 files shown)


crates/but-workspace/src/graph_manipulation.rs
```diff
@@ -324,12 +340,41 @@ pub(crate) fn connect_segment_to_edges<M: RefMetadata>(
...
+    desired_order: usize,
+) -> usize {
+    let used_orders = existing_orders.collect::<HashSet<_>>();
+    let mut order = desired_order;
+    while used_orders.contains(&order) {
+        order += 1;
+    }
+    order
+}
+
```
estib-vega, line 377: This is used to guard against colliding parent ordering

krlvi (approved):

PR #14272, opened 2026-06-17
refactor(cli): consolidate picker implementations

davidpdrsn:
Unifies all CLI pickers under `InputOutputChannel::{prompt_select, prompt_multi_select}`.

I also removed the vendored cli-prompts crate and implemented our own picker with ratatui. I don't think its worth pulling in a whole new dependency just for a picker. I did using [dialoguer](https://docs.rs/dialoguer/latest/dialoguer/) first but ran into a bunch of issues with that.


slarse (approved):
Approve for `push.yml` change, rest is too big to review :D

krlvi (approved):

PR #14273, opened 2026-06-17
Lite: implement new designs for "add new branch" button

OliverJAsh:
I kept the copy as "add new branch" rather than "new branch" because I think it's important we emphasise this will _apply_ a new branch rather than _switch_ to a new branch. On that note, we probably need to continue the discussion on switch vs apply.

As mentioned previously, I feel conflicted about this design because it blurs keyboard selection, i.e. it looks like something that can be selected and it widens the gap between things that can be selected ("Changes" and stacks). But maybe it's okay.

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/OutlineTree.module.css
+++ b/apps/lite/ui/src/routes/project/$id/workspace/OutlineTree.module.css
@@ -14,6 +14,10 @@
 	text-align: center;
 
 	&:not([data-disabled]):hover {
+		background-color: light-dark(
+			color-mix(in srgb, var(--fill-gray-bg) 5%, var(--bg-2)),
+			color-mix(in srgb, var(--fill-gray-bg) 12%, var(--bg-2))
+		);
 		color: color-mix(in srgb, var(--text-2) 50%, var(--text-1));
 	}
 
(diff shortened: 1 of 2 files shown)


apps/lite/ui/src/routes/project/$id/workspace/OutlineTree.module.css
```diff
@@ -4,6 +4,28 @@
...
+	display: grid;
+	padding-inline: 12px;
+	padding-block: 10px;
+	border-radius: var(--radius-l);
+	background-color: var(--bg-2);
+	color: var(--text-2);
+	text-align: center;
+
+	&:hover {
+		background-color: var(--bg-3);
```
OliverJAsh: @PavelLaptev I wasn't sure what to use here. WDYT?

PavelLaptev: Looks good. There are a few things I want to change, I'll push a few commits.

PavelLaptev:
@OliverJAsh I'm also not entirely sure what the final hover state should look like, but I've made a few tweaks to improve the current implementation.
On hover:
- Text: Now slightly darker.
- Swapped out `bg-3` for dark mode. Because it's a surface token, it makes the background darker instead of lighter in dark mode. I adjusted it so it properly darkens in light mode and lightens in dark mode.

Feel free to tweak or refactor the code if you prefer a different approach, I was focused on getting the visual results.

image


https://github.com/user-attachments/assets/4826c2b9-a180-4a4f-8d68-e23c303dcd78

PavelLaptev (commented):
👍 LGTM

PR #14275, opened 2026-06-17
Initial implementation of a graph based workspace projection

Caleb-T-Owens:
This PR:

- Adds a function on the Editor to get a graph-based frame of reference for what the workspace looks like
- Adds a but-workspace level function backed by sapling-renderdag to provide line-drawing instructions
- Adds a new API that can return the new line-drawing information to the backend.

Future PRs will:

- Expose this information in `WorkspaceState`
- Include PR information as part of the reference rows
- Use PR information as part of the reference rows
- Replace duplicate logic between the workspace created here and the new upstream_integration.rs logic.
- Make calculating integration statuses optional

## Why on the Editor?

I've gone back and forth on whether the `Editor` is a suitable place to pull this information out of.

Having talked over the challenges involved with in dealing with segments, Estib and I have concluded that it is preferable to use the Editor as the basis for constructing this new graphy view of the workspace.

Some reasons behind this decision:

It removes the need to rationalise around segments in both the creation AND the consumption of the graph.
The majority of consumers of the workspace want to use it as a frame of reference for the mutations they're about to perform, so providing a representation based on Editor selectors is a very natural to consume.
Concerns that were discussed:

It's weird to depend on an Editor in read-only cases. This is very valid, but I think it's something that we can solve elegantly by improving the Editor such that write permissions are only needed when we actually going to call materialize().
If we've got a bunch of Selectors, the conversion to workspace types will end up needing to look up all the commits. It's going to however need to do that anyway.

## sapling-renderdag and pre-digested graphs

Another topic that has been discussed in depth has been whether or not we should be providing full graphs to the frontend, or some "pre-digested" version.

After a successful experiment that made use of the `sapling-renderdag` library (which both JJ and sapling) use to provide line drawing instructions to the TS frontend, we have become reasonably sure that this is a direction that we want to peruse.

Using a pre-digested graph will avoid duplication in graph rendering between our different clients. It also generally eliminates the need to do any graph traversals or logic in the frontends, keeping them primarily presentational.

## Size of this PR

This PR is roughly about +1600 -200 excluding tests and moved code. This is larger than I was initially hoping for, but my hope is that the code is still generally understandable. If not - let's change it.

The key logic sits in https://github.com/gitbutlerapp/gitbutler/blob/ce098b2b2361c2e688e188a822f6aa7cddc4f928/crates/but-rebase/src/graph_rebase/workspace.rs.

The workspace creation happens in two phases:

- Categorising the topology. Based on some rev-sets that try to describe the workspace in general terms, we try to identify the different parts of the workspace.
- Calculating upstream information. Based on both the target ref, we determine whether commits and references are integrated, and additionally whether commits are in the "local and remote" state.

Most of this logic is the same as the existing `but-workspace/src/upstream-integration.rs` stack discovery logic, except that it additionally handles "above workspace" commits and the cases when the `target_ref` is not provided.


Caleb-T-Owens:
Having talked over the challenges involved with in dealing with segments, Estib and I have concluded that it is preferable to use use the `Editor` as the basis for constructing this new graphy view of the workspace.

Some reasons behind this decision:
- It removes the need to rationalise around segments in both the creation AND the consumption of the graph.
- The majority of consumers of the workspace want to use it as a frame of reference for the mutations they're about to perform, so providing a representation based on Editor selectors is a very natural to consume.

Concerns that were discussed:
- It's weird to depend on an `Editor` in read-only cases. This is very valid, but I think it's something that we can solve elegantly by improving the `Editor` such that write permissions are only needed when we actually going to call `materialize()`.
- If we've got a bunch of `Selectors`, the conversion to workspace types will end up needing to look up all the commits. It's going to however need to do that anyway.

estib-vega (approved):

estib-vega:
let it rip

PR #14277, opened 2026-06-17
but-core: Don't delete the workspace root

estib-vega:
Fix safe checkout cleanup so deleting workspace-only files stops at the worktree root instead of attempting to remove the repository directory.

### DIFF
--- a/crates/but-core/src/worktree/checkout/function.rs
+++ b/crates/but-core/src/worktree/checkout/function.rs
@@ -130,7 +130,14 @@ pub fn safe_checkout(
                         return Err(err.into());
                     }
                 } else {
+                    let workdir = repo.workdir().context("non-bare repository")?;
                     for dir_to_delete in path_to_delete.ancestors().skip(1) {
+                        let Ok(relative_to_workdir) = dir_to_delete.strip_prefix(workdir) else {
+                            break;
+                        };
+                        if relative_to_workdir.as_os_str().is_empty() {
+                            break;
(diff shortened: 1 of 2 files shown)


estib-vega:
Ran into this while adding tests for but switch. My understanding is that this is more of a corner case, but still a good thing to have.

From clanky:

> We needed it because `path_to_delete.ancestors().skip(1)` walks from the deleted file’s parent upward, and that sequence includes the worktree root itself.
>
> Before the guard, deleting a workspace-only file could do this:
>
> 1. Remove `workdir/some-file`.
> 2. Walk ancestors.
> 3. Try to remove `workdir`.
> 4. Usually it would fail with `DirectoryNotEmpty`, but if the checkout left the worktree root otherwise empty, or in layouts where `.git` is not inside that directory, it could actually remove the worktree root.
>
> That mattered for `but switch` because switching from `gitbutler/workspace` to a normal branch can delete files that exist only in the workspace ref. `safe_checkout` cleans up empty parent directories after deletions, which is good, but it must stop at the worktree boundary.
>
> So the change is a defensive checkout cleanup fix: remove empty directories inside the project, never the project directory itself.

jonathantanmy2 (approved):
Makes sense. I was also thinking that the `.git/` directory would cause DirectoryNotEmpty, but I guess the `.git/` directory could be anywhere (even in regular git, we can set `$GIT_DIR` to somewhere else).

PR #14278, opened 2026-06-17
but: switch command

estib-vega:
## Usage
Switch to a local branch, workspace branch ID, or the GitButler workspace.

### Examples
Switch to a branch:
`but switch my-feature`

Switch back to the GitButler workspace:
`but switch --workspace`

Create a new branch at the target and switch to it (canned name if no name is specified):
`but switch --new my-feature`
`but switch --new`

### DIFF
--- a/crates/but/src/command/switch.rs
+++ b/crates/but/src/command/switch.rs
@@ -1,0 +1,55 @@
+use crate::{CliResult, IdMap, args::atoms::CliIdArg, utils::OutputChannel};
+
+pub fn handle(
+    ctx: &mut but_ctx::Context,
+    out: &mut OutputChannel,
+    target: Option<CliIdArg>,
+    workspace: bool,
+    new: bool,
+) -> CliResult<()> {
+    let mut guard = ctx.exclusive_worktree_access();
+
+    if workspace {
+        but_api::branch::workspace_checkout_with_perm(ctx, guard.write_permission())?;
+        if let Some(out) = out.for_human() {
+            writeln!(out, "Switched to workspace")?;
+        }
+        return Ok(());
+    }
+
+    if new {
+        let requested_name = target.map(|target| target.0);
+        but_api::branch::branch_checkout_new_with_perm(
+            ctx,
+            requested_name,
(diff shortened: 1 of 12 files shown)


estib-vega:
~~Waiting on this other PR https://github.com/gitbutlerapp/gitbutler/pull/14277 to be merged to fix the integration test issues.~~

Good to go

crates/but/src/command/switch.rs
```diff
@@ -0,0 +1,135 @@
...
+                CliId::Uncommitted(..) => "an uncommitted file",
+                CliId::PathPrefix { .. } => "a path",
+                CliId::CommittedFile { .. } => "a committed file",
+                CliId::Unassigned { .. } => "unassigned changes",
+                CliId::Stack { .. } => "a stack",
+            };
+            Err(bad_input(format!("Invalid branch. '{}' is {kind}", target.0)).into())
+        }
+    }
+}
```
slarse: We try to contain this kind of resolution to CLI atoms, this duplicates lot of that code. I think `BranchArg::resolve_branch()` should do the trick, see if you can use that instead.

I don't know if we have any existing code path from `CliIdArg` all the way to an unapplied branch, but you should be able to cobble it together from CLI atoms, perhaps with some minimal extension.

estib-vega:
Notes for reviewer: **This command should be hidden** for now, even if we add it. Since switching away from the workspace and then running other actions will trigger an error until something like: https://github.com/gitbutlerapp/gitbutler/pull/14276 lands

slarse (approved):
Looks fine to add as a hidden command, but argument parsing should be consolidated with CLI atoms.

Note that hidden commands are added to auto completion, and users are likely to find `switch` very quickly. If we want it to be less discoverable, consider prepending an unusual character (e.g. `_`) to the command name until we're ready to release it to the masses.

estib-vega:
> Looks fine to add as a hidden command, but argument parsing should be consolidated with CLI atoms.

Thanks! I'll address the changes and hide the command.

estib-vega:
Addressed the feedback

PR #14304, opened 2026-06-18
fix(cli,tui): improve errors when stdin isn't a terminal

davidpdrsn:


Byron:
@davidpdrsn Is commit2 really legacy from the start? I didn't look at the code, but would have hoped that commit2 would naturally start out as 'modern' implementation.

davidpdrsn:
@Byron it uses

```rust
but_api::legacy::workspace::head_info(ctx)
```

among other things 😢

Byron:
The `among other things 😢` might carry a lot of weight here. But if not, then `head_info` can be replaced with `ctx.workspace_…` for the same stack information. 

The reason I mention this is that most code, particularly not commit, probably isn't interested in the expensive computations it does (to do more advanced integration checks for instance).

https://github.com/gitbutlerapp/gitbutler/blob/18dc5132dab8cad159ad3e427f146c4e90946ed9/crates/but-api/src/legacy/workspace.rs#L47

So ideally, code only uses the cheapest-possible version.

Why is it like this? Why can't the WP just have a way of giving you expensive derived data and why does `head_info()` even exist? It's historical, originally it was made specifically for the Svelte UI to consume. Now it spreads around the codebase, and if the is any hindsight then it's probably that this "made for frontend" part should have been expressed in code better. Right now, it's only indirectly expressed by types that support serialisation. But that distinction also isn't 100% correct as there are some plumbing types that also support it, for convenience.

Anyway, I hope this detour helps to better understand why it's all a bit suboptimal, and I hope to one day do a pass to use the WP where `head_info` is unnecessarily used today.

davidpdrsn:
> The among other things 😢 might carry a lot of weight here.

It includes the TUI which `but commit2 --interactive` uses and the TUI is legacy because it uses `but status` which is legacy. Its turtles all the way down.

> Why is it like this? Why can't the WP just have a way of giving you expensive derived data and why does head_info() even exist? It's historical, originally it was made specifically for the Svelte UI to consume. Now it spreads around the codebase, and if the is any hindsight then it's probably that this "made for frontend" part should have been expressed in code better. Right now, it's only indirectly expressed by types that support serialisation. But that distinction also isn't 100% correct as there are some plumbing types that also support it, for convenience.

Exactly which versions of which APIs one should use remains somewhat of a mystery to me honestly 😆 We went with `head_info` because its usually what people talk about in meetings. I had the impression it was the go to "what does the workspace look like API". It contains all the data we need so it was very convenient.

Maybe there is some messaging about what exactly `but-api` is for thats missing. My impression was that if there is function in `but-api` that does the thing you need, then thats the function you're suppose to use.

cc @slarse

Byron:
>  It contains all the data we need so it was very convenient.

This is correct, as long as it doesn't contain more data than needed and can be substituted with `ctx.workspace_*`. After all, this is the lowest-level world view that is fastest to produce and cached. The `head_info` is uncached, and when called definitely has performance problems due to the expensive, uncached computation it does (it's even capped at 1s compute).

It's most definitely something we should try to nail down in Berlin when we meet, my take is to analyse what data is actually needed by the callers of `head_info()` and then bring it to `ctx.workspace_*` if inexpensive, while finding solutions for expensive, long-running computations (right now, integration check, but the same problem exists for hunk-dependencies which found their way into most 'status' calls).

slarse:
@davidpdrsn It would be worthwhile to take a stab at making `commit2` non-legacy in all regards but the `--interactive` flag. It seems like it should be possible.

PR #14306, opened 2026-06-18
fix(but-action): use configured model instead of hardcoded gpt-5-mini

### LINKED ISSUE #13400: Claude Code auto-commit hook: hardcoded 'gpt-5-mini' + ignores gitbutler.aimodelprovider
## Summary

The `but claude stop` hook silently fails to rewrite auto-commit messages for users who have configured Anthropic as their AI provider (`gitbutler.aimodelprovider=anthropic`). Two independent bugs in the auto-commit / reword path:

### Bug 1 — hook ignores `gitbutler.aimodelprovider`

**File:** `crates/but-claude/src/hooks/mod.rs` (~line 173)

```rust
if let Some(llm) = LLMProvider::default_openai() {
```

The hook hardcodes OpenAI as provider, ignoring the user's configured `gitbutler.aimodelprovider`. Other places in the codebase use the correct pattern `LLMProvider::from_git_config(&cfg)` (e.g. `crates/gitbutler-tauri/src/action.rs:71`, `crates/gitbutler-tauri/src/bot.rs:26`, `crates/but/src/command/legacy/branch/show.rs:417`). Only this hook is inconsistent.

### Bug 2 — `generate.rs` hardcodes `"gpt-5-mini"` as the model name

**File:** `crates/but-action/src/generate.rs` lines 34 and 90

```rust
let response = llm
    .structured_output::<StructuredOutput>(&system_message, chat_messages, "gpt-5-mini")?
```

The literal `"gpt-5-mini"` is passed to `llm.structured_output(...)` regardless of which provider `llm` is. When `llm` is an Anthropic provider, the Anthropic API responds:

```
HTTP 404 Not Found
{"type":"error","error":{"type":"not_found_error","message":"model: gpt-5-mini"}}
```

This Err is swallowed silently at `crates/but-claude/src/hooks/mod.rs` by `but_action::reword::commit(...).ok().unwrap_or_default()`, so the commit keeps the fallback title built at `crates/but-action/src/simple.rs:151-155`:

```rust
let commit_message = if let Some(prompt) = external_prompt {
    format!("{prompt}\n\n{change_summary}")
} else {
    change_summary.to_string()
};
```

→ The user's raw prompt (potentially in their native language) becomes the commit title.

## How to reproduce

1. Set provider to Anthropic:
   ```
   git config --global gitbutler.aimodelprovider anthropic
   git config --global gitbutler.aianthropickeyoption bringYourOwn
   ```
   and configure your own Anthropic API key in GitButler settings.
2. Enable auto-commit for Claude Code (`claude.autoCommitAfterCompletion = true`).
3. Use Claude Code in a non-English prompt. The resulting commit will have your raw prompt as its title instead of a generated English message.

Observed with: `but 0.19.8`, Anthropic API key `sk-ant-api03-…`, real-world test with `model=claude-haiku-4-5` succeeds with HTTP 200.

## Minimal fix

Both bugs in one patch:

```diff
--- a/crates/but-claude/src/hooks/mod.rs
+++ b/crates/but-claude/src/hooks/mod.rs
@@
-    if let Some(llm) = LLMProvider::default_openai() {
+    let llm_opt = gix::config::File::from_globals()
+        .ok()
+        .and_then(|cfg| LLMProvider::from_git_config(&cfg))
+        .or_else(LLMProvider::default_anthropic)
+        .or_else(LLMProvider::default_openai);
+    if let Some(llm) = llm_opt {

--- a/crates/but-action/src/generate.rs
+++ b/crates/but-action/src/generate.rs
@@
-    let response = llm
-        .structured_output::<StructuredOutput>(&system_message, chat_messages, "gpt-5-mini")?
+    let model = llm.model().unwrap_or_else(|| "gpt-5-mini".to_string());
+    let response = llm
+        .structured_output::<StructuredOutput>(&system_message, chat_messages, &model)?
```

(Same for `branch_name()` at line 90.)

For the fallback to work out-of-the-box for Anthropic users, `crates/but-llm/src/anthropic.rs::from_git_config` should also default `model` to e.g. `"claude-haiku-4-5"` when `gitbutler.aianthropicmodelname` is not set, or the `structured_output` callers should pick a sensible default per provider kind.

Happy to open a PR if this sounds right.

robertmonka:
Fixes #13400.

### Problem

`but-action/src/generate.rs` (`commit_message` / `branch_name`) passed a hardcoded `"gpt-5-mini"` to `structured_output` regardless of the configured provider. For Anthropic / Ollama / other providers `gpt-5-mini` is not a valid model, so the request returns `404` and the raw prompt ends up as the commit title.

### Fix

Add `LLMProvider::model_or_default()` which returns the configured model, or a per-provider default when none is set:

- **Anthropic** → `claude-haiku-4-5` (so the proxied path, which carries no model name, still hits a real model)
- **everything else** → `gpt-5-mini` (unchanged)

`generate.rs` now uses `model_or_default()` for both commit message and branch name generation instead of the hardcoded string.

This is scoped to the `generate.rs` path per the discussion in the issue. The other `default_openai()` / hardcoded-model sites (`legacy/ai.rs`, `branch/show.rs`) can follow separately.

### Before / after

To show the effect, model selection was run against a live OpenAI-compatible provider where `gpt-5-mini` is just as invalid as on Anthropic:

- **Before** — hardcoded `gpt-5-mini`: `404, not found`.
- **After** — the resolved model: a real commit message (e.g. `Add hello world print to main()`).

### DIFF
--- a/crates/but-action/src/generate.rs
+++ b/crates/but-action/src/generate.rs
@@ -30,8 +30,9 @@ unified diff:
     );
 
     let chat_messages = vec![ChatMessage::User(user_message)];
+    let model = llm.model_or_default();
     let response = llm
-        .structured_output::<StructuredOutput>(&system_message, chat_messages, "gpt-5-mini")?
+        .structured_output::<StructuredOutput>(&system_message, chat_messages, &model)?
         .context("Failed to generate structured content for commit message")?;
 
     Ok(response.commit_message)
@@ -83,12 +84,9 @@ pub fn branch_name(
     );
 
     let chat_messages = vec![ChatMessage::User(user_message)];
+    let model = llm.model_or_default();
     let response = llm
-        .structured_output::<GenerateBranchNameOutput>(
-            &system_message,
-            chat_messages,
-            "gpt-5-mini",
-        )?
+        .structured_output::<GenerateBranchNameOutput>(&system_message, chat_messages, &model)?
         .context("Failed to generate structured content for commit message")?;
 
     Ok(response.branch_name)
--- a/crates/but-llm/src/lib.rs
+++ b/crates/but-llm/src/lib.rs
@@ -238,6 +238,24 @@ impl LLMProvider {
         }
     }
 
+    /// Returns the configured model, or a per-provider default when none is set.
+    ///
+    /// Callers that need a concrete model name (e.g. structured output) must not
+    /// hardcode a provider-specific model, as it would 404 against other providers.
+    /// This returns the configured model when available, otherwise a sensible
+    /// default for the active provider: `claude-haiku-4-5` for Anthropic (so the
+    /// proxied path, which carries no model name, still hits a real model) and
+    /// `gpt-5-mini` for everything else.
+    pub fn model_or_default(&self) -> String {
+        if let Some(model) = self.model() {
+            return model;
+        }
+        match &self.client {
+            LLMClientType::Anthropic(_) => "claude-haiku-4-5".to_string(),
+            _ => "gpt-5-mini".to_string(),
+        }
+    }
+
     /// Creates a default OpenAI LLM provider using environment-based credentials.
     ///
     /// This is a convenience method that attempts to create an OpenAI provider


Byron:
Thanks a lot for the fix.

It looks good to me, but let's have a second pair of eyes for the details.

estib-vega (approved):
Thanks for catching this!

Byron:
I rebased the commit which still was unsigned and enabled auto merge as it seems to be the intention to merge as is.

@slarse I think we are hitting a timing-dependent lock failure [here](https://github.com/gitbutlerapp/gitbutler/actions/runs/27852973130/job/82435266934?pr=14306#step:3:1323). It could probably be 'fixed' by increasing the busy-timeout in `but-db`, but… it already stands at [5s](https://github.com/gitbutlerapp/gitbutler/blob/18dc5132dab8cad159ad3e427f146c4e90946ed9/crates/but-db/src/migration.rs#L8-L9). Another failure mode might be that it really was unfortunate timing and a write got in within a deferred transaction, but these are rare as the explicit use of transactions is rare. Just wanted to give you a heads-up, as for this instance I just plan to rerun the job.

<details><summary>Copilot Analysis with possible fix</summary>
<p>

Root cause: this failure is a **runtime lock contention in SQLite**, not a version mismatch.

From the failing job log (run/job you shared), the command reaches the compatibility smoke-test flow and then fails with:

- `Initiated a background sync...`
- `Error: database is locked`
- `Error code 5: database is locked`

This strongly indicates two `but` processes (or two commands sharing the same state dir) touched the same DB at once during the “install release → install nightly” validation sequence.

Relevant workflow location:
- `.github/workflows/push.yaml` in job **`validate-installer`** (installs release, then nightly in same job):  
  https://github.com/gitbutlerapp/gitbutler/blob/18dc5132dab8cad159ad3e427f146c4e90946ed9/.github/workflows/push.yaml

## Focused fix

Isolate `but` state per step and disable background sync during validation commands.

### Suggested patch (workflow)

In `validate-installer`, set a job-level temp home + explicit state dir, then pass no-sync flags (or equivalent env) for version checks:

```yaml
validate-installer:
  needs:
    - build-installer
    - fetch-published-versions
  permissions:
    contents: none
  env:
    LATEST_RELEASE_VERSION: ${{ needs.fetch-published-versions.outputs.release_version }}
    LATEST_NIGHTLY_VERSION: ${{ needs.fetch-published-versions.outputs.nightly_version }}
    # isolate app state for this job
    HOME: ${{ runner.temp }}/but-home
    XDG_CONFIG_HOME: ${{ runner.temp }}/xdg-config
    XDG_DATA_HOME: ${{ runner.temp }}/xdg-data
    XDG_CACHE_HOME: ${{ runner.temp }}/xdg-cache
    # if supported by CLI/app, disable background activity
    BUT_NO_BACKGROUND_SYNC: "1"
  steps:
    - run: mkdir -p "$HOME" "$XDG_CONFIG_HOME" "$XDG_DATA_HOME" "$XDG_CACHE_HOME"

    - name: Download ${{ matrix.artifact_id }} installer
      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c
      with:
        name: "but-installer-${{ matrix.artifact_id }}-${{ github.run_number }}"
        path: .

    - run: chmod +x ./but-installer

    - name: Install release
      run: ./but-installer

    - name: Validate release version
      run: but --version --no-background-sync | grep "but $LATEST_RELEASE_VERSION"

    - name: Install nightly
      run: ./but-installer nightly

    - name: Validate nightly version
      run: but --version --no-background-sync | grep "but $LATEST_NIGHTLY_VERSION"
```

## Why this resolves it

- The failure appears when CLI startup triggers background sync (seen in logs), which opens the same DB.
- For CI validation, you only need `but --version`, so background tasks are unnecessary.
- Isolated HOME/XDG directories prevent cross-step/process DB reuse from shared default locations.

If `--no-background-sync` is not currently supported, keep the env isolation and add a short retry wrapper around version checks as a fallback:

```bash
for i in 1 2 3; do
  but --version && break
  sleep $((i*2))
done
```

But the preferred fix is **disable background sync for CI version validation + isolate state dirs**.

</p>
</details>

slarse:
@Byron The fact that the background sync causes locking in scripts is known. My preferred fix would be to _remove_ the background sync - it both causes locking and is engineered to create orphans that may lock up without any supervising process being there to kill them. And there is no simple way for a user to see that they're there, either.

It's just a very problematic piece of the CLI as a whole.

Byron:
> It's just a very problematic piece of the CLI as a whole.

Agreed!

With that said:

>  it both causes locking and is engineered to create orphans that may lock up

I didn't know it can lock up, in the sense of being backgrounded and waiting forever. The DB timeout causes a shutdown, and one would expect that the worktree/repository exclusive lock will eventually be obtained.

slarse:
> > it both causes locking and is engineered to create orphans that may lock up
> 
> I didn't know it can lock up, in the sense of being backgrounded and waiting forever. The DB timeout causes a shutdown, and one would expect that the worktree/repository exclusive lock will eventually be obtained.

Maybe I phrased that poorly, there's no known reproduction of it locking up. But there are plenty of failed CI runs where it has locked up, the version upgrade/downgrade smoke tests are especially affected.

Orphaned processes that contend for interprocess locks are by design high risk for locking up indefinitely.

> The DB timeout causes a shutdown

But there is no transaction timeout, is there? And we have a bunch of codepaths that can upgrade a read lock to a write lock, right? I wouldn't be surprised if that could deadlock.

Byron:
> But there is no transaction timeout, is there? And we have a bunch of codepaths that can upgrade a read lock to a write lock, right?

As there is just a single lock in Sqlite, I don't think there can be deadlocks. There is also only busy-timeout (5s) when waiting makes sense at all. 

But I agree that the best way to deal with this particular issue is to definitely not sync like this.

PR #14307, opened 2026-06-18
Fix data loss when committing a new directory at a renamed file's old path

krlvi:
## The problem

Committing can silently delete files when, in a single commit, a tracked file is renamed away from a path and a new directory is created at that same path.

### Reproduction

1. Commit a file `A`.
2. Rename `A` → `B`.
3. Create a new directory `A/` and put one or more files in it.
4. Commit all changes.

The directory `A/` and its contents disappear from the commit. Worse, `but undo` then breaks and cannot recover them — the files survive only in the raw snapshot, not via GitButler.

### Why it happens

When GitButler builds the committed tree, it applies the changes in path-sorted order — for this scenario that's `A/one`, `A/two`, and finally the rename `B` ← `A`. The new directory entries are written first, which turns the old path `A` into a directory in the tree being built. The rename of `A` → `B` then removes the old source path `A`. By that point `A` is no longer the original blob; it is the freshly-built directory, so removing it prunes the entire `A/` subtree and drops its files from the commit.

The bug is specific to a file→directory transition combined with rename detection:

- A plain deletion of `A` (without a rename) does not trigger it.
- It reproduces consistently because the rename's destination (`B`) sorts after the new directory's children (`A/...`), so the directory is built before the source removal runs.

### Impact

- **Silent data loss** — committed directory contents vanish with no error.
- **Broken recovery** — `but undo` fails afterward, so the loss is not recoverable through normal GitButler operations.

### DIFF
--- a/crates/but-core/src/tree/mod.rs
+++ b/crates/but-core/src/tree/mod.rs
@@ -217,7 +217,15 @@ pub fn apply_worktree_changes<'repo>(
         };
         // NOTE: See copy below!
         if let Some(previous_path) = change_request.previous_path.as_ref().map(|p| p.as_bstr()) {
-            base_tree_editor.remove(previous_path)?;
+            // A sibling change in this batch may have turned `previous_path` into a directory
+            // (e.g. `A` renamed to `B` while a new `A/` is committed at the old path). Those
+            // upserts already replaced the old blob, so only remove the source if it's still a leaf.
+            if base_tree_editor
+                .get(previous_path)
+                .is_none_or(|e| !e.mode().is_tree())
+            {
+                base_tree_editor.remove(previous_path)?;
+            }
         }
         if change_request.hunk_headers.is_empty() {
             let rela_path = change_request.path.as_bstr();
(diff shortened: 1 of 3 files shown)


crates/but-core/src/tree/mod.rs
```diff
@@ -217,7 +217,15 @@ pub fn apply_worktree_changes<'repo>(
         };
         // NOTE: See copy below!
         if let Some(previous_path) = change_request.previous_path.as_ref().map(|p| p.as_bstr()) {
-            base_tree_editor.remove(previous_path)?;
+            // A sibling change in this batch may have turned `previous_path` into a directory
```
jonathantanmy2, line 220: How did a sibling change in this batch turn `previous_path` into a directory without throwing an error? That seems like something we should fix.

I think what should happen is:
 - a sibling change can only upsert a file if what it's overriding is also a file
 - we should sort the changes in the batch such that deletions happen before additions

krlvi: Yeah... I was just investigating this report by @slarse https://discord.com/channels/1060193121130000425/1272844201914142834/1517150632899776662

jonathantanmy2: I sent #14312 which also fixes this (I copied over the test case from this PR). I prefer my fix because this fix still relies on the behavior that if we upsert into `A/one`, we automatically delete `A`, and I find that behavior error-prone.

PR #14315, opened 2026-06-18
Refine ToggleGroup item sizing and colors

PavelLaptev:
style: tweak layout and element styles

- Increase height of the item component
- Update icon and text styles to match buttons


PavelLaptev:
cc @OliverJAsh

PR #14318, opened 2026-06-19
Follow-up to 14312

Byron:
This is a follow-up to #14312 which uses `remove_leaf()` instead of `remove()` for added control.

Doing so would have prevented the bug we were seeing automatically.

Please note that I refrained from putting it into more places, despite having found some, because I'd have to think really hard if `remove_leaf()` is truly correct there 😅.

There is also [a uboot](https://github.com/GitoxideLabs/gitoxide/pull/2604) that greatly improves the performance of `packedrefs` file reading, a contribution that speeds up the AUR package repo by 4 seconds (of a 12s fetch operation). This performance optimisation will be most noticeable when there are 100k+ refs I suppose.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4339,7 +4339,7 @@ dependencies = [
 [[package]]
 name = "gix"
 version = "0.84.0"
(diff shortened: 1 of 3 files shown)


estib-vega (approved):
Trusting you. Changes seem focused enough 🤞

PR #14334, opened 2026-06-19
Lite: show updates in native message box

OliverJAsh:
This makes the app feel more native.

@slarse I might need your help testing this with the production builds, if you have time we could quickly pair.

file-311f8331f98703ad9953f1debc8141c3


slarse:
Ftr, tested on Linux and works fine as before.

On macOS the auto-updater seems to fail on installing when starting from a bundle built locally. So we'll just have to wait until two new nightlies have been released to see if this actually works.

PR #14351, opened 2026-06-21
fix: allow gpg.ssh.program to use tilde

tobeycodes:
## 🧢 Changes

Updates gpg program to use `trusted_path` rather than `trusted_program` so that gpg signing works with a path containing a tilde.

It would appear to me the underlying functions are quite similar and both do a filter on `filter_config_section` but transparently this is my first time reading gitoxide and it's definitely possible I am overlooking something

1. https://github.com/GitoxideLabs/gitoxide/blob/a507ee0e9b7023598516a1ada1d99e8b622acdde/gix/src/config/snapshot/access.rs#L63-L83
2. https://github.com/GitoxideLabs/gitoxide/blob/main/gix/src/config/cache/access.rs#L260
3. https://github.com/GitoxideLabs/gitoxide/blob/main/gix/src/config/cache/access.rs#L528

## ☕️ Reasoning

I use a script in my dotfiles repo that will act differently depending on my machine. On my Mac it will use 1password. In my virtual machines and servers it will sign without 1password since there is no GUI. These machines do not always use the same username so an absolute path is not always the same. 

**Before**

image

**After**

image

### DIFF
--- a/crates/but-core/src/commit/mod.rs
+++ b/crates/but-core/src/commit/mod.rs
@@ -1,6 +1,4 @@
-use std::{
-    borrow::Cow, collections::HashSet, io::Write, path::Path, path::PathBuf, process::Stdio,
-};
+use std::{collections::HashSet, io::Write, path::Path, path::PathBuf, process::Stdio};
 
 use anyhow::{Context as _, bail};
 use bstr::{BStr, BString, ByteSlice};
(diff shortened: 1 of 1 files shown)


Byron:
Thanks so much! The before-after pictures are particularly useful 🙏.

Indeed, also Git treats these values [as paths](https://github.com/git/git/blob/7760f83b59750c27df653c5c46d0f80e44cfe02c/gpg-interface.c#L802), and not as program, so this change is what it takes to align both.

Byron (approved):

PR #14352, opened 2026-06-21
skill: document `but pr` in SKILL.md

### LINKED ISSUE #14290: skill: `SKILL.md` doesn't mention `but pr`, agents fall back to `gh pr create`
### Version

0.20.0

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

When I ask my coding agent (Claude-style, GitButler skill installed) to "open a PR for this branch," it runs `gh pr create --draft ...` instead of `but pr new <branch> --draft ...`. Every time. The skill is loaded and `but pr new` works fine in the same environment.

`but pr new` is documented in [`references/reference.md` § `but pr`](https://github.com/gitbutlerapp/gitbutler/blob/master/crates/but/skill/references/reference.md#L376-L394) and used in [`references/examples.md`](https://github.com/gitbutlerapp/gitbutler/blob/master/crates/but/skill/references/examples.md), but `SKILL.md` itself — which agents read first and rely on most heavily — never mentions any `but pr` subcommand. The Git-to-But Map also has no `gh pr create` → `but pr new` row. So the agent's only "open a PR" command is the `gh pr create` it knows from training data. When retrospecting with the agent, it suggests that the lack of obvious "recipe" for pull requests makes it never look into the references at all.

_No response_

_No response_

```shell

```

ameyypawar:
`SKILL.md` — the skill file coding agents read first — never mentioned any `but pr` subcommand, so agents fall back to `gh pr create` instead of `but pr new`. (`but pr` is documented in `references/reference.md`, just not surfaced where agents look first, and the Git-to-But Map had no `gh pr create` row.)

This surfaces it in the three places agents rely on, mirroring the existing `reference.md` docs:

- a `but pr new` line under **Command Patterns**
- an **Open a pull request** Task Recipe
- the `gh pr create` → `but pr new` row in the **Git-to-But Map**

Doc-only change; `prettier` and `cargo check -p but` pass locally.

Fixes #14290

### DIFF
--- a/crates/but/skill/SKILL.md
+++ b/crates/but/skill/SKILL.md
@@ -53,6 +53,7 @@ but <mutation> ...
 - Tear off a branch: `but move <branch-name-or-id> zz` (`zz` = unassigned; branch name or branch CLI ID)
 - Push: `but push <branch-name>` — always specify the branch; bare `but push` pushes ALL branches when run non-interactively
 - Pull: `but pull --check` then `but pull`
+- Open a PR: `but pr new <branch-name>` — pushes the branch and opens the PR in one step (no separate `but push`); add `--draft` for a draft. In non-interactive shells pass `-m "<title>"`, `-F <file>`, or `-t` (use default content) to skip the editor. Needs `but config forge auth`.
 
 ## Task Recipes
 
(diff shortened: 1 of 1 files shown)


krlvi:
i will run some benchmarks and adjust before taking this

krlvi:
this is now implemented in https://github.com/gitbutlerapp/gitbutler/pull/14387 

thanks for flagging this

PR #14358, opened 2026-06-22
grit transport

schacon:
This adds a `gitbutler.grit` config setting that uses the Grit transport library for push/fetch operations instead of fork/exec'ing out to Git.

While slower and Claude'd, it seems to work. Using it as a fallback option if `git` is not present may not be the _wurst_ approach.

```
❯ time GRIT_NET_DEBUG=1 butd fetch                                                              took 12s
Assuming you meant to check for upstream work, running `but pull --check`
Fetching from upstream remotes...
[grit-net] ssh connect git@github.com:gitbutlerapp/gitbutler.git (service=git-upload-pack, request protocol v2)
[grit-net] ssh connected: protocol v0, 14483 ref(s) advertised
[grit-net] fetch_remote: begin — protocol v0, 1 refspec(s), tags=Following, depth=None
[grit-net] fetch_remote: remote advertised 14483 ref(s) (v0/v1 advertisement)
[grit-net] fetch_remote: 2879 matched ref(s), want 0 object(s) [+56.764917ms since begin]
[grit-net] fetch_remote: retain_following_tags 304.36475ms
[grit-net] fetch_remote: apply 113.8285ms — repo_open 74.292µs, prune 42ns, packed_load 14.390667ms, resolve 66.587611ms, classify 32.396972ms (19 ancestry-checked), write 0ns (0 written)
[grit-net] fetch_remote: done — 2865 ref update(s), default branch 'master' [+474.99775ms total]
Checking integration statuses...
Checking base branch status...

Base branch:    origin/master
Upstream:       13 new commits on origin/master

  280a3b9fc4 Merge pull request #14351 from tobeycodes/fix/allow-gpg-ssh-program-to-u
  8ffb540ccf Merge pull request #14354 from gitbutlerapp/uncommit-diff-output  Add di
  73b850ed65 Merge pull request #14348 from tobeycodes/patch-1  fix: typo in readme.m
  ... (10 more)

Run `but pull` to update your branches
GRIT_NET_DEBUG=1 butd fetch  2.67s user 0.27s system 63% cpu 4.629 total
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -538,7 +538,7 @@ dependencies = [
  "serde_json",
  "serde_path_to_error",
  "serde_urlencoded",
- "sha1",
+ "sha1 0.10.6",
  "sync_wrapper",
  "tokio",
  "tokio-tungstenite",
@@ -2262,6 +2262,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "const-oid"
+version = "0.10.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a6ef517f0926dd24a1582492c791b6a4818a4d94e789a334894aa15b0d12f55c"
+
+[[package]]
 name = "const-random"
 version = "0.1.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2882,6 +2888,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "56254986775e3233ffa9c4d7d3faaf6d36a2c09d30b20687e9f88bc8bafc16c8"
 
 [[package]]
+name = "difflib"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6184e33543162437515c2e2b48714794e37845ec9851711914eec9d308f6ebe8"
+
+[[package]]
 name = "digest"
 version = "0.10.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


Byron:
> Using it as a fallback option if git is not present may not be the wurst approach.

I agree 😋.

And what would be really, really great is if `grit` could ship with a built-in `ssh` compatible transport, i.e. one that doesn't rely on the binary. This would probably be most useful for Windows people who have never setup Git and use GitButler as their first (and only, after that :)) Git client.

`grit-ssh` is probably going to land long before `gix-ssh` can be a thing.

estib-vega:
> This would probably be most useful for Windows people who have never setup Git
>

Yes plssssss

schacon:
Hm. I _suppose_ it's possible, but then you really do get down a hole of like which key types are supported, ssh agents, etc. I'm curious if this is worth the trouble vs just encouraging HTTPS use (or even via the `grit auth` thing I've demo'd). Like if someone is on Windows and doesn't have Git for Windows installed, what are the chances they're going to figure out how to create and use SSH keys?

Byron:
> Like if someone is on Windows and doesn't have Git for Windows installed, what are the chances they're going to figure out how to create and use SSH keys?

Right! I was never asking the question, but blindly adopted the Cargo stance of "Cargo wants to be independent, so it's using `git2` built-in SSH support which comes with `libgit2`". I think that users can configure Cargo by providing their SSH pub key (like GitButler did), and then that's passed to `git2` when Git dependencies are fetched through a SSH URL. The exact use-case of a Windows user having keys, but not having any SSH binary was never explained at least in what I read.

So it's true, if `grit` wanted to support that, it would also have to provide key generation facilities, and a way to safely store and retrieve the passphrase (which could be done with `git credential` I suppose).
The value proposition here would be "easy access to SSH based serves for Windows users with a turn-key solution", without wanting to become/have a fully-integrated `ssh` program.

PR #14372, opened 2026-06-22
Bump design-core and update list item CSS vars

PavelLaptev:
In Figma, I started using `list item hover` and `selected but not in focus` opacity states in several places, so I converted them into variables for better consistency.
This PR hooks those same variables up in the code, pulling them from the `design-core` package.


PavelLaptev:
cc @OliverJAsh

PR #14398, opened 2026-06-23
Update field styles

PavelLaptev:
### Fields
- Added field styles.
- Added Component Storybook stories.

### Commit Box Updates
- Updated the commit box to match the styling of our other fields for better visual consistency across the page.

image

### Other Changes
- Added a new CSS `--focus-ring` variable.
- Applied the new focus ring to buttons as well.
- Improve "lines added/removed" component readability.


apps/lite/ui/src/global.css
```diff
@@ -13,4 +15,9 @@ body {
 
 button {
 	app-region: no-drag;
+
+	&:focus-visible {
+		outline: none;
+		box-shadow: var(--focus-ring);
```
OliverJAsh: We can probably use `outline` instead of `box-shadow`.

apps/lite/ui/src/routes/project/$id/workspace/OutlineTree.module.css
```diff
@@ -138,11 +140,16 @@
 	&:focus {
 		outline: none;
 	}
+
+	&::placeholder {
+		color: var(--text-3);
```
OliverJAsh, line 145: Perhaps this should be applied to all input/tetxareas.

OliverJAsh (approved):

apps/lite/ui/src/global.css
PavelLaptev: done

apps/lite/ui/src/routes/project/$id/workspace/OutlineTree.module.css
PavelLaptev: If we go this route, should we implement more than just the global placeholder colot and cover all the basics for inputs/textareas? For example:

```css
input,
textarea {
	outline: 1px solid var(--border-2);
	background-color: var(--bg-1);
	color: var(--text-1);
	transition: outline-color var(--transition-fast);

	&::placeholder {
		color: var(--text-3);
	}

	&:hover:not(:focus, :disabled) {
		outline-color: var(--border-1);
	}

	&:focus {
		outline: var(--focus-ring);
	}

	&:disabled {
		cursor: not-allowed;
		opacity: 0.5;
	}
}
```

OliverJAsh: I think it would be best to apply the style in our field components rather than globally, so we don't have to undo a bunch of styles if we don't want the design system's styles for whatever reason. For example: inline reword/rename.

PR #14423, opened 2026-06-24
feat(but): make hunk IDs semi-stable

slarse:
The old hunk IDs were basically just globally indexed, making them incredibly unstable. Committing or discarding the first hunk ID would shift all subsequent IDs by 1, requiring a status after each operation. And also making it very easy to accidentally reference the wrong hunk as all "shifted" hunks would have indexes that existed before the mutation, but at that point referencing other hunks.

The new hunk IDs are derived from the content of the hunk, including any context lines. This makes them rather stable, and collisions are unlikely as you need to have identical additions, deletions _and_ context lines to cause one.

Additionally, hunk IDs are now always namespaced to the containing file, further reducing the probability of a collision.

A qualified hunk reference takes the form `<file_id>:<hunk_id>`, where `<hunk_id>` is typically the first 1-2 characters of the hunk's hash.

In the event of a hash collision, the hunk ID is extended with `<collision_index>-<num_collisions>`. See the docs of `assign_content_based_hunk_short_ids` for details.


crates/but/src/id/mod.rs
```diff
@@ -1151,24 +1251,45 @@ impl UncommittedFile {
...
+                    .short_id_hunk_assignments
+                    .iter()
+                    .filter(|(short_id, _)| *short_id == *element)
+                    .map(|(short_id, hunk_assignment)| {
+                        let cli_id = CliId::UncommittedHunkOrFile(UncommittedHunkOrFile {
+                            id: short_id.to_owned(),
+                            hunk_assignments: NonEmpty::new(hunk_assignment.to_owned()),
+                            is_entire_file: false,
+                        });
+                        Box::new(Leaf { cli_id }) as Box<dyn Node<'a> + 'a>
```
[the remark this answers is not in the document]
slarse: What even is this code suggestion ...

crates/but/src/id/mod.rs
```diff
@@ -489,12 +492,15 @@ impl IdMap {
...
+                )?;
+            }
+
+            for (short_id, hunk_assignment) in &uncommitted_file.short_id_hunk_assignments {
                 uncommitted_hunks.insert(
-                    new_short_id,
+                    format!("{}:{}", uncommitted_file.short_id, short_id),
                     UncommittedHunk {
                         hunk_assignment: hunk_assignment.clone(),
                     },
```
slarse, line 506: I'm looking to get rid of this separate map of `uncommitted_hunks`, but that had a bit of a blast radius that I don't want to deal with in this PR. So I'll handle that separately.

crates/but/src/id/mod.rs
```diff
@@ -527,6 +533,92 @@ impl IdMap {
...
+    ///     - `<collision_index>` is the 0-based index of the hunk relative to the other colliding
+    ///       hunks, in input order.
+    ///     - Example: `0#0-2` and `0#1-2`
+    ///
+    /// We include the number of collisions in the collision disambiguation to prevent
+    /// commit/discard of a leading colliding hunk from causing trailing colliding hunks to get new
+    /// IDs that previously pointed to other colliding hunks.
+    ///
+    /// Note that hunks that lack a diff follow the same rules, only that `<prefix>="q"` always.
+    fn assign_content_based_hunk_short_ids<'a>(
```
slarse, line 564: This function contains the entire logic for hunk ID assignment, and is meant to be called on a file-by-file basis.

crates/but/tests/but/command/absorb.rs
```diff
@@ -93,31 +93,31 @@ fn uncommitted_hunk() -> anyhow::Result<()> {
...
     env.but("diff a.txt")
         .assert()
         .success()
         .stderr_eq(snapbox::str![])
         .stdout_eq(snapbox::str![[r#"
-────────╮
-j0 a.txt│
-────────╯
+──────────╮
+nk:2 a.txt│
```
slarse, line 103: Most test changes just look like this

crates/but/tests/but/command/rub.rs
```diff
@@ -276,7 +276,7 @@ Uncommitted changes
   ],
   "stacks": [
     {
-      "cliId": "k0",
+      "cliId": "j0",
```
slarse, line 279: The reason stack IDs change in tests is that they're sourced from the same global ID pool that hunk IDs used to be sourced from.

slarse (commented):
All production code changes are contained in the first commit, with the second commit just updating tests to adapt to the new hunk IDs.

davidpdrsn (approved):
Love it!

crates/but/src/id/mod.rs
```diff
@@ -1151,24 +1241,45 @@ impl UncommittedFile {
...
+                    Ok(vec![Box::new(Leaf { cli_id })])
+                } else {
+                    Ok(vec![])
+                }
+            }
+            _ => {
+                let matches = self
+                    .short_id_hunk_assignments
+                    .iter()
+                    .filter(|(short_id, _)| short_id == element)
```
slarse, line 1270: It would be good to have the full ID available here, because then we'd be able to do a prefix match on the full ID rather than an exact match on the short ID. With this exact match approach, you might perform a mutation that causes some other hunk's ID to get shortened due to less overlap with an adjacent ID, and that in turn makes the initial short ID not work anymore. Even though it's more specific than the new shortened version.

I'll look into exposing the full IDs as well tomorrow. For now an exact match is the best we can do.

crates/but/src/id/mod.rs
```diff
@@ -1151,24 +1241,45 @@ impl UncommittedFile {
...
-            && let Some((short_id, hunk_assignment)) = self.short_id_hunk_assignments.get(index)
-        {
-            let cli_id = CliId::UncommittedHunkOrFile(UncommittedHunkOrFile {
-                id: short_id.to_owned(),
-                hunk_assignments: NonEmpty::new(hunk_assignment.to_owned()),
-                is_entire_file: false,
-            });
-            return Ok(vec![Box::new(Leaf { cli_id })]);
+        match element.strip_prefix('#') {
+            Some(maybe_index) if let Ok(index) = usize::from_str(maybe_index) => {
```
jonathantanmy2, line 1253: Is this branch meant for IDs like `file.txt:#0` to access the first hunk in the file? If we don't need this, we can remove it in another PR. If we need it, we should add a test (also in another PR).

jonathantanmy2 (approved):
Thanks - looks good. I have one question, but that can be addressed in another PR.

crates/but/src/id/mod.rs
slarse: Yes, exactly, that's for indexing (which used to be the only thing that was possible).

I've adapted the old parsing tests for indexing (see e.g. [this one](https://github.com/gitbutlerapp/gitbutler/pull/14423/changes#diff-077981916c46d233afc2815b920727daa55b133b32d906842f0c6082f96a7784R1273)), but did not add any e2e tests as I'm not actually sure we want to keep it. It's just that @krlvi seemed to think it was a good thing to retain and it costs very little. As we have indexing for the collisions by necessity, I thought "file-global" indexing made sense for consistency as well.

So, for indexing in particular, test coverage remains the same as before.

PR #14452, opened 2026-06-25
Order stacks by the workspace commit's parents, drop parent_order

mtsgrd:
Stack display order now follows the workspace commit's **real parent array** instead of a cached `parent_order` edge weight — which had gone stale (it held metadata order), so octopus/multi-stack workspaces showed the wrong order.

The parent array is the **single source of stack order**. Ordering happens once, at graph-build time, so nothing downstream re-sorts: the graph's first-parent traversal, its debug display, and the `but st` lane order all read the same already-correct order off the edges.

### 1. Drop `parent_order`, read the real parents
- display reads the actual `parent_ids`; `Edge.parent_order`, its accessor, and the `rebuild_edges_in_parent_order` heuristic are all gone
- `preferred_parent_order_by_commit` re-keyed to segment ids — still steers the merge-walk to the entrypoint through diamonds, just no weight
- first parent from `parent_ids[0]`; `but-rebase` derives its step order the same way

### 2. Order once, at graph-build time
- `but-graph` orders the ws-commit's outgoing edges by the real parent array in `post.rs`: `stack_entry_commit` walks first-parent edges (skipping empties) to find each lane's entry, then sorts by `(by_parent, by_metadata)` — metadata order stays as a secondary tie-break for lanes the parent array can't tell apart (e.g. two branches sharing one base)
- the workspace projection's re-sort is **deleted** — it reads order straight off the edges; the invariant now holds by construction
- `remerged_workspace_commit_v2` (the non-graph writer) sorts its parent list by `stack.order` instead of `HashMap::values()`, so the real app produces parents in a deterministic order matching what the display reads

### 3. Empty branches
- give each empty branch its own real ws-commit parent, so two empties on **distinct** bases stay distinct instead of collapsing (one base ↔ one lane); a second empty on an already-used base walks back to a distinct unused base (concise out-of-commits error)
- `but-rebase` gives commit-less empties distinct, increasing `Edge.order` values from a counter, so the step graph never has tied parent orders and the old tie-break `v.reverse()` is removed
- follow the first-parent edge by graph order (not dst id) when walking the excluded side, so upstream commits aren't miscounted for empty stacks
- resolve each branch's own base after integrated-commit pruning instead of falling back to the workspace merge base

### 4. Fixtures + snapshots
- fixture helpers check out the first arg as the first parent deterministically; one-empty scenarios list the empty branch first (a no-op `git merge` of an ancestor would otherwise silently drop it)
- the `three-stacks` move scenario gives its stacks distinct bases, so emptied stacks resolve to different parents instead of collapsing onto one
- multi-stack display flips to the true order across but-graph / but-workspace / but — snapshots re-accepted; the `move`/`squash` commit-order asserts are updated for the new order

✅ but-graph / but-workspace / but-rebase / but green; `cargo fmt`, `clippy -D warnings`, and `pnpm begood` all clean.

> [!NOTE]
> The construction layer can still collapse two empty branches that genuinely share one git base onto a single ws-commit parent (the array then has fewer slots than stacks). This PR sidesteps that with distinct-base fixtures; the long-term fix is to have move/remerge materialise empties as distinct parents even on a shared base. Tracked in the branch handover.

### 5. Drive-by: `first_ordered_parent` (but-rebase)
Reference resolution in `materialize`/`commit`/`reword`/`rebase` only needs the first parent, but called `collect_ordered_parents` (a full pruned DFS over every parent) and took `.first()`. Added `first_ordered_parent`, which shares the DFS but stops at the first commit. No behaviour change; the all-parents call sites are unchanged.

### DIFF
--- a/crates/but-graph/src/api.rs
+++ b/crates/but-graph/src/api.rs
@@ -59,10 +59,6 @@ impl Graph {
     /// `dst_commit_id` can be provided if the connection is to a future commit that isn't yet available
     /// in the `segment`. If `None`, it will be looked up in the `segment` itself.
     ///
-    /// `parent_order` is the 0-based position of `dst_commit_id` among the source commit's
-    /// parents. For a merge commit with parents `[A, B]`, the edge to `A` must use `0`,
-    /// and the edge to `B` must use `1`, even if traversal discovers `B` before `A`.
-    ///
     /// Return the newly added segment.
     pub fn connect_new_segment(
         &mut self,
@@ -71,7 +67,6 @@ impl Graph {
         dst: Segment,
         dst_commit: impl Into<Option<CommitIndex>>,
         dst_commit_id: impl Into<Option<gix::ObjectId>>,
-        parent_order: u32,
     ) -> SegmentIndex {
         let dst = self.inner.add_node(dst);
         self.inner[dst].id = dst;
@@ -82,7 +77,6 @@ impl Graph {
             dst,
             dst_commit,
             dst_commit_id.into(),
-            parent_order,
         );
         dst
     }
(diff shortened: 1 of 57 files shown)


mtsgrd:
Closing in favor of a consolidated draft PR from `commit-graph-experiment` — this branch's commits are being folded into that single-branch series (link to follow).

mtsgrd:
Consolidated draft PR: #14623

PR #14464, opened 2026-06-26
Use global selection colors for diff

samhh:
Closes GB-1657. We can continue to refine, we haven’t discussed the diff styling in detail yet.


OliverJAsh:
This is much better!

Related, regarding themes: [GB-1644](https://linear.app/gitbutler/issue/GB-1644/diff-library-theme-clashes-with-dark-mode-styles)

PR #14467, opened 2026-06-26
docs(cli,squash2): write docs for squash2

davidpdrsn:


crates/but/src/args/squash2.rs
```diff
@@ -1,18 +1,89 @@
...
+    /// This cannot be used when `TARGET` is the uncommitted area (`zz`).
     #[clap(long, short = 'u', group = "commit_message")]
     pub use_target_message: bool,
+
+    /// Use the message of the source(s).
+    ///
+    /// The message of the target will be discarded.
+    ///
+    /// Cannot be used if `<SOURCES>` are not committed or if moving committed changes between
+    /// commits.
```
slarse: Should also not be usable when `TARGET` is the uncommitted area?

crates/but/src/args/squash2.rs
```diff
@@ -1,18 +1,89 @@
...
     #[clap(long, group = "commit_message")]
     pub use_source_message: bool,
+
+    /// The target to squash into.
+    ///
+    /// If `TARGET` is a commit the sources will be added to the commit.
+    ///
+    /// If `TARGET` is a branch the sources will be added to the first commit on the branch.
```
slarse, line 59: @krlvi and I discussed yesterday that it might be time to expose change IDs as CLI IDs for commits. I'm looking into that next week. That might scratch the itch of committing to the tip of a branch (as the change ID remains the same even after multiple squashes) to the point that we might not need this special case.

slarse (approved):
Looks good, just spotted one thing that might be incomplete and had a remark regarding upcoming change IDs as cli IDs.

crates/but/src/args/squash2.rs
davidpdrsn: Cool! In that case we should expose them in `but status` as well.

PR #14399, opened 2026-06-23
Replace change status icon with kebab menu on file row hover and selection

PavelLaptev:
Swaps the 'change status' icon menu for kebab menu to prevent layout shifting on hover.

image

Reference (Conductor app):

https://github.com/user-attachments/assets/39f0564c-ced4-4d68-bd70-5cf49ead0526

PR #14469, opened 2026-06-26
Styles file list item

PavelLaptev:
* Reverted [https://github.com/gitbutlerapp/gitbutler/pull/14399](<https://github.com/gitbutlerapp/gitbutler/pull/14399>)
* I fixed the width of the change status icon to keep the kebab menu in the same position when hovering over different files.

Before:

https://github.com/user-attachments/assets/d4ab10b8-1a71-4449-8368-43f71e51fe59

After:

https://github.com/user-attachments/assets/d229297e-7088-4b3a-bbb4-92e754aca845

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/FilesTree.module.css
+++ b/apps/lite/ui/src/routes/project/$id/workspace/FilesTree.module.css
@@ -16,9 +16,11 @@
 
 .fileStatusBadge {
 	align-self: center;
+	width: 12px;
 	font-weight: 700;
 	font-size: 11px;
 	font-family: var(--text-fontfamily-base);
+	text-align: right;
 
 	&[data-char="A"] {
 		color: var(--change-status-addition);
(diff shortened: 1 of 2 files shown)


PavelLaptev:
cc @OliverJAsh

apps/lite/ui/src/routes/project/$id/workspace/FilesTree.module.css
```diff
@@ -16,16 +16,11 @@
...
 	font-size: 11px;
-
 	font-family: var(--text-fontfamily-base);
-
-	.fileRow:hover &,
-	.fileRow:focus-within &,
-	.fileRow[aria-selected="true"] & {
-		display: none;
-	}
+	text-align: right;
```
OliverJAsh, line 23: This is a small detail but I wonder if it would look better horizontally aligned with the other status badges, as if it was an icon. Happy either way!

OliverJAsh (approved):

apps/lite/ui/src/routes/project/$id/workspace/FilesTree.module.css
PavelLaptev: I thought about it, but decided to stick to the original right-aligned version.

image

I think we can tweak it later after using it for a bit.

PavelLaptev:
I also swapped the link and the kebab button, so the kebab menu won't shift the link button on hover.

image

OliverJAsh:
Nice, I'll make a change to always show the dependency icon button (in another PR).

PR #14477, opened 2026-06-26
Add open in browser ctx menu items & hotkeys

samhh:
Closes GB-1656.


OliverJAsh:
We saw a significant performance improvement when we hoisted hotkeys out of the rows, so I think we should aim to preserve that hoisting. We do have a precedent for looking up the selection in the head info, see for example `selectedBranchSegment`.

How about this?: [https://github.com/gitbutlerapp/gitbutler/pull/14479](<https://github.com/gitbutlerapp/gitbutler/pull/14479>)

PR #14479, opened 2026-06-26
Lite: hoist hotkeys

OliverJAsh:


samhh (approved):
Good shout, I forgot about that with the hotkeys.

In the future we should make `headInfo` contents cheaply available for lookups rather than iterating with each new selection with e.g. `findCommit`.

OliverJAsh:
Yeah I agree. I'd like to do that sooner rather than later otherwise performance is going to gradually decline.

PR #14487, opened 2026-06-26
Fix oplog restore leaving spurious uncommitted changes

krlvi:
## Problem

Updating the workspace — or any operation that moves the workspace commit — and then restoring or undoing it through the oplog left that operation's file changes behind in the worktree as spurious uncommitted changes. They were safe to discard, but shouldn't show up in the first place.

## Cause

`restore_snapshot` repointed `gitbutler/workspace` at the snapshot commit *before* checking out the snapshot's worktree. Both checkout backends — cv3 `safe_checkout` and the legacy git2 `checkout_workdir_tree` — diff from the current `HEAD` tree, so once the ref had been moved the from/to trees were identical, the diff was empty, and the worktree was never reverted.

## Fix

Capture the worktree's baseline tree before any refs move, and defer the workspace-ref update until after the checkout. The checkout then diffs from the pre-restore commit and reverts the worktree correctly. The cv3 path is given the captured baseline explicitly; the legacy path benefits because `HEAD` still points at the pre-restore commit during the checkout.

Note: the cv3 backend is covered by the regression tests here; the legacy backend isn't reproducible in the test harness (git2's index reconciliation masks it on small scenarios) and was verified by hand against a real repository.

### DIFF
--- a/crates/but/tests/but/command/undo.rs
+++ b/crates/but/tests/but/command/undo.rs
@@ -149,7 +149,6 @@ fn can_undo_but_discard_rename() {
 }
 
 #[test]
-#[ignore = "Test harness runs with cv3 feature flag, and but_core::worktree::safe_checkout_from_head does not restore the worktree file A for some reason. https://linear.app/gitbutler/issue/GB-1403/run-test-both-with-and-without-cv3-feature-flag"]
 fn can_undo_unapply() {
     let env = Sandbox::init_scenario_with_target_and_default_settings("one-stack");
     env.setup_metadata(&["A"]);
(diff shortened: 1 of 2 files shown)


crates/but/tests/but/command/undo.rs
```diff
@@ -149,7 +149,6 @@ fn can_undo_but_discard_rename() {
 }
 
 #[test]
-#[ignore = "Test harness runs with cv3 feature flag, and but_core::worktree::safe_checkout_from_head does not restore the worktree file A for some reason. https://linear.app/gitbutler/issue/GB-1403/run-test-both-with-and-without-cv3-feature-flag"]
 fn can_undo_unapply() {
```
krlvi, line 152: fyi @davidpdrsn

PR #14495, opened 2026-06-28
fix(tui): make arrow keys do the same as hjkl

davidpdrsn:


krlvi:
I wish our GUI could do this ❤️‍🩹

davidpdrsn:
> I wish our GUI could do this ❤️‍🩹

Yes 🫤

PR #14543, opened 2026-06-30
Prioritise display of file names in files panel

samhh:
Copies how we render file labels in the diff headers in overflow scenarios, which I personally find far more useful. Assuming this is what we want, let it be known that this causes the label text to shift when the kebab pops in or out.

Before:

Screenshot 2026-06-30 at 17 30 40

After:

Screenshot 2026-06-30 at 17 32 56

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/FilesTree.module.css
+++ b/apps/lite/ui/src/routes/project/$id/workspace/FilesTree.module.css
@@ -14,6 +14,22 @@
 	}
 }
 
+.filePath {
+	color: var(--text-3);
+
+	/* Overflow ellipsis on the left, pioritising filename. */
+	direction: rtl;
+	text-align: left;
+}
+
+.pathLast {
+	color: var(--text-1);
+
+	[data-selection-scope]:focus .fileRow[aria-selected="true"] & {
+		color: var(--text-1-invert);
+	}
+}
+
 .fileStatusBadge {
 	align-self: center;
 	width: 12px;
--- a/apps/lite/ui/src/routes/project/$id/workspace/FilesTree.tsx
+++ b/apps/lite/ui/src/routes/project/$id/workspace/FilesTree.tsx
@@ -280,6 +280,10 @@ const FileRow: FC<
 		selectProjectHasCheckedCommits(state, projectId),
 	);
 
+	const lastSepIdx = relativePath.lastIndexOf("/");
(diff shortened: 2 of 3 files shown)


PavelLaptev:
Looks good, thank you! I added one small change to increase the space between the file name and actions. I used `margin-right`, but feel free to change it if there’s a better way.

image

PavelLaptev (approved):

OliverJAsh:
Alternatively, we could do something like VS Code:



Zed has similar but slightly different:



This might be of interest as well:

[Raphael Schaad on Twitter / X](https://x.com/raphaelschaad/status/2034009580678553923)

apps/lite/ui/src/routes/project/$id/workspace/FilesTree.module.css
```diff
@@ -14,6 +14,22 @@
...
+	text-align: left;
+}
+
+.pathLast {
+	color: var(--text-1);
+
+	[data-selection-scope]:focus .fileRow[aria-selected="true"] & {
+		color: var(--text-1-invert);
+	}
+}
```
OliverJAsh, line 31: Instead of this, perhaps we can re-use `fadedText`. (I've been trying to keep the row component styles encapsulated in the row component.)

OliverJAsh (approved):

OliverJAsh:
> let it be known that this causes the label text to shift when the kebab pops in or out

This is quite noticeable. I think this wouldn't be a problem if we did something like VS Code / Zed pictured above. cc @PavelLaptev @samhh

samhh:
I'm down for that. Would you do that only here or also in the diff headers for consistency?

OliverJAsh:
Linear does it for both. I like that personally, WDYT?

PavelLaptev:
Looks good. Let's try. We need to make a few changes to improve readability:
- Increase the spacing between the file name and the file path.
- Decrease the file path opacity from `0.5` to `0.4`.

You can find the updated specs here in [Figma](https://www.figma.com/design/EBuHQGUcCaSw4Ln5uVpWkn/Lite?node-id=1810-22686&t=cHoh9ys0DsCr2GUC-1)

image

image

@OliverJAsh I remember you also wanted to show the "link" icon all the time, not just on hover. WDYT?

OliverJAsh:
gitbutlerapp/gitbutler#14561

PR #14545, opened 2026-06-30
fix(ai): sort changed files by path for deterministic commit messages

### LINKED ISSUE #14485: Sort changed files before building the AI commit message prompt
## Description

When generating an AI commit message, changed files are added to the prompt in a non-deterministic order. Regenerating on an unchanged working tree produces a different file order each time.

## Steps to reproduce

1. Configure Ollama as the AI provider.
2. Have a multi-file change in the working tree.
3. Click **Generate message** several times without changing anything.

## Result

The diff content is the same between requests, but the per-file order differs each time.

## Expected

A fixed file order for identical input.

## Effects

- Same changeset yields differently-ordered prompts, so messages vary between regenerations.
- Reordering invalidates the local runtime's prompt cache on each regeneration.

## Suggested fix

Sort the change set by file path before assembling the prompt.

## Related

- #8755

Env: GitButler 0.20.3 (macOS), Ollama provider.

fuleinist:
﻿## Description

Replaces the random shuffle in uildDiff() with a deterministic sort by file path.

### Problem

When generating an AI commit message, uildDiff() calls shuffle() which uses Math.random() to randomize the file order. Regenerating on the same working tree produces a different file order each time, which:
- Causes the AI to generate different commit messages for identical input
- Invalidates the local runtime's prompt cache on each regeneration

### Fix

- Remove the shuffle() function entirely
- Sort the formatted diff entries by their string representation (which starts with the file path)
- The upstream inputs are already file-tree sorted via sortLikeFileTree() — this preserves that stable order instead of overriding it with randomness

### Test update

The multi-hunk test previously accepted either ordering (shuffle was non-deterministic). Now it asserts a single expected output and verifies that both input orders produce the same result.

Fixes #14485

### DIFF
--- a/apps/desktop/src/lib/ai/service.test.ts
+++ b/apps/desktop/src/lib/ai/service.test.ts
@@ -459,14 +459,10 @@ describe.concurrent("buildDiff", () => {
 		expect(buildDiff([hunk1], 100).length).to.eq(100);
 	});
 
-	test("When provided multiple hunks, it joins them together with newlines", () => {
-		const expectedOutput1 = `${hunk1.filePath} - ${hunk1.diff}\n${hunk2.filePath} - ${hunk2.diff}`;
-		const expectedOutput2 = `${hunk2.filePath} - ${hunk2.diff}\n${hunk1.filePath} - ${hunk1.diff}`;
+	test("When provided multiple hunks, it sorts them by file path", () => {
+		const expectedOutput = `${hunk1.filePath} - ${hunk1.diff}\n${hunk2.filePath} - ${hunk2.diff}`;
 
-		const outputMatchesExpectedValue = [expectedOutput1, expectedOutput2].includes(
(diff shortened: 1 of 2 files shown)


Byron (commented):
Thanks a lot for taking a stab at this.

Just to reproduce what I remember, the reason for the shuffle may have been a workaround to get more spread of knowledge to be considered for commit messages, as the passed context may have a limit, at the cost of making it much less deterministic.

With that said, I think @estib-vega might be able to review this.

estib-vega:
Hey @fuleinist, we appreciate the submission!

@Byron is right, the shuffling is intentional. This is the motivation:
If the user is clicking on the 'generate' button multiple times, we can assume it's they want a different message.
Shuffling the diffs around will help them doing that.

We currently want to keep the shuffling, so will close down this PR. 
Please do ping me if you want to keep talking about this. This is our [Discord channel](https://discord.gg/WvfftQW3U)

PR #14561, opened 2026-07-01
Lite: display directory after filename

OliverJAsh:
https://linear.app/gitbutler/review/prioritise-display-of-file-names-in-files-panel-1b7f54f83e14#comment-0528b558

@PavelLaptev Feel free to tweak in a subsequent PR.

image

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/Details.module.css
+++ b/apps/lite/ui/src/routes/project/$id/workspace/Details.module.css
@@ -210,7 +210,7 @@
 
 .pathInit {
 	margin-inline-start: 6px;
-	color: var(--text-3);
+	color: var(--text-2);
 }
 
 .hunkHeaderRow {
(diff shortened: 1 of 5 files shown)


PavelLaptev:
@OliverJAsh I tweaked the opacity a little, if that works for you

OliverJAsh:
I feel like the contrast is a hard to read 🤔

PavelLaptev:
then `text-2` should work better.
Other areas you showed already use `text-2`

PavelLaptev:
Another issue is that the dark background is crushing the contrast. As and idea to explore, we can get back to bg colors for files and branches. Try avoiding a background color that is too gray, and instead use a white or light off-white background for the branches and files view + some borders?
If we are changing the "Changed files" section bg to a white bg, making this view light as well would keep the overall interface consistent 🤔 

image

OliverJAsh:
> As and idea to explore, we can get back to bg colors for files and branches

Yeah I'm also curious about this. I also wonder if we need to draw boxes around stacks or whether we could make things more compact (less boxes inside boxes) so we can show more commits at a time.

PR #14567, opened 2026-07-01
Kbd upd

PavelLaptev:
The previous design looked a bit 'dirty' on certain components when placed on non-white backgrounds.

With these changes, the KBD component looks lighter and adapts much better to all backgrounds.

Screenshot 2026-07-01 at 14 53 40

---

Before:
image

After:
image

---


PavelLaptev:
@OliverJAsh I know we're going to change the design of the current 'uncommitted changes' section, but I'd still like to get these fixes in.

OliverJAsh:
We also need to think about the command palette (cmd+k). It looks a bit broken now.

Perhaps keep the background here? I was also thinking it would be nice to keep the background in tooltips.

image

PavelLaptev:
I didn't like it because it feels too noisy, and I’d prefer to avoid having so many permutations. For now, should we just drop the KBD (keyboard shortcuts) from the buttons? They don't really work with the current design in buttons.

OliverJAsh:
Yeah I think that could work

PavelLaptev:
@OliverJAsh I've made some updates. The size of the KBD is relative depending on the usage.


https://github.com/user-attachments/assets/9940bc1d-c1b9-4bbc-b0b4-d028ea27eb82

https://github.com/user-attachments/assets/7e490e8d-76ed-482d-8d33-fa9f74ca4b8f



A few things I'm considering to add later:
- The 'invert' variant

image

- A modification to support putting KBD components inside buttons (I've temporarily removed them for now)

apps/lite/ui/src/components/Button.stories.tsx
```diff
@@ -184,3 +185,30 @@ export const WithIconStartAndEnd = meta.story({
 		</div>
 	),
 });
+
+export const WithKbd = meta.story({
```
OliverJAsh: Not sure we still need this?

OliverJAsh (approved):

apps/lite/ui/src/components/Button.stories.tsx
PavelLaptev: Right. I'll delete this

PR #14572, opened 2026-07-02
fix(but-core): record staged submodule gitlinks in commits

### LINKED ISSUE #14507: CLI: but commit and amend silently drop a staged submodule gitlink
### What happened

I added a git submodule to a repo that GitButler manages, then tried to commit it through the CLI. The `.gitmodules` file and my other code changes commit fine, but the submodule gitlink itself never makes it into any commit. It stays staged forever no matter what I do.

After `git submodule add` plus checking out a pinned tag inside the submodule, the superproject index correctly records the gitlink:

```
$ git ls-files -s cli-proxy/CLIProxyAPI
160000 b05a27e4d7087d62d9c5a96580888fdfc999f5b1 0	cli-proxy/CLIProxyAPI
```

But when I stage and commit through `but`, the resulting commit only contains `.gitmodules` and the regular file changes. The gitlink is dropped:

```
$ but stage cli-proxy/CLIProxyAPI -b my-branch
$ but commit my-branch -m "vendor submodule"
$ git show <new-commit> --stat
 .gitmodules        | 3 +++
 cli-proxy/Dockerfile | 11 ++++------
 2 files changed
```

Notice `cli-proxy/CLIProxyAPI` is missing from the commit. Running `but status` afterwards still shows it as staged:

```
[staged to my-branch]
  A cli-proxy/CLIProxyAPI
```

`but amend <file> <commit>` does not pick it up either. With the path it reports `Source 'cli-proxy/CLIProxyAPI' not found`, and with the cliId it runs without error but the commit is unchanged and the gitlink is still staged.

### Why this matters

The end result is a broken branch. The `.gitmodules` file lands but the commit that should record which commit the submodule points to never exists. On a fresh clone, `git submodule update --init` has no gitlink to resolve, so the submodule tree comes up empty. In my case that breaks a Docker build in CI that copies the submodule source, because the directory is empty.

There is no workaround through the CLI that I could find. The only way to actually record the gitlink is to drop out of GitButler and use raw git, which defeats the point of managing the repo with GitButler.

### Steps to reproduce

1. In a GitButler managed repo, run `git submodule add <url> path/to/sub`.
2. Optionally check out a specific tag or commit inside the submodule so the gitlink is non-default.
3. `but stage path/to/sub -b <branch>` then `but commit <branch> -m "..."`.
4. Inspect the commit with `git show <commit> --stat`. The gitlink path is absent.
5. `but status` still lists the gitlink as staged.

### Expected

`but commit` and `but amend` should record the submodule gitlink (mode 160000 entry) in the commit, the same way plain `git commit` does, so the recorded pointer matches the index.

### Environment

- but CLI 0.20.3
- macOS (Darwin 25.5.0)

This looks related to the submodule meta issue #10720. Filing separately because the specific failure here is the CLI commit and amend path silently dropping an already staged gitlink, rather than submodule support in general.

ameyypawar:
Fixes #14507.

## Bug
`but commit` and `but amend` silently drop a staged submodule gitlink. The `.gitmodules` file and regular changes commit fine, but the `160000` gitlink never makes it into the commit — leaving a broken branch (on a fresh clone, `git submodule update --init` has no pointer to resolve).

## Cause
In `but-core`'s `apply_worktree_changes`, a submodule gitlink is a whole-file change (no hunks), materialized via gix's `worktree_file_to_object`. That returns `None` for a submodule directory it can't open as a repository — and the CLI opens repos with **isolated** options, so opening the submodule always fails → the change is rejected (`WorktreeFileMissingForObjectConversion`) and dropped. The GUI works because its full open-options open the submodule successfully — the "two different code paths" noted in the issue.

## Fix
When `worktree_file_to_object` returns `None`, record the pinned commit from the index's authoritative `160000` entry (`EntryKind::Commit`) instead of rejecting the change.

## Test
Adds a CLI regression test that builds a real submodule, commits through GitButler, and asserts the gitlink is recorded in the tree (and no longer uncommitted). It fails without the fix and passes with it.

Note: the separate `but stage <sub> <branch>` "Branch not found" and `but amend <path> <commit>` "Source not found" resolution glitches mentioned in the issue are out of scope here — this addresses the silent gitlink drop in the commit/amend tree-build.

### DIFF
--- a/crates/but-core/src/tree/mod.rs
+++ b/crates/but-core/src/tree/mod.rs
@@ -230,10 +230,23 @@ pub fn apply_worktree_changes<'repo>(
                 Some((id, kind, _fs_metadata)) => {
                     base_tree_editor.upsert(rela_path, kind, id)?;
                 }
-                None => into_err_spec(
-                    possible_change,
-                    RejectionReason::WorktreeFileMissingForObjectConversion,
-                ),
+                // A submodule gitlink is a directory, not a worktree file, so `worktree_file_to_object`
+                // returns `None` whenever it can't open the submodule's repository - which is the case
+                // with the isolated open-options the CLI uses, so `but commit`/`amend` silently dropped
+                // staged submodules (#14507). The pinned commit is authoritative in the index, so record
+                // it directly from the `160000` entry rather than rejecting the change.
+                None => match index
+                    .entry_by_path(rela_path)
+                    .filter(|entry| entry.mode == gix::index::entry::Mode::COMMIT)
+                {
+                    Some(entry) => {
+                        base_tree_editor.upsert(rela_path, EntryKind::Commit, entry.id)?;
+                    }
+                    None => into_err_spec(
+                        possible_change,
+                        RejectionReason::WorktreeFileMissingForObjectConversion,
+                    ),
+                },
             }
         } else if let Some(worktree_changes) = &worktree_changes {
             let Some(worktree_change) = worktree_changes.iter().find(|c| {
--- a/crates/but/tests/but/command/commit.rs
+++ b/crates/but/tests/but/command/commit.rs
@@ -1678,6 +1678,86 @@ fn committing_to_existing_branch_via_short_id() {
         .success();
 }
 
+#[test]
+fn commit_records_staged_submodule_gitlink() -> anyhow::Result<()> {
+    // Regression for #14507: `but commit` silently dropped a staged submodule gitlink. The CLI
+    // opens repositories with isolated options, so gix's `worktree_file_to_object` can't open the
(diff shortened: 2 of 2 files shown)


Byron:
Thanks for giving it a shot.

However, I have to close this as I don't think it's the correct fix at all, it looks like a very specific workaround.

While it's great that there it adds a test in the `but` CLI, in the respective issue I [also pointed out](https://github.com/gitbutlerapp/gitbutler/issues/14507#issuecomment-4837773139) that this works in the UI.

So any fix should check what the UI does, and see if that can be ported over to the CLI. If `but-core` changes are needed, they also require their own tests.

PR #14579, opened 2026-07-02
fix(but): resolve bare paths for staged files in `commit -p` (#13318)

### LINKED ISSUE #13318: but commit -p doesn't accept file path when file is staged
### Version

0.19.7

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

The `-p` parameter of `but commit` accepts file paths but only when the file is unstaged.

`but status`

```
╭┄zz [unstaged changes]
┊     no changes
┊
┊  ╭┄i0 [staged to kr-branch-1]
┊  │ oo A foo.txt
┊  │
┊╭┄kr [kr-branch-1] (no commits)
├╯
┊
┴ 6df1bee [gb-local/main] 2026-04-15 Initial empty commit
```

`but commit -p foo.txt`

```
Error: Invalid file ID(s):
  'foo.txt' not found. Run 'but status' to see available file IDs.
```

_No response_

```shell

```

sidfishus:
## 🧢 Changes

Fix bare-path resolution in `IdMap` so `but commit -p foo.txt` works for staged files.
Add unit + integration tests for bare path (positive) and `zz:` prefix (negative).

## ☕️ Reasoning

**`non_colon_uncommitted_filename`:** I thought creating a separate test for cases where there's no colon / prefix makes more sense instead of adding them to the existing `colon_uncommitted_filename`. That's the reproducer for the bare-path case (#13318).

**`colon_uncommitted_filename`:** I only added `zz:assigned` → empty here, a guard so `zz:` stays scoped to the unassigned bucket and doesn't regress.

**`commit.rs`:** `commit_with_bare_path_on_staged_file` is the reproducer for that style of test. I also added `commit_with_zz_prefix_on_staged_file_fails` as the matching negative case. I didn't add comments because the tests are self explanatory and this is consistent with the other tests.

**Issue discussion:** Per #13318, fix the reported case where a single staged foo.txt wasn't found. Qualified IDs (oo, i0:foo.txt) still work. Duplicate paths across areas still hit the existing ambiguity error; no new disambiguation logic in this PR.

This seems like a suitable issue to pick up for my first contribution.

AI Notice: I have peer programmed this with an AI agent. I've spent a lot of time reproducing and understanding the problem, and considering how best to name the functions (I got confused between uncommitted files and uncommitted/unassigned (`zz`)). This is what I've finalised on, but happy to be led by somebody with more experience of the codebase.

## 🎫 Affected issues

Fixes: #13318

### DIFF
--- a/crates/but/src/id/mod.rs
+++ b/crates/but/src/id/mod.rs
@@ -729,11 +729,31 @@ impl IdMap {
 /// Private methods to individually parse what can appear on both side of a
 /// colon. (Some of them can also appear alone.)
 impl IdMap {
-    fn parse_uncommitted_filename<'a>(
+    /// Bare path token (e.g. `foo.txt`): any entry in [`Self::uncommitted_files`] with this path,
+    /// whether assigned to a stack or in the uncommitted area (`zz`).
+    ///
+    /// For stack-scoped resolution (e.g. `branch@{stack}:file.txt`), use
+    /// [`StackWithId::parse`] instead.
+    fn parse_uncommitted_filename<'a>(&'a self, element: &str) -> Vec<Box<dyn Node<'a> + 'a>> {
+        self.parse_uncommitted_filename_where(element, |_| true)
+    }
+
+    /// After [`UNCOMMITTED`] (`zz:`): entries in [`Self::uncommitted_files`] that are in the
+    /// uncommitted area (not assigned to a stack).
+    fn parse_uncommitted_area_filename<'a>(&'a self, element: &str) -> Vec<Box<dyn Node<'a> + 'a>> {
+        self.parse_uncommitted_filename_where(element, |hunk_assignment| {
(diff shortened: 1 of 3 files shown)


Byron:
Thanks a lot for giving it a shot!

I think this is an area where @jonathantanmy2 would have the most insight.

sidfishus:
To help with review and give proof here's the 2 new automated tests failing without the changes:

test-fail-without-fix-compact

And the end to end run failing without the changes:

e2e-without-fix

And then the 2 new tests succeeding with fix:

test-pass-with-fix-compact

And the end to end run succeeding with the fix:

e2e-with-fix

Byron:
Thanks a lot @sidfishus, this helps for sure!

Let me also reel in @davidpdrsn who has been working no a new implementation of `but commit`, to see if this fix is already included there as well.

PR #14581, opened 2026-07-02
Don't selection follow drag target

samhh:
Closes GB-1705.

### DIFF
--- a/apps/lite/ui/src/outline/mode.ts
+++ b/apps/lite/ui/src/outline/mode.ts
@@ -33,6 +33,7 @@ export type KeyboardTransferOperationMode = {
 /** @public */
 export type PointerTransferOperationMode = {
 	source: Operand;
+	target: Operand | null;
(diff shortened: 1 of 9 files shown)


OliverJAsh:
I'm unsure about 6e1daf7f273a64d272961b73bf13d060c18bb432. It adds some complexity to the code, and we're still showing the selection in the details panel, so I think there's still some value in showing selection styles in some form, so that it's clear which item is being shown in the details panel. This also matches the behaviour in VS Code / Finder.

Perhaps it would help if we used the selected + blurred style instead of selected + focused. This seems to be the behaviour in Finder.

OliverJAsh:
Another reason to keep selection styles is they will be relevant in the future when we add multi-selection, because dragging will use the mulit-selection.

OliverJAsh:
`restoreSelection` isn't needed for pointer transfer mode now. Do we move it?

PR #14587, opened 2026-07-02
Add initial CI status to workspace view

samhh:
Towards GB-1339.

The polling configuration is a bit naive, we can iterate.


samhh:
NB for observing refetch behaviour consider setting `refetchIntervalInBackground`. With this in mind perhaps we should lean harder on `staleTime` instead.

PR #14597, opened 2026-07-03
Make workspace tests hermetic against user-level settings

mtsgrd:
Tests were reading the real user `settings.json` — flip `unapplyV3Pgm` in your settings and the `but` TUI snapshot tests fail (`Expected to be in open workspace mode`). Root cause: `Context::from_repo` loaded AppSettings from the user config dir, and ~40 test call sites used it.

### What changed

`from_repo` is gone; callers now pick one:

```rust
Context::from_repo_for_testing(repo)             // compiled-in defaults, no disk
Context::from_repo_with_settings(repo, settings) // explicit, for production
```

- Tests that need the real settings-loading constructors (`discover`, `new_from_project_handle`, but-debug's `handle_args`) wrap in the new `but_testsupport::isolated_app_data_dir`, which points `E2E_TEST_APP_DATA_DIR` at a temp dir.
- The `but` binary loads settings once per process (`crate::app_settings()`, a `OnceLock`) instead of re-reading the file in `init_ctx`.
- CI: the rust-test job redirects `E2E_TEST_APP_DATA_DIR` and a final step fails if any test created files there, so this can't quietly regress.

### Verification

Ran the whole workspace suite with `E2E_TEST_APP_DATA_DIR` pointing at a nonexistent dir and checked it stays absent. One intentional exception: but-core's login-shell test runs your real shell, so it does whatever your zshrc does (on CI: nothing).

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -272,6 +272,10 @@ jobs:
     env:
       CARGO_TERM_COLOR: always
       GITBUTLER_TESTS_NO_CLEANUP: "1"
+      # Redirect the app-data dir so any test that loads user-level settings writes here.
+      # The final step asserts nothing did - see 'Assert test hermeticity'.
+      # (a literal path - the `runner` context is not available in job-level `env`)
+      E2E_TEST_APP_DATA_DIR: /tmp/hermeticity-guard
     defaults:
       run:
         shell: bash
@@ -335,6 +339,13 @@ jobs:
         run: |
           set -x
           (cd crates/gitbutler-filemonitor/vendor/debouncer && cargo test)
+      - name: Assert test hermeticity
+        run: |
+          if [ -e "$E2E_TEST_APP_DATA_DIR" ]; then
+            echo "::error::Tests wrote to the user-level app-data dir. Use Context::from_repo_for_testing or but_testsupport::isolated_app_data_dir instead of settings-loading constructors:"
+            find "$E2E_TEST_APP_DATA_DIR"
+            exit 1
+          fi
 
   rust-test-but-api-macros:
     needs: changes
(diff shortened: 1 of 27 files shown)


krlvi (approved):
good fix

.github/workflows/push.yaml
```diff
@@ -275,6 +272,10 @@ jobs:
     env:
       CARGO_TERM_COLOR: always
       GITBUTLER_TESTS_NO_CLEANUP: "1"
+      # Redirect the app-data dir so any test that loads user-level settings writes here.
+      # The final step asserts nothing did - see 'Assert test hermeticity'.
+      # (a literal path - the `runner` context is not available in job-level `env`)
+      E2E_TEST_APP_DATA_DIR: /tmp/hermeticity-guard
```
Byron, line 278: Nice one!

Byron (commented):
I love the CI check for tests that broke isolation!

I only took a brief look at the implementation and wanted to note that there already are some facilities to isolate a context, but the focus was on the `gix::Repository` being isolated. If memory serves, there also is path helper that allows to not read from an environment variable, which should also be used there.

With that said, all that this was done without knowledge of `temp-env`, which solves the "set an env-var only for one test" problem with fork magic. Most of our tests are probably heavy enough to not make that extra fork weigh it down much.

mtsgrd:
@Byron sorry, can I just clarify if you're saying we should land this or if I should make some changes? *edit: let's fix forward if necessary.

Byron:
> @Byron sorry, can I just clarify if you're saying we should land this or if I should make some changes? *edit: let's fix forward if necessary.

Sorry for the late response. Merging is fine, and I don't tink anything is in need of fixing here.

PR #14613, opened 2026-07-04
Run TUI commands in the active project directory

oaleynik:
## 🧢 Changes

Runs commands launched from `but tui` command mode in the active GitButler project directory.

This applies to both built-in `but` command mode and shell command mode.

## ☕️ Reasoning

When `but tui` is launched with `-C <repo>`, the TUI displays and operates on that repo through its active `Context`. However, command mode spawned child processes without setting their working directory, so they inherited the shell’s original cwd.

That could make a user run a command while looking at repo A in the TUI, but accidentally execute it in repo B.

The fix sets the child process cwd to `ctx.workdir_or_fail()?` before spawning.

https://github.com/user-attachments/assets/9719d8a7-cb17-49b4-af21-41fbb0a670a1

### DIFF
--- a/crates/but/src/command/legacy/status/tui/app/command_mode.rs
+++ b/crates/but/src/command/legacy/status/tui/app/command_mode.rs
@@ -1,5 +1,6 @@
 use std::{ffi::OsString, process::Command};
 
+use but_ctx::Context;
 use crossterm::event::Event;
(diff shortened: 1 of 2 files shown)


davidpdrsn:
Superseded by https://github.com/gitbutlerapp/gitbutler/pull/14877

PR #14620, opened 2026-07-05
fix(forge): honor project forge_override for self-hosted forges

### LINKED ISSUE #14319: GitLab integration broken in 0.20.1 (self-hosted instance
### Version

0.20.1

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

# GitLab integration broken in 0.20.1 (self-hosted instance)
  
## Steps to reproduce / behavior
 
- On 0.20.1, I get the following API error: `No forge could be determined for this repository.`
- After reverting back to 0.20.0, everything works correctly again.
 
## Additional details
 
In **Project Settings → Project**, there is a "Forge override" section with "GitLab" selected. However, on 0.20.1 the "Configure GitLab integration" section is missing — this section is present in 0.20.0.

ameyypawar:
Fixes #14319. Opening as a **draft** for review — the direction was confirmed by the maintainer investigation in the issue and by the reporter, so I'd welcome a self-hosted test and any steer on the shape before marking it ready.

## Bug
0.20.1 regressed self-hosted GitLab: a project with `forge_override = gitlab` shows "No forge could be determined" and the GitLab settings section disappears.

## Cause
547b873 moved forge resolution into Rust, but `but_forge::derive_forge_repo_info(url)` never consumed the project's `forge_override`. For a self-hosted host that isn't a recognized keyword (`gitlab.com`/`gitlab.*`) and before an account is configured, it returns `None` — so `forge_info()` returns `None` and `ForgeForm.svelte` hides the GitLab config (gated on `forgeInfo?.name === "gitlab"`).

## Fix
Resolve the forge type in one place, `resolve_forge_name`, in priority order:
1. the URL host (`github.com`, `gitlab.*`, …),
2. a configured account whose custom host matches,
3. the project's saved `forge_override` — **last resort**, mirroring pre-0.20.1's `if forgeType === "default" && forgeOverride`.

A recognized host or a matching account always wins, so a stale override can't mislabel a repo whose forge is already known. Owner/repo still come from the URL's `GenericProvider` regardless of how the type is resolved, so the override path produces the same `ForgeRepoInfo` the account path would.

The override (`legacy_project.forge_override`) is threaded through `forge_info`, `compare_branch_url`, and the forge API paths in but-api and the CLI auth path. The pure-URL identity comparison (`remote_urls_match`) passes `None`.

## Tests
- `resolve_forge_name` precedence — the override resolves a self-hosted host; a recognized host and a matching account each win over the override; none → `None`
- end-to-end `derive_forge_repo_info` on a self-hosted URL (owner/repo parsed correctly)
- `build_base_url` builds the self-hosted web URL from the override-derived info
- `from_override_str` is case/whitespace-insensitive

`cargo check --workspace` and clippy are clean.

## Notes / open questions
- Host detection is substring-based, so a self-hosted host embedding another forge's domain (e.g. `git.azure.company.com`) is still detected as that forge and the override won't correct it — a pre-existing limitation, left out of scope here.
- Account-vs-override ordering: a configured account (which carries credentials) is treated as authoritative over the saved preference. This doesn't affect the reported case (no account configured).
- I threaded an `Option<ForgeName>` parameter through `derive_forge_repo_info` and its callers; happy to switch to a centralized `ctx`-level helper if that's preferred.

The reporter offered to test a build on their self-hosted instance — that would be the strongest confirmation.

### DIFF
--- a/crates/but-api/src/legacy/forge.rs
+++ b/crates/but-api/src/legacy/forge.rs
@@ -23,6 +23,15 @@ pub fn push_remote_url(project_meta: &ProjectMeta, repo: &gix::Repository) -> Re
     project_meta.push_remote_url(repo)
 }
 
+/// Parse the project's saved forge preference (the Settings forge picker's
+/// `forge_override`) into a `ForgeName`. It's the last-resort fallback passed to
+/// `derive_forge_repo_info` so a forge the remote URL and configured accounts
+/// can't identify - e.g. self-hosted GitLab before an account is added - still
+/// resolves (#14319). `None` means no usable preference.
+fn forge_override(saved: Option<&str>) -> Option<ForgeName> {
+    saved.and_then(ForgeName::from_override_str)
+}
+
 fn review_template_content(file: FileInfo) -> Result<String> {
     if file.size.is_none() {
         return Ok(String::new());
@@ -49,7 +58,10 @@ pub fn pr_templates(ctx: &but_ctx::Context, forge: ForgeName) -> Result<Vec<Stri
 pub fn forge_provider(ctx: &Context) -> Result<Option<ForgeName>> {
     let project_meta = ctx.project_meta()?;
     let repo = ctx.repo.get()?;
-    let forge_repo_info = but_forge::derive_forge_repo_info(&remote_url(&project_meta, &repo)?);
+    let forge_repo_info = but_forge::derive_forge_repo_info(
+        &remote_url(&project_meta, &repo)?,
+        forge_override(ctx.legacy_project.forge_override.as_deref()),
+    );
     Ok(forge_repo_info.map(|info| info.forge))
 }
 
@@ -60,7 +72,10 @@ pub fn forge_provider(ctx: &Context) -> Result<Option<ForgeName>> {
 pub fn forge_info(ctx: &Context) -> Result<Option<but_forge::ForgeInfo>> {
     let project_meta = ctx.project_meta()?;
     let repo = ctx.repo.get()?;
-    Ok(but_forge::forge_info(&remote_url(&project_meta, &repo)?))
+    Ok(but_forge::forge_info(
+        &remote_url(&project_meta, &repo)?,
+        forge_override(ctx.legacy_project.forge_override.as_deref()),
+    ))
 }
 
 /// Web compare URL for a branch — drives the "Open in browser"
@@ -81,6 +96,7 @@ pub fn forge_compare_branch_url(
         &base,
         &branch,
         fork.as_deref(),
+        forge_override(ctx.legacy_project.forge_override.as_deref()),
     ))
 }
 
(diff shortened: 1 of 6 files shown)


PetrAltman:
I'm the reporter of #14319. Built this branch on Windows (`x86_64-pc-windows-msvc`) against my self-hosted GitLab and can confirm the main fix works: the "Configure GitLab integration" section is back in Project Settings and the "No forge could be determined" error is gone. Thanks for this!

One remaining issue that looks related to URL construction, in case it's in scope here:

The **"View MR"** button on the virtual branch (in the side column) opens GitLab at a truncated host — just `gitlab` instead of the full `gitlab.internal.example.com`. The short host actually resolves on my internal network (DNS search domain), but the page then fails with an SSL certificate error, because the cert is issued for the full FQDN `gitlab.internal.example.com`, not the short name. So the truncation is the root cause; the cert error is just the symptom.

The **"Open MR in browser"** button in the merge column opens the correct full URL (`gitlab.internal.example.com/...`) with no cert error.

So the two buttons seem to build the MR URL via different paths: the merge-column one resolves the self-hosted host correctly, but the virtual-branch "View MR" one keeps only the first host label and drops the rest of the domain. Happy to grab logs or test a follow-up change if useful.

(Host redacted — my real instance is an internal `*.int.*` domain.)

PetrAltman:
Update / root-cause find, reporter of #14319 here.

I tracked down what actually triggered this on my setup: my git `origin` remote was pointing at a short host (`gitlab`) instead of the full FQDN (`gitlab.internal.example.com`). After fixing origin to use the FQDN, the GitLab integration works correctly even on 0.21 (which doesn't include this PR).

That seems to explain the original error: in 0.20.1 forge resolution became host-based, and my short origin host matched neither the built-in `gitlab.*` patterns nor my configured account host (`gitlab.internal.example.com`), so it resolved to no forge — while the saved forge override wasn't consulted as a fallback. Which is exactly what this PR adds.

So the PR still looks valuable as a robustness fallback (honor the override when the origin host can't be matched — mirrors, SSH host aliases, etc.), even though my specific breakage came down to a misconfigured origin.

One related note on the two MR buttons: "View MR" (virtual branch, side column) appears to build its URL from the raw origin host, while "Open MR in browser" (merge column) uses the resolved forge instance URL. That's why they differed for me — with origin fixed to the FQDN, both work now. Might still be worth having "View MR" use the resolved forge URL for consistency, but it's no longer broken on my end.

ameyypawar:
Thanks for building it and confirming — glad the GitLab section and forge detection are back on your self-hosted instance.

And thanks for chasing the root cause. It lines up with the fix: when the origin host doesn't match a built-in pattern or a configured account host — as a short or aliased host won't — 0.20.1 resolved to no forge and never consulted the saved override. This PR restores that fallback, so a saved forge preference still resolves in those cases (mirrors, SSH host aliases, short internal hostnames).

The "View MR" host truncation looks like a separate area from this PR (which changes forge resolution, not MR-URL construction) — probably best as its own issue. I'd want to trace both buttons' URL paths before proposing a fix, so it targets the right spot.

Byron:
Thanks @ameyypawar for the fix, and @PetrAltman for taking a look.

Something I'd feel uncomfortable with is that support for the forge override now is explicitly needed in `but-api`, so it can easily be forgotten as well. It feels like it belongs closer to `but-forge`. With that said, @estib-vega would know best if this is the right direction.

PR #14623, opened 2026-07-06
Workspace graph: one substrate

mtsgrd:
People working in the workspace-graph core said it was hard to read. Size wasn't the problem. **No stage could be understood on its own** — each one repaired the output of the stage before it, and the last stage wrote its result back over the first stage's input. To answer "why is this branch in that lane?", you had to hold the whole pipeline in your head, in order, including the parts that had already been overwritten.

This replaces that machine with one where **every stage only reads the one before it**.

One fact organizes everything below: **everything GitButler shows is derived from git — except emptiness, and its order.** An empty branch's place in a stack, the order among branches sharing a commit, an empty stack's slot in the workspace merge: git cannot say these, so they are declared. Stacking, branch order, stack order — all ancestry and the merge's parent array. The system is shaped accordingly: a vanilla-git core, plus a small store for the facts git can't hold.

## The pipeline, both ways

Five steps, unchanged. What changes is what each one is allowed to do.

| | Step | master | this branch |
|---|---|---|---|
| **1** | **Read** | the repository + the declaration | *same inputs; the walk stops at convergence* |
| **2** | **Build** | a segment graph, then ten fixup traversals that rewrite each other | two tables, then four phases that only add |
| **3** | **Derive** | one 1,489-line producer deciding everything at once | a frame stage, a partition engine, a presence table |
| **4** | **Describe** | a stack is a chain; identity is scored | a stack is a DAG; identity is a lookup |
| **5** | **Write** | the view is copied back over the declaration | only operations write; the view never does |

<details>
<summary><b>The same comparison, drawn</b> — every stage side by side, with what each one is no longer allowed to do</summary>

</details>

---

## 1 · Read

The traversal's two stopping questions used to share one budget; they are now separate concepts with separate owners (`walk/`'s module docs state the rules). **Convergence** — have the seeds met? — is tracked by goals and ends the walk. The commit budget is **pagination**: how much history below the workspace↔target floor to materialize. The two never trade against each other.

One measured consequence: a target's local that is provably behind the target is a *fact carried on its seed* (one generation-cutoff ancestry check inside the commit-graph file, ~150 ns per visited commit) — not a corridor to walk. Previously that corridor was 98% of a 4,664-commit walk on this repository, growing with every upstream commit. Interleaved against current master on the same machine (warm, three runs each): this repository's graph build is **12.3 ms → 9.2 ms, 25% faster**; a merge-heavy 500k-commit stress repository goes **697 ms → 589 ms, 15% faster**. The saving is the corridor the walk no longer visits.

---

### One indirection less: `ObjectId` and `FullName`, with nothing in between

The things we actually hold are a commit id and a ref name. The thing the graph spoke was `petgraph::graph::NodeIndex` — so every question was a translation: identity in, handle out, resolve the handle to reach anything.

```rust
// master — you hand it the two identities you own…
fn segment_by_ref_name_or_commit_id(
    &self,
    ref_name: &gix::refs::FullNameRef,
    commit_id: gix::ObjectId,
) -> (Option<SegmentIndex>, Option<(SegmentIndex, CommitIndex)>)   // …and get back handles to resolve again
```

…implemented as a linear scan over `node_weights()`. Two consequences, and the second is why the core read as opaque:

- an index is only meaningful in the build that produced it, so anything holding one had to be fixed up whenever the structure moved
- **you could not read a query and know what it was about** — you had to trace where its index came from

And it did not stay contained. `SegmentIndex` was `pub`, and it spread: **298 uses across 32 files, escaping but-graph into seven other crates.** A library's addressing scheme had become the project's vocabulary.

Now the identity *is* the address:

```rust
fn node(&self, id: ObjectId) -> Option<&Commit>
```

The row number still exists — `by_id: HashMap<ObjectId, usize>` — but it is private and never leaves. `SegmentIndex` does not exist at all. Underneath are two plain tables:

```
CommitGraph ─────────────────────────────┐     RefLayout ──────────────────────────┐
  nodes           the rows               │       groups      per commit, ordered   │
  parent_edges    one per parent,        │                   ref groups            │
                  in parent order        │       facts       per ref, table order  │
  children        derived adjacency      │       amended_ws_parents                │
  generations     depth from a root      │                   the merge's slots     │
  by_id / by_ref  indexes, rebuildable   │                                         │
  tombstoned      soft delete            │   ← joined on commit id                 │
```

<details>
<summary><b>The leak this removes</b> — a full graph traversal that existed only to undo a container's iteration order</summary>

`post.rs`, on master:

> "Rebuild outgoing edges so petgraph traversal follows the source commit's `parent_order`. Petgraph traverses edges in **reverse creation order**, so edges are re-added from last parent to first parent."

```
git's order          what PetGraph hands back          what pass 1 restores
  p1 , p2       →          p2 , p1              →          p1 , p2
                          (reversed)                (re-inserted, one by one)
```

A full walk of the graph to arrive back at the order git already gave. Now the order is a column — `parent_edges[c] = [p1, p2]` — so there is nothing to repair and no pass that repairs it.
</details>

### Ten rewriting traversals become four additive phases

`init/post.rs` was 2,204 lines running ten passes over the whole graph in a fixed order — and their comments *were* the schedule:

```
1  rebuild_edges_in_parent_order
2  update_entrypoint_commit_id
3  fixup_remote_tracking_refs_and_maybe_split_segments   "before anything, cleanup the graph"
4  fixup_workspace_segments        "should be first — what follows names the segments it creates"
5  fixup_segment_names             "all non-workspace fixups must come first"
6  workspace_upgrades
7  ad_hoc_branch_stack_upgrades
8  set_entrypoint_to_ref_name      "after all the virtual branches were added"
9  improve_remote_segments         "better not to do this before workspaces are sorted"
10 compute_generation_numbers      "finally, once all segments were added"
```

Replaced by `build/`, one file per phase:

```
facts  →  plan  →  materialize  →  layout
```

> [!IMPORTANT]
> **The structural invariant:** no pass rewrites another pass's structure. Every mutation is additive construction, or a materializer applying a decision the plan already made — so no later phase can silently undo what an earlier one decided.

**A reference used to be structural.** Naming a commit mid-segment forced a split (`split_segment`, `maybe_split_segments`), so the graph's shape depended on which refs happened to exist. Now a ref is a row in a table — same commits, one entry added — and the graph has one shape for a given history, whatever refs exist.

---

### Naming is decided once, and written down

Given one commit carrying three branches — `A` (has metadata), `main` (the target's local), `B` (nothing recorded) — master decided the winner inline across post-processing passes 5 and 9, recorded nothing, and would name it differently if those passes ran in the other order.

Now it's one ordered ladder in `disambiguated_ref_at`, where **a tier only answers when exactly one branch fits** — a tie isn't broken, it falls through:

| | rule | on this example |
|---|---|---|
| 1 | not integrated, and exactly one branch carrying metadata | **→ A** |
| 2 | exactly one remote-tracked branch | two match — defer |
| 3 | on a stack top: metadata wins; among several, the target's own local | not a stack top |
| 4 | exactly one branch, whatever it is | three here — defer |

The verdict is stored in `RefLayout` at build time, so the next read doesn't decide again — and a test can assert it on its own.

### The frame becomes a stage instead of a local variable

`WorkspaceFrame` on master was private to `init.rs`, built by `workspace_frame` and consumed by `workspace_stacks` in the same pass, so no answer could be inspected apart from the lanes it produced. Three of its eight fields were positions in one particular graph:

```rust
ws_tip_segment_id:      SegmentIndex           // a position
entrypoint_sidx:        Option<SegmentIndex>   // a position
lower_bound_segment_id: Option<SegmentIndex>   // a position
```

The replacement is built once and then read-only, and every field is a commit id or a ref name — so the same frame means the same thing against any graph of the repository:

```rust
tip_commit / tip_ref_info     ObjectId · RefInfo
entry_ref  / entry_commit     FullName · ObjectId
lower_bound / _ref_name       ObjectId · FullName     // the floor, and what names it
missing_target_ref            Option<FullName>        // recorded but unresolvable — reported, not dropped
```

### Lanes come from a colouring, not a walk

Master attributed commits by walking first parents down from each workspace-merge parent. A lane was whatever that route reached — and **a branch the route missed landed in no lane at all**: not shown, not addressable, with nothing reporting that it went missing.

`derive_partition()` colours each commit for the tip that owns it, stopping at the integration base:

- exactly **one** tip reaches it → that tip's exclusive lane
- **two or more** → those tips are *one stack*, and the shared tail is stored once, not duplicated
- **zero**, because it rests on the bound → still named, because the merge holds it

Stackhood becomes a fact of the graph rather than an outcome of the traversal. The engine is **total**: every view it is handed yields stacks. A producer that could answer "nothing" would need a second producer to cover the difference, and that is how you get back to five authorities.

---

### A stack may be a DAG, natively

`Stack` was `Vec<StackSegment>` — adjacency positional, `(i, i+1)`. A fork had nowhere to live: a flat list carries no edges, so it became two stacks or one leg was lost. **The shape the UI showed was the only shape the record could hold.**

```rust
pub dag_edges: Option<Vec<(usize, usize)>>
```

- **in-stack fork** — the legs are siblings, listed under the merge, never re-threaded by adjacency
- **converged multi-tip** — tips converging above the base are one stack, shared tail stored once
- `None` for an unforked stack, so the common case costs nothing — and today's linear lanes are *derived* per tip by `expand_lanes`, never stored

### Identity is a lookup, not a competition

The id itself is stored in `VirtualBranches.toml`, written once when the stack was created — both columns read the same file. What differs is how a freshly derived lane is matched back to it.

| | master | this branch |
|---|---|---|
| how | every ref in view is scored; the highest wins | the lane takes the id of the declared stack owning a branch it reaches |
| stability | add a ref and the scores move — and which ref names the segment was itself settled by the ordered passes | the declared branches are input; nothing the read does can move them |
| result | the same lane can come back wearing a different id | the same lane comes back wearing the same id |

This is not cosmetic: the id is how the app recognises a lane across reads — what you had selected, the draft you were typing into.

---

## 5 · Write

The same `but_graph::Workspace` value was asked to be three things at once:

1. the picture on screen
2. the argument to every branch operation
3. **the saved declaration** — copied back by `workspace_reconciliation_input → reconcile_projected_stacks → make_heads_match`

Job 3 is gone. A rendering decision — which of several refs names a commit — can no longer become a durable workspace fact that the next derivation reads back. Operations record their own intent instead:

```
you run: but apply A   (while standing on main)

master              1. merge rebuilt  2. view derived  3. write-back copies the lanes in
this branch         1. apply writes the declaration    2. merge rebuilt, view derived from it
```

Both end at `stacks = [ main , A ]` — one *because that is what got drawn*, the other *because apply said so*.

Job 2 also narrows: operations still take the workspace for graph context, but they read the **total, pre-prune derivation** rather than the pruned display, so what the display leaves out can no longer change where an operation anchors.

The rules that took over the recording are stated in `but-workspace`'s crate docs:

- **Name what the merge holds** — every merge parent is named, so an applied branch owning no commits is still shown
- **Record what is applied** — applying a second branch declares the one you were already standing on
- **Never strand the user** — integration mints a stand-in rather than emptying the workspace; unapply refuses the last stack

---

## The editor: refs as positions, not nodes

The rebase engine is replaced (`but-rebase/graph_rebase/`). Master's editor wires references into ancestry as nodes; here a ref is a **position on a commit**. The difference is what nobody has to remember:

- a rebase rewrites ids in place, so refs follow **by standing still** — "rewrote the graph but forgot to move the branch" is not a bug this engine can have
- refs have no edges, so commit surgery **cannot** touch them — a parent list is a stored array, with no ref chains hiding inside it
- ref updates and metadata retirement are both *derived* from the finished graph, never maintained by hand — which fixes two live bugs (PR associations stranding under dead branch names; remove-then-recreate erasing fresh metadata)

Both worlds are visible in any verb — every line classifies itself by spelling. Inserting a commit *above* a target, the refs standing on it come along in the same statement:

```rust
let new_idx = self.store.commits.add_commit(spec);              // vanilla
self.store.commits.redirect_children(target_commit, new_idx);   // vanilla
self.store.commits.push_parent(new_idx, target_commit);         // vanilla
ref_ops::reposition_refs(                                       // extension —
    &mut self.store, target_commit, new_idx, Carry::Preserve);  //   the refs come along
```

Inserting *below* the same target, the extension line you'd expect does not exist — carry statements name parent entries by stable id, so the extension's facts follow the transplant unaided:

```rust
let new_idx = self.store.commits.add_commit(spec);              // vanilla
self.store.commits.transplant_parents(target_commit, new_idx);  // vanilla
self.store.commits.push_parent(target_commit, new_idx);         // vanilla
```

Measured against master's editor on the same operations: **93 → 0** caller-side reaches into raw storage, a **433-line** caller-side surgery module deleted, **5 calls → 1 verb** to move a commit. The GitButler layer itself is an annex, not a model: a pure-vanilla editor is **18% away, all subtraction** — the engine changes by zero lines. Reading: [`editor-worlds.md`](https://github.com/gitbutlerapp/gitbutler/blob/commit-graph-experiment/editor-worlds.md) (guided, from code excerpts) · [`refs-as-positions.md`](https://github.com/gitbutlerapp/gitbutler/blob/commit-graph-experiment/refs-as-positions.md) (the measurements).

---

## Scope

```
349 files changed, +53,938 / −40,979

crates/but-graph    +29,107 / −24,539     the model swap (effectively a rewrite)
crates/but-rebase    +9,681 /  −5,184     the graph editor
everything else      the operations layer, CLI, docs
```

Thirteen files in `but-graph/src` are deleted outright — 11,664 lines: all of `init/`, plus `segment.rs`, `api.rs`, `utils.rs`, `statistics.rs` and `projection/workspace/`.

## Testing

- CI is green across the board: **3,600+ Rust tests**, clippy `-D warnings` in all three feature configurations, `cargo doc`, all four Playwright shards, the blackbox suite, and the Windows check
- **A walk-invariants fuzzer is in the repository** ([`but-graph/tests/graph/fuzz.rs`](https://github.com/gitbutlerapp/gitbutler/blob/commit-graph-experiment/crates/but-graph/tests/graph/fuzz.rs)): random DAG repositories — managed workspaces included — asserting the model's laws: the entry is always collected, derivation is deterministic, the stored partition owns each commit exactly once, unlimited walks withhold only below-floor history. 2,000 repositories green at time of writing
- Performance, interleaved against current master: **15% faster** on a merge-heavy 500k-commit stress repository, **25% faster** on this one (numbers and cause in **1 · Read**)

## Reading it

Start with **the guided tour** — one small repository through every pipeline stage as asserted snapshots ([`but-graph/tests/graph/tour.rs`](https://github.com/gitbutlerapp/gitbutler/blob/commit-graph-experiment/crates/but-graph/tests/graph/tour.rs)), with a mutation chapter in [`but-rebase/tests/rebase/graph_rebase/tour.rs`](https://github.com/gitbutlerapp/gitbutler/blob/commit-graph-experiment/crates/but-rebase/tests/rebase/graph_rebase/tour.rs). Then the diff is stratified; in dependency order:

1. `but-graph/src/commit_graph.rs` + `ref_layout.rs` — the two tables
2. `but-graph/src/build/` — the four phases and the invariant
3. `but-graph/src/projection/frame.rs` — what the view is of
4. `but-graph/src/projection/partition.rs` — the colouring. If you read one file, read this one
5. `but-rebase/src/graph_rebase/` — the editor; its guided reading is `editor-worlds.md` at the repository root
6. `but-meta` + `but-workspace/src/branch/` — the write-back's removal, and the operations that took over

[`crates/WORKSPACE_MODEL.md`](https://github.com/gitbutlerapp/gitbutler/blob/commit-graph-experiment/crates/WORKSPACE_MODEL.md) is the shortest way in — glossary, the model's eight rules with their enforcement points, and the layering guidance. The snapshot glyphs every test speaks in are documented in `but-testsupport/src/graph.rs`.

<details>
<summary><b>Two decisions made during the rebase</b> — worth reading as part of the argument rather than as cleanup</summary>

- **Commit ids come from the derived stack, not the pruned display.** When `Workspace::stacks` changed element type, a hunk-dependency consumer stopped compiling. Two fixes compiled; the one taken reads ids from `SegmentStack` rather than reaching for `display_stacks()`, because that analysis is operation-facing and routing it through the pruned display is precisely the coupling this branch removes. A conflict resolved on principle rather than on what builds.
- **Branches that advanced outside the workspace are reported, not renamed.** A branch committed to directly points past the workspace merge. Its lane used to be silently anonymous, and naming the segment after that ref would be a lie. The display now derives an `advanced_outside` pair — the ref name, and the commits the workspace is missing — while the segment stays unnamed, because *a segment is named by the ref that points at it*. Same instinct as `missing_target_ref`: report the fact rather than quietly drop it.
</details>

## Known follow-ups

Deliberately not in this change:

> [!WARNING]
> **Stale metadata naming a deleted branch** makes the rebase materialize a ref pointing at the null object, after which the workspace stops deriving. Reachable whenever a branch is deleted outside GitButler while `vb.toml` still lists it.

- **`StackId` still crosses the legacy API boundary.** The core and the modern API target by ref name; `but-api/src/legacy/*` accepts a `StackId` and resolves it at the boundary. It retires with Tauri.
- **The CLI forces single-branch mode off for itself**, because every `but` command refuses to run outside `gitbutler/*`. Deleting that override is the acceptance test for the CLI alignment work.
- **The editor-op fuzzers are not yet in the repository** (the walk-invariants fuzzer is). They should land before anyone relies on CI to catch a regression in that area.

<details>
<summary><b>Where this converges</b> — the target picture this branch is one step toward</summary>

</details>


Byron:
Thanks for spiking this! To my mind it's super valuable to show it can be done, and if I'd start a greenfield project, I'd also aim for keeping the data structure unified, and I'd also not have a segmented graph because overall it's not worth it (or there is not enough evidence that it is).

Assuming this PR is good as it in every regard, I ran my final most gruesome test to check performance. The baseline for my nefarious gitlab repository in `master` is ~700ms for a workspace projection, with 690ms going into the traversal.

On this branch, the same run takes ~15s, most of which be accounted to the commit traversal.

So once this PR is a candidate for merging, this would certainly be something to look into.

<details><summary>Detailed runs</summary>
<p>

## Baseline

~700ms

```
❯ but-debug -ttt -C ~/dev-gb/gitlab graph
Workspace(📕🏘️:0:gitbutler/workspace[🌳] <> ✓refs/remotes/origin/master⇣31609 on e9f20c7) {
    id: 0,
    kind: Managed {
        ref_info: RefInfo {
            ref_name: FullName(
                "refs/heads/gitbutler/workspace",
            ),
            commit_id: Some(
                Sha1(bedfe4ffce8c3c224d9ee64891958866c68873c6),
            ),
            worktree: Some(
                Worktree {
                    kind: Main,
                    owned_by_repo: true,
                },
            ),
        },
    },
    stacks: [],
    metadata: Some(
        Workspace {
            ref_info: RefInfo { created_at: "2023-01-31 14:55:57 +0000", updated_at: None },
            stacks: [],
            target_ref: "refs/remotes/origin/master",
            target_commit_id: Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3),
            push_remote: "origin",
        },
    ),
    target_ref: Some(
        TargetRef {
            ref_name: FullName(
                "refs/remotes/origin/master",
            ),
            segment_index: NodeIndex(1),
            commits_ahead: 31609,
        },
    ),
    target_commit: Some(
        TargetCommit {
            commit_id: Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3),
            segment_index: NodeIndex(3),
        },
    ),
}
INFO     run [ 718ms | 1.75% / 100.00% ]
TRACE    ┝━ new_ondemand_git2_repo [ 333ns | 0.00% ] gitdir: "/Users/byron/dev-gb/gitlab/.git"
TRACE    ┝━ new_ondemand_db [ 83.0ns | 0.00% ] project_data_dir: "/Users/byron/dev-gb/gitlab/.git/gitbutler"
TRACE    ┝━ new_ondemand_app_cache [ 84.0ns | 0.00% ] cache_dir: Some("/Users/byron/Library/Caches/com.gitbutler.app.dev") | cache_mode: Disk
DEBUG    ┕━ Context::workspace_from_head [ 705ms | 0.00% / 98.25% ]
DEBUG       ┕━ Context::workspace_and_db_with_perm [ 705ms | 0.16% / 98.25% ]
TRACE          ┝━ DbHandle::new_at_path [ 540µs | 0.07% / 0.08% ]
TRACE          │  ┕━ run_migration [ 22.5µs | 0.00% ]
TRACE          ┝━ DbHandle::new_at_path [ 290µs | 0.04% / 0.04% ]
TRACE          │  ┕━ run_migration [ 15.5µs | 0.00% ]
TRACE          ┝━ Graph::from_commit_traversal [ 691ms | 91.47% / 96.20% ] tip: Sha1(bedfe4ffce8c3c224d9ee64891958866c68873c6) | ref_name: "refs/heads/gitbutler/workspace"
TRACE          │  ┕━ post_processed [ 33.9ms | 4.72% ]
WARN           │     ┕━ 🚧 [warn]: Ignoring stack Some(RefInfo { ref_name: FullName("refs/heads/gitbutler/target"), commit_id: Some(Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3)), worktree: None }) (NodeIndex(3)) as it is pruned
TRACE          ┝━ Graph::into_workspace [ 12.2ms | 1.71% ]
WARN           │  ┕━ 🚧 [warn]: Ignoring stack Some(RefInfo { ref_name: FullName("refs/heads/gitbutler/target"), commit_id: Some(Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3)), worktree: None }) (NodeIndex(3)) as it is pruned
TRACE          ┕━ DbHandle::new_at_path [ 498µs | 0.07% / 0.07% ]
TRACE             ┕━ run_migration [ 25.6µs | 0.00% ]
```

## This branch

~15s

```bash
❯ cargo run --release --bin but-debug -- -ttt -C ~/dev-gb/gitlab graph
    Finished [`release` profile [optimized]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 0.31s
     Running `target/release/but-debug -ttt -C /Users/byron/dev-gb/gitlab graph`
Workspace(📕🏘️:0:gitbutler/workspace[🌳] <> ✓refs/remotes/origin/master⇣31609 on e9f20c7) {
    id: 0,
    kind: Managed {
        ref_info: RefInfo {
            ref_name: FullName(
                "refs/heads/gitbutler/workspace",
            ),
            commit_id: Some(
                Sha1(bedfe4ffce8c3c224d9ee64891958866c68873c6),
            ),
            worktree: Some(
                Worktree {
                    kind: Main,
                    owned_by_repo: true,
                },
            ),
        },
    },
    stacks: [],
    metadata: Some(
        Workspace {
            ref_info: RefInfo { created_at: "2023-01-31 14:55:57 +0000", updated_at: None },
            stacks: [],
            target_ref: "refs/remotes/origin/master",
            target_commit_id: Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3),
            push_remote: "origin",
        },
    ),
    target_ref: Some(
        TargetRef {
            ref_name: FullName(
                "refs/remotes/origin/master",
            ),
            segment_index: 47653,
            commits_ahead: 31609,
        },
    ),
    target_commit: Some(
        TargetCommit {
            commit_id: Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3),
            segment_index: 1,
        },
    ),
}
INFO     run [ 15.3s | 0.19% / 100.00% ]
INFO     ┝━ ThreadSafeRepository::discover() [ 7.04ms | 0.01% / 0.05% ]
INFO     │  ┕━ open_from_paths() [ 5.33ms | 0.03% / 0.03% ]
INFO     │     ┕━ gix_odb::Store::at() [ 643µs | 0.00% ]
INFO     ┝━ ThreadSafeRepository::open() [ 266µs | 0.00% / 0.00% ]
INFO     │  ┕━ open_from_paths() [ 228µs | 0.00% / 0.00% ]
INFO     │     ┕━ gix_odb::Store::at() [ 118µs | 0.00% ]
TRACE    ┝━ new_ondemand_git2_repo [ 208ns | 0.00% ] gitdir: "/Users/byron/dev-gb/gitlab/.git"
TRACE    ┝━ new_ondemand_db [ 84.0ns | 0.00% ] project_data_dir: "/Users/byron/dev-gb/gitlab/.git/gitbutler"
TRACE    ┝━ new_ondemand_app_cache [ 42.0ns | 0.00% ] cache_dir: Some("/Users/byron/Library/Caches/com.gitbutler.app.dev") | cache_mode: Disk
DEBUG    ┕━ Context::workspace_from_head [ 15.2s | 0.00% / 99.76% ]
DEBUG       ┕━ Context::workspace_and_db_with_perm [ 15.2s | 0.02% / 99.76% ]
TRACE          ┝━ DbHandle::new_at_path [ 2.78ms | 0.02% / 0.02% ]
TRACE          │  ┕━ run_migration [ 198µs | 0.00% ]
TRACE          ┝━ DbHandle::new_at_path [ 278µs | 0.00% / 0.00% ]
TRACE          │  ┕━ run_migration [ 10.5µs | 0.00% ]
TRACE          ┝━ Graph::from_commit_traversal [ 15.2s | 99.49% ] tip: Sha1(bedfe4ffce8c3c224d9ee64891958866c68873c6)
TRACE          ┝━ Graph::into_workspace [ 36.0ms | 0.24% ]
WARN           │  ┕━ 🚧 [warn]: Ignoring stack Some(RefInfo { ref_name: FullName("refs/heads/gitbutler/target"), commit_id: Some(Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3)), worktree: None }) (1) as it is pruned
TRACE          ┕━ DbHandle::new_at_path [ 423µs | 0.00% / 0.00% ]
TRACE             ┕━ run_migration [ 19.5µs | 0.00% ]
```

</p>
</details>

mtsgrd:
@Byron thanks for flagging so early, I'll replicate and see if I can figure out what the material difference is(!) I'd also love to go over some of the details with you once I've had a chance to do another pass or two, I only _just_ managed to tie it together end to end last night.

mtsgrd:
@Byron would you mind running that perf test again? It should have come down by quite a bit, albeit not fully matching the 0.6s.

Byron:
I think it will be best if you had my test-setup, which is usually a local target branch that is far behind its remote counterpart. With the latest version from here, the issue is similar to what it was before, albeit improved at ~10s.

In theory, traversing the entire commit-graph once (in my repo) can't take more than 1.5s even without commit-graph cache. So reaching 10s means it's another class of performance issue entirely. What makes the current implementation fast is:

* decode and parse each commit at most once
* use a `git commit-graph` when present (this is trivial and I'd expect this to be present here already)
* try to reach relevant parts of the graph as fast as possible, and stop traversal as early as possible

### Update

It turns out (and was already visible in the performance trace) that the time is spent in the segment graph conversion. Everything else is fast as it was before. And knowing that the segment graph is only used for visualization in SVGs and tests, it's completely non-critical. Also I am pretty sure the performance of this can be improved as well.

So for me the only question left is why there are about 8k added lines of code - it should be about the same or less, ideally.

<details><summary>Details</summary>
<p>

```bash
❯ cargo run --release --bin but-debug -- -ttt -C ~/dev-gb/gitlab graph
    Finished [`release` profile [optimized]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 0.36s
     Running `target/release/but-debug -ttt -C /Users/byron/dev-gb/gitlab graph`
Workspace(📕🏘️:0:gitbutler/workspace[🌳] <> ✓refs/remotes/origin/master⇣31609 on e9f20c7) {
    id: 0,
    kind: Managed {
        ref_info: RefInfo {
            ref_name: FullName(
                "refs/heads/gitbutler/workspace",
            ),
            commit_id: Some(
                Sha1(bedfe4ffce8c3c224d9ee64891958866c68873c6),
            ),
            worktree: Some(
                Worktree {
                    kind: Main,
                    owned_by_repo: true,
                },
            ),
        },
    },
    stacks: [],
    metadata: Some(
        Workspace {
            ref_info: RefInfo { created_at: "2023-01-31 14:55:57 +0000", updated_at: None },
            stacks: [],
            target_ref: "refs/remotes/origin/master",
            target_commit_id: Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3),
            push_remote: "origin",
        },
    ),
    target_ref: Some(
        TargetRef {
            ref_name: FullName(
                "refs/remotes/origin/master",
            ),
            segment_index: 47653,
            commits_ahead: 31609,
        },
    ),
    target_commit: Some(
        TargetCommit {
            commit_id: Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3),
            segment_index: 1,
        },
    ),
}
INFO     run [ 10.4s | 0.28% / 100.00% ]
INFO     ┝━ ThreadSafeRepository::discover() [ 7.48ms | 0.02% / 0.07% ]
INFO     │  ┕━ open_from_paths() [ 5.64ms | 0.05% / 0.05% ]
INFO     │     ┕━ gix_odb::Store::at() [ 784µs | 0.01% ]
INFO     ┝━ ThreadSafeRepository::open() [ 270µs | 0.00% / 0.00% ]
INFO     │  ┕━ open_from_paths() [ 221µs | 0.00% / 0.00% ]
INFO     │     ┕━ gix_odb::Store::at() [ 114µs | 0.00% ]
TRACE    ┝━ new_ondemand_git2_repo [ 208ns | 0.00% ] gitdir: "/Users/byron/dev-gb/gitlab/.git"
TRACE    ┝━ new_ondemand_db [ 125ns | 0.00% ] project_data_dir: "/Users/byron/dev-gb/gitlab/.git/gitbutler"
TRACE    ┝━ new_ondemand_app_cache [ 84.0ns | 0.00% ] cache_dir: Some("/Users/byron/Library/Caches/com.gitbutler.app.dev") | cache_mode: Disk
DEBUG    ┕━ Context::workspace_from_head [ 10.4s | 0.00% / 99.65% ]
DEBUG       ┕━ Context::workspace_and_db_with_perm [ 10.4s | 0.03% / 99.65% ]
TRACE          ┝━ DbHandle::new_at_path [ 2.52ms | 0.02% / 0.02% ]
TRACE          │  ┕━ run_migration [ 51.6µs | 0.00% ]
TRACE          ┝━ DbHandle::open_existing_read_only_at_path [ 26.4µs | 0.00% ]
TRACE          ┝━ DbHandle::new_at_path [ 217µs | 0.00% / 0.00% ]
TRACE          │  ┕━ run_migration [ 11.1µs | 0.00% ]
TRACE          ┝━ Graph::from_tip [ 10.4s | 0.16% / 99.39% ] seed: Sha1(bedfe4ffce8c3c224d9ee64891958866c68873c6)
TRACE          │  ┝━ walker::traverse [ 640ms | 6.13% ]
TRACE          │  ┝━ CommitGraph::from_walk_outcome [ 14.0ms | 0.13% ]
TRACE          │  ┝━ enrichment_inputs [ 257ms | 2.46% ]
TRACE          │  ┝━ gather_and_plan [ 30.9ms | 0.14% / 0.30% ]
TRACE          │  │  ┝━ facts [ 9.14ms | 0.09% ]
TRACE          │  │  ┕━ chain_plan [ 7.56ms | 0.07% ]
TRACE          │  ┝━ graph_from_commit_graph [ 9.40s | 0.00% / 90.02% ] commits: 80086
TRACE          │  │  ┕━ segment_data::build [ 9.40s | 90.02% ]
TRACE          │  ┝━ apply_ad_hoc_orders [ 83.0ns | 0.00% ]
TRACE          │  ┝━ derive_ref_positions [ 14.7ms | 0.14% ]
TRACE          │  ┕━ render_segment_graph [ 3.50ms | 0.03% ]
TRACE          ┝━ Graph::into_workspace [ 20.8ms | 0.20% ]
WARN           │  ┕━ 🚧 [warn]: Ignoring stack Some(RefInfo { ref_name: FullName("refs/heads/gitbutler/target"), commit_id: Some(Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3)), worktree: None }) (1) as it is pruned
TRACE          ┕━ DbHandle::new_at_path [ 431µs | 0.00% / 0.00% ]
TRACE             ┕━ run_migration [ 19.4µs | 0.00% ]

gitbutler.commit-graph-experiment on  commit-graph-experiment [$≡] took 10s
❯ but-debug -ttt -C ~/dev-gb/gitlab graph
Workspace(📕🏘️:0:gitbutler/workspace[🌳] <> ✓refs/remotes/origin/master⇣31609 on e9f20c7) {
    id: 0,
    kind: Managed {
        ref_info: RefInfo {
            ref_name: FullName(
                "refs/heads/gitbutler/workspace",
            ),
            commit_id: Some(
                Sha1(bedfe4ffce8c3c224d9ee64891958866c68873c6),
            ),
            worktree: Some(
                Worktree {
                    kind: Main,
                    owned_by_repo: true,
                },
            ),
        },
    },
    stacks: [],
    metadata: Some(
        Workspace {
            ref_info: RefInfo { created_at: "2023-01-31 14:55:57 +0000", updated_at: None },
            stacks: [],
            target_ref: "refs/remotes/origin/master",
            target_commit_id: Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3),
            push_remote: "origin",
        },
    ),
    target_ref: Some(
        TargetRef {
            ref_name: FullName(
                "refs/remotes/origin/master",
            ),
            segment_index: NodeIndex(1),
            commits_ahead: 31609,
        },
    ),
    target_commit: Some(
        TargetCommit {
            commit_id: Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3),
            segment_index: NodeIndex(3),
        },
    ),
}
INFO     run [ 686ms | 3.22% / 100.00% ]
TRACE    ┝━ new_ondemand_git2_repo [ 375ns | 0.00% ] gitdir: "/Users/byron/dev-gb/gitlab/.git"
TRACE    ┝━ new_ondemand_db [ 83.0ns | 0.00% ] project_data_dir: "/Users/byron/dev-gb/gitlab/.git/gitbutler"
TRACE    ┝━ new_ondemand_app_cache [ 125ns | 0.00% ] cache_dir: Some("/Users/byron/Library/Caches/com.gitbutler.app.dev") | cache_mode: Disk
DEBUG    ┕━ Context::workspace_from_head [ 664ms | 0.01% / 96.78% ]
DEBUG       ┕━ Context::workspace_and_db_with_perm [ 664ms | 0.59% / 96.78% ]
TRACE          ┝━ DbHandle::new_at_path [ 3.16ms | 0.45% / 0.46% ]
TRACE          │  ┕━ run_migration [ 63.9µs | 0.01% ]
TRACE          ┝━ DbHandle::open_existing_read_only_at_path [ 40.0µs | 0.01% ]
TRACE          ┝━ DbHandle::new_at_path [ 374µs | 0.05% / 0.05% ]
TRACE          │  ┕━ run_migration [ 14.4µs | 0.00% ]
TRACE          ┝━ Graph::from_commit_traversal [ 651ms | 91.50% / 94.79% ] tip: Sha1(bedfe4ffce8c3c224d9ee64891958866c68873c6) | ref_name: "refs/heads/gitbutler/workspace"
TRACE          │  ┕━ post_processed [ 22.6ms | 3.29% ]
WARN           │     ┕━ 🚧 [warn]: Ignoring stack Some(RefInfo { ref_name: FullName("refs/heads/gitbutler/target"), commit_id: Some(Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3)), worktree: None }) (NodeIndex(3)) as it is pruned
TRACE          ┝━ Graph::into_workspace [ 5.62ms | 0.82% ]
WARN           │  ┕━ 🚧 [warn]: Ignoring stack Some(RefInfo { ref_name: FullName("refs/heads/gitbutler/target"), commit_id: Some(Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3)), worktree: None }) (NodeIndex(3)) as it is pruned
TRACE          ┕━ DbHandle::new_at_path [ 369µs | 0.05% / 0.05% ]
TRACE             ┕━ run_migration [ 17.8µs | 0.00% ]
```

</p>
</details>

Byron:
I had Codex run an analysis on where the SLOC additions are located.

I had it check for SegmentData is used as well, and it seems to still be in the critical path, i.e. is used for the workspace projection. Maybe that changes the next time this is pushed.

----

Compared with local `master` using `master...HEAD`:

`219 files changed, 26120 insertions(+), 17833 deletions(-)`, net `+8287`.

The net growth mostly comes from:

| Area | Added | Deleted | Net |
|---|---:|---:|---:|
| `crates/but-graph` | 15,342 | 10,931 | `+4,411` |
| `crates/but-rebase` | 6,023 | 2,728 | `+3,295` |
| `crates/but-workspace` | 3,636 | 3,579 | `+57` |
| everything else | 1,119 | 595 | `+524` |

Main source: `but-graph` replaced old `src/init` code with larger new graph/build/walk machinery.

Big buckets:

- `crates/but-graph/src/build/*`: `+4,509`
- `crates/but-graph/src/walk/*` new files: `+3,068`
- new graph modules like `commit_graph.rs`, `segment_graph.rs`, `ref_layout.rs`: `+1,563`
- deleted old `crates/but-graph/src/init/*`: `-5,648`
- `crates/but-rebase/src/graph_rebase/editor_graph.rs`, `layout.rs`, `positions.rs`: `+1,883`
- `but-rebase` test expansion/churn: net `+838`

Largest individual net additions:

- `crates/but-graph/src/walk/mod.rs`: `+1,638`
- `crates/but-graph/src/build/segment_data.rs`: `+1,454`
- `crates/but-graph/src/walk/walker.rs`: `+1,055`
- `crates/but-graph/src/commit_graph.rs`: `+1,037`
- `crates/but-rebase/src/graph_rebase/editor_graph.rs`: `+928`
- `crates/but-graph/src/build/plan.rs`: `+895`
- `crates/but-graph/src/build/mod.rs`: `+835`

So the short answer: the ~8k net increase is mostly new `but-graph` graph construction/walking infrastructure plus new `but-rebase` editor graph/layout machinery, partly offset by deleting the old `but-graph/src/init` implementation.

mtsgrd:
Thanks @Byron! As a last piece I'll see if there's a way we can rely solely on CommiGraph and fold in the SegmentGraph type and related machinery. That should in theory get the line counts closer to parity.

Byron:
While I still didn't dare to look at the code, I think with +3k lines and the performance problem fixed (just ~10% slower), I don't see any obvious issue with it anymore 🎉.

<details><summary>Details</summary>
<p>

```
❯ hyperfine 'but-debug -t -C ~/dev-gb/gitlab graph' './target/release/but-debug -t -C ~/dev-gb/gitlab graph'
Benchmark 1: but-debug -t -C ~/dev-gb/gitlab graph
  Time (mean ± σ):     621.9 ms ±  12.7 ms    [User: 358.9 ms, System: 261.4 ms]
  Range (min … max):   611.4 ms … 654.3 ms    10 runs

Benchmark 2: ./target/release/but-debug -t -C ~/dev-gb/gitlab graph
  Time (mean ± σ):     692.8 ms ±   9.1 ms    [User: 500.4 ms, System: 1236.9 ms]
  Range (min … max):   681.0 ms … 710.5 ms    10 runs

Summary
  but-debug -t -C ~/dev-gb/gitlab graph ran
    1.11 ± 0.03 times faster than ./target/release/but-debug -t -C ~/dev-gb/gitlab graph

gitbutler.commit-graph-experiment on  commit-graph-experiment [$≡] took 13s
❯ but-debug -t -C ~/dev-gb/gitlab graph
Workspace(📕🏘️:0:gitbutler/workspace[🌳] <> ✓refs/remotes/origin/master⇣31609 on e9f20c7) {
    id: 0,
    kind: Managed {
        ref_info: RefInfo {
            ref_name: FullName(
                "refs/heads/gitbutler/workspace",
            ),
            commit_id: Some(
                Sha1(bedfe4ffce8c3c224d9ee64891958866c68873c6),
            ),
            worktree: Some(
                Worktree {
                    kind: Main,
                    owned_by_repo: true,
                },
            ),
        },
    },
    stacks: [],
    metadata: Some(
        Workspace {
            ref_info: RefInfo { created_at: "2023-01-31 14:55:57 +0000", updated_at: None },
            stacks: [],
            target_ref: "refs/remotes/origin/master",
            target_commit_id: Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3),
            push_remote: "origin",
        },
    ),
    target_ref: Some(
        TargetRef {
            ref_name: FullName(
                "refs/remotes/origin/master",
            ),
            segment_index: NodeIndex(1),
            commits_ahead: 31609,
        },
    ),
    target_commit: Some(
        TargetCommit {
            commit_id: Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3),
            segment_index: NodeIndex(3),
        },
    ),
}
INFO     run [ 674ms | 100.00% ]
WARN     ┝━ 🚧 [warn]: Ignoring stack Some(RefInfo { ref_name: FullName("refs/heads/gitbutler/target"), commit_id: Some(Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3)), worktree: None }) (NodeIndex(3)) as it is pruned
WARN     ┕━ 🚧 [warn]: Ignoring stack Some(RefInfo { ref_name: FullName("refs/heads/gitbutler/target"), commit_id: Some(Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3)), worktree: None }) (NodeIndex(3)) as it is pruned

gitbutler.commit-graph-experiment on  commit-graph-experiment [$≡]
❯ ./target/release/but-debug -t -C ~/dev-gb/gitlab graph
Workspace(📕🏘️:gitbutler/workspace[🌳] <> ✓refs/remotes/origin/master⇣31609 on e9f20c7) {
    kind: Managed {
        ref_info: RefInfo {
            ref_name: FullName(
                "refs/heads/gitbutler/workspace",
            ),
            commit_id: Some(
                Sha1(bedfe4ffce8c3c224d9ee64891958866c68873c6),
            ),
            worktree: Some(
                Worktree {
                    kind: Main,
                    owned_by_repo: true,
                },
            ),
        },
    },
    stacks: [],
    metadata: Some(
        Workspace {
            ref_info: RefInfo { created_at: "2023-01-31 14:55:57 +0000", updated_at: None },
            stacks: [],
            target_ref: "refs/remotes/origin/master",
            target_commit_id: Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3),
            push_remote: "origin",
        },
    ),
    target_ref: Some(
        TargetRef {
            ref_name: FullName(
                "refs/remotes/origin/master",
            ),
            commits_ahead: 31609,
        },
    ),
    target_commit: Some(
        TargetCommit {
            commit_id: Sha1(e9f20c7c9e73ed04592e317f9021b39808f839d3),
        },
    ),
}
INFO     run [ 853ms | 100.00% ]
```

</p>
</details>

PR #14633, opened 2026-07-06
fix(desktop): show update banner again on manual check after dismiss (#12898)

### LINKED ISSUE #12898: "Check for updates" doesn't do anything if new version is available
### Version

0.5.1940

### Operating System

Windows

### Distribution Method

msi (Windows)

### Describe the issue

If you open GitButler and a new version is available, it will show that in a popup on the left bottom corner. However, if you close that popup and press the "`Check for updates`" button, no new popup will appear.

1. Open older GitButler version
2. See "`new version available`" popup
3. Close popup
4. Click "`check for updates`"
5. Popup doesn't appear

Version popup should reappear showing a version is available. Now, you need to restart the application to see it again.

```shell

```

sidfishus:
## 🧢 Changes

Fix **Check for updates** so the update banner reappears after the user dismissed it.

- Show the banner on manual checks even when `seenVersion` matches (`isUnseenVersion || manualCheck`).
- Reset `manualCheck` in the `finally` block of `checkForUpdate` so it only applies to the current call.

Add unit and component tests for the dismiss → manual-check flow.

Also adjust the existing `should display up-to-date on manual check` test in `AppUpdater.test.ts` (see Reasoning).

## ☕️ Reasoning

Dismissing the banner clears the UI but leaves `seenVersion` set, which correctly suppresses background auto-checks for a version the user already saw. **Check for updates** is explicit though.

Manual checks now bypass the `seenVersion` gate. Auto-check behaviour is unchanged.

**`AppUpdater.test.ts`:** The existing `should display up-to-date on manual check` test is unrelated to the dismiss bug, but it started failing with this PR. On render, `AppUpdater` triggers a background auto-check; the test also fires a manual check. Resetting `manualCheck` in `finally` means a late-finishing mount auto-check can wipe the manual result. The fix is to let the mount auto-check finish first (`advanceTimersToNextTimerAsync`), then run the manual check which is also the same pattern as the other tests in that file.

**Manual testing:** The unit/component tests mock `checkUpdate`, but clicking through in `pnpm dev:desktop` is not possible without workarounds: the real Tauri `checkUpdate` hangs indefinitely in dev. For local click-through I temporarily stubbed `checkUpdate` behind `VITE_STUB_UPDATER=true` so it returned a fake update (`99.99.99-stub`) immediately. Flow tested: banner on startup → dismiss → **File → Check for updates…**. With the fix commented out, the banner stayed hidden (bug). With the fix restored, it reappeared. The stub was removed before this PR.

## 🎫 Affected issues

Fixes: #12898

### DIFF
--- a/apps/desktop/src/components/shared/AppUpdater.test.ts
+++ b/apps/desktop/src/components/shared/AppUpdater.test.ts
@@ -63,11 +63,34 @@ describe("AppUpdater", () => {
 		expect(button).toBeVisible();
 	});
 
+	test("should display download button again on manual check after dismiss", async () => {
+		vi.spyOn(backend, "checkUpdate").mockReturnValue(
+			mockUpdate({
+				version: "1",
+				body: "release notes",
+			}),
+		);
+
+		render(AppUpdater, { context });
+		await vi.advanceTimersToNextTimerAsync();
+
+		expect(screen.getByTestId("download-update")).toBeVisible();
+
+		updater.dismiss();
+		await vi.advanceTimersToNextTimerAsync();
+		expect(screen.queryByTestId("update-banner")).toBe(null);
+
+		await updater.checkForUpdate(true);
+		await vi.advanceTimersToNextTimerAsync();
+
+		expect(screen.getByTestId("download-update")).toBeVisible();
+	});
+
 	test("should display up-to-date on manual check", async () => {
 		vi.spyOn(backend, "checkUpdate").mockReturnValue(mockUpdate(null));
 		const { getByTestId } = render(AppUpdater, { context });
-		updater.checkForUpdate(true);
 		await vi.advanceTimersToNextTimerAsync();
+		await updater.checkForUpdate(true);
 
 		const button = getByTestId("got-it");
 		expect(button).toBeVisible();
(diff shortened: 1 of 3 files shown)


Byron (approved):
Thanks a lot for the fix and the manual testing performed here. It's always great to also see before/after videos, which can help to reduce the amount of manual testing the reviewer has to conduct.

Having looked only at the code, this looks like a surgical fix with added test coverage. While I didn't verify the tests actually break if the fix is removed, I think the final check could be conducted by @mtsgrd who is more in the UI-flow than me.

mtsgrd:
Awesome, looks great, thank you @sidfishus!

sidfishus:
AppUpdater test fail
AppUpdater test pass

https://github.com/user-attachments/assets/6563c5a3-5eeb-4b9a-ad91-d3402f1ebae0


https://github.com/user-attachments/assets/447a995d-bd48-4380-b662-e2b33f9aa4e8

updater test fail
updater test pass

Hi @Byron @mtsgrd ,

It's a bit late now, but I've attached screenshots / video of without the fix and with the fix for the 2 new tests, as well as the manual testing. I'll remember to do this for the next PR I submit, thanks for the feedback and review.

To manually test I made Tauri override 'IBackend.checkUpdate' and return a new version of 99.99.99-stub so on start up it sees the version change.

Thanks

PR #14645, opened 2026-07-07
checkout: only allow checkout from HEAD

jonathantanmy2:
It's enforced that checkout is only allowed from HEAD anyway.


Caleb-T-Owens (approved):
Nice

PR #14661, opened 2026-07-07
Render image diffs in the but tui details pane

krlvi:
Changed images previously showed only a "no diff available" placeholder in the details pane. They now render as pictures: before on top, after below, at equal scale, each with a caption of its dimensions and file size. Works for uncommitted changes, commits, committed files, and branch diffs.

Rendering uses ratatui-image's graphics support — kitty, iTerm2, or sixel where the terminal supports one. The capability query must run before entering the alternate screen and is only sent when environment variables suggest a graphics-capable terminal, because ratatui-image's query thread swallows a keystroke and breaks raw mode when the query goes unanswered. Terminals without graphics keep the placeholder text, as does the but-2 ANSI renderer, which shares the diff-rendering engine but never requests image decoding.

Notable decisions:

- Images are encoded for the terminal once at full display size via SlicedProtocol; scrolling renders only the visible rows with an offset (browser-style clipping), so scrolling never re-encodes or re-uploads and does not flicker.
- Decoded images are cached by blob ids — worktree files by path/size/mtime so edits invalidate naturally — and pre-scaled off the UI thread; terminal encodes are cached by content fingerprint and debounced until the selection rests, keeping commit-to-commit navigation smooth.
- Frames are bracketed in synchronized-update escapes (mode 2026) so terminals present them atomically; unsupported terminals ignore this.
- The image decoding crates are added to the dev-profile opt-level overrides, following the existing gix precedent — unoptimized PNG decoding made debug builds unusably slow.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -600,6 +600,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
+name = "base64-simd"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "339abbe78e73178762e23bea9dfd08e697eb3f3301cd4be981c0f78ba5859195"
+dependencies = [
+ "outref",
+ "vsimd",
+]
+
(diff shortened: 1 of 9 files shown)


krlvi:
@davidpdrsn leaving this for you here to use as a parts bin

PR #14666, opened 2026-07-08
fix(but): remove S from the default less options

slarse:
S triggers paging if the content is too wide. This is extremely annoying for everyone involved, it's better to just let it wrap.

Unfortunately, there is no way to just truncate without paging in less.


slarse:
@krlvi FYI :)

PR #14673, opened 2026-07-08
fix(but): render anonymous workspace segments instead of wedging status

### LINKED ISSUE #14497: Stale parent in workspace merge commit wedges CLI ("Could not find branch CLI id '' in IdMap") and shows "Unnamed segment"
### Version

App 0.20.4 / `but` CLI 0.20.3

### Operating System

macOS (Apple Silicon, Darwin 25.5.0)

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

A stale parent in the GitButler workspace (octopus) merge commit puts the whole project into an unrecoverable state from both the GUI and the CLI.

**Symptoms**

- GUI shows a lane titled **"Unnamed segment"** holding the commits of a branch that was supposed to be unapplied.
- `but status`, `but status -fv`, `but apply <branch>`, and `but unapply <branch>` all fail (or silently no-op) with:
  ```
  Error: Could not find branch CLI id '' in IdMap
  ```
- GUI **Update workspace** (integrate upstream) fails:
  ```
  API error: (workspace_integrate_upstream)
  Failed to merge bases while cherry picking commit <workspace-commit>.
  Encountered a conflict while merging the commit's merge bases: ...
  ```
  CLI `but pull` reports: `Error during integration: merge conflict when computing workspace tree`.

**Root cause (after digging in)**

The `gitbutler/workspace` merge commit had 6 parents. Five matched the recorded stack heads (the target base + 4 applied stacks). The 6th parent was an **old head of a branch that had been unapplied/removed** — its SHA matched no current stack head in `vb_stack_heads`.

```
workspace commit parents:
  46ebe1c4  local-tmp-branch          (head matches)
  d59a2ed6  target base               (matches)
  2c1c4590  <stale reflection head>   (NO stack claims this) <-- orphan
  be451d74  jira-skills-kan           (head matches)
  68ba8dd7  static-skills-reminder    (head matches)
  708b9ba9  auto-mode                 (head matches)
```

The unapplied branch's recorded head was `c07c757` while the workspace parent was the older `2c1c4590` (same commit message, different SHA — the branch had since been amended/rebased, and it was also already merged upstream via yet another SHA). Because that parent maps to no stack, the lane is rendered with an **empty CLI id**, and the `IdMap` lookup throws, which aborts `status` and every mutation that first enumerates the workspace — so there is no in-product way to apply, unapply, or integrate out of the state.

This looks like it was created when a branch that was already merged upstream (via different SHAs) was unapplied/removed without the workspace merge commit being rebuilt to drop its parent.

**How I recovered** (for reference — all in-product paths were blocked)

1. Manually set the offending stack `in_workspace = false` in both `virtual_branches.toml` and `.git/gitbutler/but.sqlite` (`vb_stacks`).
2. With the orphan stack no longer in the applied set, `but pull --check` reported the 4 real branches as `[ok]` and excluded the orphan; `but pull` then advanced the base and rebased the 4 branches cleanly, dropping the stale parent.

### How to reproduce

Approximate sequence (the exact trigger seems to be unapply/remove of an already-merged branch):

1. Have several applied virtual branches in the workspace.
2. One branch gets merged upstream, but the merge on the remote uses different commit SHAs than the local branch (e.g. squash/rebase-on-merge or post-review amends).
3. Amend/rebase that local branch so its head advances, then unapply or remove it.
4. The workspace merge commit retains the branch's **previous** head as a parent, which now matches no stack head.
5. Open the GUI (see "Unnamed segment") and/or run `but status` (see the `IdMap` error). Try **Update workspace** -> integration fails.

### Expected behavior

- Unapplying/removing a branch should rebuild the workspace merge commit so it never retains a parent that no stack head claims.
- A workspace parent that maps to no stack should be handled gracefully (e.g. dropped or surfaced as a recoverable warning) rather than throwing `Could not find branch CLI id '' in IdMap` and wedging `status` + every mutation.
- There should be an in-product recovery path (e.g. `but workspace rebuild`/`but clean` that re-derives the merge commit from the recorded stack heads) instead of requiring hand-edits to `virtual_branches.toml` / `but.sqlite`.

### Relevant log output

```
$ but status
╭┄zz [unassigned changes] (no changes)
┊
┊╭┄lo [local-tmp-branch]
┊●   46ebe1c4f8 local
├╯
┊
Error: Could not find branch CLI id '' in IdMap

$ but unapply henry.bui/feat/<branch>
Failed to unapply branch. Branch 'henry.bui/feat/<branch>' not found in any applied stack

$ but pull
Found 1 upstream commits on origin/develop
Updating 4 active branches...
Error during integration: merge conflict when computing workspace tree
Error: merge conflict when computing workspace tree
```

ameyypawar:
Fixes #14497.

## Bug
When the `gitbutler/workspace` merge commit has a parent that no stack head claims — an **anonymous segment** — `but status` aborts with `Could not find branch CLI id '' in IdMap`. Because the same enumeration backs every workspace mutation, the whole project is wedged with no in-product recovery. A minimal reproduction is in the issue thread: `git branch -d <applied-branch>` deletes the ref, and the next `but status` throws.

## Cause
An anonymous segment has no name, so `populate_branch_short_ids` assigns it no CLI id and it isn't in the `IdMap`. `but status` looked its empty id up in the `IdMap`, found nothing, and the error propagated out — aborting `status` and every mutation that enumerates the workspace first.

## Fix
Tolerate the anonymous segment at render:
- `StatusOutputLineData::Branch.cli_id` becomes `Option<Arc<CliId>>` (`None` = anonymous); the status enumeration detects a nameless segment via `branch_name().is_none()` and skips the `IdMap` lookup.
- It renders as `[unnamed segment]` with a hint to recover by naming its tip (`git branch <name> <commit>`) — the recovery described in the issue thread.
- Every interactive TUI consumer (commit/move/stack modes, cursor selection) treats a `None` id as a non-target — you can't commit to, move, or stack an unnamed segment.

## Tests
Adds a regression test that deletes an applied branch's ref, asserts `status` renders the unnamed segment rather than erroring, and that re-creating the branch re-binds it. It also smoke-tests `--format json`, which builds ids from `short_id` directly and so never threw.

`cargo test -p but` (468 tests) and clippy are clean.

## Scope
This fixes the `status` wedge (the anonymous-segment `IdMap` throw). Preventing the orphan parent from being retained in the first place — rebuilding the workspace merge commit on unapply, as the issue also suggests — is a separate, complementary change in `but-workspace`, out of scope here.

### DIFF
--- a/crates/but/src/command/legacy/status/mod.rs
+++ b/crates/but/src/command/legacy/status/mod.rs
@@ -1233,33 +1233,53 @@ fn print_group(
                 .unwrap_or_default();
 
             let branch = segment.branch_name().unwrap_or(BStr::new("")).to_string();
-            let branch_cli_id = lookup_cli_id_for_short_id(
-                &status_ctx.id_map,
-                &repo,
-                &segment.short_id,
-                |id| matches!(id, CliId::Branch { .. }),
-                "branch",
-            )?;
+            // The branch CliId is its name, so a nameless segment gets no short
+            // id assigned (see `populate_branch_short_ids`) and isn't in the
+            // IdMap. Such an anonymous segment - e.g. a `gitbutler/workspace`
+            // merge-commit parent whose branch was deleted or unapplied - must be
+            // tolerated here instead of throwing "Could not find branch CLI id ''
+            // in IdMap", which used to wedge `status` and every workspace
+            // mutation that enumerates it first (#14497).
+            let is_anonymous = segment.branch_name().is_none();
+            let branch_cli_id = if is_anonymous {
+                None
+            } else {
+                Some(lookup_cli_id_for_short_id(
+                    &status_ctx.id_map,
+                    &repo,
+                    &segment.short_id,
+                    |id| matches!(id, CliId::Branch { .. }),
+                    "branch",
+                )?)
+            };
             let mut branch_suffix = Vec::new();
             branch_suffix.extend(ci_spans);
             if let Some(branch_status) = branch_status {
                 branch_suffix.push(branch_status);
             }
             branch_suffix.extend(review_spans);
+            if is_anonymous {
+                branch_suffix.push(Span::styled(
+                    " (unnamed - run `git branch <name> <commit>` to recover)",
+                    t.hint,
+                ));
+            }
             if !no_commits.is_empty() {
                 branch_suffix.push(Span::raw(" "));
                 branch_suffix.push(Span::styled(no_commits, t.hint));
             }
 
+            let branch_name = if is_anonymous {
+                Vec::from([Span::styled("unnamed segment", t.hint)])
+            } else {
+                Vec::from([Span::styled(branch, t.local_branch), Span::raw(workspace)])
+            };
             output.branch(
                 Vec::from([Span::raw(format!("┊{notch}┄"))]),
                 BranchLineContent {
                     id: Vec::from([Span::styled(segment.short_id.clone(), t.cli_id)]),
                     decoration_start: Vec::from([Span::raw(" [")]),
-                    branch_name: Vec::from([
-                        Span::styled(branch, t.local_branch),
-                        Span::raw(workspace),
-                    ]),
+                    branch_name,
                     decoration_end: Vec::from([Span::raw("]")]),
                     suffix: branch_suffix,
                 },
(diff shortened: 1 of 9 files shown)


Byron:
Thanks for giving this a shot!

For now I feel uncomfortable enough about this symptomatic to close this PR without further action, also given its linting errors, in favor of continuing the discussion in the related issue. I also invite @davidpdrsn to go ahead with this PR if it's deemed a viable fix in principle after all.

davidpdrsn:
I agree with @Byron that we probably can't merge this.

Anonymous segments is something that we're discussing internally but there are a bunch of API and UX things to figure out. So until we have more clarity around that I wouldn't wanna merge this.

PR #14675, opened 2026-07-08
branch rename: Rename uses the new API

estib-vega:
- Introduce branch_rename.
- This correctly re-checks out the HEAD if in single branch mode
- This correctly updates the metadata


crates/but-core/src/branch/safe_delete.rs
```diff
@@ -28,6 +28,15 @@ impl SafeDelete {
...
 
 impl SafeDelete {
+    /// Return the worktree directories whose `HEAD` points to `rn` (directly or indirectly),
+    /// or `None` if no worktree does. Unlike [`delete_reference()`](Self::delete_reference) this
+    /// performs no mutation, so it can be used to check the precondition up front.
+    pub fn worktree_dirs_with_ref(&self, rn: &gix::Reference) -> Option<&[PathBuf]> {
+        self.worktrees_by_ref
+            .get(&rn.inner.name)
+            .map(|paths| paths.as_slice())
+    }
```
estib-vega, line 38: This is the only thing that falls under the code owners authority

Issue #14260, opened 2026-06-17
`UpdateWorkspaceBase` leaves workspace in unrecoverable state when rebase fails with EMFILE ("too many open files")


endercrest:
### Version

0.20.0

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Summary

When updating the workspace base (trunk) on a large repo with many lanes, the rebase can fail partway with **"too many open files" (EMFILE)** because the app inherits macOS's default soft file-descriptor limit of **256**. The failed operation is **not rolled back atomically** — it leaves GitButler's database and git refs in an inconsistent, half-committed state that **cannot be recovered from within the UI**. The workspace shows an **"Unnamed segment" that cannot be unapplied or interacted with**.

This has happened twice on the same repo. Recovery required manual `but.sqlite`/ref surgery.

### Actual behavior

The operation half-commits. Observed inconsistent state:

- **Database (`but.sqlite`)**: `default_target_sha` advanced to the new trunk, but **all stacks set to `in_workspace = 0`** (0 applied).
- **Git refs**: `refs/heads/gitbutler/workspace` still pointed at a workspace commit that **merged the rebased lane heads** (13 parents), and `refs/heads/gitbutler/target` was **left at the old base** — i.e. the workspace ref and the DB's `default_target_sha` disagreed.
- **UI**: the merged-but-untracked commits rendered as an **"Unnamed segment"** that could not be unapplied or otherwise acted on. All other lanes appeared unapplied.

### Two underlying issues

1. **Non-atomic failure handling (primary).** A rebase failure during `UpdateWorkspaceBase` should be transactional — either fully applied or fully rolled back. Currently a mid-rebase error leaves DB ↔ refs desynced and the UI stuck.
2. **App-imposed fd limit (contributing).** The macOS GUI default of 256 open files is too low for rebasing a large repo with many lanes.

### Workaround

- Launch the binary with a raised limit: `ulimit -n 65536 && /Applications/GitButler.app/Contents/MacOS/gitbutler-tauri`, then retry — the rebase completes.
- To recover an already-broken state: undo the failed operation via the oplog snapshot, or manually realign `gitbutler/target` + `gitbutler/workspace` refs to the DB's `default_target_sha`.

### How to reproduce (Optional)

1. On a large monorepo with many applied lanes, run **Update Workspace Base** (update trunk).
[...]

PR #14687, opened 2026-07-09
fix(but-api): raise the open-file limit at startup to avoid EMFILE

ameyypawar:
Addresses the EMFILE trigger in #14260.

## Bug
On a large repo with many stacks, `UpdateWorkspaceBase` (updating trunk) can fail partway with **EMFILE ("too many open files")**: the app inherits macOS's default `RLIMIT_NOFILE` soft limit of **256**, and each `gix`/`git2` repository memory-maps a potentially large number of pack files on first object access (with several repositories open at once). When the rebase fails mid-operation, the workspace is left in an inconsistent, hard-to-recover state.

## Fix
Raise the soft open-file limit toward the hard limit once at startup, from the desktop app, the CLI, and the local API server. The value is capped at 65536 — the `ulimit -n 65536` that resolves the issue in practice. It's best-effort (never lowers the limit, treats a failure as non-fatal) and a no-op on Windows (which has no comparably low default).

- `but_api::fd_limit::raise_soft_limit()` wraps `rlimit::increase_nofile_limit`, which handles the macOS quirk where `rlim_max` is `RLIM_INFINITY` but the real ceiling is `kern.maxfilesperproc`.
- Called at the top of the `gitbutler-tauri`, `but`, and `but-server` mains — before anything opens a repository.

## Scope
This removes the EMFILE **trigger**. The separate, primary issue in #14260 — non-atomic failure handling (a rebase failure of any cause leaving the database and refs desynced) — is out of scope here, so this intentionally doesn't close the issue.

## Verification
- `cargo doc` (`-D warnings`), `cargo clippy`, and `cargo test -p but-api` are clean; `but`, `gitbutler-tauri`, and `but-server` compile.
- The raise is exercised under a lowered limit: from `ulimit -n 400`, the helper raised the soft limit to 61440 (the macOS `kern.maxfilesperproc` ceiling — well above the 256 that triggers the crash).
- A full EMFILE-cascade repro isn't practical (it needs a very large many-pack repository), so this prevents the trigger rather than reproducing the end-to-end failure.

Note: `but-napi` (the Electron/node host) has no Rust entrypoint to hook, so raising its limit would need to be done by the node host.

### DIFF
--- a/crates/but-api/Cargo.toml
+++ b/crates/but-api/Cargo.toml
@@ -165,5 +165,9 @@ snapbox.workspace = true
 tempfile.workspace = true
 tokio = { workspace = true, features = ["full"] }
 
+[target.'cfg(unix)'.dev-dependencies]
+# Used by tests/fd_limit.rs to read RLIMIT_NOFILE and assert the never-lowers invariant.
+rlimit = "0.10.2"
+
 [lints]
 workspace = true
--- a/crates/but-api/Cargo.toml
+++ b/crates/but-api/Cargo.toml
@@ -153,6 +153,10 @@ but-schemars.workspace = true
 objc2-app-kit = { version = "0.3.2", default-features = false, features = ["std", "NSWorkspace"] }
 objc2-foundation = { version = "0.3.2", default-features = false, features = ["std", "NSString", "NSURL"] }
 
+[target.'cfg(unix)'.dependencies]
+# Raise the open-file soft limit at startup so large rebases don't hit EMFILE (#14260).
+rlimit = "0.10.2"
+
 [dev-dependencies]
 but-api = { path = ".", features = ["legacy"] }
 but-testsupport.workspace = true
--- a/crates/but-api/src/fd_limit.rs
+++ b/crates/but-api/src/fd_limit.rs
@@ -28,38 +28,3 @@ pub fn raise_soft_limit() {
         Err(err) => tracing::warn!("could not raise the open-file soft limit: {err}"),
     }
 }
-
-#[cfg(all(test, unix))]
-mod tests {
-    use rlimit::Resource;
-
-    // We assert only the safety invariant (never lowers), not that the target
-    // is reached: the OS-permitted maximum is platform-specific and, on macOS,
-    // is `kern.maxfilesperproc` - which is well below `RLIM_INFINITY`, so the
-    // achieved soft limit can legitimately be below `TARGET_NOFILE` even when
-    // the reported hard limit is unlimited.
-    #[test]
-    fn raise_soft_limit_runs_and_never_lowers() {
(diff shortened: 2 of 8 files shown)


crates/but-api/src/fd_limit.rs
```diff
@@ -0,0 +1,65 @@
...
+    // is reached: the OS-permitted maximum is platform-specific and, on macOS,
+    // is `kern.maxfilesperproc` - which is well below `RLIM_INFINITY`, so the
+    // achieved soft limit can legitimately be below `TARGET_NOFILE` even when
+    // the reported hard limit is unlimited.
+    #[test]
+    fn raise_soft_limit_runs_and_never_lowers() {
+        let (soft_before, _hard) = Resource::NOFILE.get().expect("read NOFILE limit");
+
+        // The public helper is best-effort and must not panic.
+        super::raise_soft_limit();
```
Byron: This test should be on integration level, as it exercises a public API.

crates/but/src/main.rs
```diff
@@ -1,5 +1,6 @@
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
+    but_api::fd_limit::raise_soft_limit();
```
Byron: The CLI should probably not do this unconditionally, or at all.

Byron (changes requested):

While I think that this PR is a good approach, let's have a second pair of eyes and check with @slarse to see what he thinks particularly about adding this to the CLI. My feeling is that there it should rely on what users have configured, and `but` should respect this.

@codex review

Byron (approved):
Thanks for addressing the earlier review. This looks good to me now, but I definitely think @slarse should have the final word on this.

crates/but-api/Cargo.toml
```diff
@@ -161,5 +165,9 @@ snapbox.workspace = true
 tempfile.workspace = true
 tokio = { workspace = true, features = ["full"] }
 
+[target.'cfg(unix)'.dev-dependencies]
+# Used by tests/fd_limit.rs to read RLIMIT_NOFILE and assert the never-lowers invariant.
+rlimit = "0.10.2"
```
slarse, line 170: Redundant, already a production dependency


```suggestion
```

slarse (approved):
@Byron I don't really see an issue with raising the soft limit for CLI use if we know this is a failure case for GitButler (and we can't do anything about the large amount of open files?). It's a soft limit for a reason, raising it isn't something I would personally view as bad practice unless it's used as a bandaid for leaking file descriptors.

A user of the CLI won't know to raise it until they've already ended up in a bad state so might be good to raise it for them.

Also, what about Lite? It would suffer from the same problem?

The thing that does strike me as a bit unwarranted in this PR is bringing in a dependency for raising the limit, it shouldn't be more than a handful lines of code even including all of the awkward C API error handling.

FYI commits are missing signatures :)

PR #12980, opened 2026-03-22
feat: hook ownership model with external manager detection

ichoosetoaccept:
Closes #12748

## Summary

Adds a hook ownership model so GitButler can coexist with external hook managers (now: prek, later: pre-commit, lefthook, husky, etc.) instead of unconditionally overwriting their hooks.

This PR ships the **plumbing + `but` CLI** milestone described in https://github.com/gitbutlerapp/gitbutler/issues/12748#issuecomment-2746847817. It does **not** include `but doctor`, Desktop UI integration, or onboarding changes — those are separate follow-ups.

### New `but-hooks` crate

All hook detection and management logic moved into a dedicated `but-hooks` crate, following the `gitbutler-` → `but-` migration. Existing consumers in `gitbutler-repo` continue working via re-exports.

- **Hook manager detection** — content-based signature matching (inspects hook file contents) with a config+binary fallback for managers that are configured but haven't installed hooks yet. Currently supports prek; adding new managers requires a single enum variant + match arms for `owns_hook`, `has_config`, `is_available`, and `integration_instructions`.
- **Managed hook installation** — installs GitButler's guard hooks with ownership checks. Non-GitButler hooks are preserved (not overwritten) unless `--force-hooks` is used, in which case the original is backed up to `<hook>-user`.
- **Project-local binary detection** — checks `.venv/bin/` and `node_modules/.bin/` in addition to system PATH, so project-local hook manager installations are detected.

### `but hook` subcommands

Composable CLI entry points for external hook managers to invoke:

- **`but hook pre-commit`** — blocks `git commit` on `gitbutler/workspace` with a helpful error directing the user to `but commit`.
- **`but hook post-checkout`** — prints an informational message when leaving the workspace branch.
- **`but hook pre-push`** — blocks `git push` on `gitbutler/workspace` with a helpful error directing the user to `but push`.
- **`but hook status`** — diagnostic command showing hook mode (managed/external/disabled/unconfigured), per-hook ownership, orphaned hook detection, and context-aware recommendations. Supports human, shell, and JSON output.

### Updated lifecycle commands

- **`but setup`** — detects external hook managers before installing. If detected: persists opt-out config, cleans up any previously-installed GitButler hooks, and prints integration instructions. `--no-hooks` / `--force-hooks` flags provide manual overrides.
- **`but teardown`** — respects hook ownership during cleanup. Externally-managed hooks are left untouched.

### Desktop backend (gitbutler-branch-actions)

[...]

PR #14698, opened 2026-07-11
feat: register managed hooks via git config on git 2.54+

### LINKED ISSUE #12748: `but setup` displaces hooks owned by hook managers (prek, pre-commit, husky) without detecting them, and provides no escape hatch
### Version

0.19.5

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

## Summary

`but setup` unconditionally overwrites `pre-commit` and `post-checkout` hooks without checking whether a hook manager (prek, pre-commit, husky, lefthook, etc.) already owns them. It renames existing hooks with a `-user` suffix and installs `GITBUTLER_MANAGED_HOOK_V1` wrappers in their place. While `but setup` does print a note that it is "Installing Git hooks", it does not name the affected files, does not warn that existing hooks will be displaced, and does not detect pre-existing hook managers.

The resulting state is fragile in several distinct ways.

### 1. Insufficient warning on `but setup`

`but setup` does print a note:

> Installing Git hooks to help manage commits on the workspace branch

However it does not:
- Name which hook files will be affected (`pre-commit`, `post-checkout`)
- Mention that **existing hooks will be renamed** to a `-user` suffix
- Detect whether a hook manager (prek, pre-commit, husky) is already installed and owns those files
- Offer a way to skip hook installation (`--no-hooks`) for projects that manage hooks independently

A developer running `but setup` in a project where prek already owns `pre-commit` gets no indication that their hook runner has just been demoted to `pre-commit-user` and wrapped by a GitButler shim.

### 2. Inconsistent coverage — `pre-push` is left untouched

GitButler wraps `pre-commit` and `post-checkout`, but ignores `pre-push` entirely. This means pre-push hooks continue running through the hook manager's full pipeline (including stash-before-hooks). When combined with a stale `gitbutler/workspace` HEAD commit, this causes linters to run against stale file content — see the stale-workspace-HEAD issue filed separately (#12750).

### 3. `prek install` after `but setup` silently breaks workspace guard

Running `prek install` a second time (e.g., after `git clone` on a new machine following the project's README) produces:

```
Hook already exists at `.git/hooks/pre-commit`, moved it to `.git/hooks/pre-commit.legacy`
Migration mode: prek will also run legacy hook `.git/hooks/pre-commit.legacy`
prek installed at `.git/hooks/pre-commit`
```

Result:
```
pre-commit         → fresh prek runner
pre-commit.legacy  → GitButler wrapper  ← prek calls this in migration mode, which calls...
pre-commit-user    → old prek runner    ← ...this orphaned duplicate: prek runs TWICE
post-checkout      → fresh prek runner  ← GitButler cleanup logic GONE
post-checkout-user → prek runner (orphaned)
```

In prek's default migration mode, the GitButler workspace guard still runs (via `.legacy`), but prek is invoked twice — once directly, once via the wrapper calling `pre-commit-user`. Running `prek install --overwrite` removes `.legacy` entirely, at which point the workspace guard is silently gone with no error or warning.

### 4. No escape hatch

There is no `but setup --no-hooks` flag, no config option to disable hook management, and no way to tell GitButler "I already manage hooks through prek — provide your logic as a hook I can call."

## Environment

- GitButler CLI: `but` (latest)
- OS: macOS 15 / macOS 26 beta
- Hook manager tested: [prek](https://github.com/j178/prek) v0.3+ (pre-commit compatible)
- Also affects: pre-commit (uses `.git/hooks/`), Husky (both `.git/hooks/` for v4 and `.husky/` for v5+ — same displacement pattern regardless of `core.hooksPath`)

### With prek (hooks in `.git/hooks/`)

```sh
# Fresh repo with prek already managing hooks
git init /tmp/hook-test
git commit --allow-empty -m "init"
cat > /tmp/hook-test/prek.toml << 'EOF'
minimum_prek_version = "0.3.3"
default_install_hook_types = ["pre-commit", "pre-push", "post-checkout"]
[[repos]]
repo = "builtin"
hooks = [{ id = "trailing-whitespace" }]
EOF
prek install -C /tmp/hook-test
```

Hooks before `but setup`:
```
.git/hooks/pre-commit    → prek runner
.git/hooks/pre-push      → prek runner
.git/hooks/post-checkout → prek runner
```

```sh
but setup   # run in /tmp/hook-test
```

Hooks after `but setup`:
```
.git/hooks/pre-commit         → GITBUTLER_MANAGED_HOOK_V1 wrapper  ← replaced, no file-level warning
.git/hooks/pre-commit-user    → prek runner (demoted, non-standard filename)
.git/hooks/pre-push           → prek runner                         ← untouched (inconsistent!)
.git/hooks/post-checkout      → GITBUTLER_MANAGED_HOOK_V1 wrapper  ← replaced, no file-level warning
.git/hooks/post-checkout-user → prek runner (demoted, non-standard filename)
```

### With Husky (hooks in `.husky/`, `core.hooksPath = .husky`)

`but setup` is aware of `core.hooksPath` — when set to `.husky`, it correctly installs its wrappers into `.husky/` rather than `.git/hooks/`. However, the same displacement pattern applies:

```sh
# Fresh repo with Husky-style hooks
git init /tmp/husky-test && git commit --allow-empty -m "init"
mkdir /tmp/husky-test/.husky
echo '#!/bin/sh\necho "husky: running pre-commit"' > /tmp/husky-test/.husky/pre-commit
chmod +x /tmp/husky-test/.husky/pre-commit
git -C /tmp/husky-test config core.hooksPath .husky
but setup   # run in /tmp/husky-test
```

Result — `.husky/` after `but setup`:
```
.husky/pre-commit      → GITBUTLER_MANAGED_HOOK_V1 wrapper  ← original displaced
.husky/pre-commit-user → original Husky hook (50 bytes)
.husky/post-checkout   → GitButler post-checkout handler    ← injected
.git/hooks/            → empty (correctly unaffected)
```

The GitButler GUI setting "Enable Husky hooks" is orthogonal to this — it controls whether GitButler's **internal** hook execution calls `.husky/` scripts, not how `but setup` behaves.

## Expected behaviour

`but setup` should:

1. **Detect existing hook managers** before overwriting. If `pre-commit`, `post-checkout`, or `pre-push` are already owned by a recognised hook manager (prek, pre-commit, husky, etc.), print a warning and instructions rather than silently renaming.

2. **Provide composable hook scripts** — ship the workspace guard and post-checkout cleanup as standalone scripts (e.g., installed to `.git/gitbutler/hooks/`) that can be called from any hook manager config.

3. **Support `--no-hooks`** (or a config flag) to skip all hook installation for teams that manage hooks independently.

4. **Handle `pre-push` consistently** — either wrap it the same way as `pre-commit`/`post-checkout`, or don't wrap any hooks at all and rely on the composable scripts approach.

## Suggested integration pattern

The correct integration for a prek-managed project looks like this (currently requires manual setup with no official guidance):

```toml
# prek.toml
[[repos]]
repo = "local"
hooks = [
    { id = "gitbutler-workspace-guard",
      name = "GitButler Workspace Guard",
      language = "system",
      entry = "nu .git/gitbutler/hooks/workspace-guard.nu",   # hypothetical
      pass_filenames = false,
      always_run = true,
      stages = ["pre-commit"] },
]
```

GitButler should ship these scripts and document this pattern as the recommended integration for projects using hook managers.

```shell

```

ichoosetoaccept:
## 🧢 Changes

On git ≥ 2.54, GitButler registers its workspace-guard hooks through [config-based hooks](https://git-scm.com/docs/githooks#_configuration_based_hooks) instead of writing into `.git/hooks/`:

- Guard scripts live in `<git-common-dir>/gitbutler/hooks/`, registered via `hook.<name>.event` / `hook.<name>.command` in local config.
- `.git/hooks/` is never touched, so prek / pre-commit / husky / lefthook keep working unchanged — git runs config hooks first, then the hooks-directory hook.
- Installing on 2.54+ also cleans up hooks that older GitButler versions wrote into `.git/hooks/` (signature-checked) and restores any displaced user hook from its backup. Installation re-runs on every workspace update, so after a git upgrade this migration happens on the user's next GitButler operation — no manual step. Until then the old file hooks simply keep working.
- `git config gitbutler.installHooks false` opts out on any git version, and `but setup` reports the skip in both its human and JSON output.
- `but setup` and `but teardown` report hooks they couldn't install or remove, in both the human and JSON output, instead of silently claiming success.

For git < 2.54 nothing changes by default: the file-based installation stays exactly as it is today (hooks run under whatever `git` the user invokes, so support is detected via `git --version`, not the library GitButler links). The only version-independent additions are the opt-out and teardown's honest reporting. Checking out away from `gitbutler/workspace` still cleans everything up (`git switch` and `git checkout` both fire the post-checkout hook — verified with both).

About two-thirds of the added lines are tests. They run against real repositories, including end-to-end with a real git ≥ 2.54: an actual `git commit` on the workspace branch is blocked, an actual `git checkout` away removes the config entries and scripts (self-skipping on older CI git).

What I validated personally, with a release build of this branch on macOS:

- [x] Disposable repo with a local bare remote and a `pre-commit` hook installed by an actual `prek install`: `but setup`, inspected the config entries and both hook directories, `git commit` blocked with `GITBUTLER_ERROR` (prek's own hook still ran and passed alongside it), `git checkout main` removed the config entries and scripts — then the same staged change committed successfully through prek. The prek hook's checksum was identical before, during, and after.
- [x] Second repo, exiting via `but branch new` + `but teardown` instead of the checkout: same clean end state, same untouched checksum. (One caveat: `but teardown` straight after `but setup`, with no branch created, stops at "Finding active branch to check out" before reaching the hooks — pre-existing behavior, `--checkout-to` gets past it and hook cleanup is then complete.)
- [x] First validated on git 2.54.0; after Homebrew moved this machine to 2.55.0 I re-ran both scenarios and the automated suite against it — everything still passes
- [x] The pre-2.54 fallback and the upgrade migration, in a Debian container (reproduction script in a comment below): with git 2.53.0 `but setup` installs today's file-based hooks — user hook displaced to `pre-commit-user` but still chained and running, commit blocked; after switching to git 2.55.0 the next install registers the config hooks, sweeps the signed file hooks, and restores the user hook byte-identical; checkout away then removes everything. 18 scripted checks, all green.

The core of the manual run:

```console
$ but setup
$ git config --local --get-regexp '^hook\.'
hook.gitbutler-pre-commit.event pre-commit
hook.gitbutler-pre-commit.command '.../.git/gitbutler/hooks/pre-commit'
...
$ git commit -m test
GITBUTLER_ERROR: Cannot commit directly to gitbutler/workspace branch.
$ git checkout main
NOTE: You have left GitButler's managed workspace branch.
$ git config --local --get-regexp '^hook\.'   # exits 1 — nothing left
```

## ☕️ Reasoning

This is the surgical alternative discussed in #12748 after @Alxandr pointed out git 2.54's config-based hooks. The ownership-model PR (#12980) made *sharing* `.git/hooks/` slots safe; config-based hooks make the slots uncontested by construction, so nearly all of that machinery becomes unnecessary — "nearly", because a small signature-checked remnant survives as the migration sweep that removes hooks older GitButler versions wrote and restores displaced user hooks. None of the rest has to be merged only to be removed as git ≥ 2.54 spreads.

The biggest structural win: the old approach needed adapter code per hook manager — a detection signature, config matching, and integration instructions for prek first, then again for pre-commit, husky, lefthook, hk, and whatever comes next. Here that whole axis disappears. GitButler never inspects or even notices the other tool, so every hook manager — including ones that don't exist yet — is supported with zero code. This honestly is my favorite part of this new PR compared to the old one. A close second favorite is that this PR contains drastically less code and attempts to contribute something that should hold for a long time, very much unlike the first attempt which would have added maintenance burden to the project, no matter how well-meaning.

Git 2.54 shipped in April 2026, so package-manager gits (Homebrew, winget, Arch, Fedora…) are already past it — my machine is on 2.55 — while slower channels (Apple's Command Line Tools, Debian/Ubuntu LTS) will lag for a while. That's fine: users below 2.54 keep today's file-based behavior and lose nothing, and the share on the config-based path only grows from here.

Two review notes:

- This landed bigger than the couple-hundred lines you mused about — a good part of the implementation is the hook scripts themselves. It was bigger still: the CLI exit-code plumbing that came with the reporting is no longer part of this PR.
- The new integration tests keep the `test_` prefix you asked to avoid on #12980: they extend the existing `managed_hooks_tests.rs`, where all 21 existing tests use it, and matching the file seemed better than mixing conventions or renaming untouched tests. Glad to rename if you prefer.

## 🎫 Affected issues

Fixes: #12748

### DIFF
--- a/crates/but/src/command/legacy/setup.rs
+++ b/crates/but/src/command/legacy/setup.rs
@@ -21,6 +21,14 @@ struct SetupResult {
     project_status: ProjectStatus,
     /// The target branch configuration
     target: Option<TargetInfo>,
+    /// Hook installation problems; when non-empty, some GitButler hooks are not
+    /// installed and direct commits to the workspace branch may go unblocked.
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    hook_warnings: Vec<String>,
+    /// Hook installation was skipped via `gitbutler.installHooks=false`, so
+    /// commits to the workspace branch are not blocked.
+    #[serde(skip_serializing_if = "std::ops::Not::not")]
+    hooks_skipped: bool,
 }
 
 #[derive(Debug, Serialize)]
@@ -424,24 +432,53 @@ pub(crate) fn repo(
         but_api::legacy::virtual_branches::switch_back_to_workspace_with_perm(ctx, perm)?;
     }
 
-    // Install managed hooks to prevent accidental git commits
-    if let Ok(repo) = ctx.repo.get()
-        && let Err(e) = gitbutler_repo::managed_hooks::install_managed_hooks(&repo)
-        && let Some(out) = out.for_human()
-    {
-        writeln!(
-            out,
-            "  {}",
-            t.attention.paint(format!(
-                "Warning: Failed to install GitButler managed hooks: {e}"
-            ))
-        )?;
+    // Install managed hooks to prevent accidental git commits. Partial
+    // successes matter as much as hard errors here: they mean some hook could
+    // not be installed or cleaned up, so tell the user instead of dropping
+    // the warnings.
+    let mut hooks_skipped = false;
+    let mut hook_warnings = Vec::new();
+    if let Ok(repo) = ctx.repo.get() {
+        hook_warnings = match gitbutler_repo::managed_hooks::install_managed_hooks(&repo) {
+            Ok(gitbutler_repo::managed_hooks::HookInstallationResult::PartialSuccess {
+                warnings,
+            }) => warnings,
+            Ok(gitbutler_repo::managed_hooks::HookInstallationResult::SkippedByConfig) => {
+                hooks_skipped = true;
+                Vec::new()
+            }
+            Ok(_) => Vec::new(),
+            Err(e) => vec![format!("Failed to install GitButler managed hooks: {e}")],
+        };
+        if let Some(out) = out.for_human() {
+            for problem in &hook_warnings {
+                writeln!(
+                    out,
+                    "  {}",
+                    t.attention.paint(format!("Warning: {problem}"))
+                )?;
+            }
+            if hooks_skipped {
+                writeln!(
+                    out,
+                    "  {}",
+                    t.attention.paint(
+                        "Note: Skipped Git hook installation (gitbutler.installHooks=false). Commits made directly to the workspace branch will not be blocked."
+                    )
+                )?;
+            }
+        }
     }
 
     // if we switched - tell the user what this is all about
     if pre_head_name != "gitbutler/workspace"
         && let Some(out) = out.for_human()
     {
+        let hooks_note = if hooks_skipped {
+            "- Skipping Git hooks (disabled via gitbutler.installHooks)"
+        } else {
+            "- Installing Git hooks to help manage commits on the workspace branch"
+        };
         writeln!(
             out,
             "{}",
@@ -450,7 +487,7 @@ pub(crate) fn repo(
 Setting up your project for GitButler tooling. Some things to note:
 
 - Switching you to a special `gitbutler/workspace` branch to enable parallel branches
-- Installing Git hooks to help manage commits on the workspace branch
+{hooks_note}
 
 To undo these changes and return to normal Git mode, either:
 
@@ -474,6 +511,8 @@ More info: https://docs.gitbutler.com/workspace-branch
             repository_path: path::absolute(repo_path)?.display().to_string(),
             project_status,
             target: target_info,
+            hook_warnings,
+            hooks_skipped,
         };
         json_out.write_value(&result)?;
     }
--- a/crates/but/src/command/legacy/teardown.rs
+++ b/crates/but/src/command/legacy/teardown.rs
@@ -15,6 +15,10 @@ use crate::{
 struct TeardownResult {
     snapshot_id: String,
     checked_out_branch: String,
+    /// Hook cleanup problems; when non-empty, some GitButler hooks are still
+    /// installed and may keep affecting git operations.
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    hook_warnings: Vec<String>,
 }
 
 pub(crate) fn teardown(
@@ -156,17 +160,27 @@ pub(crate) fn teardown(
         writeln!(out)?;
     }
 
-    // Uninstall managed hooks before checking out
-    if let Ok(repo) = ctx.repo.get()
-        && let Err(e) = gitbutler_repo::managed_hooks::uninstall_managed_hooks(&repo)
-        && let Some(out) = out.for_human()
-    {
-        writeln!(
-            out,
-            "  {}",
-            t.attention
-                .paint(format!("Warning: Failed to uninstall Git hooks: {e}"))
-        )?;
+    // Uninstall managed hooks before checking out. `uninstall_managed_hooks`
+    // reports removal failures as PartialSuccess warnings rather than Err, so
+    // surface those too -- they mean GitButler hooks are still active.
+    let mut hook_warnings = Vec::new();
+    if let Ok(repo) = ctx.repo.get() {
+        hook_warnings = match gitbutler_repo::managed_hooks::uninstall_managed_hooks(&repo) {
+            Ok(gitbutler_repo::managed_hooks::HookInstallationResult::PartialSuccess {
+                warnings,
+            }) => warnings,
+            Ok(_) => Vec::new(),
+            Err(e) => vec![format!("Failed to uninstall Git hooks: {e}")],
+        };
+        if let Some(out) = out.for_human() {
+            for problem in &hook_warnings {
+                writeln!(
+                    out,
+                    "  {}",
+                    t.attention.paint(format!("Warning: {problem}"))
+                )?;
+            }
+        }
     }
 
     // Check out the target branch using Git directly
@@ -228,11 +242,21 @@ pub(crate) fn teardown(
 
     // Final success message
     if let Some(out) = out.for_human() {
-        writeln!(
-            out,
-            "{}",
-            t.success.paint("✓ Successfully exited GitButler mode!")
-        )?;
+        if hook_warnings.is_empty() {
+            writeln!(
+                out,
+                "{}",
+                t.success.paint("✓ Successfully exited GitButler mode!")
+            )?;
+        } else {
+            writeln!(
+                out,
+                "{}",
+                t.attention.paint(
+                    "⚠ Exited GitButler mode, but some GitButler hooks could not be removed (see warnings above)."
+                )
+            )?;
+        }
         writeln!(out)?;
         writeln!(
             out,
@@ -251,6 +275,7 @@ pub(crate) fn teardown(
         out.write_value(&TeardownResult {
             snapshot_id,
             checked_out_branch: target_branch_name,
+            hook_warnings: hook_warnings.clone(),
         })?;
     }
 
--- a/crates/but/tests/but/command/setup.rs
+++ b/crates/but/tests/but/command/setup.rs
@@ -730,3 +730,119 @@ Learn more at https://docs.gitbutler.com/cli-overview
 
 "#]]);
 }
+
+/// When the user opted out of hooks via `gitbutler.installHooks=false`, setup
+/// must say hook installation was skipped instead of claiming hooks are being
+/// installed while direct commits to the workspace branch go unguarded.
+#[test]
+fn setup_with_hooks_opted_out_says_so() {
+    let env = Sandbox::open_with_default_settings("repo-no-remote");
+    env.invoke_git("config --local gitbutler.installHooks false");
+
+    env.but("setup")
+        .assert()
+        .success()
+        .stderr_eq(snapbox::str![])
+        .stdout_eq(snapbox::str![[r#"
+Setting up GitButler project...
+
+→ Adding repository to GitButler project registry
+  ✓ Repository already in project registry
+
+→ Configuring default target branch
+  No push remote found, creating gb-local remote...
+  ✓ Created gb-local remote tracking main
+  ✓ Set default target to: gb-local/main
+
+GitButler project setup complete!
+Target branch: gb-local/main
+Remote: gb-local
+
+  Note: Skipped Git hook installation (gitbutler.installHooks=false). Commits made directly to the workspace branch will not be blocked.
+
+Setting up your project for GitButler tooling. Some things to note:
+
+- Switching you to a special `gitbutler/workspace` branch to enable parallel branches
+- Skipping Git hooks (disabled via gitbutler.installHooks)
+
+To undo these changes and return to normal Git mode, either:
+
+    - Directly checkout a branch (`git checkout main`)
+    - Run `but teardown`
+
+More info: https://docs.gitbutler.com/workspace-branch
+
+
+
+██▄      ▄██  ▀██▀▀█▄ ▀██▀ ▀██▀ █▀▀██▀▀█
+████▄  ▄████   ██  ██  ██   ██  ▀  ██  ▀
+████████████   ██▀▀█▄  ██   ██     ██
+████▀  ▀████   ██  ██  ██   ██     ██
+██▀      ▀██  ▄██▄▄█▀  ▀█▄▄▄█▀   ▄▄██▄▄
+
+The command-line interface for GitButler ⋈
+
+$ but branch new <name>                       Create a new branch
+$ but status                                  View workspace status
+$ but commit -m <message>                     Commit changes to current branch
+$ but push                                    Push all branches
+$ but teardown                                Return to normal Git mode
+
+Learn more at https://docs.gitbutler.com/cli-overview
+
+
+"#]]);
+}
+
+/// The opt-out note above only reaches humans. A JSON consumer that cannot see
+/// it would proceed as if the workspace were guarded, so the skip has to be
+/// visible in the JSON output too.
+#[test]
+fn json_output_reports_hooks_skipped() {
+    let env = Sandbox::open_with_default_settings("repo-no-remote");
+    env.invoke_git("config --local gitbutler.installHooks false");
+
+    env.but("--json setup")
+        .allow_json()
+        .assert()
+        .success()
+        .stderr_eq(snapbox::str![])
+        .stdout_eq(snapbox::str![[r#"
+{
+  "repositoryPath": "[..]",
+  "projectStatus": "alreadyexists",
+  "target": {
+    "branchName": "gb-local/main",
+    "remoteName": "gb-local",
+    "newlySet": true
+  },
+  "hooksSkipped": true
+}
+
+"#]]);
+}
+
+/// Hooks install cleanly here, so neither new field should appear -- existing
+/// JSON consumers must see exactly the payload they saw before.
+#[test]
+fn json_output_omits_hook_fields_when_install_succeeds() {
+    let env = Sandbox::open_with_default_settings("repo-no-remote");
+
+    env.but("--json setup")
+        .allow_json()
+        .assert()
+        .success()
+        .stderr_eq(snapbox::str![])
+        .stdout_eq(snapbox::str![[r#"
+{
+  "repositoryPath": "[..]",
+  "projectStatus": "alreadyexists",
+  "target": {
+    "branchName": "gb-local/main",
+    "remoteName": "gb-local",
+    "newlySet": true
+  }
+}
+
+"#]]);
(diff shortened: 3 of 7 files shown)


ichoosetoaccept:
To validate the pre-2.54 fallback and the upgrade migration end-to-end, I ran the
script below in a Debian container: it builds git 2.53.0 and 2.55.0 from source
(no package archive carries both), builds `but` from this branch, and walks one
repo through the full journey — file-based hooks under 2.53 (user hook displaced
but chained, commit blocked), the migration to config-based hooks after switching
to 2.55 (signed file hooks swept, user hook restored byte-identical), and the
checkout-away cleanup. 18 checks, all green on my machine.

To reproduce, from a checkout of this branch:

```sh
docker run --rm \
  -v "$PWD":/src:ro \
  -v "$PWD"/validate-migration.sh:/test.sh:ro \
  -v but-build-cache:/build -v but-registry-cache:/usr/local/cargo/registry \
  rust:slim-bookworm bash /test.sh
```

(The two named volumes are optional but make re-runs incremental).

<details>
<summary><b>validate-migration.sh</b> (80 lines)</summary>

```bash
#!/usr/bin/env bash
# Runs INSIDE the container. Validates pre-2.54 file-based behavior and the 2.53→2.55 migration.
set -uo pipefail
say() { printf '\n===== %s =====\n' "$*"; }
ok()  { echo "OK   $*"; }
bad() { echo "FAIL $*"; FAILURES=$((FAILURES+1)); }
FAILURES=0

say "install build deps"
apt-get update -qq >/dev/null
apt-get install -y -qq build-essential curl ca-certificates pkg-config libssl-dev zlib1g-dev cmake perl libdbus-1-dev >/dev/null 2>&1

say "build git 2.53.0 and 2.55.0 from source"
for v in 2.53.0 2.55.0; do
  curl -sL "https://mirrors.edge.kernel.org/pub/software/scm/git/git-$v.tar.gz" | tar xz -C /tmp
  make -C /tmp/git-$v -j"$(nproc)" prefix=/opt/git-${v%.0} NO_TCLTK=1 NO_GETTEXT=1 NO_CURL=1 NO_EXPAT=1 NO_PERL=1 install >/dev/null 2>&1 \
    && ok "built git $v" || bad "building git $v"
done
/opt/git-2.53/bin/git version && /opt/git-2.55/bin/git version

say "build but from the branch (debug)"
export CARGO_TARGET_DIR=/build
export PATH=/opt/git-2.55/bin:$PATH   # build scripts may invoke git
(cd /src && cargo build -p but >/tmp/build.log 2>&1) || { bad "cargo build"; tail -40 /tmp/build.log; }
BUT=/build/debug/but
[ -x "$BUT" ] && ok "but binary built" || { bad "no but binary"; exit 1; }

say "create test repo with a user pre-commit hook"
export E2E_TEST_APP_DATA_DIR=/tmp/appdata NOPAGER=1
mkdir -p /tmp/appdata
export PATH=/opt/git-2.53/bin:$PATH
git init -q --bare /tmp/origin.git
git init -q -b main /tmp/repo
cd /tmp/repo
git config user.email t@e.c && git config user.name t && git config commit.gpgsign false
printf '#!/bin/sh\necho USER-HOOK-RAN\nexit 0\n' > .git/hooks/pre-commit
chmod +x .git/hooks/pre-commit
SHA_BEFORE=$(sha1sum .git/hooks/pre-commit | cut -d' ' -f1)
echo hi > f && git add f && git commit -qm init
git remote add origin /tmp/origin.git && git push -qu origin main 2>/dev/null

say "PHASE 1: git 2.53 — expect FILE-BASED hooks, user hook displaced+chained"
git version
"$BUT" setup >/tmp/setup1.log 2>&1 || { bad "but setup (2.53)"; tail -5 /tmp/setup1.log; }
if git config --local --get-regexp '^hook\.' >/dev/null 2>&1; then bad "config entries present on 2.53"; else ok "no config entries on 2.53"; fi
grep -q GITBUTLER_MANAGED_HOOK .git/hooks/pre-commit && ok "managed file hook installed" || bad "no managed file hook"
[ -f .git/hooks/pre-commit-user ] && ok "user hook displaced to pre-commit-user" || bad "no -user backup"
[ -f .git/hooks/post-checkout ] && grep -q GITBUTLER_MANAGED_HOOK .git/hooks/post-checkout && ok "managed post-checkout installed" || bad "no managed post-checkout"
echo x >> f && git add f
OUT=$(git commit -m nope 2>&1); RC=$?
echo "$OUT" | grep -q USER-HOOK-RAN && ok "displaced user hook still chained+ran" || bad "user hook did not run"
echo "$OUT" | grep -q GITBUTLER_ERROR && [ $RC -ne 0 ] && ok "commit blocked on 2.53" || bad "commit NOT blocked on 2.53 (rc=$RC)"

say "PHASE 2: upgrade to git 2.55 — expect config hooks + migration sweep"
export PATH=/opt/git-2.55/bin:$PATH
git version
"$BUT" setup >/tmp/setup2.log 2>&1 || { bad "but setup (2.55)"; tail -5 /tmp/setup2.log; }
N=$(git config --local --get-regexp '^hook\.' | wc -l)
[ "$N" -eq 4 ] && ok "4 config entries registered" || { bad "expected 4 config entries, got $N"; git config --local --get-regexp '^hook\.'; }
[ -d .git/gitbutler/hooks ] && ok "scripts dir created" || bad "no scripts dir"
SHA_AFTER=$(sha1sum .git/hooks/pre-commit | cut -d' ' -f1)
[ "$SHA_BEFORE" = "$SHA_AFTER" ] && ok "user hook restored byte-identical" || bad "user hook differs after migration"
[ ! -e .git/hooks/pre-commit-user ] && ok "-user backup consumed" || bad "-user backup still present"
[ ! -e .git/hooks/post-checkout ] && ok "legacy post-checkout swept" || bad "legacy post-checkout still present"
OUT=$(git commit -m nope2 2>&1); RC=$?
echo "$OUT" | grep -q GITBUTLER_ERROR && [ $RC -ne 0 ] && ok "commit blocked on 2.55 via config hook" || bad "commit NOT blocked on 2.55 (rc=$RC)"
echo "$OUT" | grep -q USER-HOOK-RAN && ok "user hook ran from hooks dir" || bad "user hook did not run on 2.55"

say "PHASE 3: checkout away — expect full cleanup, user hook intact"
git checkout main 2>&1 | tail -2
if git config --local --get-regexp '^hook\.' >/dev/null 2>&1; then bad "config entries survived checkout"; else ok "config entries removed"; fi
[ ! -e .git/gitbutler/hooks ] && ok "scripts dir removed" || bad "scripts dir survived"
SHA_FINAL=$(sha1sum .git/hooks/pre-commit | cut -d' ' -f1)
[ "$SHA_BEFORE" = "$SHA_FINAL" ] && ok "user hook still byte-identical" || bad "user hook mutated"
OUT=$(git commit -m works 2>&1); RC=$?
[ $RC -eq 0 ] && echo "$OUT" | grep -q USER-HOOK-RAN && ok "normal commit works again through user hook" || bad "post-cleanup commit failed (rc=$RC)"

say "RESULT"
[ "$FAILURES" -eq 0 ] && echo "ALL CHECKS PASSED" || echo "$FAILURES CHECK(S) FAILED"
exit "$FAILURES"
```

</details>

Byron:
Thanks so much for making this happen! As opposed to the previous PR, this one seems 'small' enough to review with options to split the teardown portion into its own PR.

As this affects the CLI which I am not very familiar with, I think it should be reviewed (or at least taken-over) by someone who is. So let me reel in @slarse to see what he thinks this PR can go.

Thanks everyone!

ichoosetoaccept:
Hey @Byron 
My pleasure, sorry it took so long! This PR is what's been blocking me from actually trying but CLI myself on real projects so I hope this can land soon and of course I'm happy to make any changes you or @slarse deem necessary to help this move forward, including splitting out the teardown logic.

I'm standing by.

Cheers!

slarse:
Sorry for the wait, I've put this on my "must do before end of week" list of tasks!

slarse (commented):
Hi @ichoosetoaccept,

I've had a relatively close look at the code now and like @Byron I think the config-based hooks implementation looks like an outstanding improvement over the current state of affairs. I don't really have anything to note about the hooks implementation itself, it looks solid to me. Big props for taking the time to contribute this!

That being said, I have some concerns with the structural changes to the CLI. `CliError` is meant to convey semantic meaning about what went wrong, but the `Exit` variant is more of a command that carries a _lot_ of meaning in comments but none in code. We also already exit non-zero where we handle `CliResult` - why push that further up for this?

I'd be very happy if we could split this into one PR with the meaningful functional changes to Git hooks - that's all great stuff - and another PR that contains the more structural changes to the CLI where I have some strong opinions about what it should look like. Would that be possible? We can continue the review in that other PR later.

On a side note, I think this carries some amount of risk so I would defer merging it until after the next release, so we can try it out in nightly for a while before it goes public. If there's something wrong with this the blast radius is potentially large as it runs during `but setup`.

ichoosetoaccept:
Thanks @slarse — split done, this PR is now hooks-only. Also rebased onto current master, and updated the description to match.

The exit-code plumbing is gone from here. What's left outside the git 2.54 path is just the opt-out setting, plus setup and teardown telling you when a hook couldn't be installed or removed rather than quietly reporting success.

On the structural one: you mentioned you have a view on how it should look, so I'd rather learn what you had in mind in terms of direction before working on a new PR.

And 0.22.0 is out now. Is that the release you mentioned we should wait for?

slarse:
Hi @ichoosetoaccept,

Sorry, I've been on vacation and should have made a handoff here. My bad! I'll be picking this up today or tomorrow and we'll get it merged as soon as possible.

ichoosetoaccept:
Hi @slarse
Not a problem at all! I also have to keep checking state of this manually.. Github never informs me of new comments despite having all the notification settings on. Not sure what's going on there. Anyway, I'll check back every couple of days to see if you discover any changes are needed or if it was merged. Thank you!

slarse:
@ichoosetoaccept I've gone through this all rather carefully now and I think it's all but ready to go. Thanks for making this effort!

I just wanted two minor details in:

* Resolve Git hooks directory with `git rev-parse --git-path hooks` instead of rolling your own
* Strengthen a test to check that config hooks are _also_ removed when both config and legacy hooks are present

I unfortunately can't push to this PR, I think it's because it's made via an organization ("Maintainers are allowed to edit this pull request." doesn't seem to work in that case). I've attached the patches here, could you apply them?

[0002-test-assert-dual-hook-teardown-test-completely-remov.patch](https://github.com/user-attachments/files/31504616/0002-test-assert-dual-hook-teardown-test-completely-remov.patch)
[0001-fix-use-git-to-resolve-LEGACY_HOOKS_DIR.patch](https://github.com/user-attachments/files/31504617/0001-fix-use-git-to-resolve-LEGACY_HOOKS_DIR.patch)

Also, please rebase on master. It should be clean, I had no conflicts on doing so.

In terms of timing, we have a patch release coming up tomorrow, and I want to merge this _just after_ that release so that we can play with it in nightly for a bit before it goes out in a mainline release.

ichoosetoaccept:
@slarse all done!
Apologies for the odd situation where you can't push to a PR on your own project/repo.
But nevermind that.. I applied your patches, rebased on master and pushed. All done from my side. LMK if you need anything else!

Byron:
> Apologies for the odd situation where you can't push to a PR on your own project/repo.

Actually no need to ever apologize for not ticking that box, after all it allows to push into *your* fork. On the other hand, I am glad it works that way and not the other way around where everyone pushes PRs into this repository :D.

slarse:
@ichoosetoaccept Could you sign the commits? Then we're ready to merge :)

ichoosetoaccept:
@slarse signed and pushed. I use [chezmoi](https://www.chezmoi.io/) to manage dotfiles for both personal and work Macs and apparently had git signing turned off globally for both (likely to avoid git signing collisions between Github (personal) and Gitlab (work)).

In any case, should be ok now!

Note: I haven't rebased this time. We're not THAT much behind master. If you do want this, lmk and I'll rebase and push.

Thanks to you both 🙂.

slarse:
> @slarse signed and pushed. I use [chezmoi](https://www.chezmoi.io/) to manage dotfiles for both personal and work Macs and apparently had git signing turned off globally for both (likely to avoid git signing collisions between Github (personal) and Gitlab (work)).
> 
> In any case, should be ok now!
> 
> Note: I haven't rebased this time. We're not THAT much behind master. If you do want this, lmk and I'll rebase and push.
> 
> Thanks to you both 🙂.

Thanks!

No need to rebase again unless we hit a conflict (unlikely). Just waiting for the next release before I merge this :)

slarse (approved):
LGTM, thanks @ichoosetoaccept for the revisions!

PR #14715, opened 2026-07-13
Recompute the merge base when needed

estib-vega:
## Changes

- Recompute `targetCommitId` when `but config target` switches to a different target ref, while preserving it when the same ref advances.
- Add workspace-level and CLI end-to-end regression coverage for target base recomputation.

## Motivation

- Prevent target and status calculations from using the previous branch's merge-base after changing targets.

## Validation

- Ran focused workspace and CLI tests, `cargo fmt`, targeted `cargo check`, and Clippy with warnings denied.

### DIFF
--- a/crates/but-workspace/src/init.rs
+++ b/crates/but-workspace/src/init.rs
@@ -21,8 +21,10 @@ use but_core::{
 ///   back-fills the legacy metadata in `meta`,
 /// * set `log.excludeDecoration = refs/gitbutler` in the repository-local Git config.
 ///
-/// The target commit id is only computed - as the merge-base between `HEAD` and
-/// `target_ref` - if it isn't already set; an existing value is never overwritten.
+/// The target commit id is computed as the merge-base between `HEAD` and `target_ref`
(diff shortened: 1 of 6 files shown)


crates/but-workspace/src/init.rs
```diff
@@ -83,9 +85,13 @@ pub fn set_target_ref_and_init_project(
...
 
+    let target_ref_changed = project_meta
+        .as_ref()
+        .and_then(|meta| meta.target_ref.as_ref())
+        .is_none_or(|existing| existing.as_ref() != target_ref);
     let sha = match project_meta.as_ref().and_then(|meta| meta.target_commit_id) {
-        Some(existing) => existing,
-        None => {
+        Some(existing) if !target_ref_changed => existing,
+        _ => {
```
estib-vega, line 94: This is the actual fix: Recalculate the target commit ID if the ref changed

estib-vega:
We decided not to got for this approach, since this is intended to be used for changing the target ref even when we have non-empty workspaces.

PR #14720, opened 2026-07-13
Lite: refactor selection selectors

OliverJAsh:


apps/lite/ui/src/projects/project.ts
```diff
@@ -23,6 +24,7 @@ import {
 	type OutlineMode,
 	type TransferMode,
 } from "#ui/outline/mode.ts";
+import { navigationIndexIncludes, type NavigationIndex } from "#ui/workspace/navigation-index.ts";
```
OliverJAsh, line 27: I'm not sure how we feel about importing navigation index into this module, but I think it's necessary if we want derived state to live inside selectors. cc @samhh

PR #14725, opened 2026-07-13
Log IPC invocations as network requests in devtools

samhh:
See commits. I'll be using this in my megamerge for a few days before potentially merging. Ideally there'd be less code in preload, we could potentially extract some stuff to a package boundary.

In the past this code wouldn't have existed, and if it did it would have been a non-neglible maintenance burden. Neither any longer applies.

Observed bugs:

- [ ] `open-in-web-browser` is logged as cancelled

### DIFF
--- a/apps/lite/electron/src/main.ts
+++ b/apps/lite/electron/src/main.ts
@@ -1,6 +1,12 @@
 import { checkForUpdates, registerUpdater } from "./updater.js";
 import WatcherManager from "./watcher.js";
 import {
+	ipcTraceCompleteChannel,
+	ipcTraceCompletionPath,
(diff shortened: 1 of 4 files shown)


OliverJAsh:
> Ideally there'd be less code in preload, we could potentially extract some stuff to a package boundary.

If we merge, please can we do this beforehand? 🙏

I already feel like I'm drowning in complexity.

Actually I would probably prefer if we didn't merge this. 🙈

samhh:
I don't mind not merging this, but in that case I'll keep this open as a draft and in my megamerge for a while, potentially indefinitely.

I suspect merging will become extremely compelling if/when we support remote clients.

PR #14727, opened 2026-07-14
checkout: remove extraneous code

jonathantanmy2:
This extra code confuses both humans and Copilot.


estib-vega (approved):

jonathantanmy2:
I ended up restoring the `write()` of the index - it seems to auto-write in most, but not all, cases, so best to put it back to be safe.

PR #14745, opened 2026-07-14
feat(but): show committed file IDs using short change ID if available

slarse:
If unavailable, we fall back to the short commit ID.

The commit ID remains usable regardless of if the change ID is what's
printed or not.

### DIFF
--- a/crates/but/tests/but/command/status.rs
+++ b/crates/but/tests/but/command/status.rs
@@ -1285,3 +1285,35 @@ fn status_in_edit_mode_delegates_to_resolve_status() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn status_file_prefixed_with_change_id_when_available_and_commit_id_otherwise() {
+    let env = Sandbox::init_scenario_with_target_and_default_settings("one-stack");
+    env.setup_metadata(&["A"]);
+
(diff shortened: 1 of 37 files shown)


crates/but/tests/but/command/status.rs
```diff
@@ -1285,3 +1285,35 @@ fn status_in_edit_mode_delegates_to_resolve_status() -> anyhow::Result<()> {
...
+┊│     1:p A B
+┊●   9477ae7 add A
+┊│     9:t A A
+├╯
+┊
+┴ 0dc3733 (common base) 2000-01-02 add M
+
+Hint: run `but help` for all commands
+
+"#]]);
```
slarse, line 1318: Here's what it looks like. The tip commit has a change ID available so it's used for the short ID of the committed file. The other commit does not have a change ID so the committed file's short ID uses the commit ID instead.

The commit ID remains usable regardless.

I'll see if we can do something about the alignment when some commits have change IDs available and others do not.

slarse:
@davidpdrsn What do you think about this, should we be using change ID to present committed file IDs?

davidpdrsn:
> @davidpdrsn What do you think about this, should we be using change ID to present committed file IDs?

I love it! I'm all in on change ids.

PR #14748, opened 2026-07-14
fix(but): implement discard for path-prefix IDs

### LINKED ISSUE #14703: CLI: `but discard <path-prefix>` panics on an unimplemented `todo!()`
### Version

Built from `main` (current `master`).

### Operating System

macOS

### Distribution Method

Built from source

### Describe the issue

Running `but discard` with a *path-prefix* ID — the kind `but status` shows for a directory that groups several uncommitted files — panics the CLI.

In `crates/but/src/command/legacy/discard.rs`, the `CliId::PathPrefix` match arm is a bare `todo!()`. Every other `CliId` variant is handled: `UncommittedHunkOrFile` / `Uncommitted` perform the discard, and `Branch` / `Commit` / `CommittedFile` / `Stack` return a clear error. Only the path-prefix case is unimplemented, so it panics instead of doing anything.

Path-prefix IDs are a first-class, user-visible ID (`but status` shows them for directories that group several uncommitted files), and `but rub` / `but unapply` already treat them as discardable groups of uncommitted changes — `discard` simply never finished the arm.

### How to reproduce

1. In a workspace with uncommitted changes, create several files under one directory, e.g. `path/to/first.txt` and `path/to/second.txt`.
2. Run `but discard path/to/`.
3. The CLI panics (`not yet implemented`) instead of discarding the files under the prefix.

### Expected behavior

`but discard <path-prefix>` discards all uncommitted changes under the prefix — consistent with how `but commit` and `but rub` already accept the same path-prefix IDs.

ameyypawar:
`but discard <path-prefix>` panicked: the `CliId::PathPrefix` arm in `crates/but/src/command/legacy/discard.rs` was a bare `todo!()`, even though every other ID variant is handled and `but rub` / `but commit` already accept the same path-prefix IDs.

Following slarse's note on the issue, this implements the arm via `DiffSpecBuilder::push_changes_from_path_prefix` — the same builder path those commands use — so a path prefix discards the uncommitted hunks it resolves to.

## Tests (`crates/but/tests/but/command/discard.rs`)
- `discard_by_path_prefix_removes_only_matching_files` — discards the files under the prefix and leaves those outside it. Without the fix it fails with the panic (exit 101).
- `discard_bare_path_prefix_leaves_stack_assigned_changes` — a bare prefix resolves to unassigned hunks only, so the stack-assigned file is left; a follow-up discard-all then confirms that file was still discardable (its survival is the resolver excluding assigned hunks, not `discard` failing on them).

Fixes #14703

### DIFF
--- a/crates/but/src/command/legacy/discard.rs
+++ b/crates/but/src/command/legacy/discard.rs
@@ -49,9 +49,13 @@ pub fn handle(ctx: &mut Context, out: &mut OutputChannel, id: &str) -> Result<()
                 CliId::UncommittedHunkOrFile(uncommitted) => {
                     builder.push_hunk_assignments(uncommitted.hunk_assignments)?;
                 }
-                CliId::PathPrefix { .. } => todo!(),
+                CliId::PathPrefix {
+                    id,
+                    hunk_assignments,
+                } => {
+                    builder.push_changes_from_path_prefix(&id, &hunk_assignments)?;
+                }
                 CliId::Uncommitted { .. } => {
-                    // Discard all uncommitted changes.
                     builder.push_hunk_assignments(worktree_changes.assignments.clone())?;
                 }
                 CliId::Branch { .. } => {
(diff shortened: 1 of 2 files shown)


davidpdrsn (approved):
Thanks!

I've cleaned up the code a bit.

PR #14754, opened 2026-07-14
Remove graduated Svelte feature flags

krlvi:
Remove three default-on frontend feature flags and their deprecated false-side Svelte paths. Keep backend and command API definitions intact.


estib-vega (approved):
makes sense

PR #14762, opened 2026-07-14
Replace Redux with Context

samhh:
_WIP_: This is an experimental PR which we very well may not merge. If we go in this direction we'll likely want to rethink React Query as well. There are also likely some as yet unfixed behaviour changes, including tearing.

As of the latest experimental performance optimisation commit, here's a before and after of navigating once down the same identical, modest outline. The larger deferred render commits to the right pertain to the diff. Ideally this'd be tested again against a much larger outline. (NB we'll have virtualisation in the future, but we need to optimise this regardless. Additionally the approach is master is currently minimally optimised.)

before

after

And here's a before and after on that same outline scrolling from top to bottom, holding down the arrow key:

before

after


samhh:
Opted to stick with what we know absent clear direction on where React is going. Hopefully Fir happens and then we can revisit.

PR #14764, opened 2026-07-14
Group projects by recency in the header project selector

PavelLaptev:
The project selector in the app header showed all projects as one flat list, which gets hard to scan with many repos. This PR groups them into Recent and Other projects.

## Changes

* Select / OptionsGroup (UI package): added support for labeled section headers inside the options list
* ProjectsService: tracks up to 5 recently opened projects in localStorage; entries are removed on project deletion
* AppHeader: builds grouped options from the recents list

## Behavior details

* Grouping only kicks in with 8+ projects — smaller lists stay flat, since headers would just add noise
* Recents keep a stable order: opening a project already in the list doesn't reorder it, avoiding items jumping right after you click them
* Falls back to a flat list if either group would be empty

image

https://github.com/user-attachments/assets/0969412b-c00a-4218-86fb-3f867377380b

### DIFF
--- a/apps/desktop/src/components/views/AppHeader.svelte
+++ b/apps/desktop/src/components/views/AppHeader.svelte
@@ -74,6 +74,9 @@
 
 	const recentProjectIds = projectsService.recentProjectIds;
 
+	// Below this number of projects a flat list is easier to scan than groups.
+	const MIN_PROJECTS_FOR_GROUPING = 8;
+
 	const mappedProjects = $derived.by(() => {
 		const allProjects = projects.response ?? [];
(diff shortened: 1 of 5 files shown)


PavelLaptev:
@krlvi here is my POC. Can you please check it before I merge?

PavelLaptev:
> @krlvi here is my POC. Can you please check it before I merge?

Sorry, you didn't react on my messages, so I merged it. Let me know if there is something you want to change.

krlvi:
its fine - if something breaks we will just fix it. and ux wise it is better

PR #14770, opened 2026-07-15
feat(but): add hidden `but _expand` command for debugging

slarse:
`but _expand <cli-id>`  expands `<cli-id>` to all matching resources.


slarse:
@davidpdrsn FYI, I need this for debugging and testing the ID mapping. Might be useful to you if you run into ID-related issues.

Issue #13345, opened 2026-04-16
"Really couldn't find <stack-id> in the current workspace projection" after virtual_branches.toml corruption


jwstover:
### Version

0.19.7

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

After a session of heavy concurrent `but` CLI usage (with the GUI also open), my `.git/gitbutler/virtual_branches.toml` picked up two forms of corruption (see **Relevant log output** below for the raw TOML entries):

1. A real stack (`3d15e53b-…`, `source_refname = fix/set-end-user-org-on-assets`) had its `[[heads]]` array written **six times** with the same `CommitId`, with two of the entries named `fix/set-end-user-org-on-assets-1` and four named `fix/set-end-user-org-on-assets`.
2. A second bogus stack with UUID `00000000-0000-0000-0000-000000000005` appeared alongside it — no `source_refname`, `in_workspace = true`, and the same six duplicated heads.

From that point on, any CLI command that reads the workspace projection failed with:

```
Error: Really couldn't find 3d15e53b-9f41-4e1a-a252-dedf25718014 in the current workspace projection
```

Broken: `but status --json`, `but clean`, `but clean --dry-run`, `but push`, `but push <branch>`, `but push --dry-run`. Still worked: `but status` (non-JSON), `but oplog list`. That narrow gap is what let me recover — I could see the workspace but not mutate it.

`but status` also displayed six phantom "(no commits)" entries for the same branch name — one per corrupt head:

```
┊╭┄g0 [fix/set-end-user-org-on-assets] (no commits)
┊├┄h0 [fix/set-end-user-org-on-assets] (no commits)
┊├┄i0 [fix/set-end-user-org-on-assets] (no commits)
┊├┄j0 [fix/set-end-user-org-on-assets] (no commits)
┊├┄k0 [fix/set-end-user-org-on-assets-1] (no commits)
┊├┄l0 [fix/set-end-user-org-on-assets-1]
┊●   eeae21e1b2 Add nil-validity comment ...
```

### How I recovered

Since CLI commands couldn't touch the projection, the fix was manual:

1. Backed up `.git/gitbutler/virtual_branches.toml` and created backup refs under `refs/backup/<ts>/` for every `refs/gitbutler/*` branch tip plus the `gitbutler/workspace` tip, so no commits could be lost.
2. Closed the GUI.
3. Edited the TOML by hand:
[...]

PR #14784, opened 2026-07-15
fix(but-meta): collapse within-stack duplicate heads

ameyypawar:
`enforce_constraints` (but-meta) dedups head names across stacks, but a *"preserve if they map to the same projected segment"* carve-out — meant for two **different** stacks legitimately sharing a segment — also fired for a **single stack repeating a branch name** (same name → same segment → preserved). So the duplicate-head corruption reported in #13345 survived cleanup, and `but status` showed the phantom "(no commits)" head entries.

This drops within-stack repeats unconditionally: a stack is a series of distinct branches and must never list the same one twice.

## Scope
This collapses the **duplicate heads** only. The `Really couldn't find <uuid>` wedge in #13345 also involves a shadow duplicate *stack*, which this does **not** remove — so this **Addresses** the issue rather than fully fixing it. I've opened a note on the issue to get direction on the shadow-stack part before the planned vb.toml removal.

## Tests (`crates/but-meta/tests/meta/ref_metadata_legacy.rs`)
- `removes_within_stack_duplicate_heads_even_when_mapped_to_a_segment` — one stack lists `shared` twice under a real projected segment (the exact condition that fires the carve-out); asserts it keeps `shared` once. Reverting only the fix makes it fail (the stack keeps `shared` twice).
- The existing `removes_duplicate_heads_…` and `preserves_duplicate_heads_…` still pass, so cross-stack behavior is unchanged.

Addresses #13345

### DIFF
--- a/crates/but-meta/src/legacy/mod.rs
+++ b/crates/but-meta/src/legacy/mod.rs
@@ -198,11 +198,20 @@ impl Snapshot {
                 }
             }
 
+            // A single stack must never list the same branch name twice: that's
+            // always corruption (e.g. from a racy write, #13345), never a legitimate
+            // shared segment. Drop within-stack repeats unconditionally - ahead of the
+            // cross-stack `preserve_duplicate` carve-out below, which only makes sense
+            // for two *different* stacks resolving to the same projected segment.
+            let mut seen_in_this_stack = HashSet::<String>::new();
             let head_indices_to_remove: Vec<_> = stack
                 .heads
                 .iter()
                 .enumerate()
                 .filter_map(|(head_idx, head)| {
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks for the fix!

While I didn't validate that this is really addressing https://github.com/gitbutlerapp/gitbutler/issues/13345, the fix will make sure that such incorrect branch metadata is fixed.

It's notable that this should only matter to old code still running, of which there should be less. If it's new code choking on this, it's a problem to solve in `but-graph` instead. After all, it reconciles the metadata with reality, and this metadata is considered untrusted and incorrect, merely by it being stale and/or written by old code which may mess things up in its own way.

PR #14786, opened 2026-07-15
Lite misc

samhh:
Tidy up operands a bit, most notably removing hierarchical IDs (primarily stack IDs). I toyed with stringly operands however they're not very ergonomic given a) we need file parents in identity and b) hunks carry additional information that's not necessary for identity.

cc @OliverJAsh for visibility

### DIFF
--- a/apps/lite/ui/src/operands.ts
+++ b/apps/lite/ui/src/operands.ts
@@ -117,10 +117,6 @@ export const operandFileParent = (operand: Operand): FileParent | null =>
 	);
 
 export const operandContains = (a: Operand, b: Operand) => {
-	if (operandEquals(a, b)) return true;
-
 	const bFileParent = operandFileParent(b);
-	if (bFileParent && operandEquals(a, bFileParent)) return true;
-
-	return false;
+	return bFileParent && operandEquals(a, bFileParent);
 };
--- a/apps/lite/ui/src/operands.ts
+++ b/apps/lite/ui/src/operands.ts
@@ -2,9 +2,9 @@ import { Match } from "effect";
 import type { HunkLineSelection } from "#ui/hunk.ts";
 
 /** @public */
-export type BranchFileParent = { stackId: string; branchRef: Array<number> };
+export type BranchFileParent = { branchRef: Array<number> };
 /** @public */
-export type CommitFileParent = { stackId: string; commitId: string };
+export type CommitFileParent = { commitId: string };
 
 export type FileParent =
 	| { _tag: "UncommittedChanges" }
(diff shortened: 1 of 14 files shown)


apps/lite/ui/src/operands.ts
```diff
@@ -36,12 +34,12 @@ export type StackOperand = {
 };
 
 /** @public */
-export type BranchOperand = StackOperand & {
+export type BranchOperand = {
 	branchRef: Array<number>;
 };
 
 /** @public */
-export type CommitOperand = StackOperand & {
```
OliverJAsh, line 44: I thought the stack ID was needed here to disambiguate the selection when the commit appears multiple times. Granted that's a temporary problem though.

apps/lite/ui/src/operands.ts
```diff
@@ -117,10 +117,6 @@ export const operandFileParent = (operand: Operand): FileParent | null =>
 	);
 
 export const operandContains = (a: Operand, b: Operand) => {
-	if (operandEquals(a, b)) return true;
```
OliverJAsh, line 120: FWIW I based this on `contains`: https://developer.mozilla.org/en-US/docs/Web/API/Node/contains

apps/lite/ui/src/operands.ts
samhh: Yes, I’ve removed it on the basis that this is temporary and not a supported state. Arguably even then it’s not appropriate for commit identity and better belongs encoded alongside the operand in selection state.

OliverJAsh: > not a supported state

I do wonder if we should still preserve support for it.

> better belongs encoded alongside the operand in selection state

I keep thinking that selection state shouldn't use `Operand`, especially when we separate selection for uncommitted files and stacks, in which case the stack selection is just branch or commit.

apps/lite/ui/src/operands.ts
```diff
@@ -1,99 +1,63 @@
...
+	| ({ _tag: "Commit" } & CommitOperand)
+	| ({ _tag: "File" } & FileOperand)
+	| ({ _tag: "Hunk" } & HunkOperand);
 
-/** @public */
-export const commitFileParent = ({ commitId }: CommitFileParent): FileParent => ({
-	_tag: "Commit",
-	commitId,
-});
+export type FileParent = Extract<Operand, { _tag: "UncommittedChanges" | "Branch" | "Commit" }>;
```
OliverJAsh, line 12: https://github.com/gitbutlerapp/gitbutler/pull/14786/changes/d766336f2c2f8ea7100141d08b849ff58c41d541

> operandContains depends upon this implementation detail anyway

How so?

samhh: Bit fuzzy now, but: if operands hypothetically needed to carry more data than file parents then we wouldn't be able to perform the conversion. This wasn't an issue in practice because they represent the same things, albeit a subset in the case of file parent. I think that's what I meant.

PR #14787, opened 2026-07-15
Lite misc

samhh:
Various fixes to end the day.

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/CommandPalette.tsx
+++ b/apps/lite/ui/src/routes/project/$id/workspace/CommandPalette.tsx
@@ -96,6 +96,7 @@ export const CommandPalette: FC<Props> = ({ open, onOpenChange }) => {
 			getItemKey={(x) => x.id}
 			getItemLabel={(x) => x.name}
 			getItemType={(x) => <Kbd hotkey={x.hotkey} />}
+			itemToStringValue={(x) => `${x.group}: ${x.name}`}
 			items={items}
 			open={open}
 			onOpenChange={onOpenChange}
--- a/apps/lite/ui/src/routes/project/$id/workspace/Details.module.css
+++ b/apps/lite/ui/src/routes/project/$id/workspace/Details.module.css
@@ -25,20 +25,16 @@
 
 .title {
 	display: grid;
-	grid-template-rows:
-		/* Match regular button height. */
-		28px
-		auto;
+	/* Match regular button height. */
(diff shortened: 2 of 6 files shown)


OliverJAsh:
Regression from b8fbe94db98a9965bd342453d80c65c1290a0430:

image

OliverJAsh:
> Regression from [b8fbe94](https://github.com/gitbutlerapp/gitbutler/commit/b8fbe94db98a9965bd342453d80c65c1290a0430):

It looks like this was fixed in https://github.com/gitbutlerapp/gitbutler/pull/14816/changes/e7fbfd2d7925e60ce1da282c80c01e7c79800490 using truncation.

I'm not sure if we want truncation though, hence my previous implementation to let it wrap (and preserve vertical alignment when doing so). I like seeing the full commit/PR title.

samhh:
Could you provide a screenshot? I don't think I observed the previous behaviour.

OliverJAsh:
Before b8fbe94db98a9965bd342453d80c65c1290a0430:

Screenshot 2026-07-17 at 11 34 39

After b8fbe94db98a9965bd342453d80c65c1290a0430:

Screenshot 2026-07-17 at 11 35 00


After e7fbfd2d7925e60ce1da282c80c01e7c79800490:

Screenshot 2026-07-17 at 11 34 03

OliverJAsh:
Let's leave it for now.

@PavelLaptev It would be good to think about how we want to handle/display long commit/PR titles.

samhh:
Yeah, I was wondering what would happen if it was three or four lines long. You can't maintain horizontal alignment.

We should discourage such long commit titles & PR/branch titles, though we'll have to handle this issue for narrow viewports regardless.

OliverJAsh:
Yeah sorry I meant "commit/PR titles wider than the viewport", not necessarily just "long titles".

PavelLaptev:
I think we can:
1. Show the full message, regardless of its length.

image

2. Show a truncated version and display the full title on hover.

image

image

Right now I prefer showing the full title, I think it's better UX even if the look of it could be not always nice.

samhh:
We still have to consider overflow behaviour for short viewports.

OliverJAsh:
If we wrap, when would it overflow?

samhh:
Overflow was the wrong word. I mean the diff will fall out of view faster on a short viewport, and generally be more cramped.

That's probably fine actually.

PR #14791, opened 2026-07-15
fix(skill): Update installs only on setup

reneleonhardt:
## 🧢 Changes

- Make ordinary commands report stale agent skills without rewriting them.
- Run the automatic skill update only after successful explicit project setup.
- Remove implicit updates from status-after-mutation JSON handling.
- Cover read-only commands, ordinary mutations, and successful setup.

## ☕️ Reasoning

Agent-skill repair changes global configuration and therefore belongs at an explicit lifecycle boundary. Inspection and unrelated mutations should report stale state without silently repairing it; successful setup is where installation repair is expected and can be reported clearly.

Concrete read-only examples are `but alias list`, `but config user`, and `but config target` when the latter is called without a branch or `--push-remote`. These inspection forms may report that an agent skill is stale, but they do not repair or rewrite it. Their corresponding add, set, unset, or target-changing forms remain mutations.

For agentic coding, keeping that boundary explicit prevents ordinary commands from unexpectedly requiring permission to modify global tooling state.

## 🧪 Reproduction

With an installed GitButler agent skill older than the bundled version:

```sh
but alias list
```

Before this fix, listing aliases silently rewrote the stale global skill instead of only reporting that an update was available.

## ✅ Verification

- Focused agent-skill notice tests: 5 passed, 0 failed.
- Tests prove ordinary observation and mutation preserve the stale file while successful setup updates it.
- The exact regression passed with default features and `--no-default-features`; the latter proves the test no longer invokes unavailable legacy setup.
- `cargo fmt --check --all` passed.

### DIFF
--- a/crates/but/src/command/skill/freshness.rs
+++ b/crates/but/src/command/skill/freshness.rs
@@ -38,12 +38,7 @@ impl AgentSkillNotice {
 }
 
 pub(crate) fn agent_skill_notice(current_dir: &std::path::Path) -> Option<AgentSkillNotice> {
-    let update = agent_skill_freshness_check();
-    let hint = agent_skill_install_hint(current_dir);
-    match update {
-        Some(failed @ AgentSkillNotice::UpdateFailed(_)) => Some(failed),
-        update => hint.or(update),
-    }
+    agent_skill_install_hint(current_dir)
 }
 
 pub(crate) fn agent_skill_update_notice() -> Option<String> {
--- a/crates/but/src/lib.rs
+++ b/crates/but/src/lib.rs
@@ -1349,10 +1349,21 @@ async fn match_subcommand(
             };
             let mut ctx = but_ctx::Context::from_repo_with_settings(repo, app_settings.clone())?;
             let mut guard = ctx.exclusive_worktree_access();
-            command::legacy::setup::repo(&mut ctx, &args.current_dir, out, guard.write_permission())
-                .context("Failed to set up GitButler project.")
-                .emit_metrics(metrics_ctx)
-                .map_err(CliError::from)
+            let result = command::legacy::setup::repo(
+                &mut ctx,
+                &args.current_dir,
+                out,
+                guard.write_permission(),
+            )
+            .context("Failed to set up GitButler project.");
+            if result.is_ok()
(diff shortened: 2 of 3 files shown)


Byron:
Thanks a lot!

The way I see it is that read-only commands could do writes and fixes, which this PR improves on.
Due to it being mostly about skills, it's probably best for @krlvi to take a look.

krlvi (changes requested):
this is interesting. in a way, the skill update should probably be linked to the binary update. But in general it is super important that there is some kind of mechanism for updating the skill on new verison of the app. The main reason is that commands and outputs are still changing quite a bit, and having an old version of the skill with a new version of the binary would be suuuuper bad. 
Im gonna mark this red while I think a bit more about this

krlvi:
a path forward could be to make the skill files somehow an asset that gets distributed with the app, and then skill installation is a matter of creating symlinks. That way updating the app binary would guarantee that agents are not operating with stale skill files. 
There are still big api changes to the but CLI. For example, with the next release, the agents will operate on stable change IDs for commits, instead of the commit shas, which means the steering must be quite different

reneleonhardt:
Thanks, that direction makes sense.

For agentic coding in sandboxes, ordinary `but` commands must never repair
skill installations or overwrite agent configuration. The explicit lifecycle
is `but agent setup` / `but skill install` to install, and `but skill check` to
diagnose or explicitly update. A linked app-owned skill asset could preserve
that contract: app updates change only the asset target, while normal CLI
invocations remain non-mutating.

If that is the intended architecture, should I treat this PR as superseded and
maintain the sandbox-safe behavior in my fork until it lands, or would you
still like a narrow no-write boundary independently?

reneleonhardt:
Closing this because upstream has chosen a different skill-update lifecycle.
#14836 refreshes discovered global skill installations during CLI invocation,
and commit steering is moving to stable change IDs. The setup-only policy in
this PR no longer matches that architecture.

The sandbox constraint remains: ordinary commands may not have permission to
rewrite global agent configuration. I will retain the no-implicit-write behavior
in my fork and revisit it if app-owned or distributed skill assets provide
freshness without command-time writes.

PR #14795, opened 2026-07-15
fix(pull): Synchronize configured local target

reneleonhardt:
## 🧢 Changes

- Resolve the existing local branch that is actually configured to track the fetched target.
  - Fast-forward that branch with ancestry, linked-worktree, and compare-and-swap guards.
  - Leave same-named but unconfigured branches and unrelated invalid tracking entries untouched.
  - Reject ambiguous tracking when multiple local branches target the same remote branch.
  - Cover successful synchronization, divergence, alternate local names, ambiguity, and checked-out-worktree rejection.
- Record the local-target transition in the oplog so undo and redo restore the
  local ref together with workspace, index, and project metadata.
- Keep snapshot preparation state-neutral and roll back the complete unpublished
  effect when integration or publication fails.

## ☕️ Reasoning

The local tracking branch is also a user-visible base for ordinary Git and worktree tooling. A successful pull should leave that ref aligned with its fetched target whenever a fast-forward is safe.

This matters for agentic coding because isolated worktrees are commonly created from the local tracking branch immediately after pulling.

## 🧪 Reproduction

With local `master` behind its configured upstream and not checked out:

```sh
but pull
git rev-parse master 'master@{upstream}'
```

Before this fix, `but pull` reported success while the two object IDs remained different. A worktree subsequently created from `master` therefore started from the stale commit.

## ✅ Verification

- Fourteen focused local-target tests passed.
- The stale snapshot-preparation regression and two oplog publication tests
  passed.
- Failure injection preserves the target ref, workspace, index, legacy
  metadata, oplog head, and next undo/redo selection.
- Formatting and diff checks passed.
- Independent correction review found no Critical or Important issues.

### DIFF
--- a/crates/but/src/command/legacy/pull/mod.rs
+++ b/crates/but/src/command/legacy/pull/mod.rs
@@ -2,9 +2,15 @@ mod json;
 
 use std::fmt::Write;
 
+use anyhow::Context as _;
 use bstr::ByteSlice;
 use but_core::{DryRun, RepositoryExt};
 use but_ctx::Context;
+use gitbutler_oplog::{
+    LocalTargetSnapshot, OplogExt,
+    entry::{OperationKind, SnapshotDetails},
+};
+use gix::refs::transaction::PreviousValue;
 use json::{BaseBranchInfo, BranchStatusInfo, PullCheckOutput, UpstreamCommit, UpstreamInfo};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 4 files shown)


Byron:
Thanks for contributing Screenshot 2026-07-16 at 15 50 56 😅?

While I give every PR the benefit of the doubt and my best to find reviewers or review it myself, with this one I think it's fair to point out the contributing guidelines which mention to open an issue first to discuss the approach. With this PR I don't know where to start, and don't feel comfortable to invite others. 

Thus, I will heavy-handedly close it while inviting a dialogue in an issue or here. Also CC @davidpdrsn as GH suggestion, maybe he knows more and thinks this should be reopened and reviewed.

reneleonhardt:
I replaced this implementation with a focused design and branch.

The replacement preserves the product fix: `but pull` safely synchronizes the
configured local tracking branch, rolls it back if workspace integration fails,
and restores it with ordinary undo and redo. It retains ambiguity, divergence,
linked-worktree, tracking, and compare-and-swap guards.

The diff is reduced from 1,669 to 768 insertions by excluding generalized oplog
publication and restore transactions, which remain separate follow-up work.

Design issue: #14823

Replacement PR: #14824

PR #14796, opened 2026-07-15
fix(commit): Report linked worktrees as unsupported

reneleonhardt:
## 🧢 Changes

- Detect linked worktrees before project initialization.
- Return a specific unsupported-context error instead of suggesting `but setup`.
- Document the worktree-local Git fallback and cover the diagnostic with a regression test.

## ☕️ Reasoning

`but commit` depends on workspace mutation semantics that are not available from an independent linked worktree. The CLI should report that capability boundary directly without changing project or stack metadata.

For agentic coding, a specific rejection prevents an automated session from treating setup as the remedy. Actual linked-worktree commit support remains separate work.

## 🧪 Reproduction

From a linked worktree on branch `linked-mutation`:

```sh
but commit linked-mutation -m test
```

Before this fix, the command reported that GitButler setup was required, even though setup could not enable linked-worktree commits.

## ✅ Verification

- The linked-worktree CLI regression passed.
- Existing main-worktree commit coverage passed, confirming the normal command path is unchanged.
- The merged PR's Rust tests, lint, blackbox, Playwright, compatibility, and platform checks passed.
- Maintainer review approved the explicit rejection and documented Git fallback.
- The merged commit is signed and verified.


Byron (approved):
Thanks a lot for the fix!

At this time, it's certainly best to actively refuse working on such a situation when also saying that we plan to support this natively. 

CC @krlvi for the skill change, which seems reasonable to me. It might also point at an issue with `but setup` from linked worktrees, even though it fails gracefully already:

```
❯ but setup
Setting up GitButler project...

→ Adding repository to GitButler project registry
Error: Failed to set up GitButler project.

Caused by:
    The repository at . is a non-main worktree. GitButler requires the main worktree.
```

For now I'd think we can merge this, maybe it's not a bad thing to prevent the agent to run into trouble with `but` that it then tries to fix/workaround somehow.

PR #14797, opened 2026-07-15
fix(workspace): Ignore direct managed ancestry

reneleonhardt:
## 🧢 Changes

- Report an ancestor workspace commit only when commits actually exist outside it.
- Keep direct and advanced ad-hoc refs out of ancestor-workspace state.
- Restore submodules at the checkout destination already selected by worktree traversal instead of resolving the path again through the repository.
- Cover direct and one-commit-ahead ad-hoc repository scenarios.

## ☕️ Reasoning

Consumers use ancestor-workspace state to decide whether commits outside a managed workspace commit must be reconstructed. That state must be absent when an ad-hoc ref has no reconstructible outside segment.

Submodule discard already receives the authoritative checkout destination from worktree traversal. Resolving it again through the repository can change the path base in a linked worktree and fail with `prefix not found`.

For agentic coding, these fixes prevent automation from reconstructing nonexistent ancestry and make discard behavior reliable from linked worktrees.

## 🧪 Reproduction

```sh
cargo test -p but-workspace direct_workspace_ref_has_no_ancestor_workspace_commit
cargo test -p but-workspace advanced_direct_workspace_ref_has_no_ancestor_workspace_commit
```

Before the ancestry fix, a direct managed-workspace ref produced an `AncestorWorkspaceCommit` with an empty `commits_outside` list. During maintainer validation from a linked worktree, four submodule-discard cases also failed with `prefix not found` before the checkout path correction.

## ✅ Verification

- Repository-level direct and one-commit-ahead ad-hoc ancestry regressions passed.
- Existing submodule-discard coverage passed with the corrected checkout destination.
- The merged PR's Rust tests, lint, blackbox, compatibility, and platform checks passed.
- All merged commits are signed and verified.

### DIFF
--- a/crates/but-workspace/src/ref_info.rs
+++ b/crates/but-workspace/src/ref_info.rs
@@ -532,24 +532,6 @@ fn ancestor_workspace_commit_if_outside(
     })
 }
 
-#[cfg(test)]
-mod ancestor_workspace_tests {
-    use super::ancestor_workspace_commit_if_outside;
-
-    #[test]
-    fn empty_outside_commits_are_not_ancestor_workspace_state() {
-        let actual = ancestor_workspace_commit_if_outside(
-            Vec::new(),
-            Some((but_graph::SegmentIndex::new(0), 0)),
-        );
-
-        assert!(
-            actual.is_none(),
-            "the managed commit itself must not be reported as an ancestor"
-        );
-    }
-}
-
 /// Gather information about graph and the workspace that might be associated with it,
 /// based on data in `repo` and `meta`. Use `options` to further configure the call.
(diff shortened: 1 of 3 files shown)


Byron (approved):
Thanks a lot for attempting a fix.

The scenario here is this:

debug-graph-00

Additionally, I added this one (with one commit ahead, but a non-managed reference)

debug-graph-00

And I think this fix is correct, all under the assumption that it's indeed correct that AdHoc workspaces shouldn't deal with workspace commits.

It's also notable that this extra handling comes from `RefInfo`, parts of which certainly need an overhaul.

PR #14798, opened 2026-07-15
fix(cli): Preserve remote state

reneleonhardt:
## 🧢 Changes

- Preserve the configured push remote when the target branch changes without an explicit replacement.
- Add `but config push-remote [remote]` to view or update the publication remote without changing the target.
- Retain upstream's combined `but config target <branch> --push-remote <remote>` behavior.
- Reject unknown remotes or a standalone update without a configured target before changing project metadata.
- Prune deleted remote-tracking branches whenever the shared fetch path refreshes a remote.

## ☕️ Reasoning

Target configuration and publication routing are separate state. Changing one must not silently reset the other, and a fork remote should be configurable without reinitializing the target branch.

This is especially important for agentic coding: automation must be able to select the user's publication remote explicitly without rewriting unrelated workspace state or accidentally targeting the canonical upstream repository.

The shared fetch path must also remove remote-tracking refs whose remote branches were deleted. Otherwise pull, land, refresh, push dry-run, and desktop or server refreshes can report success while exposing stale remote state.

## 🧪 Reproduction

With an `origin` remote already configured:

```sh
but config push-remote origin
but config push-remote
```

Before this fix, there was no non-legacy standalone command for changing the push remote without resetting the target.

The combined form remains available when both values should change:

```sh
but config target upstream/master --push-remote origin
```

After a branch has been deleted from the configured remote:

```sh
but fetch
git branch -r
```

Before this fix, GitButler's shared fetch invocation omitted `--prune`, so the deleted branch could remain visible locally after a successful refresh.

## ✅ Verification

- Lower workspace and API tests cover target preservation, successful updates, unknown remotes, and the missing-target error.
- CLI and parser tests cover setting and inspecting the effective push remote while retaining the combined target-plus-push-remote form.
- The shared fetch regression proves `--prune` is included without changing configured refspec order.
- Rust tests and lint, SDK generation checks, Node checks, blackbox tests, and Playwright checks passed on the pull request.


Byron:

As it would take me quite some time to figure out the current desired way of working, it's probably easier to follow GitHubs recommendation and see if @estib-vega has some intuition about this PR. Thank you.

estib-vega:
I like it in principle, for UX reasons. 
We should do it in using non-legacy APIs for this, though. I can take this over

estib-vega (approved):

estib-vega:
Thanks for the contribution @reneleonhardt, it's highly appreciated

PR #14807, opened 2026-07-16
test(but): centralize source of truth for agent environment variables

slarse:
I had agents constantly rerunning tests and setting `PI_CODING_AGENT=` to unset the harnes env variable, as that was polluting test output. This turned out to be caused by drift between the environment variables we look for, and the environment variables we unset when running `but`s tests.

This PR centralizes a single source of truth for agent environment variables and ensures they're all unset when running tests.


slarse:
@krlvi FYI merging this

PR #14808, opened 2026-07-16
but status: make change IDs the unmistakable primary commit identifier

krlvi:
`but status -v` showed the commit sha directly next to the change ID (`tww 41ef713 …`), which makes it easy for agents to grab the wrong identifier: the sha resolves today but goes stale after any amend, while the change ID is stable.

- Verbose commit lines demote the sha to a labelled note after the timestamp: `kyn Kiril Videlov 2026-07-08 … (sha 50c9fa2)`. Commits without a change ID (e.g. upstream-only) keep the sha as their leading identifier.
- One-shot output pads committed-file ID prefixes to match the change ID shown on the commit line (`k:ww` → `kyn:ww`), so the parent/child linkage is visible without color. The TUI keeps the minimal IDs.
- Agent invocations get one extra hint line spelling out the grammar: the first token on each line is the ID to use in commands, and (in verbose mode) that the parenthesized sha changes on every amend.

Both the padded and minimal ID forms resolve in the parser via prefix matching, so nothing changes for what users can type.

The GitButler skill file update describing the new output format will come in a follow-up PR.

### DIFF
--- a/crates/but/src/command/legacy/status/mod.rs
+++ b/crates/but/src/command/legacy/status/mod.rs
@@ -34,7 +34,7 @@ use crate::{
         upstream::{self, BranchStatus as UpstreamBranchStatus},
         workspace_target,
     },
-    id::{SegmentWithId, ShortId, StackWithId, TreeChangeWithId},
+    id::{ChangeIdWithShortId, SegmentWithId, ShortId, StackWithId, TreeChangeWithId},
     tui::text::truncate_text,
     utils::{
         InputOutputChannel, OutputChannel, WriteWithUtils, shorten_hex_object_id,
@@ -198,6 +198,7 @@ struct StatusContext<'a> {
     branch_merge_statuses: BTreeMap<String, UpstreamBranchStatus>,
     has_branches: bool,
     is_agent_invocation: bool,
+    render_mode: StatusRenderMode,
     is_paged: bool,
     should_truncate_for_terminal: bool,
     id_map: IdMap,
@@ -566,6 +567,7 @@ fn build_status_context<'a>(
         has_branches,
         is_agent_invocation: matches!(format, OutputFormat::Agent)
             || crate::utils::detect_agent::detect().is_some(),
+        render_mode,
         is_paged,
         should_truncate_for_terminal,
         id_map,
@@ -684,6 +686,15 @@ fn print_hint(
             "Hint: ◐ means rewritten locally vs upstream.",
             crate::theme::get().hint,
         )]))?;
+    }
+
+    if status_ctx.is_agent_invocation {
+        let id_hint = if status_ctx.flags.verbose {
+            "Hint: the first token on each line is the ID to use in commands. The sha in parentheses is informational and changes on every amend."
+        } else {
+            "Hint: the first token on each line is the ID to use in commands."
+        };
+        output.hint(Vec::from([Span::styled(id_hint, crate::theme::get().hint)]))?;
     }
 
     let hint_text = if not_on_workspace {
@@ -1385,10 +1396,7 @@ fn print_group(
                     status_ctx,
                     stack_with_id.id,
                     commit.short_id.clone(),
-                    commit
-                        .change_id
-                        .as_ref()
-                        .map(|change_id| change_id.short_id.clone()),
+                    commit.change_id.as_ref(),
                     &inner.inner,
                     CommitChanges::Workspace(&commit.tree_changes_using_repo(&repo)?),
                     classification,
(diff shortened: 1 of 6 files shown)


crates/but/src/command/legacy/status/mod.rs
```diff
@@ -1534,10 +1540,17 @@ fn print_commit(
 
     let upstream_commit = matches!(commit_changes, CommitChanges::Remote(_));
 
+    // One-shot output pads file ID prefixes to match the change ID shown on
+    // the commit line; the TUI keeps the minimal IDs.
+    let padded_file_id_prefix = match output {
+        StatusOutput::Immediate { .. } => change_id.map(ChangeIdWithShortId::padded_short_id),
+        StatusOutput::Buffer { .. } => None,
+    };
```
davidpdrsn, line 1550: I believe there is another way of checking if we're rendering the TUI 🤔 using `output` feels a bit brittle

davidpdrsn: Yes, `build_status_context` gets a `render_mode` which we probably should store on `StatusContext` and use instead.

crates/but/src/command/legacy/status/mod.rs
```diff
@@ -1632,9 +1645,14 @@ fn print_commit(
     if status_ctx.flags.show_files.show_files_for(commit.id) {
         match commit_changes {
             CommitChanges::Workspace(tree_changes) => {
+                let display_id =
+                    |short_id: &str| match (&padded_file_id_prefix, short_id.split_once(':')) {
+                        (Some(prefix), Some((_, rest))) => format!("{prefix}:{rest}"),
+                        _ => short_id.to_owned(),
+                    };
```
davidpdrsn: nit: maybe this is just me but I generally dislike inline closures like this and always tell my clanker to make standalone functions instead.

crates/but/src/id/mod.rs
```diff
@@ -265,6 +265,20 @@ pub struct ChangeIdWithShortId {
...
 }
 
+impl ChangeIdWithShortId {
+    /// The short ID padded with further change ID characters to a minimum
+    /// length of three, matching how the change ID is displayed on commit
+    /// lines.
+    pub fn padded_short_id(&self) -> String {
+        let mut id = self.short_id.clone();
+        let full = self.change_id.to_string();
+        if let Some(padding) = full.get(id.len()..3.min(full.len())) {
```
davidpdrsn: This `3` should probably be in a shared constant.

crates/but/src/command/legacy/status/mod.rs
krlvi: Good call — `render_mode` is now stored on `StatusContext` and `print_commit` matches `StatusRenderMode` instead of the output enum. Behavior is unchanged: the TUI refresh path builds its context with `Tui` mode, and the post-TUI final print reuses the TUI-built lines.

crates/but/src/command/legacy/status/mod.rs
krlvi: Extracted to a standalone `displayed_file_id()` function.

crates/but/src/id/mod.rs
krlvi: Added `MIN_DISPLAYED_CHANGE_ID_CHARS` in `id/mod.rs` and used it in `padded_short_id()`.

crates/but/src/command/legacy/status/mod.rs
```diff
@@ -686,6 +686,15 @@ fn print_hint(
         )]))?;
     }
 
+    if status_ctx.is_agent_invocation {
+        let id_hint = if status_ctx.flags.verbose {
+            "Hint: the first token on each line is the ID to use in commands. Commit IDs are stable change IDs; the sha in parentheses changes on every amend."
+        } else {
```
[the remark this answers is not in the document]
krlvi: Reworded to "The sha in parentheses is informational and changes on every amend" — no longer claims every commit line has a change ID.

crates/but/src/command/legacy/status/mod.rs
```diff
@@ -1762,20 +1777,31 @@ fn display_cli_commit_details(
...
+        // When a change ID is shown it is the primary identifier; the sha is
+        // demoted to a labelled note after the timestamp so the two cannot be
+        // mistaken for one another.
+        let (sha, sha_note) = if change_id_spans.is_empty() {
+            (Vec::from_iter([start_id, end_id]), None)
+        } else {
+            (
+                Vec::new(),
+                Some(Span::styled(
+                    format!(" (sha {}{})", start_id.content, end_id.content),
```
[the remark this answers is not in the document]
krlvi: Intentionally deferred — the bundled skill update describing the new output format lands in a follow-up PR, as noted in the PR description. Sequencing it separately keeps this PR about the output change itself.

PR #14813, opened 2026-07-16
fix(config): Refresh stored target metadata

reneleonhardt:
## 🧢 Changes

- Add `but config target --refresh` to update the stored target commit from the
  already configured remote-tracking ref.
- Keep `but config target --push-remote <remote>` working without a positional
  target branch.
- Preserve target identity, refs, workspace, branch tips, index, and worktree.
- Reject refresh while named or anonymous stacks are applied, and reject a
  simultaneous target change.
- Normalize fully qualified remote-ref input.

## ☕️ Reasoning

Fetching can advance the configured remote-tracking ref while the project keeps
an older stored target commit. Re-selecting the same target intentionally
preserves that commit, leaving no supported metadata-only repair.

An explicit refresh provides a bounded recovery path without pulling, moving a
protected local default branch, or disturbing workspace state. This is useful
for human and agentic coding workflows that separate remote refresh from
workspace integration.

## 🧪 Reproduction

```sh
but pull --check
but config target --refresh
```

Before this fix, the first command could advance the configured tracking ref
while the stored target commit remained stale. Re-selecting the same target was
intentionally a no-op.

## ✅ Verification

- Focused workspace, parser, CLI, and push-remote compatibility tests passed.
- The tests cover idempotence, applied-stack rejection, and preservation of
  refs, HEAD, workspace, index, worktree, target name, and push remote.
- Formatting, warnings-denied Clippy with `--no-default-features`, focused diff
  review, and commit-signature verification passed.

### DIFF
--- a/crates/but-workspace/src/init.rs
+++ b/crates/but-workspace/src/init.rs
@@ -11,6 +11,38 @@ use but_core::{
     ref_metadata::ProjectMeta,
 };
 
+/// Refresh the stored target commit from the configured remote-tracking ref.
+///
+/// Returns `true` when metadata changed. This does not move refs or alter the
+/// workspace; callers must reject applied stacks and hold exclusive repository
+/// access before invoking it.
+pub fn refresh_target_commit_id(
+    repo: &gix::Repository,
+    meta: &mut impl RefMetadata,
+) -> Result<bool> {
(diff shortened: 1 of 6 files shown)


Byron:
Thanks for tackling this. 

However, I think it's best to close it and wait for a couple of other fixes to land that might affect it positively.

PR #14814, opened 2026-07-16
fix(workspace): Drop preserved anonymous remnants

reneleonhardt:
## 🧢 Changes

- Drop a projected workspace parent after unapply when the preserved named
  branch already contains that history.
- Keep independent and diverged anonymous workspace history unchanged.
- Cover every workspace disposition and the related multi-parent apply path.

## ☕️ Reasoning

Unapplying a named stack could leave an anonymous lane that preserved no unique
commits. The lane had no usable CLI selector, blocked target changes, and made
agentic coding workflows require manual recovery even though the named branch
already retained the complete history.

The containment check removes only redundant projected history. It remains
conservative when histories diverge, are disjoint, or cannot be compared.

## 🧪 Reproduction

1. Advance a named branch ref beyond its currently projected workspace tip.
2. Unapply that branch, then unapply the remaining empty stack.
3. Run `but status -fv`.

Before this change, status rendered an anonymous lane and then failed with
`Could not find branch CLI id '' in IdMap`; changing the target remained blocked.

## ✅ Verification

- Format check passed.
- All 53 focused apply/unapply tests passed.
- Clippy passed with warnings denied.
- The signed-tip regression test passed.
- A disposable CLI reproduction removed the redundant parent, preserved named
  and component refs plus linked-worktree staged state, and allowed target
  configuration afterward.

### DIFF
--- a/crates/but-workspace/src/branch/unapply.rs
+++ b/crates/but-workspace/src/branch/unapply.rs
@@ -468,9 +468,10 @@ pub(crate) mod function {
         workspace_ref_name: &FullNameRef,
         ws_md: &but_core::ref_metadata::Workspace,
         disposition: WorkspaceDisposition,
-        excluded_anonymous_tip_id: Option<gix::ObjectId>,
+        unapplied_branch_tip_id: Option<gix::ObjectId>,
     ) -> anyhow::Result<WorkspaceRefUpdateAfterUnapply> {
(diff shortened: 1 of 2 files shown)


Byron:
Thanks for tackling this. I will take a look!

Byron:
I am sorry, but I don't think I will have time to review this anymore, particularly after time has passed. Instead, I hope to run into the problem myself and can fix it with local reproducer.

Also, the changes in #14890 might positively affect this as well.

PR #14820, opened 2026-07-16
Own but skill files

krlvi:
Assign @krlvi as the code owner for files under crates/but/skill.


Byron:
Oh, I caused this, didn't I 😅.

PR #14822, opened 2026-07-16
feat(but): use change ID in `_commit2` output

slarse:
For the `but2` commands, we're pretending like change IDs are always persisted, which we plan for them to be shortly. Hence, change ID is not optional in this output.


slarse:
Closing, handling all `but2` commands later.

PR #14824, opened 2026-07-16
fix(pull): Synchronize configured local target

### LINKED ISSUE #14823: `but pull` should synchronize the configured local tracking branch
### Version

0.21.0

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

## Problem

`but pull` fetches and integrates the configured remote target but can leave its
local tracking branch behind. A subsequent Git worktree or agentic-coding task
created from that branch then starts from stale history despite a successful
pull.

## Scope

The change adds one optional local-target transition to the existing pull
snapshot. Generalized oplog publication and restore transactions remain
separate follow-up work.

With the configured local branch behind its upstream:

```sh
but pull
git rev-parse master 'master@{upstream}'
```

Pull reported success and updated the GitButler workspace base, but the
configured local tracking branch remained 41 commits behind.

After successful workspace integration, `but pull` should:

- resolve the unique local branch configured to track the target;
- fast-forward it only when safe;
- reject ambiguity, divergence, concurrent changes, and linked-worktree
  checkouts;
- roll the ref back if integration fails; and
- include the transition in ordinary undo and redo.

An unconfigured same-named branch must remain unchanged.

```shell
# Observed in a registered clean workspace:

$ but pull
Found 33 upstream commits
Rebase successful

$ git rev-parse master 'master@{upstream}'
7afbe90ed5fff300b991de865a575f118aa984e9
96620db4af911ceb3e1ca7c466b30f8a9844b671

$ git rev-list --left-right --count master...master@{upstream}
0	41
```

reneleonhardt:
# fix(pull): Synchronize configured local target

## 🧢 Changes

- After successful integration, fast-forward the unique safe local branch that
  tracks the target from the shared API used by CLI and GUI clients.
- Leave ambiguous, diverged, or linked-worktree branches untouched, and keep
  state consistent through failures, undo, and redo.

## ☕️ Reasoning

`but pull` could report success while leaving the local target behind its
upstream. New isolated worktrees then started from stale history, especially in
agentic coding workflows.

## 🎫 Affected issues

Fixes: #14823

## 🧪 Reproduction

```sh
but pull
git rev-list --left-right --count master...master@{upstream}
```

Before: `0 N` with `N > 0`, despite pull reporting success. After: `0 0`.

## ✅ Verification

- Focused API, CLI, failure, and undo/redo tests cover safe updates and rejected
  branch states.
- Formatting and Clippy passed; two independent reviews found no important
  issues.
- The integrated candidate passed 570 scoped tests with 14 configured skips,
  the release build, and installed smoke checks inside and outside the sandbox.

### DIFF
--- a/crates/but-api/Cargo.toml
+++ b/crates/but-api/Cargo.toml
@@ -155,7 +155,7 @@ objc2-foundation = { version = "0.3.2", default-features = false, features = ["s
 
 [dev-dependencies]
 but-api = { path = ".", features = ["legacy"] }
-but-testsupport.workspace = true
+but-testsupport = { workspace = true, features = ["sandbox-but-api"] }
 gitbutler-oplog.workspace = true
 snapbox.workspace = true
 tempfile.workspace = true
--- a/crates/but-api/src/workspace.rs
+++ b/crates/but-api/src/workspace.rs
@@ -7,6 +7,7 @@ use std::{
 };
 
 use crate::WorkspaceState;
+use anyhow::Context as _;
 use bstr::ByteSlice;
 use but_api_macros::but_api;
 use but_core::{
@@ -15,7 +16,7 @@ use but_core::{
 };
 use but_error::AnyhowContextExt as _;
 use but_forge::ForgeReview;
-use but_oplog::legacy::{OperationKind, SnapshotDetails};
+use but_oplog::legacy::{LocalTargetSnapshot, OperationKind, SnapshotDetails};
 use but_rebase::graph_rebase::mutate::RelativeTo;
 use but_serde::BStringForFrontend;
 use but_workspace::{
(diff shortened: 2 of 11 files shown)


krlvi:
we should definitely fix this, but this is not the correct layer to implement it on - it should be on the other side of the API such that other clients like the GUI would do the right thing.

reneleonhardt:
Thanks—agreed. I moved the behavior out of the CLI and into the shared
workspace integration API used by both CLI and GUI clients. The local-target
edit now joins the workspace edit in the same guarded ref transaction, so a
failure leaves both untouched; the CLI only fetches, invokes the shared
operation, and renders its outcome.

The replacement is one signed commit (+842/-33; 294 lines of focused pull
coverage). The direct shared-API test and ten CLI scenarios pass, as do
formatting and Clippy for the affected packages. I’d appreciate another look.

reneleonhardt:
One final review found an edge case that could still leave the local target and
workspace out of sync. I fixed it and added focused coverage for failures and
undo/redo.

The behavior remains in the shared workspace API as requested, with the CLI
only calling it. The updated branch is one signed commit and passes formatting,
Clippy, focused tests, and integrated smoke checks. Thanks again for the
direction.

Byron:
Thanks, but I think the idea is to wait on feedback for the issue, and based on that, create a PR. Having an issue *and* a PR for the same thing is just more work.

@krlvi #14823 was also closed, but maybe there is something I am missing. Let's hash it out over there.

PR #14830, opened 2026-07-17
docs(but): update AGENTS.md with new command structure instructions

slarse:
Instructions were a tad bit out of date

### DIFF
--- a/crates/but/AGENTS.md
+++ b/crates/but/AGENTS.md
@@ -6,16 +6,15 @@ For CLI work that touches graph/workspace/branch/stack/commit relationships,
 reachability, ordering, operation targets, or Git graph/history/ref-placement
 changes, also read `crates/WORKSPACE_MODEL.md`.
 
-## CLI I/O
+## Command structure and I/O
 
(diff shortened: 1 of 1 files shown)


crates/but/AGENTS.md
```diff
@@ -6,16 +6,14 @@ For CLI work that touches graph/workspace/branch/stack/commit relationships,
...
-  `out.for_human()` for human text, `out.for_shell()` for shell-friendly output,
-  and `out.for_json()` with `write_value(...)` for JSON.
-- Do not read `std::io::stdin()` directly in command or business logic. For
-  interactive input, gate with `out.can_prompt()` or use
-  `out.prepare_for_terminal_input()`; for piped or machine input, accept
-  `read: impl std::io::Read` so tests can inject data. Keep `stdin().lock()` at
-  top-level CLI wiring.
+## Command structure and I/O
+
+- Commands should follow the structure of `_commit2`
```
krlvi: what does this mean?

krlvi: is it about suffixing with 2 and prefixing with _ ? Is it the way it's implemented?

PR #14833, opened 2026-07-17
Lite: add GUI setting for files visible + misc

OliverJAsh:
Towards [GB-1770: Persist client-side state to local storage (e.g. files)](https://linear.app/gitbutler/issue/GB-1770/persist-client-side-state-to-local-storage-eg-files)


samhh:
Early feedback: this isn't a "setting", this is more so persisted application state. PR drafts is the one example of persisted state we currently have.

OliverJAsh:
I imagined it was a setting in the same way that other toggles are, like `diffStyle`. Where do you draw the line for toggles like this?

samhh:
It is a blurred line. I think I categorise them based upon expected UI patterns.

I was struggling to define the line. I got this from prompting and it resonates:

> The distinction is usually **meaning and ownership**, not persistence.
>
> * **Settings** express user preferences about how the app should behave: theme, default sort order, notification choices, editor font size. They are intentionally chosen, relatively stable, and often portable across devices or workspaces.
> * **Application state** records where the app currently is or what has happened: selected item, open panels, navigation history, drafts, cached data, window position, workflow progress. It changes as a consequence of normal use and is often tied to a session, document, or device.
>
> A useful test:
>
> > If the app lost this value, would the user say “my preference was reset” or “the app forgot where I was”?

If I bootstrap GitButler on a new device I'd ideally like my diff style to be remembered. Files visible is context-dependent and therefore wouldn't.

Speaking of which, I feel it makes sense that on master this state is per-project. Having files visible or not in one project has no bearing on the next, for me.

Possibly in either case we have different user expectations.

OliverJAsh:
I opened [GB-1770: Persist client-side state to local storage (e.g. files)](https://linear.app/gitbutler/issue/GB-1770/persist-client-side-state-to-local-storage-eg-files) for us to implement this properly as ephemeral but persisted application state as opposed to a setting.

PR #14834, opened 2026-07-17
Preserve empty branch ownership when reordering

estib-vega:
## Why?
In single-branch mode: Dragging an empty branch (B) under a non-empty branch (A), would result in B owning the commits from A and A becoming empty.

That's no longer the case.

## Summary

- move empty ad-hoc branch refs when a reorder crosses commit-owning segments
- preserve metadata-only reorders for empty branches that already share a commit target
- add API and Playwright regressions for `A <- B <- C (empty)` becoming `A <- C (empty) <- B`

### DIFF
--- a/crates/but-workspace/src/branch/move_branch.rs
+++ b/crates/but-workspace/src/branch/move_branch.rs
@@ -224,10 +224,11 @@ pub(super) mod function {
     ///
     /// In single-branch (ad-hoc) mode there is no workspace commit, and the tip-to-base order of
     /// branches lives in the `branch_order` metadata table rather than in workspace metadata. Empty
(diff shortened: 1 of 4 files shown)


crates/but-workspace/src/branch/move_branch.rs
```diff
@@ -292,7 +299,7 @@ pub(super) mod function {
             });
         }
 
-        if !subject_segment.commits.is_empty() {
+        if move_requires_graph_update {
```
estib-vega, line 302: This is the actual fix. The rest are tests

Caleb-T-Owens (approved):
:shipit:

PR #14844, opened 2026-07-17
Checkable commits

samhh:
Closes GB-1334.

Builds atop Olly's work on checking commits to support performing the operations (squash, move). Also adds support for shift ranges i.e. shift+click and shift+space.

The one thing left is squash message ordering, which I'll follow up in the ticket.


OliverJAsh:
We should probably at least repair/clear the checked commit state when the user explicitly deletes (discards) commits, like we do for selection. WDYT?

samhh:
Agree, that makes sense. Feel there are probably some thin abstractions we can define around state cleanup:

1. Wipe the specified operands from state anywhere they're present e.g. when deleting commits.
2. Check that all operands in state are present e.g. when the workspace updates.

Repair needs more holistic thought incl/ around identity and oplog traversal.

OliverJAsh:
This feels similar to `updateRewrittenCommitReferences` except for deletions instead of rewrites.

PR #14846, opened 2026-07-17
Moving after a mark can follow upwards as well as downwards

Caleb-T-Owens:
https://github.com/user-attachments/assets/c03115fb-40c1-4daf-9276-cfdd8d39a080

The following logic is now applied:

If I'm at a segment edge:

- If the only row above/below is now in the opposite state, we move the cursor there
- If the only other row is now the same state, we don't move.

If I'm in the middle:

- If the cursor's row as turned into the same state as it's neighbours, it remains the same.
- If the cursor's row has turned into the opposite state as BOTH it's neighbours, it goes down
- If the cursor's row has turned into the same as one of it's neighbours, it goes to the other neighbour


davidpdrsn (approved):
Awesome!

PR #14847, opened 2026-07-17
performance(but): avoid rebuilding entire IdMap to get change ID

slarse:
Changes to `theme.rs` introduced a relatively significant performance hit to many commands when formatting output for humans. They'd recompute the entire `IdMap` from context, which entails rebuilding the workspace projection. I measured  a 5-8% performance hit to `but commit` from this computation, which is too much for just output formatting.

This commit somewhat reverts that by reusing pre-mutation change IDs when available, otherwise trying to fetch from an existing `IdMap`, or if that is not available fetching the commit from the repo.

For the most part, we're trusting that change IDs stay the same. This is overall not well structured and we should take more care with the `but2` commands, which I've left unimplemented and deferred for GB-1771.

### DIFF
--- a/crates/but/tests/but/command/move.rs
+++ b/crates/but/tests/but/command/move.rs
@@ -40,7 +40,7 @@ fn move_commit_before_another_commit() -> anyhow::Result<()> {
         .assert()
         .success()
         .stdout_eq(str![[r#"
-Moved 1#2 → before 1#1
+Moved 1 → before 1
 
 "#]]);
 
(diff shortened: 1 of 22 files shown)


crates/but/tests/but/command/move.rs
```diff
@@ -220,7 +220,7 @@ Hint: run `but help` for all commands
     .assert()
     .success()
     .stdout_eq(str![[r#"
-Moved 2 commits → after 1#2
+Moved 2 commits → after 1
```
slarse, line 223: Some of these outputs turned a bit weird as they contain collisions, and we're not resolving from the `IdMap` here. It's mostly fine because collisions aren't that likely, but it's a topic for improvement.

PR #14855, opened 2026-07-17
Lite: improved uncommitted files

OliverJAsh:
Towards [GB-1732: Show uncommitted files in outline panel aka sidebar](https://linear.app/gitbutler/issue/GB-1732/show-uncommitted-files-in-outline-panel-aka-sidebar)

@samhh This PR should be in a good place, but I'm keen for a thorough code review.

There's a bunch of small things to think about still (see below) but I don't think there are any blockers, so if it looks good then please merge in my absence and I'll get back on it when I return. Cheers!

For when I return:
- [ ] selection scope focus indicator e.g. when empty
- [ ] bug: on click, flash on focus change (pointer down) before selection changes (pointer up) - not sure how we can avoid this
- [ ] restore ability to absorb/cut all uncommitted files, either via replacement shortcuts OR multi-selection
- [ ] clicking on header should forward focus to tree OR widen scope of tree element
- [ ] decouple headers from `Row` (`Row` is designed for selectable rows, not headers)
- [ ] rename outline tree to stacks tree
- [ ] rename outline mode to ??
- [ ] rename outline panel to ??
- [x] can we remove commit targets now?
- [ ] can we refine selection types? `Operand` is much wider than we need now
- [ ] all-in-one diff (@samhh probably best you do this one)
- [ ] drag area includes commit form
- [ ] uncommitted file as active source: outline is clipped due to scroll area

### DIFF
--- a/apps/lite/ui/src/hotkeys.ts
+++ b/apps/lite/ui/src/hotkeys.ts
@@ -139,10 +139,6 @@ export const workspaceHotkeys = {
 } satisfies Record<string, HotkeyWithMeta>;
 
 export const outlineHotkeys = {
-	absorb: {
-		hotkey: "A",
-		meta: { group: "Uncommitted changes", name: "Absorb" },
-	},
 	amendCommit: {
 		hotkey: "Shift+A",
 		meta: { group: "Commit", name: "Amend commit" },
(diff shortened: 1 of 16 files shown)


samhh (approved):
The way we currently defer the details panel is a bit more noticeable now. Another small one for the list!

PR #14874, opened 2026-07-19
Fold sidebar button update

PavelLaptev:
- [simplify icons design](https://github.com/gitbutlerapp/gitbutler/commit/5a4b6b653344bf7e31e77ed5acaf17388ae0b3d3)
- [Avoids "double signaling" for the outline toggle](https://github.com/gitbutlerapp/gitbutler/commit/5e6a7dd76e93817836e37097f9fc83af5469e790). A Toggle gets pressed styling on top of the changing icon, which creates double signaling. In this pattern, only the icon should change, the button's styling (e.g., its background) should stay the same. While "hide sidebar" is technically a toggle, it behaves more like an action button and sits apart from other toggle-style controls.

https://github.com/user-attachments/assets/af4cd28e-6d8b-44aa-bc3c-8eee42a7d0ee

---

As a reference for the icons I took the Obsidian app:

https://github.com/user-attachments/assets/652bdbbb-d36d-4a0e-afb9-b8e369af0539


PavelLaptev:
@OliverJAsh We discussed whether this button should have all the attributes of a toggle button, like changing its background when toggled. I don't think that's needed for the sidebar icon, it works fine without it.

PR #14884, opened 2026-07-20
Add cherry pick to workspace crate

samhh:
One small step towards GB-1535. but-api next?

### DIFF
--- a/crates/but-workspace/src/commit/cherry_pick.rs
+++ b/crates/but-workspace/src/commit/cherry_pick.rs
@@ -1,0 +1,56 @@
+//! Cherry-pick commits into a workspace graph.
+
+use anyhow::bail;
+use but_core::{RefMetadata, commit::Headers};
+use but_rebase::commit::DateMode;
+use but_rebase::graph_rebase::{
+    Editor, Selector, Step, SuccessfulRebase, ToCommitSelector, ToSelector as _,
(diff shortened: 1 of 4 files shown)


estib-vega (approved):

crates/but-workspace/src/commit/cherry_pick.rs
```diff
@@ -0,0 +1,56 @@
...
+};
+
+/// Cherry-pick commits above or below a commit, or below a reference, in the
+/// workspace graph.
+///
+/// Sources are deduplicated and ordered parent-first. Child commits, and the
+/// target commit, if applicable, are rebased atop the cherry-picked commits.
+pub fn cherry_pick_commits<'ws, 'meta, M: RefMetadata>(
+    mut editor: Editor<'ws, 'meta, M>,
+    source_commit_ids: impl IntoIterator<Item = gix::ObjectId>,
```
Caleb-T-Owens: To aid composability, could this function take an iterator of `ToCommitSelector`

An `Editor` after `.rebase()` is called returns a `SuccessfulRebase` and some other selectors. The idea is that you can then turn a `SuccessfulRebase` back into an `Editor` and then make use of that and the returned selectors to call into another function.

The but-transaction crate is often making use of this in the CLI and TUI

Caleb-T-Owens (approved):
Danke!

PR #14917, opened 2026-07-21
Fix unexpected vertical auto-scroll-on-drag of whole MultiStackView

cbjeukendrup:
rather than only horizontal scrolling in MultiStackView and vertical in individual stack views.

Behaviour without this PR:

https://github.com/user-attachments/assets/9da38150-ca85-4823-ba16-6d26e0ad2e63


krlvi:
Thanks! It appears to fix the issue! <3

PR #14932, opened 2026-07-22
Repair rather than derive stale state

samhh:
Prep for checked files in a roundabout way, see commits. Substantially improves commit selection stability (e.g. undo), towards GB-1723. Arguably simplifies the mental model, but does not fully embrace this new approach yet. We don't currently bother correlating commits except in selections, though we could; there are probably some abstractions waiting for us here somewhere.


samhh:
Very keen to discuss and get feedback @OliverJAsh. I'm merging now since it unblocks file checking and almost all the code is new and essentially isolated so it won't be too hard to revert if we want to go another way.

PR #14936, opened 2026-07-22
Make CLI ignore assignments

davidpdrsn:
We've been talking about removing assignment support in the CLI for a while and
now its time to make it happend!

The end goal I'm working towards is that the CLI ignores stack assignments and
simply treats all changes as uncommitted, like Lite does. So if you assign
something in Tauri and then look at `but status` things wont be assigned.

I do this by adding a new type called `WorktreeHunk` which is like
`HunkAssignment` except it doesn't have `stack_id` and `branch_ref_bytes`. I
then change `IdMap` to use that type instead of `HunkAssignment`. I then update
all the code to not look at those fields.

I also

- Remove `but stage`.
- Remove `but unstage`.
- Remove assignment related routes from `but rub`.
- Remove the `--only` flag from `but commit`.

I focused on getting things to build. There is likely lots of additional
refactorings that we can do but that can wait until this has landed.

I also feel like `CliId::Stack` should be removed but we can discuss that. I
didn't change it here.

### Review

This PR doesn't pass CI yet but it should be in a reviewable state. I would
appreciate getting a second pair of eyes before I start fixing the tests.

I recommend reviewing commit by commit and focusing on the first two commits
that touch the `IdMap` and `DiffSpecBuilder`. All other changes are resulting
from those and are very mechanical.

All changes are done by hand and not using a clanker.


krlvi:
all 3 points here sound reasonable

krlvi:
skimming over this - looks reasonable

crates/but/src/id/mod.rs
```diff
@@ -1620,20 +1665,17 @@ impl CliId {
 pub struct UncommittedFile {
     /// The shortest ID that can be used to unambiguously refer to this file.
     pub short_id: ShortId,
-    /// Every element has the same [HunkAssignment::stack_id] and [HunkAssignment::path_bytes],
-    /// so the first assignment can be used to obtain both.
-    short_id_hunk_assignments: NonEmpty<(UnqualifiedHunkId, HunkAssignment)>,
+    /// Every element has the same [`HunkAssignment::path_bytes`] so the first assignment can be used
+    /// to obtain both.
```
jonathantanmy2: s/both/it/
since there is no more `stack_id`

crates/but/src/id/parser.rs
```diff
@@ -189,20 +189,14 @@ fn try_parse_range(
...
 }
 
-fn get_all_files_in_display_order(ctx: &mut Context, id_map: &IdMap) -> anyhow::Result<Vec<CliId>> {
+fn get_all_files_in_display_order(id_map: &IdMap) -> anyhow::Result<Vec<CliId>> {
     // First, files assigned to branches (they appear first in status display),
     // then uncommitted files (they appear last in status display)
-    let (_guard, _, workspace, _) = ctx.workspace_and_db()?;
-    let stack_ids: Vec<Option<but_core::Id<'S'>>> =
-        workspace.stacks.iter().map(|stack| stack.id).collect();
     let mut positioned_files: Vec<(usize, &BStr, CliId)> = id_map
```
jonathantanmy2: The first field (the `usize`) is no longer necessary since it was used only to sort by stack (but we can clean that up after the PR, so it's not urgent).

crates/but/src/id/uncommitted_info.rs
```diff
@@ -1,15 +1,14 @@
...
 use bstr::BString;
-use but_hunk_assignment::HunkAssignment;
 use nonempty::NonEmpty;
 
-use crate::id::id_usage::UintId;
+use crate::id::{WorktreeHunk, id_usage::UintId};
 
 /// Information about uncommitted files.
 pub(crate) struct UncommittedInfo {
     /// Uncommitted hunks partitioned by branch assignment and filename.
```
jonathantanmy2: partitioned by filename (delete "branch assignment and")

crates/but/src/id/uncommitted_info.rs
```diff
@@ -23,24 +22,20 @@ type PartitionKey = (Option<BString>, BString);
 impl UncommittedInfo {
     /// Partitions hunk assignments by branch assignment and filename.
```
jonathantanmy2: same as above

guillenotfound:
I think removing some of this is of course a breaking change as you mentioned in the release notes. Is there a more in depth rationale about these changes?

Not judging them by any means, but it breaks the workflows I was used to, so now if I want to pre-stage things I don't think that's possible anymore.

On the other hand this might open a door, I use `git add -p` or vim-figitive all the time, using butler breaks these flows specially if I'm working on different things. Would it make sense to only commit staged changes if there are otherwise you are required to pass the hunks/files that you want to commit as per the new flow?

Probably this is not the place for this discussion anyways so I'll be happy to join Discord? If that's a better place to have it 😊

davidpdrsn:
> Probably this is not the place for this discussion anyways so I'll be happy to join Discord? If that's a better place to have it 😊

Yes, please join our Discord if you want to talk more about it, but I can provide an explanation here.

> On the other hand this might open a door, I use `git add -p` or vim-figitive all the time, using butler breaks these flows specially if I'm working on different things.

This was one of the primary reasons we removed support for assignments: it doesn't work with third-party tools. Git fundamentally cannot stage to different branches, so GitButler tracked assignments separately, which didn't work with other tools that relied on the Git index.

Of course tools like vim-figitive still doesn't know how to commit to GitButlers parallel branches but thats a problem regardless. I personally use `but tui` for that.

> Would it make sense to only commit staged changes if there are otherwise you are required to pass the hunks/files that you want to commit as per the new flow?

We think a flow where you directly commit the changes you want is easier. With GitButler, you don't need to first stage hunks and then commit. You can just commit the hunks directly. You can do this via the CLI with `but commit hunk-a hunk-b ...`, with `but commit --interactive`, or in `but tui`.

The commits are also easy to change afterwards with `but squash/reword/amend`, etc. You can essentially treat commits in a more fluid and mutable way which makes staging unnecessary.

PR #14946, opened 2026-07-23
feat(cli): add skill for writing new cli commands

davidpdrsn:
Tested a bunch with this prompt

> Create a new `but overview` command that prints:
> 
> For each applied branch it prints
> - The number of files changes, lines added, and lines removed on the branch. Use our existing branch diff apis
> - How many commits are on the branch
> 
> Also take a list of branch arguments. If empty show overview for all branches. If non-empty show overview only for branches given. Error if called with an unapplied branch
> 
> Support all output formats
> 
> Also write tests
> 
> I'm gonna go walk the dog so you can't ask me questions. Use your best judgement and keep going until its done

Makes sol xhigh generate decent results that can easily be shaped.

### DIFF
--- a/.agents/skills/cli-commands/SKILL.md
+++ b/.agents/skills/cli-commands/SKILL.md
@@ -1,0 +1,239 @@
+---
+name: cli-commands
+description: Use when adding or modifying CLI command (`but` commands) under `crates/but/src`.
+---
+
+Imagine we're implementing a new `commit3` command. The high level structure
+for that must be as follows.
+
+## Arguments
+
+Arguments live in `crates/but/src/args`. For our commit command that would be
+`crates/but/src/args/commit3.rs`:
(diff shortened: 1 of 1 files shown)


slarse (approved):
Looks good! This overlaps a little with the changes I made to `crates/but/AGENTS.md`, maybe remove the stuff I added there and just point to this skill?

davidpdrsn:
> Looks good! This overlaps a little with the changes I made to `crates/but/AGENTS.md`, maybe remove the stuff I added there and just point to this skill?

Yeah maybe. I'll look into that.

krlvi:
Nice! Also mentioning that this skill exists and should be considered inside the AGENTS.md is probably a good idea. And keeping this in a skill means we dont fill up the context until needed.

PR #14952, opened 2026-07-24
fix(cli): fix squashing into branch stacked below empty branch

davidpdrsn:


crates/but/src/command/legacy/squash2.rs
```diff
@@ -670,7 +670,12 @@ pub fn resolve_target(
                         continue;
                     };
                     if ref_info.ref_name == branch_name {
-                        return if let Some(commit) = ref_info.commit_id {
+                        let commit = segment
+                            .commits
+                            .first()
+                            .map(|commit| commit.id)
+                            .or(ref_info.commit_id);
```
davidpdrsn, line 677: Feels like this might be a bug in the graph. The branch is not empty yet `ref_info.commit` is `None` 🤔 I'm gonna try and write a more focused test case and send to someone.

PR #14968, opened 2026-07-24
refactor(lite): replace openInEditor with openInProgram

slarse:
Adds new `openInProgram` and `listPrograms` functions to the Lite IPC bridge.

Some things to note:

* `openInProgram` replaces `openInEditor`.
* `listEditors` is still used - an editor _is_ a program with category "editor"
* `listPrograms` is as of yet unused, but it's there if you want it to be able to list every category of app
* `openInProgram` does not _fully_ unlock the underlying API, as that can also open multiple files with a single invocation. If we feel we need that for lite, we'll need a slightly more drastic change to "what to open" is passed in. So I didn't do it in this initial refactor.


samhh (approved):
> `openInProgram` does not fully unlock the underlying API, as that can also open multiple files with a single invocation. If we feel we need that for lite, we'll need a slightly more drastic change to "what to open" is passed in. So I didn't do it in this initial refactor.

That would be helpful for multiselection/checking e.g. I'm reviewing some code and want to more closely inspect n files in my editor without having to invoke the action n times.

PR #14976, opened 2026-07-24
perf(lite): stop rebuilding the whole diff view on every render

krlvi:
`getDiffView` walks every file and hunk of a changeset to build the code view items, the hunk navigation index and the lookup maps. It ran on every render of the Diff component, so merely moving between files in the list rebuilt the model for all of them.

Building it inside the query's `combine` gives the result a stable identity: react-query caches it on the results and the combine reference, and the compiler memoizes the inline closure by its captures. The closure captures the changes and the outline selection, both stable across a file selection, so `getDiffView` now runs when the changeset changes rather than on every pass.

`changesetKey` and `fileParent` move inside it for that reason — derived in render they were rebuilt every time, which would have defeated the memoization.

Selecting a branch, whose diff spans the whole branch, is where this was worst. Switching files measured 64-78ms per click and now measures 10-13ms, unchanged whether the branch list holds one row or seven hundred.

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/Details.tsx
+++ b/apps/lite/ui/src/routes/project/$id/workspace/Details.tsx
@@ -198,11 +198,34 @@ const mkCodeViewItem = (
 	};
 };
 
+const changesetKeyOf = (outlineSelection: Operand): string =>
+	Match.value(outlineSelection).pipe(
+		Match.tags({
+			Branch: ({ branchRef }) => decodeBytes(branchRef),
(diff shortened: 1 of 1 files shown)


krlvi:
Hmm, @samhh im not sure if we actually want this. im gonna close

samhh:
I need to think about if this is the best way to solve this but you've identified a real performance issue. The difference is visible on a large branch if you hold down shift+arrow to rapidly navigate files.

PR #15028, opened 2026-07-27
Diff annotations w/ prompt copying

samhh:
This is a _super early experimental_ version of something I wondered about over the weekend. I'd like to ship early and see how it feels before spending any longer on it. If this doesn't work out then on the plus side it's given me some exposure to how Pierre handles annotations.

In a nutshell, instead of hopping between your agent and GitButler and having to manually describe where your feedback relates to, you can type your targeted feedback in GitButler instead and copy a prompt with all of that information captured.

Screenshot 2026-07-27 at 17 03 31

Here's how the prompts currently look:

```md
# Feedback

1. apps/lite/ui/src/annotation.ts:15 (additions) in refs/heads/localannos

Here is my first piece of feedback.

2. apps/lite/ui/src/routes/project/$id/workspace/Details.tsx:217 (additions) in refs/heads/localannos

Here is my second piece of feedback.

I sure am waffling on.
```

### DIFF
--- a/apps/lite/ui/src/annotation.ts
+++ b/apps/lite/ui/src/annotation.ts
@@ -1,0 +1,103 @@
+import { decodeBytes } from "#ui/api/bytes.ts";
+import type { FileParent } from "#ui/operands.ts";
+import type { AnnotationSide } from "@pierre/diffs";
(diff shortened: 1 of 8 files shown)


krlvi:
:shipit:

PR #15071, opened 2026-07-29
Speed up Lite comment creation

krlvi:
Keep parsed diff data stable when comments change and insert newly created comments directly into the query cache.

@samhh wdyt?

### DIFF
--- a/apps/lite/ui/src/annotation.ts
+++ b/apps/lite/ui/src/annotation.ts
@@ -3,7 +3,7 @@ import type {
 	CommentCreateParams,
 	CommentUpdateParams,
 } from "#electron/ipc.ts";
-import type { QueryKey } from "#ui/api/queries.ts";
+import { commentsQueryOptions, type QueryKey } from "#ui/api/queries.ts";
 import { decodeBytes } from "#ui/api/bytes.ts";
 import { errorMessageForToast } from "#ui/errors.ts";
 import type { FileParent } from "#ui/operands.ts";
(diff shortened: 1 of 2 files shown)


samhh (changes requested):
Ah, I hadn't noticed this before working on #15075!

I've given that other branch a little extra love. If you compare the two locally it's noticeably snappier by implementing optimistic updates.

Agent comparison:

---

Area | `lite-faster-comments` | `sah-branch-6` |
|---|---|---|
| Comment creation | Updates cache after IPC succeeds | Inserts immediately before IPC completes |
| Focus timing | After backend response | Immediate, using a client-generated UUID |
| Backend changes | None | Adds client-supplied IDs through Rust and generated SDK |
| Archive | Waits, then refetches | Optimistically removes, then reconciles |
| Diff construction | TanStack `combine` | Explicit `useMemo` |
| Size | 2 files, +49/−37 | 6 files, +201/−137 |

The React Compiler output confirms both implementations preserve the expensive base diff across comment updates:

- Lite caches `getDiffView()` through `useSuspenseQueries.combine`.
- Sah caches it through `useMemo`.
- Both separately memoize `withAnnotations()` and preserve unannotated item identities.

So the main difference is perceived latency, not base-diff rebuilding. Sah is genuinely optimistic; Lite merely eliminates the follow-up comment-list request.

[...]

If choosing one branch unchanged, I’d choose `sah-branch-6`: it delivers the actual immediate-create behavior and includes the necessary backend contract. `lite-faster-comments` is smaller, but materially less optimistic.

PR #15075, opened 2026-07-29
Snappy annotations

samhh:
Feels snappy like the non-SDK version used to.

For review: I tweaked the comment API in d21767d to accept a client-supplied ID in order to enable optimistic creation.


krlvi (approved):
:shipit:

PR #15085, opened 2026-07-29
discard commit: allow workdir-target conflicts

jonathantanmy2:
It's more complicated than I was hoping - there were some things that need to be plumbed, and the plumbing was longer than expected.

If this is too big for reviewers, I could look into doing some refactorings first to make the actual change simpler. I already did some refactoring PRs, but perhaps not enough.

---

The sides of the conflicts will be written to the index.

An earlier version of this commit reverted some of the changes in
https://github.com/gitbutlerapp/gitbutler/commit/9547b84b99f2adb10700e1dfcc80a91944d2456e (revive `but worktree` subcommands, 2026-06-11) (which
changed some tests because at that time, GitButler didn't support
checkout causing workdir-target conflicts) but this commit no longer
does so as allowing workdir-target conflicts is no longer on for all
commands by default, but must be configured individually (I did it
this way because there is no generic way to report to the user that a
conflict has occurred).

This currently only works when discarding commits (not other things).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3831,7 +3831,7 @@ dependencies = [
 [[package]]
 name = "git2"
 version = "0.21.0"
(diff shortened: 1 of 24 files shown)


jonathantanmy2:
Superseded by #15232

PR #15096, opened 2026-07-30
Fix diff metrics regression

samhh:
Fixes regression in #15086.

cc @PavelLaptev We need to check that these numbers line up whenever we tweak the styling of the diff. I can explain how this works over a call if that helps. 🙂


PavelLaptev:
Overlooked this, sorry

PR #15103, opened 2026-07-30
Switch between modes in the TUI

davidpdrsn:
Allows switching between commit, squash, and move mode without first going to normal mode.


krlvi:
stacking works :)

davidpdrsn:
> stacking works :)

It does! Its so nice :D

PR #15107, opened 2026-07-30
Handle narrow sidebar

PavelLaptev:
### Summary

The workspace sidebar (outline panel) broke down at narrow widths: the commit button overflowed, and branch row meta items (status text, PR link, Push button) collided instead of adapting. This PR makes the sidebar degrade gracefully when resized down.

### Changes

**Commit button adapts to narrow panels**
- The commit form is now a CSS container; below 300px the "Commit" label is hidden, leaving the icon/kbd only.
- When the label is hidden, a tooltip shows "Commit" (with the hotkey) on hover — the tooltip is suppressed while the label is visible.

**Branch row push status truncates with ellipsis**
- Status text ("Nothing to push", "Some unpushed", …) now truncates with an ellipsis instead of overflowing into the PR link and Push button.
- Meta items no longer shrink (`flex-shrink: 0`); only the status text yields space via a dedicated shrinkable item, so buttons and links keep their full size.

**Outline panel sizing**
- Min width raised from 240px → 260px, default width lowered from 500px → 420px.

### Testing

- `pnpm isgood` passes; typecheck clean.
- Manually resized the sidebar to min width and verified: label hides at the breakpoint, tooltip appears only then, and branch rows truncate without overlap.

### Screenshots

https://github.com/user-attachments/assets/612f71ef-1213-41b8-8a1b-69ec90779813

### What's left

Screenshot 2026-07-31 at 00-02-37

I hold off on this bug. Once Kiril unlocks the 'Upstream' tab, we can properly update the header design. Should be soon.

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/CommitForm.tsx
+++ b/apps/lite/ui/src/routes/project/$id/workspace/CommitForm.tsx
@@ -80,6 +80,7 @@ export const CommitForm: FC<{
 
 	const commitTextareaRef = useRef<HTMLTextAreaElement | null>(null);
 	const formRef = useRef<HTMLFormElement | null>(null);
+	const commitButtonLabelRef = useRef<HTMLSpanElement | null>(null);
 
 	const { data: draftMessage } = useQuery(draftCommitMessageQueryOptions(projectId));
 	const { mutate: persistDraftMessage } = usePersistDraftCommitMessage();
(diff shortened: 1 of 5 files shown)


PavelLaptev:
cc @samhh

apps/lite/ui/src/routes/project/$id/workspace/CommitForm.tsx
```diff
@@ -354,15 +371,29 @@ export const CommitForm: FC<{
...
 
 					<div className={styles.dropdownButton}>
-						<Button
-							className={getButtonClassName({ variant: "pop" })}
-							focusableWhenDisabled
-							type="submit"
-							disabled={!canCommit}
+						<Tooltip.Root
+							// Only show the tooltip when the container query hides the label.
+							open={commitTooltipOpen && commitLabelHidden}
```
samhh, line 376: I was investigating the use of a resize observer and `getComputedStyle`, which is a code smell (perf).

I might be having a Friday moment but I don't think this tooltip can ever actually be visible...

PavelLaptev: Thanks for flagging this! Looks like I missed it. I tested it earlier and it was working, but I’ll make sure to fix it in the next PR.

PR #15108, opened 2026-07-30
Add Create PR / Land actions to branch context menu

cbjeukendrup:
Expose whichever action (Create PR or Land) is hidden by the
'Land branches directly' toggle as a context menu item, so both
remain reachable regardless of the setting.

Scherm­afbeelding 2026-07-31 om 01 13 03


krlvi:
Seems reasonable.

PR #15156, opened 2026-08-04
feat: ✨ expose GitButler GUI and CLI as flake packages

robertodr:
## 🧢 Changes

I've changed `flake.nix` such that it can be ingested in other flakes to provide `gitbutler` and `but`. I got help from GPT-5.6 Sol.

## ☕️ Reasoning

I use NixOS and the official derivation for `gitbutler` is a few version behind and `but` is not even available. Ingesting them directly from an official project flake would help me keep up with newer releases.

Running:

```
NIXPKGS_ALLOW_UNFREE=1 nix build --impure .#gitbutler

NIXPKGS_ALLOW_UNFREE=1 nix build --impure .#but
```

does produce the expected working output.

### DIFF
--- a/NIX.md
+++ b/NIX.md
@@ -1,0 +1,30 @@
+# Installing GitButler with Nix
+
+GitButler can be consumed as a flake input:
+
+```nix
+{
+  inputs.gitbutler.url = "github:gitbutlerapp/gitbutler";
+
+  outputs = {nixpkgs, gitbutler, ...}: let
+    system = "x86_64-linux";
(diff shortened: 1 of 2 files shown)


robertodr:
Closing, as the `llm-agents.nix` flakes already provided up-to-date versions of both the [GUI](https://github.com/numtide/llm-agents.nix/blob/main/packages/gitbutler/package.nix) and the [CLI](https://github.com/numtide/llm-agents.nix/blob/main/packages/but/package.nix)

PR #15175, opened 2026-08-04
rebase: add options parameter to `materialize()`

jonathantanmy2:
The changes touch many files, but besides the change to `crates/but-rebase/src/graph_rebase/materialize.rs`, the changes are trivial, so hopefully this is easy to review.

---

In the future, it will need another parameter that controls how conflicts during checkouts are reported. Therefore, perform a preliminary refactor to add an options struct parameter.


Caleb-T-Owens (approved):
Nice!

PR #15212, opened 2026-08-06
Flesh out forge support in lite

mtsgrd:
The PR tab in lite grows up from a title/description form into an actual pull request view, closely following the Figma design.

Fittingly, this description uses the same GitHub-flavored markdown the branch teaches lite to render — the screenshot below is lite drawing a table just like the one here.

## The tour

| Area | What you get |
| --- | --- |
| Metadata panel | Status badge, author, reviewers with verdicts, labels, checks, branches, created/updated — a bordered card like the design |
| Markdown | Rendered descriptions + comments, shiki syntax highlighting, edit mode with Cancel |
| Activity | Comments, review submissions, pushed commits, and review requests interleaved with connector dashes, plus a composer. Edit or delete your own comments, quote-reply, and long content folds behind *Show more* |
| Reactions | Chips on the PR and its comments with who-reacted tooltips; toggle your own or add from an animated emoji popover |
| Merging | Method picker (remembered per project), working auto-merge toggle, a tooltip explaining *why* merge is blocked, Close under a kebab |
| Management | Add/remove labels and request/withdraw reviewers from the panel's `+` buttons |
| Freshness | Polls while focused, sleeps in the background, catches up on refocus. Posting, reaction toggles, and the auto-merge switch apply optimistically — instant UI, clean rollback on failure, and the composer draft survives a failed post |

The markdown gets GitHub-parity sanitization (rehype-sanitize's GitHub-modeled schema): safe HTML like the `<details>` block below folds in lite exactly like it does here, while scripts, styles, event handlers, and unsafe URLs are stripped — all pinned by tests. Images inline when GitHub hosts them (same privacy story as github.com — GitHub's image CDNs don't leak viewer requests to authors); any other host renders as a link and is never fetched.

<details>
<summary>Under the hood</summary>

New forge endpoints end to end (GitHub client → but-forge → but-api → SDK → electron IPC): comments list/create/edit/delete, review submissions, issue-timeline events, reaction reads and writes (kinds are an open set of forge-native names, ready for GitLab's arbitrary award emoji), repo labels, assignable users, label add/remove, review request/withdraw, a current-login lookup so the UI knows which comments are yours, plus auto-merge state on `ForgeReview` with a DB migration and struct-version bump (the e2e review-seeding fixture follows the bump). GitLab/Bitbucket read as empty for the new listings instead of erroring; writes are GitHub-only for now, gated through new `ForgeCapabilities` entries rather than forge-name checks.

The PR view's pieces live in their own modules now (`PullRequestTab`, `Panel`, `Comments`, `Reactions`), and every relative timestamp carries its absolute time in a tooltip — rendered by the app's own tooltip component, since native `title` tooltips never show in the Electron window.

</details>

image

Left for follow-up branches:

- [ ] Fire the remaining write paths at a real PR (merge, labels, reviewer requests — comments and reactions have been exercised live)
- [ ] GitLab parity for the new listings


krlvi (approved):
LGTM

Issue #15209, opened 2026-08-06
`but commit` never runs pre-commit / commit-msg / post-commit hooks (no call site in the CLI)


artile:
### Summary

`but commit` never runs the `pre-commit`, `commit-msg` or `post-commit` hooks. Not "runs them and ignores the result" — the CLI has no call site for them at all. For a repository whose quality gates live in a pre-commit hook, this means committing through `but` silently skips every check that committing through git enforces.

Please run the commit-time hooks from the CLI, with a `--no-hooks` / `--no-verify` opt-out, the way `but push` and `but pr new` already do.

### Evidence

Measured on `but 0.22.0` (installed via `brew install --cask gitbutler`), on a repository that sets `core.hooksPath=.githooks`.

The hook was instrumented to append to a log file, regenerate a docs index and `git add` it — the shape of a real gate hook:

- `but commit -b exp-hooks2 -m "..." wm` → `Created commit vtp on new branch 'exp-hooks2'`. The log file was **never created**. The resulting commit contains one file; the file the hook stages is absent.
- Control, same hook, same repository, `git commit` → the log file is written and `git diff --cached --name-only` shows both the edited file and the file the hook staged.

The source agrees. `pre_commit_with_tree()` in `crates/gitbutler-repo/src/hooks.rs` has exactly one caller — `crates/but-api/src/legacy/repo.rs`, the desktop application's API. Searching `crates/but/src` for a pre-commit call site returns nothing. `commit_msg()` and `post_commit()` in the same file are reachable only from that same desktop path, so all three are CLI-dead. `--no-hooks` exists on `but push` and `but pr new`, but there is no flag on `but commit` to turn hooks on, because there is nothing to turn on.

### Second, smaller problem — worth fixing in the same change

Even the desktop path discards what a hook stages. `pre_commit_with_tree()` backs up the index file, replaces it with the commit's tree, runs the hook, and then restores the backup with a plain rename:

```rust
std::fs::rename(backup_path, index_path).context("failed to restore pre-commit index")?;
```

Any `git add` the hook performed is inside the index that just got overwritten. This is a common and load-bearing hook pattern — a formatter, a lockfile updater, or a generated-index regenerator stages its output so that it lands in the same commit. Wiring the CLI up to the current implementation would produce a hook that runs but whose output silently vanishes, which is arguably worse than not running it. Reconciling the hook's index writes back into the commit's change set would make the feature actually usable.

### Why this matters for agent workflows

[...]

PR #15222, opened 2026-08-06
Restore commit hooks in the CLI

ameyypawar:
## 🧠 Changes

`but commit` runs the repository's `pre-commit` hook before writing the commit and `post-commit` after, with `-n`/`--no-hooks` (alias `--no-verify`) to bypass — restoring what `45a8387f40` ("feat(but): replace old commit command with commit2") dropped. The flag declaration, the failure wording and the "post-commit failures don't undo the commit" rule are taken from the code commit2 replaced. The TUI's commit runs them too.

A hook that rewrites files needs care. `DiffSpec` documents that hunk headers failing to match the worktree are dropped, and `push_changes_from_uncommitted_area()` expands every change into hunks — so even `but commit -m` with no ids carries headers a hook can invalidate, and committing the specs computed beforehand would carry less than was asked for without saying so. So:

- with nothing singled out, the specs for **those same paths** are rebuilt from the worktree as the hook left it;
- when files or hunks were singled out, the commit is refused and the worktree is left untouched;
- only specs carrying hunk headers are watched, since additions, deletions, symlinks, binaries and too-large files are read whole while the commit is built and cannot go stale.

`retired_syntax.rs` listed `--no-hooks` among the retired forms with no modern equivalent, so `but commit -c -m "msg" --no-hooks` was refused rather than translated. Both grammars spell it identically, so it now carries over.

Separately, `join_output` tested `stderr.is_ascii()` where it meant `is_empty()`, so every ASCII-only hook message was discarded and reported as "hook produced no output". That one also affects the desktop and `but push`.

## ☕️ Reasoning

Before commit2, `but commit` ran all three commit hooks. The rewrite carried none across, so a repository whose gates live in `pre-commit` has had them silently skipped since 24 July — and the CLI is the surface the bundled agent skill directs agents to in preference to `git`.

PR #15139 documents the invariant a hook breaks: the changes handed to `create_commit` "must describe the pre-commit state — in practice, run this before any other operation that could change what the worktree is based on". A `pre-commit` hook is exactly such an operation, running between the specs being computed and the commit being made. Rebuilding restores that invariant; refusing is the answer when rebuilding would overrule a choice the user made.

On parity with `git`: there isn't any, and I would rather say so than imply it. `git` commits the index, so a hook's edits land only if the hook stages them. GitButler has no staging, and nothing here can tell a staging hook from one that merely writes, so the rule chosen is "the files already going into this commit, as the hook left them" — narrower than committing whatever the hook touched, wider than ignoring it. If you would rather it refuse in both cases, that is a smaller change than this one.

`commit-msg` is not restored here. commit2 creates the commit and *then* rewords it, because the editor shows the commit's own diff, and that reword runs inside a transaction that owns the `Context` which `hooks::commit_msg` needs. Restoring it means moving message resolution ahead of commit creation, and deciding what the hook is handed — a message read back from the commit has no trailing newline, so the common `echo ... >> "$1"` sign-off hook welds itself onto the subject line. That belongs in its own change.

Hooks default on, matching `but push` and `git`. The desktop's `runCommitHooks` defaults to false, but it is frontend-only `localStorage` with no Rust counterpart, so the CLI cannot read it; worth knowing the two surfaces differ.

### A note on the test isolation fix

`e71c886798` pins `core.hooksPath` per test repository. libgit2 only honours `GIT_CONFIG_NOSYSTEM`/`GIT_CONFIG_GLOBAL` when a repository is opened `FROM_ENV`, which `but` never does, so the sandbox's config isolation does not reach any hook path. On a machine with `core.hooksPath` set globally, the hook these tests write was skipped and the developer's own hook ran instead — I reproduced both halves. That is worth knowing beyond this PR: every libgit2-read config key escapes the sandbox today, and the general fix belongs in `but-testsupport`. Happy to send that separately. The failure mode was pointed out by @sensei-woo.

## 📌 Todos

Known and deliberately out of scope. None of these are introduced by this change - they are
pre-existing, and several are live in the desktop today. Listed so they are not mistaken for
oversights, and I am happy to take any of them on separately:

- **The desktop still has the underlying defect.** It runs the same pre-hook specs through `pre_commit_hook_diffspecs` and then commits those same pre-hook changes, so a hook that reformats a file still drops the hunks it moved. This PR fixes the CLI only, which means the two surfaces now behave differently — arguably the more valuable follow-up, and it belongs in `but_core` rather than here.
- Two hazards this makes reachable by default again, both predating commit2 and both live in the desktop: a hook that itself invokes `but` deadlocks, because `git` sets `GIT_INDEX_FILE`/`GIT_EDITOR` and `git2-hooks` does not, so the re-entry escape hatch never triggers; and abnormal termination during a hook leaves a stale `index.gitbutler-hook-backup` that blocks later commits until it is removed by hand.
- Hook output on success is discarded, and `post-commit` failures only reach `tracing`. Both should go to stderr, as `git` does.
- Change detection compares file bytes, where `worktree_file_to_object` would be filter-aware and constant-memory. Only files with hunk headers are read, so binaries and large files are already excluded, but it is the better primitive.
- `but amend`, `squash` and the other commit-creating commands still run no hooks.

## 🎫 Affected issues

Addresses: #15209

### DIFF
--- a/crates/but/skill/references/reference.md
+++ b/crates/but/skill/references/reference.md
@@ -165,6 +165,7 @@ but commit --above <target> -m "message" <id>  # Place the commit above a commit
 but commit --below <target> -m "message" <id>  # Place the commit below a commit or branch
 but commit -b <branch> --no-message <id>     # Commit without a message
 but commit --empty -b <branch> -m "message"  # Insert an empty commit
+but commit -b <branch> -m "message" --no-hooks # Commit without running commit hooks
 ```
 
 **Where the commit goes:** `-b`/`--branch`, `-A`/`--above`, and `-B`/`--below` are mutually exclusive.
@@ -174,6 +175,11 @@ but commit --empty -b <branch> -m "message"  # Insert an empty commit
 - With no branches applied, a new branch is created. With one applied stack, the commit goes to its top branch's tip. With more than one stack, a targeting flag is **required** — otherwise the command fails with "Unclear where to commit. Found more than one stack". The gate is stacks, not branches: several branches stacked together take an untargeted commit on the stack's top branch.
 
 **Important:** `but commit -b <branch> -m "msg"` with no IDs commits ALL uncommitted changes. Pass IDs to commit only specific files or hunks.
+
+**Hooks:** the repository's `pre-commit` hook runs before the commit is written, and `post-commit`
+after it. A failing `pre-commit` aborts the commit and its output is shown; `--no-hooks` (alias
+`--no-verify`) skips both, matching `but push`. A failing `post-commit` is reported but does not
+undo the commit.
 
 `but commit` is not supported from linked worktrees. Use Git directly for the worktree-local commit, and do not run `but setup` there.
 
--- a/crates/but/src/args/commit.rs
+++ b/crates/but/src/args/commit.rs
@@ -83,6 +83,10 @@ pub struct Platform {
     #[clap(short, long, group = "changes_to_commit")]
     pub interactive: bool,
 
+    /// Bypass commit hooks
+    #[clap(short = 'n', long = "no-hooks", visible_alias = "no-verify")]
+    pub no_hooks: bool,
+
     /// One or more changes to commit.
     ///
     /// A change can either be a file or a hunk.
--- a/crates/but/src/command/legacy/commit.rs
+++ b/crates/but/src/command/legacy/commit.rs
@@ -11,9 +11,11 @@ use but_transaction::{IntermediateCommitCreateResult, Transaction};
 use but_workspace::{RefInfo, branch::create_reference::Anchor, commit::ChangeSource};
 use gitbutler_oplog::entry::{OperationKind, SnapshotDetails};
 use gitbutler_repo::hooks::{ErrorData, HookResult};
+use gix::bstr::{BString, ByteSlice};
 use gix::refs::FullName;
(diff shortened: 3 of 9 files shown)


sensei-woo:
Thanks for working on this. Could you please also restore the `commit-msg` hook? It was dropped alongside `pre-commit` and `post-commit` during the command rewrite, and we rely on it for Claude Code/Codex attribution. Without it, #15209 is only partially fixed.

sensei-woo:
I put together the `commit-msg` follow-up as a stacked draft: https://github.com/ameyypawar/gitbutler/pull/2

It runs the hook after provided/editor message resolution but before the workspace transaction commits, so hook edits are retained and rejection rolls everything back. Merging that PR into this branch will update #15222 directly.

ameyypawar:
Took the `core.hooksPath` fix — credited in `e71c886798`. Good catch: libgit2 ignores the sandbox's config isolation, so a global hooksPath had my own hook running during the test suite.

Holding off on the rest. `message_raw()` has no trailing newline, so a hook doing `echo "Signed-off-by: …" >> "$1"` produces `add a thingSigned-off-by: …` where git gives `add a thing\nSigned-off-by: …`. And `no_hooks_skips_commit_msg_hook` passes with the `run_commit_msg_hook` call removed — it only asserts the marker file is absent.

Both fixable. Worth sending as your own PR on top of this one.

PR #15227, opened 2026-08-07
but-db: document that schema-version bumps are reserved for planned breaks

krlvi:
Bumping `SchemaVersion` locks every older binary out of the database, so the established policy (settled after #14948) is to leave stale columns and tables in place rather than bump for cleanup.

This captures that policy where migration authors will actually see it: on both `SchemaVersion` enums, on `M::up`, on the removed-tables list, and as a short rule in `crates/AGENTS.md` — to prevent accidental bumps during routine cleanup. Comments are deliberately brief; a planned, coordinated break remains legitimate when genuinely needed.


krlvi:
This is only comments so im getting this in

PR #15234, opened 2026-08-08
Derive lite's IPC wiring from the SDK

mtsgrd:
lite's three IPC wiring files restated what the SDK already declares — 69 params interfaces, 102 handler registrations, and ~100 channel strings written out **twice**, because the sandboxed preload could not import the map that held them.

This makes the wiring derive from the SDK's own declarations instead.

> [!NOTE]
> Registering a new endpoint is now the `#[but_api(napi)]` attribute you were writing anyway, plus **one line of TypeScript**.

## Adding an endpoint

**Before** — six edits across three files, for every endpoint:

```diff
  // apps/lite/electron/src/ipc.ts
+ export interface BranchDiffParams {          // ① restate every name and type
+ 	projectId: string;
+ 	branch: string;
+ }
+ 	branchDiff: (params: BranchDiffParams) => Promise<TreeChanges>;   // ② the member
+ 	branchDiff: "workspace:branch-diff",                              // ③ the channel

  // apps/lite/electron/src/preload.cts
+ 	branchDiff: (params) =>                                           // ④ the forwarder
+ 		ipcRenderer.invoke("workspace:branch-diff", params) as Promise<TreeChanges>,

  // apps/lite/electron/src/main.ts
+ import { branchDiff } from "@gitbutler/but-sdk";                    // ⑤ the import
+ senderValidatingHandle(                                             // ⑥ the handler
+ 	liteIpcChannels.branchDiff,
+ 	(_e, { projectId, branch }: BranchDiffParams) => branchDiff(projectId, branch),
+ );
```

**After** — the Rust attribute, and this:

```diff
  // apps/lite/electron/src/ipc.ts
  export const exposedEndpoints = [
+ 	"branchDiff",
  ] as const;
```

The payload type, the renderer-facing signature, the preload forwarder and the main-process handler all follow from the name.

<details>
<summary>Proof, not assertion</summary>

Added `"listBranches"` to that list, changed nothing else, and confirmed `ipc.ts` was the only modified file. It typechecked, and after a rebuild the renderer had a 106-member surface where `window.lite.listBranches({ projectId, filter: null })` returned 1631 branches. Then reverted.

</details>

## How

napi-rs declares endpoints positionally, so TypeScript could see each parameter's **type** but not its **name** — which is exactly why 69 params interfaces existed. `#[but_api(napi)]` already parses those names to generate its wrapper; it now also records them in an `inventory` registry, and `but-ts` emits them beside the types as `apiParamNames`.

Pairing the two recovers the payload shape without restating a single type:

```ts
export type PayloadFor<K extends Endpoint> = { ... };  // names × positional types
```

The generator parses the names back out of the generated declarations and **fails the build** if they disagree with what was registered, so the two halves cannot drift.

Separately, bundling the preload (vite lib mode) removed the sandbox constraint that forced the duplicated channel strings. An endpoint's name is now its channel.

## What it deletes

| file | before | after | |
| --- | ---: | ---: | ---: |
| `ipc.ts` | 741 | 238 | **−503** |
| `main.ts` | 1064 | 566 | **−498** |
| `preload.cts` | 357 | 149 | **−208** |
| **total** | **2162** | **953** | **−1209** |

| | files | added | deleted |
| --- | ---: | ---: | ---: |
| hand-written | 20 | 672 | **1645** |
| generated (new files only) | 4 | 424 | 0 |

**2.4 lines deleted for every line added**, hand-written. Gone: 104 channel strings, 69 params interfaces, 102 handler registrations, 102 preload forwarders.

## Notes for review

- **The generated SDK diff is purely additive.** `index.d.ts` is byte-identical to master; only the new `apiParamNames.{js,d.ts}` files appear. No existing endpoint's signature or behaviour changes, and `apps/desktop` is untouched.
- **Dispatching by name is safer than the positional calls it replaces.** `commitMoveChangesBetween(projectId, sourceCommitId, destinationCommitId, …)` takes two commit ids that are both strings — a hand-written handler could transpose them silently. A payload cannot.
- **Completeness is a compile error, not a runtime one.** A member wired on one side only, or a handler whose payload can't supply its call's arguments, fails to typecheck rather than surfacing as `No handler registered`. Both guards were checked by deliberately breaking them.
- **The exposed list stays a decision.** 12 registered endpoints are deliberately not exposed — `branchLand`, `removeBranch`, `workspaceCheckout`, `commitCherryPick` among them. Adding a name is the moment the renderer is allowed to call it.
- **Seven handlers stay written out**, and each is electron's own rather than the SDK's: clipboard, app version, `shell.openExternal` (with its protocol check), the watcher manager, GUI settings read/write, plus askpass — which isn't a `#[but_api]` endpoint, so it has no registered names.

<details>
<summary>Verification</summary>

Typecheck, oxlint, prettier and knip clean. 54 lite tests, 388 desktop tests, 121 Rust tests across `but-api` and `but-ts`, `cargo fmt --check`, `cargo check --all-targets` on all touched crates.

Because this rewires every IPC call, it was also exercised live over CDP after a rebuild: every dispatch category (derived scalar, zero-arg, multi-argument, main-process overrides, imperative handlers), a watcher subscribe/unsubscribe round-trip, and a forge panel render. Endpoints that can't be invoked safely were probed with a bogus project id — failing at project lookup rather than argument conversion is what proves the parameters resolved by name.

</details>


samhh:
Thank you for this, much better! I had a go at this a while ago but was missing the Rust params piece.

PR #15235, opened 2026-08-08
Add a diff minimap

mtsgrd:
A ruler down the right side of the diff panel that shows the whole change at
once, so you can see where you are and jump around without scrolling blind.

image

## What you see

- Every added and removed run, drawn where it actually is in the diff
- Marks as wide as the lines are long, with indentation left blank, so you get
  a rough sense of the shape of the code
- A thin line between files, plus a file type icon where there is room
- Two columns in side-by-side mode, one in unified — same as the diff
- A lens over the part that is currently on screen

## What you can do

- Grab the lens and drag it, the way you would a scrollbar
- Or press anywhere on the track to jump there
- Hover to see which file you are over

## When it shows up

- Only once the diff is at least two screens tall. Below that the lens covers
  most of the ruler and says nothing the scrollbar would not, so the normal
  scrollbar stays instead
- Works the same whether you show all files at once or one at a time
- Very large change sets are turned away rather than mapped, counted off the
  patch so nothing is parsed to find out

## How it works

- Positions come from the diff viewer's own layout, so files it has not
  rendered yet are still placed correctly. The within-file model was checked
  against the viewer's own exact line positions and matched on every file, hunk
  and mark measured
- It is painted on a canvas, so a few hundred files cost nothing in DOM, and
  scrolling only moves the lens
- Once there is less than a pixel per line it averages them, each weighted by
  how much of the pixel it covers, the way an image is scaled down

## Known limits

- With line wrapping on, marks shift a little as you scroll, because the viewer
  only estimates line heights until it renders them. Wrapping off, which is the
  default, is exact

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/Details.module.css
+++ b/apps/lite/ui/src/routes/project/$id/workspace/Details.module.css
@@ -148,6 +148,19 @@
 
 .diffContentsContainer {
 	display: grid;
+	/* Code, then the minimap ruler — which collapses the column when it has
+	   nothing to map. */
+	grid-template-columns: minmax(0, 1fr) auto;
+
+	/* Drop the native scrollbar only while the minimap is showing, so a diff too
+	   short to map keeps one. */
+	&:has([data-minimap-navigable="true"]) .diffContents {
+		scrollbar-width: none;
+
+		&::-webkit-scrollbar {
+			display: none;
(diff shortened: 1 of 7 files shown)


mtsgrd:
Feeling pretty confident this is a good direction, landing now and will fix forward/remove if any issues come up.

Issue #15154, opened 2026-08-04
Squash-merged stack is never detected as integrated, and the conflicted residue permanently blocks re-detection


yunjae-park1111:
### Version

`but` CLI 0.22.0 (analysis done against the `release/0.22.0` tag; the decisive files are byte-identical on `master` @ `5ed3782`, 2026-08-03)

### Operating System

macOS (Apple Silicon)

### Summary

Follow-up to #14298 with a source-level analysis. I'm filing it separately because it covers a second, so-far-unreported defect: **after the first failed integration of a squash-merged stack, the workspace enters a state that can never be detected as integrated again** — there is no automatic recovery path. Fixing squash detection alone (e.g. reviving #14323) would not heal workspaces that are already in this state.

Related: #14298 (squash → conflict on integrate), #15127 (`(no changes)` residue never cleaned up), #8457 (same symptom on the old backend), #14323 (closed-unmerged fix attempt for the detection half).

### How to reproduce

1. Create a stack with 2 commits A and B that touch overlapping lines; push; open a PR.
2. Squash-merge the PR on GitHub (the target branch gains one new commit whose content is A+B). The merge itself is clean — GitHub reports no conflicts, CI is green.
3. `but pull`.

### Observed symptoms

**1. The lane turns into a zombie instead of disappearing.**

After the pull, the merged stack is still applied. Both commits were rebased onto the new base (which already contains their squashed sum), so both now show `(no changes)`; the first one is additionally stored `{conflicted}`. Sanitized `but status` right after the pull (healthy unrelated lanes omitted, names redacted):

```
┊╭┄ ab [dep-bump-branch] CI: ✅ #1234: fix: bump dependencies to clear CVEs
┊◐   cd fix: adjust install step so the bump takes effect (no changes)
┊◐   ef fix: bump dependencies to clear CVEs (no changes) {conflicted}
├╯
┊
┴ 1a2b3c4 (common base) 2026-08-04 fix: bump dependencies to clear CVEs (#1234)

Hint: ◐ means rewritten locally vs upstream.
```

Three things worth pointing out in that snapshot:

- The common base **is** the squash commit of this very branch — the content is demonstrably already downstream of us, yet the lane above it survives.
- GitButler itself renders `CI: ✅ #1234` (the merged PR) on the exact lane it flags `{conflicted}`. The forge association is clearly known; it just doesn't feed the integration decision.
[...]

PR #15241, opened 2026-08-09
fix: use squash matches during upstream integration

### LINKED ISSUE #14298: Conflicted files when using the new integrate upstream model
### Version

0.20.1

### Operating System

macOS

### Distribution Method

dmg (Mac OS - Apple Silicon)

### Describe the issue

When I try to squash and merge a pull request that contains separate file edits to the same file lines, integrating upstream after merging will give a conflicted file error.

1. Make changes to the same file lines in separate commits
2. Make a pull request
3. Squash and merge the pull request
4. Fetch and integrate upstream
5. Got the error of conflicted files.

Image

I did not expect to get a merge conflict in this quite simple case.

```shell

```

VanjaRo:
## 🧢 Changes

- Use cumulative changeset matches during upstream integration.
- Remove local commits only when their combined content is present upstream.
- Keep unmatched branches and commits local.
- Add regression coverage for direct and managed workspaces, shared stack bases, multi-segment squashes, net-zero heads, and near matches.

## ☕️ Reasoning

`RefInfo` already detected squash-integrated commit ranges by comparing their combined content with upstream. The integration path did not use that result and checked only individual commits.

Consequently, GitButler could report a squashed range as integrated and then replay the same commits during integration, producing unnecessary conflicts or leaving stale branch state.

This change applies cumulative content matching to the integration path. Matching is attempted within named branch segments first. A squash spanning adjacent segments is accepted only when the complete range is eligible and exactly matches upstream. Unmatched or net-zero live segments remain local.

This prevents incorrect replay of squash-integrated commits. It does not repair workspaces already affected by an earlier integration.

## 🎫 Affected issues

Fixes #14298
Related to #15154

### DIFF
--- a/crates/but-core/src/changeset.rs
+++ b/crates/but-core/src/changeset.rs
@@ -92,8 +92,28 @@ impl<T: ChangesetCommit + ?Sized> ChangesetCommit for &T {
 
 /// Similarity matches between workspace commits and upstream commits, computed from commit IDs.
 pub struct SimilarityByCommitIds {
-    /// Upstream commit IDs keyed by the workspace commit ID that matched them.
+    /// Upstream commit IDs keyed by workspace commits proven integrated individually or by range.
     pub matches_by_workspace_commit: HashMap<gix::ObjectId, gix::ObjectId>,
+    upstream_lut: Identity,
+    time_used: Duration,
+}
+
+impl SimilarityByCommitIds {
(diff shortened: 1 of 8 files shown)


krlvi:
Thanks @VanjaRo We are working on a possibly more complete approach in https://github.com/gitbutlerapp/gitbutler/pull/15309
Closing this here and we will continue the work in the other PR

PR #15248, opened 2026-08-09
Show the pull request a branch already has

mtsgrd:
Selecting a branch in the branches explorer showed a **Diff / Pull Request** toggle whose form could not work: `publish_review` derives the base from the branch's position in a workspace stack, and an unapplied branch has none.

The pane dispatched on the *operand* — what is selected, not what it is — and both outline tabs mint a `Branch`. It now takes a `DetailsSelection`:

```
AppliedBranch | UnappliedBranch | Commit | UncommittedFile
```

`WorkspacePage` already resolves selection per tab, so it names what it resolved. Applied and unapplied are the words the rest of the app uses — `unappliedStacks` is what fills the branches tab, and `ListedBranch` was already taken by the SDK type those stacks are made of.

## What the branches tab shows now

| | header | body |
|---|---|---|
| branch has a pull request | Diff / Pull Request toggle | diff, or the full review |
| branch has none | title only | diff |

A review is keyed by branch name, so the description, conversation, checks, reviewers and labels were already reachable — nothing new was needed to read them. The toggle appears only alongside a review, so it is never a half-dead control. Opening a review is still not offered here; apply the branch and use the workspace, where the base exists.

`AppliedBranchDetails` and `UnappliedBranchDetails` share `BranchTabToggle`, `useBranchTabHotkeys`, `BranchDiff`, `BranchTitleRow` and `ReviewView`. Editing the description stays the applied view's affordance; the review's own controls work in both.

`Operand` keeps its shape. It is the argument type of `getOperations`, and an unapplied branch is precisely what operations cannot act on — so the distinction belongs where the two tabs converge, not in a dozen exhaustive matches that could not act on it. Its `Branch` case is now commented to say it means an *applied* branch, which it already did: `operandLabel` asserts the ref resolves to a segment.

## Verification

Driven against the running dev app over CDP: an unapplied branch with a PR shows both tabs and the full review (description, checks, reviewers, labels, activity); one without shows title and diff only; an applied branch is unchanged and its pull-request tab still loads.

Checked against the committed tree in a scratch worktree, not the working copy: typecheck, `oxlint` over the whole repo, `prettier`, `knip:prod`, `knip:non-prod` and the lite unit tests (77/77) all clean.

### DIFF
--- a/apps/lite/ui/src/operands.ts
+++ b/apps/lite/ui/src/operands.ts
@@ -3,6 +3,10 @@ import type { HunkLineSelection } from "#ui/hunk.ts";
 
 export type Operand =
 	| { _tag: "UncommittedChanges" }
+	/**
+	 * Applied to the workspace: no operation can act on an unapplied branch, and
+	 * `operandLabel` asserts the ref resolves to a segment.
+	 */
 	| ({ _tag: "Branch" } & BranchOperand)
 	| ({ _tag: "Commit" } & CommitOperand)
 	| ({ _tag: "File" } & FileOperand)
@@ -37,7 +41,9 @@ export const uncommittedChangesOperand: Operand = {
 	_tag: "UncommittedChanges",
 };
 
-export const branchOperand = ({ branchRef }: BranchOperand): Operand => ({
+export const branchOperand = ({
+	branchRef,
+}: BranchOperand): Extract<Operand, { _tag: "Branch" }> => ({
 	_tag: "Branch",
 	branchRef,
 });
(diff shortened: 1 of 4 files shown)


samhh:
From a UX perspective there are two issues:

1. There is now layout shift when navigating between a branch with a PR and one without one.
2. I might want to define a draft PR, and perform other actions, on a branch that's not applied. I don't personally see the value in this limitation; perhaps it speaks to a wider habit of how we treat the workspace and everything else as two different worlds.

PR #15264, opened 2026-08-10
Add root `--skill` flag printing the bundled skill text

schacon:
`but --skill` writes the bundled base skill text (`SKILL.md`) to stdout and exits, so agents can load the GitButler skill into context directly (e.g. `but --skill >> prompt.md`) without installing skill files first.

## What changed

- New root-only `--skill` flag on `Args`. It short-circuits right after argument parsing, like `--version`: `but --skill` prints the skill even if a subcommand follows, and `but status --skill` is rejected since the flag is not global.
- The output reuses the installer's `prepare_skill_content` path via a new `command::skill::base_skill_text()`, so the CLI version is injected into the frontmatter exactly as `but skill install` writes it, and the embedded file is UTF-8 validated.
- Output is raw stdout — no pager, truncation, or output-channel formatting — so it pipes cleanly.
- `but skill --help` long help now cross-references `but --skill` for discoverability.

## Decisions

- Only `SKILL.md` is printed, not the `references/` files; the flag targets bootstrap/context-loading, where the base skill is the entry point.
- The bundled skill docs themselves don't mention the flag: an agent that already has the skill installed has no use for printing it.

## Testing

Integration test asserts `but --skill` output equals the embedded `SKILL.md` with the version swapped in, deriving the expected version from `but --version` so it holds for both `dev` and release builds.

### DIFF
--- a/crates/but/src/args/mod.rs
+++ b/crates/but/src/args/mod.rs
@@ -44,6 +44,9 @@ pub struct Args {
     /// Use when the next step needs IDs or details from the resulting workspace.
     #[clap(long, global = true)]
     pub status_after: bool,
+    /// Print the bundled agent skill text (SKILL.md) to stdout and exit.
+    #[clap(long)]
+    pub skill: bool,
     /// Subcommand to run (`but <COMMAND>`).
     ///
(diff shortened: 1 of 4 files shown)


krlvi:
Since we already have a `but skill` command, how about this being `but skill show`?

PR #15275, opened 2026-08-10
Don't block committing when the workspace has no branches

PavelLaptev:
In Lite, a workspace with zero stacks was a dead end for committing. The commit
target list is built from the workspace stacks, so with none it was empty, the
target was null, and Commit stayed disabled. "Start commit" still expanded the
form and accepted a message that could never be submitted — the separate "Add
new branch" action was the only way forward, and nothing on screen said so.

Committing now creates the branch itself, the way the desktop app already does.

## How

Lazily, on submit: the empty-workspace path calls `branchCreate` with
`newRef: null` (letting Rust generate and normalize the name), selects the new
branch in the outline, then commits onto it. Opening the form writes no ref, so
a user who changes their mind leaves no branch behind.

To name that branch before it exists, this adds `branch_canned_name` to
`but-api` — one call to the existing `unique_canned_refname`, the very
generator `branch_create` falls back to — and wires it through the generated
SDK, the Electron IPC allowlist, and a query invalidated by anything that moves
a ref. The target chip shows that name with a `new` badge and is disabled,
since there is nothing to pick; if the endpoint is unavailable it degrades to a
generic "New branch".

`hasNoBranches` is a prop of its own rather than a synthetic target item:
target items carry an `Operand` that drives the outline selection, and a branch
that doesn't exist yet has none. Amend is deliberately untouched — it still
requires a commit that already exists — and `isBranchCreatePending` folds into
the pending state so the form cannot double-submit across the two mutations.

## Testing

`pnpm -F @gitbutler/lite check`, `cargo check -p but-api`, oxlint, oxfmt,
prettier and knip all pass.

Driven over CDP against a dev app on a project with zero stacks: the target
chip is `aria-disabled` and announces the branch it will create, the commit
options chevron is enabled where it was previously dead, and expanding the form
gives an enabled Commit button.

## TODO
It would be nice do display a draft branch. Will do it in the next PR


krlvi (approved):
lgtm

PavelLaptev:
> lgtm

🙏

PR #15280, opened 2026-08-10
branch delete: clean up no-commit paths to target

jonathantanmy2:
Currently, when a branch is deleted, an edge between the workspace
commit and the target (or the local branch corresponding to the target
ref) remains. Ensure that we do not have that edge.

### DIFF
--- a/crates/but-transaction/src/lib.rs
+++ b/crates/but-transaction/src/lib.rs
@@ -4,7 +4,7 @@ use std::{
 };
 
 use anyhow::Context as _;
-use bstr::BStr;
+use bstr::{BStr, BString, ByteVec};
 use but_api::WorkspaceState;
 use but_core::{
     DiffSpec, DryRun, RefMetadata, commit::CommitIdentifiers, ref_metadata, sync::RepoExclusive,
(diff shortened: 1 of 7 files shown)


crates/but-transaction/src/lib.rs
```diff
@@ -339,8 +339,84 @@ where
...
+                    }
+                }
+                Some(child_selector)
+            };
+
+            if let Some(must_disconnect_child) = must_disconnect_child {
+                editor.remove_edges(must_disconnect_child, ref_selector)?;
+            } else {
+                editor.replace(ref_selector, but_rebase::graph_rebase::Step::None)?;
+            }
```
Caleb-T-Owens, line 418: Should we always be replacing the ref_selector with None?

Caleb-T-Owens: In the tests, it seems like only removing edges results in a lingering `A` reference.

I'm wondering whether or that that is expected.

jonathantanmy2: Ah, good catch! Not sure how I missed that.

Caleb-T-Owens (approved):

PR #15283, opened 2026-08-11
Add Playwright infra to Lite

samhh:
Closes GB-1829. See details in fabc7bd67d9830c28fedbd1803565b6927355555. Between GitHub Actions, Turborepo, and Electron our tooling conspires to make this nontrivial.

The slow Rust build portion of [the new job](https://github.com/gitbutlerapp/gitbutler/actions/runs/31506013972/job/93827711286) should be faster once this is merged and a cache entry from master is created.

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -27,6 +27,7 @@ jobs:
       docs: ${{ steps.filter.outputs.docs }}
       but_installer: ${{ steps.filter.outputs.but_installer }}
       install_script: ${{ steps.filter.outputs.install_script }}
+      lite_e2e: ${{ steps.filter.outputs.lite_e2e }}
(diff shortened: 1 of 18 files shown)


estib-vega:
@samhh I think it'd make more sense to move the `lite/e2e` directory under the top-level `e2e` one. So that all e2e tests are there are well. What do you think?

samhh:
I suppose it depends if we view these tests as "test every GitButler product end to end" or "test Lite end to end". Notably at least the Web E2E tests already live separately.

I'm disinclined because then `:/e2e/` would need to know how to build and launch Lite, breaking package boundaries a bit.

estib-vega (approved):
Fair point. Not too important either 👍

PR #15313, opened 2026-08-13
Handle Shift-click on commit rows

krlvi:
## Context

Shift-clicking an enabled Lite commit row selected it instead of extending the checked range; only the checkbox forwarded the modifier.

Fixes GB-1830: https://linear.app/gitbutler/issue/GB-1830/shift-click-entire-row-for-multichecking

## Change

Route Shift-clicks on ordinary row content through the existing commit-checkbox range behavior. Interactive controls and normal row selection keep their existing handling.


samhh:
I'm unsure we need the complexity of `isFromNonRowBody`. The unit tests are also probably unnecessary; if we really want to test this it's most easily tested from a user's perspective with an E2E test.

krlvi:
> I'm unsure we need the complexity of `isFromNonRowBody`. The unit tests are also probably unnecessary; if we really want to test this it's most easily tested from a user's perspective with an E2E test.

fair! lets simplify it - this got merged before the feedback

PR #15342, opened 2026-08-15
fix: make Open in browser work from the first try.

ValeraS:
## 🧢 Changes
Disable "Open in browser" menu item until `branchUrl` is evaluated.

## ☕️ Reasoning
`branchUrl` is Svelte's `$direved` value, that is lazy and not (re-)evaluated until it is actually read. `branchUrl` is only read inside the menu item’s click handler. On the first click, reading it initialises `compareBranchUrlQuery`, which starts the async `forgeCompareBranchUrl` request—but its response is still `undefined` and onClick handler does nothing.


krlvi (approved):
thanks!

PR #15349, opened 2026-08-16
Lite: run the panes edge to edge, with the app's own scrollbar

PavelLaptev:
## 🧢 Changes

- Panes now scroll natively, with a single scrollbar styled globally in `global.css`. This replaces Base UI's `ScrollArea` and the `Scroller` wrapper, and removes `.overlayScrollbar`, `.scrollAreaRoot` and friends from `ui.module.css`.
- Rows and cards run edge to edge in the outline, upstream and branches tabs — no boxed cards, no inset gutters, and none of the `margin-inline` stretch hacks that used to compensate for them.
- Upstream rows are squared off and the boxed card around them is gone.
- The stack card header is now a labelled strip; fold-all moved from its own button into the stack menu.
- The push button is separated from the branch row meta, and author meta truncates when space is tight.
- The diff pane is pinned to the diff area's second grid row.
- Fixed a double-dimmed rail connector between folded branches.
- Popups keep the scrollbar gutter on their own background, via `--scrollbar-track-bg` / `--scrollbar-track-border` tokens that popup surfaces restate.
- The stack menu stays reachable outside the default mode; only the items that mutate the stack gate themselves.

## ☕️ Reasoning

The panes were built on Base UI's `ScrollArea`, which gave a true overlay scrollbar but needed a wrapper component and a fair amount of CSS to imitate a native one. Because that overlay reserved no space, content had to be pushed back under it with negative margins, and the boxed cards existed partly to hide where those gutters landed.

Scrolling natively removes all of it: one scrollbar definition in `global.css` applies everywhere, the gutter is only taken while content actually overflows, and rows are free to run to the pane edge. The gutter carries the panel background with a hairline dividing it from the content, so it reads as part of the chrome rather than as something floating over it.

## 📸 Before / after

Captured from the same app at the same viewport (1545×766, 2× DPR), with only this branch applied or unapplied between the two runs — so each pair differs by styling alone.

### Workspace sidebar

Stacks edge to edge, the stack card header as a labelled strip, and the separated push button.

| Before | After |
| --- | --- |
|  |  |

### Upstream tab

Squared rows, with the boxed card dropped.

| Before | After |
| --- | --- |
|  |  |

### Branches tab

Rows run edge to edge instead of sitting in rounded cards. Filtered to `pl-branch` so both runs list exactly the same branches.

| Before | After |
| --- | --- |
|  |  |

### Uncommitted file rows

| Before | After |
| --- | --- |
|  |  |

### Popup scrollbar

The project picker with its list overflowing: the gutter takes the popup's own background rather than the panel background, and drops the divider that would otherwise cut across it.

| Before | After |
| --- | --- |
|  |  |

### Diff pane

Pinned to the diff area's second grid row.

| Before |
| --- |
|  |

| After |
| --- |
|  |


krlvi:
This does look much cleaner indeed! Thanks for making this happen

PavelLaptev:
> This does look much cleaner indeed! Thanks for making this happen

Thanks! I still owe you the Upstream tab change, but I decided to put that in the next PR.

PR #15359, opened 2026-08-17
Merge gitbutler-filemonitor into gitbutler-watcher

Caleb-T-Owens:
`gitbutler-filemonitor` had a single workspace consumer (`gitbutler-watcher`), so the crate is folded into watcher instead of kept as a standalone package.

- Moved file-monitor modules, watch-plan, tests/fixtures, and the vendored `gitbutler-notify-debouncer` into `crates/gitbutler-watcher`
- Watcher now owns `WatchMode`, `spawn`, and `InternalEvent`; existing `gitbutler_watcher::WatchMode` callers are unchanged
- Deleted `crates/gitbutler-filemonitor` and dropped it from workspace dependencies
- CI vendored-debouncer tests now run from `crates/gitbutler-watcher/vendor/debouncer`

Validation: `cargo check -p gitbutler-watcher --all-targets` and `cargo test -p gitbutler-watcher --lib` (11 passed).


Caleb-T-Owens:
Withdrawn: this only moved gitbutler-filemonitor into gitbutler-watcher rather than eliminating the logic. Crate removal must delete unused code, not relocate it.

PR #15363, opened 2026-08-17
Redesign the Upstream and Branches tabs in Lite

PavelLaptev:
Both outline tabs get the headers their designs call for, and the branch
list gains the filter behaviour the file lists already had.

**Upstream tab** — three headed bands divide the listing: incoming
changes, the workspace's branches against them, and the older target
history behind both. The update prose and button move out of the graph
into their own band, as does the control paging further back. The older
section restores cursor-based paging that was dropped after the tab's
first version, now keyed under the base listing's query root so a fetch
or workspace update invalidates the pages with it — 10 commits on
arrival, 25 per "load more".

**Branches tab** — a header replaces the permanent search field and lone
filter button, and the search collapses into it, taking the header's
place only while open so narrowing the list costs no vertical space.
The stored query becomes nullable, null meaning closed, matching how the
file lists hold theirs.

**Starting a branch** — the `+` created a workspace lane and nothing
else, leaving no way to start a branch the way a plain checkout does. It
now opens a menu with both, since neither is the obvious default. The
second uses the existing `branch_checkout_new` endpoint; `Mod+Shift+N`
is bound beside `Mod+N` so the accelerators the menu shows are real
wherever the outline is.

### Decisions worth a look

- `useFileFilter`/`FileFilterRow` become `useListFilter`/`ListFilterRow`,
  with rows named by key rather than by path, so the branches tab shares
  the keyboard loop instead of copying it. No behaviour change for the
  two file lists.
- The Upstream tab's section headers are its own component rather than
  `SectionHeaderRow`: shorter and quieter, because three of them fall
  inside one list. The two now differ, which is intentional but means
  the shared "Stack header" in Figma is only half implemented.
- The older section's rows keep the target's rail colour rather than
  reporting individually that the workspace has them, following the
  design.
- `buildItems` used to drop the commit run trailing the deepest fork
  point. Those are older integrated commits by definition, so they now
  head that section — otherwise the paged history starts below them and
  the rail draws across a hole.

### Verified

Screenshots of the changed surfaces are in a comment below. Typecheck,
oxlint, knip, prettier and the unit tests pass.

The paging depth itself is not exercised: the screenshot fixtures have
only one commit of older history, so the 10-on-arrival / 25-per-press
split and the "load more" control are unproven against a real backlog.


PavelLaptev:
## 📸 UI screenshots

Captured 8 surfaces, 6 changed.

### branches-tab

| Before | After |
| --- | --- |
|  |  |

### commit-form

| Before | After |
| --- | --- |
|  |  |

### details-pane

| Before | After |
| --- | --- |
|  |  |

### uncommitted

| Before | After |
| --- | --- |
|  |  |

### upstream-tab

| Before | After |
| --- | --- |
|  |  |

### workspace-sidebar

| Before | After |
| --- | --- |
|  |  |

<details>
<summary>2 surfaces unchanged</summary>

`project-picker`, `settings`

</details>

PR #15365, opened 2026-08-17
Add Lite local repository onboarding

estib-vega:
## Summary

- add a native directory picker and shared Lite flow for adding local repositories from onboarding or the project picker
- expose `addProject` through N-API, invalidate the projects cache, and regenerate the SDK bindings
- add Lite-specific test IDs and a Playwright onboarding journey that verifies the checked-out branch is unchanged

## Why

Lite's first-run experience previously had no way to add a local repository. This makes onboarding actionable while preserving Lite's normal branch instead of switching to `gitbutler/workspace`.

## Visuals
Screenshot 2026-08-19 at 09 42 29
Screenshot 2026-08-19 at 09 43 09


krlvi (approved):

mtsgrd:
I'm just a bit curious about the added dependency on @gitbutler/ui? It looks like it's for reusing test ids.

estib-vega:
> I'm just a bit curious about the added dependency on @gitbutler/ui? It looks like it's for reusing test ids.

Yep, exactly. 
Like for desktop, we moved the test IDs to the ui package, as most of the components were defined there.

I did the same for Lite since I expect that eventually its components will be there

PR #15369, opened 2026-08-17
Lite: tighten up a few workspace UI details

PavelLaptev:
Four small presentational cleanups in the Lite workspace UI. No behaviour changes.

- **Commit form "new" marker** now uses the shared `Badge` component instead of a
  bespoke CSS chip.
- **Upstream tab** drops its per-section bands in favour of a single pane heading.
- **Outline branch header** no longer draws a separator dot before the `Push`
  button; the button takes a 4px left margin instead, so the row reads as
  status-then-action rather than as two facts of equal weight.
- **Panel padding** — the details actions row and the shared section header row
  both move onto `var(--panel-padding-inline)`, so the bands down a panel line up.
  With the actions row on a uniform 8px, the `filesHidden` variant asked for the
  inset it already had, so it and its only consumer are gone.

The padding item is a 2px shift — look for it rather than at it.

Validation: `pnpm -F @gitbutler/lite check` and `pnpm knip:prod` pass. No unit or
e2e test references the removed rule or class.


PavelLaptev:
## 📸 UI screenshots

Captured 8 surfaces locally against the seeded fixtures. Before is this PR's base
(`lite-ui-tweaks`), not `master`, so the diff below is this PR's own change only.

Surfaces were picked by reading the diff rather than by byte-comparing the runs: the
fixtures build fresh repositories per run, so SHAs and relative timestamps differ between
captures and every surface compares as "changed". The outline sidebar is the only surface
this change reaches.

### workspace-sidebar
The per-stack header row ("Stack of 3 branches") is gone; branch rows move up to meet the
section heading.

| Before | After |
| --- | --- |
|  |  |

<details>
<summary>7 surfaces not shown</summary>

`branches-tab`, `commit-form`, `details-pane`, `project-picker`, `settings`, `uncommitted`, `upstream-tab` — no code path from this change reaches them.

</details>

PR #15374, opened 2026-08-17
Discard multiple commits

samhh:
SDK update that is in turn used to support the functionality in Lite.


krlvi (approved):
this looks reasonable!

PR #15387, opened 2026-08-18
Rest emptied stack bottoms on the commit they point to

krlvi:
In single-branch mode, when the checked-out branch sits at the target tip (e.g. the canned branch created once all stacks integrate), its commits are pruned as integrated and its stack ends up empty. The stack's `base` kept its collection-time value — the first commit of whatever segment came next along the first parent, typically the stale local tracking branch (`master`) or any old local ref below. `head_info` then reported the branch as resting far below the target, and Lite's Upstream tab drew every target commit above the branch as if it were behind.

| before | after |
|---|---|
| `origin/master` → "6 commits between" → `kv-branch-3` | `kv-branch-3` directly at `origin/master` |

`recompute_last_segment_base` now handles an emptied bottom: if the graph segment still owns commits (pruned as integrated), the branch rests on the commit it points to; a truly empty ref keeps resting on its ancestor. The projection recomputes every stack's base once, after all pruning and truncation passes, rather than only when upstream advanced — so a pruned bottom always names the parent of its bottom-most remaining commit.

Note for graph owners: this also refreshes the base of non-empty stacks whose integrated bottom commit was pruned when upstream has not advanced (e.g. `B on fafd9d0` → `on 91bc3fc`, the parent of B's bottom-most displayed commit). Every changed snapshot is an emptied/pruned bottom now naming the commit it actually rests on.

### DIFF
--- a/crates/but-api/tests/api/branch_checkout.rs
+++ b/crates/but-api/tests/api/branch_checkout.rs
@@ -257,12 +257,12 @@ fn checkout_new_returns_head_info_matching_fresh_head_info() -> anyhow::Result<(
 
     snapbox::assert_data_eq!(
         crate::support::workspace_graph(&ctx)?,
-        snapbox::str![[r"
+        snapbox::str![[r#"
 ⌂:0:new-branch[🌳] <> ✓refs/remotes/origin/main on 5374caf
-└── ≡:0:new-branch[🌳] {1}
+└── ≡:0:new-branch[🌳] on 5374caf {1}
     └── :0:new-branch[🌳]
 
-"]]
+"#]]
     );
 
     #[cfg(feature = "graph-workspace")]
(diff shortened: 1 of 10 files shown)


estib-vega:
Can't review the but-graph code confidently, but looking at the changes (that seem to keep the base information only) and if the tests pass, it's good from my side.
Would like to give it a spin on nightly though, to test out any potential repercussions

estib-vega (approved):
Would want to wait for another review for someone with more context on the changes to but-graph, cc @Caleb-T-Owens

krlvi:
Narrowed after further validation: a truly empty ref (no commits in its graph segment) now keeps its collection-time base exactly as before, so graph-init reconciliation (`candidates_for_independent_branches_in_workspace`) sees identical inputs. Only segments emptied by pruning, and stacks whose pruned bottom moved, get a refreshed base.

Validation: full `cargo test --workspace` green (109 test binaries); `head_info` diffed old-vs-new binary on a real multi-stack managed workspace (byte-identical), on the same workspace with two integrated stacks applied (byte-identical), and on the single-branch-at-target-tip case (only the base changes, `fade02a7` → the target tip).

PR #15412, opened 2026-08-19
Image rendering in diffs

samhh:
Closes GB-1824.

We need to discuss the security implications of `getWorkspaceFile`, because secrets may be present (e.g. `.env`) but uncommitted and the API makes these available to the renderer. I suspect Tauri already has this problem.

This SDK function is used in this PR to get the blob of uncommitted images. Lite will also need this function for uncommitted additions in partial diff hydration (i.e. load whole file on demand).

We could potentially solve this by hiding files not visible to Git from this function. That limits the surface of attack to uncommitted secrets that haven't been gitignored yet.

Exploit example:

```diff
diff --git a/apps/lite/ui/src/routes/project/$id/workspace/ImageDiff.tsx b/apps/lite/ui/src/routes/project/$id/workspace/ImageDiff.tsx
index 2c900631e7..f8693bc4d3 100644
--- a/apps/lite/ui/src/routes/project/$id/workspace/ImageDiff.tsx
+++ b/apps/lite/ui/src/routes/project/$id/workspace/ImageDiff.tsx
@@ -2,7 +2,7 @@ import type { PayloadFor } from "#electron/ipc.ts";
 import type { FileParent } from "#ui/operands.ts";
 import type { TreeChange } from "@gitbutler/but-sdk";
 import { queryOptions, useQuery } from "@tanstack/react-query";
-import type { FC } from "react";
+import { type FC, useEffect } from "react";
 import styles from "./ImageDiff.module.css";
 
 type ImageSource =
@@ -138,6 +138,12 @@ export const ImageDiff: FC<{
 	fileParent: FileParent;
 	version: number;
 }> = ({ projectId, change, fileParent, version }) => {
+	useEffect(() => {
+		void window.lite
+			.getWorkspaceFile({ projectId, relativePath: ".jj/repo/config-id" })
+			.then(({ content }) => console.log(".jj/repo/config-id", content));
+	}, [projectId]);
+
 	const sources = imageSources(change, fileParent, version);
 	const before = useImageUrl(projectId, sources.before);
 	const after = useImageUrl(projectId, sources.after);

```

```
.jj/repo/config-id <file contents>
```


krlvi:
oooh nice!

samhh:
I intend to merge these PRs on the following basis:

1. It should not block us in dev, especially because...
2. ...it is not a regression versus Tauri.
3. I will ensure we discuss this in the next Lite meeting, if we haven't already done so by then. (Edit: I'm going to have a crack at resolving the issue, but I still don't consider it a blocker.)

PR #15423, opened 2026-08-19
Move Lite test IDs into Lite

estib-vega:
## Summary

- move `LiteTestId` from `@gitbutler/ui` into Lite UI source
- share the local test IDs with the instrumented React components and Lite E2E tests
- remove Lite's `@gitbutler/ui` dependency and the corresponding UI package step from the Lite E2E workflow

## Why

Lite only depended on `@gitbutler/ui` for four test IDs. That caused the Lite E2E workflow to package unrelated Svelte UI code before testing the React app.

The local definition uses an erasable const object because Lite enables `erasableSyntaxOnly`.


mtsgrd (approved):

samhh:
Perfect shouldn't be the enemy of good - any E2E test is an improvement - but worth noting that E2E test IDs can often be an anti-pattern. Better where possible to test against what the user actually sees, for example a button with the text "add project".

estib-vega:
> Better where possible to test against what the user actually sees, for example a button with the text "add project".
>

@samhh  I disagree here. That would impose implicit constraints on the UI. We couldn't have buttons with the same label, and changing the labels would involve changing the tests.

I agree that there's a balance in between 0 and everything has a dedicated test ID, but I'm happy to look for that balance.
Is there a specific concern you have with Test IDs?

samhh:
It does imply that changing copy also means updating tests.

Conflicts aren't often an issue since you can query within a subtree of the DOM, see e.g. Lite's branches spec. Arguably if copy conflicts it's signalling a UX issue.

It's better for maintaining accessibility, and slightly increases the coverage of what you're testing since you're more closely mimicking what the user sees and does. See:

* [https://playwright.dev/docs/locators#:\~:text=locators-,We%20recommend%20prioritizing%20role,technology%20perceive%20the%20page.,-Locate](<https://playwright.dev/docs/locators#:~:text=locators-,We%20recommend%20prioritizing%20role,technology%20perceive%20the%20page.,-Locate>)
* [https://tkdodo.eu/blog/test-ids-are-an-a11y-smell](<https://tkdodo.eu/blog/test-ids-are-an-a11y-smell>)
* [https://testing-library.com/docs/queries/about/#priority](<https://testing-library.com/docs/queries/about/#priority>)

estib-vega:
I get more your point now. Thanks for the links! :)

---

**Some extra context**

I think I've landed in the opinion of we shouldn't over do it, but there was a reason we had to: 
In the previous GUI, we re arranged it a good 3 times (by hand without AI like cavemen) and that meant also having to find the right DOM selectors.

We had a very dense and high-depth component trees (*are you selecting the stack container or the branch header?*), with lots of repeated labels (e.g. the repeated 'start commit button').

We introduced the test IDs in order to make that process easier for us to change and to still have the confidence that we didn't get false positives

> Conflicts aren't often an issue since you can query within a subtree of the DOM, see e.g. Lite's branches spec. Arguably if copy conflicts it's signaling a UX issue.
>

Probably. I think more than an issue, the nature of the application features we want to support demand a high UX price.
I think it's a bit better in lite, but since we're still trying out things, I expect that changing the UI/UX with the certainty that we can trust our E2E will be an important thing for us to have.

---

That said, I agree, adding test IDs for everything is not needed especially for buttons that probably won't be duplicate.
I'll remove some of the test IDs and merge this if that's ok.

PR #15458, opened 2026-08-20
Keep arrow-key file navigation ahead of the details pane

mtsgrd:
Holding an arrow key over the uncommitted file list stuttered every few files, and the list could not keep up with the key.

Moving the file cursor also re-pointed the details pane in the same keystroke — a diff-cursor write plus a scroll through the virtualised viewer. That is more work than a key repeat leaves between moves, so the pane, not the list, set the pace, and each re-point blocked a frame.

The pane now waits for the cursor to settle. A cursor arriving out of stillness re-points at once, so single steps stay immediate; one arriving mid-run defers, and every further move cancels it, so a held key re-points once on release rather than at every file it passes.

Also memoises `absorptionTargetCommitIds` and `appliedAddressSpace` in `PageBody`. Rebuilt every render, those handed `WorkspaceListsContext` a fresh value, which re-rendered every row in the lists on any cursor move. `uncommittedAddressSpace` beside them was already memoised this way.

### Measured

Arrow key held over the uncommitted file list at a 75ms repeat, 11 presses, three runs of each side, app window foregrounded. A "hitch" is a long task landing during the hold, which blocks a frame.

| | before | after |
|---|---|---|
| presses that land | 10 of 11 | 11 of 11 |
| hitches during the hold | 4–5 | 1 |
| blocked by those hitches | 221–280ms | 54–59ms |
| gap between moves | mean 97ms, range 69–136ms | mean 62ms, range 54–81ms |

The one remaining hitch is the first step's re-point, and it is deliberate — it is what keeps a single press immediate. It shows up as the first gap of ~80ms before the rest settle to ~60ms.

### Notes for review

- `startTransition` cannot do this job. The diff cursor lives in redux, and React always schedules external-store updates synchronously, so a transition never downgrades them. Measured as a no-op before going this way.
- Deferring rather than cancelling does not help either: every move finishes before the next key arrives, so nothing is ever dropped. Also measured as a no-op.
- Cursor timing alone read as flat while the stutter was still plainly visible. Long-task observation is the measure that matched what you can see.

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/Page.tsx
+++ b/apps/lite/ui/src/routes/project/$id/workspace/Page.tsx
@@ -30,7 +30,7 @@ import { useAddLocalRepository } from "#ui/components/useAddLocalRepository.ts";
 import { ResizeHandle } from "#ui/components/ResizeHandle.tsx";
 import { globalHotkeys, workspaceHotkeys } from "#ui/hotkeys.ts";
 import { writeLastOpenedProject } from "#ui/project.ts";
-import { useAppDispatch, useAppSelector } from "#ui/store.ts";
+import { useAppDispatch, useAppSelector, useAppStore } from "#ui/store.ts";
 import type { ProjectForFrontend, RefInfo } from "@gitbutler/but-sdk";
 import { useHotkey, useHotkeys, type UseHotkeyDefinition } from "@tanstack/react-hotkeys";
 import {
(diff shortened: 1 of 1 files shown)


samhh (commented):
`useEffect` is usually a code smell; most of them in our codebase were introduced by agents. See: [https://react.dev/learn/you-might-not-need-an-effect](<https://react.dev/learn/you-might-not-need-an-effect>)

In this case rather than playing with timers I'd try deferring to React with another `useDeferredValue`. Having said that I'm not sure we want the details pane to be lagging behind (IIUC), it should be able to immediately render the header.

Possibly I need to sit down with this properly for a bit to get a better holistic view.

mtsgrd:
Makes sense, thank you.

PR #15462, opened 2026-08-20
Tidy the file lists' scroll affordances

PavelLaptev:
Tidy up the scroll affordances on the two file lists — the uncommitted list in the workspace and the file tree in the details pane.

### The fade ramp only shows once you scroll

The gradient at the bottom of the uncommitted list appeared whenever the list overflowed, so an untouched list already showed a fade. It now waits until the list is actually scrolled away from the top (`scrollable: top` *and* `scrollable: bottom`).

Its pseudo-element also only pulled back 4px of its 14px height, leaving a permanent 10px gap under the last row whether or not the ramp was visible. It now cancels its full height, so it claims no flow space and overlays the rows instead. The scroll padding is kept at the ramp's full height so keyboard navigation doesn't park the selected row underneath it.

### The separator runs edge to edge

The scrollbar reserves a gutter that content cannot paint into, so the sticky separator stopped a scrollbar's width short of the panel edge. The gutter's own band and hairline used to make that read as deliberate; on lists whose rows share the panel surface it just looked like a notch.

`.scrollerWithSeparator` now flattens the gutter and has the scrollbar track paint the missing stretch itself, under the same scroll state as the sticky half, so both halves appear and disappear as one line. `.scrollerFlatGutter`, briefly a separate class, was folded in — it had exactly the same two call sites, and the flat gutter is what makes the two-half line necessary in the first place.

`.scroller` keeps its banded gutter by default: the stacks, upstream, and branches lists put `--bg-1` cards on a `--bg-2` panel, where that hairline still marks where the content surface ends.

### Verification

The scrollbar-track half rests on a container query reaching `::-webkit-scrollbar-track:vertical`, which is unusual enough to be worth checking rather than assuming. Rendered the shipped rule verbatim in Chromium with the gutter confirmed reserved (190px client vs 200px offset): nothing at rest, unbroken edge-to-edge line when scrolled. Note that macOS Chrome uses overlay scrollbars by default, so reproducing this needs `-AppleShowScrollBars Always` passed to the browser process — otherwise `::-webkit-scrollbar` never renders and every case looks identical.

Typecheck and Prettier pass. Not yet eyeballed in a running build.

### DIFF
--- a/apps/lite/ui/src/components/ui.module.css
+++ b/apps/lite/ui/src/components/ui.module.css
@@ -30,8 +30,18 @@
 	overscroll-behavior: contain;
 }
 
-/* Shows a separator line at a scroller's top edge while it is scrolled down. */
+/* Shows a separator line at a scroller's top edge while it is scrolled down,
+   and keeps the scrollbar gutter free of a band of its own so rows and
+   scrollbar sit on the same surface, with only the thumb to show for it.
+
+   The two go together: the gutter is reserved, content cannot paint into it,
+   so the sticky line stops a scrollbar's width short of the edge. The track
+   paints that last stretch itself, under the same scroll state, and the two
+   halves appear and disappear as one line. */
 .scrollerWithSeparator {
+	--scrollbar-track-bg: transparent;
+	--scrollbar-track-border: transparent;
+
 	container-type: scroll-state;
 
 	&:before {
@@ -47,6 +57,12 @@
 
 		@container scroll-state(scrollable: top) {
 			opacity: 1;
+		}
+	}
+
+	&::-webkit-scrollbar-track:vertical {
+		@container scroll-state(scrollable: top) {
+			border-top: 1px solid var(--border-section);
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


PavelLaptev:
### Scroll affordances — before / after

Each image shows the list **at rest** (left) and **scrolled** (right).

**Before**

before

**After**

after

Three differences:

- **At rest** — the ramp used to fade `util_1.ts` before the list had been touched; it now stays clear until you scroll.
- **At rest** — the gutter's grey band and hairline are gone on lists whose rows share the panel surface; only the thumb remains.
- **Scrolled** — the separator used to stop a scrollbar's width short at the top right; it now runs edge to edge, the scrollbar track painting the stretch content cannot reach.

The 10px gutter stays reserved on both sides, so this is like-for-like.

> [!NOTE]
> These are **not** captures of the running app. The `e2e/tests/screenshots.spec.ts` harness needs a local Electron run, so this is a standalone page rendered in Chromium from the branch's actual CSS — base rules taken from the pre-change commit, head rules from the branch, with the resolved design tokens and `global.css`'s scrollbar rules. The rows are stand-ins. It demonstrates the CSS behaviour; it does not verify the surrounding app layout.
>
> macOS Chrome defaults to overlay scrollbars, which would hide the effect entirely — the browser was launched with `-AppleShowScrollBars Always`, and the reserved gutter asserted (190px client vs 200px offset) so neither side silently loses its scrollbar.

PR #15463, opened 2026-08-20
Give Lite more than one build target

mtsgrd:
`apps/lite` was built as a single app: one vite config, one entry, one electron binary. This lets it build **several targets from the same `ui/src`**, so the same workspace UI can run somewhere other than the desktop app — the first of those being a panel for an agent harness.

**The shape**

- `ui/src` becomes the shared library, and each target is an entry that composes it. The renderer, preload and main binaries stay exactly as they were; the harness adds two more: a panel bundle (a single self-contained IIFE, React 19 and the diff worker inlined) and a node host bundle.
- Targets diverge by **composition, not conditionals**. The rule that keeps it honest: if anything in `ui/src` ever needs to ask which host it is running on, a seam is missing.

**The panel**

- runs the app's *own* router over a memory history — no URL bar, but search validation, the project guard and the watcher-arming loader all run exactly as they do in the app
- `Panel` is the workspace route's component: the app's lists, minus the Details pane and window chrome
- the host answers every endpoint through the shared endpoint table, resolves which project to show (deepest worktree wins), and runs real project watchers with per-subscription event channels

**What it asked of shared code** — deliberately small:

- `routes.tsx` exports `createRouteTree({ workspace })`, so each entry names its own workspace surface (the app passes `Page`)
- `createAppRouter` takes an optional history
- `Page`'s applied-list derivation moved into `applied-address-space.ts`, so a second target can share it without importing the page

**Verification**

A jsdom rig mounts the real panel over a fake transport: the lists render from fixtures, a failed push raises its declared toast, and a watcher event refreshes the files. `pnpm test` runs it after the ui suite.

A stdio host CLI the plugin spawns, and the plugin sources themselves, ride along so the panel can be driven end to end.


PavelLaptev:
@mtsgrd can you please attach sreenshots if there are new components or CSS changes

mtsgrd:
@PavelLaptev sure, but this is a draft for now.

PR #15470, opened 2026-08-20
Preserve panel focus while dragging

samhh:
Closes GB-1794 & GB-1795.


samhh:
cc @PavelLaptev Let me know if you spot any new oddities with focus or drag and drop in particular.

PR #15482, opened 2026-08-21
Bring edit mode to lite

mtsgrd:
Edit mode checks a commit out into the working directory so its files can be edited with any tool, then rewrites the commit on save (rebasing whatever sat above it). The desktop app has had it for ages, and the CLI's `but resolve` flow uses it too — lite just could not see it: run `but resolve` in a terminal and the app would keep drawing a workspace that no longer existed.

**Backend: one word away**

- the six mode endpoints in `but-api` were already `#[but_api]` — they just gained `napi` generation, so they reach the SDK like every other call
- a new `OperatingMode` cache tag ties it together: the mode query provides it, the three mode mutations invalidate it (the mode reads a metadata file the watcher cannot see), and the `gitHead` watcher event — which previously invalidated nothing — now invalidates it too
- the `gitHead` event already carries the new head and mode, so lite pushes that payload straight into the cache instead of refetching. Entering edit mode from a terminal flips the app on its own

**Lite: a mode gate and a surface**

- while HEAD is parked on the edit ref, the workspace page swaps to an edit-mode surface: the commit's files with their conflict states, what changed since entering, save and cancel
- entry points: **Edit Commit** in the commit context menu, and an **Open Edit Mode** button on the conflict bar — which already said "these files can only be resolved in edit mode" without offering a way there
- guardrails: saving while conflicted files are untouched asks first, as does cancelling with changes on disk

**Verified live** against a fixture repo over CDP: enter → the surface swaps in by itself; edit a file on disk → the watcher refreshes the changed list; cancel → workspace restored, worktree clean. The save path reuses the same backend `but resolve finish` exercises.


PavelLaptev:
@mtsgrd screenshots please

mtsgrd:
## Screenshots

Edit mode is a **new screen**, so there is no "before" to compare against — this is the surface as the branch introduces it. Captured locally at `78e9e333e9`, this PR's head.

### Edit mode

The conflicted case: a commit whose change could not be applied, checked out with its markers on disk.

edit mode

Added to the screenshot catalogue (`apps/lite/e2e/tests/screenshots.spec.ts`), so later pull requests touching this screen get it automatically.

<details>
<summary>What this does not cover</summary>

Only the new surface was captured. The other eight catalogue surfaces were **not** compared, because the base pass needs this branch unapplied and that was blocked by unrelated uncommitted files in the working tree. So this says nothing about whether the branch changes any existing screen.

Also out of reach for this harness, as always: native context menus (they are OS windows), dark mode (the fixture seeds the light theme), and hover/focus states.
</details>

PR #15489, opened 2026-08-21
Show the file-row tooltip only when it earns its place

mtsgrd:
Every file row offered a hover tooltip that just repeated its own label. It now opens only when it can add something the row cannot show.

- the row measures its label (`scrollWidth` vs `clientWidth`) through a ResizeObserver, so window resizes and layout shifts re-decide whether the path is ellipsized
- any scroll dismisses the tooltip — it does not follow the list, and would otherwise linger outside the scroll container. The listener only exists while a tooltip is open, so a long file list holds at most one document listener rather than one per row
- conflicted uncommitted rows keep their get-out hint ("Resolve the conflict, then right-click → Mark as Resolved") in the tooltip

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/FileRow.tsx
+++ b/apps/lite/ui/src/routes/project/$id/workspace/FileRow.tsx
@@ -11,7 +11,7 @@ import { classes } from "#ui/components/classes.ts";
 import { changesFileHotkeys } from "#ui/hotkeys.ts";
 import { Toolbar, Tooltip } from "@base-ui/react";
 import { Match } from "effect";
-import type { ComponentProps, FC } from "react";
+import { type ComponentProps, type FC, useEffect, useRef, useState } from "react";
 import styles from "./FileRow.module.css";
 import treeStyles from "./FilesTree.module.css";
 import { Row, RowCheckbox, RowLabel, RowLabelContainer, RowToolbar } from "./Row.tsx";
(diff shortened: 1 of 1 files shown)


samhh:
Base UI tooltips support grouping whereby the open delay is skipped once you're already displaying a sibling. We've somewhat broken that, where now a user has to go through that delay again if they're not quick enough to move from A to C in this example:

```
- long row A
- short row B
- long row C
```

That aside, on the code front `useEffect` continues to be a smell. In this case we can probably get away with measuring once in an event handler, dropping much of the ceremony:

```diff
diff --git a/apps/lite/ui/src/routes/project/$id/workspace/FileRow.tsx b/apps/lite/ui/src/routes/project/$id/workspace/FileRow.tsx
index 198f10b6a0..f47ef6cb55 100644
--- a/apps/lite/ui/src/routes/project/$id/workspace/FileRow.tsx
+++ b/apps/lite/ui/src/routes/project/$id/workspace/FileRow.tsx
@@ -11,7 +11,7 @@ import { classes } from "#ui/components/classes.ts";
 import { changesFileHotkeys } from "#ui/hotkeys.ts";
 import { Toolbar, Tooltip } from "@base-ui/react";
 import { Match } from "effect";
-import { type ComponentProps, type FC, useEffect, useRef, useState } from "react";
+import { type ComponentProps, type FC, useRef } from "react";
 import styles from "./FileRow.module.css";
 import treeStyles from "./FilesTree.module.css";
 import { Row, RowCheckbox, RowLabel, RowLabelContainer, RowToolbar } from "./Row.tsx";
@@ -73,38 +73,29 @@ export const FileRow: FC<
 		uncommit,
 	});
 
+	const hasConflictHint = item._tag === "Conflict" && fileParent._tag === "UncommittedChanges";
 	// An uncommitted conflict is a state to get out of, so the row says how.
-	const rowTooltip =
-		item._tag === "Conflict" && fileParent._tag === "UncommittedChanges"
-			? `${relativePath} — Resolve the conflict, then right-click → Mark as Resolved`
-			: relativePath;
+	const rowTooltip = hasConflictHint
+		? `${relativePath} — Resolve the conflict, then right-click → Mark as Resolved`
+		: relativePath;
 	const lastSepIdx = relativePath.lastIndexOf("/");
 	const directoryPath = lastSepIdx !== -1 ? relativePath.slice(0, lastSepIdx) : null;
 	const fileName = lastSepIdx !== -1 ? relativePath.slice(lastSepIdx + 1) : relativePath;
-
-	// The tooltip only repeats the label, so it opens only when the label is
-	// ellipsized and cannot say it all. Measured as it opens, never stale.
-	const [rowTooltipOpen, setRowTooltipOpen] = useState(false);
 	const rowLabelRef = useRef<HTMLDivElement>(null);
-	const labelIsTruncated = () => {
-		const label = rowLabelRef.current;
-		return label !== null && label.scrollWidth > label.clientWidth;
-	};
-
-	useEffect(() => {
-		if (!rowTooltipOpen) return;
-		// A tooltip does not follow the list, so any scroll dismisses it. The
-		// listener exists only while a tooltip is open: one per list, not row.
-		const dismiss = () => setRowTooltipOpen(false);
-		document.addEventListener("scroll", dismiss, true);
-		return () => document.removeEventListener("scroll", dismiss, true);
-	}, [rowTooltipOpen]);
 
 	return (
 		<Tooltip.Root
-			open={rowTooltipOpen}
-			onOpenChange={(open) => setRowTooltipOpen(open && labelIsTruncated())}
 			disableHoverablePopup
+			onOpenChange={(open, eventDetails) => {
+				const label = rowLabelRef.current;
+				if (
+					open &&
+					!hasConflictHint &&
+					(label === null || label.scrollWidth <= label.clientWidth)
+				) {
+					eventDetails.cancel();
+				}
+			}}
 		>
 			<Tooltip.Trigger
 				// We pass the ID here instead of including it with the other props as a

```

Further, though it's unlikely to matter in this case, I'd be generally wary of reading calculated layout information in JS as it can contribute to layout thrashing.

PR #15492, opened 2026-08-21
Various fixes

samhh:
So much scope for regressions to have crept in here, though I did do a lot of manual validation.


samhh:
cc @mtsgrd 1a5f6865948075abbc7c480146ec8f6f7a8b21ce may be of interest. Diff range state is probably too hot to naively put right into the URL ([or maybe not?](https://nuqs.dev/docs/limits)), though I think with this simplified shape it should be easier to make some sort of interop happen.

PR #15503, opened 2026-08-22
Fix the upstream tab's incoming count

PavelLaptev:
The number on the Upstream tab counted two different things depending on which
page was shown, and went stale once the tab had been visited.

**It changed when you clicked the tab.** `incomingCount` fell back to the
graph's `commitsAhead` while another page was shown, and used the first-parent
row count once the tab was open. `commitsAhead` sums every segment's commits,
so a merged PR contributed its merge commit *plus* everything the merge brought
in — a one-commit PR showed as `2` from the workspace tab and `1` from the
upstream tab, listing a single row either way. It now always counts the
first-parent rows, so the badge, the page label, and the rows agree.

**It went stale after the first visit.** Both queries behind the hook were
`enabled: active`. Once the tab had been opened, the cached listing kept
winning the fallback, and the invalidation that follows a `gitFetch` reached no
enabled observer — so commits that arrived while the workspace tab was shown
did not reach the badge until the tab was clicked. The listing query now stays
enabled; the item and address-space derivation is still gated on the tab being
active, so an inactive tab only pays for the count. The listing itself is a
bounded local first-parent walk, already refetched on every fetch event.

**The badge caps at `99+`.** It sits inside a tab toggle, where a third digit
takes width from the tab labels, and being a hundred commits behind is
ordinary. The upstream page still states the exact count.

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/Sidebar.tsx
+++ b/apps/lite/ui/src/routes/project/$id/workspace/Sidebar.tsx
@@ -45,6 +45,14 @@ const adjacentPage = (tab: PageId, offset: -1 | 1): PageId => {
 	return assert(pageOrder[(index + offset + pageOrder.length) % pageOrder.length]);
 };
 
+/**
+ * Counts past this are shown as `99+`: the badge sits inside a tab, where a
+ * third digit takes width from the tab labels, and at that size the number is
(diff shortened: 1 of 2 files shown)


PavelLaptev:
cc @krlvi

PR #15505, opened 2026-08-23
Sort uncommitted files by last modified and show how recently each one changed

mtsgrd:
Adds a recency view to the uncommitted file list, so the files you just touched are the ones you see first.

### What it does

- **Sort by Last Modified** in the Uncommitted header's menu (next to View as List/Tree) sorts the list newest-first, off by default. It only changes the order — the list/tree shape setting stays in force.
- Every row carries an abbreviated age badge (`now`, `3m`, `2h`, `5d`, `3w`), sitting right of the kebab column.
- The badge's **contrast fades with age** rather than disappearing: full strength for a change just written, easing to a floor that stays readable. The curve is logarithmic, because the gap between five minutes and an hour matters far more when scanning than the gap between five days and a week.
- Rows older than an hour add the full time ago (`5 hours ago`) to their hover tooltip.
- Changes newer than about a minute pulse, via an inset-shadow animation that leaves hover and selection fills visible underneath. The clock ticks every 30s, so the pulse can outlast the minute by up to that much — it is decoration, and per-row timers would cost more than the imprecision.

### Where the times come from

`changesInWorktree` and the watcher's worktree payload now carry `modificationTimes` — epoch milliseconds keyed by the change's path. They are gathered in one pass in `but-core`'s UI conversion, one `symlink_metadata` per changed file, rather than a query per row.

A path with nothing on disk to stat is simply absent, so **deletions and files removed from the worktree carry no time** and sort last. That is deliberate: a file that no longer exists has no honest last-modified time, and the nearest proxy (the parent directory's mtime) is shared across sibling deletions and silently wrong if anything else in the folder changed.

### Notes

- `modification_times` is a `BTreeMap`, so the serialized key order is stable.
- The `but-core` snapshot test asserts *which* paths carry a time; the values themselves are wall-clock and match as wildcards, so the test does not depend on fixture checkout time.
- The ticking clock that ages the labels mounts only while the recency view is on.

### DIFF
--- a/apps/lite/harness/browser/Panel.tsx
+++ b/apps/lite/harness/browser/Panel.tsx
@@ -65,6 +65,8 @@ export const Panel: FC = () => {
 		filter: uncommittedFilesFilter,
 		mode: uncommittedFilesDisplayMode,
 		collapsedDirectories: uncommittedFilesCollapsedDirectories,
+		// The panel has no recency toggle, so it lists in path order.
+		recentFirst: false,
 	});
 	const uncommittedAddressSpace = fileTreeAddressSpace(uncommittedFileRows);
 
--- a/apps/lite/harness/tests/fixtures.ts
+++ b/apps/lite/harness/tests/fixtures.ts
@@ -88,6 +88,8 @@ export const fixtureFileChange = (path: string): TreeChange => ({
 export const fixtureWorktreeChanges = (changes: Array<TreeChange>): WorktreeChanges => ({
 	changes,
 	ignoredChanges: [],
+	// Fixed mtimes so modifiedAtMs derivations are deterministic.
+	modificationTimes: Object.fromEntries(changes.map((change) => [change.path, AUTHORED_AT])),
 	assignments: [],
 	assignmentsError: null,
 	dependencies: null,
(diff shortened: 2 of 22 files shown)


mtsgrd:
## 📸 UI screenshots

Captured 1 surfaces, 1 changed.

> **Captured ad hoc.** This state — the recency toggle on, over files of staged ages — is not in the screenshot catalogue and is **not covered going forward**; the spec that produced it was deleted after the run. Treat it as a photograph of this change, not a regression guarantee.

> **Slightly stale header:** since this capture, the clock toggle visible in the header moved into the ⋯ menu as “Sort by Last Modified” (off by default). The rows, badges and fade are unchanged.

Clipped to the uncommitted panel alone, so the branch and commit lists are out of frame. The file mtimes are stamped onto the fixture (20s, 8m, 95m, 5h and 30h old) rather than waited for, so the badges read `now`, `8m`, `1h`, `5h` and `1d`, and the contrast fade across that range is visible in one shot.

### recency-view

krlvi:
This has some performance implications. I'd suggest keep the `modification_times()` in core, but remove it from `but_core::diff::ui::worktree_changes()`. 
On the API you can add `worktree_changes_with_modification_times` which composes `but_core::diff::worktree_changes` and `but_core::diff::ui::modification_times`.

Otherwise looks cool, and maybe useful as well

PavelLaptev:
@mtsgrd can we move the toggle button into a kebab menu for now. And make this mode off by default?
If you want to merge it. If not you can wait until I check it

krlvi (approved):
lets try this

PR #15526, opened 2026-08-24
Provide source links in but-ts output

samhh:
Allows frontend to go straight to the source rather than grepping around. See commit. Soft stack dependency to avoid codegen conflict.


Caleb-T-Owens (approved):

samhh:
If this causes any problems, anyone feel free to revert!

PR #15529, opened 2026-08-24
feat(cli): _diff2 support for worktrees

davidpdrsn:


davidpdrsn:
cc @Caleb-T-Owens this just aligns the behavior between `but diff some-worktree` and `but _diff2 some-worktree`. It doesn't address the lack of support for worktree diffs we discussed during the eng meeting on august 24

PR #15532, opened 2026-08-24
Remove segment IDs from projection debug output

jonathantanmy2:
This makes it easier to test alternative implementations of graphs that do not create the segments in the same order (but still create the same graph).


krlvi (approved):

krlvi:
seems reasonable

PR #15545, opened 2026-08-25
Lite UI fixes: keyboard keys, textarea padding, menu label

PavelLaptev:
A few small UI corrections in Lite, plus a design-core bump.

- **Keyboard keys**: `text-box-trim` on `.key` was inert because the element was `inline-flex` — the declaration only applies to block containers and inline boxes, and the text child landed in an untargetable anonymous flex item. The key is now a block centered with `text-align`; `.keys` still handles vertical centering. `FocusScopeKbd` pins its key to a fixed 15px square where the base padding fought the trimmed cap-height box, so padding is zeroed there and the glyph is centered with `align-content`. The key face is also restyled: flat tinted background and a single outline ring instead of the drop shadow, with a `min-width` so single glyphs stay square-ish.
- **Textarea padding**: the textarea was inheriting the single-line input's 8px inline / 6px block padding; the design specifies 12px on every edge.
- **Stack menu label**: "Unapply Whole Stack" → "Unapply Stack" (test expectation updated alongside).
- **Dependency**: bump `@gitbutler/design-core` 3.12.5 → 3.13.2 in the Lite app.

No behavior changes beyond the styling and the menu wording.

### DIFF
--- a/apps/lite/package.json
+++ b/apps/lite/package.json
@@ -136,7 +136,7 @@
 		"@base-ui/react": "^1.6.0",
 		"@base-ui/utils": "^0.3.1",
 		"@gitbutler/but-sdk": "workspace:*",
-		"@gitbutler/design-core": "3.12.5",
+		"@gitbutler/design-core": "3.13.2",
 		"@pierre/diffs": "^1.3.5",
 		"@pierre/theming": "^1.0.1",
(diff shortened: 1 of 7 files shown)


PavelLaptev:
No need for screenshots here, the changes are purely cosmetic or in Storybook.

PR #15559, opened 2026-08-25
Lite: add a Snackbar component

PavelLaptev:
One line of consequence, floated over the surface that caused it: a glyph,
a sentence, and optionally a close button fenced off behind a divider.
Where a Toasts card states a title and a description, a snackbar has only
the sentence, so it can sit close to the thing it is about.

Every variant wears the same surface — the info, danger and safe glyphs
are the only thing saying how the news landed.

Screenshot 2026-08-25 at 20-29-56

### DIFF
--- a/apps/lite/ui/src/components/Snackbar.module.css
+++ b/apps/lite/ui/src/components/Snackbar.module.css
@@ -1,0 +1,66 @@
+.snackbar {
+	display: inline-flex;
+	column-gap: 8px;
+	align-items: center;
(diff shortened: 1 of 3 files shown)


PavelLaptev:
Folded into #15560. The component had no production consumer on this branch — its only importer lives one branch up — so `knip --strict` correctly flagged `Snackbar.tsx` as an unused file. Shipping it together with the toolbar work that uses it.

PR #15570, opened 2026-08-26
lite: default diff view to selected file only

krlvi:
Flips the `unidiff` GUI setting default so the details pane shows only the selected file's diff by default. Users who already toggled the setting keep their stored choice.

### DIFF
--- a/apps/lite/ui/src/settings.ts
+++ b/apps/lite/ui/src/settings.ts
@@ -32,7 +32,7 @@ export const defaultSettings = {
 		dark: "github-dark-default" satisfies BundledTheme,
 	},
 	theme: "system",
-	unidiff: true,
+	unidiff: false,
 } satisfies Partial<GUISettings>;
 
 export const clampAutoFetch = (ms: number): number =>


samhh:
Are there any other review tools that default to a single file? It strikes me as very uncommon.

I flipped the files list to tree by default since there seems to be a consensus amongst us. Is this the same excluding myself?

krlvi:
> Are there any other review tools that default to a single file? It strikes me as very uncommon.
> 
> I flipped the files list to tree by default since there seems to be a consensus amongst us. Is this the same excluding myself?

sorry i was a bit trigger happy here. I discovered yesterday this was possible and was super happy about it so.... made it the default. 
The main issue i felt is that it was never clear what I am looking at at any given time... so maybe treated a symptom rather than the cause

krlvi:
we can revert this, i pressed buttons before reading messages 🙈

samhh:
Let's take the temperature in the meeting in half an hour and then decide. 🙂

krlvi:
btw one think i realized is that the "single file at a time / selected file only" diff should probably only apply to when the file list is visible - otherwise it's confusing

samhh:
> btw one think i realized is that the "single file at a time / selected file only" diff should probably only apply to when the file list is visible - otherwise it's confusing

Counter-example: I want to review one file at a time for whatever reason, and I have limited viewport width.

krlvi:
> > btw one think i realized is that the "single file at a time / selected file only" diff should probably only apply to when the file list is visible - otherwise it's confusing
> 
> Counter-example: I want to review one file at a time for whatever reason, and I have limited viewport width.

wouldnt the UI then mislead you that there is only one file in the commit? For the space constrained option, the user can fold other files i guess

samhh:
> wouldnt the UI then mislead you that there is only one file in the commit? For the space constrained option, the user can fold other files I guess

🤔

cc @PavelLaptev Is there another way we could solve this without conditionally overriding the user's setting? Or is that actually preferable?

PavelLaptev:
There is a ticket for it: https://linear.app/gitbutler/issue/GB-1910/lite-switch-between-unified-and-single-file-diff-from-the-details

I think if we move this out of settings and onto the diff itself, it'd be more discoverable. Some of these other settings could live outside settings too:

image

Similar to how Linear does it:

image

> btw one think i realized is that the "single file at a time / selected file only" diff should probably only apply to when the file list is visible - otherwise it's confusing

Let's not overcomplicate this with extra conditions. What happens if I'm looking at my diffs with the file list open and then hide it? Coupling the setting to file-list visibility just moves the confusion: hide the list, the pane silently expands to everything, and now "selected file only" means two different things depending on an unrelated panel. That's not really a setting anymore.

And "just fold other files" doesn't answer @samhh's counter-example — folding is manual work you redo on every commit; the setting is set once.

I'd rather we work on the actual issue: "which file am I on". Here's how Linear solves it:

Screenshot 2026-08-26 at 19-50-00

https://github.com/user-attachments/assets/bcdfbad4-9c3b-4dea-b75b-12c6ed3e452b

Or Dropbox Paper:

https://github.com/user-attachments/assets/4a44295d-19a9-4dad-951c-af3d749cb82c

Might even be a nice minimap replacement.

Another option for single-file mode: add prev/next controls and show which file you're on. Quick draft the agent made:

image

krlvi:
alright! lets roll back the changes i did and focus on the actual problem

PR #15577, opened 2026-08-26
lite: apply selected-file diff only while the file list is visible

krlvi:
Follow-up to #15570. The selected-file-only diff mode relies on a visible file list to say which file the pane is showing. When that list is hidden — the files panel toggled off, or the sidebar's uncommitted rows behind full-window mode — a lone unlabelled diff is confusing, so the pane falls back to rendering every file.

The stored setting alone still drives the scroll-on-activation behavior in the page, since file activation can only come from a visible list, where the pane follows the setting.


samhh:
https://github.com/gitbutlerapp/gitbutler/pull/15570#issuecomment-5426702036

PR #15583, opened 2026-08-26
feat(but): add support for resolving committed hunk IDs

slarse:
Adds preliminary support for resolving committed hunk IDs. Committed hunk IDs are currently not exposed anywhere, you just need to know what they are to use them. The only command that can currently use committed hunk IDs is `_expand`, so meaningful support in e.g. `squash` and `move` remains to be implemented.

The current implementation is _very_ inefficient if multiple IDs need to be resolved, as every ID is individually parsed and independently requires tree and blob level diffing, even if they belong to the same committed file. We need some form of caching moving forward. I'm thinking a per-command explicit caching strategy for the commands that do resolve multiple IDs, and those that don't simply need not be concerned with caching. I don't want hidden caching as it's so easy to create subtle bugs if the underlying state changes, and you don't even know there's a cache involved.

We're also currently dropping rename information in parsing (it's present in the TreeChange) - I will add that in as it's needed.

### Review notes

@davidpdrsn I think it's especially important you check this from the perspective of the TUI.

* I've added in basically no support for the TUI yet. In a couple of places I've just added a line or two that would be easy to forget to add later, while now it's obvious as the `CommittedHunk` variant is missing.
* `CommittedHunk` is a distinct type from `CommittedFile`. This fits the parsing patterns much better than making it a single `CommittedHunkOrFile`. I did consider it might not work with how selection operates in the TUI, but at a glance it looks like each selected uncommitted hunk becomes its own `UncommittedHunkOrFile` that are coalesced only later, which we could adapt to work for distinct `CommittedHunk` and `CommittedFile`.
* Rust nit-picks welcome as usual

@Caleb-T-Owens Could you quickly browse over from a worktree perspective? In my mind, as committed files are already scoped to commits, these changes should have no impact on worktrees. Also trying it out in practice, resolution seems fine.

### DIFF
--- a/crates/but/src/id/mod.rs
+++ b/crates/but/src/id/mod.rs
@@ -209,7 +209,7 @@ fn short_ids_from_tree_changes(
     tree_changes: Vec<but_core::TreeChange>,
 ) -> anyhow::Result<Vec<(NonEmpty<but_core::TreeChange>, ChangeId, ShortId)>> {
     let FileInfo { changes } = FileInfo::from_tree_changes(tree_changes)?;
-    let mut short_ids: Vec<(NonEmpty<but_core::TreeChange>, ChangeId, ShortId)> = Vec::new();
+    let mut short_ids = Vec::new();
     for (path, changes) in changes {
         // Committed files are namespaced under their commit's ID, so they never
         // compete with the uncommitted namespace and need no source of their own.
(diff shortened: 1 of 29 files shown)


crates/but/src/id/mod.rs
```diff
@@ -232,33 +232,136 @@ fn short_ids_from_tree_changes(
...
+    committed_file: CommittedFileId,
+    /// The short CLI ID for this file (typically 2 characters)
+    short_id: ShortId,
+    /// Tree change from the commit's first-parent diff.
+    ///
+    /// This should in practice always contain a single change, but according to the docs on
+    /// [FileInfo::changes], there have been cases where a single path resolved to multiple tree
+    /// changes. As we currently cannot structurally guarantee this does not happen, we need this
+    /// awkward list here.
+    tree_changes: NonEmpty<but_core::TreeChange>,
```
slarse: I'm very uncertain if we want to propagate this "should not but has been observed to happen" scenario anymore.

crates/but/src/id/mod.rs
```diff
@@ -1655,6 +1758,23 @@ pub struct UncommittedHunkOrFile {
...
 
+#[derive(Debug, Clone)]
+/// A committed file or hunk.
+pub struct CommittedHunk {
+    /// The committed file to which the hunk belongs.
+    pub committed_file: CommittedFileId,
+    /// The short ID of this hunk.
+    pub id: ShortId,
+    /// The hunk itself.
+    pub hunk: but_core::SingleHunk,
```
slarse, line 1769: @Caleb-T-Owens `but_core::SingleHunk` is expressly documented to be for uncommitted changes. But... Seems to work fine for committed changes, too. Documentation inaccuracy or am I committing sins against all of butmanity and there's a different type to use?

crates/but/src/id/tests.rs
```diff
@@ -1678,6 +1789,51 @@ fn uncommitted_path() -> anyhow::Result<()> {
     Ok(())
 }
 
+/// This test represents a bad state: This shouldn't really happened, but if it does we don't want
+/// to just lose changes.
```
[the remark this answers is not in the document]
slarse: Oh sod off.

crates/but/src/id/mod.rs
Caleb-T-Owens: Single hunk is fine for use to represent any hunk.

The main intention of the type was to avoid using `struct HunkAssignment`.

I just let my LLM appease the mandatory doc linters.

Caleb-T-Owens (approved):
LGTM.

WRGT worktrees, ideally commits in worktrees should be handled identically to normal worktrees which seems to be what we've got here.

davidpdrsn (approved):
Great work! I'm very excited about this :)

> I've added in basically no support for the TUI yet.

Sounds good! Now that the TUI and CLI share implementations for mutations it hopefully wont be too hard to add support piece by piece.

> CommittedHunk is a distinct type from CommittedFile. This fits the parsing patterns much better than making it a single CommittedHunkOrFile.

Honestly I think having separate types is way nicer. I'm not sure why we have a single UncommittedHunkOrFile type.

PR #15591, opened 2026-08-27
Fix onboarding without Git identity

krlvi:
Allow onboarding to create its internal references when no repository identity is configured. Existing committer attribution is preserved, with GitButler used only as the missing-identity fallback.

Linear: [GB-1916](https://linear.app/gitbutler/issue/GB-1916/onboarding-silently-stalls-when-setting-the-target-branch-without-a)

### DIFF
--- a/crates/gitbutler-branch-actions/src/base.rs
+++ b/crates/gitbutler-branch-actions/src/base.rs
@@ -12,7 +12,7 @@ use but_error::{Code, bail_precondition};
 use but_graph::FirstParent;
 use gitbutler_project::{FetchResult, Project};
 use gitbutler_reference::{Refname, RemoteRefname};
-use gitbutler_repo::first_parent_commit_ids_until;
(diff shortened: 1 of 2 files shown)


crates/gitbutler-branch-actions/src/base.rs
```diff
@@ -268,10 +268,10 @@ pub(crate) fn set_base_branch(
             meta.set_workspace(&workspace)?;
             drop((workspace, meta));
             if !branch_matches_target {
-                repo.reference(
```
krlvi, line 271: @Byron in theory it could be nice if gix repo.reference could take a committer as an optional argument that would be in the reflog

Byron: There is just the thing and I will post a follow-up: `gitoxide.committer.nameFallback` and `gitoxide.committer.emailFallback` can just be configured and it will automatically use these anywhere it needs a committer.

PR #15602, opened 2026-08-27
Fix branch move checkout in single-branch mode

estib-vega:
## Summary

- route branch moves through the checkout-aware API because transactions cannot persist ad-hoc branch order or apply the new tip checkout
- add CLI regression coverage matching the desktop single-branch move scenarios, with before/after `but status` snapshots
- add a reusable three-branch ad-hoc fixture

## Impact

Moving a branch to the top of a single-branch stack now checks out that branch and keeps the complete reordered stack projected.

fixes GB-1847

### DIFF
--- a/crates/but/src/command/legacy/move.rs
+++ b/crates/but/src/command/legacy/move.rs
@@ -918,12 +918,31 @@ fn resolve_sources(
     }
 }
 
+/// Branch moves bypass `but-transaction`, which cannot persist the branch order or check out the
+/// new tip returned by single-branch moves, and use the checkout-aware `but-api` path instead.
 pub fn run(
     ctx: &mut Context,
     meta: &mut impl RefMetadata,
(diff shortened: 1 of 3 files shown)


crates/but/src/command/legacy/move.rs
```diff
@@ -918,12 +918,31 @@ fn resolve_sources(
...
+        )?;
+        return Ok((
+            MoveOutcome::StackBranch {
+                source_branch: op.source_branch.clone(),
+                target_branch: op.target_branch.clone(),
+            },
+            result.workspace,
+        ));
+    }
+
```
estib-vega: @davidpdrsn we need to route the branch move outside the but-transaction to ensure that the HEAD is updated.

Ideally the but-transaction can update HEAD eventually, but now this fixes moving branches in SBM

davidpdrsn (approved):

PR #15603, opened 2026-08-27
fix(but): clear error when an anonymous segment is used as a `branch new` target

### LINKED ISSUE #15596: `but branch new <name> --above <segment>` fails with "Reference name cannot end with a slash" when the segment has no branch reference
### Version

0.22.1

### Operating System

Linux

### Distribution Method

deb (Linux)

### Describe the issue

**Context**

I ran `git reset --hard origin/feature/docker` in a repo that was not yet under GitButler management, then ran `but` to trigger the interactive setup. GitButler switched the repo to `gitbutler/workspace` and created the workspace, but the 4 pre-existing commits from `feature/docker` ended up in an anonymous stack segment (shown as `g0` in `but status`), with no branch reference attached:

```
╭┄ zz [uncommitted] (no changes)
┊
┊╭┄ g0
┊●   ovo Auto-refresh dependencies
┊●   krl Minor improve container
┊●   qko Update volumes for update hot reload
┊●   xro Migrate dev stack to Docker Compose
├╯
...
```

`but unapply g0` correctly refuses this, with a helpful hint:

```
Error: Cannot unapply segment 'g0' because it has no branch reference

Hint: Run `but branch new <name> --above <commit-id>` with the segment's top commit ID from `but status`, then run `but unapply <name>`.
```

**Bug**

Following that hint literally but passing the *segment label* (`g0`) instead of the *commit ID* (as one might reasonably do, since `g0` also appears as a valid target elsewhere, e.g. in `but status` output) fails with an unrelated, unhelpful error:

```
$ but branch new to-delete --above g0
Error: Reference name cannot end with a slash
```

This error message gives no indication of what went wrong or how to fix it. It appears `--above` accepts `g0` as a `BRANCH_OR_COMMIT` argument without validating that it resolves to something with an actual name, then tries to build a git ref (something like `refs/heads/<empty>/...` or similar) from the segment's empty/anonymous name, which naturally ends in `/` and is rejected by git's ref-name validation.

1. Create a stack segment with no branch reference (e.g. via `git reset --hard <sha>` before `but setup`, or any path that leaves commits applied without a branch name — reproduces as segment `gN` in `but status`).
2. Run `but branch new <name> --above <segment-label>` (e.g. `--above g0`), using the *segment label* rather than a commit ID or branch name.
3. Observe: `Error: Reference name cannot end with a slash`, with no hint pointing back to using a commit ID instead.

Either:
- `--above`/`--below` should accept a bare segment label like `g0` as equivalent to "top commit of that segment" (since it's already displayed as a valid-looking identifier in `but status` and is the identifier used elsewhere, e.g. `but unapply g0`), or
- If a segment label is not a valid target for `--above`, the CLI should reject it up front with a clear error (e.g. "`g0` is a segment with no branch reference; use its top commit ID instead"), rather than surfacing an internal ref-construction failure.

cmatheo:
## Summary

`but branch new <name> --above/--below <target>` crashed with a confusing
"Reference name cannot end with a slash" error when `<target>` was the label
of an anonymous stack segment (e.g. `g0`) — a segment that has no branch
reference. The label resolved to a branch with an empty name, and that empty
name was fed straight into `refs/heads/` ref construction, producing an
invalid, unhelpful error.

This returns a clear `bad_input` error with a hint instead, matching the
existing pattern already used for the same situation in `but unapply`.

Fixes #15596

## Changes

- `resolve_above_below_target` (`crates/but/src/command/legacy/branch/new.rs`)
  now detects an empty-named branch resolution (anonymous segment) and
  returns:
  ```
  Cannot use '<segment>' as a branch target because it is an anonymous stack segment

  Hint: Run `but status` to find the segment's top commit, then run `but branch new <name> --above/--below <commit-id>` with that commit ID
  ```
- Added a regression test using the existing `one-stack-anonymous-segment`
  fixture.

## Test plan

- [x] `cargo check -p but --all-targets`
- [x] `cargo test -p but branch::new::`
- [x] `cargo test -p but` (689 passed, 0 failed)

### DIFF
--- a/crates/but/src/command/legacy/branch/new.rs
+++ b/crates/but/src/command/legacy/branch/new.rs
@@ -137,7 +137,7 @@ fn resolve_above_below_target(
     id_map: &IdMap,
     target: CliIdArg,
 ) -> CliResult<NewStackedBranchTarget> {
-    let target = target
+    let resolved = target
         .resolve_in_workspace(
             repo,
             id_map,
@@ -146,9 +146,16 @@ fn resolve_above_below_target(
         )?
         .into_branch_or_commit()?;
 
-    Ok(match target {
+    Ok(match resolved {
         BranchOrCommit::Commit(commit) => NewStackedBranchTarget::Commit(commit),
         BranchOrCommit::Branch(branch_arg) => {
+            if branch_arg.0.is_empty() {
+                return Err(bad_input(format!(
+                    "Cannot use '{target}' as a branch target because it is an anonymous stack segment"
+                ))
+                .hint("Run `but status` to find the segment's top commit, then run `but branch new <name> --above/--below <commit-id>` with that commit ID")
+                .into());
+            }
             NewStackedBranchTarget::Branch(branch_arg.resolve_local_branch_name()?)
         }
     })
--- a/crates/but/tests/but/command/branch/new.rs
+++ b/crates/but/tests/but/command/branch/new.rs
@@ -54,6 +54,24 @@ Hint: Run `but status` for applicable targets.
 }
 
 #[test]
+fn rejects_anonymous_segment_as_target() {
+    let env =
+        Sandbox::init_scenario_with_target_and_default_settings("one-stack-anonymous-segment");
+    env.setup_metadata(&["A"]);
+
+    env.but("branch new --above g0 new-branch")
+        .assert()
+        .failure()
+        .stderr_eq(str![[r#"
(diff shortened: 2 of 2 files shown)


slarse:
Hi @cmatheo,

Closing for the same reason as in https://github.com/gitbutlerapp/gitbutler/pull/15604#issuecomment-5498965437. Thanks for your contribution - it made us finally act on this oversight!

PR #15604, opened 2026-08-27
feat(but): accept anonymous segment labels for `branch new --above/--below`

cmatheo:
feat(but): accept anonymous segment labels for `branch new --above/--below`

## Summary

Stacked on #15603 (clear error message for the same situation).

An anonymous stack segment (like `g0`) has no branch reference, so even with
the clearer error message from the previous PR, `but branch new --above/--below
g0` still couldn't be used directly — users had to run `but status`, find the
segment's top commit ID by hand, and pass that instead.

This resolves an anonymous segment label directly to its top commit, so it
can be used as a target the same way a branch or commit ID can.

Relates to #15596

## Changes

- Added `IdMap::anonymous_segment_top_commit` (`crates/but/src/id/mod.rs`),
  which resolves a CLI element to the top-most commit of the anonymous
  segment it names, if any.
- `resolve_above_below_target` (`crates/but/src/command/legacy/branch/new.rs`)
  tries this resolution first, before falling back to the normal
  branch/commit resolution (whose empty-name guard from the previous PR
  remains as a safety net for the theoretical case of an anonymous segment
  with no commits).
- Updated the regression test added in the previous PR to assert success
  instead of the old error.

## Test plan

- [x] `cargo check -p but --all-targets`
- [x] `cargo test -p but branch::new::`
- [x] `cargo test -p but` (689 passed, 0 failed)

### DIFF
--- a/crates/but/src/command/legacy/branch/new.rs
+++ b/crates/but/src/command/legacy/branch/new.rs
@@ -137,6 +137,10 @@ fn resolve_above_below_target(
     id_map: &IdMap,
     target: CliIdArg,
 ) -> CliResult<NewStackedBranchTarget> {
+    if let Some(commit) = id_map.anonymous_segment_top_commit(&target.0, repo)? {
+        return Ok(NewStackedBranchTarget::Commit(commit));
+    }
+
     let resolved = target
         .resolve_in_workspace(
             repo,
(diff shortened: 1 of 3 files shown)


krlvi:
hey @cmatheo - discussed briefly with @davidpdrsn yesterday - we are still pondering how we want these to work from a product stand point

slarse:
Hi @cmatheo,

As @krlvi pointed out above it's currently unclear what direction we want to go with anonymous segments. For now @davidpdrsn and I have concluded that we'll only support anonymous segments in `but reword`, and all other commands will reject them. This needs to be fixed at a more central parsing level and be supported by the type system, so I've concocted a solution in https://github.com/gitbutlerapp/gitbutler/pull/15672.

I'm therefore closing this PR. We appreciate your contribution, but it's not quite the right approach here.

PR #15611, opened 2026-08-28
Fix operation history crashing on every load

macgyver13:
28a8593f5d replaced the oplog transport type while porting the list to Lite. The old gitbutler_oplog::entry::Snapshot carried #[serde(rename = "id")] on its commit_id; the new but_api::legacy::oplog::json::Snapshot does not, so the wire field became commitId. The desktop's hand-written Snapshot interface still read id, leaving every entry.id undefined.

The snapshot list is keyed by that id, so two undefined keys threw each_key_duplicate and replaced the whole history view with 'Something went wrong' as soon as a second snapshot existed.

Renames the desktop type and its uses to commitId, matching the DTO and the generated SDK that Lite already consumes.

### DIFF
--- a/apps/desktop/src/components/history/SnapshotCard.svelte
+++ b/apps/desktop/src/components/history/SnapshotCard.svelte
@@ -195,7 +195,7 @@
 
 	const historyService = inject(HISTORY_SERVICE);
 	const snapshotDiff = $derived(
-		historyService.snapshotDiff({ projectId, snapshotId: entry.id, childId }),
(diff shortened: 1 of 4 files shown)


krlvi:
Thanks @macgyver13 for detecting and fixing this! And also thank you for running the nightly builds, it's really helpful in catching stuff like this

krlvi (approved):
verified the issue & the fix! thanks

krlvi:
just fyi @samhh - in case this interferes with the lite app

PR #15612, opened 2026-08-28
Configure fallback committer when opening repositories

Byron:
The good thing about this solution is that it works for all current and future reference creations while no Git identity is set.

----
Set gitoxide's committer name and email fallbacks on repositories opened through
Context. This allows reference reflogs before a user configures a Git identity
while preserving configured identities.

Restores direct reference creation, superseding
https://github.com/gitbutlerapp/gitbutler/pull/15591 and undoing
073c77063834177e3674639709cb1d13c32c252b.

### DIFF
--- a/crates/but-ctx/src/lib.rs
+++ b/crates/but-ctx/src/lib.rs
@@ -17,6 +17,7 @@ use but_core::{
 use but_path::AppChannel;
 use but_settings::AppSettings;
 use but_utils::OnDemandCache;
+use gix::config::tree::Key as _;
 use tracing::instrument;
 
 /// Legacy types that shouldn't be used.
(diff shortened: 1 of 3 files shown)


krlvi (approved):

krlvi:
nice this is cleaner - i had no idea it was possible

samhh:
[I only have `user.*` set](https://github.com/samhh/dotfiles/blob/81f435ee91a73dbe450acd5b539854c2cd46456e/home/vcs.nix#L229-L231), not `committer.*`, causing my commits today to be attributed to GitButler and therefore un-mergeable (currently e.g. https://github.com/gitbutlerapp/gitbutler/pull/15624). I believe this should be considered a bug?

Byron:
Sorry for not testing by hand, I didn't think it could do that at all, how naive. Rest assured that the fix will also come with at least one assertions that will catch this in future.

PR #15621, opened 2026-08-28
Refresh the workspace whenever HEAD moves

mtsgrd:
The head event declared only the operating mode stale, so a HEAD move refreshed
the mode and nothing else. The workspace still updated — but only because `git
checkout` writes the reflog as well, and that raises the separate `gitActivity`
event.

Moving between two branches is where it shows. The mode reads
`OutsideWorkspace` before and after, so nothing about the mode changes and
`gitActivity` was doing all the work. Suppressing each event in the client, on
the old code:

| suppressed | branch → branch refresh |
| --- | --- |
| `gitActivity` | **fails** |
| `gitHead` (the whole head event) | passes |

The head event contributed nothing. With this change the first row passes too.

- `GitHead` now declares what a repository change declares, plus the mode, so
  the refresh is stated rather than inherited from a sibling event that happens
  to arrive with it.
- The regenerated `cacheTags` diff is the one `gitHead` line. It was generated
  from a tree holding only this change, so other in-flight API work in the
  shared workspace is not swept into it.
- Three e2e tests: workspace → branch → detached → workspace, a bare `git
  symbolic-ref` rewrite of HEAD, and branch → branch with the mode unchanged.

No UI change. An earlier revision of this branch added a HEAD indicator to the
sidebar; that was dropped.

### DIFF
--- a/apps/lite/e2e/tests/head-moves.spec.ts
+++ b/apps/lite/e2e/tests/head-moves.spec.ts
@@ -1,0 +1,108 @@
+import { execFileSync } from "node:child_process";
+import path from "node:path";
+import { processEnvironment, type LiteTestEnvironment } from "../setup.ts";
+import { expect, test } from "../test.ts";
+
+/**
+ * Move HEAD from outside the app, the way a terminal `git checkout` does.
+ *
+ * The app is told about this by the watcher rather than by its own mutations,
+ * so nothing here touches the UI: the assertions are about what the workspace
+ * shows once the events have landed.
+ */
+const git = (environment: LiteTestEnvironment, ...args: Array<string>): void => {
+	execFileSync("git", args, {
+		cwd: path.join(environment.workdir, "local-clone"),
+		env: processEnvironment({ GIT_CONFIG_GLOBAL: environment.gitConfig }),
+		stdio: "pipe",
+	});
+};
+
+const checkout = (environment: LiteTestEnvironment, ...args: Array<string>): void =>
+	git(environment, "checkout", ...args);
+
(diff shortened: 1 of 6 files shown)


mtsgrd:
## 📸 UI screenshots

Withdrawn: the sidebar HEAD indicator these showed has been reverted, so this
branch no longer has a visual change. The screenshots that were here would have
shown UI that is not in the PR.

PavelLaptev:
@krlvi What do you think about this? If I remember correctly, you were against this in Tauri.

Screenshot 2026-08-28 at 17-56-04

mtsgrd:
It feels like something should indicate you're on gitbutler/workspace when in workspace mode, but wouldn't be necessary when you're in single branch mode. :/

PavelLaptev:
@krlvi @mtsgrd Another issue is that we're running out of space. Should we redesign the header? Or move it from the header to the stack section instead?

Screenshot 2026-08-28 at 18-00-25


Some header drafts from the past https://www.figma.com/design/EBuHQGUcCaSw4Ln5uVpWkn/Lite?node-id=2010-7070&t=O4He6X7ZqA7dfo6x-1

PavelLaptev:
I actually agreed with @mtsgrd 
I think we need an indicator. But we don't want it to look like a warning or something that doesn't look right.

krlvi:
I would be keen to solve the need for this in some other way. We dont think its desirable for workspace vs single branch to feel like "modes".

Very soon we will be looking at how worktrees fit in the picture. I would not wanna do this adjustment here until he have had some consideration around WTs first

PR #15624, opened 2026-08-28
Fix focus change re-renders

samhh:
Currently blocked from merging by this: https://github.com/gitbutlerapp/gitbutler/pull/15612#issuecomment-5455863023

If someone with elevated permissions would like to merge please go ahead. 🙂


krlvi:
@samhh sorry for the blockage. It is not fixed in the latest version of the app - so if you just rebase or edit something here, the committer will be fixed

PR #15644, opened 2026-08-29
Add a test-results skill for coding agents

mtsgrd:
Teaches agents working in this repo how to use tests.but.dev:

- **After pushing**: one call (`/api/report/branch?branch=X`) returns every suite's latest run with failures inlined — full failure output plus a `knownFlaky` verdict, so an agent can decide retry-vs-investigate without downloading CI logs.
- **Fixing flakes**: the evidence-ranked leaderboard (`retryFlakeRuns` = proof-grade) plus `/api/report/test/{id}` with failure clusters, disagreeing commit shas, and reproduction/verification guidance.

The report endpoints are already live.

### DIFF
--- a/.agents/skills/test-results/SKILL.md
+++ b/.agents/skills/test-results/SKILL.md
@@ -1,0 +1,73 @@
+---
+name: test-results
+description: Query CI test results, failures, and flake evidence from tests.but.dev. Use after pushing a branch to check whether CI failures are yours or known-flaky, when investigating a failing or flaky test, or to pick which flaky test to fix.
+---
+
+All six CI suites (cargo, vitest, playwright, playwright-ct, lite-e2e,
+blackbox) report every test result — including full failure output — to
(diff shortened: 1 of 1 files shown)


.agents/skills/test-results/SKILL.md
```diff
@@ -0,0 +1,73 @@
+---
+name: test-results
+description: Query CI test results, failures, and flake evidence from tests.but.dev. Use after pushing a branch to check whether CI failures are yours or known-flaky, when investigating a failing or flaky test, or to pick which flaky test to fix.
```
krlvi, line 3: we probably dont want this triggering 
`Use after pushing a branch` 

I view this `tests.but.dev` as a debug thing, and would prefer my agent to never interact with it unless im debugging a problem with CI

mtsgrd:
That makes sense, I'll get back to this pr later today.

PR #15645, opened 2026-08-29
PR activity notifications in lite

mtsgrd:
Draws the eye to pull requests with new activity, remembers what has been seen, and files what happened into a notification bell.

**The bell** sits in the window's top-right corner with a red dot while unread notifications wait. Its panel lists entries richly, newest first — a comment with author and snippet reads differently from "pushed 3 commits" or an approval. Clicking an entry marks it read and jumps to the PR tab (or the forge page when the branch isn't local); opening the panel alone marks nothing read.

One axis decides what files: **is a human waiting on you?** Someone else's comment, a verdict, a review request naming you, and `@you` mentions — loud on any open PR, applied or not. Commits are quieter facts the bell still carries; requests naming someone else, dismissed verdicts and your own actions are silence. A kind's items on one review coalesce into one entry per poll.

Seen state is per-review watermarks in local storage — deliberately disposable and per-machine:

- Reviews are stamped seen the first time they are listed, so a wiped store reads as "everything is read" rather than a wall of stale dots, and only activity after a review first appeared can be unread.
- Marks never move backwards; viewing a PR tab advances one after a short dwell, re-armed on window focus so an unfocused view doesn't count. Quiet surfaces stay: dots on the branch row's PR chip and the Details PR tab, a count on the Workspace sidebar tab.
- Each new conversation entry wears a small `New` marker, and the markers are **per-item**: an entry counts as looked at only once it has sat in the focused viewport for a beat. Items the dwell advances past unseen are recorded as *skips* — the complement is stored on purpose, since a list of read items would grow with everything ever read once one stubborn item pinned the mark — so their markers and the dot survive until each is actually seen. Skips are capped; dropping one merely reads as seen.
- It started as a `review_seen` table in but-db; [krlvi's review](https://github.com/gitbutlerapp/gitbutler/pull/15645#discussion_r3886170014) pointed out that disposable per-machine state doesn't need a migration or a round trip, so the table, its endpoints and its cache tag are gone.

Detection rides the existing 60s review-listing poll: a `modifiedAt` bump past the watermark triggers one targeted comments/submissions/timeline fetch for that PR only, and the detector baselines from the stored watermarks, so activity that landed while the app was closed still announces itself once. One dial in Settings → General: `Pull request activity: loud | quiet | off` (loud shows the bell, quiet keeps only the dots, off hides it all).

Testable without a forge: the classifier and the stores are pure modules driven by fabricated events in vitest, and the harness panel mounts the real detector so `panel.test.tsx` runs the whole loop — poll, classify, file, dot — against fake forge handlers. Also verified live against real GitHub data.

Landing the click on the exact comment rides the stacked diff-comments PR, where inline comments become addressable. OS notifications and a dock badge for the unfocused window are deliberately left for a later follow-up.

### DIFF
--- a/apps/lite/ui/src/review-inbox.ts
+++ b/apps/lite/ui/src/review-inbox.ts
@@ -1,0 +1,192 @@
+/**
+ * @file The notification inbox behind the bell.
+ *
+ * One entry per review, kind and poll — the detector writes what happened,
+ * structured, and the bell renders it richly. Same store discipline as the
+ * watermarks: local storage, disposable, capped, validated at parse, pure
+ * in-memory snapshots.
+ */
+
+import { useSyncExternalStore } from "react";
+
+export type InboxKind =
+	| "comment"
+	| "mention"
+	| "approved"
+	| "changesRequested"
+	| "reviewRequested"
+	| "committed"
+	| "merged"
+	| "closed";
+
+export type InboxEntry = {
+	id: string;
+	kind: InboxKind;
+	review: number;
+	reviewTitle: string;
+	unitSymbol: string;
+	sourceBranch: string;
+	htmlUrl: string;
+	author: string | null;
+	/** How many items this entry coalesces — "3 comments". */
+	count: number;
+	snippet: string | null;
+	at: string;
+	seen: boolean;
+};
+
+const inboxKinds: ReadonlyArray<string> = [
+	"comment",
+	"mention",
+	"approved",
+	"changesRequested",
+	"reviewRequested",
+	"committed",
+	"merged",
+	"closed",
+];
+
+const storageKey = (projectId: string) => `pr_activity_inbox:v1:${projectId}`;
+
+/** Newest kept; the overflow was old news nobody opened. */
+const inboxCap = 100;
+
+const listeners = new Set<() => void>();
+let cached: { key: string; entries: Array<InboxEntry> } | null = null;
+
+// Storage can throw — disabled, partitioned, or over quota — and a throw
+// here would crash every subscriber's render. The inbox degrades instead.
+const storageGet = (key: string): string | null => {
+	try {
+		return localStorage.getItem(key);
+	} catch {
+		return null;
+	}
+};
+const storageSet = (key: string, value: string): void => {
+	try {
+		localStorage.setItem(key, value);
+	} catch {
+		// The in-memory copy still serves this session.
+	}
+};
+
+const parseEntries = (raw: string | null): Array<InboxEntry> => {
+	if (raw === null) return [];
+	try {
+		const stored: unknown = JSON.parse(raw);
+		if (!Array.isArray(stored)) return [];
+		// Ordered here, not just at write time: a list stored by an older
+		// build keeps whatever order it had until something new files.
+		return stored
+			.filter((entry): entry is InboxEntry => {
+				if (typeof entry !== "object" || entry === null) return false;
+				const e = entry as Record<string, unknown>;
+				return (
+					typeof e.id === "string" &&
+					typeof e.kind === "string" &&
+					inboxKinds.includes(e.kind) &&
+					typeof e.review === "number" &&
+					typeof e.reviewTitle === "string" &&
+					typeof e.unitSymbol === "string" &&
+					typeof e.sourceBranch === "string" &&
+					typeof e.htmlUrl === "string" &&
+					(e.author === null || typeof e.author === "string") &&
+					typeof e.count === "number" &&
+					(e.snippet === null || typeof e.snippet === "string") &&
+					typeof e.at === "string" &&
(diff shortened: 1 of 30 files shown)


mtsgrd:
## 📸 UI screenshots

Captured 10 catalogue surfaces against seeded fixtures, base vs head. One shows a real change; the rest of this feature only appears with a forge, so those were captured from the running app against live GitHub data — see below. (Updated after the toast → notification-bell redesign.)

### settings

The **Pull request activity** dial (loud / quiet / off) in Settings → General.

| Before | After (live capture) |
| --- | --- |
|  |  |

### The notification bell — live captures

Activity now files into a bell in the sidebar header instead of toasts. A red dot says unread notifications wait:



The panel over live GitHub data — title first, per-kind icons, coalesced counts ("left 9 comments on #15651"), blue unread dots, compact times. Clicking an entry marks it read, closes the panel, and jumps to that PR's tab:



Every entry kind, rendered from entries seeded into the store (same components as production; fabricated data, same convention as the earlier fake-forge toast shots) — mention, changes requested, approved, coalesced comments, commits, merged:



### Unread dots and Workspace badge — live capture

The running app against live GitHub data, with two watermarks wound back: the accent dot on the `pr-activity` row's PR chip, and the unread count on the Workspace tab (top left).



<details>
<summary>Other surfaces</summary>

Byte-identical: branches-tab, project-picker, uncommitted-rows, upstream-tab, workspace-sidebar (its dot needs a forge, hence the live capture above).

Differed only by capture noise, not by this branch: commit-form (per-run panel width), details-pane (randomized placeholder text and panel width), edit-mode (fixture-generated commit hash).

The bell and dot states are ad-hoc live captures — not covered by the screenshot catalogue going forward.

</details>

crates/but-db/tests/db/migration.rs
```diff
@@ -527,6 +527,12 @@ CREATE TABLE `hunk_assignments`(
 	PRIMARY KEY(`path`, `hunk_header`)
 );
 
+-- table review_seen
```
krlvi: @mtsgrd i wonder if this class of state "seen x" is better off living in local storage - more flexibility, no worries around migrations etc

PavelLaptev:
Os it okay that we see notifications only for applied branches? @krlvi

apps/lite/ui/src/review-inbox.ts
```diff
@@ -0,0 +1,192 @@
...
+const storageKey = (projectId: string) => `pr_activity_inbox:v1:${projectId}`;
+
+/** Newest kept; the overflow was old news nobody opened. */
+const inboxCap = 100;
+
+const listeners = new Set<() => void>();
+let cached: { key: string; entries: Array<InboxEntry> } | null = null;
+
+// Storage can throw — disabled, partitioned, or over quota — and a throw
+// here would crash every subscriber's render. The inbox degrades instead.
```
samhh, line 58: Was Local Storage over IndexedDB (as elsewhere) an intentional choice?

apps/lite/ui/src/review-inbox.ts
```diff
@@ -0,0 +1,192 @@
...
+
+const parseEntries = (raw: string | null): Array<InboxEntry> => {
+	if (raw === null) return [];
+	try {
+		const stored: unknown = JSON.parse(raw);
+		if (!Array.isArray(stored)) return [];
+		// Ordered here, not just at write time: a list stored by an older
+		// build keeps whatever order it had until something new files.
+		return stored
+			.filter((entry): entry is InboxEntry => {
```
samhh, line 82: User-defined type guards are unsafe and unergonomic. Validation of this many properties would benefit from a validation library; we already have the arktype dependency for this purpose.

Having said that, it's also not really necessary to validate this data. We own it, we've versioned the key, it's about as trustworthy in terms of its type as the data we get from the SDK.

apps/lite/ui/src/review-inbox.ts
```diff
@@ -0,0 +1,192 @@
...
+/** The inbox, newest first — a stable array identity between writes. */
+export const useInboxEntries = (projectId: string, enabled: boolean): Array<InboxEntry> =>
+	useSyncExternalStore(enabled ? subscribeInbox : subscribeNothing, () =>
+		enabled ? readEntries(projectId) : emptyInbox,
+	);
+
+const emptyInbox: Array<InboxEntry> = [];
+
+/** How many entries are unseen — the bell's dot. */
+export const useInboxUnseenCount = (projectId: string, enabled: boolean): number =>
```
samhh, line 189: If you squint we're reimplementing the RQ API. Taken with the other comments you can compare against for example how draft PRs are implemented.

I should probably add a bunch of these notes to our `AGENTS.md`.

samhh (commented):
(Nothing major, just noted some inconsistency between feature implementations.)

PR #15647, opened 2026-08-29
fix(tui): support squashing marked commits into self

davidpdrsn:


krlvi:
image
this is due to a bug from yesterday - but the latest nightly should no longer do that

davidpdrsn:
> this is due to a bug from yesterday - but the latest nightly should no longer do that

Good to know! I haven't built in a few days.

PR #15651, opened 2026-08-29
Show a review's diff comments in lite

mtsgrd:
Comments left on a diff line were fetched by nothing, so a reviewer who only commented inline showed up as an empty "Reviewed" card, and their `@you` never raised a mention. This PR reads them, renders them everywhere they belong, answers them, and moves the compact timeline into the side panel.

**Diff comments**
- `listReviewThreads` reads them through GraphQL's `reviewThreads` — REST reports neither the thread a comment belongs to nor whether it was resolved (half of Copilot's threads on the previous PR were resolved, and REST would have shown every one as open).
- The Pull Request tab nests threads under the review they were posted with, quotes the code each hangs on, and says why a thread is not on the diff. The branch diff renders open threads on the line they were left on and pins them on the minimap.
- A thread whose line no longer says what the forge quoted is marked `Outdated` — the forge's own flag only knows the pushed head, so a branch amended since keeps threads looking current while their line numbers point at moved code.
- Replies post from either surface through `addPullRequestReviewThreadReply`. Mentions inside diff comments now count, so an `@you` on a line files like one left in the conversation.

**Activity in the side panel**
- The compact happenings — opened, commits, review requests — move out of the conversation into an Activity section, newest first, five shown with a "Show N more" reveal, scrolling inside itself.
- The timeline now comes from GraphQL's `timelineItems` rather than REST: only there does a commit resolve to a GitHub account, and without a login the reader's own pushes could not be told apart from anyone else's — they wore "New" markers and filed into the notification bell.
- Agent feedback stands out: any bot author gets an Agent chip and a tinted card, detected generically (the forge's `isBot` plus the `[bot]` login suffix). Fixes the REST mapping that compared GitHub's `Bot` user type against lowercase `bot`, which left `isBot` false for every bot.

**Timestamps**
- `RelativeTime` ticks on a single shared clock per interval, so a settled conversation stops saying "2 minutes ago" an hour later, and a page of timestamps costs one timer.

Right-clicking quoted code copies it or opens the file at that line; a selection anywhere in the app now offers copy.


PavelLaptev:
@mtsgrd please don't forget to add screenshots

krlvi (approved):

mtsgrd:
## 📸 UI screenshots

Live captures from the running app against this PR's own review data (the feature only appears with a forge, so these are photographs of the real thing rather than fixture captures). All ad-hoc — not covered by the screenshot catalogue going forward.

### A diff thread, nested under its review

An open Copilot thread with its Reply, and the code it hangs on quoted beneath — dimmed directory, bright file and line, real line numbers:



### Review cards

Agent authors get the chip and tinted card; resolved conversations fold away; suppressed comments stay countable:



### Activity in the side panel

The compact timeline — commits and review requests, newest first, five shown with a reveal. Commits resolve their author through GraphQL now, so your own pushes carry no "New" marker:



**Not shown**: inline thread cards on the branch diff and their minimap pins — every open thread on this PR is currently `Outdated` (the branch has been rebased past all of them), and the diff drops outdated threads by design, so there is nothing to photograph. The rendering is exercised by `review-threads.test.ts` and was verified live before the rebases.

PR #15668, opened 2026-08-31
Newest first and a folded composer in the PR conversation

mtsgrd:
The conversation now reads newest first, with the composer at the top.

- The conversation and the Activity feed sort newest first at the source (the panel previously reversed by hand); items sharing a timestamp — commits pushed in one batch — keep that reading through a reverse before the stable sort
- The composer sits right under the description and folds to one input-shaped row until focused; focusing reveals the full markdown composer with focus handed straight to the textarea
- A draft arriving from outside unfolds it too: a reply quote, or a failed submit restoring its text
- Submitting folds it back (optimistic — a forge refusal re-unfolds it with the draft intact), as does Escape or leaving it while empty
- Your own just-posted comment appears directly under the composer, where the new order puts "now"

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/PullRequestComments.module.css
+++ b/apps/lite/ui/src/routes/project/$id/workspace/PullRequestComments.module.css
@@ -249,7 +249,7 @@
 .composer {
 	display: flex;
 	flex-direction: column;
-	margin-block-start: 14px;
+	margin-block-end: 14px;
 	border-radius: var(--radius-card);
 	outline: 1px solid var(--border-2);
 	outline-offset: -1px;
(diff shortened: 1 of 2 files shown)


mtsgrd:
### 📸 UI screenshots

Live captures from the dev app with the branch applied (after) and unapplied (before), photographed on #15621's conversation. The PR conversation needs a real forge, so it is not in the fixture-backed catalogue — these are one-off captures, not standing coverage.

#### Conversation — composer placement and order

Before: no composer under the description; comments read oldest first. After: the composer folds to one input row right under the description, and comments read newest first (krlvi's 16:33 comment now leads; the oldest sinks to the bottom).

| Before | After |
| --- | --- |
| conversation before | conversation after |

#### Composer — folded until focused

Before: the full composer always open at the bottom of the feed. After: focusing the folded row reveals the same rich composer in place, with the caret in the textarea; it folds back on submit, Escape, or leaving it empty.

| Before (bottom of feed) | After (focused, top of feed) |
| --- | --- |
| composer before | composer focused |

PR #15679, opened 2026-09-01
Show and resolve branch divergence in lite

mtsgrd:
A diverged branch shows its remote as a second rail in the workspace outline: a collapsed `origin/<name>` row with the ahead count, expandable to the remote-only commits. Head info's `commitsOnRemote` was rendered nowhere, and force push was the only resolution on offer.

- A branch whose history was only rewritten, nothing new either side, reads "Rewritten" with no button: the branch row's force push is the whole answer.
- Integrate on the leg and Update From Remote in the branch menu open a dialog led by a diagnosis of what the remote has, with three answers over one outline of the previewed outcome (a dry-run `applyBranchIntegration`): **Keep mine** force-pushes the branch as it is; **Combine both** rebases your work onto the remote's new commits, leaving out the remote's versions of commits you rewrote so neither side lands twice; **Take theirs** replaces the branch with the remote. Conflicts are flagged before anything applies.
- Under Combine both every incoming commit carries a checkbox; unticking one leaves it out and parks it under "Left out".
- Integrating is not publishing, and the dialog says so: the footer states whether a plain or force push follows, and once the integration lands the dialog offers that push in place, with Done to leave it for later.

Verified by `branch-update.spec.ts`, which applies every answer against seeded divergences and asserts the resulting git history, conflict markers included.

### DIFF
--- a/apps/lite/e2e/diverge.ts
+++ b/apps/lite/e2e/diverge.ts
@@ -1,0 +1,79 @@
+import { execFileSync } from "node:child_process";
+import { appendFileSync, writeFileSync } from "node:fs";
+import path from "node:path";
+import { paths, processEnvironment, type LiteTestEnvironment } from "./setup.ts";
+
+/**
+ * branch1 applied and diverged: one commit only here, two only on the remote,
+ * both sides editing `a_file` so a rebase conflicts. Reload the app afterwards.
+ */
+export const divergeBranch1 = (environment: LiteTestEnvironment): void => {
+	const but = process.env.BUT ?? path.join(paths.repoRoot, "target/debug/but");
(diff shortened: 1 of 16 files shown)


mtsgrd:
## 📸 UI screenshots

Captured 5 surfaces, 5 changed.

Captured ad hoc from a scratch checkout of master plus this commit, so the sidebar is master's; these surfaces are not in the screenshot catalogue and are not covered going forward.

### workspace-sidebar-diverged

| Before | After |
| --- | --- |
|  |  |

### branch-update-dialog

### branch-update-done

### workspace-sidebar-incoming

### workspace-sidebar-rewritten

PR #15694, opened 2026-09-01
fix: small file row polish in Lite

PavelLaptev:
Two small UI fixes in the Lite workspace rows:

- The file row status icon now uses `--fill-pop-bg` instead of `--text-pop`, matching the fill styling used by the related status badge colors.
- The row label container's column gap tightens from 10px to 8px.


PavelLaptev:
No need for screenshots here. Changes are too minor.

PR #15699, opened 2026-09-02
Fix _expand IDs in linked worktrees

krlvi:
## Context

Trigger: run `but _expand` with a commit ID displayed by status from inside that linked worktree.

Symptom: the command aborts at the main-worktree state guard, so custom `but-*` commands cannot resolve the displayed commit.

## Change

Build the expansion map from the canonical main-worktree context when worktree support is enabled, while keeping object parsing tied to the invoking repository. Preserve feature-off linked-worktree behavior, reject bare repositories cleanly when canonical worktree state is unavailable, and cover both repository shapes with focused tests.

Fixes GB-1960

### DIFF
--- a/crates/but/src/command/expand.rs
+++ b/crates/but/src/command/expand.rs
@@ -125,9 +125,26 @@ impl CliOutput for ExpandOutcome {
 }
 
 pub fn handle(ctx: &but_ctx::Context, cli_id: CliIdArg) -> CliResult<ExpandOutcome> {
-    let guard = ctx.shared_worktree_access();
-    let id_map = IdMap::new_from_context(ctx, guard.read_permission())?;
     let repo = ctx.repo.get()?;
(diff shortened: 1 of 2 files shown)


krlvi:
@davidpdrsn do you like this solution (from your linear). Feel free to close if not

slarse:
@krlvi I'ma pull an executive decision and close this. This is basically a one-off hack doing what @Caleb-T-Owens handles generally with https://github.com/gitbutlerapp/gitbutler/pull/15705.

The ticket itself is also a bit unclear to me. If it's saying that commits in worktrees cannot be expanded from the main worktree, that's incorrect/outdated, it works fine. If it's saying that `but` is currently non-functional when CWD is the linked worktree, that is correct but not an issue with `_expand` per se. @davidpdrsn could elaborate :)

davidpdrsn:
> If it's saying that but is currently non-functional when CWD is the linked worktree

It says this.

PR #15705, opened 2026-09-02
But acts as if run in main worktree

Caleb-T-Owens:


Caleb-T-Owens:
I've updated the PR so that `but setup` can't be run from linked worktrees. It seems a bit more nonsensical

krlvi (approved):

PR #15711, opened 2026-09-02
feat(but): support committed hunks in `but discard`

slarse:


crates/but/tests/but/command/discard.rs
```diff
@@ -783,6 +783,369 @@ Hint: run `but help` for all commands
...
+        .success()
+        .stdout_eq(snapbox::str![[r#"
+──────────────────────────╮
+ 1#0:q:q renamed_file.txt │
+──────────────────────────╯
+
+No diff available - file is either empty, binary, or too large
+
+"#]]);
+}
```
slarse, line 1069: The unihunk diff of a renamed file looks, let's say, _not great_. Will fix separately.

crates/but/tests/but/command/discard.rs
```diff
@@ -783,6 +783,369 @@ Hint: run `but help` for all commands
...
+
+No diff available - file is either empty, binary, or too large
+
+"#]]);
+
+    env.but("discard 1:p:q")
+        .assert()
+        .success()
+        .stdout_eq(snapbox::str![[r#"
+Discarded changes from B from 1 to create 1
```
slarse, line 1122: This status message doesn't make much sense anymore now that we use change IDs. Will fix separately.

PR #15716, opened 2026-09-02
Remember branch diff/pr selection

Caleb-T-Owens:


apps/lite/ui/src/branch-tab.test.ts
```diff
@@ -0,0 +1,24 @@
...
+	it("loads the stored pick, so repeating it writes nothing", async () => {
+		localStorage.setItem(key, "pr");
+		vi.resetModules();
+		const { writeBranchTab } = await import("./branch-tab.ts");
+		const setItem = vi.spyOn(Storage.prototype, "setItem");
+		writeBranchTab("pr");
+		expect(setItem).not.toHaveBeenCalled();
+		setItem.mockRestore();
+	});
+});
```
samhh, line 24: Nit: These suites that agents generate are pretty low value, I usually delete them. User-visible behaviour is best tested in E2E.

apps/lite/ui/src/branch-tab.ts
```diff
@@ -0,0 +1,46 @@
...
+		localStorage.setItem(storageKey, tab);
+	} catch {
+		// The in-memory copy still serves this session.
+	}
+	for (const listener of listeners) listener();
+};
+
+/** The last tab picked, or `null` before any pick, when each surface leads with its own default. */
+export const useChosenBranchTab = (): BranchTab | null =>
+	useSyncExternalStore(subscribe, () => chosen);
```
samhh, line 46: Ideally for me we'd use React Query for this sort of thing. It's consistent with everything else, ready to scale as new requirements arise, has tooling niceties, etc. In that case you can look at some examples dotted about using IDB instead of Local Storage.

Having said that it's not a blocker.

samhh (approved):

PR #15731, opened 2026-09-03
fix(lite): a folded branch stays folded when renamed

cbjeukendrup:
When renaming a folded branch, it would become unfolded; when renaming it back to the original name, it would become folded again. Fix this by migrating the folded state to the new branch name upon successfully renaming a branch.

Resolves [my comment about this on Discord from this morning](https://discord.com/channels/1060193121130000425/1481308709001891961/1545032628095426640)


samhh:
Thank you!

PR #15734, opened 2026-09-03
test(but): derive initial test change IDs from commit content

slarse:
Change IDs have historically been hard coded to `1` in tests. Now that we're using change IDs as the primary identifier for commits, that's massively inconvenient for tests.

This PR adds a magic `content-hash` value for `gitbutler.testing.changeId` that causes the change ID to be deterministically generated from the content of the commit (before writing the change ID to it, of course, for practical reasons of possibility).

The change ID is initially set to a placeholder, as in the current code flow the change ID header is "requested" before the commit is fully formed. For pretty obvious reasons. The header is then replaced with a content derived Change ID upon writing the commit to disk.

The updates to tests are _massive_. I've put LLMs hard at work in several cycles to detect test regressions. I've caught some. Some may be left, but this is such a valuable improvement that it's worth the risk of some borked tests.

> **Important:** You need to delete the cached test fixtures (in `generated-do-not-edit` directories) for tests to pass after pulling this in.

### DIFF
--- a/crates/but-core/src/commit/mod.rs
+++ b/crates/but-core/src/commit/mod.rs
@@ -86,23 +86,30 @@ impl Headers {
     }
 
     /// Create a new instance, with the following rules for setting the change id header:
-    /// 1. Read `gitbutler.testing.changeId` from `config` and if it's a valid u128 integer, use it as change-id.
-    /// 2. generate a new change-id
+    /// 1. If `gitbutler.testing.changeId` is `content-hash`, use a placeholder that is replaced
+    ///    with an ID derived from the final commit content by [`create()`].
+    /// 2. If `gitbutler.testing.changeId` is a valid u128 integer, use it as change-id.
+    /// 3. Generate a new change-id.
     ///
     /// This produces a stored header value. For the deterministic fallback used when headerless
     /// commits still need a change-id, see [`Self::ensure_change_id()`].
     pub fn from_config(config: &gix::config::Snapshot) -> Self {
+        let use_content_hash = config
+            .string(TESTING_CHANGE_ID_KEY)
+            .is_some_and(|value| value.as_slice() == TESTING_CHANGE_ID_CONTENT_HASH);
         Headers {
-            change_id: Some(
+            change_id: Some(if use_content_hash {
+                ChangeId::from(BString::from(CONTENT_HASH_CHANGE_ID_PLACEHOLDER))
+            } else {
                 config
-                    .integer("gitbutler.testing.changeId")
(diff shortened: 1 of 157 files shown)


crates/but/tests/fixtures/scenario/shared.sh
slarse, line 1: Key change: this causes fixtures to use content-derived change IDs.

crates/but-core/src/commit/mod.rs
slarse, line 1: The functional changes are all in here.

crates/but-rebase/src/merge.rs
```diff
@@ -103,7 +103,7 @@ pub(crate) fn octopus(
         )
     } else {
         crate::commit::update_committer(repo, &mut target_merge_commit)?;
-        Ok(repo.write_object(target_merge_commit)?.detach())
```
slarse: This has no test impact, but if we don't change this then anything that takes this codepath in tests won't correctly replace the change ID placeholder. So I think this is the way to go?

crates/but/src/command/legacy/status/tui/tests/copy_tests.rs
```diff
@@ -95,7 +103,8 @@ fn copies_every_commit_value() {
 
     tui.input(COPY_MORE);
     tui.input([KeyCode::Down; 2]);
-    tui.input(KeyCode::Enter).assert_copied_text_eq("1");
+    tui.input(KeyCode::Enter)
+        .assert_copied_text_eq("omyuzposspozzzlpxqkzstlkmxpvxzqs");
```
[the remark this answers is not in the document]
slarse: Eh, no, this is fine.

crates/but-core/tests/core/commit.rs
```diff
@@ -324,10 +324,69 @@ Headers {
 mod create {
     use anyhow::Context as _;
     use bstr::ByteSlice;
```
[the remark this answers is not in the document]
slarse: This PR doesn't even touch that... So no.

crates/but-rebase/src/merge.rs
```diff
@@ -103,7 +103,7 @@ pub(crate) fn octopus(
...
+        but_core::commit::create(repo, target_merge_commit, None, SignCommit::No)
```
slarse, line 106: And one change here - this was sidestepping the materialization-time substitution of the change ID placeholder.

krlvi (approved):

davidpdrsn (approved):
:shipit:

PR #15745, opened 2026-09-04
Gracefully handle large changesets

samhh:
Closes GB-1947. Defers awkward UX interactions around unloaded patches for later as it's an edge case unlikely to be encountered by real users. Poor backend performance in `changesInWorktree` is tracked separately. Further performance improvements are viable but not blocking.


apps/lite/ui/src/api/queries.ts
```diff
@@ -528,6 +534,63 @@ export const treeChangeDiffsQueryOptions = ({ projectId, change }: PayloadFor<"t
...
+				for (
+					let batchStart = 0, batchSize = 64;
+					batchStart < changes.length;
+					batchStart += batchSize, batchSize *= 2
+				) {
+					yield await pMap(
+						changes.slice(batchStart, batchStart + batchSize),
+						(change) => window.lite.treeChangeDiffs({ projectId, change }),
+						{ concurrency, signal },
+					);
```
mtsgrd, line 587: @Caleb-T-Owens I think I remember we talked about concurrency wrt to read locks.. do you know if these will execute in parallel?

samhh: Here are some numbers from an agent in my test repo. Concurrency 9 is the comparison as my system has 18 logical cores.

---

For **queryFn alone**, excluding the file scan and React rendering, medians across three runs on 5,021 files:


| Configuration | First 64-file batch | All complete |
| -- | -- | -- |
| Master (concurrency 9) | **110ms** | **4.04s** |
| Concurrency 1 | **358ms** | **28.50s** |

Concurrency 1 takes **7.1× longer** overall.

The separate live-render measurement (one run each):


| Configuration | First batch painted | All rendered |
| -- | -- | -- |
| Master | 2.37s | 6.87s |
| Concurrency 1 | 2.54s | 32.77s |

---

I then pointed the agent at your comment here since I'm not familiar with the Rust code.

---

The code supports concurrency:

* The N-API wrapper runs each synchronous Rust call through `tokio::task::spawn_blocking`.
* `tree_change_diffs` obtains a repository and computes the patch without explicitly taking a worktree lock.
* Even where shared read locks are used, they **allow multiple readers simultaneously**; writers block them. That’s documented in `shared_worktree_access`.

The **28.50s → 4.04s** result demonstrates that overlapping these requests substantially improves throughput. It doesn’t establish precisely how much comes from parallel CPU work versus overlapping I/O, context setup, and IPC—but it argues strongly against the whole operation being serialized by a lock.

So my answer to that comment would be: **yes, this path supports parallel execution, and the Deno benchmark shows a substantial practical benefit.**

PR #15748, opened 2026-09-05
fix(lite): let the stacks and branches lists scroll past the selected row

cbjeukendrup:
The workspace stacks list would reach a point it could not scroll past: it moved for one frame, snapped back, and stayed there for as long as you kept scrolling. The branches list does the same once a few branches are unfolded.

## Cause

Both lists pin the selected row into the virtual range so it stays mounted wherever the list is scrolled to, and `getRangeExtractorWithIndices` appended that pin *after* the range instead of putting it in its place. So once the range slid past the pinned row, the row moved between the end of React's keyed children and its place among them on every render — and React answers a re-order by tearing the moved rows' layout effects down and setting them up again.

That matters because those rows virtualise their own commits against the list's single scroller. `@tanstack/react-virtual` assumes a virtualizer *owns* its scroller: whenever its effects re-attach, `_willUpdate` writes the offset it last remembered back onto the element.

```js
if (this.scrollElement !== scrollElement) {
  this.cleanup();
  ...
  this._scrollToOffset(this.getScrollOffset(), { adjustments: undefined, behavior: undefined });
}
```

Re-attaching on every render, each mounted row stamped an offset the user had already scrolled away from. And because a frame's scroll writes coalesce into a single scroll event, the stamped offset was the one every virtualizer read back as the truth — so none of them ever learned where the user actually scrolled, and they kept re-stamping the same number.

Measured in the running app at the stuck point: nine `scrollTo({top: 264})` calls per frame from `Virtualizer._willUpdate → _scrollToOffset`, while the element had just been set to 276.

## The fix ~, and the follow-up~

**`fix(lite): keep a pinned virtual row in its place in the list`** is the fix. Ordering the extractor's indices removes the per-render re-order, and with it the effect churn that drove the whole thing. Instrumenting every virtualizer bound to the stacks scroller afterwards, a scroll produces **zero** re-attachments and zero writes, where before it produced one per mounted row per frame.

DOM order is semantic in these lists anyway — they are `role="tree"`, so it is what a screen reader reads and what sequential navigation follows, and a pinned row appended after the range was announced last among its siblings.

~The `refactor` commit on top is not needed to unstick the lists, and is worth having anyway. A borrowed scroller with several virtualizers on it is only safe while nothing re-attaches their effects mid-scroll, and when that happens is not something these lists decide — `Activity` tears them down and sets them up again on every tab switch. So the nested virtualizers are given a `scrollToFn` that does nothing: they read the scroller, and only the virtualizer that owns it moves it.~

~That also lets the one write they made on purpose go. A selected commit row now reveals itself through `Row`, the way a branch row already did and every unvirtualised row does, which the extractor's pin makes reliable by keeping it mounted — so the hand-rolled `scrollToIndex` effects and the refs that remembered whether `Activity` had already replayed them are gone. Net, the two commits remove more code than they add.~

## Tests

`getRangeExtractorWithIndices` gets unit coverage for where a pin lands, and two e2e tests drive a list taller than its panel with the first row selected. Both fail on `master` and pass on the first commit alone — they cover the behaviour, not either mechanism, so they stay honest if the ordering fix is ever revisited.

Not affected, checked: the upstream list, `FilesTree`, `PickerDialog` and `ProjectPicker` each run a single virtualizer that owns its own scroller.

Fixes the regression from #15692 (stacks) and the same shape in #15682 (branches).

### DIFF
--- a/apps/lite/e2e/tests/branches-list.spec.ts
+++ b/apps/lite/e2e/tests/branches-list.spec.ts
@@ -1,0 +1,97 @@
+import { expect, test } from "../test.ts";
+
+test.describe("branches list", () => {
+	test.use({ scenario: "project-with-many-unapplied-branches.sh" });
+
+	// Each unfolded branch runs a commit virtualizer on the borrowed list
+	// scroller, so with the selected branch scrolled away from, they used to
+	// stamp the offset they last remembered back over the scroll the user made.
+	test("scrolls to the last branch while the first one is selected", async ({ appWindow }) => {
+		// Seeding a project this size and then unfolding and driving it by hand
+		// costs more than the default budget allows for on a loaded CI runner.
+		test.slow();
+
+		await appWindow
+			.getByRole("group", { name: "Pages" })
+			.getByRole("button", { name: "Branches" })
+			.click();
+
+		const branches = appWindow.getByRole("treeitem", { name: /^branch-\d+$/ });
+		await expect(branches.first()).toBeVisible();
+
+		// `unfold` asks every branch the list currently has mounted to show its
+		// commits, which is what puts a commit virtualizer on the shared scroller.
+		// It rides along with the scroll reading because both need the same
+		// element and each round trip costs the same as the whole batch.
+		const scroll = (unfold = false): Promise<{ offset: number; end: number }> =>
+			appWindow.evaluate((unfoldMounted) => {
+				// Anchored on a branch row: the hidden sidebar tabs are still in the
+				// document, and they look just like this list from the outside.
+				const row = document.querySelector('[role="treeitem"][aria-label^="branch-"]');
+				const tree = row?.closest('[role="tree"]');
+				const scroller = tree?.parentElement;
+				if (!tree || !scroller) throw new Error("Branches list has no scroller");
+
+				if (unfoldMounted) {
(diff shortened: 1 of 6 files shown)


cbjeukendrup:
Screen recording showing the behaviour before these changes:

https://github.com/user-attachments/assets/f09837e8-dad7-41d7-9bd5-a4d8048051e0

Also reproduced it on the branches tab, when some branches are unfolded. After these changes I can't reproduce it anymore.

samhh (changes requested):
Thanks! For posterity I could only repro the issue in a dev build due to `<StrictMode />`'s double-firing effect behaviour.

I believe there are some regressions however due to the second commit specifically:

1. Scrolling to the selected commit can be off by some distance e.g. expand a branch, select a commit, scroll really far down, down arrow.
2. Navigating down the workspace commits (down arrow) can cause some selected commits to be visibly behind the merge base dock.

cbjeukendrup:
Thanks for taking a look! 

I decided to take the second commit out of this PR, since it is not necessary to fix the issue, and I'm afraid I don't have time at the moment to dive deep enough into it to confidently fix it (also after the recent big changes to the workspace tab). 

Fwiw, it seems that there are currently also some keyboard navigation scrolling issues without this PR, so it's probably best if I don't get in the way there :)

samhh (approved):

PR #15749, opened 2026-09-06
Correct folded branch commit rings

krlvi:
FYI @PavelLaptev

## Context

In Lite's branch lists, a folded branch containing one or two commits always shows three circles despite the commit count. The rail overstates how many commits are hidden, so users have to rely on the numeric count or unfold the branch.

## Change

Show one ring for one hidden commit, two for two, and the compact three-ring group for larger counts. Both workspace and unapplied branch lists supply their existing commit count. Preserve numeric counts, unfolding, colors, graph connectors, and the three-ring fallback when the count is unknown.

### DIFF
--- a/apps/lite/ui/src/components/GraphSegment.test.tsx
+++ b/apps/lite/ui/src/components/GraphSegment.test.tsx
@@ -1,0 +1,44 @@
+/** @vitest-environment jsdom */
+
+import { GraphSegment } from "./GraphSegment.tsx";
+import { renderToStaticMarkup } from "react-dom/server";
+import { describe, expect, it } from "vitest";
+
+const render = (groupCount?: number) => {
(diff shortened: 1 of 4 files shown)


PavelLaptev:
@krlvi As we previously discussed, the folded ring doesn't count the number of commits, it just indicates whether the branch is folded or not. If there is only one commit, it will always show a single dot, making it hard to understand if it's folded. There needs to be another solution, which I promised to address in this ticket: https://linear.app/gitbutler/issue/GB-1979/rework-branch-rows-to-fit-pr-and-push-buttons

The branch row will be redesigned anyway.

krlvi:
@PavelLaptev this was the bot that saw the conversation on discord. We can close this PR

PR #15736, opened 2026-09-03
Draw the stacks section as a graph

mtsgrd:
The workspace's stacks section becomes a git graph: the stack cards stay full width, and rails run around them into an upstream section below, with the target's incoming commits and the merge base's history.

**Layout**

- One main line runs up the left into the top card. Every other card, and a target that has moved on, sits one gap to the right and bends onto the line in the gap under it. A row is indented only where the line runs behind it.
- The plan is pure and unit-tested. `Graph/layout.ts` orders the cards, cuts the target line at the stacks' bases, folds long runs (10 rows, then 20 per ask) and emits the SVG paths. `Graph/usePlan.ts` feeds it from the queries and the fold state in `projects/graph.ts`.

**Drawing**

- Rows draw their own rails. `GraphSegment` is two-tone now, so the stretch between two icons takes the lower icon's colour and push states read along the rail.
- Cards draw everything inside their edges: the stub over the floor, and the main line passing behind a forked card.
- One SVG draws only the gaps between cards, from the virtualiser's measurements (`Graph/Rails.tsx`).
- The upstream section draws its own rails in CSS (`Graph/Section.module.css`): it is not virtualised, its folds animate, and the folded target card docks at the scroller's foot.

**Behaviour**

- The target's card folds the incoming commits. Folded, it docks so the news stays in view, and its toggle scrolls to its place before opening.
- The merge base row says how many commits are new, folds the history below the base (loaded a page at a time, forgotten on fold) and carries the Update button.
- Cards and section rows share the applied cursor: arrow keys walk from the cards into the upstream rows, the fold key closes the fold the cursor sits in, and a target commit opens in Details.

Verified with the unit tests in `Graph/layout.test.ts`, the lite gates, and the committed branch alone in a scratch worktree.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #15756, opened 2026-09-06
Dock the merge base row, not the target's card

mtsgrd:
Follow-up to #15736. The merge base row now says how many commits are new and carries Update, so it is the row to keep in view while the stacks scroll.

- The merge base row is sticky while its history is folded. Stuck at the scroller's foot it gets 4px of air all round, a hairline on top, and no line into its chevron, since what scrolls past above is not the line it would claim to come from. At its place it is unchanged. Its toggle scrolls to its place before opening, as the target's card did.
- The target's card stays in flow. It draws the gap under itself, the leg's bend and the main line through it, as a small inline SVG, and the ref row draws its gap the same way. The docked row carries nothing outside its box, so docking needs no scroll tracking: a scroll-state container query tells the row's own pseudo-elements whether it is stuck.
- A row scrolled into view clears the docked row's height.

Verified in a fixture with master five commits ahead: docked over the cards, opened from docked with the scroll landing at the bottom, at its place under the card, and with the incoming card open.

### DIFF
--- a/apps/lite/ui/src/components/GraphSegment.module.css
+++ b/apps/lite/ui/src/components/GraphSegment.module.css
@@ -16,8 +16,7 @@
 	}
 }
 
-/* A stretch of the rail in another status's colour, given by its own status;
-   without one it keeps the glyph's. */
+/* A stretch of the rail in another status's colour; without one, the glyph's. */
 .tone {
 	color: var(--commit-status-color);
 }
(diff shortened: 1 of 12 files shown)


mtsgrd:
### Screenshots

Captured ad hoc at `c4792cb57c` from the `project-with-remote-branches` fixture with three stacks and five commits incoming on `origin/master`. Not part of the screenshot catalogue, so later pushes are not compared against these.

**Merge base folded, docked at the scroller's foot**

merge base docked

**Merge base open, at its place under the target's card**

merge base open

PR #15768, opened 2026-09-07
Push before creating a pull request in Lite

PavelLaptev:
A forge can only open a review on a branch it has. Lite's create-PR form used to answer that by disabling its button whenever the branch or its parent had never been pushed, with nothing saying why. Desktop and the CLI both push first and then create; Lite now does the same.

## What changed

- The publish mutation takes the push the PR has to wait for and runs it ahead of the forge call when any branch down the stack still has something to push. This matches the desktop rule rather than only covering the first push, so the PR always opens against the current commits. Force is used when any of those branches needs it, the same way the sidebar's Push button decides.
- The PR source is the name the branch landed under on the remote, which differs from the local name when the branch tracks another remote.
- A failed push toasts as "Failed to push" rather than "Failed to create pull request", and the push endpoint's caches are invalidated alongside the review ones.
- The button reads "Push and create a PR" while a push is pending, so the extra step is visible before it happens.
- The only remaining disable here is conflicted commits, which cannot be pushed. The target branch now shows the parent by name even before that parent is pushed, since the push covers it.

One case this opens up is handled in the stacked follow-up #15771: a branch with no commits cannot open a PR, since a forge refuses a PR whose head adds nothing to its base, and pushing first would only leave an empty branch on the remote. That PR holds the button on such a branch with the reason in its label.

## Verification

Typecheck, unit tests, oxlint, oxfmt, prettier, and knip are clean. Checked in the dev app on a branch that had never been pushed: the button is enabled and labelled "Push and create a PR", with the target shown as the project's base branch. The push-then-publish path is the same Rust path `but pr new` and the sidebar's Push button already use.

Related: GB-1959, which lists "PR panel, no pull request" as an empty state to design. With the gate gone there is no such state left on applied branches.

## Screenshots

This button

Screenshot 2026-09-07 at 13 22 38

### DIFF
--- a/apps/lite/ui/src/api/mutations.ts
+++ b/apps/lite/ui/src/api/mutations.ts
@@ -52,9 +52,7 @@ import type {
 	TreeChange,
 } from "@gitbutler/but-sdk";
 import { type QueryClient, useMutation, useQueryClient } from "@tanstack/react-query";
-import type { PayloadFor } from "#electron/ipc.ts";
 import type { GUISettings } from "#electron/settings.ts";
-import { invalidateDeclared } from "#ui/api/tags.ts";
 import { moveDraftPR } from "#ui/pr.ts";
 import { presentableOperation } from "#ui/snapshot.ts";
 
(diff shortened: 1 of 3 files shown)


apps/lite/ui/src/api/mutations.ts
```diff
@@ -215,12 +217,70 @@ export const useBranchCheckoutNew = () => {
...
+}
+
+/**
+ * Creates a pull request, pushing the branch and its ancestors first when
+ * `push` says any of them still has something to push — a forge can only
+ * open a review on a branch it has. The PR's source is then the name the
+ * branch landed under on the remote, which differs from the local one when
+ * the branch tracks another remote.
+ */
+export const usePublishReview = (projectId: string) => {
```
samhh: At first glance it'd be cleaner not to prescribe pushing in this hook and handle this UX dependency at the call sites instead.

PavelLaptev: Agreed, done in 314e801093. `usePublishReview` is back to its plain shape and the form runs the existing push hook first, then publishes on its result. That also drops the custom error class, the hand-written toast and the manual invalidation, and the sidebar now sees the push under its own mutation key while it runs.

PR #15776, opened 2026-09-07
Remove the Upstream tab from Lite

PavelLaptev:
The stacks graph already lists the incoming commits and the merge base's history under the stacks, so the Upstream tab repeated what the workspace page shows. This removes the tab and everything only it used.

## What goes

- The tab and its count badge in the sidebar. `[` and `]` now cycle between Workspace and Branches.
- The upstream list, its hook, and the redux slice holding its expanded segments.
- The `upstream` page id and cursor from the URL layer. A stale `?page=upstream` link opens the workspace instead of erroring.
- The upstream selection and details placeholder on the workspace page.
- The upstream screenshot test and the Upstream badge assertion in the sidebar e2e test, with the mocked target commits that only fed that badge.

## What stays

The graph's upstream section borrowed the commit row from the upstream list. It moves to `Graph/TargetCommitRow.tsx`, trimmed to the applied list the graph uses.

## Verification

`pnpm -F @gitbutler/lite check`, unit tests, oxlint, knip, oxfmt, and prettier all pass. In the dev app the tab bar shows Workspace and Branches, and the merge base history under the stacks still renders its commit rows.

## Screenshots

Before:

image

After:

image


mtsgrd (approved):

PavelLaptev (commented):
Reviewed the full diff against the stacks graph, the URL cursor layer, and the e2e tests.

**Correctness**

- The moved `TargetCommitRow` keeps every behaviour the graph relied on: applied-list cursor, `treeitem` semantics, `inert` for rows a pending operation drops, `railEnds`, and scroll-into-view (previously gated on `list === "applied"`, which is now the only list, so it is unconditional). `status` became required, and both call sites in `Section.tsx` already pass it.
- `validateSearch` in `routes.tsx` now drops `page=upstream` and the `upstream` param, so stale URLs saved by `writeLastPlace` degrade to the workspace page rather than erroring. `WorkspaceCursorSnapshot.page` narrows to `"branches"`, which matches the only remaining non-default page.
- `remapSearchCommits` / `remapSearchBranch` iterate `["applied", "unapplied"]` only; no other param carried addresses.
- `sidebar.spec.ts` still asserts the Workspace tab has no badge, which is the property the test is named for; the mocked target commits only existed to light the Upstream badge and are gone with it.

**Cleanup applied in this pass**

- `GraphSegment` still carried the `groupHead` and `parentHead` glyphs, which only the deleted shared-history expander drew. Removed them along with the `isGroupGlyph` helper they justified. Amended into the commit.

**Left as is, deliberately**

- `inbox.svg` / `inbox-fill.svg` stay in the icon set. Icons are a curated library rather than dead code, and `inbox-fill` was unused before this PR too.
- `groupCenteredGlyph` and the `centered` prop remain in use by the graph's elided rows.

**Verification**: `pnpm -F @gitbutler/lite check`, unit tests (342 + 42), oxlint, knip, oxfmt, prettier all clean after the amend.

PR #15782, opened 2026-09-07
Draw the graph's gutter row by row

mtsgrd:
The stacks graph's rails were drawn by four mechanisms: rows drew their glyphs, cards drew a stub and a through line with pseudo-elements, one SVG drew the gaps between cards from the virtualiser's measurements, and the upstream section drew its own with pseudo-elements. Now every row draws its whole gutter and a card draws the gap under it, so nothing is drawn from measured pixels, and no fold, dock or virtualisation can put a line out of step with its row. The look is unchanged.

- A row's gutter is a run of columns at a 12px pitch: one per line running behind the row, then the glyph's own. `GraphSegment` takes `behind`, the count of those columns, drawn as the translucent line the cards drew over their rows.
- The gap under a card is a row too, `GraphGap`: the main line through it, and the card's own line bending onto it. A stack card renders its gap inside its measured element, so the virtualiser has no gap of its own and `Rails.tsx` goes.
- The section's headers draw the rail around their chevron with the `control` glyphs, and the target's card is built like a stack card: an air row, its header, the incoming rows, a stub over its floor, then the gap with the leg's bend.
- `layout.ts` keeps the plan and the section's constants. The fold overlays and the commit checkbox sit on the glyph's column.

This is the mechanism a DAG renderer needs: a second lane is more columns per row, not another drawing layer. Verified against the previous look to the pixel, with the lite gates and the fixture captures.


PavelLaptev:
@mtsgrd screenshots pls

mtsgrd:
There's no visual change in this one!

PavelLaptev:
> There's no visual change in this PR!

Sorry, the bot tagged me here.

> 🔵 Needs a closer look
> It makes broad, tightly-coupled changes across virtualization, layout constants, and rendering/CSS for the graph gutter, which warrants final human verification of visual and interaction correctness.

This confused me as well.

PR #15783, opened 2026-09-07
Put the uncommitted files at the top of the graph

mtsgrd:
One section on the workspace tab instead of two, as `but status` draws it: the uncommitted files head the trunk, and every stack card is a lane off it.

## What changed

- The uncommitted files are a card at the head of the graph's scroller: a header row with the trunk's head and a fold, the file rows with the trunk as their rail, the last-commit line and the commit form as rows on it. The filter row sits under the header rather than in its place.
- Every stack card is a lane off the trunk, the top one included, and its gap bends onto the trunk.
- The card's head, header and filter row, sticks at the scroller's top while its rows scroll under it, so the filter stays in reach. The filter button goes while the card is folded.
- Once the card has scrolled out above, a docked row stands in for it at the scroller's head, without chevron or toolbar. A click scrolls back to the card.
- The panel splitter, the "Stacks and branches" header with its fold and summary, and the two-panel fold state go. One heading row, "Workspace", keeps the new-branch menu.

## Decisions to confirm

- The "Workspace" heading row above the scroller stays for now, since the other tabs each have one.
- The commit form stays at the card's foot and scrolls with it. Docking it at the scroller's head, as the merge base row docks at its foot, is the follow-up if a long file list makes it a reach.
- A clean worktree shows "no changes" beside the header, with no empty row under it.

## Not in this change

- Docking the commit form.
- The fold hotkey on a file row folding the card.

## Verification

- Lite typecheck, oxlint, prettier, unit tests, and the `workspace-focus` and `commit-form` e2e specs pass.
- Checked live: the head stuck with its hairline, the docked row taking over as the card leaves, a keyboard walk keeping the selected row clear of the stuck head, the fold hiding the filter button.


mtsgrd:
## 📸 UI screenshots

Captured 4 surfaces, 4 changed.

### uncommitted-long-list

| Before | After |
| --- | --- |
|  |  |

### uncommitted-small-list

| Before | After |
| --- | --- |
|  |  |

### uncommitted-docked

### uncommitted-head-stuck

Captured ad hoc for this pull request, not by the catalogue, so these surfaces are not covered going forward. "Before" is `origin/master`; the two scrolled states have no counterpart there.

- **small list**: two added files and one modified.
- **long list**: 37 files, at the top.
- **head stuck**: the long list scrolled 240px; the card's header holds at the scroller's top.
- **docked**: the top branch grown to 17 commits and the list scrolled past the card; the stand-in row takes the scroller's head.

PR #15785, opened 2026-09-07
Turn signing on in fixture scripts without dropping the harness isolation

mtsgrd:
Fixes the `workspace_commit_should_not_be_allowed_to_conflict` flake (13 CI failures, a bare ENOENT): the signed fixture scripts replaced `GIT_CONFIG_COUNT` with their own entries, dropping gix-testtools' isolation (`maintenance.auto=false`), so a detached `git maintenance` raced the fixture copy.

- the scripts now append `commit.gpgsign=true` to `GIT_CONFIG_PARAMETERS`, which git applies after `GIT_CONFIG_COUNT` and after the harness's own parameters, so the isolation survives and signing stays on
- bumps gix to `4ec7c08e47`, upstream main as of today, which includes GitoxideLabs/gitoxide#2982: gix-testtools now passes its isolation through `GIT_CONFIG_PARAMETERS`, so a script can no longer drop it by setting `GIT_CONFIG_COUNT`. Crate versions are unchanged; the 22 intervening commits include two `gix-config-value` parity fixes marked breaking (boolean numbers, `~` expansion)
- commit hashes and snapshots are unchanged
- deletes two signed line-offset fixtures in but-workspace that nothing referenced

Verified with `cargo check --workspace --all-targets` on the new rev, `cargo test -p but-rebase -p but-core` with fixture caches cleared so every script re-ran under the new harness, and a local git check of both harness shapes.


mtsgrd:
@krlvi could we land this?

slarse (approved):
This one has been bothering me, nice.

The one thing that bothers me about this fix is that it's very fragile. There's nothing stopping us from introducing this same error in new fixtures.

mtsgrd:
Ok, merging now, but can you think of a way we can make this future proof?

crates/but-core/tests/fixtures/scenario/single-signed.sh
```diff
@@ -11,11 +11,15 @@ git config user.signingKey "$PWD/signature.key"
...
-export "GIT_CONFIG_VALUE_1=main"
+# Append to the entries gix-testtools passes via GIT_CONFIG_COUNT. Replacing
+# them drops its maintenance.auto=false; each commit then spawns a detached
+# 'git maintenance' that races the fixture copy.
+base_config_count=${GIT_CONFIG_COUNT:-0}
+export "GIT_CONFIG_KEY_${base_config_count}=commit.gpgsign"
+export "GIT_CONFIG_VALUE_${base_config_count}=true"
+export "GIT_CONFIG_KEY_$((base_config_count + 1))=init.defaultBranch"
+export "GIT_CONFIG_VALUE_$((base_config_count + 1))=main"
+export "GIT_CONFIG_COUNT=$((base_config_count + 2))"
```
slarse: Looking at this again, this would also benefit from an `append_git_config` helper function, like the `but-rebase` tests get.

slarse (commented):
@mtsgrd I'd say a blanket disallowing of using `GIT_CONFIG_*` in any scenario is a good start. We could put a check like this in CI, for example:

```bash
if rg --files-with-matches 'GIT_CONFIG_KEY' | grep 'fixtures/scenario' | grep -v 'shared.sh'; then
    echo 'Found scenarios that inappropriately override GIT_CONFIG_* variables, use append_git_config instead'
    exit 1
else
    echo 'All scenarios abiding by the GIT_CONFIG_* rules'
    exit 0
fi
```

That would require scenarios to use a shared helper. It's not _rock solid_ as it fails if the scenarios are put in some subdirectory that doesn't follow the `fixtures/scenario` convention, but it will work for existing places where we have scenarios. WDYT?

I also realize I'm not powerful enough to approve this PR, hehe :D

crates/but-core/tests/fixtures/scenario/single-signed.sh
Byron: https://github.com/GitoxideLabs/gitoxide/pull/2982 will address this centrally, and make overriding these variables possible without a rain dance. It will take a moment to land in GB, but when it does I will make sure to role this back.

mtsgrd:
I put this back to draft while we await the next gix release.

Byron:
@mtsgrd I think it's completely fine to update `gix` in this PR - the fix is already merged I believe.

mtsgrd:
@Byron I've added gix version bump to this pr. Can you review and possibly approve?

crates/but-core/tests/fixtures/scenario/single-signed.sh
```diff
@@ -11,11 +11,9 @@ git config user.signingKey "$PWD/signature.key"
...
 echo "*.key*" >.gitignore
 
-export "GIT_CONFIG_COUNT=2"
-export "GIT_CONFIG_KEY_0=commit.gpgsign"
-export "GIT_CONFIG_VALUE_0=true"
-export "GIT_CONFIG_KEY_1=init.defaultBranch"
-export "GIT_CONFIG_VALUE_1=main"
+# Appended to git's command-scope config, which outranks the harness's commit.gpgsign=false
+# however gix-testtools passes it.
+export GIT_CONFIG_PARAMETERS="${GIT_CONFIG_PARAMETERS:+$GIT_CONFIG_PARAMETERS }'commit.gpgsign=true'"
```
Byron, line 16: While I don't love that this leaks into fixtures this way, which is something I didn't think of while naively assuming that GIT_CONFIG_COUNT (etc) stacks with GIT_CONFIG_PARAMETERS, I think it's acceptable given the limited amount of alternatives.

The format also has its benefits as it's inarguably less verbose and more readable.
It's probably also confusing that those who try to use `GIT_CONFIG_COUNT` will just see it silently ignored, which is something that speaks for the previous solution of fiddling with global configuration files, but which would also bring back the verbose `GIT_CONFIG_COUNT`.

Long story, let's go with this :D.

Byron (approved):
Thanks, looks good to me.

crates/but-core/tests/fixtures/scenario/single-signed.sh
mtsgrd: Ok, great. :)

mtsgrd:
@krlvi could you approve this?

krlvi (approved):

PR #15787, opened 2026-09-07
Add an optional setting to put the file list right of the diff

mtsgrd:
I think it would feel more natural to have the changes file list on the right?

This adds a switch under Appearance → Files, "File list right of the diff". It defaults to off, so nothing changes until it is turned on. On, the commit, branch and uncommitted views show the diff first with the file list to its right, keeping it away from the sidebar's own file list.

### DIFF
--- a/apps/lite/electron/src/settings.ts
+++ b/apps/lite/electron/src/settings.ts
@@ -26,6 +26,7 @@ const guiSettingsV1 = type({
 	"dryRunOperations?": "boolean",
 	"editorId?": "string",
 	"fileDisplayMode?": "'list' | 'tree'",
+	"filesPanelRight?": "boolean",
 	"lineDiffType?": "'word-alt' | 'word' | 'char' | 'none'",
 	"minimap?": "boolean",
 	"pathFirst?": "boolean",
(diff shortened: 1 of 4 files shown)


mtsgrd:
## 📸 UI screenshots

Captured 9 catalogue surfaces, 1 changed, plus the settings page captured from the dev app.

### settings-appearance

The new switch under Appearance → Files. Captured from the running dev app rather than the fixture harness, since the Appearance page is not in the catalogue.

| Switch off (default) | Switch on |
| --- | --- |
|  |  |

### details-pane

The layout the switch controls. "Before" is also what the pane looks like with the switch off; "After" is the switch on.

| Before | After |
| --- | --- |
|  |  |

The commit view shares this layout and is not in the catalogue; it was checked live in the dev app.

<details>
<summary>8 surfaces unchanged</summary>

`branches-tab`, `commit-form`, `pr-form` and `uncommitted` are byte-identical. `edit-mode`, `project-picker`, `settings` and `workspace-sidebar` differ only in run-to-run noise (a fixture commit id, a text caret, "Ns ago").

</details>

PR #15788, opened 2026-09-07
Take the keychain prompt off Lite's launch path

mtsgrd:
Lite nightly took 8.5 s to show its window on the first launch of each build. The tracing log and `securityd` show why: `initMetrics` is awaited before the window is created, it read the profile through `get_user`, and that checks the token in the keychain. The shared token item trusted only the desktop nightly, so each new Lite build got a keychain prompt and the window waited behind it.

- `getUserProfileLocal` reads the stored profile only; the token is checked by the first call that needs it
- the remote failure-limit fetch is no longer awaited before the window
- the login-shell environment is fetched async while Electron boots, instead of synchronously at module load (about 0.5 s of zsh startup)
- the `but` scheme gets `codeCache` and the window `bypassHeatCheck`, so the 3 MB renderer bundle is not recompiled from source on every launch

Measured on a local packaged build of this branch, warm launch, spawn to event:

| Stage | Before | After |
|---|---|---|
| Page loaded | 1.6 s, plus the prompt when it appears | 0.78 s |
| Window visible | | 0.74 s |
| Keychain prompt at launch | on each build the item does not trust | none |

The workspace stays empty until `changes_in_worktree` returns, 1.9 to 5 s on a large dirty tree here. That is separate work on the Rust side.

### DIFF
--- a/apps/lite/electron/src/main.ts
+++ b/apps/lite/electron/src/main.ts
@@ -49,7 +49,9 @@ import { type GUISettings, readSettings, writeSettings } from "./settings.js";
 import { initMetrics, metricsOnLogin, shutdownMetrics, withApiCommandCapture } from "./metrics.js";
 import { apiParamNames } from "@gitbutler/but-sdk/api-param-names";
 
-Object.assign(process.env, interactiveLoginShellEnvironment());
+// Started now so the shell's startup overlaps with Electron's; applied once ready, before
+// anything that could spawn a process.
+const shellEnvironment = interactiveLoginShellEnvironment();
 
 const isHeadless = process.env.GITBUTLER_LITE_HEADLESS === "true";
 if (isHeadless && process.platform === "darwin") app.setActivationPolicy("accessory");
@@ -590,6 +592,7 @@ if (!app.requestSingleInstanceLock()) {
 
 void app.whenReady().then(async () => {
 	initLogging();
+	Object.assign(process.env, await shellEnvironment);
 	applyGUISettings(await readSettings());
 	await initApplicationNamespace(null);
 	configureAskpass();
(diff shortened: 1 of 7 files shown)


apps/lite/electron/src/main.ts
```diff
@@ -590,6 +596,7 @@ if (!app.requestSingleInstanceLock()) {
 
 void app.whenReady().then(async () => {
 	initLogging();
+	Object.assign(process.env, await shellEnvironment);
```
Byron, line 599: I think this is great as it won't risk the environment not being configured when the app is ready. This also seems to fix a bug as previously it didn't await it.

crates/but-api/src/legacy/users.rs
```diff
@@ -114,10 +114,13 @@ impl From<User> for UserProfile {
...
 
 /// The signed-in account, or `None`. Credentials stay in this process.
+///
+/// Only the stored profile is read: the keychain, which may prompt on macOS, is left to the
+/// first call that needs the token.
 #[but_api(napi)]
 #[instrument(err(Debug))]
 pub fn get_user_profile_local() -> Result<Option<UserProfile>> {
-    Ok(get_user()?.map(Into::into))
+    Ok(gitbutler_user::get_user()?.map(Into::into))
```
Byron, line 123: This also constitutes a bugfix, but merely because `get_user()` has massive side-effects. That's likely to bite another time elsewhere if `get_user()` is still called.

Byron (approved):
This is great, but I also I can only speak for certain parts. Maybe @samhh has ideas about the Chromium parts.

mtsgrd:
Thanks Byron, let me go ahead and land this, build a new nightly, and if it doesn't have the expected impact or if it breaks anything I'll come back to it.

samhh (approved):
Overlaps with some changes I had in flight, so for posterity:

- Not awaiting that function in metrics speeds up launch by ~4%.
- The two code caching properties (atop another change of mine, anyway) speed up warm launches by ~6%.

Also `bypassHeatCheck` remains lazy, which isn't quite clear from the clanked comment.

PR #15791, opened 2026-09-07
Lite: switch a PR's status from its badge

PavelLaptev:
Updates the PR panel's Status row to match the [design](<https://www.figma.com/design/EBuHQGUcCaSw4Ln5uVpWkn/Lite?node-id=4855-52769>): the status badge sits in the section header next to the PR link and doubles as the status switcher.

* The badge grows a chevron and opens a native menu with **Open**, **Draft** and **Closed**, the current one checked. Draft and Open use the draftiness mutation, Closed uses the review update the toolbar menu already used to close and reopen. Reopening restores the draftiness the PR was closed with, so landing on the other of Open and Draft chains a second step once the PR is open again. A merged PR keeps a plain badge with no chevron.
* The standalone "Convert to draft" / "Mark as ready" button is gone; the menu replaces it.
* Both the badge and the PR link get hover states. The badge deepens its own tint using the same transition as buttons, now shared as a `--transition-button` token. The light gray Badge variant derives its tint from the shared opacity variable so it can lift on hover; its resting look is unchanged, though on an inverted selected row it now lifts along with the colored variants.
* No pointer cursors

Also includes the two earlier commits on this branch for the empty Reviewers and Labels sections and the Branches row. The gray stand-in shapes in the empty sections went partly because they looked like a skeleton loader: gray circles and bars where the content would be are what every app draws while loading, so the section read as stuck rather than empty. An "Add reviewers" button says what fills it instead, and a control is the one thing a skeleton never shows. That rule is now in `DESIGN.md` under Empty states.

## Screenshots

Before:

image

After:

image

### DIFF
--- a/apps/lite/DESIGN.md
+++ b/apps/lite/DESIGN.md
@@ -54,6 +54,16 @@ thing; dark mode is handled by the tokens. Note that selected rows reach these
 styles through CSS in `Row.module.css` rather than by passing the variant, so
 selection can restyle without a re-render.
 
+## Cursors
+
+**No pointer cursors.** Lite is a desktop app, and desktop apps keep the arrow
+over buttons, menus and rows; the hand is a web convention for links out to a
+page. Don't set `cursor: pointer` on a control, and don't reintroduce it by
+resetting a `<button>` — the browser default for buttons is already the arrow.
+Interactivity is shown by the hover state, not the cursor. The cursors that do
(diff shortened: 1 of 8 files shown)


PavelLaptev:
cc @samhh
I remember you mentioned this looked like a skeleton state, tried to fix it here.

PR #15794, opened 2026-09-07
Call the merge base row Base, and count what it is behind

mtsgrd:
Wording only, on the graph's foot row.

- The row reads **Base**, or the ref name with a quiet **base** caption when the ref sits on the base.
- A moved-on target says **3 behind** rather than **3 new**.
- The fold's accessible name follows: "Unfold the base's history".


mtsgrd:
image

PR #15796, opened 2026-09-07
Check out a remote-tracking branch through its local tracking branch

mtsgrd:
## Problem

Lite's apply-conflict toast offers "Switch to branch instead", and it failed for every branch that only exists on the remote:

- the apply picker hands `refs/remotes/origin/<name>` to `apply`
- a conflict-aborted apply never creates the local branch
- `branchCheckout` refused anything outside `refs/heads`

```
Failed to switch branch
Can only check out local branches under refs/heads, got 'refs/remotes/origin/lite-workspace-base-rows'
```

## Change

- `branch_checkout` accepts a `refs/remotes/` ref and checks out its local tracking branch, creating it at the remote commit with `branch.<name>.remote/merge` config when missing. This is what `git checkout <name>` and a successful `apply` already do.
- An existing local tracking branch is used as is, even when it diverged from the remote.
- Refs that are neither local nor remote-tracking branches (tags etc.) are still rejected.
- The tracking-config recipe moves out of `apply` into `but_workspace::branch` so both paths share it.
- Generated SDK typings regenerated for the doc comment; no endpoint signature changed.

Lite needed no code change: the existing `branchCheckout` call now works. `but switch` keeps its own local-only guard.

## Tests

- remote-only ref creates and checks out the local tracking branch with tracking config
- remote ref with an existing local switches to it without moving it
- tags are rejected

### DIFF
--- a/crates/but-workspace/src/branch/mod.rs
+++ b/crates/but-workspace/src/branch/mod.rs
@@ -561,8 +561,9 @@ pub(crate) fn setup_local_tracking_configuration(
     let mut config = repo.config_file_mut(repo.common_dir().join("config"))?;
     let mut section =
         config.section_mut_or_create_new("branch", Some(local_tracking_ref.shorten()))?;
-    // Only edit the configuration if truly empty, let's not overwrite user data.
-    if section.num_values() == 0
+    // Leave tracking the user already configured alone; other keys in the section are no reason to.
+    if section.value("remote").is_none()
+        && section.value("merge").is_none()
         && let Some((upstream_branch, remote)) =
             repo.upstream_branch_and_remote_for_tracking_branch(remote_tracking_ref)?
     {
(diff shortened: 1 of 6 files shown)


crates/but-workspace/src/branch/mod.rs
```diff
@@ -545,6 +545,81 @@ pub(crate) fn try_find_validated_ref<'repo>(
...
+        local_tracking_ref.as_ref(),
+        commit_id,
+        gix::refs::transaction::PreviousValue::MustNotExist,
+        format!(
+            "GitButler creates local tracking for {}",
+            remote_tracking_ref.as_bstr()
+        ),
+    )?;
+    Ok(local_tracking_ref)
+}
```
Caleb-T-Owens, line 631: Should we make the reference before updating the config?

PR #15800, opened 2026-09-08
Forget a forge account without its token, and find the credential helper in dev builds

mtsgrd:
Two bugs behind "the Forget button in Lite's GitHub settings does nothing".

**Forget was a silent no-op when the token could not be read.** `delete_*_access_token` looked the account up by first retrieving its token from the keystore. An account whose token was unreadable (stored by another build kind, or a keychain refusing the read) was never found, so the entry stayed in `forge_settings.json` and the call still returned success. Lite refetched the list and showed the same account. Now the stored entry is matched by identifier and removed, then the secret is deleted; a missing secret was already a no-op there. Same shape for GitHub, GitLab and Bitbucket, with tests for the missing-token case on GitHub and GitLab.

**Debug builds never saw a `credential.helper` set in the git installation's config.** Debug builds on macOS keep forge tokens through git credential helpers. `but-secret` loads git's global config through gix, which skips the installation config unless `git_binary` is enabled. A helper configured only there (Homebrew's `/opt/homebrew/etc/gitconfig`) was invisible, so storing a token silently wrote nothing and every read came back empty. Anyone with the helper in `~/.gitconfig` never hit this. Loading the config with `git_binary` enabled fixes it; verified with an in-process persist/retrieve/delete round-trip.

No API surface changes.


mtsgrd:
@Byron could you take a look at this? @samhh found it fixes a problem introduced by https://github.com/gitbutlerapp/gitbutler/commit/8875cbd756.

krlvi (approved):

PR #15801, opened 2026-09-08
Take Lite's transition timings from design-core tokens

PavelLaptev:
Lite's transition durations and the popup curve now come from design-core instead of two hand-rolled variables in global.css.

design-core 3.17.0 adds `--transition-fast` (80ms), `--transition-medium` (150ms), `--transition-popup` as an alias of medium, and `--easing-popup`, the hard ease-out popups open with. Each carries a description in ⚛️ Lite Core saying what it is for.

**Wiring**

- Bump `@gitbutler/design-core` to 3.17.0.
- Drop the local `--transition-fast` and `--transition-popup` from global.css. They shared names with the tokens and would have overridden them. `--transition-button` stays, since it is a Lite composite built on the fast token.
- The old local popup value bundled duration and curve, while the token is duration only, so every popup transition now pairs `var(--transition-popup)` with `var(--easing-popup)`: modal, dropdown, backdrop and toolbox.
- Consumers of `--transition-fast` need no change. The old value's `ease` is the CSS default.

**Literals moved onto tokens**

- `--transition-medium`: both Switch transitions (120ms), the Settings chevron (0.15s), the Graph section fold (180ms, keeps its `ease-out`), and the DiffMinimap fade-in (0.15s).
- `--transition-fast`: the Button icon opacity fade (0.08s).

Left as literals on purpose: the minimap's 0.4s fade-out (no slow tier yet), a 50ms fade in ui.module.css, and the spinner and fresh-change keyframes, which are loops rather than transitions.

**DESIGN.md**

A new Motion section records the two tiers and what each is for, that popup is medium with a curve and always travels with `--easing-popup`, that easings outside popups stay keywords (with the Figma preset mapping for `ease`), that loops and holds are not transitions, and that anything that moves respects reduced motion. Like the Empty states section, it points at the Lite Core token descriptions as the other copy of the rules.

No motion changes. Every edit swaps a literal for a token of the same value, and the popup curve moves from `cubic-bezier(0.45, 1.005, 0, 1.005)` to `(0.45, 1, 0, 1)`, which is what Figma exported.

### DIFF
--- a/apps/lite/DESIGN.md
+++ b/apps/lite/DESIGN.md
@@ -64,6 +64,50 @@ Interactivity is shown by the hover state, not the cursor. The cursors that do
 change are the ones that describe a gesture: `text` over editable text, `grab`
 and `grabbing` while dragging, and the resize cursors on a splitter.
 
+## Motion
+
+**Two speeds, both tokens.** Every transition takes its duration from
+design-core. `--transition-fast` (80ms) is for a state change on a control that
+stays where it is: hover and press on a button, the outline of a focused field,
+a small opacity fade. `--transition-medium` (150ms) is for something that moves
+or changes shape: a switch thumb, a chevron turning, a section folding, the
+minimap fading in. A move on the fast tier reads as a jump with a flicker on
+it; a hover on the medium tier feels laggy. Don't write a duration by hand. If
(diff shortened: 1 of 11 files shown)


PavelLaptev:
it's timing only. So no screenshots in this one.

PR #15813, opened 2026-09-08
Ask about usage metrics on the Lite welcome page

mtsgrd:
The desktop app asks before it starts collecting; Lite started with the shared default and never asked. The welcome page now carries the question, and adding the first repository is the answer.

- A **Usage metrics** switch on the welcome page, shown until onboarding is marked complete. Adding the first repository marks it.
- The switch writes the shared setting straight away through three new local endpoints (`getAppSettings`, `updateTelemetry`, `updateOnboardingComplete`), and the main process starts or stops collection to match, so no relaunch is needed.
- The same row sits under a **Telemetry** heading in General settings, so the choice can be revisited.
- The onboarding e2e spec now checks the switch is there and that the add marks onboarding complete.

### DIFF
--- a/apps/lite/e2e/tests/onboarding.spec.ts
+++ b/apps/lite/e2e/tests/onboarding.spec.ts
@@ -1,4 +1,5 @@
 import path from "node:path";
+import type { LiteElectronApi } from "../../electron/src/ipc.ts";
 import { LiteTestId } from "../../ui/src/testIds.ts";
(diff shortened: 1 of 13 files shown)


mtsgrd:
## Screenshots

Welcome page, captured ad hoc at 81ef473 (the page is not in the screenshot catalogue, so this is a one-off, not ongoing coverage).

### Before

welcome-before

### After

welcome-after

PR #15818, opened 2026-09-08
Lead notification rows with the event, and notify the desktop

mtsgrd:
Two changes to Lite's PR-activity notifications, both driven by actionability: what happened, to what, and by whom.

## Rows

- The first line is now the event and its actor: `Changes requested by alice`, `3 comments from dave`, `Review requested by erin`.
- The second line is the target: branch name, then `#N`. The PR title moves to the row's tooltip.
- The snippet, when there is one, stays on the third line.
- Loud kinds (comments, mentions, verdicts, review requests) take the strong text color. Quiet kinds (pushes, merges, closes) sit muted.
- Icons carry meaning: warn for changes requested, safe for approved, pop for mentions.

## Desktop notifications

- Loud entries also show as a system notification, through a new `showNotification` host member. The main process decides whether the window is focused, because a freshly loaded document reports `document.hasFocus()` true while the window sits behind another app. A click focuses the window and lands on the entry, on the same path as a bell click.
- A refused notification is logged with the OS reason. On macOS the ad-hoc signed dev Electron is always refused (`UNErrorDomain error 1`), so this can only be seen on a signed build.
- More than three loud entries in one poll collapse into one summary notification.
- The review listing stops polling in an unfocused window, which is exactly when a desktop notification is the only way to be heard. The detector polls in the background while the setting is on.
- New `desktopNotifications` setting, on by default, disabled unless PR activity is set to Loud.

## Notes

- The loud/quiet split is one table, `inboxKindAttention`, used by both the row weight and the desktop gate.
- Entry navigation moved from the bell into the detector hook so the harness panel, which mounts the hook but not the bell, exercises the notification click end to end.
- Bot comments are still loud, and your own PRs on unapplied branches still only surface mentions. Both are worth a follow-up.
- The app is expected to be running for a desktop notification to arrive: hidden or behind another app is fine, quit is not.

### DIFF
--- a/apps/lite/electron/src/endpoint-table.ts
+++ b/apps/lite/electron/src/endpoint-table.ts
@@ -10,7 +10,12 @@ import { exposedEndpoints, type Endpoint, type LiteElectronApi, type PayloadFor 
  */
 
 /** Members the renderer implements itself; they have no host-side handler. */
-type RendererOnlyKey = "onAskpassPrompt" | "onDeepLink" | "onFullScreenChange" | "platform";
+type RendererOnlyKey =
+	| "onAskpassPrompt"
+	| "onDeepLink"
+	| "onFullScreenChange"
+	| "onNotificationClick"
+	| "platform";
 
 /** Handlers needing the transport event itself, or taking variadic arguments. */
 type ImperativeKey =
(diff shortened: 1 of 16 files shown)


mtsgrd:
## 📸 UI screenshots

Captured 9 catalogue surfaces plus one ad-hoc surface; 2 changed.

### notification-bell

Captured ad hoc with a seeded inbox (one entry of every kind); this surface is not in the catalogue and is not covered going forward.

| Before | After |
| --- | --- |
|  |  |

### settings

| Before | After |
| --- | --- |
|  |  |

<details>
<summary>8 catalogue surfaces unchanged</summary>

`branches-tab`, `commit-form`, `details-pane`, `pr-form`, `uncommitted`, and three that differed only by capture noise (a relative time, a fixture sha, a caret blink): `edit-mode`, `project-picker`, `workspace-sidebar`.

</details>

PR #15821, opened 2026-09-08
Serve the but skill from the CLI

krlvi:
Bare `but skill` prints the skill guide embedded in the binary, with
`reference`, `concepts`, and `examples` subcommands for the reference
files and `--full` to print everything. The content is always the one
matching the installed binary, so agents can read it without an
installed copy going stale.

A hidden `--stub` flag on `skill install` and `agent setup` writes a
single SKILL.md that keeps the trigger frontmatter and tells the agent
to run `but skill`. Stubs carry `stub: true`, count as complete
installations, and are refreshed as stubs. The flag is team-internal
while the approach is evaluated against the full install.

Pointers inside the skill files now name `but skill <doc>` instead of
file paths, so they resolve for both installed and served copies.


krlvi:
Opened by mistake while publishing a stacked branch; closing. The branch is unchanged.

PR #15822, opened 2026-09-08
Project the integrated workspace against the new target

krlvi:
In Lite, updating the workspace base briefly showed the incoming target commits inside the work-in-progress stack, marked as integrated, before the view snapped to the correct state.

`integrate_upstream_with_hints` hands back a `project_meta` pointing at the target's new tip, but the rebase keeps re-projecting from the workspace graph's own `project_meta`, which still holds the old target. `materialize` and the dry-run overlay both read from there, so the workspace state returned by the mutation sits on the old base with the just-integrated target commits folded into the stacks. Lite renders that response directly, and only the watcher-triggered refetch of head info reads the persisted meta.

The graph's `project_meta` is now moved to the target's tip before the editor is created, so the returned workspace, and the dry-run preview, are projected against the new base. The sibling test that already materializes this scenario asserts the resulting workspace.

Lite additionally refetches the target line when the update succeeds, so the base section does not keep showing the old incoming commits until the watcher delivers.


krlvi:
Superseded: re-opened against master from an independent branch.

PR #15830, opened 2026-09-08
Lite: drop the unread dots from the pull request tab and branch row

PavelLaptev:
Removes the teal dot from the Pull Request tab and from the PR chip in the branch row. The notification bell stays.

## Motivation

**It is not clear what the dot means.** There is no label and no hint. A new user sees a colored dot and has to guess whether it means new comments, a failed check, or something else.

**Clicking it leads nowhere.** A dot promises that something new is waiting behind it. So when you click the tab or the row, you expect to be taken to that thing, or at least to see it highlighted. Right now nothing happens. The tab opens as usual and you have to scan the whole page to find what changed. Showing the dot without that second half is doing half the job, and half a feature is worse than none.

**It is flaky.** The dot lights up for your own pushes and comments, because it only compares the PR's last-modified time against a timestamp saved on this machine. It goes away only after the tab has been open and focused for a second, so a quick look leaves it on. And two computers disagree about what is new.

Screenshot 2026-09-09 at 01-07-27

On the video below I can make the dot disappear.

https://github.com/user-attachments/assets/f1a9a914-be31-4a9d-bac9-a3bf4c7a6613

**In the branch row it competes with the CI tick.** There may be room for an unread marker in the row, but a green dot is not it. The tick is green and the dot is green, so they look like one thing and neither reads clearly. And the row can already hold several buttons, so one more indicator squeezed in next to them is not a good idea without a design that decides what belongs there.

Screenshot 2026-09-09 at 01-07-14

The bell still lists the same activity, so nothing is lost. If we bring unread markers back, they should come with a proper design: a clear meaning, a look that does not clash with the checks, and a click that takes you to what is new.

## Changes

- The Diff / Pull Request toggle and both of its call sites drop the unread flag, the dot, and the subscriptions and extra review query that fed it.
- The branch row drops the dot, its hover title, and the now-unused CSS classes.
- Seen tracking stays, since the bell rests on it.
- With the dots gone the "Quiet" level shows nothing, so the settings hint and code comments say so. Folding Quiet into Off changes stored settings and is left for a separate decision.
- The harness test checks the inbox directly instead of the dot; the e2e sidebar test drops its dot assertion.

### DIFF
--- a/apps/lite/e2e/tests/sidebar.spec.ts
+++ b/apps/lite/e2e/tests/sidebar.spec.ts
@@ -93,11 +93,6 @@ test("keeps unread PR activity off the Workspace tab", async ({ appWindow, elect
 	await appWindow.reload();
 
 	await expect(appWindow.getByRole("button", { name: "Notifications, 1 unread" })).toBeVisible();
-	await expect(
-		appWindow
-			.getByRole("treeitem", { name: "C", exact: true })
-			.getByTitle("New activity on this pull request"),
-	).toBeVisible();
 	const pages = appWindow.getByRole("group", { name: "Pages" });
 	await expect(pages.getByRole("button", { name: "Workspace", exact: true })).toHaveText(
 		"Workspace",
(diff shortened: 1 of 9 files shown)


PavelLaptev:
cc @mtsgrd @krlvi

PR #15828, opened 2026-09-08
Show linked worktrees in the Lite sidebar graph

mtsgrd:
Linked git worktrees appear in the workspace sidebar as lanes, so work happening in another checkout of the repo is visible from the main one. Gated on the `worktreeManipulation` feature flag, like the CLI.

## What it does

- A worktree resting on a commit in an applied stack draws as a lane inside that card, above the commit it rests on; one resting below the workspace draws as its own card under the stacks. Same placement rule as `but status`.
- Inside a lane: the worktree name, then `Uncommitted files` with the usual badge and line totals, the checked-out branch, and the commits only that worktree has.
- Rows are the sidebar's own: files and commits select and show their diff, the branch shows its details, arrow keys walk them, and a worktree file is addressable in the URL as `worktree-file:<name>:<path>`.
- Worktree files commit and amend into the workspace by drag or keyboard cut; the change leaves the worktree checkout. The branch row takes commits the way the CLI commits and moves onto a worktree.
- Menus: a worktree file opens, reveals, and copies its path against its own checkout; the branch row copies the branch name or worktree path, opens a terminal or the file manager there, and archives or removes the worktree. A commit inside a worktree keeps copy and open-in-browser only.
- Settings: the flag has a switch under Experimental, and a project page `Worktrees` lists active and archived worktrees with archive, unarchive, and remove.

## Backend

- `tree_change_diffs_from_source` diffs a change against the checkout its `ChangesSource` names; `tree_change_diffs` reads the main checkout, which shows the wrong file for a worktree's uncommitted change. SDK regenerated.
- `updateFeatureFlags` is wired through the Electron bridge like the other app settings.

## Notes

- The `UncommittedChanges` address gained an optional worktree name; file addresses inherit it, so identity keys and the changes source follow the parent. The main card is unchanged.
- The applied tree's row wrappers moved to `WorkspaceLists/TreeItem.tsx` so lanes render into the same wrappers as the cards.
- Not in this PR: discarding or absorbing a worktree's changes (no backend), checking worktree files, and a worktree on a stack of branches, which shows one branch like the TUI does.

## Verified

- `pnpm -F @gitbutler/lite check`, the Lite unit tests, and `knip` pass; `cargo test -p but-workspace --test workspace worktrees` passes.
- Live against a plain `git worktree add` of this repo: lanes, diffs, keyboard walking, and a cut-and-commit of a worktree file onto a workspace branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #15840, opened 2026-09-09
Show a linked worktree's stacked branches in Lite

mtsgrd:
Follow-up to #15828. A worktree's commits were listed flat under the branch it has checked out, so a commit sitting under a lower branch in the same checkout showed up under the wrong name and the lower branch was not shown at all.

- `WorktreeInfo::segments` splits a worktree's commits at each local branch met on the first-parent walk down from its `HEAD`. The first segment is the checked-out branch; each one below is a branch stacked underneath. Same rule as the stack projection: remote-tracking refs and tags never start a segment.
- `WorktreeInfo::commits()` is the flat view; the CLI and the ID map keep using it, so `but status` output is unchanged.
- Lite's worktree card renders a branch row per segment, selectable with the worktree branch menu. Navigation and the graph layout read commits through the segments.
- The worktree fixture gains a checkout with two branches; the projection test asserts both segments by name.

Not in this PR: a per-branch heading in the `but status` worktree lane, which needs a CLI id for the row.

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/Graph/layout.test.ts
+++ b/apps/lite/ui/src/routes/project/$id/workspace/Graph/layout.test.ts
@@ -63,7 +63,7 @@ const worktree = (name: string, base: Worktree["base"], commits: Array<string>):
 	refName: null,
 	head: commits[0] ?? "",
 	base,
-	commits: commits.map(ownCommit),
+	segments: [{ refName: null, commits: commits.map(ownCommit) }],
 });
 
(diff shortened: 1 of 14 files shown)


mtsgrd:
### Screenshots

Captured ad hoc from the dev app against a scratch worktree of this repository, with `worktree-scratch` checked out on top of `worktree-scratch-lower`. After only. The worktree card is not in the screenshot catalogue, so this surface is not covered going forward.

**worktree-card**

worktree-card

PR #15855, opened 2026-09-09
Count the incoming commits on the upstream card header in Lite

mtsgrd:
The target's card header named the branch and nothing else, so how much had landed upstream only showed once the card was unfolded, or on the Base row's "N behind" further down.

- The section header gains a `badge` slot beside its label, in the position the uncommitted header uses for its file count.
- The target's card fills it with the plan's incoming count, in the same lightGray badge, with "N new commits" as the spoken label.

No CSS changed.


mtsgrd:
### Screenshots

Captured ad hoc from the dev app against this repository's own target, which had one new commit at the time. After only. The upstream card is not in the screenshot catalogue, so this surface is not covered going forward.

**upstream-header**

upstream-header

PR #15858, opened 2026-09-09
Make the hand cursor an appearance setting in Lite

mtsgrd:
Lite keeps the arrow over its controls, as desktop apps do. The harness panel showed the hand on links only because the app's `global.css` is not loaded there, and a dozen component modules still set their own `cursor: pointer`.

- `handCursor` GUI setting, off by default, with a "Hand cursor" switch on the Appearance page
- one `--control-cursor` property per host: the app sets it from the setting through a root attribute, the harness panel pins the hand since it sits in a web page
- `control-cursor.css`, imported right after design-core by both hosts, puts it on buttons, links, `summary` and the button, menu-item and option roles
- component modules stop setting `cursor: pointer` or pinning `cursor: default` on controls, so the setting reaches them; disabled controls keep `not-allowed`, rows keep the arrow
- DESIGN.md's Cursors section describes the property and the setting

Behaviour change: buttons that showed the hand before (the row fold toggle, the "add description" link in the PR form) now follow the setting and default to the arrow.

Verified in the running dev app over CDP: flipping the switch moves every button and link between `default` and `pointer`, and the value persists across reloads.

### DIFF
--- a/apps/lite/DESIGN.md
+++ b/apps/lite/DESIGN.md
@@ -58,11 +58,18 @@ selection can restyle without a re-render.
 
 **No pointer cursors.** Lite is a desktop app, and desktop apps keep the arrow
 over buttons, menus and rows; the hand is a web convention for links out to a
-page. Don't set `cursor: pointer` on a control, and don't reintroduce it by
-resetting a `<button>` — the browser default for buttons is already the arrow.
-Interactivity is shown by the hover state, not the cursor. The cursors that do
-change are the ones that describe a gesture: `text` over editable text, `grab`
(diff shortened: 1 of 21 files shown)


mtsgrd:
## 📸 UI screenshots

Captured 1 surfaces, 1 changed.

### settings-appearance

| Before | After |
| --- | --- |
|  |  |


The Appearance page was captured ad hoc for this pull request: it is not in the screenshot catalogue, so later changes to it are not covered.

krlvi (approved):
neat

PR #15864, opened 2026-09-09
Keep the PR meta line inline beside the author

PavelLaptev:
The reviewer picker change in #15846 gave the shared user row `flex: 1` so a long login truncates before the new remove button. The PR description's meta line reuses that same row, so the author grew to fill the whole line and pushed the commit count and updated time to the far edge.

A1CF3F4F-2840-4EDA-A7B9-2956D30D2E02


PavelLaptev:
cc @mtsgrd

PR #15866, opened 2026-09-09
Lite: move the upstream count and Integrate onto the target's row

krlvi:
In the Lite workspace sidebar, the Base row showed "n behind" and an Update button, while the commits they refer to are folded under the origin/master row above it. Seeing what is new and bringing it in were two rows apart.

Now the target's row (origin/master) carries the count as "n new" and an **Integrate** button, whose tooltip reads "Integrate the latest from origin/master into the base". The Base row is a plain fold toggle again; its docked stand-in at the panel foot keeps only the fold.

Same mutation as before, relocated and relabelled. This is the minimal tweak agreed on in Discord, not the larger reframing (leading with "upcoming changes" and demoting the ref name), which still needs design work.

Before / after are in a comment below.

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/Graph/Section.module.css
+++ b/apps/lite/ui/src/routes/project/$id/workspace/Graph/Section.module.css
@@ -67,7 +67,7 @@
 }
 
 /* Folded, the merge base row's stand-in docks at the scroller's foot while
-   the row is out of view below, keeping the count and Update in view; its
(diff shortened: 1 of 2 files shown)


krlvi:
## 📸 UI screenshots

Captured ad hoc against `project-with-remote-branches.sh` with two commits added to the remote's master and fetched. This state is not in the screenshot catalogue, so it is not covered going forward.

### target-moved-on

| Before | After |
| --- | --- |
|  |  |

### Hovering Integrate

PR #15868, opened 2026-09-09
Show each branch of a linked worktree in the Lite sidebar

mtsgrd:
Follows #15865, which projects a linked worktree's history into stack segments. Lite still flattened those into one branch row over every commit, so a branch stacked under the checked-out one was not shown and its commits sat under the wrong name.

- The worktree lane renders a branch row per segment, with the commits under the branch that owns them. A detached `HEAD`'s anonymous first segment has no branch row.
- Keyboard navigation lists the rows in the same order.

Not in this PR: the segments' push status and remote-only commits, which the lane does not surface yet.


mtsgrd:
### Screenshots

Captured ad hoc from the dev app against a scratch worktree of this repository, with `worktree-scratch` checked out on top of `worktree-scratch-lower`. After only. The worktree card is not in the screenshot catalogue, so this surface is not covered going forward.

**worktree-card**

worktree-card

PR #15874, opened 2026-09-09
Warn before workspace updates cause conflicts

mtsgrd:
- Show a warning triangle before the target ref when pulling would cause conflicts.
- List affected branches, commits, and files in a popup.
- Show no preview indicator during checks or after failures. Keep `Pull latest` available.
- Combine rapid changes and run one check at a time. Skip stale checks that haven't started.
- Share the preview and pull state between the target row and its docked copy.
- Preview conflicts without changing commits or working files.
- Return conflicted files for both previews and applied updates.

Local checks passed:

```text
Rust core/rebase tests + API regression in both SDK modes
31 focused UI tests + 31 Electron tests (9 skipped)
Full Lite tests and typecheck
Lint, format, Knip, and React Compiler checks
```

Fixes GB-2029.

### DIFF
--- a/apps/desktop/src/lib/stacks/stackInvalidation.test.ts
+++ b/apps/desktop/src/lib/stacks/stackInvalidation.test.ts
@@ -33,6 +33,7 @@ const headInfo = {
 const integrationResult = {
 	workspaceState: { headInfo, replacedCommits: {}, checkoutConflictOccurred: false },
 	worktreeConflicts: [],
+	commitConflicts: {},
 	targetCommits: null,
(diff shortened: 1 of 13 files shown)


mtsgrd:
- One-off captures of the running Lite app after updating to the new target row.
- The warning triangle sits to the left of the ref name.
- Only confirmed conflicts show a warning. The target row has no preview indicator while checking or after failure.
- `Pull latest` stays available.

| Before | Warning |
| --- | --- |
| Target row before this change | Warning triangle to the left of origin/master |

**Conflict details**

Popup lists the affected branch, commits, and files

PR #15857, opened 2026-09-09
Next nightly download links

samhh:
Closes GB-2019. Anyone feel free to change the copy/visuals or merge; we tentatively plan to merge this on Friday.

https://gitbutler-j6c7a0x00-gitbutler.vercel.app/nightly

PR #15876, opened 2026-09-09
Link each Next nightly to its commit on the website

PavelLaptev:
The Next Nightly block on /nightly shows the build's version and release date, but unlike the stable hero above it there is no link to the commit the build came from, because the Lite updater manifests do not carry one.

Screenshot 2026-09-09 at 23-01-08

This wires that up end to end:

- The Lite workflow passes the run's commit to electron-builder as `releaseInfo.releaseNotes`, which electron-builder writes into every updater manifest as a `releaseNotes` field. The step also gets `shell: bash` so the env var expands on the Windows runner.
- The nightly page's server load reads that field back, accepting it only when it is a full 40-character sha.
- The Next meta line ends with the short sha linked to the commit on GitHub, using the same link styling as the hero with a white override for the blue background.

The link stays hidden until a nightly built after this change has published a manifest carrying the sha, so nothing changes on the page until the next Lite nightly run.

Stacked on #15857, which restyles the section this builds on.

### DIFF
--- a/.github/workflows/lite.yml
+++ b/.github/workflows/lite.yml
@@ -88,7 +88,11 @@ jobs:
           APPLE_TEAM_ID: ${{ runner.os == 'macOS' && secrets.APPLE_PROVIDER_SHORT_NAME || '' }}
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-        run: pnpm exec electron-builder --publish always
+          RELEASE_SHA: ${{ github.sha }}
+        shell: bash
+        # The commit lands in the updater manifest under `vendor.sha`, where the
+        # website's nightly page reads it back to link each build to its commit.
+        run: pnpm exec electron-builder --publish always --config.releaseInfo.vendor.sha="$RELEASE_SHA"
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         name: Upload artifacts
         with:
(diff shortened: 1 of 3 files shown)


PavelLaptev:
@samhh I also added this to include the build SHA, similar to what we have for the nightly block. It's not strictly necessary, though.

.github/workflows/lite.yml
```diff
@@ -88,7 +88,11 @@ jobs:
           APPLE_TEAM_ID: ${{ runner.os == 'macOS' && secrets.APPLE_PROVIDER_SHORT_NAME || '' }}
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-        run: pnpm exec electron-builder --publish always
+          RELEASE_SHA: ${{ github.sha }}
+        shell: bash
+        # The commit lands in the updater manifest as `releaseNotes`, where the
+        # website's nightly page reads it back to link each build to its commit.
+        run: pnpm exec electron-builder --publish always --config.releaseInfo.releaseNotes="$RELEASE_SHA"
```
samhh: We might want to use `releaseNotes` in the future. `vendor.sha` or similar might be better?

samhh (approved):

.github/workflows/lite.yml
PavelLaptev: Good call. Switched to `releaseInfo.vendor.sha`, which electron-builder spreads into the manifest as a nested `vendor: { sha }` key, and the nightly page reads it back from there. `releaseNotes` stays free.

slarse (approved):

PR #15899, opened 2026-09-10
feat(but): show tree status in `but diff` output

slarse:
This changes the hunk header rendering to mimic the status output, showing the tree status as A, D, R or M. That information was not generally available during rendering, so I've propagated the minimal `TreeStatusKind` enum to every struct that needs it. Copying it around does not seem like a big deal.

In addition to that, I've made some necessary changes to the hunk extraction for tree changes. Previously we'd just put all the hunks in one big list (`but_core::hunks_from_changes`), but that lost the information about which hunk came from which change. The new `but_core::changes_with_hunks` retains this association and it's now propagated up the call stack, all the way to the `IdMap` and into the `CliID`s.

### Before
current

### After
colorize_change_char

This also goes hand in hand with the new look for `but status` that was introduced in https://github.com/gitbutlerapp/gitbutler/pull/15892 - they look "the same".

For more examples, just look at any of the hundreds of changed TUI snapshots :)


crates/but/src/utils/diff_rendering.rs
```diff
@@ -1160,38 +1185,43 @@ fn render_signature(
...
+
+    if let Some(status) = status {
+        path_line.push(status_letter_kind(status, theme));
+        path_line.push(Span::raw(" "));
+    }
+
+    path_line.extend([
+        Span::raw(path.to_string()),
+        Span::raw(" "),
+        Span::styled("|", theme.border),
```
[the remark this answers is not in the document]
slarse: Ermagerd ...

krlvi:
@slarse did you consider putting the status after the file name?

slarse:
> @slarse did you consider putting the status after the file name?

@krlvi Yes, a fair deal actually. But I decided against it mostly because I want the layout to be consistent with `but status`, and there I'd find it pretty strange to have the status after the filename. Here's what it currently looks like:
new_status

There's also the fact that the status is a tiny, fixed-width symbol that displaces the filepath by precisely two characters (the status + a space), while the filepath is variable width and can potentially displace the status very far.

The thing that speaks _for_ putting the status last is that, as mentioned in the [corresponding `but status` PR](https://github.com/gitbutlerapp/gitbutler/pull/15892), the status is probably the least important piece of information we're displaying here.

But from the perspectives of layouting and consistency, I found it best to have the status before the filename.

Note that I also discarded the idea of putting status information on a separate line altogether as that IMO bloats the output disproportionately to how important that piece of information is.

crates/but-core/src/hunks.rs
slarse, line 1: @Caleb-T-Owens 👀

crates/but/src/utils/change_source.rs
slarse, line 1: @Caleb-T-Owens and also 👀

crates/but/src/id/mod.rs
```diff
@@ -872,8 +872,20 @@ impl IdMap {
...
+                            .into_iter()
+                            .map(|hunk| (UnqualifiedHunkId::default(), hunk)),
+                    )
+                    // Note: This expect holds because but_core::SingleHunk::from_tree_change is
+                    // guaranteed to produce at least one hunk per change, and that's what's used in
+                    // but_core::changes_with_hunks. We could get rid of this expect by encoding
+                    // that in a NonEmpty return value from but_core::SingleHunk::from_tree_change
+                    // and propagate that up the stack, but that requires adding the nonempty crate
+                    // as a dependency in but_core. Decided "not worth it" for now.
+                    .expect("Each tree change produces at least one hunk"),
```
slarse: @Caleb-T-Owens And also 👀 , thoughts?

Caleb-T-Owens: Is this true - what if I touch an empty file? Does that contain one empty hunk?

It would be nice to encode in `but_core` if it's actually true

slarse: > Is this true - what if I touch an empty file? Does that contain one empty hunk?

Yes, that is how `but_core::SingleHunk::from_tree_change` is written.

https://github.com/gitbutlerapp/gitbutler/blob/c5394f0b25ba3b68231f2d9caee73f75936409e9/crates/but-core/src/hunks.rs#L55-L61

But as you also think we should push it down, let's do that.

slarse: Contract pushed all the way down with https://github.com/gitbutlerapp/gitbutler/pull/15899/changes/6fc63655518b39cac0a2e3a799078d01b5409fbf

Caleb-T-Owens (approved):
neat

PR #15914, opened 2026-09-11
Fix Lite stalls from refetching unchanged blob diffs

krlvi:
Fixes [GB-2046](https://linear.app/gitbutler/issue/GB-2046/next-nightly-stalls-and-saturates-cpu-when-file-events-repeatedly).

File watcher events invalidated every diff query, including an open committed diff whose content had not changed. For a large commit, each refresh repeated native diff work and synchronous hashing of the full changes array inside the streamed query's cache lookups. Repeated unrelated file writes caused CPU saturation and roughly 300 ms renderer stalls.

### Reproduction and impact

The isolated macOS fixture attached to the issue reproduces this on Next Nightly 0.0.200 and the source build before this fix:

1. Create a repository with 5,000 tracked text files plus `.gitignore`, and 23 stable untracked files.
2. Keep the 5,001-file baseline commit selected in the diff view.
3. Repeatedly create and delete an unrelated `fixture-noise.lock` file at 750 ms intervals for 25 seconds (33 writes).

With the same fixture and workload, comparing the source build before and after:

| Observation during file churn | Before | After |
| --- | ---: | ---: |
| Renderer response time, p95 | 273 ms | 5 ms |
| Renderer long tasks | 31 | 0 |
| Main-process CPU time over 25 seconds | 112.43 s | 34.92 s |

CPU time accumulates across threads. These are local reproduction results; the remaining work includes normal watcher processing.

### Fix

Mark diff queries according to whether they read from the worktree, using the same null-destination-object-ID rule as the Rust diff engine. Preserve the watcher's changed paths through N-API so ordinary file events can skip blob-only diffs.

Index events still refresh broadly: they carry no changed paths, and staged attributes can affect blob diffs. Missing paths or query metadata also keep the existing broad invalidation behavior. Live and mixed worktree diffs continue refreshing, including edits that leave their change descriptors unchanged; editing and restoring a file in the reproduction app updated the displayed diff in both directions.

This keeps the change confined to watcher payload forwarding and Lite query invalidation. It adds no debounce or scheduling mechanism and leaves live-worktree invalidation broad.

### DIFF
--- a/apps/lite/harness/tests/panel.test.tsx
+++ b/apps/lite/harness/tests/panel.test.tsx
@@ -152,7 +152,10 @@ test("a watcher event refreshes the uncommitted files", async () => {
 	worktree = fixtureWorktreeChanges([fixtureFileChange("src/new-file.ts")]);
 	const event: WatcherEvent = {
 		name: "worktreeChanges",
-		payload: { type: "worktreeChanges", subject: { changes: worktree } },
+		payload: {
+			type: "worktreeChanges",
+			subject: { changes: worktree, changedPaths: ["src/new-file.ts"] },
+		},
 	};
 	panel.push(eventChannel, event);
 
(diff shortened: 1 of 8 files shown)


krlvi:
FYI @samhh

PR #15916, opened 2026-09-11
Add a saved History display menu in Lite

mtsgrd:
## 🧢 Changes

Add a kebab to the sidebar's History section to choose its default view: the last five commits or the last 12 hours. Save the preference, load enough history for the time window, and keep Show more available in both modes.

## ☕️ Reasoning

The time window uses commit time and keeps intervening commits in graph order. Switching modes resets pagination, including when an earlier request finishes afterwards.

Validation: Lite typecheck, lint, formatting, both Knip checks, 553 unit tests, and 33 end-to-end tests passed (9 skipped). After extracting the menu component, typecheck and all 67 graph tests passed; React Compiler memoization was checked.

### DIFF
--- a/apps/lite/electron/src/settings.ts
+++ b/apps/lite/electron/src/settings.ts
@@ -29,6 +29,7 @@ const guiSettingsV1 = type({
 	"fileDisplayMode?": "'list' | 'tree'",
 	"filesPanelRight?": "boolean",
 	"handCursor?": "boolean",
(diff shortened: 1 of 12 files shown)


mtsgrd:
## 📸 History menu

Before and after, using the same seeded workspace and viewport. All nine catalogue captures passed on each side; this comparison shows the changed History header. Relative timestamps differ between runs. The native OS menu popup is outside the screenshot harness's scope.

| Before | After |
| --- | --- |
|  |  |

PR #15918, opened 2026-09-11
Stop worktree history at its shared base

mtsgrd:
Worktrees branching below the target could lose their base and display the repository's entire history. Keep the target traversal alive while a disconnected tip still needs its ancestry, preserving existing cutoffs for tips already in workspace or integrated history.

Adds a regression fixture with an older worktree fork and a small traversal limit. Validated with 226 graph, workspace, and CLI status tests, plus the running Lite app: 13,272 worktree commits reduced to 2.

### DIFF
--- a/crates/but-graph/src/init/types.rs
+++ b/crates/but-graph/src/init/types.rs
@@ -81,12 +81,17 @@ impl Limit {
         {
             return false;
         }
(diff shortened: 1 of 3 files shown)


Caleb-T-Owens (approved):
Seems reasonable

PR #15917, opened 2026-09-11
Keep commit controls in a resizable Workspace pane

mtsgrd:
## 🧢 Changes

Starting a commit splits the Workspace into a resizable upper pane for uncommitted files and the composer, and an independently scrolling lower pane for stacks. The initial split keeps the commit button in place; canceling preserves the visible file position and draft. The collapsed commit button remains sticky and slides beneath the Uncommitted files header.

## ☕️ Reasoning

Keep files and commit controls accessible while browsing stacks, with minimal movement when entering or leaving commit mode. Reuse the existing panel and resize-handle components.

Validation: Lite typecheck, lint, formatting, both Knip checks, focused commit-form, stack scrolling, and keyboard-focus E2E tests.

PR #15919, opened 2026-09-11
Keep collapsed upstream chevrons clear of the mainline

mtsgrd:
## 🧢 Changes

Show incoming upstream commits as a branch even when collapsed. Keep the grey mainline continuous beside the chevron and use the upstream yellow for the full connector back to it.

Adjust the collapsed chevron's spacing while keeping the label stationary and the expanded chevron aligned with the commit line.

## ☕️ Reasoning

Make the collapsed branch relationship clear without interrupting the mainline or shifting the expanded card's alignment.

Validation: verified collapsed and expanded states over CDP; Lite typecheck, lint, formatting, both Knip checks, and Graph Section tests passed before this unchanged commit was split out of #15917.

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/Graph/Section.module.css
+++ b/apps/lite/ui/src/routes/project/$id/workspace/Graph/Section.module.css
@@ -113,6 +113,14 @@
 	--row-label-offset: calc(-3px + var(--row-fold-chevron-offset, 0px));
 }
 
+.collapsedBranchHeader {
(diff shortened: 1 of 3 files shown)


mtsgrd:
## 📸 Upstream collapsed

Before (`8861f86241`) and after (`040ed50dd2`), captured ad hoc from the same seeded fixture before this unchanged graph fix was split into its own PR. This state is not in the permanent screenshot catalogue.

| Before | After |
| --- | --- |
|  |  |

PR #15926, opened 2026-09-11
Keep stack order stable on refresh

mtsgrd:
Stacks could change order on refresh for two unrelated reasons. One commit each, and each has a test that failed before its fix.

### 1. Moving a graph connection put its stack at the front

- Stack order is the order of the workspace's outgoing connections; petgraph visits the newest first.
- `move_incoming_edges` re-added the one connection it moved, so that stack jumped to the front. It runs after stacks are ordered.
- It now rebuilds all of the source's outgoing connections in order. That covers all three callers: the two worktree paths, and the entrypoint path that needs no worktree.

```text
Saved:            A, B   (worktree on B)
Refresh, before:  B, A
Refresh, after:   A, B
```

### 2. `but commit` saved the stack order after releasing its lock

- The TOML store writes on drop. `but commit` re-binds its lock guard after `meta`, so the lock goes first and the TOML lands a moment later. The other seven transaction commands drop in the safe order.
- The app's watcher waits on that lock and re-reads the moment it is free: new refs, old order. The TOML write then triggers a second re-read.
- `but commit -b <new>` asks for position 0, so the new stack showed up last and then jumped to the front.
- `RefMetadata::flush` (no-op by default, the TOML store overrides it) now runs inside the transaction, once the refs have moved. Drop order no longer matters.
- A failed flush is logged, not returned: the refs already moved, and drop would only have logged too.
- Swapping two `let`s in `commit.rs` would fix today's case. Flush also covers `with_transaction`, which always releases its own lock before the caller's handle drops.

```text
Requested:  A, N, B
Before:     A, B, N, then A, N, B on drop
After:      A, N, B
```

### Not covered

- The second fix fits the jump I see in Lite while agents commit from the CLI, but I haven't measured it in a running app.
- A read that doesn't wait for the lock can still land between the refs moving and the flush.

### Verified

```sh
cargo test -p but-graph worktree_ref_as_later_stack_top_preserves_stack_order
cargo test -p but-transaction creating_stack_persists_order_before_metadata_is_dropped
cargo test -p but-meta flushed_metadata_remains_writable_until_drop

# The other workspace response format.
cargo test -p but-transaction --features but-api/graph-workspace
```


mtsgrd:
@Caleb-T-Owens I think this is what I've brought up in the past, would be great if you could have a look?

mtsgrd:
repro unclear, will re-open if it resurfaces.

PR #15929, opened 2026-09-11
Hide the duplicate sidebar fetch button

mtsgrd:
Keep row toolbars hidden with their parent. Adds coverage for docked fetch visibility; 40 E2E tests pass.


mtsgrd:
Ad hoc Electron captures. Before temporarily restores the old toolbar visibility declaration in the fixture; after uses the fix. The temporary capture spec was removed.

| Before | After |
| --- | --- |
|  |  |

PR #15931, opened 2026-09-11
Keep diverged branch actions inside narrow rows

mtsgrd:
When a diverged branch has a PR and CI checks, Integrate overflows a narrow sidebar even with the button labels hidden. Allow rows containing Push and Integrate to wrap when crowded.

Verified in Electron at 260–600px with folded and expanded branches. Before-and-after screenshots are attached in the comment below.


mtsgrd:
### Diverged branch at a 260px sidebar width

Before the fix, Integrate extends past the branch row. After, the metadata wraps and both actions fit.

| Before | After |
| --- | --- |
| Integrate overflows the branch row | Metadata wraps and both branch actions fit |

Captured directly from the running Electron app using the diverged-branch fixture and mocked PR/CI data. This state was captured ad hoc, outside the shared screenshot catalogue.

PR #15933, opened 2026-09-11
Show expandable pull request checks in Lite

mtsgrd:
Put Checks directly below PR status and let its chevron reveal every job, with status, a compact live elapsed time for running jobs, and a fixed duration for completed jobs. Keep failures first in the indicator and job list, and cap the expanded list with an edge-to-edge scroll area.

Only Checks and Reviewers collapse. Their chevrons use the shared icon button style, align with the section inset, and disappear when empty; section titles remain plain text.

Validation: Lite typecheck, 570 unit/harness tests, 40 E2E tests (9 skipped), lint, both Knip checks, and formatting passed. Focused browser checks covered live timing, completion updates, all statuses, keyboard controls and scrolling, geometry, and empty states in both PR panels. React Compiler output was inspected for stable query selection and row memoization.

### DIFF
--- a/apps/lite/ui/src/routes/project/$id/workspace/PullRequestPanel.module.css
+++ b/apps/lite/ui/src/routes/project/$id/workspace/PullRequestPanel.module.css
@@ -38,32 +38,25 @@
 }
 
 .heading {
+	display: flex;
+	align-items: center;
 	color: var(--text-2);
 }
 
-/* Only the stacked layout folds the card; beside the column it has the
-   height to show everything. */
-.panelFooter {
-	display: none;
-	padding-inline: 14px;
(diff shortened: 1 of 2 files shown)


mtsgrd:
## 📸 UI screenshots

One-off captures of two PR details states in the running Lite app, using the same seeded PR/check data and viewport. These states are not in the shared screenshot catalogue; no permanent coverage was added.

macOS scrollbars were set to **Always** for both runs, then restored to the original setting. The original view showed only failed jobs, so both before images show that view.

### Checks collapsed

| Before | After |
| --- | --- |
|  |  |

### Checks expanded

| Before | After |
| --- | --- |
|  |  |

PR #15940, opened 2026-09-12
feat(lite): add an operations log button to the sidebar header

PavelLaptev:
The operations log has always been there behind ⇧⌘O and the command palette, but with no visible way in. This gives it a button, plus the `history` icon it needed.

- New `history` icon from ⚛️ Lite Core — a clock face broken at the upper left, which reads as time you can step back through rather than time passing.
- `clock` gets the same hands so the two read as a pair; its minute hand also moves back onto the 16px grid the set is drawn to. Its only consumer is the "Awaiting review" state in `PullRequestPanel`.
- The button goes in the header's action cluster, left of settings — restoring a snapshot rewrites the whole project, so it belongs with the other project-wide dialogs rather than on a row or in a context menu.
- It mirrors the settings button beside it, and takes its label and keycaps from the hotkey's own meta, so the tooltip, the `aria-label` and the command palette entry stay one string. Gated on `noOperationPending`, matching the hotkey.

Verified against the running dev app after a full reload: the button opens the picker, ⇧⌘O still works, Escape closes. `check`, the 46 unit tests, oxlint, knip and both formatters are clean.

Undo and redo are still keyboard-only, and "Operations log" is still the wording in all three places — both left alone deliberately, since either is a change of its own.


PavelLaptev:
## 📸 UI screenshots

The new button in the sidebar header, left of the settings gear:

| Before | After |
| --- | --- |
|  |  |

It sits in the header, so it also shows on the branches tab and with the commit form open — same button, not worth repeating here.

Captured locally on macOS, not in CI.

PR #15941, opened 2026-09-12
feat(but): add an experimental `but panel` workspace view

PavelLaptev:
A proof of concept: a narrow, read-only workspace view on localhost, for keeping the relevant project's workspace in a side pane next to an agent chat instead of switching to the client.

- `but panel [--port N] [--no-open]` serves a compiled-in page on 127.0.0.1, prints its URL (also with `--json`) and opens the browser; Ctrl-C stops it
- One server shows any project: the page names it with `?project=<path>`, and running `but panel` in another project reuses the running server and points at that project's page; the project name in the header switches between your GitButler projects (the full list needs `legacy`)
- Branches with their PR and CI, commits, their files and diffs, uncommitted changes, and linked worktrees, refreshing every 3s
- A search box finds files across workspace commits, uncommitted and worktree changes; right-clicking a file opens it with the programs `but open` suggests
- Opening is the only action and the only `POST`, accepted only with the panel's own origin, so another page in the browser can't launch programs
- Built without `legacy`: data comes from `get_workspace`, `changes_in_worktree_with_perm`, `worktrees_list_with_perm`, `commit_details_with_line_stats`, `tree_change_diffs` and `tree_change_diffs_from_source`
- Except PRs and CI, which only `but_api::legacy::forge` exposes today; that lookup is gated behind `legacy` and reads the forge cache only, so CI can be missing until something refreshes it
- std-only server, GET only, rejects foreign `Host` headers (DNS rebinding)
- Hidden while the shape settles, so `crates/but/skill` doesn't mention it yet; worktrees need the `worktreeManipulation` flag, and archived ones are listed without their changes

## How to try

```sh
gh pr checkout 15941
cargo run -p but -- -C /path/to/a/gitbutler/project panel
```

Your browser opens that project's page on `localhost:7789`; run it again from another project to see that one too. Afterwards run your release `but` once (e.g. `but status`): a dev `but` rewrites your installed GitButler agent skill to `dev`, and the release build restores it.


PavelLaptev:
@krlvi I haven't worked on the design at all yet. I just wanted to check whether this makes sense. The current setup is inconvenient because `panel` isn't part of a release yet, but I imagine it should end up as simple as `but panel`.

If you give it a green light, I'll work on the UI. Another option is a separate `but-panel` repo that combines `but status` with git commands to build the picture. From what I understand, though, wiring it into the Rust code directly is the better approach.

krlvi:
this looks really really cool! thanks for prototyping this! you should demo it during summit and lets talk about it!

PavelLaptev:
> this looks really really cool! thanks for prototyping this! you should demo it during summit and lets talk about it!

Nice. I'll go ahead and tweak the UI then.

PR #15949, opened 2026-09-14
Lite: file tree: give directory rows (most of) the features from file rows

cbjeukendrup:
* Context menu
    Scherm­afbeelding 2026-09-14 om 12 27 04

* 'Reviewed' check marks
    Scherm­afbeelding 2026-09-14 om 12 30 38

---
- <kbd>&nbsp;2&nbsp;</kbd> #15993
- <kbd>&nbsp;1&nbsp;</kbd> #15949 👈


PavelLaptev:
@cbjeukendrup thanks for the PR!
How folder "checkmarks" work?

cbjeukendrup:
A folder is marked 'reviewed' iff all files under it are reviewed:

https://github.com/user-attachments/assets/03349043-bfb4-410c-86f0-1be02adbba70

PavelLaptev:
>  Schermopname.2026-09-21.om.09.37.00.mov

Gotcha. I wonder if we should add the ability to check files and folders via right-click? Currently, it only works via the UI button, and reviewing a folder isn't possible yet.

cbjeukendrup:
I could look into that. 'Marking a folder (un)reviewed' means 'marking all files under it (un)reviewed', right? 

(Or do you think we should give folders their own reviewed state? I'm not sure that'd be a good idea, because the folder is in a way only a transient UI thing: if the file tree changes (files added/removed), the folder might be split, or disappear its parent.)

PavelLaptev:
> 'Marking a folder (un)reviewed' means 'marking all files under it (un)reviewed', right?

yes, a folder just represents its files. Technically I'm not sure how feasible it is, but I thought it would be consistent and nice for UX. @samhh wonder wdyt

cbjeukendrup:
How would this be?

https://github.com/user-attachments/assets/b83d08c5-e37e-4805-b12f-88e8dc40594a

PR #15976, opened 2026-09-16
feat(lite): fail check on CSS custom properties nothing defines

PavelLaptev:
A `var(--x)` nothing defines doesn't error; the property quietly falls back, which is how `--radius-m` shipped square corners. This catches that class of bug in `check` and fixes the ones it found.

- `scripts/check-tokens.mjs` gathers every defined name from design-core, the diff library and Lite's own CSS and TS, and fails `pnpm -F @gitbutler/lite check` on any `var(--x)` matching none, with the nearest real name as a hint
- `--radius-m` → `--radius-card` on the conflict editor, the image diff frame and the edit-mode file list
- `--font-mono`, `--clr-scale-succ-40` and `--diffs-background` → the tokens that exist, so mono text, the resolved hint and the gutter seam render as intended
- AGENTS.md says how a component is checked in Storybook: the iframe URL, the derived story id, the dark theme, computed styles over eyeballing


PavelLaptev:
that's about checking CSS and fixing broken values.

PR #15984, opened 2026-09-18
fix(tui): Fix ANSI escape code in "Discard commit" popup

cbjeukendrup:
Before: 
Scherm­afbeelding 2026-09-18 om 02 25 52

After:
Scherm­afbeelding 2026-09-18 om 02 25 57

(Generated using Claude)


slarse:
Thanks for the contribution @cbjeukendrup! I've assigned @davidpdrsn to review this as the TUI is his brainchild.

davidpdrsn (approved):
I've pushed some changes now that removes a bit of duplication.

Thanks!

PR #16000, opened 2026-09-21
ci: `CSS` label, screenshots only if visible, and UI review requests

PavelLaptev:
Two pieces of pull request automation for design work, both of which ask rather than insist: the screenshots reminder stops labelling every Lite CSS change, and changes to the UI libraries request a design-system review.

**Screenshots**

- Not every CSS change shows on screen, so the reminder stops labelling Lite CSS pull requests `screenshots needed`
- pr-labeler marks any pull request with CSS in it `CSS`
- The reminder applies no label; it comments once, in one paragraph, and asks for screenshots only if the change is visible
- `screenshots needed` stays as a label to apply by hand, and the comment words itself for that case
- The `lite-screenshots` skill adds `screenshots` and leaves `CSS` alone

**UI review requests**

- A pull request touching `packages/ui-react/**`, `packages/ui-svelte/**` or `apps/lite/DESIGN.md` requests a review from Pavel. It only asks: no merge is blocked and no check fails
- Not CODEOWNERS on purpose: under the `master protection` ruleset an owned path needs an owner's approval, and GitHub doesn't count an author's approval of their own pull request. Pavel writes most changes to these paths, so owning them would block his own
- Asks once: skipped if the review is already requested or given, on drafts, on Pavel's own pull requests, and on forks (read-only token)


slarse (approved):
This is a pretty pragmatic approach, wonder if we should generalize this to other systems. @davidpdrsn wdyt for `but/tui`?

PR #16001, opened 2026-09-21
refactor(lite): move the components into @gitbutler/ui-react

PavelLaptev:
Lite's components were a folder inside the app, so `but panel` and the MCP app each restyle the same workspace by hand. They move to `packages/ui-react`, a source-only package (no build; each consumer's Vite compiles it). No behaviour changes. Follows #15994, which cut the components' ties to the app.

- **The move (commit 1):** 572 files, 458 of them pure renames. 113 files plus the icon, illustration and logo folders, `base.css` and `time.ts`. In Lite the change is one import prefix: `#ui/components/X.tsx` → `@gitbutler/ui-react/X.tsx`.
- **Stays in Lite:** the Markdown components, mention suggestions, the picker dialog, `useAddLocalRepository`, and `useCopied` (it writes the clipboard through `window.lite`, so it only works in Electron).
- **Tooling (commit 2):** oxlint lints the package with Lite's config, re-exported so it can't drift; `check-tokens` reads the package's source too; the `lite_e2e` filter in `push.yaml` gains `packages/ui-react/**`, since a filter that matches nothing fails silently. pr-labeler labels a change to the package `@gitbutler/ui-react`, as it does the Svelte one.
- **Storybook stays in Lite** and lists the package's stories under the same `components/` titles, so story ids don't change. Lite's preview can't re-export the package's: Storybook reads the file's source for the `definePreview` call, and a re-export silently dropped the theme toolbar and dark mode. The package exports the shared settings and each preview calls `definePreview` with them.
- **Checked:** tests are the same 573 (545 in Lite, 28 in the package); React Compiler still compiles the moved files under Lite's Vite config; all 125 stories index and render in both themes.
- knip ignores `packages/*`, so the moved components are no longer checked for unused exports.


slarse (approved):
Only looked at the CI file changes.

PR #16002, opened 2026-09-21
ci: ask for a design-system review on UI library changes

PavelLaptev:
Requests a review from the design system's maintainer when a pull request touches the component libraries or Lite's design rules. It only asks: no merge is blocked and no check fails.

- Watches `packages/ui-react/**`, `packages/ui-svelte/**` and `apps/lite/DESIGN.md`.
- Not CODEOWNERS on purpose: under the `master protection` ruleset an owned path needs an owner's approval, and GitHub doesn't count an author's approval of their own pull request. Pavel writes most changes to these paths, so owning them would block his own.
- Asks once: skipped if the review is already requested or given, on drafts, on Pavel's own pull requests, and on forks (read-only token).
- Modelled on `lite-screenshots.yml`: same triggers, same pinned actions, `persist-credentials: false`, and `pull-requests: write` only on the job that needs it.


PavelLaptev:
Moved into #16000, which now carries this commit alongside the screenshots reminder change.

Issue #15870, opened 2026-09-09
CLI: Cannot forge auth on headless Linux machines via CLI


anaisbetts:
### Version

0.22.3

### Operating System

Linux

### Distribution Method

deb (Linux)

### Describe the issue

GitButler CLI seems to use Secret Storage on Linux which doesn't work on many headless CLI setups. Furthermore the error is pretty inscruitable:

Error: Authentication failed

Caused by:
    0: Failed to persist access token
    1: Platform secure storage failure: DBus error: The name is not activatable
    2: DBus error: The name is not activatable
    3: The name is not activatable

### How to reproduce (Optional)

1. SSH into a headless Linux box
2. `but config forge auth`

### Expected behavior (Optional)

It would be nice if I could say "It's Fine Actually, just save it to a dotfile"

### Relevant log output (Optional)

```shell

```

slarse:
Hi @anaisbetts,

Thanks for raising this. I've found that this is slightly broader, as in sandboxes with minimal runtimes libdbus is often not available. This makes `but` unable to execute at all (linker fails), even if it's not storing any secrets.

I'm currently trying to make `but` execute in more and more minimal environments, so I'll be looking at solving this within a few weeks. I both want to make the dependency on libdbus itself optional, as well as the presence of a secret service.

slarse:
@anaisbetts We've added the `BUT_CREDENTIALS_DIRECTORY` environment variable to override the keychain storage and simply store credentials as unencrypted plaintext files. The only protection on the directory/files is ownership and relatively stringent permission checks, I really can't think of anything else to do there.

This will not kick in automatically - if the user wants to store credentials in unencrypted plaintext they have to make that choice explicitly.

PR #16006, opened 2026-09-22
deps(linux): replace libdbus dependency with zbus

slarse:
This is Linux-only migration from the `libdbus`-reliant backend to a `zbus` one.

For context, `zbus` is a pure-rust implementation of an API for interacting with D-Bus, which in turn is a Linux IPC thingamajigg. We currently exclusively use D-Bus for secrets storage.

Currently, if a machine lacks libdbus (a shared library), the dynamic linker vomits and we can't even start `but`. By migrating to `zbus`, we gain a lot of flexibility how we deal with credentials storage.

Fixes GB-2058, relates to #15870


crates/but-secret/tests/linux_keyring.rs
```diff
@@ -0,0 +1,84 @@
...
+    round_trip("sync")?;
+    tokio::runtime::Builder::new_current_thread()
+        .build()?
+        .block_on(async { round_trip("current-thread") })?;
+    tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(2)
+        .build()?
+        .block_on(async { round_trip("multi-thread") })?;
+    Ok(())
+}
```
slarse, line 84: The value of keeping these tests around is a bit dubious, as they never execute in CI. They "work on my machine" currently, and I think I'll want them for at least a little bit. But the risk that they break without us knowing is high, and setting up automated runs of them is a bit too much effort.

krlvi (approved):

PR #16008, opened 2026-09-22
Allow naming branches created when committing with `--above/--below`

davidpdrsn:
Previously you couldn't name branches created with `but commit --above/--below TARGET`. This PR adds support for that with `but commit --above/--below TARGET --branch NAME`.

It does somewhat muddy what `--branch` does but having lived with the new commands for a while I think its fine. I think the intention of `--above foo --branch bar` is pretty clear.


slarse (approved):
Looks fine!

